PR #2, opened 2026-03-22
feat: add get_image tool with HTTP-based image access for remote ComfyUI

freeyoung:
* Add fetchImage/uploadImageHttp to client layer for pure HTTP image transfer.
* New get_image tool fetches via /view API and saves to current workspace.
* Rewrite upload_image to try HTTP first with filesystem fallback.
* Expand image filenames in get_history output for easy get_image chaining.

### DIFF
--- a/src/comfyui/client.ts
+++ b/src/comfyui/client.ts
@@ -205,3 +205,51 @@ export async function getHistory(
   const res = await client.fetchApi(path);
   return res.json() as Promise<Record<string, HistoryEntry>>;
 }
+
+/**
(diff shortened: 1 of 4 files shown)


artokun:
> * Add fetchImage/uploadImageHttp to client layer for pure HTTP image transfer.
> 
> * New get_image tool fetches via /view API and saves to current workspace.
> 
> * Rewrite upload_image to try HTTP first with filesystem fallback.
> 
> * Expand image filenames in get_history output for easy get_image chaining.

Thanks! Works great

PR #51, opened 2026-06-23
Added 'scripts' into the 'files' array.

NeoAnthropocene:
Hi @artokun,

Thanks for releasing NPM `comfyui-mcp` package` 0.17.0`! Isaw ytou made some additions on it.
Unfortunately, the `npx -y comfyui-mcp` command is still crashing with an EOF error in MCP clients because the installation is still failing - at least on my side.

While you successfully added `"packs"`, `"model-settings.json"`, and `"model-settings.user.jsonc.example"` to the `"files"` array in your last commit, the `"scripts"` directory was left out.

Because `"scripts"` is missing from the published npm package, the `postinstall` hook (`node scripts/postinstall.mjs`) crashes with a `MODULE_NOT_FOUND` error, which kills the MCP server startup in the background and returns an `EOF` error.

To finally fix this, you just need to add `"scripts" `to the `"files"` array in package.json and publish `0.17.1` (or whatever the package manager writes for it):
```
  "files": [
    "dist",
    "plugin",
    "packs",
    "scripts",
    "model-settings.json",
    "model-settings.user.jsonc.example"
  ],
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "dist",
     "plugin",
     "packs",
+    "scripts",
     "model-settings.json",
     "model-settings.user.jsonc.example"
   ],


artokun:
Shipped in **v0.17.1** — thanks again, @NeoAnthropocene! 🙏

Great catch. To be clear about the impact: this wasn't cosmetic — `v0.17.0`'s `files` allowlist had dropped `scripts/` while `package.json` still ran `postinstall: node scripts/postinstall.mjs`, so **every `npm install` / `npx -y comfyui-mcp` crashed on the missing file**. Restoring `scripts/` (your PR) fixes the install and also ships `sync-agents.mjs` so `npm run sync-agents` works from an install.

To make sure this class of regression can't reach npm again, I added a **pack-and-install smoke test** (`scripts/smoke-install.mjs`): it packs the tarball exactly as `npm publish` would, installs it into a clean project (running the postinstall hook), and boots the entrypoint. It's now a CI job *and* a gate before publish — it would have failed on 0.17.0. Confirmed it passes on the 0.17.1 release.

Appreciate you keeping the packaging honest! 🎉

NeoAnthropocene:
The `smoke-install.mjs` has been a helpful automated fail-safe addition, well thought!

Please keep up this repo alive, it's a hidden gem!

PR #114, opened 2026-07-02
feat: --force-remote flag + keep generations.db out of CWD for remote ComfyUI

andyhite:
## Summary

**`--force-remote` flag**
- Adds a `--force-remote` CLI flag / `COMFYUI_MCP_FORCE_REMOTE` env var that overrides the loopback-hostname heuristic used to decide local vs. remote mode
- Fixes a footgun for port-forwarded remote ComfyUI installs (e.g. dstack on RunPod) that are legitimately reachable at `localhost:8188` — without this flag, such setups are misclassified as local, enabling `COMFYUI_PATH` auto-detection and local-only tools against a filesystem that doesn't actually have ComfyUI on it
- Also threads the flag into the panel orchestrator's `env-capabilities.ts` LOCAL/REMOTE classifier so the agent's system-prompt label stays consistent with `isRemoteMode()`, including for the `connect <url>` runtime path

**Keep `generations.db` out of CWD for remote ComfyUI**
- `GenerationTracker.defaultDbPath()` gated the "good" DB location on `config.comfyuiPath` (the local install dir), which is `undefined` by design in remote/cloud mode — so the DB fell back to `process.cwd()` and littered whatever directory the orchestrator was launched from. This happened **even when remote was detected correctly**; it was a missing data-dir fallback, not a detection bug.
- The fallback now writes to a stable per-user data dir (`~/.comfyui-mcp`, override with `COMFYUI_MCP_DATA_DIR`), scoped per instance via a new `getInstanceSlug()` (`host_port`, or `comfy-cloud` in cloud mode) so multiple remotes keep separate histories instead of sharing one file.
- Propagates the remote signal to the **spawned child MCP**: the orchestrator now sets `COMFYUI_MCP_FORCE_REMOTE=1` in the child's env whenever it isn't handing it a `COMFYUI_PATH`. A loopback-hosted remote (RunPod/dstack port-forward) was otherwise re-detected as LOCAL in the child — the second path by which the DB ended up in CWD or a stray local install.

## Test plan
- [x] `npm run build`
- [x] `npm test` (971/971 passing) — adds `src/__tests__/services/generation-tracker-path.test.ts` (per-instance / data-dir path resolution, explicit-path precedence, local-install unchanged) and `getInstanceSlug()` cases in `src/__tests__/config.test.ts`, alongside the existing `--force-remote` coverage
- [x] README updated for `--force-remote` (flag table entry, deployment-modes table, dstack/RunPod callout)

## Notes / follow-ups
- The new `COMFYUI_MCP_DATA_DIR` env var is **not yet documented** in the README — worth a small docs follow-up.
- Loopback-hosted remotes that all forward to the same local port (e.g. two pods both on `localhost:8188`) share an instance slug, since the URL can't distinguish them. A distinct proxy URL (the normal RunPod case) gets a unique slug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+### Added
+
+- **`--force-remote` flag** (`COMFYUI_MCP_FORCE_REMOTE=1`) — classify a loopback
+  `--comfyui-url` as remote, for dstack/RunPod-style port-forwards where a remote
+  ComfyUI is reachable at `localhost:8188`. The orchestrator forwards it to spawned
+  agents.
+
+### Changed
+
+- **`generations.db` for remote/cloud/undetected targets** now lives under
+  `~/.comfyui-mcp/instances/<host_port>/` (override with `COMFYUI_MCP_DATA_DIR`)
+  instead of the current working directory. Loopback hosts share one
+  `localhost_<port>` slug. Existing remote users' CWD `generations.db` is not
+  migrated — the history restarts empty at the new location.
+
 ## [0.22.0] - 2026-06-29
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -553,12 +553,23 @@ The server auto-detects your ComfyUI installation and port. Override with enviro
 | Mode | Trigger | Local FS / process tools? |
 |------|---------|----------------------------|
 | **Local** | default | yes |
-| **Remote** | `--comfyui-url` / `COMFYUI_URL` points at a non-loopback host | no — server skips `COMFYUI_PATH` auto-detection so stale local installs can't silently absorb uploads |
+| **Remote** | `--comfyui-url` / `COMFYUI_URL` points at a non-loopback host, or `--force-remote` is set | no — server skips `COMFYUI_PATH` auto-detection so stale local installs can't silently absorb uploads |
 | **Cloud** | `COMFYUI_API_KEY` is set (targets [Comfy Cloud](https://cloud.comfy.org)) | no — HTTP primitives route via `cloud.comfy.org` over `X-API-Key`; WebSocket and local-only tools throw `CLOUD_UNSUPPORTED` |
+
+Some setups (e.g. [dstack](https://dstack.ai) driving ComfyUI on RunPod) port-forward
+a remote ComfyUI back to `localhost:8188`, so the loopback check above gets it
+wrong — the install isn't actually local. Pass `--force-remote` (or set
+`COMFYUI_MCP_FORCE_REMOTE=1`) alongside `--comfyui-url`/`COMFYUI_URL` to force
+remote mode regardless of hostname:
+
+```bash
+npx -y comfyui-mcp@latest --comfyui-url http://localhost:8188 --force-remote
+```
 
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `COMFYUI_URL` | | Full ComfyUI URL, e.g. `https://comfy.example.com:8443` — overrides `COMFYUI_HOST`/`PORT`/`SSL` and skips auto-detection. A **path prefix is preserved** (e.g. `https://host/comfyapi`) for reverse-proxied instances. Non-loopback hosts opt into **remote mode**. |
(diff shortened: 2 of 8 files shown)


artokun (changes requested):
Thanks @andyhite — both halves of this are genuinely wanted. The dstack/RunPod port-forward misclassification is real, the \`--force-remote\` flag in \`config.ts\` is clean and well-tested, and the \`generations.db\` relocation (\`~/.comfyui-mcp/instances/<host_port>/\` + \`COMFYUI_MCP_DATA_DIR\`) is exactly the right shape. One change needs rework before this lands, and you'll want a rebase first.

## 1. Please rebase onto latest `main`

`v0.24.0` just landed with a large change to the same orchestrator region (any-LLM backend support). I ran a merge simulation — **your branch rebases cleanly, no conflicts** (nearest hunks are ~29 lines apart). But the rework below matters *more* now, because the two spawn sites you touched serve the new Ollama/OpenAI-compatible backend too.

## 2. The auto-`FORCE_REMOTE` in the orchestrator regresses default local panel users

In `src/orchestrator/index.ts`, both spawn sites now do:

```ts
...(comfyuiPath ? { COMFYUI_PATH: comfyuiPath } : { COMFYUI_MCP_FORCE_REMOTE: "1" }),
```

The catch: the orchestrator's `comfyuiPath` comes **only** from `process.env.COMFYUI_PATH` — it does no auto-detection — and the panel launcher never sets it. Pre-PR, a default local panel user (no `COMFYUI_PATH` env, ComfyUI on `127.0.0.1:8188`) still got full local tools because the **child's own** `resolveComfyUIPath()` auto-detected the install. Post-PR that same user is silently forced remote: no `download_model`, no `apply_manifest` installer packs, no local model scans. That's the most common panel configuration.

And the trade isn't needed — your DB fix in this same PR already covers the port-forward case's worst symptom (CWD littering), and the *explicit* `--force-remote` flag covers the rest.

**Suggested rework** — only forward force-remote when the user opted in or the URL is genuinely non-loopback (apply at both spawn sites, ~L631 and ~L763):

```ts
...(comfyuiPath
  ? { COMFYUI_PATH: comfyuiPath }
  : isForceRemoteFlagSet() || !isLoopbackHost(new URL(comfyuiUrl).hostname)
    ? { COMFYUI_MCP_FORCE_REMOTE: "1" }
    : {}),
```

(`isForceRemoteFlagSet()` is yours; for the loopback check, export the existing loopback-hostname predicate from `config.ts` rather than duplicating the list.) This also fixes a real gap in the current diff: `--force-remote` passed as an **argv flag** to the orchestrator never reaches the child at all — only the env-var form inherits. With this rework the flag forwards explicitly.

## 3. Your setup, spoon-fed (works on `v0.24.0` + this PR rebased)

For the dstack/RunPod port-forward where ComfyUI is remote but reachable at `localhost:8188`:

```bash
# Panel / orchestrator (agent on your laptop, ComfyUI on the pod):
npx -y comfyui-mcp@latest connect http://localhost:8188 --force-remote

# Plain MCP server (Claude Code / Cursor config):
{ "command": "npx", "args": ["-y", "comfyui-mcp", "--force-remote"],
  "env": { "COMFYUI_URL": "http://localhost:8188" } }

# env-var form, if you prefer:
COMFYUI_MCP_FORCE_REMOTE=1 COMFYUI_URL=http://localhost:8188 npx -y comfyui-mcp
```

That gets you correctly-classified remote mode (no stale-install upload absorption, no local-only tools), and your DB fix keeps `generations.db` in `~/.comfyui-mcp/instances/localhost_8188/`.

## Minor (non-blocking)

- `localhost` vs `127.0.0.1` produce different instance slugs (`localhost_8188` ≠ `127.0.0.1_8188`) — consider normalizing loopback hosts in `getInstanceSlug()`.
- Existing remote users' CWD `generations.db` history is silently orphaned by the move — worth a changelog line.
- One new test writes into the real `~/.comfyui-mcp` (cleaned up, but it touches the developer's actual home dir) — a tmp-dir + `COMFYUI_MCP_DATA_DIR` override would isolate it.
- `COMFYUI_MCP_DATA_DIR` README entry (you flagged it) — happy to take that as a follow-up if you'd rather keep this PR tight.

Rework the spawn-site conditional + rebase and this is good to go. 🙏

andyhite:
@artokun all feedback items addressed, ready for review again 👍

artokun (approved):
Rework is exactly right — default loopback panel users keep child auto-detection, opt-in/non-loopback forwards cleanly, and the argv-flag gap is closed. Thanks for turning the minors around too (test isolation, DATA_DIR docs, slug normalization, changelog). Landing it — this ships in the next release. 🚀

artokun:
out in `comfyui-mcp@0.24.1` 🚀 — your dstack setup is now just:

```bash
npx -y comfyui-mcp@latest connect http://localhost:8188 --force-remote
```

thanks for the clean turnaround on the rework 🙏

PR #138, opened 2026-07-07
fix(runpod): aria2 download sidecar — Manager's built-in downloader ran at <1-4 MB/s

artokun:
## Symptom (live pod b0ndoon4j7tcto, image 1.6, 2026-07-07)

"The pod cannot download models." Reality: a 12.9 GB `krea2_turbo_fp8.safetensors` install had been running for 40+ minutes at **792 kB/s – 3.5 MB/s** with 2 more tasks wedged behind it (Manager's install queue is serial), so every subsequent install looked dead.

## Root cause

Measured on the affected pod (via its Jupyter kernel):

| Path | Speed |
|---|---|
| Manager's built-in download of the file | **0.8–3.5 MB/s** |
| `curl` of the **same HF file**, same pod | **14.7 MB/s** |
| `curl` of another HF repo | **83 MB/s** |
| `curl` → `/workspace` (write path) | **62 MB/s** |
| `dd` 4 MB blocks → `/workspace` | **309 MB/s** |

Not HuggingFace, not the volume: ComfyUI-Manager's fallback downloader (`manager_downloader.py` → torchvision `download_url`, single stream, tiny chunks) collapses against the MooseFS network volume.

## Fix

Manager **natively** hands model downloads to an aria2 RPC daemon when `COMFYUI_MANAGER_ARIA2_SERVER` is set (multi-connection segmented fetch). This PR bakes that sidecar into the image:

- **Dockerfile**: `apt aria2` + venv `aria2p` as a required late layer; integrity gate asserts both. (Manager `import aria2p`s at startup whenever the env var is set — a missing package would crash ComfyUI at boot, hence hard-required and double-guarded.)
- **post_start.sh §4.4**: loopback, secret-gated `aria2c` RPC daemon (`--split=16 --max-connection-per-server=16 --min-split-size=8M`, `--file-allocation=none` for MooseFS, `--allow-overwrite`/`--auto-file-renaming=false` so a partial file from the old downloader can't cause a `.1` rename mismatch). Env exported **only** when the daemon started and `aria2p` imports; `ARIA2_DISABLE=1` opts out.
- **test_image.sh**: static aria2c/aria2p assertions + boot-1 assertion that the sidecar is actually running (a silent fallback to the slow path is exactly the bug).
- **README**: env-var table additions (incl. the already-honored `HF_TOKEN`/`HUGGINGFACE_TOKEN` alias) and a "Fast model downloads" section with the field numbers.

Manager's aria2 dir handling was verified against its source: it passes the absolute `/workspace/models/<cat>` dir + filename straight to the daemon, which runs in the same container — no path mapping needed.

## Rollout

Needs an image release (**1.7**) via `deploy-dockerhub.sh` (runs `test_image.sh` as the gate). Existing pods can't pick this up without the new image — the env must be exported before ComfyUI launches.

### DIFF
--- a/docker/runpod/Dockerfile
+++ b/docker/runpod/Dockerfile
@@ -331,6 +331,15 @@ RUN apt-get update && apt-get install -y --no-install-recommends aria2 \
  && rm -rf /var/lib/apt/lists/* \
  && "${VPIP}" install --no-cache-dir aria2p
 
+# hf_transfer — Rust parallel downloader for the OTHER model-fetch path: aria2
+# only covers Manager's install-model; many custom nodes (Impact, WAS, …) and
+# ComfyUI internals download via huggingface_hub, which stays single-stream
+# unless HF_HUB_ENABLE_HF_TRANSFER=1 (env set in §10) AND the hf_transfer
+# package is importable — the hub RAISES at download time if the flag is set
+# without the package, so they ship together. hf_xet accelerates Xet-backed
+# repos the same way (hub uses it automatically when present).
+RUN "${VPIP}" install --no-cache-dir hf_transfer hf_xet
+
 # -----------------------------------------------------------------------------
 # custom_nodes PERSISTENCE seed. Move the image's baked custom_nodes (the Agent
 # Panel + ComfyUI's builtins) to a SEED dir. At boot, post_start.sh symlinks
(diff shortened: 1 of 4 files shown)


artokun:
Added in `a092cf4`: **hf_transfer + hf_xet** baked with `HF_HUB_ENABLE_HF_TRANSFER=1` in the runtime ENV block. aria2 covers Manager's install-model route only — custom nodes (Impact, WAS, …) and ComfyUI internals that fetch via `huggingface_hub` stay single-stream without this. The flag and package ship together deliberately (`huggingface_hub` raises at download time if the flag is set without the package); the integrity gate and `test_image.sh` both assert `import hf_transfer`.

PR #197, opened 2026-07-12
feat(orchestrator): loopback MCP console + credential slots — ported from MichaelDanCurtis fork

artokun:
Ports MichaelDanCurtis's Stack 4 — the loopback MCP console + credential slots — onto current main.

## What this adds

- **Loopback HTTP MCP console** (`src/orchestrator/panel-console-http.ts`): a small control-plane HTTP server on bridge port + 2 (default 9182), bound strictly to `127.0.0.1`. Serves a landing page, `GET /api/status`, and is advertised to the ComfyUI panel via `console_url` on the bridge `backends` frame.
- **Console token gate**: the orchestrator mints a per-run `randomBytes(24)` token, passes it to the console, and advertises it to the panel as `console_token`. `/api/secrets` and `/credentials` require it (query param or `Authorization: Bearer`).
- **Credential slots** (`src/services/panel-secrets.ts`): `CREDENTIAL_SLOTS` UI facade over the existing allowlisted secret stores, with alias fan-out (e.g. HuggingFace writes both `HF_TOKEN` and `HUGGINGFACE_TOKEN`), `setPanelSecret`, `maskSecret`, and `listPanelSecretsMasked` (masked display only — raw values never leave the store).
- **`GET/POST /api/secrets`**: token-gated masked listing + slot writes, with a hardened `readJsonBody` (settles on oversize/closed bodies instead of hanging).
- **`GET /credentials`**: compact key-entry page served with a `frame-ancestors` CSP limited to `127.0.0.1:8188` / `localhost:8188` so the panel can embed it.
- **`GET /api/lora-preview?id=…`**: serves LoRA catalog preview thumbnails with path-containment checks.

Credit: **Ported from MichaelDanCurtis/comfyui-mcp (commits 1efda5e…a305156) with thanks — authorship preserved via cherry-pick.**

## Conflict resolutions

- `src/orchestrator/index.ts` (1efda5e, 2bc70a2): `pushReadiness` had diverged on main (customEndpointConfigured arg + the llama.cpp live re-probe re-push). Kept main's block and added `console_url` / `console_token` to **both** `bridge.push` frames (including the async llamacpp re-push). Kept main's richer `pathNote` ready-log line, adding the console URL to it.
- `src/services/panel-secrets.ts` (2d4d1da): agent allowlist had diverged — merged main's `COMFYUI_MCP_CUSTOM_API_KEY` with the fork's GLM/Zhipu/Kimi keys.
- `src/orchestrator/panel-console-http.ts` (f8632cc, 293ee99, a305156): the fork's file history interleaves an out-of-scope vault/PhotoMap commit (4848fb7), whose content leaked into the 3-way merges. Kept only this stack's changes each time.

## Dropped

- `docs/design/connections-hub.md` hunk from 5f045b1 (file doesn't exist on main — doc-only noise).
- Vault/PhotoMap console surface from fork commit 4848fb7 (not in this stack): `/api/vault`, `/api/photomap`, `listTrainingPacks`/`photomapHealth` imports, and the landing-page "Vault & PhotoMap" section.

## Added beyond the pick list

- `src/services/lora-catalog.ts` copied verbatim from the fork (its commit 9c6aa7c, LoRA-manager stack): `/api/lora-preview` and its test hard-depend on it. Content-identical to the fork file, so if the LoRA-manager stack lands separately the add/add merge is clean.
- One follow-up commit (821bd72): the ported preview path-containment check compared against `previewsRoot + "/"`, which never matches Windows backslash paths — every preview 403'd. Now uses `path.sep`.

## Security review (port audit)

- ✅ Binds loopback only: `host` defaults to `127.0.0.1` and the orchestrator never overrides it; never exposed off-host.
- ✅ Token is always minted and passed by the orchestrator; `/api/secrets` and `/credentials` reject without it. `/api/lora-preview`, `/api/status`, and the landing page are unauthenticated but loopback-only and read-only (status/thumbnails).
- ⚠️ Two soft spots worth knowing (as ported, not regressions): `tokenOk` returns true when **no** token is configured (dev-open fallback — unreachable via the orchestrator path, which always mints one), and the token is accepted as a `?token=` query param (URL-log exposure on the local machine).

## Tests

- `npx tsc --noEmit`: clean
- `npx vitest run`: **111 files, 1235/1235 passed** (includes the new `panel-console-http`, `console-secrets`, and `panel-secrets-slots` suites)
- No MCP tool descriptions changed → no `docs:gen` output.

### DIFF
--- a/src/__tests__/orchestrator/console-secrets.test.ts
+++ b/src/__tests__/orchestrator/console-secrets.test.ts
@@ -1,0 +1,45 @@
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { randomUUID } from "node:crypto";
+import { startPanelConsoleHttpServer, type PanelConsoleHttpServer } from "../../orchestrator/panel-console-http.js";
+
+const TOKEN = "test-console-token";
+let srv: PanelConsoleHttpServer;
+const base = () => srv.url;
+
+describe("console /api/secrets", () => {
+  beforeEach(async () => {
+    process.env.COMFYUI_MCP_PANEL_SECRETS = join(tmpdir(), `secrets-${randomUUID()}.json`);
+    srv = await startPanelConsoleHttpServer({ port: 0, bridgePort: 9180, comfyuiUrl: "http://127.0.0.1:8188", token: TOKEN });
+  });
+  afterEach(async () => { await srv.stop(); });
+
+  it("401s without the token", async () => {
+    const r = await fetch(`${base()}/api/secrets`);
+    expect(r.status).toBe(401);
+  });
+
+  it("lists masked slots with the token", async () => {
+    const r = await fetch(`${base()}/api/secrets?token=${TOKEN}`);
+    expect(r.status).toBe(200);
+    const body = await r.json();
+    expect(body.slots.find((s: any) => s.id === "openrouter")).toBeTruthy();
+    expect(body.slots.every((s: any) => s.masked === null || typeof s.masked === "string")).toBe(true);
+  });
+
+  it("sets a key and reflects it masked; rejects unknown slot", async () => {
+    const ok = await fetch(`${base()}/api/secrets?token=${TOKEN}`, {
+      method: "POST", headers: { "content-type": "application/json" },
+      body: JSON.stringify({ slot: "civitai", value: "civ_key_123456789" }),
+    });
+    expect(ok.status).toBe(200);
+    expect((await ok.json()).masked).toBe("civ_…789");
+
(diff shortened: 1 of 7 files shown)


artokun:
Post-merge live verification (console started standalone from merged dist on :19790):
- landing 200; /credentials 401 without token / 200 with; /api/secrets 401/200 ✅
- secret round-trip: POST {slot,value} → masked `sk-t…890`, GET never leaks raw value ✅
- loopback-only bind confirmed ✅

One gap found: **no clear/delete path for a set slot** — empty value is rejected ("slot and value required") and there's no clear flag, so a slot can only be overwritten, never removed. Minor (loopback+token surface) but worth a follow-up so users can revoke a key without hand-editing ~/.comfyui-mcp/panel-secrets.json.

Issue #210, opened 2026-07-13
Per-workflow agent session permanently orphaned after panel tab-id scheme migration (v0.7.0+ deterministic ids vs legacy random UUID)


artokun:
### What happened / root cause
An existing per-workflow chat session's orchestrator agent is created bound to a fixed `tabId` at spawn time (see `dist/orchestrator/panel-tools.js` comment: "Each agent gets its own server bound to its tab id ... no tab_id juggling for the model", and `makePanelToolCtx(bridge, tabId, workflowTargets)`). All panel_* tool calls route through `bridge.send(cmd, { tabId: ctx.tabId })`, and `UiBridge.resolveTarget(tabId)` in `dist/services/ui-bridge.js` throws hard when the exact/prefix id is not found among currently connected tabs (no fallback).

The session in question was originally bound to a legacy random-UUID tab id (e.g. `6eccc826-592e-4abb-b280-35434e00ddd1`). Mid-session, `comfyui-mcp-panel` was updated from an older commit to v0.8.0 (fb87531), which had introduced (in v0.7.0, commit range including per-workflow-session work) a NEW deterministic per-workflow tab-id scheme: `tmp:<uuid>` for unsaved/unpersisted workflow tabs, flipping to `wf:<hash>` once the workflow is identified/saved (see `web/js/comfyui-mcp-panel.js` around line 616-670 and the tmp:→wf: flip comment near line 10136).

After a `panel_reload` (frontend scope) — or independently, a ComfyUI restart — the browser tab reconnected using the NEW id scheme (`tmp:7eab...` then `wf:workf...`), which never matches the OLD literal `tabId` the orchestrator agent was permanently bound to. Every subsequent panel_* call fails with:
```
no connected tab with id "6eccc826-592e-4abb-b280-35434e00ddd1". Connected: wf:workf (“image_flux2_fp8”)
```
Critically, `panel_reload` (both orchestrator and frontend scope) is ALSO routed through the same stale-tabId bridge call, so it fails identically — the session cannot self-heal or recover once orphaned. The only recovery path is for the user to close/reopen the chat/tab from outside the broken session.

### Steps to reproduce
1. Open a workflow tab, start an agent chat session (agent binds to tab's current id).
[...]

PR #211, opened 2026-07-14
fix(bridge): route old tab ids to migrated tabs when a socket re-hell…

GOPALGTM:
## Problem

A connected panel tab can reconnect the same WebSocket with a new tab_id
(e.g. random UUID → deterministic "wf:" scheme). Agents already bound to
the old tab_id lost their connection — bridge.send() returned
"no connected tab".

## Fix

- **UiBridge**: per-socket `_tabIdMigration` Map (oldId → newId). When a
  socket re-hellos with a different tab_id, the old id migrates to the new.
- **resolveTarget()**: checks the migration map before failing on unknown id.
- **PanelAgentManager.rebindAgent()**: moves agent session from old tab id
  to new one after migration.
- **orchestrator/index.ts**: calls rebindAgent on hello when tab_id changes.

## Tests

2 new tests in `ui-bridge.test.ts` — single tab migration and two-tab
cross-contamination check. All 23 tests pass.

### DIFF
--- a/src/__tests__/services/ui-bridge.test.ts
+++ b/src/__tests__/services/ui-bridge.test.ts
@@ -362,6 +362,85 @@ describe("UiBridge (multi-tab)", () => {
     a.close();
   });
 
+  it("resolves via tab-id migration when a socket re-hellos under a new scheme (tmp:→wf:)", async () => {
+    // Simulate the bug scenario: a tab first connects with a random-UUID tab id
+    // (the old scheme), an agent binds to it, then the SAME socket re-hellos
+    // under a deterministic tmp:/wf: scheme (the new scheme). bridge.send() with
+    // the OLD id must still resolve to the new connection.
+    const sock = await connectPanel(); // open socket, no hello yet
+    autoReply(sock, "old-tab");
+    await vi.waitFor(() => expect(bridge.connected()).toBe(false));
+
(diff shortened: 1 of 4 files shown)


artokun:
Thank you @GOPALGTM — this was a genuinely good fix for a nasty orphaning bug, and it's now **merged via #212 with your commit cherry-picked intact** (your authorship is preserved in main's history).

Why the indirect route: review on top of your work found three things worth hardening before release, and I couldn't push to your branch —

1. **Migration chains**: the field failure in #210 re-helloed *twice* (`uuid → tmp:<uuid> → wf:<hash>`), and the single-hop map lookup landed on the dead intermediate id. Your map now path-compresses on insert, so every historical id resolves to the live tab in one step (test added for the exact reported chain).
2. **Rebind trigger**: the workflow-*title* heuristic on the orchestrator side could cross-rebind agents between two identically-titled tabs (two 'Unsaved Workflow's are routine). Your same-socket bridge migration was the right authority — the bridge now stamps `migrated_from` on the hello and the orchestrator rebinds from that signal instead.
3. `rebindAgent` now *migrates* the persisted session id rather than clearing it, so the rebound conversation still resumes across a process restart.

The per-socket migration map design and the `rebindAgent` state-carry were spot on — exactly the right architecture. Hope to see more PRs from you. 🙏

PR #222, opened 2026-07-15
fix(converter): serialized-widget nodes (LTXDirector) no longer scramble widget values

artokun:
## Problem

\`convertUiToApi\` maps \`widgets_values\` onto input names positionally in \`object_info\` input order. Nodes with a **custom serialized-widget layout** (\`properties.has_serialized_properties === true\` — WhatDreamsCost's LTXDirector/LTXSequencer, kijai's PromptRelay) pack extra/reordered widgets into \`widgets_values\`, so the first unaccounted slot shifts every later widget by one.

Field impact (live session, 128-node LTX director workflow): LTXDirector resolved with \`frame_rate:"seconds"\`, \`display_mode:768\`, \`custom_width:1344\`, \`divisible_by:18\` — an unrunnable graph from panel_strip_workflow / strip_workflow / convert / any headless enqueue of a UI graph.

## Fix

These nodes also write their authoritative **named** values into \`node.properties\`. After the positional pass, when the node declares \`has_serialized_properties\`, override widget inputs from properties. Surgical by design:
- gated on the flag — normal nodes with stale property copies are untouched (regression-tested)
- matched against \`widgetNames\` — non-input properties (\`cnr_id\`, \`timeline_data\`, UI toggles) cannot leak into the prompt

## Verification

- New regression tests from the real capture's shape: flag-on → authoritative values win; flag-off → positional mapping untouched
- Full suite: 125 files / 1369 tests green
- Real-data check: the actual 191KB director workflow + live object_info now resolves \`frame_rate:24, display_mode:"seconds", 768×1344, divisible_by:32, img_compression:18\` (66 nodes, matching the session's report)
- Live panel test via Chrome: see PR comment

### DIFF
--- a/src/__tests__/services/workflow-converter.test.ts
+++ b/src/__tests__/services/workflow-converter.test.ts
@@ -202,3 +202,117 @@ describe("convertUiToApi — bypass / mute resolution", () => {
     expect(workflow["4"].inputs.images).toEqual(["1", 0]); // resolved through the bus
   });
 });
+
+describe("convertUiToApi — serialized-widget nodes (has_serialized_properties)", () => {
+  // Real-world shape: WhatDreamsCost's LTXDirector packs extra/reordered widgets
+  // into widgets_values (23 slots, timeline JSON included), so positional mapping
+  // shifts every widget after the first unaccounted slot — frame_rate came out
+  // "seconds", display_mode 768, divisible_by 18 (field bug, 2026-07-14). The
+  // node's authoritative named values live in node.properties.
+  const DIRECTOR_INFO = {
+    LTXDirector: {
+      input: {
+        required: {
+          model: ["MODEL"],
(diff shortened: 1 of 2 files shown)


artokun:
**Live E2E through the real browser canvas (Chrome + live panel bridge): PASS**

- `panel_load_workflow path:_stripscratch.json` → 128 nodes on the visible canvas (loaded into its own tab)
- `panel_strip_workflow {}` (no args — live-canvas capture via the new graph_serialize) → 66 nodes
- LTXDirector resolved: `{"frame_rate":24,"display_mode":"seconds","custom_width":768,"custom_height":1344,"divisible_by":32,"img_compression":18}` — all authoritative values, previously scrambled to `"seconds"/768/1344/…/18-shifted`.

Merging; release held per request.

PR #232, opened 2026-07-16
fix(queue-monitor): reconnect watchdog WS after ComfyUI retarget

artokun:
## What

`QueueMonitor` stayed disconnected from ComfyUI after the orchestrator retargets ComfyUI on a local panel `hello` (`127.0.0.1` → `localhost`). Root cause: `start()` early-returned whenever `this.url` was already set, but `stop()` never cleared it. The retarget path calls `stop()` + `start(newUrl)` (`orchestrator/index.ts`), so `start()` saw the stale URL and never reconnected.

## Impact

- `queue_status` broadcast (#229) reported **`connected:false` forever** → the mobile app's live queue bar never appeared.
- The local-Ollama **VRAM pause** (`onRunStart`/`onRunEnd`) silently stopped firing after the first retarget.

Latent since the watchdog landed (`783787c`, 2026-06-27) — **not** a #229 regression; #229's `queue_status` broadcast is what surfaced it. #229's own E2E harness used a direct URL with no retarget, so it passed.

## Fix

- `start()` reconnects on a URL change or after a `stop()` (idempotent only for the same live URL).
- WS handlers guarded with `this.ws !== ws` so the retarget's async close of the *old* socket can't null the freshly-opened one or schedule a spurious reconnect (a race the old early-return masked).

## Verification

- Isolated dist test — `start(127.0.0.1)` → `stop()` → `start(localhost)` → `snapshot().connected === true` (was `false`).
- Live orchestrator — after the real `127.0.0.1`→`localhost` retarget, the bridge now broadcasts `queue_status connected=true` (was `connected=false`).
- `tsc` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,18 @@ All notable changes to this project are documented here. This project adheres to
   passes the `prompt_id` it observed, cancel_job only interrupts a still-matching
   running job, and pending jobs are never touched
 
+#### Fixed
+- **QueueMonitor stayed disconnected after a ComfyUI retarget** — the watchdog
+  WS never reconnected once the orchestrator retargeted ComfyUI (e.g. the
+  `127.0.0.1`→`localhost` swap on a local panel `hello`), because `start()`
+  early-returned on the stale URL after the retarget's `stop()`. That left
+  `queue_status` permanently `connected:false` (so the mobile queue bar never
+  appeared) and silently disabled the local-Ollama VRAM pause. `start()` now
+  reconnects on a URL change or after `stop()`, and the WS handlers are guarded
+  against a superseded socket so the retarget's async close can't null the new
+  connection. Latent since the watchdog landed (2026-06-27); surfaced by the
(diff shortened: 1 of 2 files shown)


artokun:
### codex:rescue review — 1 issue found & fixed

codex confirmed the reconnect logic, the socket-identity guards, and no VRAM-pause regression. It caught one real issue I introduced with the supersession guard:

> `stop()` nulls `this.ws` without clearing `state.connected`; the old socket's guarded `close` handler now returns early before clearing it — so `snapshot().connected` stays stuck `true` after `stop()` (through a retarget gap, or indefinitely if the reconnect fails).

Fixed in `5e52ae1`: `stop()` now clears `state.connected` directly.

Re-verified (isolated dist): `start(127.0.0.1)`→true, `stop()`→**false**, `start(localhost)`→true, `stop()`→false. And live: orchestrator broadcasts `queue_status connected=true` after the real retarget, from the fully-fixed build.

PR #234, opened 2026-07-18
refactor(orchestrator): provider registry for simple OpenAI api-key backends

artokun:
## Why

Adding one OpenAI-compatible **API-key** provider today means editing **~10 scattered sites** that all copy-paste the same identity. The recently-added `moonshot` provider is the poster child — `grep -rn "moonshot" src` on the base branch hits every one:

1. `agent-backend.ts` — `BackendId` union + a `*_CAPABILITIES` const (all just `{...OLLAMA_CAPABILITIES}`)
2. `<name>-backend.ts` — a near-identical `class extends OllamaBackend`
3. `code-provider-auth.ts` — a `resolve*Credentials()` (glm/moonshot identical shape)
4. `orchestrator/index.ts` — model var, `makeBackend` branch, `getProbeBackend` branch, `KNOWN_BACKENDS` Set, keyed-providers boot log, `currentModelFor` branch, connect-ack (`isXx` bool + agentLabel + ready + degraded)
5. `panel-console-http.ts` — a **second** `KNOWN_BACKENDS` list
6. `backend-readiness.ts` — a readiness branch
7. `panel-secrets.ts` — a `CREDENTIAL_SLOTS` entry + `AGENT_SECRET_ENV_ALLOWLIST`

## What

Introduce **one source of truth** — `src/services/openai-provider-registry.ts`, a **pure-data leaf module** (imports nothing from services/orchestrator, so both layers can consume it without a cycle) that describes each simple api-key provider once: `id`, `slotLabel`/`slotHelp`, `envKeys`, `modelEnv`/`defaultModel`, `ackFallbackLabel`, `readyMessage`/`degradedMessage`, `simpleKeyAuth`. Then the scattered lists **derive** from it:

- **Both `KNOWN_BACKENDS` lists** splice in `OPENAI_KEY_PROVIDER_IDS` (non-registry providers untouched).
- **`backendReadiness`** — a single "ready iff one of `envKeys` is set" loop replaces the glm + moonshot branches (bespoke non-registry branches untouched).
- **`CREDENTIAL_SLOTS` + `AGENT_SECRET_ENV_ALLOWLIST`** derive their glm/kimi/moonshot entries.
- **One generic `makeOpenAiKeyBackend` factory** collapses the near-identical `GlmBackend`/`MoonshotBackend` classes across `makeBackend` **and** `getProbeBackend` (they only differed by id/host/apiKey/model). Both class files are **deleted**.
- **keyed-providers boot log, `currentModelFor`, connect-ack ready/degraded text** all read registry fields.
- **`resolveGlmCodeCredentials`/`resolveMoonshotCredentials`** now share one `resolveKeyedCredentials` helper (public signatures, error messages, and `__testing` bases unchanged).

### Before → after (sites touched to add one api-key provider)

| | Before | After |
|---|---|---|
| Sites to edit | ~10 across 7 files | **1 registry entry** (+ a 2-line resolver wrapper for the throw-if-absent credential path) |

## Consolidated vs. deliberately left bespoke

- **Fully consolidated: `glm`, `moonshot`** — structurally identical simple api-key providers (resolve env key → `OllamaBackend` openai). Registry drives every one of their sites.
- **Metadata-only: `kimi`** — its auth code path is special (OAuth **or** `KIMI_API_KEY` dual-auth via `KimiBackend` + `resolveKimiCodeOAuth` + its own readiness branch), so `simpleKeyAuth: false` keeps that hand-written. The registry still unifies kimi's list/slot/allowlist/model/connect-ack metadata.
- **Left bespoke: `openrouter`, `custom`** — OpenAI-compatible too, but they **degrade gracefully with no key** (rather than throwing), carry a **different readiness shape** (`cli` mirrors key/URL presence), and have **up-front connect-ack guards**. Forcing them into the table would be a bad abstraction.
- **Untouched: `claude`, `codex`, `chatgpt`, `gemini`, `grok`, `ollama`, `lmstudio`, `llamacpp`, `copilot`** — CLI/OAuth/SDK/local providers that don't fit one table.

## Behavior-preserving

Pure de-duplication: no provider's env var names, model defaults, base URLs, labels, or user-facing messages change (per-provider ready/degraded text is preserved as registry fields; a plain `OllamaBackend` already reports the same `OLLAMA_CAPABILITIES` the removed `GLM_/MOONSHOT_CAPABILITIES` spreads carried). All public exports the tests import keep working.

## Verify

- `npm run build` — tsc clean
- `npx vitest run` — **128 files, 1443 tests passed** (matches baseline)

## Risks / for the maintainer to review

- The registry lives in `src/services/` (not `orchestrator/`) so `services/panel-secrets` can import it without a layering inversion / import cycle. Named `openai-provider-registry.ts` to avoid colliding with the unrelated `src/experimental/provider-registry.ts`.
- `agentLabel` for a registry provider is `openAiKeyProviderModel(reg) ?? … ?? ackFallbackLabel`; the model helper never returns nullish, so the `??` fallbacks are unreachable — this exactly mirrors the old `glmModel ?? …` (also always truthy). Kept for parity.
- `resolveOpenAiKeyCredentials(id)` in `code-provider-auth` is the one place mapping a `simpleKeyAuth` id → its resolver; a new simple provider needs its 2-line resolver wrapper added there.


artokun:
Rebased onto v0.42.0 main (moonshot commit auto-dropped as upstream) and taken through the adversarial review gate: **CLEAN** — behavior-preservation verified field-by-field against `origin/main` history (env precedence chains, readiness tuple shapes, the #209 ctor-throw handling, kimi's untouched dual-auth, connect-ack strings, both KNOWN_BACKENDS lists), plus rebase-composition checks against the GPT-5.6 / blind-gate / gemini-#242 drift (the blind spawn env rides inside makeHttpBackendMcpServers, so registry-built backends inherit it). Full suite 1529/1529; live smoke shows all 14 providers advertised with sane ready flags and the registry-derived boot log.

PR #237, opened 2026-07-19
feat(training): CLI LoRA trainer P1 — character LoRA on FLUX.1-dev via GPU-Docker train_* tools (E2E-proven)

artokun:
# CLI LoRA trainer — P1: local character-LoRA slice (train_* tools + GPU Docker + E2E-proven)

Plan: `~/.claude/plans/eager-hatching-cerf.md`. Wraps **ostris ai-toolkit's `run.py`** (unchanged, upstream) in a **headless GPU Docker image**; the **LLM drives training through `train_*` MCP tools + a purpose-built skill**. Local-first (RunPod = P4, deferred), narrow first slice: **character LoRA on FLUX.1-dev**.

## What lands in P1

- **`docker/trainer/Dockerfile`** — lean headless image (CUDA 12.8.1 devel, torch 2.9.1/cu128, ai-toolkit pinned via `AI_TOOLKIT_REF`, `ENTRYPOINT ["python","run.py"]`). No Node/UI.
- **`src/services/ai-toolkit.ts`** — container driver: docker/GPU preflight (`trainerDoctor`), `buildTrainerImage`, `startTraining` (`docker run --gpus all` + bind mounts, streamed stdout → progress ticks), `stopTraining` (honest envelope), `containerRunning` liveness probe.
- **`src/services/training-config.ts`** — ai-toolkit YAML generator (schema confirmed by codex review against the official `train_lora_flux_24gb.yaml`).
- **`src/services/training-jobs.ts`** — job registry + lifecycle:
  - Per-job JSON under `~/.comfyui-mcp/training/jobs/`; **refreshed from disk on every read** so the orchestrator's cross-process `call_tool` client (mobile `train_status`) sees live state; throttled progress+log snapshots.
  - Dataset staging: validate-all-then-atomic-swap; refuses to restage a dir an active job trains from.
  - Cancel: pre-persisted state (no finalize race), liveness-verified stop, honest revert-to-running on failure, retry for stale cancelled records, rollback reconcile.
  - Owner-death recovery keyed on the **final save** (periodic checkpoints prove nothing).
  - **Output handoff:** final `.safetensors` → ComfyUI `models/loras/` + LoRA catalog upsert (trigger keyword, FLUX.1-dev base, sample preview, `missing:false`) — skipped honestly if the ComfyUI instance changed mid-run.
- **`src/tools/train.ts`** — `train_list_flows` / `train_prepare_dataset` / `train_start` / `train_status` / `train_cancel` / `train_build_image` / `train_doctor`. Registered as the `training` TOOL_GROUP (appended; order preserved). `train_list_flows` + `train_status` whitelisted for mobile `call_tool` (read-only). `train_cancel` returns `ok:false` when the container isn't confirmed stopped.
- **`plugin/skills/train-character-lora/SKILL.md`** — teaches the LLM the whole flow (dataset guidance, trigger word, defaults, monitoring, failure modes).
- **`package.json`** — ships `docker/` so `train_build_image` works from the npm package.

## E2E proof (RTX 4090, this branch's code)

Real run, not a mock: built `comfyui-mcp-trainer:latest` (ai-toolkit `a0224793`), staged a 6-image character dataset, `train_start` → 200 steps on FLUX.1-dev.

- `train_doctor` all green (docker 29.6.1, `--gpus all` → RTX 4090, image, HF_TOKEN).
- Live `train_status` from a **separate process** reported the running job correctly (cross-process path).
- Job **completed** (final loss 0.196) → `e2e_smoke.safetensors` (172 MB, rank-16 Flux LoRA, `ss_base_model_version: flux.1`, trigger `ohwx`) copied to `models/loras/`; catalog entry `loras-e2e-smoke` + sample preview.
- LoRA **loaded in a live ComfyUI Flux workflow** (`LoraLoaderModelOnly`) and steered the output — A/B vs strength-0 control (control render confirms fp8 softness is settings, not the LoRA).

The E2E caught two real bugs, both fixed in this branch: `/dataset` must be mounted **rw** (ai-toolkit writes `.aitk_size.json` into it) and the progress parser needs a **loss reading** to trust step counts (dataset bars like `6/6` masqueraded as training steps).

## Review & tests

- **6 rounds of codex review, 19 findings addressed** (cross-process registry freshness/liveness, cancel/finalize races in both directions, recovery misclassification, mid-run retarget guard, cancel retry, terminal progress linger, cancel envelope honesty).
- 51 trainer tests (config gen, progress parse, dataset staging, registry/cancel/recovery lifecycle). `tsc` clean. Full suite: 1504 passing, 1 **pre-existing unrelated** `node-dev` failure (verified it fails without this branch's changes too).

## Deferred (per plan)

- P2: broaden flows/models (Z-Image Turbo, style/slider/edit).
- P3: mobile Training tab + panel wiring to `train_*` over `call_tool`.
- P4: RunPod — same image on a pod, `src/services/runpod.ts`. **No cloud spend yet.**

Note: the E2E artifacts remain on the rig — `models/loras/e2e_smoke.safetensors` (+ catalog entry), the training dataset/jobs under `~/.comfyui-mcp/training/`, and Flux support models downloaded for the validation render (`flux1-dev-fp8`, `clip_l`, `t5xxl_fp8`, `ae`). Delete `e2e_smoke*` freely; the rest is generally useful.

### DIFF
--- a/docker/trainer/Dockerfile
+++ b/docker/trainer/Dockerfile
@@ -12,14 +12,15 @@
 # Run (bind config + dataset + output + HF cache; --gpus all is required):
 #   docker run --rm --gpus all \
 #     -v /path/config.yml:/config.yml:ro \
-#     -v /path/dataset:/dataset:ro \
+#     -v /path/dataset:/dataset \
 #     -v /path/output:/output \
 #     -v $HOME/.cache/huggingface:/root/.cache/huggingface \
 #     -e HF_TOKEN=$HF_TOKEN \
 #     comfyui-mcp-trainer:latest /config.yml
 #
 # The config's training_folder must be /output and datasets[].folder_path /dataset
-# (the driver in src/services/ai-toolkit.ts wires these paths).
+# (the driver in src/services/ai-toolkit.ts wires these paths). /dataset must be
+# writable: ai-toolkit caches (.aitk_size.json, latents) inside it.
 
 FROM nvidia/cuda:12.8.1-devel-ubuntu24.04
 
--- a/docker/trainer/Dockerfile
+++ b/docker/trainer/Dockerfile
@@ -1,0 +1,64 @@
+# Headless GPU LoRA-training image — runs ostris ai-toolkit's `run.py` directly.
+#
+# Derived from ai-toolkit's own Dockerfile (same CUDA base, torch, arch list) but
+# STRIPPED of the Next.js UI: we drive training via the CLI + a generated YAML
+# config, so no Node/UI layers. One image, two future targets — local `docker run`
+# now, the same image on a RunPod pod later.
+#
+# Build (from repo root):
+#   docker build -t comfyui-mcp-trainer:latest \
+#     --build-arg AI_TOOLKIT_REF=<commit-or-tag> docker/trainer
+#
+# Run (bind config + dataset + output + HF cache; --gpus all is required):
+#   docker run --rm --gpus all \
+#     -v /path/config.yml:/config.yml:ro \
+#     -v /path/dataset:/dataset:ro \
+#     -v /path/output:/output \
+#     -v $HOME/.cache/huggingface:/root/.cache/huggingface \
+#     -e HF_TOKEN=$HF_TOKEN \
+#     comfyui-mcp-trainer:latest /config.yml
+#
+# The config's training_folder must be /output and datasets[].folder_path /dataset
+# (the driver in src/services/ai-toolkit.ts wires these paths).
+
+FROM nvidia/cuda:12.8.1-devel-ubuntu24.04
+
+ENV DEBIAN_FRONTEND=noninteractive \
+    PYTHONUNBUFFERED=1 \
+    PYTHONUTF8=1 \
+    PIP_NO_CACHE_DIR=1 \
+    HF_HUB_ENABLE_HF_TRANSFER=1 \
+    # Ampere (30xx) through Blackwell (50xx/GB) — 8.9 = RTX 4090.
+    TORCH_CUDA_ARCH_LIST="8.0 8.6 8.9 9.0 10.0 12.0"
+
+# 1) System packages (heaviest, most cache-stable layer).
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        python3 python3-venv python3-pip python3-dev \
+        git git-lfs cmake build-essential ffmpeg libgl1 libglib2.0-0 ca-certificates \
+    && rm -rf /var/lib/apt/lists/* \
+    && ln -sf /usr/bin/python3 /usr/local/bin/python
+
+# 2) PyTorch (cu128) — pinned to ai-toolkit's tested versions, before source so a
+#    code bump doesn't re-download torch.
+RUN pip install --break-system-packages --no-cache-dir \
+        torch==2.9.1 torchvision==0.24.1 torchaudio==2.9.1 \
+        --index-url https://download.pytorch.org/whl/cu128 \
+    && pip install --break-system-packages --no-cache-dir hf_transfer
+
+# 3) ai-toolkit source (pin AI_TOOLKIT_REF for reproducible builds) + its
+#    requirements. Requirements are installed from the cloned file so they layer
+#    after torch.
+ARG AI_TOOLKIT_REPO=https://github.com/ostris/ai-toolkit.git
+ARG AI_TOOLKIT_REF=main
+WORKDIR /opt
+RUN git clone --recurse-submodules "${AI_TOOLKIT_REPO}" ai-toolkit \
+    && cd ai-toolkit \
+    && git checkout "${AI_TOOLKIT_REF}" \
+    && git submodule update --init --recursive \
+    && pip install --break-system-packages --no-cache-dir -r requirements.txt
+
+WORKDIR /opt/ai-toolkit
+
+# Fail fast + loud if the container is started without a GPU.
+# run.py takes the config path as its single positional arg.
+ENTRYPOINT ["python", "run.py"]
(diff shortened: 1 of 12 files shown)


artokun:
## Independent codex:rescue review (Opus, post-handoff check)

Great work overall — E2E-proven, well-architected (cross-process persistence + recovery, retarget-safe handoff), 51 tests green. An independent adversarial pass surfaced issues the earlier rounds missed. Triaged:

### Pre-merge (recommend fixing before merge)
1. **`train_status` is not truly read-only** (and it's mobile-whitelisted as read-only). `train_status → getJob/listJobs → refreshRegistry` treats a running record with no *local* handle + stopped container as "owner-dead" and then **does the handoff** (copy LoRA + `lora_catalog.upsert` + persist `completed`). A healthy owner is in exactly that state briefly between Docker exit and `finalizeJob`, so a mobile poll can (a) trigger side-effects from a "read" and (b) race the real owner's finalization. → Make `train_status` genuinely side-effect-free; gate recovery/handoff to an explicit path or a single owner, guarded by a lock/CAS. (`training-jobs.ts:204-256`, `train.ts:149-163`, `orchestrator/index.ts:647`)
2. **Cancel races finalization → LoRA still published.** `finalizeJob` checks `diskCancelled()` only before handoff; a cancel persisted after that check still proceeds to handoff + `completed`. No lock spans check→handoff. (`training-jobs.ts:513-558`, `744-768`)
3. **Dockerfile default `AI_TOOLKIT_REF=main` is not pinned**, but `train_build_image` claims it is and omits the arg. A fresh npm-installed build can differ from the E2E-proven ai-toolkit commit. → Pin the default to the E2E'd commit (a0224793). (`Dockerfile:53`, `train.ts:193-207`)

### Hardening (fine as follow-ups)
4. `train_start` mounts `datasetPath` **rw** without constraining it under `datasetsRoot()` — any accessible dir w/ an image can be exposed to container writes. (`training-jobs.ts:574-577`, `ai-toolkit.ts:183`)
5. Registry `persist()` is non-atomic + swallows errors; failed persist doesn't block launch → a container can become unfindable/uncancellable after owner death. (`training-jobs.ts:151-157`, `620`)
6. Dataset replacement is check-then-replace (no lock); shared `.staging-${pid}` path collides on concurrent same-name prepares. (`training-jobs.ts:311-350`)
7. `train.log` is unbounded (no rotation). (`training-jobs.ts:376-383`)

### Plausible (verify)
- Retarget detection compares host:port slug only; if a supported retarget changes the ComfyUI models root while keeping host+port, the LoRA copies to the old dir but catalogs against the new target. Confirm every retarget path changes host/port.

Docker args are built as an array (spawn, not shell) — no shell-injection surface found. Nice.

artokun:
## Review findings addressed — all fixed, pushed as `ba690e2`

Thanks for the adversarial pass — finding #1 especially was a genuine architectural flaw (my recovery-in-read-path design). Everything is resolved, each with regression tests; the fix itself then went through its own adversarial iteration (see "follow-up" below). Final state: **53 trainer tests green, full suite 1526✓, tsc clean, final codex review: no actionable defects.**

### Pre-merge — all fixed

**1. `train_status` not read-only** → redesigned. Recovery/handoff no longer happens on reads for healthy-owner states. A job now records `ownerPid`; handoff/fail requires the owner **provably dead** (pid liveness) **or** its liveness lease expired (10-min `updatedAt` heartbeat, refreshed by progress AND log activity). A status poll observing a container-gone job with a healthy owner just reports the record — the owner persists its terminal state itself. Recovery runs under a per-job CAS lockfile, re-reads the record inside, and only recovers on proof (final save, not checkpoints). `training-jobs.ts:230-290, 372-400`

**2. Cancel races finalization → LoRA published** → serialized by the same per-job lock. Cancel persists `cancelled` **under the lock**; finalize's cancel-check → handoff → persist all happen inside it; in-flight cancels register in `pendingCancels` (concurrent cancels **coalesce** — the second joins the first; finalize can't out-race an in-flight cancel; a foreign cancel adopted mid-flight **continues to docker stop**). `training-jobs.ts:1040-1130, 1200-1330`

**3. `AI_TOOLKIT_REF=main` unpinned** → pinned to the E2E'd full commit `a0224793cef5d5073c8ed0b8cdb838a84fd1cba0` (kept your comment from `2e53287`; resolved the rebase to the full SHA). `Dockerfile:61`

### Hardening — all fixed

**4. `datasetPath` rw-mount unconstrained** → must be staged under `datasetsRoot()` (error points at `train_prepare_dataset`). Containment uses realpaths (symlink/junction escape rejected — junction test skips cleanly where symlinks are unavailable) and volume-probed case semantics (no platform assumptions; basename-only probe so mounted volumes classify correctly).

**5. Non-atomic swallowing persist** → atomic unique-tmp + rename, returns success; a failed first persist **aborts the launch** (no untracked container).

**6. Dataset check-then-replace race** → serialized in-process (`preparingNames`, set before first await) **and** cross-process (`.prep-lock` file, token-guarded like the job lock, 12h cap so slow legitimate copies aren't broken into; staging suffix randomized).

**7. Unbounded `train.log`** → rotates at 5MB, per-file cadence (concurrent jobs can't starve each other).

### Verify-note (retarget slug = host:port)

The handoff pins `lorasDir` **at job start** (persisted on the record), so the copy always lands in the original instance's models dir regardless of later retargets. The catalog upsert is skipped (with a warning) when the instance slug changed. A same-host:port models-root change can't desync copy vs catalog: the catalog file is slug-keyed (same slug = same file), the dir is start-pinned.

### Follow-up: the fix went through its own adversarial iteration

The lock machinery itself was hardened across several review rounds: locks carry ownership tokens (pid+timestamp+seq, content-matched takeover — rename/check-then-act was replaced by a claim-serialized, token-verified protocol), release is token-checked and claim-serialized, empty locks + expired claims are reclaimed, lock-creation non-contention errors (EACCES/ENOSPC) surface immediately instead of burning budgets, live progress snapshots persist under the lock, acquirers honor active takeover claims, claim waits honor the caller's budget, finalized-record adoption is full-object. Final codex review of the resulting diff: **no actionable defects**.

Issue #241, opened 2026-07-20
Codex pin hides GPT-5.6; newer Desktop cache can crash app-server


JusticeWay:
## Summary

There are now **three related compatibility gaps** in the embedded Codex backend:

1. `comfyui-mcp@0.40.0` bundles `@openai/codex@~0.141.0`, so the Agent Panel does not receive the account-available GPT-5.6 models:
   - `gpt-5.6-sol`
   - `gpt-5.6-terra`
   - `gpt-5.6-luna`
2. The backend and companion panel filter out the new reasoning levels `max` and `ultra`.
3. Updating only to stable Codex `0.144.6` is **not a complete workaround** when Codex Desktop shares a newer `models_cache.json`: the model handshake can succeed, but the first real turn may make the app-server exit while renewing the cache TTL.

The currently verified end-to-end workaround is `@openai/codex@0.145.0-alpha.24`, with `max` / `ultra` enabled. This returned the expected live model catalog and completed a real `gpt-5.6-sol` turn through the Agent Panel bridge.

## Environment

- Windows 11
- Node.js 24.18.0
- `comfyui-mcp@0.40.0`
- bundled `@openai/codex@0.141.0`
- Codex Desktop cache with `client_version: 0.145.0`
- ChatGPT-authenticated Codex account with access to GPT-5.6 Sol/Terra/Luna
- verified working end to end: `@openai/codex@0.145.0-alpha.24`

The current `main` branch still declares:

```json
"@openai/codex": "~0.141.0"
```

## Reproduction A: GPT-5.6 models are missing

1. Start the orchestrator:

   ```shell
   npx -y comfyui-mcp@0.40.0 connect
   ```

2. Open ComfyUI Agent Panel and select the ChatGPT/Codex backend.
3. Open the model picker or inspect the bridge `models` handshake.

### Actual result

Sol, Terra, and Luna are absent because the orchestrator starts the bundled Codex 0.141.x app-server.

### Expected result

The panel should preserve every model returned by the authenticated account's current Codex `model/list`.

## Reproduction B: 0.144.6 can crash on a current Desktop cache

1. Let current Codex Desktop create/update its shared `models_cache.json`.
2. Override the panel runtime to `@openai/codex@0.144.6`.
3. Select `gpt-5.6-sol` and send a real message.

The model picker/handshake may already show the GPT-5.6 family, but the turn fails with:

```text
codex app-server exited unexpectedly (exit 1)
[...]

PR #253, opened 2026-07-21
fix(orchestrator): a per-tab start failure can never self-exit the orchestrator

### LINKED ISSUE #250: Invalid API key on an OpenAI-dialect provider kills the WHOLE orchestrator at agent start (all tabs)
Found while live-testing the provider registry (#234) — **pre-existing** behavior, present since at least v0.38.0 (not a registry regression; the refactor preserves it faithfully).

## Repro
1. Set a syntactically-plausible but invalid key, e.g. `MOONSHOT_API_KEY=sk-fake` (readiness correctly flips the provider ready — key presence is all it can check).
2. Panel/mobile tab selects moonshot (or glm/custom/openrouter with a bad key) and sends any message.
3. `OllamaBackend.prepare()` throws (`endpoint ... rejected the key (http 401)`) → PanelAgent "failed to start" → the orchestrator **self-exits**: `self-exit (tab ... agent failed to start ...) — closing the bridge so a fresh orchestrator can take over.`

## Impact
Every OTHER tab on the orchestrator dies too — including healthy sessions on different providers (observed live: a gpt-5.6-sol session was killed by another tab's bad moonshot key). A recoverable, per-tab configuration error escalates to a process-wide restart.

## Expected
A 401/unreachable at agent start should degrade THAT tab only — same UX as the keyless path after #209 (degraded ack + a visible "check your key in the API Keys card" say), leaving other tabs untouched. The self-exit-for-takeover path should be reserved for genuinely unrecoverable orchestrator states.

## Notes
- The #209 fix covers the ctor/keyless case (ValidationError → brokenBackend → degraded). This is the keyed-but-invalid case one step later, in prepare()/first-connect.
- Registry rows/readiness/degraded-ack all behaved correctly in the same test — this is purely the start-failure escalation policy.

artokun:
## What / why

Fixes #250.

An invalid API key on an OpenAI-dialect provider (moonshot/glm/custom/openrouter) killed the WHOLE orchestrator at agent start, taking down every other tab.

**Repro (from the issue):** set `MOONSHOT_API_KEY=sk-fake` (readiness flips ready — key *presence* is all it can check), select moonshot on a tab, send any message. `OllamaBackend.prepare()` throws (`endpoint ... rejected the key (http 401)`) → PanelAgent "failed to start" → `onAgentFatal` → `requestSelfExit` ("closing the bridge so a fresh orchestrator can take over") → every other tab dies, including healthy sessions on other providers.

This is the same class of failure as the keyless ctor path fixed in #209 (ValidationError → `brokenBackend` → degraded ack + visible say), just one step later — in `prepare()`/first-connect instead of the constructor.

## How the fix routes start-failures

- **`src/orchestrator/panel-agent.ts`** — the manager's `settle()` on a start reject no longer fires `onAgentFatal`. It routes to a new optional **`onStartFailure(tabId, message)`** callback (falling back to a per-tab `onSay` notice when unwired). The dead agent slot was already cleared, so the tab stays recoverable: fix the key → re-send or Disconnect → Connect spawns a fresh agent.
- **`src/orchestrator/index.ts`** — wires `onStartFailure`: pushes an honest say naming the provider (`⚠️ The <backend> agent could not start: <error> — Check your <provider> API key in the API Keys card (<ENV_VAR>), then Disconnect → Connect to retry.`, registry-driven for glm/kimi/moonshot, bespoke copy for openrouter/custom) plus a `degraded` ack to THAT panel tab only, and logs it as tab-local. **No `requestSelfExit`.**

## Self-exit deliberately preserved

Checked every route into `requestSelfExit` (it has exactly one caller: `onAgentFatal`):
- **Kept fatal:** the bounded self-restart give-up (`agent session kept dropping (self-restart gave up)`) — the "bridge open but no panel agent responded" wedge that the self-exit was built for.
- Unrelated paths (uncaughtException handler, connect-time degraded acks) untouched.

## Test evidence

New: `src/__tests__/orchestrator/start-failure-per-tab.test.ts` (follows the restart-coalesce harness pattern) —
1. a backend whose `prepare()` rejects with a fake 401 fires `onStartFailure` with the honest error, does **not** fire `onAgentFatal`, and a second tab's healthy agent keeps processing turns;
2. the `onSay` fallback (manager without `onStartFailure`) is still per-tab, still no fatal;
3. **recovery:** after the factory starts returning a healthy backend ("key fixed"), a new send on the SAME tab spawns a working agent;
4. **preserved:** the instantly-dropping-session backend still trips the give-up threshold → `onAgentFatal`.

Gates:
- `npx tsc --noEmit` — clean
- `npx vitest run` — **135 files / 1541 tests passed** (incl. the 4 new)

### DIFF
--- a/src/__tests__/orchestrator/start-failure-per-tab.test.ts
+++ b/src/__tests__/orchestrator/start-failure-per-tab.test.ts
@@ -1,0 +1,226 @@
+// Issue #250 — a backend whose prepare() rejects at agent start (e.g. an
+// OpenAI-dialect provider 401ing on an invalid API key) must degrade THAT TAB
+// ONLY. Before the fix, the manager routed every start failure to onAgentFatal,
+// which the orchestrator wires to a process-wide self-exit ("closing the bridge
+// so a fresh orchestrator can take over") — killing every OTHER tab too,
+// including healthy sessions on different providers.
+//
+// Contract under test:
+//  (a) a start failure must NOT fire onAgentFatal (no self-exit) and must not
+//      disturb another tab's live agent;
+//  (b) it produces the per-tab degraded notification (onStartFailure when
+//      wired, else the onSay fallback);
+//  (c) the tab's slot stays RECOVERABLE: after the user fixes the key, a new
+//      send on the SAME tab spawns a fresh agent that works;
+//  (d) the genuinely process-level path is preserved: the bounded self-restart
+//      give-up ("session keeps ending immediately") still fires onAgentFatal.
+
+import { describe, expect, it, beforeAll } from "vitest";
+import type {
+  AgentBackend,
+  AgentEvent,
+  BackendStartOptions,
+  ModelChoice,
+} from "../../orchestrator/agent-backend.js";
+import { CLAUDE_CAPABILITIES } from "../../orchestrator/agent-backend.js";
+
+let PanelAgentManager: typeof import("../../orchestrator/panel-agent.js").PanelAgentManager;
+
+beforeAll(async () => {
+  ({ PanelAgentManager } = await import("../../orchestrator/panel-agent.js"));
+});
+
+/** A backend whose prepare() rejects — the #250 repro: OllamaBackend.prepare()
+ *  throwing `endpoint ... rejected the key (http 401)` on an invalid key. */
+class Rejecting401Backend implements AgentBackend {
+  readonly id = "moonshot" as const;
+  readonly capabilities = CLAUDE_CAPABILITIES;
+  prepareCalls = 0;
+
+  async prepare(): Promise<void> {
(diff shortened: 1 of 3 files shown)


artokun:
Review gate: the Codex CLI review task wedged (no session writes for 26+ min — same failure mode as #241-era tasks), so a Claude adversarial review substituted, per established precedent. Verdict was FIX FIRST with one MAJOR (stuck 120s thinking spinner on the degraded tab — fixed in edb9df2 by pushing turn:done from onStartFailure) plus a cheap-insurance backend.close guarantee (also edb9df2). Non-blocking findings filed as #255 #256 #257. Gates after fixes: tsc clean, 1541/1541 tests pass.

PR #260, opened 2026-07-21
fix(orchestrator): start-failure follow-ups — rebind-safe settle, held mail, real spinner clear

### LINKED ISSUE #255: rebind race can strand a prepare-failing agent under its new key (settle() guard early-return)
Pre-existing, surfaced by the #253 adversarial review (finding 2). settle()'s guard (agents.get(tabId) !== agent) early-returns when rebindAgent() moved a still-starting agent to a new key; if prepare() then rejects, the dead agent stays mapped under the new key — hasLiveAgent true, queue never drained, no warning, tab wedged until reset. Not a #253 regression (the old fatal also skipped this window), but now that start failures are survivable it deserves a fix: settle should locate the agent by identity across keys, or rebind should re-arm settle for the new key.

Also from the same review (finding 5): the orchestrator's onStartFailure wiring (hint selection, composite-key -> panelTabOf mapping, ack+turn push) has no test coverage — add one when fixing this.

### LINKED ISSUE #256: messages sent while an agent is failing to start are dropped without any ack
From the #253 adversarial review (finding 4). Messages queued into a doomed agent between spawn and prepare()-reject die with it: no seen ack, no failure ack, and the working receipt already cancelled the panel's delivery-failure timer — the bubble sits muted-'queued' forever. The per-tab ⚠️ say mitigates. Fix idea: settle(err) should flush the agent's queued/held messages back into heldMessages (or nack them) so the next successful start re-delivers.

### LINKED ISSUE #257: turn state:'idle' frame is a no-op — panel only handles working/done
From the #253 adversarial review (finding 7). src/orchestrator/index.ts (~line 3208) pushes { type: 'turn', state: 'idle' } to 'clear the working spinner', but the panel's turn handler only recognizes 'working' and 'done' — the clear never happens. Either push state:'done' there or teach the panel 'idle'.

artokun:
Follow-ups from the adversarial review of #253 (findings 2, 4, 5, 7).

## #255 — rebind race stranded a prepare-failing agent under its new key

**What:** `settle()` in `PanelAgentManager.spawn()` guarded on its spawn-time key (`agents.get(tabId) !== agent`). If `rebindAgent()` moved a still-starting agent to a new key (panel tab-id migration) and `prepare()` then rejected, settle early-returned: the dead agent stayed mapped under the new key — `hasLiveAgent` true, queue never drained, no `onStartFailure`, tab wedged until reset.

**Fix:** settle now locates the agent by **identity** (scan of the agents map) and cleans up / fires `onStartFailure` / `onAgentFatal` under whatever key **currently** maps the agent. If the agent is no longer mapped at all (replaced by `restartAgentResume`), settle is still a no-op.

**Also (finding 5):** the orchestrator's inline `onStartFailure` wiring had zero test coverage. The frame construction — hint selection via `openAiKeyProvider`, composite-key → panel-tab split, `say` + degraded `ack` + `turn:done` push order — is extracted to `src/orchestrator/start-failure-notice.ts` (index.ts now just pushes the built frames) and unit-tested in `start-failure-notice.test.ts`.

## #256 — messages into a failing start died silently

**What:** anything queued into a doomed agent between spawn and `prepare()`-reject died with it: no seen ack, no failure ack, bubble stuck muted-queued forever.

**Fix:** `prepare()` rejects **before the channel is ever consumed**, so every message since the spawn is still in the agent's queue. `settle(err)` captures them (via `takePending()`, before `stop()`) into a manager-level held-mail map; the **next spawn on the same key re-delivers them first**, in order, with their original `mid`s — so the fresh agent's dequeue fires the normal `seen` acks and the panel bubbles flip from queued to read. Redelivery semantics were chosen over nacking to match the existing held-mail precedent (`heldDuringGen` / `restartAgentResume` carry-over): the message is answered, not bounced. If the retry fails too, settle re-captures — no loss loop. Held mail migrates on `rebindAgent()` (in the durable pass — it exists precisely when no live agent does), and is dropped on explicit `reset()` / `stopAll()`.

`takePending()`'s return type now includes `mid` (the runtime items always carried it), so `restartAgentResume` carry-over also preserves seen-tracking.

## #257 — `turn state:'idle'` was a no-op

**What:** the local-VRAM hold path (`index.ts`, held-during-gen branch) pushed `{ type: 'turn', state: 'idle' }` to clear the working spinner, but the panel's `onTurn` handler only recognizes `'working'` and `'done'` — the spinner ran until the 120s safety timeout.

**Fix:** push `state: 'done'`. Verified read-only against the panel repo (`web/js/comfyui-mcp-panel.js`, `onTurn`): `'done'` hides the spinner, disarms the mid-task resume nudge (correct here — no turn is in flight; the held message re-delivers on render end with a fresh `'working'`), and snapshots the graph in a try/catch (harmless).

Fixes #255
Fixes #256
Fixes #257

## Test evidence

- `npx tsc --noEmit` — clean
- `npx vitest run` — **137 files / 1551 tests passed**
- New: `start-failure-rebind-heldmail.test.ts` — rebind-during-failing-start → slot cleaned under the new key + `onStartFailure(newKey)` fired + recoverable; queued-during-failing-start → re-delivered in order after a successful retry with seen acks for the original mids; held mail follows a rebind.
- New: `start-failure-notice.test.ts` — hint selection (moonshot/glm/kimi via registry, openrouter/custom/generic fallbacks), composite-key → panel tab split (incl. `wf:`-prefixed ids and bare keys), exact frame sequence say → degraded ack → `turn:done` (pinning that it is not `idle`).
- Updated: `start-failure-per-tab.test.ts` recoverability case — the retry turn now (correctly) also contains the re-delivered first message, ordered ahead of the retry.

### DIFF
--- a/src/__tests__/orchestrator/start-failure-notice.test.ts
+++ b/src/__tests__/orchestrator/start-failure-notice.test.ts
@@ -1,0 +1,87 @@
+// Issue #255 (review finding 5) — the orchestrator's onStartFailure wiring had
+// no test coverage. The frame construction now lives in start-failure-notice.ts
+// (extracted from index.ts's inline handler) so this can pin down:
+//   • hint selection via the OpenAI-key provider registry (openAiKeyProvider),
+//     with the openrouter / custom / generic fallbacks;
+//   • composite-key (`panelTabId::backend`) → panel tab mapping, including the
+//     split-on-LAST-separator rule and the bare-key fallback;
+//   • the exact frame sequence pushed to the panel: honest say → degraded ack →
+//     turn:"done" (NOT "idle" — the panel's turn handler only recognizes
+//     working/done, so an idle frame never clears the spinner; issue #257).
+
+import { describe, expect, it } from "vitest";
+import {
+  backendOfKey,
+  buildStartFailureNotice,
+  panelTabOfKey,
+  startFailureHint,
+} from "../../orchestrator/start-failure-notice.js";
+
+describe("composite agent key → panel tab mapping", () => {
+  it("splits `panelTabId::backend` on the LAST separator", () => {
+    expect(panelTabOfKey("tab-1::moonshot")).toBe("tab-1");
+    expect(backendOfKey("tab-1::moonshot", "claude")).toBe("moonshot");
+  });
+
+  it("a bare key (no separator) maps to itself + the default backend", () => {
+    expect(panelTabOfKey("tab-plain")).toBe("tab-plain");
+    expect(backendOfKey("tab-plain", "claude")).toBe("claude");
+  });
+
+  it("a tab id that itself contains '::'-free prefixes still splits correctly", () => {
+    // Deterministic panel ids look like "wf:abc" / "tmp:abc" — the colon inside
+    // the tab half must not confuse the split.
+    expect(panelTabOfKey("wf:abc123::glm")).toBe("wf:abc123");
+    expect(backendOfKey("wf:abc123::glm", "claude")).toBe("glm");
+  });
+});
+
+describe("startFailureHint provider selection", () => {
+  it("names a registered OpenAI-key provider's slot + primary env key", () => {
+    // moonshot / glm / kimi live in the openai-provider-registry.
+    expect(startFailureHint("moonshot")).toContain("MOONSHOT_API_KEY");
+    expect(startFailureHint("glm")).toContain("GLM_API_KEY");
+    expect(startFailureHint("kimi")).toContain("KIMI_API_KEY");
+    expect(startFailureHint("moonshot")).toContain("API Keys card");
+  });
+
+  it("falls back for openrouter, custom, and unknown backends", () => {
+    expect(startFailureHint("openrouter")).toContain("OPENROUTER_API_KEY");
+    expect(startFailureHint("custom")).toContain("Custom endpoint");
+    expect(startFailureHint("claude")).toBe(
+      "Check the provider's credentials/login, then Disconnect → Connect to retry.",
+    );
+  });
+});
+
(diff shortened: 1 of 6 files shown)


artokun:
Addressed all 3 MAJORs from the Codex review in 071fe28:

1. **Restart gate vs held mail** — `PanelAgentManager.hasHeldMail()` added; `allIdle()` now returns false while failed-start mail is parked, and the `SelfRestarter` gate in `index.ts` checks `!manager.hasHeldMail()` explicitly alongside `heldDuringGen`. An auto-restart can no longer tear down (`stopAll()` erases the map) while re-delivery is pending.

2. **Cancel/reorder reach held mail** — `cancelQueued()` falls through to the held-mail entry by `mid` when no live agent has the message; `reorderQueue()` applies the panel's order to held entries with the same stable-sort semantics. A deleted prompt parked by a failed start is no longer re-delivered next spawn.

3. **Turn-aware spinner clear** — the held-during-gen `turn:"done"` is now gated on `!manager.isTurnActive(key)` (busy **or** pending → a real turn owns the spinner and pushes its own done); when a turn is active, nothing is pushed (the idle no-op is dropped entirely). The start-failure path's `turn:"done"` stays unconditional — no turn can be in flight when `prepare()` rejects, so the notice test is unchanged.

Regression tests added for all three (held mail blocks `hasHeldMail()`/`allIdle()`; cancel-by-mid + reorder-by-mid reach held mail, cancelled prompt gets no seen ack; `isTurnActive` true mid-turn and with a queued follow-up, false when idle/absent).

Gates: `npx tsc --noEmit` clean; `npx vitest run` **137 files / 1555 tests passed**.

PR #261, opened 2026-07-21
fix(queue-status): foreign-run attribution + sub-tick run completions in the queue_status broadcast

### LINKED ISSUE #258: queue_status broadcast carries no prompt_id on ComfyUI 0.28+ — clients can't attribute runs
Found during on-device verification of the mobile Diagnose button (artokun/comfyui-mcp#249, mobile PR #21). Against ComfyUI 0.28.0 the orchestrator's queue_status frames are always running:false / prompt_id:null with depth transitions only — passive watchers never learn WHICH run is executing or just finished. The mobile app had to grow an anonymous diagnose mode (id-less diagnose_run + burst baseline/dedupe) to work at all. Fix idea: QueueMonitor should enrich its broadcast from GET /queue (queue_running carries [number, prompt_id, ...]) so frames regain prompt_id on modern ComfyUI.

### LINKED ISSUE #259: runs shorter than the QueueMonitor poll tick never reach queue_status subscribers
Found during on-device verification of mobile PR #21: a run that fails in under ~1-2s can start and finish entirely between the broadcaster's 1 Hz ticks — no queue_status transition is ever emitted, so the mobile queue bar (and the new Diagnose affordance) never see it. Wire-level, affects all passive consumers. Fix idea: on each tick also diff /history's tail (or subscribe to ComfyUI's own websocket executing/executed events) so sub-tick runs still produce a transition.

artokun:
Fixes #258
Fixes #259

## Root cause (#258)

Verified wire-level against the live ComfyUI **0.28.0**: a passive, non-originating WS client receives **only `status` frames** (queue_remaining) — `execution_start` / `executing` / `execution_*` **and** `progress` / `progress_state` are all sid-scoped to the client that queued the prompt. Probe result for a foreign run (frame type counts on the passive socket): `{"status": 4, "crystools.monitor": 11}` — nothing carrying a `prompt_id`. So `QueueMonitor`'s adopt-from-`progress_state` path (its only attribution source; it never polled HTTP) can never fire for foreign runs, leaving frames at `running:false / prompt_id:null` with depth-only transitions. Meanwhile `GET /queue`'s `queue_running` entries do carry the prompt_id (`[number, prompt_id, ...]` — confirmed live).

## Approach

`QueueMonitor` keeps its WS (still the source of node/progress detail for older servers and status/queue_remaining everywhere) and gains a best-effort `poll()` the orchestrator awaits before each 1 Hz broadcast tick:

- **#258 — `GET /queue`**: adopt `queue_running[0][1]` as the running prompt_id + take the true depth (running+pending). Clear the run when the queue reports empty (with a stale-response race guard).
- **#259 — `GET /history?max_items=8` tail diff**: new ids since the last poll are completions — this works **regardless of sid-scoping** (the ws `executing`/`executed` events verifiably do NOT reach foreign clients on 0.28, so history diffing was the only reliable option). Primed on first look so pre-existing history never replays; deduped against the WS `execution_success/error/interrupted` events (older ComfyUI / own runs) via a bounded set. Status mapping: `success` / `error` / `interrupted` (from `status.messages`).

The broadcaster drains completions each tick and replays each as its own frame, so a run that started AND finished entirely between ticks still reaches every subscriber exactly once.

## Wire compat (panel / mobile)

**Additive only** — no existing field changes meaning. New `queue_status` fields:

- `last_completed_prompt_id: string | null` (sticky — also seeds late-connecting tabs)
- `last_completed_status: "success" | "error" | "interrupted" | null`
- `last_completed_at: number | null` (ms epoch observation time, staleness hint)

Checked consumers: `comfyui-mcp-panel/web/js` does **not** consume `queue_status` bridge frames at all (only the unrelated `nodes_queue_status` tool). The mobile app can later replace its anonymous-diagnose burst/baseline dance by keying Diagnose off `last_completed_prompt_id` + `last_completed_status === "error"`, and drop the id-less mode.

## Live verification (ComfyUI 0.28.0, real dist code driving monitor + broadcaster as a passive observer)

- Foreign 20-step run → frame `running:true, prompt_id:"20cb1354-…", queue_depth:1`, then a completion frame `last_completed_status:"success"` ✅ (#258)
- Fully-cached rerun (completes in <100 ms, entirely between ticks; `/queue` observed empty the whole time) → completion frame `last_completed_prompt_id:"542d95ad-…", last_completed_status:"success"` still emitted ✅ (#259)
- Shared queue respected: only self-enqueued 512×512 v1-5 runs, nothing cancelled, running orchestrator untouched.

## Gates

- `npx tsc --noEmit` clean
- `npx vitest run`: 136 files / 1551 tests green (includes new unit tests: /queue adoption + clear + fetch-failure resilience; history prime/diff/dedupe/interrupted mapping; broadcaster completion replay incl. burst + idle→idle sub-tick case)

## Notes

- A long foreign render with zero broadcast progress signal now gets *tracked* (it previously didn't exist to the watchdog); its stall clock only refreshes on real signals, so the stall report stays conservative for foreign runs.
- Burst semantics: up to 8 history-tail entries per tick; each drained completion is replayed as a distinct frame (bounded queue of 20).

## Codex review hardening (second commit)

Parallel /queue+/history fetches (one-timeout worst case); 32-entry history tail with a visible saturation warning; timestamp-based priming so runs finishing during startup aren't swallowed; poll generation counter + completion-state reset across stop()/start() retargets; completions ordered by ComfyUI's monotonic queue number (prompt[0]).

Also fixes the interrupted false-failure chain server-side (**comfyui-mcp-mobile#23**): `diagnose_run` now reports `Status: interrupted` (with an Interrupted section) for runs with `execution_interrupted` and no `execution_error`, instead of echoing status_str "error" — mobile's classifier will stop showing "Render failed" for cancels.

### DIFF
--- a/src/__tests__/tools/diagnose-run.test.ts
+++ b/src/__tests__/tools/diagnose-run.test.ts
@@ -194,3 +194,61 @@ describe("diagnose_run — failure explanation", () => {
     expect(validateWorkflowMock).not.toHaveBeenCalled();
   });
 });
+
+describe("diagnose_run — interrupted runs (mobile#23)", () => {
+  // ComfyUI records an interrupted run with status_str "error" plus an
+  // execution_interrupted message and NO execution_error. Echoing the raw
+  // "error" made cancelled runs read as failures downstream (mobile's
+  // classifier keys on the exact word "error" → "Render failed").
+  const interruptedEntry = (queueNumber: number, id: string) => ({
+    prompt: [queueNumber, id, { "1": { class_type: "KSampler", inputs: {} } }, {}, []],
+    outputs: {},
+    status: {
+      status_str: "error",
+      completed: false,
+      messages: [
+        ["execution_start", { prompt_id: id, timestamp: 1 }],
+        ["execution_interrupted", { prompt_id: id, node_id: "5", node_type: "KSampler" }],
+      ],
+    },
+  });
+
+  it("reports Status: interrupted (not error) with an Interrupted section, no Runtime-failure section", async () => {
+    getHistoryMock.mockResolvedValue({ p: interruptedEntry(1, "p") });
+    const text = (await getHandler("diagnose_run")({ prompt_id: "p" })).content[0].text;
+    expect(text).toContain("**Status**: interrupted");
+    expect(text).not.toContain("**Status**: error");
+    expect(text).toContain("### Interrupted");
+    expect(text).toContain("interrupted/cancelled at node 5 (KSampler)");
+    expect(text).not.toContain("### Runtime failure");
+    expect(text).not.toContain("_No runtime error recorded");
+  });
+
+  it("a genuine execution_error wins over an interrupted marker in the same entry", async () => {
+    const e = interruptedEntry(1, "p") as { status: { messages: unknown[] } };
+    e.status.messages.push([
+      "execution_error",
+      { node_id: 7, node_type: "VAEDecode", exception_type: "RuntimeError", exception_message: "boom" },
+    ]);
+    getHistoryMock.mockResolvedValue({ p: e });
+    const text = (await getHandler("diagnose_run")({ prompt_id: "p" })).content[0].text;
+    expect(text).toContain("**Status**: error");
+    expect(text).toContain("### Runtime failure");
+    expect(text).not.toContain("### Interrupted");
+  });
+
+  it("formatRunOutcome maps an interrupted entry without node info too", async () => {
+    const { formatRunOutcome } = await import("../../tools/diagnostics.js");
+    const lines = formatRunOutcome({
+      prompt: {},
+      outputs: {},
+      status: {
+        status_str: "error",
+        completed: false,
+        messages: [["execution_interrupted", {}]],
+      },
+    } as never);
+    expect(lines[0]).toBe("**Status**: interrupted");
+    expect(lines.join("\n")).toContain("### Interrupted");
+  });
+});
--- a/src/orchestrator/index.ts
+++ b/src/orchestrator/index.ts
@@ -3282,8 +3282,16 @@ export async function runPanelOrchestrator(): Promise<void> {
   const queueStatusBroadcaster = createQueueStatusBroadcaster(
     () => QueueMonitor.snapshot(),
     (frame) => void bridge.push(frame),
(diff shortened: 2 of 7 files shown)


artokun:
Addressed all five codex review findings in 0651c1f:

1. **Parallel fetches** — `/queue` + `/history` now run via `Promise.all`; worst-case poll is one 2.5s timeout, not a serial ~5s. Overlap guard retained. Regression test asserts both requests are in flight before either resolves.
2. **Tail width + priming** — (a) tail raised to `max_items=32` with a `logger.warn` when a diff saturates the whole window (all entries unseen) instead of silently claiming coverage; pending-completion bound raised to 64 so a saturated diff reaches the broadcaster whole. (b) Priming now uses the server-side message timestamps against the monitor start time: a run finishing during startup is reported, only entries that genuinely predate the monitor are swallowed (verified live earlier that history messages carry ms-epoch `timestamp`; remote clock skew degrades to at worst one replayed/swallowed tail entry at startup, noted in code).
3. **Interrupted false-failure chain** — `diagnose_run` now reports `**Status**: interrupted` + an `### Interrupted` section (naming the node when recorded) when the entry has `execution_interrupted` and no `execution_error`; a genuine `execution_error` still wins. Outcome formatting extracted to exported `formatRunOutcome` with 3 new tests. This is the server-side fix for **comfyui-mcp-mobile#23** — mobile's exact-"error" classifier will now receive "interrupted" and can stop showing "Render failed" for cancels.
4. **Retarget race** — generation counter captured at poll start, checked after the awaits; `stop()`/`start()` bump it, so an in-flight poll's stale responses are stranded. `start()` also clears `lastCompleted`/pending completions/dedupe set (+ re-primes), so nothing leaks across a retarget. Two regression tests (mid-flight bump; start() reset).
5. **Ordering** — new-completion diff sorts by ComfyUI's monotonic queue number (`prompt[0]`, per history-select.ts), so bursts replay oldest→newest and `lastCompleted` is genuinely the most recent. Regression test scrambles object order and asserts queue-number order.

Gates after fixes: `npx tsc --noEmit` clean; `npx vitest run` **136 files / 1560 tests green** (9 new). Live E2E re-run against ComfyUI 0.28.0 still passes: foreign-run attribution frame + sub-tick completion frame both observed on the wire.

PR #251, opened 2026-07-21
feat(runpod): RunPod connector — manage a live pod + live status broadcast + idle auto-stop

artokun:
Wires in a **RunPod connector**: the agent manages a pod's lifecycle by ID, connects comfyui-mcp to it, and broadcasts **live status** to a control panel — with **idle auto-stop** to save GPU cost. UX modeled on [gpu-cli.sh](https://gpu-cli.sh) (`gpu status` / `gpu dashboard` / auto-stop-when-idle), credited throughout — see gpu-cli/gpu#31.

## Tools (over RunPod's GraphQL API — `RUNPOD_API_KEY` from `~/.comfyui-mcp/.env`)
| Tool | Does |
|---|---|
| `runpod_pod_status` / `runpod_list_pods` | Live pod state (status, GPU, uptime, $/hr, ComfyUI URL) |
| `runpod_pod_start` / `runpod_pod_stop` | Resume / stop a pod (keeps disk, releases GPU) |
| `runpod_pod_troubleshoot` | exists → RUNNING → GPU → port exposed → ComfyUI answering; returns the blocker + fix |
| `runpod_pod_connect` | **Live connection** — verifies health, retargets comfyui-mcp at the pod (`setComfyuiTarget`+`resetClient`), and starts the status broadcast |
| `runpod_watch` / `runpod_unwatch` | Broadcast a pod's live status without retargeting |
| `runpod_deploy_link` | The **referral** deploy link (`template=bnqtkvcer3&ref=dkx71w9b`) + gpu-cli credit |

## Live status broadcast (the "status updates")
`services/runpod-watch.ts` polls the watched pod every ~15s and pushes a change-only `runpod_status` frame (status, GPU/VRAM util, uptime, $/hr, ComfyUI URL, idle countdown) to every panel/mobile tab — mirroring the existing `queue_status` / `download_progress` broadcasters. New tabs are seeded with the current frame. Auto-starts on `runpod_pod_connect`.

## Idle auto-stop (gpu-cli parity)
While a connected pod's ComfyUI is idle (queue empty, nothing running) past `RUNPOD_IDLE_STOP_MINUTES` (default 15; `0` disables), the pod is **auto-stopped** to save GPU-time cost. The panel keeps showing it EXITED so it can be restarted.

## Referral
RunPod referral attaches at signup/deploy via a `?ref=` link, not a per-pod API param — so the connector manages existing pods over the API and hands users the deploy link for creation, which carries our template + ref code.

## Attribution
[...]

PR #263, opened 2026-07-21
feat(training): LoRA training on RunPod pods (P4) + E2E-hardened transport

artokun:
## What

Train LoRAs on RunPod pods end-to-end: `train_start` with `target: "pod"` stages the dataset + generated ai-toolkit config to a connected pod, launches pod-native training over ssh (no docker on the pod), streams progress/samples back, and delivers the finished LoRA per `deliverTo` (pod / local / both) with a LoRA-catalog upsert when local.

Stacks on #251 (RunPod connector). Panel counterpart: training-wizard Local/Pod switch (PR in comfyui-agent-panel).

## E2E proof (real A40 pod, 2026-07-21)

- Pod deployed via connector, bootstrapped via `train_bootstrap` (pinned ai-toolkit + venv)
- 32GB FLUX.1-dev snapshot uploaded (HF token lost gated access mid-run — see below), `model_path` override pointed the trainer at the pod-local dir
- `pod_e2e2`: 200 steps @512px, 6-image dataset, trigger `ohwx` — **completed, final loss 0.1608**, 9 samples coherent at steps 0/100/200
- Delivery verified in all four zones: pod `/workspace/models/loras/pod_e2e2.safetensors` (172MB), local ComfyUI loras dir, catalog entry `loras-pod-e2e2` + preview, samples mirrored to the rig
- Bonus: the harness kill mid-run exercised **owner-dead orphan recovery** for real — `train_status` pulled output back, handed off, and completed the job honestly

## Hardening (codex review rounds 1-6, all fixed)

- **tar-over-ssh transport** replaces rsync (Windows rigs lack rsync): `pipe()` kills both children on every settle path, waits for both exit codes, EPIPE is failure noise not a process crash; `--delete` semantics on dataset staging; mktemp+rename delivery can't clobber unrelated pod files; leading-dash names survive (14/14 live probes)
- **Concurrency**: carries the job-registry hardening that post-dates #237's merge (`a41d45f`)
- **Idle-stop guard**: pod idle auto-stop only applies to a pod we're rendering on (`985c72e`) — training jobs aren't ComfyUI queue jobs
- `runpod-client`: env vars as `{key,value}[]` per API (fixes PUBLIC_KEY → ssh on one-tap pods)

Tests: 80 trainer/ssh/config tests + tsc green; full suite 1648 green at `bc61319`.

## Known follow-ups

- HF token needs FLUX.1-dev license re-acceptance (401 on weights from rig AND pod since ~07-21; metadata API fine). `model_path` works around it but the default HF path stays broken until fixed.
- Idle-stop vs training: the guard covers the render-on-pod case; a training-aware idle check is a deliberate follow-up.

🤖 Generated with [Kimi Code](https://github.com/MoonshotAI/kimi-code)

---

**Update (2026-07-22):** rebased onto `fix/runpod-billing-secrets` (#272 — the #251 salvage). Until #272 merges, this PR's diff includes its two commits; merge order: **#272 → #263** (the createPod conflict resolution is already done here: billing-safe + port-3000 + `22/tcp` + `PUBLIC_KEY` env-array all coexist). Codex round 7 also landed: pod cancels are scoped to the job's own config (no more blanket `pkill run.py`), and `target: local + deliverTo: pod` is rejected up front. Known follow-ups filed: #273 (per-pod reservation race), #274 (per-pod idle-stop scoping), #275 (local-native start wiring).


artokun:
CI fix pushed (test-only): the new symlink-escape test laid its trap link inside datasetsRoot() without creating the dir first — ENOENT on all three OSes (it passed on Windows dev machines only because itWithSymlinks skips without symlink privileges). One mkdirSync before the symlink; the containment guard under test is untouched. tsc clean, file 59 passed/1 skipped locally. Codex review of the training delta is running separately — findings will follow as a comment.

artokun:
Codex adversarial review of the training delta (diffed base→head, connector divergence excluded). Verdict: **FIX FIRST**. CI note: the symlink-test ENOENT is already fixed on the branch (708be4e) — checks are green.

**Transport & secrets**
1. **BLOCKER — training-jobs.ts:739: a failed output download can still publish a truncated LoRA.** `pullPodOutput` only logs a nonzero tar/ssh exit and continues; extraction writes straight into the final dir (runpod-ssh.ts:223), and the existence-only handoff (training-jobs.ts:1097) copies the partial `name.safetensors` and can mark the job completed. With `deliverTo: "pod"`, the partial local file is then uploaded OVER the valid pod artifact. Fix: check the transfer exit code, extract into a temp dir, verify (size/checksum against the pod-side stat), then promote atomically.
2. MAJOR — runpod-ssh.ts:229: pod archives are extracted LOCALLY with no entry/symlink validation — a pod-controlled symlink at the expected `.safetensors`/sample name is preserved by tar and later followed by copyFileSync. Validate entries + temp-dir + atomic promote (same shape as fix 1).
3. MAJOR — runpod-ssh.ts:255/260: `HF_TOKEN` rides in the remote command string AND the local ssh argv — visible to local process inspection and the remote shell's cmdline. Pass it via stdin or a 0600 remote env file.

**Money & liveness**
4. MAJOR — training-jobs.ts:803 + index.ts:3335: the idle guard treats every persisted queued/running record as active WITHOUT probing, and owner-death reconciliation only runs via getJob/listJobs — if the harness dies and nobody calls train_status, auto-stop is suppressed and the pod bills forever. The watcher path needs the probing variant (or a periodic reconcile).
5. MAJOR — training-jobs.ts:809: `hasActiveTrainingJob("pod")` is global — a job on pod A suppresses idle-stop for watched pod B. Scope by pod id/endpoint.

**Concurrency & cancellation**
6. MAJOR — training-jobs.ts:1305→1365: the one-training-per-pod guard is scan-then-persist with no pod-scoped lock — two MCP processes can both see the pod free and double-launch (the per-job lock machinery added for #237 could host a pod-scoped lock).
7. MAJOR — runpod-ssh.ts:298/314 (+ caller at training-jobs.ts:799 omitting remoteConfigPath): cancel runs `pkill -f [r]un.py` — can kill unrelated pod workloads, and the post-stop probe reports "alive" if ANY run.py exists. Persist the remote PID at launch and kill/verify by PID.

**Delivery**
8. MAJOR — training-jobs.ts:1107 (+ runpod-ssh.ts:214): delivery silently, non-atomically overwrites an existing local LoRA of the same name; a copy failure can truncate the existing model. Collision-check + tmp-write + rename.

(Codex's sandbox couldn't run tsc/vitest — UNVERIFIED there; CI is green on the head. Findings are scoped to the 12-file training delta; the connector-layer findings live on #251.)

PR #285, opened 2026-07-22
feat(apps): apps_* tools — micro-app run surface for mobile/canvas-less clients

artokun:
## The Apps epic (orchestrator side)

Sibling of **comfyui-mcp-panel#114** (panel + registry) and the comfyui-mcp-mobile Apps-tab PR. Thin proxies over the panel pack's `/comfyui_mcp_panel/apps/*` routes, so the panel stays the single storage + run implementation for both desktop and mobile.

- `apps_list` / `apps_get` — read the rig's registered micro-apps (manifest: appMode inputs/outputs, deps, hideWorkflow).
- `apps_run` — patch `<nodeId>.<widget>` values into the app's stored prompt snapshot and queue it; returns `prompt_id`. `apps_run_status` polls status + outputs.
- `apps_import` — fetch a bundle from the public registry and install it as a local app (registry id becomes the local id; deps are reported, not auto-installed).

All five are added to the mobile `call_tool` whitelist — same risk posture as the already-whitelisted `enqueue_workflow` / `save_workflow` / `download_*` tools: user-initiated, validated in the tools, no model/system mutation.

**Tests:** 8 new vitest cases (proxy routing, error translation, import round-trip, URL guard); `tsc` build green; docs:gen regenerated (163 tools).

No merge/release per maintainer — awaiting codex:rescue review.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -86,6 +86,7 @@
               "tools/defaults-stats-skills",
               "tools/runpod",
               "tools/training",
+              "tools/apps",
               "tools/comfy-cli",
               "tools/skills-knowledge"
             ]
--- a/docs/tools/apps.mdx
+++ b/docs/tools/apps.mdx
@@ -1,0 +1,143 @@
+---
+title: "Apps (micro-apps)"
+description: "List, inspect, and run the panel's micro-apps — named, one-click workflow apps with exposed inputs — for canvas-less clients (mobile)."
+icon: "layout-grid"
+---
+
+<Info>5 tools. Generated from the live MCP tool schemas — do not edit by hand; run `npm run docs:gen`.</Info>
+
+## apps_list
+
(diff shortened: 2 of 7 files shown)


artokun:
### codex:rescue review — 4 findings, all fixed in `d9cbd32`

Ran `codex exec review --base main` on the branch. Findings triaged as all-real; fixes pushed, 9 vitest cases green.

**P1s**
1. **Route traversal via `prompt_id`** — `apps_run_status` interpolated the id into the URL, so `../../../../system_stats` escaped the apps route into arbitrary ComfyUI GET endpoints *with the configured auth headers* — a whitelist bypass. Fixed: shape validated at the zod boundary AND in-handler, segment URL-encoded, regression test asserts no fetch fires.
2. **SSRF via `registry_url`** — scheme-checking alone left loopback/LAN and redirect-into-LAN open on a server-side fetch. Fixed: allowlist (default registry + `COMFYUI_MCP_REGISTRY_URLS`), `redirect: "error"`.

**P2s**
3. **Unbounded bundle download** — no timeout/size cap. Fixed: 30s/15s aborts + 16MB/5MB caps on declared and actual bytes.
4. **Thumbnails lost on import** — they live at a separate registry endpoint, not in the JSON bundle. Fixed: fetched and forwarded as `thumbnail_b64` so installed apps keep their card art.

PR #294, opened 2026-07-22
fix(codex): make interrupted and retrying turns recoverable

### LINKED ISSUE #293: Codex turn can remain pending forever when app-server ignores turn/interrupt
## Environment

- `comfyui-mcp 0.45.0` / current `main` (`3077250`)
- Windows 11
- Agent Panel using the Codex app-server backend (`gpt-5.6-sol`)

## Symptom

A Codex turn can remain in the working state indefinitely while newly submitted panel messages accumulate as `pending`. Pressing Stop or using send-now does not recover it when the live app-server process stops answering `turn/interrupt`.

## Root cause

`CodexBackend.interrupt()` awaits the `turn/interrupt` JSON-RPC request without a deadline. In addition, `AppServerClient.close()` awaits `exitPromise` indefinitely, and `runTurn()` can only reach its terminal result after a protocol completion or process exit. A live-but-silent child therefore keeps all three layers waiting:

1. `PanelAgent.interrupt()` cannot arm its bounded gate fallback until backend interrupt returns;
2. the active backend generator never emits its required terminal `result`;
3. the old run never exits, so it cannot hand control to a fresh app-server and the preserved queue.

The same path can also surface a misleading provider-failure banner during intentional recycle.

## Expected behavior

- Bound `turn/interrupt` acknowledgement time.
- Detach and recycle an app-server that misses the deadline.
- Finish the active turn locally without depending on `close()` or process `exit`.
- Emit exactly one controlled terminal result and no false error banner.
- Do not consume the next queued message from the retired run.
- Bound graceful close, process-tree termination, force-kill escalation, and backend shutdown.
- Preserve the Codex thread/session id so the fresh app-server can resume it.

## Reproduction test

A deterministic regression uses a fake client where all three promises stay pending: `turn/interrupt`, `close()`, and `exitPromise`. On current main, the interrupt and active iterator do not settle. With the proposed recovery path, interrupt settles within its deadline, the iterator yields `{ type: result, ok: true, subtype: interrupted }`, exits, and the next channel item remains unread.

I have a focused PR ready with the backend fix and six Vitest cases covering the combined hang, responsive interrupts, repeated interrupts, client-replacement races, bounded shutdown, and late RPC rejection.

JusticeWay:
Fixes #293.

## What changed

- bound `turn/interrupt` with a configurable deadline (1500 ms by default)
- locally terminate the active turn with exactly one controlled `interrupted` result when the RPC deadline expires
- detach the stale client before teardown and prevent a late timeout from clearing a replacement client/session
- stop the retired backend run before it can consume the next queued turn
- make app-server close idempotent and bounded, including process-tree termination, force-kill escalation, pipe detach, and a backend-level close budget
- avoid the false provider-failure banner during intentional recycle
- honor Codex app-server `ErrorNotification.willRetry`: retryable `Reconnecting... N/5` notifications keep the turn alive and only non-retrying errors are terminal
- add eight deterministic Vitest regressions for responsive/hung/repeated interrupts, the combined active-run + hung `close()` + hung `exitPromise` case, replacement-client races, bounded backend shutdown, and late RPC rejection

## Root cause

The existing path had three unbounded dependencies: `CodexBackend.interrupt()` waited forever for the JSON-RPC acknowledgement, `AppServerClient.close()` waited forever for `exitPromise`, and `runTurn()` only finished after a protocol terminal event or process exit. A live-but-silent app-server therefore prevented `PanelAgent.interrupt()` from arming its fallback, never emitted the terminal result required by the turn gate, and left later panel messages in `pending`.

The new per-turn abort hook releases the generator independently of OS/process teardown. The old run then exits without reading the next channel item, allowing the existing PanelAgent restart path to create a fresh app-server and resume the persisted Codex thread.

## Configuration

- `COMFYUI_MCP_CODEX_INTERRUPT_TIMEOUT_MS` (default `1500`)
- `COMFYUI_MCP_CODEX_CLOSE_TIMEOUT_MS` (default `2000`)
- `COMFYUI_MCP_CODEX_FORCE_KILL_GRACE_MS` (default `500`)

## Validation

- `npm run lint` ? pass
- `npx vitest run src/__tests__/orchestrator/codex-interrupt-recovery.test.ts` ? 8/8 pass
- full `npm test` ? 1782 passed, 1 skipped; one unrelated local-environment test fails depending on host state (`node-dev` sees the installed `rg`, or the native training probe sees the running local ComfyUI process). No failures are in the Codex/backend scope.

The regression's critical fake deliberately leaves `turn/interrupt`, `close()`, and `exitPromise` pending forever. The patched iterator still emits `{ type: 'result', ok: true, subtype: 'interrupted' }`, terminates, and leaves the next queued item unread.

### DIFF
--- a/src/__tests__/orchestrator/codex-interrupt-recovery.test.ts
+++ b/src/__tests__/orchestrator/codex-interrupt-recovery.test.ts
@@ -1,0 +1,215 @@
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+
+vi.mock("../../utils/logger.js", () => ({
+  logger: {
+    debug: vi.fn(),
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+  },
+}));
+
+type BackendModule = typeof import("../../orchestrator/codex-backend.js");
+type Backend = InstanceType<BackendModule["CodexBackend"]>;
+
+let CodexBackend: BackendModule["CodexBackend"];
+
+beforeAll(async () => {
+  vi.stubEnv("COMFYUI_MCP_CODEX_INTERRUPT_TIMEOUT_MS", "25");
+  vi.stubEnv("COMFYUI_MCP_CODEX_CLOSE_TIMEOUT_MS", "25");
+  vi.stubEnv("COMFYUI_MCP_CODEX_FORCE_KILL_GRACE_MS", "10");
+  vi.resetModules();
+  ({ CodexBackend } = await import("../../orchestrator/codex-backend.js"));
+});
+
+afterAll(async () => {
+  // Allow detached close-budget checks from intentionally hung fakes to settle.
+  await new Promise((resolve) => setTimeout(resolve, 200));
+  vi.unstubAllEnvs();
+});
+
+const DEADLINE_MS = 250;
+
+async function settlesWithin(promise: Promise<unknown>, timeoutMs = DEADLINE_MS): Promise<boolean> {
+  let timer: NodeJS.Timeout | undefined;
+  try {
+    return await Promise.race([
+      promise.then(
+        () => true,
+        () => true,
+      ),
+      new Promise<boolean>((resolve) => {
+        timer = setTimeout(() => resolve(false), timeoutMs);
+      }),
+    ]);
+  } finally {
+    if (timer) clearTimeout(timer);
+  }
+}
+
+function installActiveClient(backend: Backend, client: object): void {
+  Object.assign(backend, {
+    client,
+    threadId: "thread-1",
+    turnId: "turn-1",
+  });
+}
+
+describe("CodexBackend interrupt recovery", () => {
+  it("keeps a responsive client and bounds an unresponsive one", async () => {
+    const responsive = {
+      request: vi.fn().mockResolvedValue({}),
+      close: vi.fn().mockResolvedValue(undefined),
+    };
+    const responsiveBackend = new CodexBackend();
+    installActiveClient(responsiveBackend, responsive);
+
+    await expect(responsiveBackend.interrupt()).resolves.toBeUndefined();
+    expect(responsive.request).toHaveBeenCalledWith("turn/interrupt", {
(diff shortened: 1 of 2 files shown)


JusticeWay:
Additional real-runtime failure reproduced and fixed in `359eaed`: Codex app-server 0.145.0 emits `method: error` with `willRetry: true` while reconnecting (observed message: `Reconnecting... 2/5`). The backend treated every `error` notification as terminal, emitted `{ok:false}`, and detached the handler before Codex could finish its retry. The protocol-generated type is `ErrorNotification = { error, willRetry, threadId, turnId }`. The handler now ignores only retryable errors as terminal outcomes (the notification still calls `onActivity`); `willRetry: false` retains the existing terminal path. Added positive and negative regression tests. Validation: targeted 8/8 and `npm run lint` pass; full suite 1782 passed / 1 skipped with the same unrelated host-dependent `node-dev` ripgrep assertion.

artokun (approved):
Thank you for this — it's an exceptional contribution. 🙏

The diagnosis in #293 nailed the real problem: three unbounded waits (`interrupt()`, `close()`, `runTurn()`) all blocking on OS/process events that a live-but-silent app-server never sends. Bounding all three and recovering *locally* — emitting one controlled `interrupted` result so the turn gate advances, detaching the dead client, and not letting the retired run swallow the next queued message — is exactly the right shape.

What stood out on review:
- Idempotent `close()` via a shared promise, and the `closingClients` budget so detached teardown can't leak
- SIGTERM→SIGKILL escalation that keeps the POSIX process-group kill, so the Windows shim's grandchild isn't orphaned
- Every timeout path clears its timer in `finally`
- 8 deterministic regressions, including the combined-hang fake where all three promises stay pending forever

Verified locally: the regression suite is 8/8, typecheck is clean, and the diff is tightly scoped to the two files. One nit I noted for the record (not blocking): the `willRetry === true` branch degrades to a safe no-op if the app-server never emits that field, so it can only help.

Merging with gratitude.

PR #334, opened 2026-07-28
fix(comfyui): restart_comfyui relaunch fails with spawn EFTYPE on Windows (#330)

### LINKED ISSUE #330: restart_comfyui stops Windows process but relaunch fails with spawn EFTYPE
On Windows, mcp__comfyui__restart_comfyui successfully stops the running ComfyUI process but fails to relaunch it, returning `spawn EFTYPE`. The stopped process must then be started manually.

### Steps to reproduce
1. Connect comfyui-mcp to a local Windows ComfyUI running from embedded Python.
2. Call restart_comfyui.
3. Observe `spawn EFTYPE`; the original PID is gone and port 8188 remains down.

### Exact error
spawn EFTYPE

### Fix
Applied locally: no (upstream tool issue).
Upstream-only: yes.
Suggested: inspect Windows spawn options/stdio or executable argument handling in the restart relaunch path. The captured command was a normal Python executable plus main.py and launch flags.

### Environment
Windows 11
ComfyUI 0.28.0
Python 3.13.11
Local executable and paths omitted; no secrets included.

artokun:
## Root cause

`restart_comfyui` (and `start_comfyui`) relaunch ComfyUI from the argv captured via `/system_stats`. That argv is Python's `sys.argv`, whose **argv[0] is the `main.py` script path, not the Python interpreter**. `spawnFromProcessInfo` did:

```ts
const [pythonExe, ...args] = info.argv;
spawn(pythonExe, args, { shell: false, ... });
```

so it spawned `main.py` as the executable. On Windows, execing a `.py` as a PE binary with `shell:false` fails with **`spawn EFTYPE`** (exec format error). The stop half works because it kills by PID-from-port and never touches argv — so ComfyUI stopped, but the relaunch died, the original PID was gone, and port 8188 stayed down. (On Linux/macOS a `.py` with a shebang + exec bit can accidentally run, which is why this was Windows-only.)

## The fix

`resolveLaunchCommand` now detects when argv[0] is a `.py`/`.pyw` script and resolves the real ComfyUI Python interpreter (reusing the existing `findComfyuiPython` helper — standalone-env / python_embeded / .venv under the install root inferred from `main.py`), then passes the whole argv (`main.py` + flags) as the interpreter's args. When argv[0] is already an interpreter (e.g. a child we spawned ourselves), it is spawned verbatim, so the crash-supervision path is unchanged. Also added `windowsHide:true` to match the other spawn/execFile call sites.

This makes the relaunch use a real, executable interpreter instead of handing a script to `spawn` — the divergence was that the saved argv never contained an interpreter to begin with.

## Tests

- Added a regression test asserting that when `argv[0]` is `…/main.py`, the relaunch resolves the Python interpreter and spawns `python [main.py, ...flags]` with the right options.
- `npx vitest run process-control` → 7/7 pass. `npm run build` (tsc) clean.
- Full `npm test`: 1930 pass, 1 skipped. The single failure (`node-dev.test.ts` expecting search engine `builtin` but ripgrep is installed on this machine) is a pre-existing, environment-dependent assertion unrelated to this change.

Closes #330

### DIFF
--- a/src/__tests__/services/process-control.test.ts
+++ b/src/__tests__/services/process-control.test.ts
@@ -1,4 +1,5 @@
 import { EventEmitter } from "node:events";
+import { join } from "node:path";
 import { afterEach, beforeEach, describe, expect, it, vi, type Mock } from "vitest";
 
 const mockConfig = vi.hoisted(() => ({
@@ -183,6 +184,31 @@ describe("process-control startup readiness", () => {
       }),
     );
     expect(children[0].unref).toHaveBeenCalled();
+  });
+
+  it("anchors a RELATIVE script argv[0] to the ComfyUI root (portable launcher, #330)", async () => {
+    // The standard Windows portable launcher runs `python ComfyUI\main.py` from
+    // the portable ROOT, so sys.argv[0] is relative. Since we force cwd to the
+    // nested ComfyUI dir, passing it verbatim would look for ComfyUI\ComfyUI\
+    // main.py — it must be anchored to config.comfyuiPath.
+    __processControlTestHooks.setLastProcessInfo({
+      pid: 0,
+      port: 8188,
+      argv: ["ComfyUI\\main.py", "--port", "8188"],
+      isDesktopApp: false,
+    });
+    mockSpawnedChildren();
+    mockNoPortProcess();
+    mockFetchOk(true);
+
+    const result = await startComfyUI();
+
+    expect(result.started).toBe(true);
+    expect(mockSpawn).toHaveBeenCalledWith(
+      "/fake/ComfyUI/python_embeded/python.exe",
+      [join("/fake/ComfyUI", "main.py"), "--port", "8188"],
+      expect.objectContaining({ cwd: "/fake/ComfyUI", shell: false }),
+    );
   });
 
   it("reports timeout instead of ready when bounded probes never succeed", async () => {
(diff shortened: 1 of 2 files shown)


artokun:
## Codex review (codex-cli 0.144.1)

**Verdict: REQUEST CHANGES** — one blocking finding (valid), now addressed.

> Blocking: standard Windows portable launches with relative `argv[0]` (`ComfyUI\main.py`). The patch passes that unchanged while forcing `cwd` to the nested ComfyUI root. Python therefore looks for `ComfyUI\ComfyUI\main.py`. The test masks this by using an absolute script path. Normalize the script to an absolute path or preserve the original working directory, then add a `ComfyUI\main.py` regression case.
>
> Type-check and 23 focused tests passed. The full suite had one unrelated `node-dev.test.ts` failure.

### Resolution (commit acd96e5)
Correct catch. The Windows portable launcher runs `python ComfyUI\main.py` from the portable **root**, so `sys.argv[0]` is relative — and since the relaunch forces `cwd` to `config.comfyuiPath` (the nested ComfyUI dir that holds `main.py`), a relative script would resolve to `…/ComfyUI/ComfyUI/main.py`.

`resolveLaunchCommand` now anchors a **relative** script argv[0] to `main.py` under the resolved ComfyUI root (`join(config.comfyuiPath, basename(script))`); absolute script paths are still used verbatim. Added a `ComfyUI\main.py` relative-argv regression test. process-control suite now 8/8, tsc clean.

PR #335, opened 2026-07-28
fix(orchestrator): rebind panel session tabId on explicit signal — self-heal after reconnect/reload/workflow-switch (#322, #331, #332)

### LINKED ISSUE #322: ui-bridge resolveTarget hard-fails forever after a frontend reload creates a new tab id (no self-heal path)

An agent session's panel MCP tools (panel_graph_outline, panel_query_graph, panel_find_nodes, panel_list_workflows, panel_reload, ...) all started failing permanently with `no connected tab with id "<old-id>"` after I called `panel_reload({scope:"frontend"})` mid-session.

Root cause in `dist/services/ui-bridge.js`: `createPanelMcpServer`/`makePanelToolCtx` captures a single `tabId` in closure when the orchestrator/MCP server is created for a panel session, and every `panel_*` tool call explicitly passes `{ tabId }` to `bridge.send(...)` → `resolveTarget(tabId)`.

`resolveTarget` has a tab-id **migration** mechanism (`tabMigrations` map) meant to self-heal exactly this kind of staleness, but it is only populated when the **same socket** re-hellos under a new tab id (user switches ComfyUI workflow tab in-page — see the hello handler around line 395-410). A **full frontend reload** (page refresh, or `panel_reload({scope:"frontend"})`) closes the old socket entirely and opens a **brand-new** socket with a **brand-new** tab id — there is no shared identity between old and new socket, so the same-socket re-hello path can never fire, `tabMigrations` never gets an entry for this old→new pair, and the session's captured `ctx.tabId` closure variable is never updated.

`resolveTarget(tabId)` with a truthy-but-unknown `tabId` never falls back to the "only one tab connected" auto-resolution (that fallback in the function only triggers when `tabId` is falsy) — so once this happens, **every** subsequent `panel_*` call from that agent session fails forever with the same error, including `panel_reload` itself (since it also routes through `bridge.send({tabId})`), so the session cannot even self-heal by respawning. The user has to manually Disconnect→Connect (or fully reload) the agent panel to get a fresh session bound to a fresh tab id.

### Steps to reproduce

1. Connect the agent panel to a ComfyUI tab, run any `panel_*` tool (establishes `ctx.tabId` for the session).
2. Call `panel_reload({scope:"frontend"})` (or just hit F5 on the ComfyUI browser tab) — this closes the old bridge socket and opens a new one under a new `tmp:<uuid>` id.
3. Call any `panel_*` tool again (e.g. `panel_graph_outline`).

### Exact error (scrubbed)

```
Error: no connected tab with id "tmp:0ccbd229-5c9a-47d6-b8b9-d8245c75401b". Connected: tmp:06cb (“Unsaved Workflow (2)”)
```

### Fix

applied locally: yes
upstream-only: no (patched the running install's `dist/services/ui-bridge.js`; only the built package was present, no local TS source checkout for `comfyui-mcp` itself)

Minimal, reversible patch to `resolveTarget` in `dist/services/ui-bridge.js`: when the explicit `tabId` doesn't match any connection (no exact match, no prefix match, no valid same-socket migration) AND there is exactly one connected tab, route to that tab instead of throwing (with a warn-level log noting the stale id). This mirrors the existing no-tabId fallback (`this.conns.size === 1`) for the case where a stale explicit id is passed. Does not change behavior when 0 or 2+ tabs are connected (still throws/lists tabs as before) — only closes the single-tab self-heal gap.

```diff
--- a/dist/services/ui-bridge.js
+++ b/dist/services/ui-bridge.js
@@ resolveTarget(tabId) {
             if (prefixed.length === 1)
                 return prefixed[0];
+            if (prefixed.length === 0 && this.conns.size === 1) {
+                // Stale-tabId self-heal: a full frontend reload opens a brand-new
+                // socket under a brand-new tab id -- that's a NEW connection, not a
+                // same-socket re-hello, so the tabMigrations path above (which only
+                // links old->new ids for the SAME socket) can never see it. A
+                // session bound to the old id would otherwise hard-fail on every
+                // panel_* call forever. With exactly one tab connected there is no
+                // ambiguity in practice, so route there instead of throwing.
+                logger.warn(`[ui-bridge] resolveTarget: tab_id "${tabId}" not found (likely a stale id from before a frontend reload); falling back to the sole connected tab`);
+                return this.conns.values().next().value;
+            }
             throw new Error(prefixed.length > 1
                 ? `tab_id "${tabId}" is ambiguous -- matches ${prefixed.length} tabs`
                 : `no connected tab with id "${tabId}". Connected: ${this.tabs()
                     .map((t) => `${t.tab_id.slice(0, 8)} (“${t.title}”)`)
                     .join(", ") || "none"}`);
         }
```

**Suggested proper upstream fix** (beyond this stopgap): when a `panel_reload` completes (or the bridge detects the prior socket for a session's tabId has fully disconnected with no migration), have the orchestrator re-resolve `ctx.tabId` to the newly-hello'd tab for that same logical session/workflow (e.g. by matching on workflow title, or by having the frontend reload flow pass the outgoing tabId to the new hello so the bridge can record a migration across sockets, not just same-socket re-hellos), rather than leaving `ctx.tabId` permanently pinned to a dead id.

### Environment

OS: Linux (CachyOS). ComfyUI: 0.28.0 (local). GPU: RTX 4070 Ti, 12GB VRAM. comfyui-mcp: installed via npx cache at `~/.npm/_npx/28eae33d00f3f7f1/node_modules/comfyui-mcp` (dist-only, no version string found in package.json beyond name).

### LINKED ISSUE #331: panel_* tools never self-heal after switching to a different workflow file (tabMigrations fallback doesn't fire; workflow-target "current" mode doesn't help)

An agent orchestrator session is bound to a fixed `ctx.tabId` (a panel-tab websocket id of the form `wf:workflows/<name>.json`) at session creation, in `dist/orchestrator/panel-tools.js` (`makePanelToolCtx`). When the user switches the ComfyUI panel to a DIFFERENT workflow file (closes one workflow, opens another), the panel socket re-hellos under a new `wf:<newpath>` id.

There is a designed self-heal for this in `dist/services/ui-bridge.js`: a `tabMigrations` Map (declared ~line 87, comment ~lines 79-86) is meant to record `oldTabId -> newTabId` "when a panel socket re-hellos under a NEW tab id", and `resolveTarget()` (~lines 676-714) consults `tabMigrations` as a fallback so old sessions keep working ("the session self-heals instead of throwing 'no connected tab with id …' on every panel_* call").

In practice this does NOT self-heal on a cross-workflow-file switch. Root cause: the migration-recording code path (ui-bridge.js ~lines 392-410) is explicitly and only scoped to a "same-socket re-hello":

```js
if (tabId && tabId !== msg.tab_id && this.conns.get(tabId)?.sock === sock) {
    // ... path-compress tabMigrations, then:
    this.tabMigrations.set(tabId, { to: msg.tab_id, sock });
    msg.migrated_from = tabId;
    this.conns.delete(tabId);
    ...
}
```

And `resolveTarget()`'s fallback (~lines 689-696) only honors a migration entry if the CURRENT connection is still owned by the same socket that recorded it:

```js
if (prefixed.length === 0) {
    const migrated = this.tabMigrations.get(tabId);
    if (migrated) {
        const target = this.conns.get(migrated.to);
        if (target && target.sock === migrated.sock)
            return target;
    }
}
```

If switching workflow FILES (as opposed to e.g. a panel-code id-scheme upgrade) causes a genuinely NEW socket connection (full reconnect) rather than a same-socket re-hello, no migration entry is ever recorded, and the session's `tabId` becomes permanently orphaned — every subsequent `panel_*` call throws.

We also confirmed `panel_set_workflow_target({mode:"current"})` / `panel_get_workflow_target` cannot fix this even though it looked like the right lever: per `workflow-target-store.js`, that store only controls whether `workflow_path` gets injected into `graph_*` commands. It does NOT change the `tabId` used when sending commands. `dist/orchestrator/panel-tools.js` (`makePanelToolCtx`, ~lines 230-240) hardcodes the tabId from session creation:

```js
export function makePanelToolCtx(bridge, tabId, workflowTargets) {
    const call = async (cmd, timeoutMs) => {
        try {
            const target = workflowTargets?.get(tabId);
            const routed = target ? withWorkflowTarget(cmd, target) : cmd;
            return ok(await bridge.send(routed, { tabId, timeoutMs }));
        ...
```

So `workflow_target` "current" mode is the wrong lever entirely for a stale/orphaned tab-id binding — it can only ever affect `workflow_path` injection, never which tab a command is routed to.

### Steps to reproduce

1. Open a ComfyUI workflow tab in the panel, start an agent session bound to it, and have it call several `panel_*` tools successfully (e.g. `panel_graph_outline`).
2. In the ComfyUI browser tab, close that workflow file and open/switch to a DIFFERENT workflow file (an actually different file, not the same file renamed).
3. Without creating a new agent session, call any `panel_*` tool (e.g. `panel_graph_outline`) in the SAME conversation.

Expected: the session self-heals per the `tabMigrations` design and the tool call succeeds against the newly active tab.

Actual: every `panel_*` tool call fails with "no connected tab with id …".

### Exact error (scrubbed)

```
Error: no connected tab with id "wf:workflows/Simple Klein 9B Workflow + Upscaler.json". Connected: wf:workf ("Aitrepreneur_KLEIN_EDIT_ULTRA_WORKFLOW")
```

Observed on: `panel_graph_outline`, `panel_search_nodes`, `panel_list_workflows`, `panel_reload`. Also confirmed `panel_set_workflow_target({mode:"current"})` + `panel_get_workflow_target` (which reported `mode: "current"` correctly) did NOT resolve the error, consistent with the root-cause analysis above (wrong lever).

### Fix

No live patch was attempted — this is a running orchestrator session bound to the affected code path, and we could not safely verify a fix without restarting the orchestrator mid-task (which would also just start a fresh, non-repro session). This needs an upstream engineer with a live repro to confirm which of the following is happening and fix it:

1. Investigate why the same-socket re-hello path (ui-bridge.js `tabId !== msg.tab_id && this.conns.get(tabId)?.sock === sock` check, ~line 396) doesn't fire on a cross-workflow-file switch — i.e. confirm whether the panel actually opens a NEW websocket connection per workflow file (making the current same-socket-only scoping too narrow for this case), or whether there's some other edge case in the migration chain-compression logic that swallows the entry.
2. If it's confirmed to be a genuinely new socket, the fix likely needs a complementary/looser migration path for this case (e.g. keyed by a stable panel/browser-tab identity outside of the websocket, or a short grace-period title/path correlation), since the current design's safety invariant ("only the socket that created a migration can be the target") is sound for avoiding cross-user/cross-tab leakage but appears to exclude the common close-file/open-different-file flow entirely.
3. Complementary, independent fix: make `workflow_target` "current" mode actually re-resolve to `bridge.lastActiveTabId` at call time (instead of only ever controlling `workflow_path` injection in `withWorkflowTarget`), so that when the original `ctx.tabId` itself goes stale, a session explicitly configured for "current" mode can still route commands to whatever tab is presently active — rather than being inert once the bound tabId is orphaned. This wouldn't fix the underlying migration gap, but would give users an explicit, discoverable escape hatch instead of a hard failure on every panel_* call.

Relevant files:
- `dist/services/ui-bridge.js` (tabMigrations Map decl ~L79-87; same-socket re-hello recording ~L392-413; resolveTarget fallback ~L676-714)
- `dist/orchestrator/panel-tools.js` (`makePanelToolCtx`, hardcoded `ctx.tabId` in `call()`, ~L229-240)
- `workflow-target-store.js` (confirms "current" mode only affects `workflow_path` injection, not tabId routing)

### Environment
OS: Windows 11 Pro N (10.0.26200)
comfyui-mcp: installed via npx at `C:\Users\artur\AppData\Local\npm-cache\_npx\28eae33d00f3f7f1\node_modules\comfyui-mcp`
ComfyUI: run from `F:\Ai Toolbox\ComfyUI\ComfyUI_windows_portable`

### LINKED ISSUE #332: Agent session permanently stuck on dead tabId after a full ComfyUI reconnect (not a same-socket re-hello) - panel_set_workflow_target and panel_reload can't recover it

An agent session's `tabId` used for all `panel_*` MCP tool calls is captured ONCE in a closure at `PanelToolCtx` creation time (`makePanelToolCtx(bridge, tabId, workflowTargets)` in `src/orchestrator/panel-tools.ts`), and every subsequent `call()` sends that frozen id to `UiBridge.send()` / `resolveTarget()` regardless of what happens afterward. `panel_set_workflow_target` (mode `"current"` or `"pinned"`) only writes routing metadata into a `workflowTargets` map keyed by that same frozen id - it does not, and cannot, change which literal `tabId` string is used for bridge routing. `panel_reload` is itself gated behind the same broken resolution, so it can't be used to recover either.

`UiBridge`'s `tabMigrations` mechanism (`src/services/ui-bridge.ts`, `resolveTarget()`) is the designed self-heal path for a changed tab id, but it ONLY records a migration on a SAME-SOCKET re-hello (the panel's browser WebSocket re-hellos under a new `tab_id` while staying the same underlying socket - e.g. switching ComfyUI workflow tabs live). A full ComfyUI restart, or a fresh browser tab reconnect, creates an entirely NEW socket, so no migration entry is ever recorded for that transition. The agent session is then permanently stuck: every `panel_*` call throws `no connected tab with id "<dead-id>"` for the rest of the session's lifetime, with no available recovery path from inside that session.

### Steps to reproduce

1. Start a panel agent session, connected to some ComfyUI browser tab (tabId captured in ctx).
2. Fully restart ComfyUI (e.g. `panel_restart_comfyui`), or otherwise cause the browser panel socket to drop and reconnect as a NEW WebSocket rather than a same-socket re-hello.
3. After reconnecting, any `panel_*` tool call (`panel_graph_outline`, `panel_set_workflow_target`, `panel_reload`, `panel_list_workflows`, ...) fails with `no connected tab with id "<old-tab-id>"`, even though a tab IS listed as connected and is reachable/working from the browser side (runs queued from the browser UI succeed normally).

### Exact error (scrubbed)

```
Error: no connected tab with id "3adaed5f-fa27-448f-9584-47a3b1bd67b8". Connected: tmp:091e (“Unsaved Workflow (2)”), wf:workf (“Upscale Bake-Off”)
```

Identical across `panel_graph_outline`, `panel_set_workflow_target`, `panel_open_workflow`, `panel_list_workflows`, `panel_reload` - all routed through the same frozen `ctx.tabId`.

### Fix

applied locally: no
upstream-only: no (this is comfyui-mcp orchestrator/bridge code, just not something I could safely fix myself this session)

I attempted a fix in `UiBridge.resolveTarget()`: when an explicit `tabId` has no exact/prefix/migration match, fall back to the same single-connection/`lastActiveTabId` logic already used when no `tabId` is supplied at all. This DOES resolve the reported symptom (verified: `npm run build` clean, and the dead-id path resolves instead of throwing) - but it REGRESSES an existing, deliberate security test:

`src/__tests__/services/ui-bridge.test.ts:459` - "a DEAD socket's migration alias never routes to an unrelated tab reusing the id (deterministic wf: reuse)" - asserts a dead/legacy `tabId` must NOT silently resolve to some other, unrelated tab that happens to be the sole/last-active connection. My fallback can't distinguish "this session's own tab reconnected fresh" from "an entirely unrelated tab happens to be the only one connected right now" - and the codebase's existing design (`tabMigrations` restricted to same-socket re-hellos, socket-scoped migration entries, the codex-review comments throughout `ui-bridge.ts`) is deliberately conservative about exactly this distinction. Weakening it at the `UiBridge` level to fix session recovery would be a real regression for the multi-tab / multi-viewer / remote-phone deployment cases this class is explicitly designed to support.

My diagnosis is the right fix lives one layer up, in the orchestrator (`src/orchestrator/panel-tools.ts`, `makePanelToolCtx`), not in `UiBridge`: `panel_set_workflow_target` already exists as the EXPLICIT, agent/user-initiated mechanism meant to "follow the active tab again" per its own tool description - but it currently only updates the `workflowTargets` map, not the frozen `ctx.tabId` closure that `call()` actually uses for bridge routing. A safe fix would make `ctx.tabId` mutable (or resolved through a small mutable holder object) and have `panel_set_workflow_target(mode:"current")` actually rebind it to the CURRENTLY resolved tab id (e.g. via `bridge.resolveTarget()`'s existing single-connection/`lastActiveTabId` fallback, invoked explicitly and deliberately at THAT moment as the intentional "yes, rebind me" signal) rather than baking an implicit fallback into every low-level `resolveTarget()` call. `panel_reload` would need the same rebind-on-explicit-call treatment so it can be used to recover a stuck session too.

I reverted my `ui-bridge.ts` attempt (`git checkout -- src/services/ui-bridge.ts`) rather than leave a change that builds clean but fails a deliberate security test.

### Environment

OS: Windows 11 (win32)
ComfyUI version: 0.28.3
GPU: NVIDIA GeForce RTX 2070, 8GB VRAM
comfyui-mcp: 0.48.2 (resolved npm install) / marketplace clone shows 0.48.1 in package.json
Deployment: ComfyUI Desktop, comfyui-agent-panel pack, ComfyUI-Manager, loopback bridge default port 9101

artokun:
## One root cause, three reported issues

An agent's panel MCP tools all route through a `tabId` captured **once** in a closure when the session is created (`makePanelToolCtx`). Every later `call()` sent that frozen id to `UiBridge.send()` / `resolveTarget()`.

When the panel's browser socket comes back as a **genuinely new socket with a new tab id**, no `tabMigrations` alias is recorded (that map only tracks *same-socket* re-hellos), so the session's frozen `ctx.tabId` is permanently orphaned and every `panel_*` call throws `no connected tab with id "<old-id>"` forever. The three duplicate issues are the same bug from three triggers:

- **#332** — a full ComfyUI reconnect
- **#322** — a frontend reload / `panel_reload({scope:"frontend"})`
- **#331** — switching to a different workflow **file**

`panel_set_workflow_target` previously only wrote the `workflowTargets` routing-metadata map — it never changed the tabId used for bridge routing — so it couldn't recover the session either.

## Approach: fix one layer up, NOT in resolveTarget

`UiBridge.resolveTarget()` is deliberately conservative. A naive "if no match and exactly one tab, use it" fallback there would **regress the security test** at `src/__tests__/services/ui-bridge.test.ts:459` ("a DEAD socket's migration alias never routes to an unrelated tab reusing the id") and the multi-tab / multi-viewer / remote-phone routing guarantees. `resolveTarget` is left untouched.

Instead the fix lives in the orchestrator (`src/orchestrator/panel-tools.ts`):

1. **Mutable tabId** — `makePanelToolCtx` holds `tabId` on the returned `ctx` object; `call()`/`confirm()` and every handler read `ctx.tabId` **live** rather than closing over the string.
2. **`rebindToActiveTab()`** — self-heals **only when** the current tabId no longer reaches a live tab (`bridge.canReach`), so a healthy/multi-tab session is never disturbed. It then re-points onto the active/sole tab via `bridge.resolveActiveTabId()` — the same no-tabId resolution a fresh session uses, invoked **explicitly at the consent moment**, not baked into `resolveTarget`. A pinned workflow target carries across to the new id.
3. **Explicit rebind signals** — `panel_set_workflow_target({mode:"current"})` ("follow the tab I'm looking at now") and `panel_reload` ("recover me now") both trigger the rebind. If a single active tab genuinely can't be determined (0 or 2+ ambiguous), they surface a **clear error** rather than guessing.
4. `UiBridge` gains two public helpers — `canReach(tabId)` (non-throwing) and `resolveActiveTabId()` (the no-tabId resolution, throwing the same clear errors) — both delegating to the untouched `resolveTarget`.

## Tests

- `panel-tools.test.ts`: a dead-session `panel_set_workflow_target({mode:"current"})` and `panel_reload` each rebind onto the live tab and subsequent calls succeed; a healthy multi-tab session is left untouched; an ambiguous dead case surfaces a clear error without changing tabId; a pinned target carries across the rebind.
- `ui-bridge.test.ts`: `canReach`/`resolveActiveTabId` back the reconnect (new socket, no migration) scenario, and `resolveActiveTabId` throws (no guess) on 2+ tabs with no last-active. **The security test at line 459 stays green.**

Test summary (`npm test`): **1936 passed | 1 skipped**, plus 1 **pre-existing, unrelated** failure in `node-dev.test.ts` (asserts the `builtin` search fallback but this machine has ripgrep installed, so it reports `ripgrep`; verified it fails identically on the clean base tree, without any of this PR's changes). `npm run build` (tsc) is clean.

Closes #322
Closes #331
Closes #332

### DIFF
--- a/src/__tests__/orchestrator/panel-tools.test.ts
+++ b/src/__tests__/orchestrator/panel-tools.test.ts
@@ -528,6 +528,108 @@ describe("panel-tools: workflow target (per-workflow agent)", () => {
   });
 });
 
+describe("panel-tools: session tabId self-heal (#322 reload / #331 workflow-switch / #332 reconnect)", () => {
+  // A minimal fake bridge modelling the ONE fact that matters: which tab ids are
+  // currently live. `send` routes only to a live id (throws `no connected tab`
+  // otherwise, exactly like UiBridge.resolveTarget); canReach/resolveActiveTabId
+  // mirror the real bridge's no-throw / no-tabId helpers.
+  function fakeBridge(live: Set<string>) {
+    const sent: Array<{ cmd: Record<string, unknown>; tabId?: string }> = [];
+    const bridge = {
+      send: async (cmd: Record<string, unknown>, opts?: { tabId?: string }) => {
+        const id = opts?.tabId;
+        if (id && !live.has(id)) throw new Error(`no connected tab with id "${id}"`);
+        sent.push({ cmd, tabId: id });
+        return { ok: true, routedTo: id };
+      },
+      push: () => 1,
+      canReach: (tabId: string) => live.has(tabId),
+      resolveActiveTabId: () => {
+        if (live.size === 1) return [...live][0];
+        if (live.size === 0) throw new Error("Panel not reachable: no panel connected");
+        throw new Error("Multiple panel tabs are connected and none is last active — pass tab_id.");
+      },
+    } as unknown as PanelToolCtx["bridge"];
+    return { bridge, sent };
+  }
+
+  it("panel_set_workflow_target({mode:'current'}) rebinds a DEAD session onto the sole live tab; calls then succeed", async () => {
+    const store = new WorkflowTargetStore();
+    // Only the NEW tab is live; the session was created bound to the old (dead) id.
+    const { bridge, sent } = fakeBridge(new Set(["new-live-tab"]));
+    const ctx = makePanelToolCtx(bridge, "dead-old-tab", store);
+
+    // Before rebind, a graph call to the dead tab fails.
+    const before = await defByName("panel_graph_outline").handler({}, ctx);
+    expect(before.isError).toBe(true);
+
+    const res = await defByName("panel_set_workflow_target").handler({ mode: "current" }, ctx);
+    expect(res.isError).toBeUndefined();
+    expect(ctx.tabId).toBe("new-live-tab");
+    const text = (res.content[0] as { text: string }).text;
+    expect(text).toContain("Rebound this session");
+
+    // After rebind, subsequent panel_* calls route to the live tab and succeed.
+    const after = await defByName("panel_graph_outline").handler({}, ctx);
+    expect(after.isError).toBeUndefined();
+    expect(sent.at(-1)?.tabId).toBe("new-live-tab");
+  });
+
+  it("panel_reload rebinds a DEAD session onto the sole live tab, then forwards soft_reload there", async () => {
+    const store = new WorkflowTargetStore();
+    const { bridge, sent } = fakeBridge(new Set(["reconnected-tab"]));
+    const ctx = makePanelToolCtx(bridge, "orphaned-tab", store);
+
+    const res = await defByName("panel_reload").handler({}, ctx);
+    expect(res.isError).toBeUndefined();
+    expect(ctx.tabId).toBe("reconnected-tab");
+    expect(sent.at(-1)).toMatchObject({
+      cmd: { cmd: "soft_reload", scope: "orchestrator" },
+      tabId: "reconnected-tab",
+    });
+  });
+
+  it("does NOT disturb a HEALTHY session: mode:'current' leaves a still-live tabId in place", async () => {
+    const store = new WorkflowTargetStore();
+    // Two live tabs incl. the session's own — a healthy multi-tab deployment.
+    const { bridge } = fakeBridge(new Set(["my-tab", "other-tab"]));
+    const ctx = makePanelToolCtx(bridge, "my-tab", store);
+
+    const res = await defByName("panel_set_workflow_target").handler({ mode: "current" }, ctx);
+    expect(res.isError).toBeUndefined();
+    // Untouched — no rebind (canReach true), so no hijack onto another tab.
+    expect(ctx.tabId).toBe("my-tab");
+    const text = (res.content[0] as { text: string }).text;
+    expect(text).not.toContain("Rebound");
+  });
+
+  it("surfaces a CLEAR error (no guess) when the session is dead AND no single active tab exists", async () => {
+    const store = new WorkflowTargetStore();
+    // Session's tab is dead and 2+ others are live with no last-active → ambiguous.
+    const { bridge } = fakeBridge(new Set(["a", "b"]));
+    const ctx = makePanelToolCtx(bridge, "dead-tab", store);
+
+    const res = await defByName("panel_set_workflow_target").handler({ mode: "current" }, ctx);
+    expect(res.isError).toBe(true);
+    expect((res.content[0] as { text: string }).text).toMatch(/Multiple panel tabs/);
+    // tabId is NOT silently changed to some arbitrary tab.
+    expect(ctx.tabId).toBe("dead-tab");
+  });
+
+  it("carries a PINNED workflow target across the rebind to the new tab id", async () => {
+    const store = new WorkflowTargetStore();
+    store.set("dead-old-tab", { mode: "pinned", path: "workflows/pinned.json" });
+    const { bridge } = fakeBridge(new Set(["new-live-tab"]));
+    const ctx = makePanelToolCtx(bridge, "dead-old-tab", store);
+
+    // panel_reload triggers the rebind without releasing the pin.
+    await defByName("panel_reload").handler({}, ctx);
+    expect(ctx.tabId).toBe("new-live-tab");
+    expect(store.get("new-live-tab")).toMatchObject({ mode: "pinned", path: "workflows/pinned.json" });
+    expect(store.get("dead-old-tab")).toMatchObject({ mode: "current" });
+  });
+});
+
 describe("panel_connect slot aliases (live panel finding: stripped aliases → auto-match scramble)", () => {
   it("maps from_slot_name/to_slot_name onto from_output/to_input on the wire", async () => {
     const { ctx, calls } = makeFakeCtx();
--- a/src/__tests__/services/ui-bridge.test.ts
+++ b/src/__tests__/services/ui-bridge.test.ts
@@ -478,6 +478,46 @@ describe("UiBridge (multi-tab)", () => {
     sockB.close();
   });
 
+  it("canReach + resolveActiveTabId back the orchestrator's explicit self-heal (#322/#331/#332)", async () => {
+    // A session was bound to old-tab, which then DIES and a genuinely new socket
+    // reconnects under new-tab — NO migration alias (different socket, first hello).
+    const sockOld = await connectPanel();
+    autoReply(sockOld, "old");
+    sockOld.send(JSON.stringify({ type: "hello", tab_id: "old-tab" }));
+    await vi.waitFor(() => expect(bridge.tabs()).toHaveLength(1));
+    sockOld.close();
+    await vi.waitFor(() => expect(bridge.tabs()).toHaveLength(0));
+
+    const sockNew = await connectPanel();
+    autoReply(sockNew, "new");
+    sockNew.send(JSON.stringify({ type: "hello", tab_id: "new-tab" }));
+    await vi.waitFor(() => expect(bridge.tabs()).toHaveLength(1));
+
+    // The old id is orphaned (no migration) — canReach is false, and the no-tabId
+    // resolution the orchestrator invokes at the explicit rebind lands on the sole
+    // live tab. resolveTarget itself is NOT weakened: the old id still throws.
+    expect(bridge.canReach("old-tab")).toBe(false);
+    expect(bridge.canReach("new-tab")).toBe(true);
+    expect(bridge.resolveActiveTabId()).toBe("new-tab");
+    await expect(bridge.send({ cmd: "x" }, { tabId: "old-tab" })).rejects.toThrow(/no connected tab/);
+
+    sockNew.close();
+  });
+
+  it("resolveActiveTabId throws (no guess) when 2+ tabs are connected with no last-active", async () => {
+    const sockA = await connectPanel();
+    autoReply(sockA, "A");
+    sockA.send(JSON.stringify({ type: "hello", tab_id: "tab-a" }));
+    const sockB = await connectPanel();
+    autoReply(sockB, "B");
+    sockB.send(JSON.stringify({ type: "hello", tab_id: "tab-b" }));
+    await vi.waitFor(() => expect(bridge.tabs()).toHaveLength(2));
+
+    expect(() => bridge.resolveActiveTabId()).toThrow(/Multiple panel tabs/);
+    sockA.close();
+    sockB.close();
+  });
+
   it("follows MIGRATION CHAINS: uuid → tmp: → wf: (the exact #210 field sequence)", async () => {
     // The reported failure re-helloed TWICE: legacy random UUID, then the
     // unsaved-tab tmp:<uuid> id, then the saved wf:<hash> id. The ORIGINAL id
--- a/src/orchestrator/panel-tools.ts
+++ b/src/orchestrator/panel-tools.ts
@@ -285,6 +285,21 @@ export interface PanelToolCtx {
   tabId: string;
   /** Per-tab workflow pin store (optional for tests). */
   workflowTarget?: WorkflowTargetStore;
+  /**
+   * EXPLICIT self-heal: re-point THIS session at the currently active/sole
+   * connected tab. The tabId captured at session creation is frozen; a full
+   * ComfyUI reconnect (#332), a frontend reload (#322), or switching to a
+   * different workflow FILE (#331) can surface a brand-NEW browser socket under
+   * a NEW tab id with no migration alias, orphaning the session so every
+   * panel_* call throws `no connected tab`. This rebinds `ctx.tabId` (which
+   * `call`/`confirm` read LIVE) to the active tab — but ONLY when the current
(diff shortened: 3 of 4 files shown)


artokun:
## Codex review (gpt-5-codex)

**Verdict: NEEDS CHANGES — non-blocking.** No security regression found: the fix wraps `UiBridge.resolveTarget` via new `canReach`/`resolveActiveTabId` helpers without altering it, and the invariant test at `ui-bridge.test.ts:459` is untouched. Both explicit signals (`panel_set_workflow_target({mode:"current"})`, `panel_reload`) error clearly on the ambiguous/no-single-active-tab case rather than guessing (confirmed by tests). Self-heal correct for reconnect and frontend-reload; workflow-file switch heals via the same explicit-recovery path.

Non-blocking suggestions from Codex (left for maintainer review):

1. **`canReach` readyState** — `canReach` delegates to `resolveTarget` (presence in `conns`), while `send`'s actual write also needs `readyState === OPEN`. A CLOSING-but-present socket could read as reachable and skip a rebind. Minor edge: `canReach` is intentionally identical to `resolveTarget`'s acceptance so "reachable" means "resolveTarget would succeed"; if the post-resolve `send` still fails the tool returns an error and the agent re-issues the recovery signal.
2. **Mutable `ctx.tabId` concurrency** — the orchestrator runs a session's tool calls sequentially, so the mutable holder isn't raced in practice; flagged as a documentation/assumption to make explicit.
3. **Document `mode:"current"` as explicit-recovery** — the tool description already frames it as both "follow the active tab" and the recovery signal.

Codex found **no blocking issue** and confirmed the deliberate resolveTarget-untouched design. Leaving these three for @artokun's call.

Issue #376, opened 2026-07-29
get_node_info returns fetch failed immediately after panel_restart_comfyui reconnect


artokun:
### What happened / root cause
After installing ComfyUI-LTXVideo with panel_install_node and restarting through panel_restart_comfyui, panel_set_workflow_target(mode=current) succeeded, but get_node_info(node_type="LTXV", verbose=false) returned only `fetch failed`. The tool did not provide HTTP status, endpoint, retry hint, or reconnect recovery. This looks like a post-restart reconnect race or stale ComfyUI client in comfyui-mcp.

### Steps to reproduce
1. Install a custom node through panel_install_node.
2. Restart using panel_restart_comfyui.
3. Rebind with panel_set_workflow_target(mode="current").
4. Immediately call get_node_info({node_type:"LTXV",verbose:false}).

### Exact error (scrubbed)
`fetch failed`

### Fix
applied locally: no  upstream-only: yes
The get_node_info transport should refresh/retry the ComfyUI base client after a managed restart, and return the underlying endpoint/status/cause when fetch still fails.

### Environment
Windows 11 / ComfyUI 0.29.0 / RTX 4090 24 GB / comfyui-mcp 0.48.5 / panel 0.11.4

artokun:
Fixed in **comfyui-mcp 0.48.6** (WS-2, #383) — the relaunch/model-dir/extra-paths logic now resolves the live ComfyUI server's actual base directory (`--base-directory`/`--models-directory`/`--extra-model-paths-config`) instead of a possibly-stale COMFYUI_PATH, and restart preflights before stopping. Please update (npx/global `comfyui-mcp@latest`) and reopen if it recurs on 0.48.6+.

Issue #369, opened 2026-07-29
download_model lands files in a stale COMFYUI_PATH install, not the live (Desktop) ComfyUI model dirs - reports success, model never appears


artokun:
### What happened / root cause

`download_model` resolves its destination from `COMFYUI_PATH` (the local workspace it detected) instead of from the **connected/live** ComfyUI server's actual model search paths.

On this machine the live server is **ComfyUI Desktop** (api target `http://127.0.0.1:8000`), whose models live under `base_path: C:\ComfyUI` per `~/AppData/Roaming/ComfyUI/extra_models_config.yaml`. But `get_environment` reports `workspace_path: ~/ComfyUI-Installs/ComfyUI/ComfyUI` (a stale/second install), and `download_model` wrote there.

Result: the download genuinely completed (3.79/3.79 GB, `download_status` = **done**, with a `landed at:` path), but the model was invisible to the running server - so it looked like a stale-cache problem rather than a wrong-directory problem. Notably `list_local_models` reads the live server's `/models` REST endpoint while `download_model` writes to `COMFYUI_PATH`, so the two tools disagree about where models live.

### Steps to reproduce

1. Run ComfyUI **Desktop** with models on a different `base_path` (here `C:\ComfyUI`), while a second/older manual ComfyUI checkout also exists on disk.
2. `download_model(url=<any model>, target_subfolder="diffusion_models")`.
3. `download_status` -> `done`, `landed at: <stale install>/models/diffusion_models/<file>`.
4. `list_local_models(model_type="diffusion_models")` -> file is **absent**.
5. `get_node_info("UNETLoader", verbose=true)` -> enum still lacks the file (so it is not a client-side cache).
6. Set the node's `unet_name` to the downloaded file and queue -> validation error, node red on canvas.

### Exact error (scrubbed)

From `panel_run`:

```
"node_errors": {
  "1": {
    "errors": [{
      "type": "value_not_in_list",
      "message": "Value not in list",
      "details": "unet_name: 'flux-2-klein-4b-fp8.safetensors' not in (list of length 24)"
    }],
    "class_type": "UNETLoader"
  }
}
```

Panel missing-asset banner:

```
node 1: MODEL missing - flux-2-klein-4b-fp8.safetensors (belongs in models/diffusion_models) [widget unet_name]
```

`download_status` (misleadingly successful):

```
- ef17d8915cf65f1d done  3.79/3.79 GB (100%)
[...]

Issue #368, opened 2026-07-29
restart_comfyui stops ComfyUI Desktop but cannot relaunch it


artokun:
### What happened / root cause
The restart_comfyui tool stopped a ComfyUI Desktop-managed instance, then failed to restart it because it could not determine the Desktop executable path. This leaves the user's ComfyUI offline and requires a manual launch.

### Steps to reproduce
1. Use a ComfyUI Desktop installation on Windows.
2. Call restart_comfyui.
3. Observe that the process is stopped but not started.

### Exact error (scrubbed)
ComfyUI was stopped but could not be started: Could not determine ComfyUI Desktop executable path. Please start it manually.

### Fix
Applied locally: no. Upstream-only: yes.
The tool should resolve the Desktop launcher before stopping the running instance, or refuse without stopping when a relaunch path is unavailable.

### Environment
Windows 11; ComfyUI 0.29.0; RTX 5080 16 GB; comfyui-mcp 0.48.5; panel 0.11.4.

artokun:
Fixed in **comfyui-mcp 0.48.6** (WS-2, #383) — the relaunch/model-dir/extra-paths logic now resolves the live ComfyUI server's actual base directory (`--base-directory`/`--models-directory`/`--extra-model-paths-config`) instead of a possibly-stale COMFYUI_PATH, and restart preflights before stopping. Please update (npx/global `comfyui-mcp@latest`) and reopen if it recurs on 0.48.6+.

artokun:
Recurrence reported on current releases: comfyui-mcp 0.49.5 and comfyui-mcp-panel 0.11.38 (Windows 11, ComfyUI 0.30.0, RTX 5090). After `stop_comfyui` successfully stopped a Desktop-managed install, `start_comfyui` using the captured restart information returned `Could not determine ComfyUI Desktop executable path. Please start it manually.` and did not relaunch the server. This matches the failure previously closed as fixed in #383 / 0.48.6, but occurs on a newer release.

artokun:
Additional recurrence on Windows 11 with ComfyUI 0.30.0 and RTX 5090 32 GB: `stop_comfyui` returned `stopped:true` and `has_restart_info:true` for a Desktop-managed process, but the immediate `start_comfyui` failed with `Could not determine ComfyUI Desktop executable path. Please start it manually.` The reporter has comfyui-mcp 0.49.5; panel is 0.11.38.

artokun:
[...]

Issue #367, opened 2026-07-29
restart_comfyui reports failed startup just before a healthy instance becomes ready


artokun:
### What happened / root cause
restart_comfyui reported `started:false` and `ready:false` after a 19.2-second readiness timeout, but the new ComfyUI process became reachable on port 8188 about 21 seconds after launch.

### Steps to reproduce
1. Restart a ComfyUI instance with normal custom-node startup time.
2. Observe the 20 x 1-second readiness probe time out.
3. Check again shortly afterward; the instance is healthy.

### Exact error (scrubbed)
`ComfyUI process was launched but the API did not become ready after 19152ms (20/20 probes).`

### Fix
Applied locally: no. Upstream-only: yes. Extend readiness polling or distinguish timeout from a failed relaunch.

### Environment
Windows 11; ComfyUI 0.29.0 local; RTX 5090 32 GB; comfyui-mcp 0.48.5; panel version unavailable in environment.

artokun:
Fixed in **comfyui-mcp 0.48.6** (WS-2, #383) — the relaunch/model-dir/extra-paths logic now resolves the live ComfyUI server's actual base directory (`--base-directory`/`--models-directory`/`--extra-model-paths-config`) instead of a possibly-stale COMFYUI_PATH, and restart preflights before stopping. Please update (npx/global `comfyui-mcp@latest`) and reopen if it recurs on 0.48.6+.

artokun:
Recurrence on **comfyui-mcp 0.49.4** (panel 0.11.38) on Linux with ComfyUI 0.30.0 and an NVIDIA GB10 (124546 MB VRAM). `restart_comfyui` launched the process but timed out after 59,144 ms / 60 probes; a follow-up `get_environment` a few seconds later reached the same launched PID, healthy. This shows the prior 20→60-probe change is still insufficient for this cold-start configuration.

PR #383, opened 2026-07-29
fix(desktop): resolve live ComfyUI base dir for relaunch/model-dirs/extra-paths (WS-2)

### LINKED ISSUE #370: restart_comfyui stops the running server then fails to relaunch it when COMFYUI_PATH points at a stale install

`restart_comfyui` **stopped the user's running ComfyUI and could not bring it back**, leaving the server down until it was relaunched by hand.

Root cause is a path-resolution mismatch. `COMFYUI_PATH` pointed at a stale install (`~/ComfyUI`) while the ComfyUI actually serving `http://127.0.0.1:8188` was a portable install at `D:\ComfyUI_windows_portable\ComfyUI`. `get_workspace` reported the stale path with `workspace_source: "env"` and no warning that nothing was running there:

```json
{
  "workspace_path": "~/ComfyUI",
  "workspace_source": "env",
  "api_target": "http://127.0.0.1:8188"
}
```

The stale directory was real but nearly empty — its `custom_nodes/` held 4 packs, none of which matched the 11 packs `list_installed_nodes` reported from the live server. So the two views of "the install" disagreed, and nothing surfaced that.

`restart_comfyui` then killed the live process (correctly, via the API target) but attempted the relaunch against `COMFYUI_PATH`, which has no runnable server. It reported the failure honestly but had already taken the server down:

```json
{
  "stopped": true,
  "started": false,
  "ready": false,
  "message": "ComfyUI was stopped but could not be started: ComfyUI process was launched but the API did not become ready after 19208ms (20/20 probes)."
}
```

Recovery required manually relaunching from the real install:

```powershell
Start-Process -FilePath "D:\ComfyUI_windows_portable\python_embeded\python.exe" \
  -ArgumentList "-s","ComfyUI\main.py","--windows-standalone-build" \
  -WorkingDirectory "D:\ComfyUI_windows_portable"
```

The destructive half succeeded and the recovery half failed — the worst split. A restart should be atomic-ish: if the relaunch command can't be determined, don't stop the server.

### Steps to reproduce

1. Have ComfyUI actually running from install A (e.g. a portable build on another drive).
2. Set `COMFYUI_PATH` to a different, stale install B that also exists on disk but isn't running.
3. Point the MCP at the live server (`--comfyui-url http://127.0.0.1:8188`).
4. Call `restart_comfyui`.
5. The live server from A is stopped; the relaunch is attempted against B and never becomes ready. ComfyUI stays down.

### Exact error (scrubbed)

```
restart_comfyui →
  stopped: true, started: false, ready: false
  readiness: { timed_out: true, attempts: 20, max_tries: 20, waited_ms: 19208,
               probe_url: "http://127.0.0.1:8188/system_stats" }
  message: "ComfyUI was stopped but could not be started: ComfyUI process was
            launched but the API did not become ready after 19208ms (20/20 probes)."
```

### Fix

Not locally patched — the immediate need was restoring the user's server, and the durable fix belongs in the tool's path resolution rather than in a patched `dist/` on one machine. Precise changes needed:

1. **Resolve the running instance's real path before stopping anything.** Derive it from the live server rather than from `COMFYUI_PATH` — e.g. the running process's command line for the PID bound to the API port, or a path reported by the server itself. `COMFYUI_PATH` should be a last-resort fallback, not the primary source, whenever a live API target is reachable.
2. **Refuse to stop when the relaunch is unknown.** Gate the stop on having a validated relaunch command (resolved interpreter + `main.py` that exist on disk). If it can't be built, return an error and leave the server running. Losing a restart is cheap; losing the server is not.
3. **Warn on divergence in `get_workspace`.** When a live API target is reachable and the resolved `workspace_path` doesn't look like that server's install (no `main.py`, or `custom_nodes/` disjoint from `list_installed_nodes`), surface it in the result. This mismatch was invisible until it caused an outage, and it also silently misroutes `download_model`'s local-streaming target — model files would have landed under the stale path.
4. **On a failed relaunch, attempt rollback** using the captured pre-stop config, and say plainly in the message that the server is currently down and how to start it manually.

### Secondary issues hit in the same flow

Both on **ComfyUI-Manager 3.x legacy**, which the environment line already flags as a partial feature set. The tools don't degrade gracefully against it:

- `panel_install_node` → `Manager manager/queue/task: HTTP 405`. Bare status code, no hint that the legacy Manager lacks the queue API or that a fallback path exists.
- `panel_restart_comfyui` → `POST /v2/manager/reboot → HTTP 405; GET /manager/reboot → HTTP 404`. This one is correctly non-destructive and its message is genuinely useful — worth keeping as the model for the others.
- `install_custom_node` **reported success with zero-valued counters**, making success unverifiable from the response alone:

  ```json
  {
    "mechanism": "manager-http",
    "message": "Installed \"ComfyUI-Impact-Subpack\" via ComfyUI-Manager. Restart may be required to load new nodes.",
    "details": { "total_count": 0, "done_count": 0, "in_progress_count": 0, "is_processing": false }
  }
  ```

  The install had in fact worked, but only a manual `ls` of the real `custom_nodes/` could confirm it — and that check is exactly what surfaced the stale-path problem. Suggest either verifying the target directory landed before claiming "Installed", or downgrading the wording to "install dispatched" when the counters are all zero.

### Related friction (separate, low priority)

After ComfyUI restarts while a panel tab is open, the frontend keeps its cached `/object_info`, so `panel_add_node` rejects a newly installed node type that the server already knows:

```
Error: Unknown node type "UltralyticsDetectorProvider" — check the exact class_type via panel_get_graph or panel_search_nodes
```

`get_node_info` confirmed the server had the node loaded. The error sends you to re-check the class name, which is a dead end since the name was right — the real fix is a browser refresh. A hint like "the frontend's node registry may be stale after a restart — reload the ComfyUI tab" would save the detour, or the panel could refetch `/object_info` on reconnect.

### Environment

- ComfyUI: 0.20.1 (local, portable build)
- ComfyUI-Manager: 3.x legacy
- GPU: NVIDIA GeForce RTX 5090, 32 GB VRAM (CUDA 13.0, torch 2.11.0, python 3.13)
- comfyui-mcp: 0.48.5
- panel: 0.11.4

artokun:
## Root cause

A whole family of Desktop bugs shares one cause: several path-resolving tools trust `COMFYUI_PATH` instead of the **running server's actual base directory**. On ComfyUI Desktop the live server is launched with `--base-directory` / `--extra-model-paths-config` pointing at a tree that differs from the configured `COMFYUI_PATH` (frequently a stale or second install). So relaunch spawned the wrong/stale path, model downloads + extra_paths wrote where the live server never reads, and readiness/reconnect probes raced. The authoritative source is ComfyUI itself — `/system_stats` reports the launch `argv` — which the output/input dir resolution already used; this PR extends the same approach to relaunch, model dirs, and the extra-paths YAML.

## Changes and issues closed

- **process-control (`#368`, `#370`, `#367`)**
  - `restart_comfyui` now runs a **relaunch preflight before stopping anything**: it resolves the running instance and validates a spawnable command (Desktop launcher exe located on disk, or resolved interpreter + `main.py` both exist). If it can't build/validate a relaunch it **refuses and leaves the server running** rather than stopping it with no way back. This makes restart atomic-ish — losing a restart is cheap, losing the server is not.
  - For Desktop, the preflight also resolves + validates the launcher via common install paths (fixes `#368`, where the "Comfy Desktop" branding never matched the old regex).
  - For a stale `COMFYUI_PATH` whose `main.py` doesn't exist, restart refuses instead of killing the live portable server (`#370`).
  - Readiness budget raised **20s → 60s** (`COMFYUI_STARTUP_CHECK_MAX_TRIES` default 20→60): custom-node cold starts routinely exceed 20s, so it reported `started:false` moments before a healthy instance became ready (`#367`).
- **output-dir**: new argv parsers `parseBaseDirFromArgv` / `parseModelsDirFromArgv` / `parseExtraModelPathsConfigsFromArgv` and async `resolveModelsDir()` / `resolveServerExtraModelConfig()` — source of truth is `/system_stats` argv, fallback `COMFYUI_PATH`.
- **model-resolver (`#369`, `#346`)**: `download_model` roots the destination at the **live server's** `--base-directory/models` (via `resolveModelSubfolderPreferServer`), not blindly `COMFYUI_PATH/models`, so files land where the running server sees them.
- **extra-paths (`#345`)**: `list_extra_paths` / `add_extra_path` / `remove_extra_path` prefer the file the running server actually loads (`--extra-model-paths-config`) instead of a hardcoded app-data guess. `list` now reflects reality (no more fiction), mutations hit the live file (no more silent no-op), and it warns loudly when the server's config is the Desktop auto-generated `shared_model_paths.yaml` ("do not edit manually").
- **client (`#376`)**: `getObjectInfo` resets the stale client and retries once on a transport failure, so `get_node_info` recovers from a torn-down socket after a managed restart instead of returning a bare `fetch failed`.

## Tests

- New/updated regression tests: `output-dir` argv parsers + `resolveModelsDir`/`resolveServerExtraModelConfig`; `process-control` restart preflight refuses-to-stop for a stale script path and for an unlocatable Desktop exe (asserts no kill / no relaunch / no client reset); `extra-paths` prefers the server's `--extra-model-paths-config` and warns; `object-info-cache` reset+retry behavior; `model-resolver` rooted at the stubbed server models dir.
- `npm run build` clean. `npm test`: all pass **except** one pre-existing, environment-dependent failure in `node-dev.test.ts` ("builtin fallback ... skips binary files") that also fails on the base commit (it expects the builtin search engine but ripgrep is installed on this machine) — unrelated to this change.

## Notes / follow-ups

- `get_workspace` divergence warning (item 3 of `#370`) and the "install dispatched" wording for zero-counter Manager installs are not in this PR — separable, lower-risk follow-ups.
- Panel `#188` (`panel_restart_comfyui` doesn't return) is panel-side JS and is not addressed here.

fix: resolve live ComfyUI base dir for Desktop (closes #370,#368,#369,#346,#345,#376,#367) [reports ≤mcp 0.48.5]


artokun:
### Codex review — resolved

Ran a Codex adversarial review (disclosed). It surfaced two genuine factual bugs I've now fixed in `e6d956c`:

- **`--models-directory`** — ComfyUI *does* have this flag (verified against `comfy/cli_args.py`); it overrides the models folder under `--base-directory`. `parseModelsDirFromArgv` now honors it, so `download_model` lands correctly for servers launched with it.
- **`--extra-model-paths-config` arity** — the flag is `nargs='+', action='append'` (multiple values per flag *and* repeatable). The parser previously grabbed only the next token; it now collects all of them.
- Also: `assessRelaunch` now existence-checks the macOS app bundle (parity with the Windows exe check), and the server-resolved extra-paths config is labeled `standalone`/`desktop` by path shape rather than force-`desktop`, with the contradictory generic note dropped.

Acknowledged-but-deferred (noted, not regressions introduced here):
- `findComfyuiPython` still prefers `COMFYUI_PATH` when resolving the interpreter/cwd for relaunch. The refuse-to-stop gate (validates `main.py` exists) is the primary safety; fully sourcing the interpreter from the live argv is a deeper follow-up.
- `getObjectInfo` reset+retry fires on any `getNodeDefs` error, not only transport drops — accepted tradeoff (reset is cheap; the alternative is fragile error-string classification).

PR #397, opened 2026-07-29
fix(chat): serialize structured backend/user/canvas payloads instead of [object Object] (WS-9)

### LINKED ISSUE #168: Live canvas context becomes [object Object] and panel tools are missing
A sidebar-agent turn received the live-canvas context as the literal string `[object Object]`, and the session tool inventory contained no `panel_*` tools (including `panel_graph_outline` and `panel_query_graph`). The panel was installed and ComfyUI was reachable, so the likely root cause is failed panel-context serialization and/or live-panel tool registration during the panel/orchestrator handshake.

### Steps to reproduce
1. Open ComfyUI with the comfyui-mcp sidebar panel.
2. Keep any workflow tab active.
3. Ask the sidebar agent: `?????????`
4. Observe the agent input begins with `[object Object]`.
5. Inspect the agent tool inventory: no `panel_*` tools are registered, so the live canvas cannot be queried.

### Exact error (scrubbed)
No thrown error. Silent/malformed behavior:
- Injected context: `[object Object]`
- Missing expected tools: `panel_graph_outline`, `panel_query_graph`, and all other `panel_*` calls.
- Panel status: installed and applicable.
- ComfyUI API reachable; queue empty.

### Fix

Ensure structured canvas/session context is JSON-serialized before composing the user turn instead of relying on JavaScript object string coercion. Also surface/register live `panel_*` tools whenever the sidebar handshake is active; if registration fails, emit an explicit diagnostic rather than silently falling back to headless-only tools.

### Environment
Windows 10
ComfyUI 0.28.0
NVIDIA GeForce RTX 5060 Ti 16 GB
comfyui-mcp 0.48.5
comfyui-mcp-panel 0.11.3

### LINKED ISSUE #175: User message attachments/objects stringify as [object Object] into agent prompt
Panel/orchestrator handoff to the agent turned a non-text user payload into the literal string `[object Object]` in the model prompt, so the agent lost the prior message/attachment context. User then replied agreeing something was strange (`Ja det er da mærkeligt`) without recoverable prior content.

Likely `String(obj)` / template coercion on a structured chat part (attachment, card click, media, or multi-part message) instead of JSON/text extraction.

### Steps to reproduce
1. Open ComfyUI agent panel chat.
2. Send a message that includes a non-plain-text part (attachment, structured UI payload, or multi-part content) — or continue a thread after such a part.
3. Observe agent turn receives user content as literally `[object Object]` plus any following plain text.

### Exact error (scrubbed)
Prompt user section ended as:
```
The user's first message follows.

[object Object]

Ja det er da mærkeligt
```
No exception; silent context loss.

### Fix
Serialize each chat part by kind: text → string; image → panel media ref / description; card click → stable label; never default Object.prototype.toString.

### Environment
OS darwin · GPU mps 48GB · ComfyUI 0.29.0 · comfyui-mcp 0.48.5 · panel 0.11.4 · Backend label in ENV: Claude (session actually Grok Build) · torch 2.13 · python 3.12

### LINKED ISSUE #176: Render structured backend failure as [object Object] in panel chat
After a Claude/Antigravity backend turn failed with an individual quota error, the sidebar chat rendered an additional `[object Object]` message instead of readable text. The panel appears to stringify a structured error/message object with implicit JavaScript object coercion.

### Steps to reproduce
1. Use a backend that returns a structured failure object, observed with Claude Opus via Antigravity.
2. Exhaust the provider quota so the turn fails.
3. Observe the sidebar chat after the readable quota error.
4. An extra `[object Object]` bubble is displayed.

### Exact error (scrubbed)
`agy exited with code 1. Error: Individual quota reached. Please upgrade your subscription to increase your limits.`
Then: `[object Object]`

### Fix
Normalize structured chat payloads before rendering: prefer a string message/error field; otherwise serialize safely instead of implicit object coercion. Add a regression test for structured backend failures.

### Environment
darwin / ComfyUI 0.29.0 / MPS 128 GB / comfyui-mcp 0.48.5 / panel 0.11.4.

artokun:
Fixes three "[object Object]" serialization reports where structured objects were String()-coerced instead of properly serialized before landing in chat text / the model prompt.

Closes artokun/comfyui-mcp-panel#176
Closes artokun/comfyui-mcp-panel#175
Closes artokun/comfyui-mcp-panel#168

## Root cause per issue

**#176 — structured backend failure renders as `[object Object]`** (FIXED here).
The orchestrator's canvas run-error forwarding (`PanelAgentManager.injectRunError` → `PanelAgent.injectRunError`, and `PanelAgent.injectEvent({kind:"run_error"})`) interpolated the incoming `error` straight into the chat/prompt text via a template literal. The panel forwards ComfyUI's execution error, which can be a **structured object** (`{node_id, node_type, exception_message, …}`), so `${error}` produced `[object Object]`. Fixed with a shared `toReadableText()` helper (prefers a string `message`/`error`/`text`/`detail` field; else JSON-serializes; never `Object.prototype.toString`). The backend `error` event detail also falls back to it for a non-string message.

**#175 — user attachment/multi-part becomes `[object Object]` in the prompt** (LIVES IN PANEL).
The orchestrator only ever receives `event.text` as an already-composed **string** (it guards `typeof event.text !== "string"` and rejects otherwise), so the coercion of a structured user part happens in the panel while it composes the outgoing turn text, not in this repo. No orchestrator code builds the user turn from structured parts. Flagged for the panel-side agent.

**#168 — live-canvas context becomes `[object Object]` + panel_* tools missing** (SPLIT).
- *Serialization half* — LIVES IN PANEL. The live-canvas / "MANUAL CANVAS CHANGES" context block is composed panel-side and arrives inside the string `event.text`; the orchestrator's own context injection (transcript replay) is already string-guarded. Flagged for the panel-side agent.
- *Tool-registration half* — FIXED here, and it is **independent**, not a cascade. Non-Claude backends get their `panel_*` tools from the loopback HTTP panel-MCP (`makeHttpBackendMcpServers` includes the `panel` server only when `panelMcpHttp` started). If that server fails to bind (port in use — plausible on Windows per the existing bind-coexistence note), every non-Claude tab silently degrades to headless-only tools. There is no shared throw/early-return between context composition and tool registration, so the two symptoms are coincidental, not causal. Added an explicit per-tab, first-turn chat diagnostic when `panelMcpHttp` is unavailable for a non-Claude tab, instead of failing silently. (Claude tabs host `panel_*` in-process and are unaffected.)

## Tests
- `src/__tests__/utils/to-readable-text.test.ts` — helper unit coverage incl. "NEVER returns `[object Object]` for a structured backend failure" and a cyclic-object guard.
- `src/__tests__/orchestrator/inject-error-serialization.test.ts` — drives a real `PanelAgent` with a recording backend; asserts both `injectRunError` and `injectEvent(run_error)` deliver readable text (no `[object Object]`) to the model turn.

Typecheck (`npm run lint`) clean; full suite green except one pre-existing, environment-dependent `node-dev` test (ripgrep-vs-builtin engine, untouched by this PR).

[reports ≤panel 0.11.5]

### DIFF
--- a/src/__tests__/orchestrator/inject-error-serialization.test.ts
+++ b/src/__tests__/orchestrator/inject-error-serialization.test.ts
@@ -1,0 +1,96 @@
+// Regression for panel #176: a STRUCTURED backend/ComfyUI failure object pushed
+// into a tab's agent must reach the model as readable text, never the literal
+// "[object Object]". Both injection paths are covered: injectRunError (urgent,
+// front-queued) and injectEvent({kind:"run_error"}) (normal enqueue).
+
+import { describe, expect, it, beforeAll } from "vitest";
+import type {
+  AgentBackend,
+  AgentEvent,
+  BackendStartOptions,
+  ModelChoice,
+} from "../../orchestrator/agent-backend.js";
+import { CLAUDE_CAPABILITIES } from "../../orchestrator/agent-backend.js";
+
+let PanelAgent: typeof import("../../orchestrator/panel-agent.js").PanelAgent;
+
+beforeAll(async () => {
+  ({ PanelAgent } = await import("../../orchestrator/panel-agent.js"));
+});
+
+/** Records every user-turn text the panel feeds it; ends each turn immediately. */
+class RecordingBackend implements AgentBackend {
+  readonly id = "claude" as const;
+  readonly capabilities = CLAUDE_CAPABILITIES;
+  turns: string[] = [];
+  private resolveNext: (() => void) | null = null;
+
+  /** Resolve once at least `n` turns have been read. */
+  waitTurns(n: number, timeoutMs = 2000): Promise<void> {
+    if (this.turns.length >= n) return Promise.resolve();
+    return new Promise<void>((resolve, reject) => {
+      const timer = setTimeout(() => reject(new Error(`only ${this.turns.length} turns read`)), timeoutMs);
+      this.resolveNext = () => {
+        if (this.turns.length >= n) {
+          clearTimeout(timer);
+          resolve();
+        }
+      };
+    });
+  }
+
+  async *run(opts: BackendStartOptions): AsyncGenerator<AgentEvent> {
+    yield { type: "session", sessionId: "sess-rec" };
+    for await (const turn of opts.channel) {
+      this.turns.push(turn.text);
+      this.resolveNext?.();
+      yield { type: "assistant", text: "ok" };
+      yield { type: "result", ok: true, subtype: "success" };
+    }
+  }
+  async interrupt(): Promise<void> {}
+  async listModels(): Promise<ModelChoice[]> {
+    return [];
+  }
+}
+
+function makeDeps() {
+  return {
+    mcpServers: {},
+    systemAppend: "",
+    model: "claude-test",
+    onSay: () => {},
+    onTurn: () => {},
+  };
+}
+
+describe("[object Object] never appears — structured failures serialize to readable text (#176)", () => {
+  it("injectRunError serializes a structured error object", async () => {
+    const backend = new RecordingBackend();
+    const agent = new PanelAgent("tab-inj1", makeDeps() as never, backend);
+    void agent.start();
+
+    // The #176 shape: a structured failure carrying a readable `message` field.
+    await agent.injectRunError({ message: "Individual quota reached.", code: 429 });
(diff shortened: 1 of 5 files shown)


artokun:
Superseded by #405 (current WS-9 branch fix/chat-object-serialization-ws9e). Closing this stale branch from an aborted run.

PR #410, opened 2026-07-30
feat(report-bug): fix-then-file default + async Worker submit/poll client

artokun:
## What

Two-part reporting enhancement (revised per review).

### Part A — "fix before submit" (SKILL.md)
Local fix is now the firm **DEFAULT** for OUR repos: patch the code where it runs FIRST (exactly **one** attempt — don't spiral), then file with the diff so reports arrive as near-PRs. Updated frontmatter trigger + Step 3. **Step 6 now consistently makes third-party / ComfyUI-core OFFER + ASK** (patch and file only with the user's go-ahead) — resolving the contradiction with the frontmatter; only OUR-repo defects are autonomous. Secret-scrub + engineer-PR/Worker paths intact.

### Part C — wire the client (synchronous Worker contract)
- `report_issue` tool files via the Worker for our repos and returns the real issue link; **prefilled GitHub-URL fallback** on third-party, `no_file`, and **ANY** non-OK / network failure (not just 401).
- `submitAndPoll`: the abort timeout now wraps **both `fetch()` and `.json()`** so a hung body parse can't wedge the call; poll is a fallback only (the Worker returns `url` inline).
- SKILL curl flow: `--max-time` on every curl, **prints** the resulting issue URL, generic fallback to `report_issue` on any failure, never polls with an empty `JOB_ID`.

## Depends on
Worker synchronous contract: artokun/comfyui-mcp-issue-worker#2.

## Tests / gate
- `npm run build` / lint (tsc): green.
- `src/tools/report-issue.test.ts` (17 tests): OLD-SYNC fixture (`{ok,url,number}` no job_id → no poll), fast-inline link, poll-until-done, non-OK throws, poll-error; plus the **registered tool** asserting third-party / `no_file` / non-OK 500 / 401 / network-fail / poll-error all return the prefilled URL, success returns the filed link, invalid repo → isError.
- Full suite: 2017 pass, 1 skipped. The one failing test (`node-dev.test.ts` builtin-vs-ripgrep) is pre-existing and environment-dependent (this machine has ripgrep) — unrelated.

### DIFF
--- a/plugin/skills/report-bug/SKILL.md
+++ b/plugin/skills/report-bug/SKILL.md
@@ -121,9 +121,9 @@ Then file it (no need to ask):
 - **Default path (everyone):** POST the report to our intake Worker — no GitHub
   account needed:
 
-  The Worker is an **async intake**: the POST accepts + queues the report and
-  returns a `job_id`; the issue is filed in the background. Submit, then poll
-  `/status/<job_id>` a few times to get the issue link.
+  The Worker files the issue **synchronously** and returns the link inline
+  (`{ ok, url, number, deduped?, job_id }`); the `job_id` + `/status/<job_id>`
+  poll is only a fallback for the rare case the submit didn't carry a `url`.
 
   ```bash
   # URL is baked in; override with $COMFYUI_MCP_ISSUE_WORKER_URL if set. The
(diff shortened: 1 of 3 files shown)


artokun:
### Round-2 revisions (force-pushed)

- **HIGH — poll failures now fall back.** `submitAndPoll` resolves *only* with a done result carrying a real `url`; ANY poll failure (HTTP error, timeout, unreachable, body-parse stall), retry exhaustion, or done-without-url THROWS → tool + SKILL curl loop fall back to the `report_issue` prefilled URL and never claim success without a link.
- **HIGH — autonomy scoped.** Frontmatter/intro/Step 6 now agree: OUR repos (artokun/comfyui-mcp, -panel, -issue-worker) = autonomous fix-then-file; third-party / ComfyUI-core = offer-and-ask.
- **MED — case-insensitive owner.** `isOurRepo` now lowercases owner/name, so `Artokun/comfyui-mcp` (capital A, as in git config) files via the Worker instead of being misrouted third-party. Mixed-case tests at unit + tool level.
- **MED — timeout regression locked.** The abort timer wraps `fetch()` **and** `.json()`; added stalling-`.json()` fixtures (submit + poll variants, at `submitAndPoll` and registered-tool level via `COMFYUI_MCP_ISSUE_TIMEOUT_MS`) asserting the prefilled fallback.

`report-issue.test.ts`: **23 pass**. Full suite: 2023 pass / 1 skipped (the lone `node-dev` builtin-vs-ripgrep failure is pre-existing and environment-only — ripgrep is installed on this box).

PR #590, opened 2026-08-01
fix(download): warn when a remote ComfyUI-Manager dispatch is authentication-gated (#473)

### LINKED ISSUE #473: Remote CivitAI downloads save auth/error payload as .safetensors and report success
On a remote ComfyUI target, `download_civitai_model` resolves model/version metadata with the locally configured `CIVITAI_API_TOKEN`, but then hands the raw CivitAI download URL to ComfyUI-Manager. Manager cannot receive the bearer header. It saves the authentication/error response under the requested `.safetensors` filename, and the tool reports `CivitAI model downloaded successfully`. The file appears in `/models/loras`, so the current landing watcher also treats it as success. Runtime loading then fails because the payload is not safetensors.

`panel_request_secret` documentation says storing `CIVITAI_API_TOKEN` for built-in comfyui tools fixes download 401, but that is false for remote Manager dispatches because the credential stays in the local MCP process.

### Steps to reproduce
1. Connect comfyui-mcp 0.48.13 to a remote ComfyUI host with ComfyUI-Manager in `network_mode=personal_cloud`, `security_level=normal`.
2. Store a valid CivitAI token using `panel_request_secret` for built-in server env `CIVITAI_API_TOKEN`.
3. Call `download_civitai_model` with model_id 627113, model_version_id 3174878, target_subfolder `loras`.
4. Observe tool reports success and the requested filename appears in `/models/loras`.
5. Load it through core `LoraLoader`.

### Exact error (scrubbed)
`safetensors._safetensors_rust.SafetensorError: Error while deserializing header: header too large`

Failed node: core `LoraLoader`, while calling `comfy.utils.load_torch_file(..., safe_load=True)`.

### Fix
Applied locally: no (one local experiment was attempted and then reverted after verification failed).
Upstream-only: yes.

A focused local experiment resolved the authenticated CivitAI endpoint with a bearer header and dispatched the first cross-origin redirect URL to Manager. The file still loaded as an invalid safetensors payload, so the change was reverted per the one-attempt rule.

Precise change needed:
- Remote CivitAI downloads must not rely on Manager fetching an unauthenticated raw CivitAI URL.
- Add a supported authenticated transport to the remote host: either a panel-host endpoint that receives a model descriptor and performs the CivitAI request with a host-stored token, or an MCP-to-panel streamed upload into a validated `models/<category>` destination.
- Validate the landed payload before reporting success: reject/remove HTML/JSON/auth responses, verify expected size/content type, and for `.safetensors` verify the header can be parsed (or at minimum that the first 8-byte header length is plausible and the JSON header is present).
- Do not accept filename presence alone as completion.
- Update `panel_request_secret`/`download_civitai_model` help so it does not claim a local token fixes remote Manager downloads until auth forwarding exists.

### Environment
- Remote OS: Linux aarch64
- ComfyUI: 0.29.0
- Python: 3.12
- torch: 2.13.0 / CUDA 13.0
- GPU: NVIDIA GB10, 121.6 GB VRAM
- ComfyUI-Manager: 3.x legacy upgraded/configured for personal_cloud during reproduction
- comfyui-mcp: 0.48.13
- comfyui-mcp-panel: version unavailable from remote status because `panel_reload` disconnected the only live tab

No secrets, authorization headers, or signed URLs are included.

artokun:
Closes #473

## The residual (verified on current `main`, 0.48.23)

`#511`/0.48.21 added the magic-byte/Content-Type gate (`assertModelPayloadOrThrow` + `classifyPayload`/`detectNonModelPayload`), but it lives entirely in `download-cache.ts` and runs **only when the MCP process materializes bytes locally** (HTTP stream, cache-hit, cloud). The **remote ComfyUI‑Manager dispatch** (`downloadModelViaManagerRemote` → `installModelViaManager` → `queueManagerTask("install-model")`) hands the raw URL to Manager, which fetches it **server-side and unauthenticated** (it can't carry the MCP's auth headers). A login-gated CivitAI version therefore makes Manager save a `200` HTML login page under the `.safetensors` name and report its queue task `done` — exactly the recurrence reported on **0.48.22** (a 10,036-byte `<!doctype html>` login page saved as `KNP_000003000.safetensors`). The `#577`/0.48.22 "dispatched, not verified" message shipped, yet the poisoned file still landed and `list_local_models` trusted it.

Crucially, for a generic remote host the MCP has **no primitive to sniff, size, or delete** the landed file: `/models/<dir>` returns filenames only (remote `listLocalModels` hardcodes `size:0`), there is no model-byte/header read endpoint, and `remove_model` hard-refuses in remote mode. So the local "validate-before-commit + atomic rename" cannot be reproduced on the remote path.

## The fix — a non-blocking, credential-FLIP warning

Before the remote dispatch, `probeRemoteModelPayload` predicts what Manager will land:

1. Probe **unauthenticated** (exactly what Manager does). Real model binary → `model` (public → dispatch).
2. Otherwise fetch the **same URL with the credential the local path would apply** (`localAuthHeadersFor` — explicit auth, else the auto HF/CivitAI host token). If **that** returns real model binary, the credential — not the network vantage — is what unlocks it: the URL is **authentication-gated** (IP-independent proof, the exact reported bug: a local token Manager never receives) → `non-model`.
3. Any other outcome (no credential, no flip, network error/timeout, or a location interstitial that ignores the bearer) → `inconclusive`.

On a confirmed flip the dispatch **is not refused** — it surfaces a loud, specific `WARNING` (auth-gated; Manager will save an auth/login page; remediation = set the token on the **host**, or download locally) and still dispatches.

### Why warn, not block
No MCP-side probe can prove what the differently-vantaged host receives (an IP-allowlisted host could succeed unauthenticated where the MCP fails), so a hard refusal from the MCP's vantage could block a legitimate download. The probe is designed so it **can never block or fail a legitimate dispatch**: every uncertainty path proceeds unchanged, and the probe is bounded by a manually-composed abort (caller signal + 12 s unref'd timeout) so it can't hang.

**Authoritative prevention still requires the host-side authenticated transport the issue scopes** (an MCP-to-panel streamed upload into a validated destination). This PR is the safe MCP-only hardening in the meantime — it converts a silent false success into a high-confidence, actionable warning for the exact reported scenario.

## Tests
`npm run build` + `npm test` green (only a pre-existing, unrelated `node-dev` ripgrep-vs-builtin environment failure). 22 new tests cover: flip → warn; no-flip interstitial → no warn (the MCP-vs-host vantage case); no credential → no warn; public → single probe; inconclusive → dispatch; and the timeout firing with a caller signal present.

## Codex self-gate
5 adversarial rounds (`gpt-5.6-terra`, high). Round 1 caught a transient-status false-positive (fixed → status-gated). Rounds 2–3 drove the design from a status heuristic to credential-flip and finally to warn-not-block (eliminating the MCP-vs-host vantage false-positive). Round 4 flagged a compat-path timeout P1 (fixed → manual signal composition + regression test). **Round 5: `VERDICT: PASS`, no remaining P0/P1.**

> Draft — please run your own independent adversarial review before merge.

### DIFF
--- a/src/__tests__/services/model-resolver.test.ts
+++ b/src/__tests__/services/model-resolver.test.ts
@@ -620,6 +620,72 @@ describe("downloadModel — remote mode (Manager install-model dispatch)", () =>
     expect(out).not.toMatch(/AUTHENTICATION-GATED/i);
   });
 
+  it("does NOT leak the HF token to an attacker host whose URL merely CONTAINS 'huggingface.co' (#590 P0)", async () => {
+    config.huggingfaceToken = "hf-secret";
+    // unauth → auth page (would drive a flip IF a credential were derived); but the PARSED
+    // host is attacker.example, so no HF token is derived and no request carries it.
+    fetchMock.mockImplementation((_url: string, opts?: { headers?: Record<string, string> }) =>
+      Promise.resolve(opts?.headers?.Authorization ? binaryProbeResponse() : htmlProbeResponse()),
+    );
+    await downloadModel(
+      "https://attacker.example/m.safetensors?ref=huggingface.co",
+      "checkpoints",
+      "m.safetensors",
+    );
+    expect(installModelViaManagerMock).toHaveBeenCalledTimes(1); // never blocked
+    const leaked = fetchMock.mock.calls.some((c) =>
+      String((c[1] as { headers?: Record<string, string> } | undefined)?.headers?.Authorization ?? "").includes("hf-secret"),
+    );
+    expect(leaked).toBe(false);
+  });
+
+  it("attaches the HF token by PARSED host for a genuine huggingface.co gated URL and warns on the flip", async () => {
+    config.huggingfaceToken = "hf-secret";
+    fetchMock.mockImplementation((_url: string, opts?: { headers?: Record<string, string> }) =>
+      Promise.resolve(opts?.headers?.Authorization ? binaryProbeResponse() : htmlProbeResponse()),
+    );
+    const out = await downloadModel(
+      "https://huggingface.co/org/repo/resolve/main/m.safetensors",
+      "checkpoints",
+      "m.safetensors",
+    );
+    expect(installModelViaManagerMock).toHaveBeenCalledTimes(1);
+    expect(out).toMatch(/AUTHENTICATION-GATED/i);
+    const sentHf = fetchMock.mock.calls.some(
+      (c) => (c[1] as { headers?: Record<string, string> } | undefined)?.headers?.Authorization === "Bearer hf-secret",
+    );
+    expect(sentHf).toBe(true);
+  });
+
+  it("keeps the HF token flowing to an HF_ENDPOINT mirror (pre-rewrite wasHfUrl threaded) — #590 P1", async () => {
+    const prev = process.env.HF_ENDPOINT;
+    process.env.HF_ENDPOINT = "https://hf-mirror.example";
+    try {
+      config.huggingfaceToken = "hf-secret";
+      fetchMock.mockImplementation((_url: string, opts?: { headers?: Record<string, string> }) =>
+        Promise.resolve(opts?.headers?.Authorization ? binaryProbeResponse() : htmlProbeResponse()),
+      );
+      const out = await downloadModel(
+        "https://huggingface.co/org/repo/resolve/main/m.safetensors",
+        "checkpoints",
+        "m.safetensors",
+      );
+      expect(installModelViaManagerMock).toHaveBeenCalledTimes(1);
+      // The URL was rewritten to the mirror, yet the HF token still applies (threaded
+      // wasHfUrl) → the flip is detected and the auth-gated warning fires.
+      expect(out).toMatch(/AUTHENTICATION-GATED/i);
+      const sentToMirror = fetchMock.mock.calls.some(
+        (c) =>
+          String(c[0]).startsWith("https://hf-mirror.example") &&
+          (c[1] as { headers?: Record<string, string> } | undefined)?.headers?.Authorization === "Bearer hf-secret",
+      );
+      expect(sentToMirror).toBe(true);
+    } finally {
+      if (prev === undefined) delete process.env.HF_ENDPOINT;
+      else process.env.HF_ENDPOINT = prev;
+    }
+  });
+
   it("does NOT probe a NON-model-binary destination (.zip) — dispatch proceeds, no warning", async () => {
     // A .zip is not a model-binary ext, so the probe short-circuits to inconclusive WITHOUT
     // fetching and the dispatch proceeds unchanged.
--- a/src/__tests__/services/model-resolver.test.ts
+++ b/src/__tests__/services/model-resolver.test.ts
@@ -61,6 +61,40 @@ function headersOf(callIndex = 0): Record<string, string> {
   return opts.headers;
 }
 
+// #473 remote residual: the remote (Manager) dispatch path now PROBES the exact URL
+// Manager will fetch (unauthenticated) before dispatching, and refuses on a definitive
+// auth/error payload. These build the probe responses the mocked fetch returns.
+/** Non-text first bytes + octet-stream → classifier "binary" → probe verdict "model". */
+function binaryProbeResponse(): Response {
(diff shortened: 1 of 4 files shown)


artokun:
## Security-fix round (commit 5d96a34) — credential leaks closed

Addressed the two P0 credential leaks + P1/P2 from the independent adversarial review:

- **P0 #1 — HF token via substring host match.** `localAuthHeadersFor` gated `HF_TOKEN` on `url.includes("huggingface.co")`, leaking it to e.g. `https://attacker.example/m.safetensors?ref=huggingface.co`. Now gated on the **parsed hostname** (`=== "huggingface.co" || endsWith(".huggingface.co")`); CivitAI branch was already hostname-parsed; unparseable URL → no token.
- **P0 #2 — arbitrary auth header across a redirect.** Replaced `redirect:"follow"` with **manual** redirect following in `sniffPayloadShape`; the first hop off the **original origin** drops ALL auth headers (Authorization + custom like `X-API-Key`), keeping only `Range`. A caller credential is never sent to any host but the one addressed.
- **P1 — HF_ENDPOINT false-negative.** Threaded pre-rewrite `wasHfUrl` from `downloadModel` so a mirror still gets the token.
- **P2 — cry-wolf.** Warn only when the unauth response is a CLEAR auth page (html/json body) on a **non-transient** status; 429/5xx/bare-status → inconclusive.
- Also: the two sequential probes now share **one** abort+timeout budget (total bounded, not 2×); `readResponseHead` caps buffered bytes to the sniff window.

Kept the verified-good invariants (fail-open on probe failure, cancellation-before-dispatch, advisory-only never-refuse, no credential in the warning text).

New tests: attacker-substring host (no leak) + legit parsed-host; cross-origin custom-header stripping; HF_ENDPOINT mirror; 429/503/bare-status P2 cases. `npm run build` + `npm test` green (only the pre-existing unrelated `node-dev` ripgrep env failure). Codex self-gate (gpt-5.6-terra/high): **VERDICT PASS, no remaining P0/P1**.

Ready for your independent re-review — still a **draft**, not merged.

artokun:
## Third P0 closed (commit 826137f) — LOCAL-path HF-token substring leak

You were right: I'd fixed the probe's copy but the **local streaming path** (`downloadModel`) still had the identical `url.includes("huggingface.co")` leak — a local download of `https://evil.example/m.safetensors?ref=huggingface.co` (HF_TOKEN set, no explicit auth) would send `Authorization: Bearer <HF_TOKEN>` to evil.example.

Fix: extracted a single shared authority **`isHuggingFaceHost(url)`** (parses hostname; `=== "huggingface.co" || endsWith(".huggingface.co")`; unparseable → false) and used it in **both** the remote probe's `localAuthHeadersFor` **and** the local streaming path, replacing the substring check. CivitAI decisions on both paths already go through the hostname-parsed `isCivitaiUrl`. Grep confirms **no substring host check remains** in the file (only a doc-comment mention of the anti-pattern). The `wasHfUrl` HF_ENDPOINT-mirror path is unchanged (attacker can't set it).

New tests: local-path `?ref=huggingface.co` and look-alike `huggingface.co.evil.com` → **no token**; genuine `huggingface.co` still gets it.

`npm run build` + `npm test` green (2708 pass; only the pre-existing unrelated `node-dev` ripgrep env failure). Codex self-gate (gpt-5.6-terra/high): **VERDICT PASS, no remaining P0/P1** (verified parsed-host correctness, both call sites, wasHfUrl threading, CivitAI parsing, no token logging, test coverage).

Still a **draft** — ready for your final independent pass.

artokun:
Independent adversarial codex review (gpt-5.6-terra/high) — CLEARED after 3 rounds. Caught **3 distinct credential-leak P0s** the author's own multi-round self-gate had passed: (R1) HF_TOKEN attached via `url.includes("huggingface.co")` substring → token sent to `attacker.example?ref=huggingface.co`; (R1) arbitrary auth header (X-API-Key) re-sent across a cross-origin redirect; (R2) the identical substring leak surviving on the LOCAL streaming download path. All fixed via a shared parsed-host `isHuggingFaceHost()` on both paths + manual-redirect auth-stripping on first cross-origin hop. R3 verified clean: no substring host-check remains, token injection only at the 2 intended sites, wasHfUrl mirror threading unspoofable, no token in logs/warnings. Advisory-only (never refuses a dispatch) is the accepted MCP-side hardening; the authoritative host-side transport fix is scoped in the issue. Merge-approved.

Issue #360, opened 2026-07-29
comfy_cli_jobs wait rejects documented promptId parameter


artokun:
### What happened / root cause
The comfy_cli_jobs schema documents both promptId and promptIds, but action=wait with a single promptId is rejected with `promptIds or all=true is required for jobs wait`. The implementation validates only promptIds/all for wait and ignores the documented singular field.

### Steps to reproduce
Call comfy_cli_jobs with action=wait, where=local, promptId set to a valid running ComfyUI prompt, and timeoutSeconds=50.

### Exact error (scrubbed)
`promptIds or all=true is required for jobs wait`

### Fix
Applied locally: no. Upstream-only: yes.
Normalize promptId into a one-element promptIds array before wait validation, or remove promptId from the schema if unsupported. Add a regression test for wait with promptId.

### Environment
darwin / ComfyUI 0.29.0 / MPS 128 GB / comfyui-mcp 0.48.5 / panel 0.11.4.

PR #610, opened 2026-08-01
fix(consent): survive idle-user timeout on the adult-consent card (panel#390)

### LINKED ISSUE #390: Make adult consent prompt survive inactive user timeout
`panel_request_adult_consent` displayed a blocking consent card but the MCP tool call did not resolve before the 300-second tools/call timeout. A consent prompt that waits indefinitely for user interaction can outlive the transport timeout and surface as a tool error even though no application failure occurred.

### Steps to reproduce
1. Ensure adult content mode is off.
2. Call `panel_request_adult_consent`.
3. Do not click the card within 300 seconds.
4. The tool fails with a tools/call timeout instead of resolving cleanly.

### Exact error
tool call error: tool call failed for panel/panel_request_adult_consent
Caused by: timed out awaiting tools/call after 300s

### Fix
applied locally: no
upstream-only: yes
Make the consent card non-blocking (return a card/request id immediately) or resolve before the transport deadline with a structured `{ nsfw_allowed:false, timed_out:true }` result. The persistent consent state can then be queried with `panel_get_content_mode` on the next turn.

### Environment
macOS (darwin), MPS 48 GB, RAM 48 GB
ComfyUI 0.29.1
torch 2.12.1
Python 3.13
comfyui-mcp 0.48.21
panel 0.11.19

artokun:
Closes artokun/comfyui-mcp-panel#390

## Root cause
`panel_request_adult_consent` (orchestrator, `src/orchestrator/panel-tools.ts`) rendered the 18+ consent card with a hardcoded `bridge.send({ … }, { timeoutMs: 300000 })`. That raced the enclosing MCP `tools/call` ~300s deadline with **zero margin** and used **no `ask_id`** (so no late-reply buffering). An idle user who never clicked the card blew the whole tool call as a transport error — `timed out awaiting tools/call after 300s` — instead of resolving cleanly, dead-ending the flow.

The consent card was the last direct-`bridge.send` ask that still bypassed the clamp+late-reply machinery already adopted by `confirm` (#360/#536) and `panel_ask`/`askUserWithGrace` (#300/#486).

## Fix
Mirror that established pattern for the consent card:
- Clamp the card deadline under the MCP budget via `getAskTiming()` (`ASK_TOTAL_BUDGET_CAP_MS = 285s`, safely below 300s).
- Send a stable `ask_id` so a slow-but-valid pick can be keyed to the late-reply buffer.
- On a reply-timeout, `pollLateAskReply()` for a bounded grace and **honor** a late answer (yes → grant, no → SFW).
- If still unanswered, resolve cleanly with `{ nsfw_allowed:<persisted>, timed_out:true }` **without** calling `setNsfwConsent`. The persistent decision is queryable via `panel_get_content_mode` on the next turn, or the user can be re-asked.

Non-timeout errors (no panel / transport failure) still propagate to `fail()`, exactly as before.

## Gate integrity (not weakened)
Consent is turned **ON only** by `isAffirmative(reply)` on a *real* reply — immediate or late-buffered. A timeout / no-answer path returns **before** any `setNsfwConsent` call, so:
- A timeout **never auto-grants** the gate.
- A timeout **never drops** a previously granted consent (nothing is written).

Persistent consent already lives on disk (`~/.comfyui-mcp/panel-settings.json` via `setNsfwConsent`) and genuinely survives idle/reload; this PR only stops the *in-flight prompt* from being lost to the transport timeout.

## Tests
New describe block `adult-consent card survives an idle-user timeout (#390)` (settings store isolated to a temp file):
1. Unanswered card → `{ timed_out:true, nsfw_allowed:false }`, **not** a transport error, and `getNsfwConsent().allowed` stays `false` (gate not auto-granted); forwarded `timeoutMs` clamped `< 300000` with a string `ask_id`.
2. Late-but-valid **YES** buffered after the reply timeout → grants consent.
3. Late-but-valid **NO** → stays SFW.
4. Fast YES grants / fast NO keeps SFW (gate not weakened).
5. Idle timeout does **not** drop a previously granted consent.

`npm run build` clean; `npm test` green for the touched suite (142/142 in `panel-tools.test.ts`; full suite passes except one pre-existing, unrelated `node-dev` ripgrep-vs-builtin environment assertion identical to origin/main).

## Codex self-gate
`gpt-5.6-terra` / high, read-only, embedded diff: **PASS** — "A no-answer returns before `setNsfwConsent`, preserving the persisted state; only immediate or ask-id-keyed late replies reach `isAffirmative`. Non-timeout errors rethrow to `fail()`."

Do not merge — consent-gate change, owner re-reviews.

### DIFF
--- a/src/__tests__/orchestrator/panel-tools.test.ts
+++ b/src/__tests__/orchestrator/panel-tools.test.ts
@@ -25,6 +25,7 @@ import {
   type ToolResult,
 } from "../../orchestrator/panel-tools.js";
 import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+import { markReplyTimeout } from "../../services/ui-bridge.js";
 
 type Forwarded = Record<string, unknown>;
 
@@ -1639,6 +1640,10 @@ describe("panel-tools: NSFW consent enforced server-side on CivitAI browsing lev
 
 describe("panel-tools: adult-consent card survives an idle-user timeout (#390)", () => {
   const origSettings = process.env.COMFYUI_MCP_PANEL_SETTINGS;
+  // The card's exact affirmative/decline option labels — the SAME source constants the
+  // classifier matches byte-for-byte, so tests can't drift from the real card.
+  const CONSENT_YES_LABEL = __panelToolsTestHooks.CONSENT_YES_LABEL;
+  const CONSENT_NO_LABEL = __panelToolsTestHooks.CONSENT_NO_LABEL;
 
   beforeAll(() => {
     // Isolate the persistent consent store to a throwaway file for this suite.
@@ -1838,7 +1843,7 @@ describe("panel-tools: adult-consent card survives an idle-user timeout (#390)",
   it("an EXACT decline explicitly reverts a prior grant to SFW", async () => {
     setNsfwConsent(true);
     const ctx = {
-      bridge: replyBridge("No — keep it SFW"),
+      bridge: replyBridge(CONSENT_NO_LABEL),
       tabId: "abcd",
       ensureReachable: () => {},
     } as unknown as PanelToolCtx;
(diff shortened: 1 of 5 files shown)


artokun:
## Security hardening round (adversarial re-review)

An independent adversarial review found that the first commit's reply-handling weakened the consent gate. Fixed in `ce01863`:

- **P0 — free-text replies could grant without genuine consent.** The consent card has a free-text "Other" box, so a reply (immediate *or* late-buffered) can be arbitrary text; the first commit fed it through the broad `isAffirmative()` prefix regex, so `"adult content is illegal here"`, `"On second thought, no"`, `"18 is the age but I decline"`, etc. matched → adult mode ON. **Fix:** new strict `classifyConsentReply()` — consent turns ON only on the **exact affirmative option label** or an anchored `^…$` whole-string yes token. `isAffirmative()` now stays confined to the destructive-op confirm cards (a false negative there only skips an op).
- **P1 — a timeout/unclear reply could revoke a prior grant.** Any *unclear* reply (incl. a no-answer timeout) now writes **nothing** — it neither grants nor revokes. Only an **exact decline** reverts to SFW.
- **P1 — `getNsfwConsent()` wasn't fail-closed.** `read()` casts arbitrary on-disk JSON, so a tampered/legacy truthy string (`"false"`, `"true"`, `1`, `"yes"`) was truthy → adult browsing permitted. **Fix:** `allowed` is consent **only** when strictly `=== true`.
- **P1 (2nd review pass) — `isReplyTimeoutError()` too broad.** It matched any error containing `did not reply … within N ms`, so a real transport error phrased that way would be treated as a recoverable card timeout and returned as a quiet unchanged-state success instead of `fail()`. **Fix:** anchored on the bridge's canonical `Panel tab <id> did not reply to "<cmd>" within <N> ms` shape (matching the precedent the `panel_open_workflow` ack path already uses).

### Gate integrity (net)
Adult mode turns **ON only** via a genuine explicit affirmative (exact option or strict `yes`); a timeout / unclear / stray reply **never grants and never revokes**; a tampered on-disk value **never** counts as consent; a real transport error **fails closed** (surfaces as an error, gate untouched).

New regression tests: late & immediate stray-input non-grant (incl. the exact P0 strings), prior-grant preservation on unclear/timeout, exact yes/decline, tampered-disk fail-closed (`getNsfwConsent`), and non-timeout-error propagation. `npm run build` + `npm test` green (same single pre-existing unrelated `node-dev` ripgrep env assertion).

### Codex re-review (gpt-5.6-terra / high, read-only, embedded diff): **PASS**
> Consent grants only via `classifyConsentReply()`'s exact label or anchored whole-string affirmative tokens; `isAffirmative()` remains confined to destructive-operation confirmations. All listed adversarial strings classify as `unclear`, which returns persisted state without writing. Timeout and unclear replies preserve prior consent; strict declines revoke. `getNsfwConsent()` is fail-closed with `raw?.allowed === true`. Late-reply handling retains one `ask_id`, caps deadline + grace at 285s, and rethrows non-canonical timeout errors to `fail()`.

Still draft — do not merge; consent-gate change, owner re-reviews.

artokun:
## P1c fully closed — `isReplyTimeoutError` is now an exact canonical-message matcher (`d935f3b`)

Your round-2 catch was correct: the earlier matcher was **unanchored**, so a *prefixed* transport error like `relay failed: Panel tab abcd did not reply to "ask_user" within 500 ms; reconnecting` still matched and was swallowed as a recoverable timeout. Two further tightening passes (my own adversarial codex re-review) closed the remaining edges:

- Anchored `^…$` to the bridge's **canonical whole message**, then replaced the trailing `$` with a **hard end-of-input `(?![\s\S])`** — because JS `$` (no `/m`) also matches just before a final `\n`, so `canonical + "\n"` was still slipping through. No trailing character of any kind (`\n`, `\r`, U+2028/2029, NUL) is now tolerated.
- Literal single space `within \d+ ms` (rejects `500ms` / `500\tms`); removed `.trim()`.

Runtime-verified truth table for the final regex:

| input | matches? |
|---|---|
| canonical bridge message | ✅ (stays recoverable) |
| `canonical + "\n"` | ❌ → fail() |
| leading spaces | ❌ → fail() |
| suffixed `… frozen; socket closed` | ❌ → fail() |
| prefixed `relay failed: …; reconnecting` | ❌ → fail() |
| `within 500ms` / `within 500\tms` | ❌ → fail() |

Both actual bridge emit sites (ui-bridge.ts:1514, :1569) produce the exact accepted template, so genuine card timeouts remain recoverable (late-reply polling intact for confirm/panel_ask/consent).

New tests exercise every rejection variant (incl. the isolated trailing-newline case) → fail(), and the genuine canonical timeout → recoverable. `npm run build` + `npm test` green (same single pre-existing unrelated `node-dev` ripgrep env assertion).

### Codex re-review (gpt-5.6-terra / high, read-only, embedded diff): **PASS**
> `isReplyTimeoutError()` uses a true start anchor and hard end-of-input `(?![\s\S])`; no trailing character — including `\n`, `\r`, U+2028/U+2029, or NUL — can remain. Whitespace, wrappers, and noncanonical `ms` spacing therefore reach `fail()`. Both actual bridge timeout sites emit the exact accepted template. The consent path remains fail-closed: only strict classification can grant, unclear/timeout leaves state untouched, and persisted consent requires `allowed === true`.

Ready for your merge review. Still draft — I won't merge (consent-gate change).

artokun:
## Defense-in-depth hardening (both items) — `9629e27`

Both round-3 items done. As you noted, neither is a triggerable grant-without-consent (the reply is the user's own round-scoped `ask_user` answer, no third-party injection vector) — this is belt-and-suspenders on THE consent gate.

**1. Byte-exact option identity.** `classifyConsentReply` previously did `reply.trim()` *before* the exact-label comparison, so a near-label wrapped in whitespace / BOM (U+FEFF) / line separators (U+2028/2029) / newlines would trim down to the exact affirmative and grant. Now the exact-LABEL comparison is **byte-for-byte (no trim)**; `.trim()` is applied **only** to the deliberately-loose free-text token path (`^(yes|y|i agree|…)$`), where surrounding whitespace is benign. A wrapped near-label can't match that anchored token regex (it's a long string), so it stays `unclear` → no grant.

**2. Typed reply-timeout marker (robust to any tab_id).** Instead of parsing the tab_id out of the message text, the bridge now tags **both** reply-timeout emit sites (`dispatch()`: the `remaining<=0` path and the `setTimeout` path) with `markReplyTimeout`. `isReplyTimeoutError` keys on `isReplyTimeoutTagged(err)` **first**, falling back to the exact canonical whole-message regex (tab-id segment loosened to `.+?`, still hard-anchored `^…(?![\s\S])` with a literal single space) for untagged/test errors. So a genuine late reply for a session whose tab_id contains a space is no longer lost to `fail()`. The disconnect emissions (`handleMidCommandDisconnect`: "panel genuinely gone" / "OUTCOME UNKNOWN") are **not** tagged, so a disconnect is never treated as a recoverable card timeout.

The marker only decides recoverable-vs-hard-error — consent is still written **only** downstream of `classifyConsentReply` on the (possibly late) reply, so item 2 opens no grant path; `getNsfwConsent` stays strict `=== true`.

New tests: BOM/whitespace-wrapped near-label → no grant (+ exact-label → grant); whitespace `yes`/`i agree` token → grant; spaced-tab_id typed-timeout → recoverable, late YES honored. `npm run build` + `npm test` green (2810 pass; same single pre-existing unrelated `node-dev` ripgrep env assertion).

### Codex re-review (gpt-5.6-terra / high, read-only, embedded diff): **PASS**
> Byte-exact labels: comparison precedes `.trim()`; only anchored yes-token parsing sees trimmed input. Wrapped labels cannot match that token regex. The marker only selects late-reply recovery; the state write remains downstream of `classifyConsentReply`, and no reply/unclear reply returns existing state without writing. Exactly two bridge reply-timeout emissions are tagged; disconnect paths are not. Fallback regex remains fully anchored with hard EOF, preserves canonical spacing, and accepts spaced tab IDs. Persisted consent remains strict `allowed === true`.

Files touched this round: `src/orchestrator/panel-tools.ts`, `src/services/ui-bridge.ts`, + tests. Ready for your merge re-verify — I won't merge (consent-gate change).

PR #614, opened 2026-08-01
fix(session): key unsaved-workflow resume on a durable per-instance uuid (#570)

### LINKED ISSUE #570: Session resume keyed by ephemeral tmp:<uuid> tab id — unsaved workflows always lose conversation across an orchestrator restart
## Symptom
After the orchestrator restarted (dev-build self-restart), the panel agent came back with **zero conversation memory** — it did not know what it had been doing one message earlier. The user had to paste the transcript back in manually.

`SessionStore` exists precisely to prevent this ("the orchestrator's own belt-and-suspenders copy … the only survivor when the process was killed and respawned"), but it cannot work for an unsaved workflow.

## Root cause
`SessionStore` (src/orchestrator/session-store.ts) is keyed by `tabId` (`::backend` suffixed). For an **unsaved** workflow the tab id is `tmp:<uuid>`, and that uuid is regenerated when the panel reloads. So on a restart that also reloads the panel, the tab reconnects under an id that has never been written to the store:

`panel-agent.ts:1621`
```ts
const resume = this.opts.sessionStore?.get(tabId) ?? this.pendingResume.get(tabId);
```
→ store miss, `pendingResume` miss (the panel's localStorage session belongs to the *old* tab identity) → `spawn(tabId, undefined)` → fresh session, conversation gone.

Saved workflows key on `wf:<path>`, which is stable, and resume correctly. That's why this reproduces on "Unsaved Workflow (N)".

## Direct evidence — live store, `%TEMP%/comfyui-mcp-panel-sessions-9180.json`
```json
"tmp:69c60eac-3cae-4504-adf3-e6dc87b635cb::claude":"8c44273a-07fa-4e47-8673-f1720cb9c005",
"tmp:7ca32bc6-f877-449c-81a5-e9138577b820::claude":"8c44273a-07fa-4e47-8673-f1720cb9c005",
"tmp:3d11aae7-993a-4cbf-ae69-badcf33edfcc::claude":"8c44273a-07fa-4e47-8673-f1720cb9c005",
"tmp:7beebe6e-22ec-40cd-bee2-3de61d998c02::claude":"aa5719e1-67b6-4365-93be-208466dba76b"
```
Same ComfyUI tab, same "Unsaved Workflow (6)", **four** tab ids. The first three all hold the real session `8c44273a`. The fourth — the id the tab came back under after the restart — got a brand-new session `aa5719e1`. That fourth line is the exact moment the memory was lost.

## Secondary defects visible in the same file
1. **`rebindAgent`'s `clear(oldKey)` did not run.** `tmp:69c60eac` and `tmp:7ca32bc6` are both still present even though the orchestrator logged `tab-id migration: … agent rebound` for them. The durable move only happens on a live migration event inside one process; a migration that spans a process restart leaves the old key orphaned and never populates the new one.
2. **The store never garbage-collects.** The file is 9.6 KB / ~120 entries, including `e2e-*`, `spike-mobile-render-*`, and `sendnow-*` test keys from prior runs. Unbounded growth, and stale keys make the "which entry is live" question ambiguous.

## Ruled out
Not the model switches. `setOptions` (panel-agent.ts:1629+) applies a model change **live** via the SDK's `setModel` and explicitly does not recreate the session — only an *effort* change restarts it, and that path passes the resume id. The two `Model →` switches immediately before the failure were not the cause.

## Suggested fix
Key durable resume on something stable across a panel reload — the workflow identity (title/path) or a panel-persisted install id — rather than the ephemeral `tmp:<uuid>`. Failing that, when a tab connects with an unknown `tmp:` id and the panel supplies a `hello.resume` session that the store does know, honor it instead of spawning fresh. Also: GC entries older than N days, and make the migration key-move durable across process restarts.

## Related
Same tab-id-churn family as the wedge report (#568), and adjacent to #474 (panel_reload loses live tab / does not auto-resume). This one is a distinct, independently reproducible root cause: the durable store structurally cannot hit for unsaved workflows.

## Environment
Windows 11 · RTX 4090 24GB · ComfyUI 0.29.2 (local) · torch 2.10.0 / CUDA 13.0 / py3.13 · comfyui-mcp 0.48.21 · panel version reported as `?` by the env line.

artokun:
Closes #570

## What #587 missed

#587 shipped a stable resume index for unsaved workflows keyed on `origin + title + backend`, seeded on hello as a fallback when the panel's `hello.resume` is absent. But an unsaved workflow's title is the **default `"Unsaved Workflow"`** (ComfyUI even re-derives the `"(N)"` suffix), so that key is **not unique** — two *different* unsaved workflows collide on one key. The concurrent-collision poison guard only fires when two live tabs write distinct sessions in one process; it **cannot** catch a *sequential / cross-restart* collision. So after a reset/restart the index served an **unrelated earlier same-title workflow's session** (the reopen: resumed `33c86744` for a turn instead of the conversation's `bfedfa9e`). #587 documented this as an accepted limitation pending a per-instance id.

## The fix

That per-instance id now exists. The panel already mints a **durable, reload-surviving, globally-unique** per-instance workflow uuid (`workflowStableUuid()` — #186/#386; embedded in the graph's `extra` so ComfyUI's unsaved-workflow persistence carries it across a reload; **forked** to a fresh uuid on clone/import; it is the very id that keys the durable chat transcript, which restored correctly in the report). The paired panel PR emits it as `workflow_uuid` on every hello.

The orchestrator now keys the stable index on **`wfid::<trusted-origin>::<uuid>::<backend>`** (`deriveStableKey`). Two distinct unsaved workflows can never share a uuid, so the cross-resume is **structurally impossible**, and the id survives the `tmp:<uuid>` tab-id churn that #570 is about.

## Safety — a wrong-resume is strictly worse than a lost-resume

- **Fail closed**: no valid uuid (old/pre-capability panel, or malformed) **or** no trusted origin -> `deriveStableKey` returns `undefined` and the disk fallback is skipped. The legacy `origin+title` key is **never written or read** — migrating it would reintroduce the wrong-resume. `hello.resume` still covers the common reload; a miss starts fresh (**lost**, never wrong).
- **Trusted origin only**: `bridge.tabServerOrigin` (server-observed, browser-set, unspoofable), canonicalized — never the client-supplied `hello.comfyui_url`.
- **Reset can't resurrect**: an identity regression (a valid-uuid tab re-hellos degraded) and a tab-id migration through a degraded panel `clearStable` the prior entry, so a later `new_session` + valid hello can't bring a reset conversation back.
- **Per-provider isolation**: `set_backend` recomputes the key from the stored `(origin, uuid)` identity; `onSession` writes the stable index only when the callback's backend is still the tab's current one, so a late old-agent callback can't poison the new provider's entry.
- **Poison guard unchanged**: the same-workflow-in-two-live-tabs collision still poisons rather than cross-resumes.

## Known trade-off (deliberate)

Until a panel updates, an unsaved workflow's *durable* fallback is unavailable when `hello.resume` is also absent — a **lost** resume, resolved once the paired panel ships (this PR) / self-heals. The only "compatibility path" would be reading the legacy non-unique key, which is exactly the wrong-resume this fixes. Fail-closed is the correct resolution.

## Tests

`session-store.test.ts` extended: two same-title workflows get distinct keys / never cross-resume; same instance keeps one key across a reload; per-backend partition; same-uuid-different-origin can't bridge; fail-closed on no-uuid / no-origin / malformed; origin canonicalization; identity-regression retire; backend-switch per-provider isolation. Paired panel PR: `buildHelloPayload` emits `workflow_uuid`. Full orchestrator suite green (503). (A pre-existing unrelated `node-dev.test.ts` flake fails identically on clean `main`.)

## Review

Self-gated with iterative codex adversarial review (gpt-5.6-terra/high): six cycles each caught and fixed a distinct real wrong-resume path (cross-workflow -> unsafe legacy migration -> identity-regression -> migration-path -> cross-provider-switch -> late-callback / client-origin). The final passes find **no remaining wrong-resume**; codex's residual verdict is a cross-repo-visibility note (the `workflow_uuid` producer lives in the paired panel repo, shipped here) plus the deliberate fail-closed compatibility trade-off above.

Paired with **artokun/comfyui-mcp-panel#481**. Draft — do not merge; session-loss is high-stakes and warrants a human re-review.

### DIFF
--- a/src/__tests__/orchestrator/in-place-replace-reset.test.ts
+++ b/src/__tests__/orchestrator/in-place-replace-reset.test.ts
@@ -149,4 +149,48 @@ describe("in-place workflow replacement resets the live session (#570 P0)", () =
     expect(manager.hasLiveAgent(key)).toBe(false);
     await manager.stopAll();
   });
+
+  it("P0: failed-start held mail counts as state (torn down on an identity transition)", async () => {
+    // A backend prepare() failure DROPS the agent but PARKS the queued user message in
+    // heldMessages (#256) — no live agent, no session. hasAnyState must still see it, so an
+    // in-place replacement resets it instead of re-delivering the prior workflow's message.
+    class RejectingBackend implements AgentBackend {
+      readonly id = "claude" as const;
+      readonly capabilities = CLAUDE_CAPABILITIES;
+      async prepare(): Promise<void> {
+        throw new Error("prepare rejected (bad key)");
+      }
+      async *run(): AsyncGenerator<AgentEvent> {
+        throw new Error("run must not be reached");
+      }
+      async interrupt(): Promise<void> {}
+      async listModels(): Promise<ModelChoice[]> {
+        return [];
+      }
+    }
+    const store = new SessionStore(PORT);
+    const manager = new PanelAgentManager({
+      mcpServers: {},
+      systemAppend: "",
+      model: "claude-test",
+      onSay: () => {},
+      onTurn: () => {},
+      onStartFailure: () => {},
+      onSession: () => {},
+      sessionStore: store,
+      makeBackend: () => new RejectingBackend(),
+      identityForKey: () => IDENTITY_A,
+    } as never);
+    const key = "wf:foo.json::claude";
+    manager.send(key, "A's private message queued behind a failed start");
+    // The start fails: no live agent, no durable session — but the message is PARKED.
+    await waitFor(() => manager.hasAnyState(key) && !manager.hasLiveAgent(key));
+    expect(store.get(key)).toBeUndefined();
+    expect(manager.hasAnyState(key)).toBe(true); // held mail is state
+    // The identity boundary (gated on hasAnyState) resets — the parked message is dropped,
+    // never re-delivered into the replacement workflow.
+    manager.reset(key);
+    expect(manager.hasAnyState(key)).toBe(false);
+    await manager.stopAll();
+  });
 });
--- a/src/__tests__/orchestrator/in-place-replace-reset.test.ts
+++ b/src/__tests__/orchestrator/in-place-replace-reset.test.ts
@@ -137,11 +137,11 @@ describe("in-place workflow replacement resets the live session (#570 P0)", () =
     await manager.stopAll();
   });
 
-  it("#570: a tab-id migration COLLISION (destination already live) retires the source agent — no cross-tab leak", async () => {
-    // The same workflow open in TWO tabs; one migrates onto the other's id. rebindAgent refuses
-    // (destination already has a live agent) WITHOUT moving/stopping the source, which stays live
-    // under the old id and would leak its callbacks into the destination tab via the proven
-    // old→new alias. The orchestrator's collision handling retires the orphaned source agent.
+  it("#570: a tab-id migration COLLISION resets the superseded destination then rebinds the source (no cross-tab leak)", async () => {
+    // The same workflow open in TWO tabs; the incoming socket migrates onto the other's id. The
+    // bridge has already SUPERSEDED the destination's socket, so its still-mapped agent would
+    // render into the incoming tab. Reset the superseded destination, THEN rebind the source into
+    // the freed id so the incoming tab keeps its OWN conversation.
     const store = new SessionStore(PORT);
     const manager = new PanelAgentManager({
       mcpServers: {},
@@ -155,20 +155,23 @@ describe("in-place workflow replacement resets the live session (#570 P0)", () =
       identityForKey: () => IDENTITY_A,
     } as never);
 
-    const oldKey = "tmp:A::claude"; // source tab
+    const oldKey = "tmp:A::claude"; // incoming/source tab
     const newKey = "wf:foo.json::claude"; // destination tab (already open, same workflow)
-    manager.send(newKey, "destination tab conversation");
-    manager.send(oldKey, "source tab conversation");
+    manager.send(newKey, "destination tab (B) conversation");
+    manager.send(oldKey, "incoming tab (A) conversation");
     await waitFor(() => manager.hasLiveAgent(oldKey) && manager.hasLiveAgent(newKey));
 
-    // The migration loop: rebind collides (destination live) → false; source still live.
-    const rebound = manager.rebindAgent(oldKey, newKey);
-    expect(rebound).toBe(false); // collision — destination occupied
-    expect(manager.hasLiveAgent(oldKey)).toBe(true); // source NOT moved
-    // Orchestrator collision handling: retire the orphaned source so it stops emitting.
-    if (!rebound && manager.hasLiveAgent(oldKey)) manager.retire(oldKey);
-    expect(manager.hasLiveAgent(oldKey)).toBe(false); // source retired → no cross-tab leak
-    expect(manager.hasLiveAgent(newKey)).toBe(true); // destination conversation preserved
+    // Without handling, rebind REFUSES (destination occupied) and the source stays stranded live
+    // under the old id — the leak precondition.
+    expect(manager.rebindAgent(oldKey, newKey)).toBe(false);
+    expect(manager.hasLiveAgent(oldKey)).toBe(true);
+
+    // Collision handling: reset the superseded destination (its agent is removed → it can no
+    // longer render into the incoming socket), THEN rebind the incoming source into the freed id.
+    manager.reset(newKey);
+    expect(manager.rebindAgent(oldKey, newKey)).toBe(true); // source takes the id
+    expect(manager.hasLiveAgent(newKey)).toBe(true); // incoming tab keeps its OWN (source) conversation
+    expect(manager.hasLiveAgent(oldKey)).toBe(false); // nothing stranded under the old id
 
     await manager.stopAll();
   });
--- a/src/__tests__/orchestrator/in-place-replace-reset.test.ts
+++ b/src/__tests__/orchestrator/in-place-replace-reset.test.ts
@@ -98,6 +98,45 @@ describe("in-place workflow replacement resets the live session (#570 P0)", () =
     await manager.stopAll();
   });
 
+  it("#570: a proven tab-id migration rebinds EVERY backend, so a dormant provider's session survives", async () => {
+    // Saved workflow A: Claude conversed (dormant after a switch to Codex), Codex is current, then
+    // a save/rename changes the wf: tab id. The migration must move the dormant Claude session to
+    // the new id too — not only the current (Codex) one — or switching back to Claude starts fresh.
+    const backend = new SessioningBackend();
+    const store = new SessionStore(PORT);
+    const manager = new PanelAgentManager({
+      mcpServers: {},
+      systemAppend: "",
+      model: "claude-test",
+      onSay: () => {},
+      onTurn: () => {},
+      onSession: () => {},
+      sessionStore: store,
+      makeBackend: () => backend,
+      identityForKey: () => IDENTITY_A,
+    } as never);
+
+    // Dormant Claude session + live-then-retired: model both providers having durable sessions
+    // under the OLD tab id (Claude dormant, Codex current) — bound to the same workflow identity.
+    store.set("wf:old.json::claude", "sess-claude", IDENTITY_A);
+    store.set("wf:old.json::codex", "sess-codex", IDENTITY_A);
+
+    // The proven migration loops over KNOWN_BACKENDS calling rebindAgent(old::b → new::b).
+    for (const b of ["claude", "codex"]) {
+      manager.rebindAgent(`wf:old.json::${b}`, `wf:new.json::${b}`);
+    }
+
+    // BOTH providers' sessions moved to the new id (dormant Claude included), old keys cleared.
+    expect(store.get("wf:new.json::claude")).toBe("sess-claude");
+    expect(store.get("wf:new.json::codex")).toBe("sess-codex");
+    expect(store.get("wf:old.json::claude")).toBeUndefined();
+    expect(store.get("wf:old.json::codex")).toBeUndefined();
+    // …and durably: switching back to Claude on the new id resumes its original conversation.
+    expect(new SessionStore(PORT).get("wf:new.json::claude")).toBe("sess-claude");
+
+    await manager.stopAll();
+  });
+
   it("#570: a backend switch RETIRES the prior provider (preserves its session) so switching back resumes", async () => {
     // BOTH provider-switch protocols on the SAME workflow — the explicit set_backend event AND
     // a re-hello that selects a different backend — now funnel to manager.retire(prevKey). They
--- a/src/__tests__/orchestrator/in-place-replace-reset.test.ts
+++ b/src/__tests__/orchestrator/in-place-replace-reset.test.ts
@@ -98,6 +98,43 @@ describe("in-place workflow replacement resets the live session (#570 P0)", () =
     await manager.stopAll();
   });
 
+  it("#570: a backend switch RETIRES the prior provider (preserves its session) so switching back resumes", async () => {
+    // set_backend on the SAME workflow must NOT reset the old provider — that would clear its
+    // durable exact record, and a SAVED wf: workflow has no stable-key fallback, so switching
+    // back could never resume. retire() stops the live agent but keeps the identity-bound
+    // session, so a later switch-back continues the original conversation (codex round-trip).
+    const backend = new SessioningBackend();
+    const store = new SessionStore(PORT);
+    const manager = new PanelAgentManager({
+      mcpServers: {},
+      systemAppend: "",
+      model: "claude-test",
+      onSay: () => {},
+      onTurn: () => {},
+      onSession: () => {},
+      sessionStore: store,
+      makeBackend: () => backend,
+      identityForKey: () => IDENTITY_A,
+    } as never);
+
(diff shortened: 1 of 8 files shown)


artokun:
## Update — combined-diff adversarial review (P0a / P0b / P1 + hardening)

An independent adversarial pass on the **combined** (mcp + panel) diff found the uuid keying correct but surfaced two more P0 wrong-resume paths and a P1 durability gap that bypassed it. All are now fixed, then further hardened across several review rounds. The **wrong-resume invariant is fully closed** (verified: no cross-workflow / cross-origin / cross-provider / reset-resurrection / cold-import / path-reuse resume remains).

### P0a — same-socket switch rebound one workflow's agent onto another (mcp)
The bridge stamps `migrated_from` on **any** same-socket re-hello under a new tab id (socket continuity, not workflow continuity). A workflow **switch** A→B on one browser tab was rebinding A's live agent onto B — B inherited A's conversation, B's own resume was refused, and A's session persisted under B's key.
- Compute the trusted `{server-origin, uuid}` identity **before** the migration decision; rebind **only** when both identities exist and are **equal**, else **retire** (new `PanelAgentManager.retire()`: stops the agent, **preserves** its durable session) — fail closed (old/degraded/malformed panels never rebind an unprovable continuation).
- `handleEvent()` drops all events once `closed` — a buffered backend event can't leak the retired workflow's reply/session into the switched-to tab (`retire-no-leak` test).
- Render-queued (`heldDuringGen`) messages: **explicit cancellation notice**, not a silent drop.
- `UiBridge.revokeTabMigration()` fences the old tab's route on a switch, so a stale `panel_*` call can't resolve onto the new canvas (ui-bridge regression test).

### P0b / P1 — unsaved uuid durability + cold-import fork (panel)
The uuid was resolved lazily from the graph `extra` embedded value — not durable in default panel-scoped mode (**P1**), and carried by a **copied** graph (**P0b**: a cold import/duplicate resumed the source's session).
- The unsaved uuid is now keyed on the pair **(ComfyUI tab path, reload-stable graph id)** in the panel's own localStorage, not `extra`. The path is reused on a reload (durable — P1) but **deduped** on a fresh import (`Unsaved Workflow (2).json` → new path → fresh uuid → fork — P0b). The graph id (`activeState.id`) is a second guard so a new workflow reusing a **freed path slot** can't inherit a closed one's uuid; reuse requires **both** ids present and equal (fail closed). Pure `resolveUnsavedWorkflowUuid` (unit-tested).

### Known residual (for your review — NOT a wrong-resume)
Codex flags one remaining **pre-existing** wrong-**canvas** race: a `panel_*` command dispatched to the browser **microseconds before** a same-socket switch can still execute on the new workflow after re-hello — the orchestrator cannot un-send a frame already delivered, and the bridge command frames carry no workflow generation the panel could verify. My change **improves** on the prior behavior (the route-fence closes the post-switch window; the pre-fix rebind masked this inside the larger wrong-resume bug) but cannot fully close the pre-switch in-flight window orchestrator-side. Fully closing it needs a **generation-bound panel-command protocol** (every command carries a workflow generation the panel checks before executing + invalidate pending on re-hello) — a separate cross-cutting workstream beyond this session-resume fix. Flagging for a follow-up decision; not reintroducing a wrong-resume to buy an approval.

Tests: mcp full suite green (2773 pass; 1 pre-existing unrelated `node-dev` flake excluded). Panel full unit suite green (1140). Draft — do not merge.

artokun:
## Update 2 — producer-side create-boundary fork + codex APPROVE

Two more wrong-resume/delivery paths found on the combined diff, both fixed; the combined fix now passes independent adversarial codex review (**approve** — "No ship-blocking wrong-resume or post-switch wrong-delivery path found").

### P0 (producer / panel) — a paste/duplicate copy could inherit the source's session
The unsaved identity was keyed on (ComfyUI path, graph id), but a PASTE/DUPLICATE copy **preserves both** the source's graph id and can share the synthesized `workflows/Unsaved Workflow.json` path (native temporaries share it) — so the copy presented the source's identity and the orchestrator rebound the source's live agent onto it (and resolved its stable session after restart). Path-dedup didn't hold.

Fixed by forking at the **creation boundary** instead: `installCreateBoundaryFork` wraps `app.loadGraphData` and mints a fresh `extra.comfyui_mcp.workflow_uuid` on every creation, identified by `isWorkflowCreationLoad` — ComfyUI passes the open `ComfyWorkflow` **object** as `loadGraphData`'s 4th arg for every REUSE (reload-restore, tab-switch, undo, reroute-migration, in-place repaint) and a non-object (null/string) for every CREATION (import/open/drag/template/shared-url/**duplicate**/**paste**/new-blank; external imports also set `openSource`). So a copy gets a fresh uuid; a reload round-trips the same uuid (durable — P1). The panel's own snapshot-revert opts out (`__cmcpNoFork` + passes the active workflow object). Verified against ComfyUI 1.42.15 (installed) + 1.47.2 (source): all 8 creation paths route through `loadGraphData` with a non-object 4th arg or `openSource`; every reuse passes a workflow object. The old (path,graph-id) localStorage machinery is removed; pure `isWorkflowCreationLoad` unit-tested.

### P0a refinement (consumer) — cross-provider render-queued work
`heldDuringGen` is keyed by `tab::backend`, so the workflow-switch cancellation only cleared the current provider's queue. Queue on A → switch provider → switch workflow left held work under the old provider (`onRunEnd` would respawn it for the switched-away workflow). Now the retire branch sweeps **every** backend for `migratedFrom` and emits one aggregate cancellation notice.

### Known residual (unchanged — separately tracked, NOT a wrong-resume)
A `panel_*` command already delivered to the browser microseconds **before** a switch can still act on the new canvas after re-hello — a wrong-**canvas** boundary issue needing a generation-bound panel-command protocol (a separate cross-cutting workstream). Codex confirms this is out of scope for the session-resume fix.

Tests: mcp full suite 2826 pass (1 pre-existing unrelated `node-dev` flake excluded); panel full unit suite 1162 pass. Both rebased onto the latest main (mcp 0.48.29 / panel post-0.11.29). Draft — do not merge.

artokun:
## Update 3 — the load-bearing resume-ownership P0 + saved-workflow identity binding

The independent review's core P0 (default panel-scoped chat resuming across workflows via an untrusted `hello.resume`) and the deeper saved-workflow identity edges are now fixed. **Every wrong-RESUME path is closed** (final codex round found no resume path); the only remaining codex finding is the separately-tracked in-flight wrong-**canvas** residual (below).

### Resume bound to trusted identity (server-side, load-bearing)
`hello.resume` is honored ONLY when the session it names is OWNED by the tab's trusted `{origin, uuid, backend}` — its exact-tab record (now identity-bound, see below) or its trusted stable-key entry. An unowned resume is dropped and no longer short-circuits the stable-key fallback. So the panel's global session can't cross-resume a fresh/copied workflow.

### Exact session bound to the FULL identity (origin+uuid), proven-ownership reset
The exact record (`wf:<path>::backend`) was keyed only on the tab id — a saved workflow **overwritten in place** (same path, new uuid), a **cross-origin** copied uuid on the same bridge port, or a **pre-upgrade/legacy** record could resume the prior/other workflow (via `hello.resume` *and* spawn's exact-store hit). Now each exact record is bound to the durable canonical identity `origin::uuid` (`SessionStore.Entry.u` via a manager `identityForKey` dep). On hello, an exact record is kept + resumable ONLY when **proven** — bound identity present, hello identity present, and equal. Every other case (in-place overwrite, unbound legacy record, no trusted identity, different origin) triggers a **full session boundary** (`manager.reset`: stops the live agent + clears pending resume/held mail + the durable session), before the ownership check and spawn.

### Creation-boundary fork made fail-safe
The `isWorkflowCreationLoad` classifier now recognizes a REUSE only when the `loadGraphData` 4th arg is genuinely a ComfyWorkflow (a non-array object with a string `path`); any primitive/array/mis-shaped object forks (bias to fork).

### Deliberate behavior changes (disclosed for your call)
- **Panel-scoped global chat no longer follows across workflows** — resume is bound to identity (per the #570 invariant; a persistent panel-global chat would be a separate authorized feature).
- **One-time reset on upgrade** for a saved-workflow chat whose on-disk record predates the identity binding (a lost resume, once).
- **Identity-less clients** (old panels / no trusted server origin) lose *saved-workflow* resume (reset per reload) — modern browser panels have a trusted handshake origin and prove out, so they're unaffected; unsaved tabs are unaffected (their tmp: id churns).

### Remaining residual (separately tracked — NOT a wrong-resume)
The in-flight wrong-**canvas** race: a `panel_*` command already delivered to the browser microseconds **before** a same-socket workflow switch can act on the new canvas. The orchestrator can't retract a delivered frame; the route-fence closes the not-yet-dispatched window but the full fix needs a **generation-bound panel-command protocol** (every command carries a workflow generation the panel verifies) — a separate cross-cutting workstream, out of scope for this session-resume fix.

Tests: mcp full suite 2831 pass (1 pre-existing unrelated `node-dev` flake excluded); panel full unit suite 1163 pass. Both rebased on latest main. Draft — do not merge.

artokun:
## Update 4 — full identity-boundary teardown; convergence on the deferred residual

The resume-ownership work is complete and the identity boundary now tears down **every** kind of per-tab state on an unproven workflow transition. Across the review cycle codex confirmed no wrong-**resume** remains; its final finding is the separately-tracked in-flight frame/command residual (below).

### Identity boundary — fully fail-closed, all state kinds
On hello, a tab's state is kept **only when positively proven** to be this workflow's identity: the state's identity **and** the hello's identity both present **and** equal (`origin::uuid`). Otherwise (in-place saved-workflow overwrite = same tab id, new uuid; cross-origin copied uuid; unbound pre-`u`/legacy record; or an identity-less client) everything is torn down **before** any replay/spawn/resume:
- live agent + armed pending resume + **failed-start held mail** (`manager.hasAnyState` → `manager.reset`),
- durable exact session (now bound to the full canonical `origin::uuid` via `Entry.u`),
- `heldDuringGen` render-held queue (swept across **all** providers),
- and the bridge's **buffered deliveries** — the bridge now runs `onPanelMessage` first and the reset calls `UiBridge.dropQueuedDeliveries` so a prior workflow's buffered `say`/`session`/`stream`/`show_media` can't replay into the replacement.

State that **is** proven this workflow's (a normal reload/reconnect) is untouched. A proven same-workflow **migration** (tab-id change) rebinds the agent and carries all held queues (all providers); a same-socket **workflow switch** retires + route-fences + cancels held work.

### Scope note (disclosed, for your call)
Browser panels always carry a trusted handshake WS Origin, so they prove out and are fully covered — the modern-panel norm. Only genuinely identity-less clients (non-browser/relay, or transient pre-`u`/old panels) pay a **lost resume** across reload/reconnect (never a cross-workflow leak) — the fail-closed trade-off.

### Remaining residual (separately tracked — the deferred workstream)
An **already-dispatched** frame or `panel_*` command — written to the socket microseconds **before** a same-socket switch — can still land in the newly selected workflow. Bridge frames carry no per-workflow generation, so the orchestrator cannot retract a frame already delivered to the browser; the route-fence + buffered-delivery drop close everything the server controls, but the pre-dispatched window needs a **generation-bound panel-command protocol** (every frame/command tagged with the workflow generation, validated panel-side) — a separate cross-cutting panel+bridge workstream, out of scope for this session-resume fix.

Tests: mcp full suite 2834 pass (1 pre-existing unrelated `node-dev` flake excluded), incl. new identity-transition, spawn-window live-agent, failed-start held-mail, cross-origin, pre-upgrade, and buffered-delivery-drop regression tests. Panel 1163 pass. Draft — do not merge.

artokun:
## Update 5 — structural inversion + legacy-uuid invalidation → codex **APPROVE**

The root pattern behind every prior leak ("teardown gated on an incomplete state-set") is fixed structurally, and independent adversarial codex now **approves**: *"no remaining queued or replayed cross-workflow path found."*

### P0a — inverted, unconditional-and-complete teardown (orchestrator)
The identity boundary now decides **only whether to KEEP** a tab's state, on positive proof (state identity **and** hello identity both present **and** equal `origin::uuid`). When **not** proven, it tears down **everything** for the tab id **unconditionally** — no longer gated on any state-presence check, so a channel the gate didn't enumerate (e.g. a bridge-buffered `show_media`/frame with no manager state) can't slip through. The complete, idempotent sweep: `manager.reset` across **all** providers (live agent + pending resume + failed-start held mail + durable session) + `heldDuringGen` (all providers) + `bridge.dropQueuedDeliveries` (missed frames + mailbox). A future channel is covered automatically.

### P0b — legacy copied-uuid invalidation (panel)
The create-boundary fork only forks *creation* loads, so a graph copied **before** the rollout, reloaded as a ComfyWorkflow (reuse), kept its duplicated embedded uuid. Fixed with a **fork-ownership marker** stamped whenever our fork mechanism mints a uuid; on unsaved load the embedded uuid is trusted **only when marked** — a legacy/unmarked embed is forked to a fresh identity and re-stamped (one-time strip post-upgrade, a lost resume never a leak), durable thereafter. A pre-rollout copied unsaved graph can no longer derive the source's `origin::uuid`.

### Codex final verdict: **approve**
> Ship for the targeted wrong-resume/delivery risk: no remaining queued or replayed cross-workflow path found.

### Sole remaining residual (separately tracked)
A frame/command **already written to the socket** before a workflow switch cannot be recalled by the server — the deferred **generation-bound panel-command protocol** (tag every frame/command with the workflow generation, validated panel-side). Out of scope for this session-resume fix.

Tests: mcp 2834 pass (1 pre-existing unrelated `node-dev` flake excluded); panel 1165 pass. Draft — do not merge.

PR #650, opened 2026-08-01
fix(environment): probe the ComfyUI venv interpreter, never fabricate "not installed" (#401)

### LINKED ISSUE #401: get_environment probes the WRONG Python: false 'Triton: not installed' led an agent to disable working acceleration

The session ENVIRONMENT line reported `Triton: not installed` on a machine where Triton **is** installed and working. The probe appears to inspect the WRONG Python interpreter: `get_environment` reports `local.python.executable = "python.exe"`, `version 3.11.7`, while `running_instance.python_version` is `3.12.11` and `embedded_python: false`. So the package probe reads a bare `python.exe` off PATH rather than the interpreter actually running ComfyUI (a venv here). Triton is absent from the PATH interpreter and present in the ComfyUI one.

Proof it is actually installed and functional -- repeated successful applications in the ComfyUI log, with no import error anywhere:

```
[INFO] Applying LTX2 Memory Efficient Sage Attention Patch to all transformer blocks, triton_kernels=True
[INFO] Applying LTX2 Attention Tuner Patch with custom scales to blocks: [0..47], triton_kernels=True
```

Filtering the logs for `No module named` returns only an unrelated `sam_3d_body` warning -- no triton failure.

### Impact -- this actively caused a wrong edit

The ENVIRONMENT line is presented as authoritative live machine state and is explicitly designated for "the acceleration decision". Trusting it, I told the user that nodes with `triton_kernels=true` would crash, and I **edited their 231-node workflow to disable those kernels** -- silently removing a real speedup they had deliberately enabled. The user corrected me ("si j'ai triton"); I verified via logs and reverted. A less attentive user would simply have lost the acceleration without knowing why.

### Steps to reproduce

1. Run ComfyUI from a venv whose interpreter is NOT the first `python.exe` on PATH, with `triton` installed in that venv only.
2. Read the session ENVIRONMENT line / call `get_environment`.
3. It reports `Triton: not installed`, and `local.python.version` disagrees with `running_instance.python_version`.

### Fix

applied locally: no
upstream-only: yes

Resolve the probe interpreter from the RUNNING ComfyUI rather than from PATH -- derive it from the server process, or query the ComfyUI process's own `sys.executable`. Minimum viable improvements:

1. When `local.python.version` does not match `running_instance.python_version`, suppress definitive `X: not installed` claims -- the probe is provably looking at the wrong environment. Degrade to `unknown` (the tool description already promises graceful degradation).
2. Report `Triton: unknown` rather than `not installed` whenever the probe interpreter is unverified.

The asymmetry matters: a false `not installed` makes an agent degrade a working setup, which is worse than saying nothing at all.

### Environment
- ComfyUI: 0.27.1 (local portable install)
- GPU: RTX 5090 32GB / 94GB RAM / CUDA 13.0 / torch 2.11.0.dev20260123+cu130
- comfyui-mcp: 0.48.6
- panel: shipped with comfyui-mcp 0.48.6

artokun:
## Why

`get_environment` reported `Triton: not installed` on a machine where Triton was installed and working; an agent trusted it and edited a 231-node workflow to disable real acceleration.

**PR #433 already fixed the original cause** (live-first interpreter resolution from the running server's argv, `verified`/`live` signals, `reconcileProbeState` degrading unverified negatives to `unknown`, `python_probe_trusted` on the JSON). That machinery is on `main` and this PR keeps it — it is additive, not a rework.

This PR fixes the **recurrence the reporter filed against 0.48.31** (latest comment on #401): on **ComfyUI Desktop**, `/system_stats` reports a *relative* `argv[0] = "ComfyUI\main.py"` with no `cwd`, so `liveRootFromArgv` returns UNRESOLVED, resolution falls back to `COMFYUI_PATH` = the bundle root, and picks the launcher's `standalone-env/python.exe` — while the server actually runs `<bundle>/ComfyUI/.venv/Scripts/python.exe`. It then reported `python_probe_trusted: true`, because the surviving rule *"reachable + no live root + matching python major.minor"* counted as proof. Installs made against the reported interpreter were invisible to the running server (`ModuleNotFoundError: omegaconf`).

## What changed

**Selecting the right interpreter**
- `serverRootsUnder()` — when a base is not itself a server root but `<base>/ComfyUI/main.py` exists on disk, that nested dir **is** the server root and is searched first (the Desktop and classic-portable shape). A base that has its own `main.py` is never outranked by a nested tree.
- `liveRelDirFromArgv()` + anchoring in `resolveComfyuiPython()` — a *relative* argv `main.py` dir is re-anchored on the configured base, accepted only when a `main.py` really exists there.
- The running server's own `embedded_python` self-report (ComfyUI computes it as `basename(dirname(sys.executable)) == "python_embeded"`) narrows candidates — the tiebreak between a bundle's embedded python and an install `.venv`. If it rules out every candidate we can see, the fallback is returned but never as `live`.

**Never fabricating a negative**
- `resolveComfyuiPython()` now reports `ambiguous` (several distinct environments under one root, e.g. `.venv` beside `standalone-env`) and `proven` (the sole interpreter under the install root the server named in its own absolute argv).
- `proven` is the **only** licence to state a negative: `reconcileProbeState` degrades every other `not-installed` to `unknown`. Positives always pass through untouched — a positive can't be the harmful direction here.
- `get_environment` drops *"matching major.minor ⇒ trusted"*. A live-but-unproven interpreter (inferred root, or ambiguous root) must match the running instance's python **and** — whenever ambiguous, or whenever the server reports one — its **exact** torch version before packages are reported. Otherwise `packages` is omitted and the new `local.python_probe_reason` states why, so "we couldn't determine it" is readable as distinct from "it's absent".

**Deliberately out of scope:** `resolveRootInterpreter` (which chooses the interpreter *pip installs* run under, via `manifest.ts` / `node-management.ts`) keeps its exact pre-#401 selection for every layout, guarded by a new test. With no live server to verify against, preferring a nested venv there would be an unverifiable guess in a mutating path. Making installs themselves live-first is a separate change.

## Tests

`src/__tests__/services/workspace-env.test.ts`, `src/__tests__/services/env-capabilities.test.ts`:
- Desktop relative-argv anchoring picks the **server** venv, not the bundle's launcher python (unit + end-to-end through `getEnvironment`).
- A workspace venv is always preferred over a bare PATH python.
- An argv root that is itself a server root is never overridden by a nested `ComfyUI/` tree.
- `embedded_python` tie-break both ways; a hint matching nothing drops the `live`/`proven` claim.
- Ambiguous roots (`.venv` + `venv`, `.venv` + `standalone-env`) are `ambiguous`, never `proven`.
- Inconclusive ⇒ `unknown`, never `not installed`; anchored/ambiguous probes can never emit a definitive negative.
- Anchored roots degrade to untrusted on a python **or** torch contradiction.
- `resolveRootInterpreter` does not follow the nesting.

`npm run build` clean; full suite green apart from one pre-existing, unrelated failure (`node-dev.test.ts` "builtin fallback" — asserts `engine: "builtin"` but ripgrep is installed on this machine; fails identically on `main`).

Self-gated with `codex exec -m gpt-5.6-terra` — five adversarial rounds, PASS on the last. It caught, and this diff resolves: `../python_embeded` outranking a root's own `.venv` while still claiming to be live; the nested-root preference overriding an exact argv root; version/torch matches being treated as proof of interpreter identity; and the candidate reorder leaking into the pip-install path.

Closes #401

### DIFF
--- a/docs/tools/install-environment.mdx
+++ b/docs/tools/install-environment.mdx
@@ -215,7 +215,7 @@ List known/auto-detected ComfyUI installations on this machine. Scans common ins
 
 ## get_environment
 
-Report ComfyUI environment info (mirrors `comfy-cli env`): the running instance details from /system_stats (OS, Python, ComfyUI version, GPU/VRAM — works for remote targets) plus local probes when a workspace path is available (Python version, git revision, ComfyUI-Manager version, and key pip packages like torch/CUDA). Degrades gracefully: local probes are omitted when no local path is configured or tools are unavailable.
+Report ComfyUI environment info (mirrors `comfy-cli env`): the running instance details from /system_stats (OS, Python, ComfyUI version, GPU/VRAM — works for remote targets) plus local probes when a workspace path is available (Python version, git revision, ComfyUI-Manager version, and key pip packages like torch/CUDA). The local python probe targets the interpreter the RUNNING server uses (its venv / embedded / standalone python, resolved from the live server), never a bare `python` on PATH. Degrades gracefully and NEVER guesses: when the correct interpreter can't be confirmed, `local.python_probe_trusted` is false, `local.packages` is omitted, and `local.python_probe_reason` says why — an absent package list means UNDETERMINED, never 'not installed'.
 
 <Note>This tool takes no parameters.</Note>
 
--- a/src/__tests__/services/desktop-restart.test.ts
+++ b/src/__tests__/services/desktop-restart.test.ts
@@ -74,7 +74,9 @@ beforeEach(() => {
     if (/netstat/i.test(cmd)) {
       return "  TCP    127.0.0.1:8188    0.0.0.0:0    LISTENING    4321\n";
     }
-    if (/lsof/i.test(cmd)) return "4321\n";
+    // findPidByPort probes with `lsof -nP -iTCP:PORT -sTCP:LISTEN -Fpn`, whose
+    // field output is p<pid> / n<addr:port> records — not the terse `-t` list.
+    if (/lsof/i.test(cmd)) return "p4321\nn127.0.0.1:8188\n";
     return "";
   });
   __processControlTestHooks.reset();
--- a/src/__tests__/services/env-capabilities.test.ts
+++ b/src/__tests__/services/env-capabilities.test.ts
@@ -219,7 +219,10 @@ describe("applyStats", () => {
   });
 });
 
-describe("resolveComfyuiPython (#401)", () => {
+describe("resolveComfyuiPython — BEST-GUESS selection only (#401)", () => {
+  // This resolver decides what to PROBE. It deliberately reports no authority
+  // signals any more: which interpreter the server actually runs is answered by
+  // live-interpreter.ts (we launched it / the OS says so), never by layout.
   let dir: string;
   beforeEach(async () => {
     dir = await mkdtemp(join(tmpdir(), "comfyui-py-"));
@@ -228,17 +231,19 @@ describe("resolveComfyuiPython (#401)", () => {
     await rm(dir, { recursive: true, force: true });
   });
 
-  it("returns the on-disk venv interpreter as VERIFIED", async () => {
-    const venvBin = IS_WIN ? join(dir, ".venv", "Scripts") : join(dir, ".venv", "bin");
-    await mkdir(venvBin, { recursive: true });
-    const exe = join(venvBin, IS_WIN ? "python.exe" : "python3");
+  const mkVenvAt = async (root: string): Promise<string> => {
+    const bin = IS_WIN ? join(root, ".venv", "Scripts") : join(root, ".venv", "bin");
+    await mkdir(bin, { recursive: true });
+    const exe = join(bin, IS_WIN ? "python.exe" : "python3");
     await writeFile(exe, "", "utf-8");
+    return exe;
+  };
 
+  it("returns the on-disk venv interpreter as VERIFIED", async () => {
+    const exe = await mkVenvAt(dir);
     const res = resolveComfyuiPython(dir, undefined);
     expect(res.verified).toBe(true);
     expect(res.python).toBe(exe);
-    // No argv → this is NOT the live interpreter (just the pinned workspace's venv).
-    expect(res.live).toBe(false);
   });
 
   it("finds the embedded python of a portable install as VERIFIED (not just .venv)", async () => {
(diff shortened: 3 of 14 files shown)


artokun:
## Review round 2 — all three HIGH findings fixed (pushed as `92ab546`)

Common root cause confirmed and addressed: `proven` meant *"we found the sole interpreter under a root we believe in"* but was **used** as *"we verified this is the interpreter the server is running"*. Those are now separate claims — `proven` documents itself as **location-established, not contents-verified**, and is necessary but never sufficient.

**HIGH 1 — anchored root could emit a trusted package list from the wrong install.** An anchored (relative-argv) root is now **never** trusted for reporting, regardless of fingerprint agreement. It gets its own branch ahead of the generic live branch. The reasoning: a stale `COMFYUI_PATH` aimed at a second bundle of the same shape anchors identically, and sibling Desktop installs routinely share both python and torch — no fingerprint we can gather separates them. Anchoring still drives **selection**, which is what makes a positive capability finding possible on Desktop; it just never backs a claim about the live server.

**HIGH 2 — `proven` granted outside the absolute argv root.** Resolution now tracks `locationEstablished`: the candidate is inside the server's own root, **or** it is the bundle's sibling `python_embeded` *and the server itself reported `embedded_python: true`*. Without that, the interpreter is still probed but never reported from — and critically, the `!locationEstablished` case now takes the same untrusted branch as anchored, so matching python+torch can no longer rescue a guessed location (that was the reporting half the first fix missed).

**HIGH 3 — `proven` bypassed contradictory runtime fingerprints.** `get_environment` now always cross-checks the probed python/torch against `/system_stats` before trusting, and **so does the capability path**: `probeTritonSage` additionally reports torch's version read from package **metadata** via `importlib.metadata` (never `import torch`, which would cost seconds), and `reconcileProbeState` degrades a negative on a torch contradiction as well as a python one. The panel line compares against the **raw** `pytorch_version` — `caps.torch` is stripped of its `+cuXXX` suffix, which is exactly the discriminating part.

**Consistency change, not one of the three:** a reachable server reporting **no install root at all** has strictly less location evidence than the anchored case now rejected — same collision, less to go on. That tier is now untrusted too. It reports nothing rather than a fingerprint-plausible guess.

**Deliberately kept:** absolute-argv **ambiguous** roots still corroborate via an exact torch match. Unlike the rejected cases the *location* is established (the server's own absolute argv root) and only *which* of several environments under it is open — an exact torch build discriminates between siblings in a known location. Flagging the distinction explicitly in case you want it tightened too.

**Untouched, as instructed:** `resolveRootInterpreter` / `manifest.ts` / `node-management.ts`. Still direct-root `candidatesIn` with the pre-#401 order, still guarded by its regression test.

**New tests** — the three scenarios exactly as specified, plus the two follow-ons my self-gate surfaced:
- two-bundle stale-`COMFYUI_PATH` collision where python **and** torch both agree → untrusted;
- absent `embedded_python` with only a parent-relative candidate → not `proven`, negative degrades to `unknown`, **and** not package-trusted (the reporting half);
- proven-root-vs-live-python contradiction → untrusted;
- proven interpreter's negative degrades on a torch contradiction, survives on agreement or an unreported torch;
- rootless tier untrusted even on an exact python+torch match.

`npm run build` clean; 3122 passing. Two failures, both unrelated and reproducible without this diff: `node-dev.test.ts` "builtin fallback" (asserts `engine: "builtin"`, but ripgrep is installed on this machine — fails identically on `main`) and `remote-model-landing.test.ts`'s 20s-timeout watcher, which is load-sensitive under the full parallel suite and passes in isolation.

Self-gate re-run to PASS (rounds 6-7). Round 7 confirmed both blockers dead, the torch probe safe (metadata-only, catches missing package/metadata, never imports torch), and the earlier sign-offs intact — install resolution direct-root only, Desktop selecting the anchored server venv while withholding authority, remote probing suppressed, ambiguity still requiring exact torch. Its one non-blocking note (a stale explanatory comment on the rootless tier) is fixed in the pushed commit.

artokun:
## Redesign pushed (`09ee079`) — inference deleted, ground truth only

Net **−147 lines** across src (843 added / 990 removed), and the trust decision is now one rule instead of a ladder.

### Why inference could never work — measured, not argued

On the machine that filed this bug (ComfyUI 0.29.2 Desktop standalone bundle, relative `argv[0] = "ComfyUI\main.py"`, no cwd), the two candidate interpreters are:

| interpreter | python | triton | sageattention |
|---|---|---|---|
| `ComfyUI\.venv\Scripts\python.exe` (**the server's**) | 3.13 | **True** | **True** |
| `standalone-env\python.exe` (the bundle's) | 3.13 | False | False |

Same python version, opposite capabilities. Every fingerprint available to the old design agrees between the right and wrong interpreter — so the old code's "corroboration" was structurally incapable of separating them, and the loser is exactly the confident `Triton: not installed` that stripped a user's acceleration.

### The rule

1. **We launched it** — `process-control` records the interpreter it spawned, keyed to the child PID, honored only while that PID still owns the port, cleared on exit/error.
2. **The OS tells us** — the server is a local process on our port; the OS reports the command line it was started with, and argv[0] of a running python *is* its interpreter.
3. **Otherwise UNKNOWN** — no package list, no `not-installed`, no trusted. Positives still pass through unconditionally.

### On step 2 — this is why the fix works for the reporter

`/system_stats` exposes **no** `sys.executable`, confirmed against both the live server and ComfyUI's `server.py` (which computes `embedded_python` from `sys.executable` and discards the path). So the HTTP surface is a dead end, as you found.

But the **OS process table** is a third observation channel, and it covers the case step 1 does not: artokun's ComfyUI was launched by *Desktop*, not by us, so a launch-record-only rule would leave his machine permanently UNKNOWN. Reading the process is an observation, not an inference, so it satisfies the terminating rule rather than weakening it.

One subtlety worth recording: on Windows you must read WMI **`CommandLine`**, not `ExecutablePath`. For a venv, Windows reports `ExecutablePath` as the *base* interpreter the venv trampoline loads (`…\standalone-env\python.exe` — the wrong one!) while `CommandLine`'s argv[0] is the venv python that `sys.executable` reports. Measured on the rig; the code and tests pin it.

### Live end-to-end on the reporting machine

```
resolveLiveInterpreter → C:\...\ComfyUI\ComfyUI\.venv\Scripts\python.exe  (process-table, PID 16764)
python_probe_trusted: true      python_probe_source: "process-table"
packages.torch: 2.10.0+cu130  ==  /system_stats pytorch_version: 2.10.0+cu130
```

### Deleted

`proven`, `ambiguous`, `anchored`, `locationEstablished`, `fingerprintMismatch`, torch-as-identity, and the "sole candidate under a believed root" notion. `resolveComfyuiPython` is now documented as a **best guess** that decides what to *probe* and may never back a claim — its selection still lands on the Desktop server venv, which is what makes a *positive* finding possible at all.

### Your two suggestions, both taken

- **Full `python_version`**: both sides now carry the real `sys.version` banner (probe runs `-c "import sys;print(sys.version)"`, falling back to `--version`; the capability probe prints it too and compares against the raw stats string rather than the display-truncated one). Contradiction check only, as you said — cloned venvs report identical banners.
- **`deploy_environment`**: reported on `running_instance` and folded into `python_probe_reason` so the message names the layout (e.g. `local-desktop2-standalone`) when we can't observe the interpreter. Not used for trust.

`python_probe_source` is now `launched-by-us` / `process-table` / `layout-guess`, so "we know" vs "we're guessing" is machine-readable.

### Self-gate

Round 8 FAILed with two real findings, both fixed: POSIX `lsof -ti :PORT` matched *connected* processes (a client could have been read as the server) → now `lsof -tiTCP:PORT -sTCP:LISTEN`, matching what the Windows branch already required; and the full-version check wasn't actually reaching either comparison path. Round 9 PASS.

Untouched: `resolveRootInterpreter` / `manifest.ts` / `node-management.ts`. `port-owner.ts` is a behaviour-preserving extraction that keeps the import graph acyclic (`workspace-env → live-interpreter → port-owner`).

Build clean; 3122 passing, plus 11 new tests for the ground-truth module. The one failure is the pre-existing `node-dev` ripgrep test that fails identically on `main`.

artokun:
Maintainer gate adjudication (final round). The independent review's two remaining findings are accepted, documented trade-offs, not oversights:

1. **Positives pass through unobserved** (`reconcileProbeState`): deliberate, test-pinned asymmetry. #401's harm was a false NEGATIVE that made an agent strip working acceleration; a false positive self-reveals loudly at generation time, and post-#668 probes increasingly come from the observed interpreter anyway. Failing closed on positives too would make the feature report 'unknown' in the common single-install case where the layout read is right.

2. **Relative-argv residual** (two installs, same relative argv): closing it needs process-cwd correlation; tier-2 is argument-only by design and the limitation is documented in `commandLineMatchesArgv`. Tier-1 (launched-by-us, PID + creation-time + absolute path) covers the case we control with full identity.

Everything else the gate raised across 5 rounds was fixed: tier-1 PID+start-time (completed from a stalled placeholder), token-aware argv correlation, macOS corroboration, absolute-interpreter requirement, exact absolute-path matching, and the unvalidated legacy launched channel (deleted — live-interpreter.ts is now the single launch record). CI green on the final head.

Issue #528, opened 2026-07-31
Runtime checker and validator keep stale object_info after custom-node restart


artokun:
After installing ComfyUI-GGUF 1.1.10 and comfyui-videohelpersuite 1.7.9, restarting the Desktop-managed ComfyUI, and confirming the live server exposes /object_info/UnetLoaderGGUF and /object_info/VHS_VideoCombine, comfyui-mcp still uses a stale node registry. get_node_info('GGUF') returns no match, check_workflow_runtime reports UnetLoaderGGUF and VHS_VideoCombine as unknown, and validate_workflow says the graph has no output even though VHS_VideoCombine is an output node. The panel frontend initially also needed a manual frontend reload before panel_add_node accepted UnetLoaderGGUF. Repro: install both packs, restart ComfyUI, query object_info directly (nodes present), then call get_node_info/check_workflow_runtime/validate_workflow. Environment: Windows 11, ComfyUI 0.29.0, comfyui-mcp 0.48.12, comfyui-mcp-panel da5fa15, RTX PRO 4000 Blackwell 24 GB, Python 3.13, torch 2.10.0 CUDA 13.0. No secrets included.

Issue #464, opened 2026-07-30
check_workflow_runtime returns unknown for bundled local-only pack when custom nodes are not installed


artokun:
### What happened / root cause
check_workflow_runtime({pack:"qwen-image-edit"}) returned runtime:"unknown" because several custom-node class types were absent from /object_info. However list_packs explicitly guarantees all bundled packs are local/free. This makes the runtime guard contradict pack metadata and incorrectly demands a paid-credits confirmation for a known local pack.

### Steps to reproduce
1. On a clean ComfyUI without the qwen-image-edit custom nodes installed, call list_packs and confirm qwen-image-edit runtime:"local".
2. Call check_workflow_runtime({pack:"qwen-image-edit"}).
3. Observe runtime:"unknown", usesApiNodes:null, and guidance to ask about paid credits.

### Exact error
No thrown error; misleading classification and guidance.

### Fix
applied locally: no
upstream-only: yes
Suggested change: when the input is a bundled pack, trust/enforce the pack's declared local runtime (or classify unknown custom nodes as local for bundled packs after verifying the pack manifest contains no API nodes). Preserve unknown behavior for arbitrary graph input.

### Environment
Windows 11; ComfyUI 0.28.0; RTX 5070 Ti Laptop 12 GB; CUDA 13.0; torch 2.13.0+cu130; Python 3.13; comfyui-mcp 0.48.12; panel 0.11.10.

Issue #424, opened 2026-07-30
panel_update_node cannot update legacy ComfyUI-Manager itself (HTTP 405)


artokun:
### What happened / root cause
Trying to update the installed `comfyui-manager` pack through `panel_update_node` on ComfyUI-Manager 3.x LEGACY fails at the legacy Manager queue endpoint with HTTP 405. This blocks the in-panel upgrade path from Manager 3.x to v4+.

### Steps to reproduce
1. Run ComfyUI 0.29.0 with ComfyUI-Manager 3.x LEGACY.
2. Call `panel_update_node({id:"comfyui-manager", version:"latest", channel:"default", mode:"remote"})`.
3. Observe immediate HTTP 405.

### Exact error (scrubbed)
`Error: Manager manager/queue/task: HTTP 405`

### Fix
Applied locally: no. Upstream-only: yes.
Suggested change: detect the legacy endpoint response and provide a local git/comfy-cli fallback when the connected Manager pack itself is the update target; otherwise return an explicit upgrade instruction.

### Environment
Windows 11; ComfyUI 0.29.0; NVIDIA GeForce RTX 5060 Ti 16 GB; CUDA 13.0; torch 2.9.1; Python 3.12; ComfyUI-Manager 3.x LEGACY; comfyui-mcp 0.48.7; sidebar panel connected, exact panel build unavailable.

artokun:
The panel-side ComfyUI-Manager 3.x **legacy dialect** was completed in **panel 0.11.9** (comfyui-mcp-panel #250) — legacy reboot route, legacy self-update on 405, and a legacy list fallback. If this report was against a legacy Manager, please update the panel to 0.11.9 and confirm whether it's resolved. Note: these panel_* tool behaviors are served by the panel, so the fix lands via the panel package. If it persists on 0.11.9, reopen with the panel/mcp versions + the exact error and we'll investigate the orchestrator side.

artokun:
Recurrence on a newer stack: comfyui-mcp **0.48.32** (after the v0.48.20 fix) and panel **0.11.28** (after the 0.11.9 legacy-dialect work) on Windows 11 / ComfyUI Desktop 0.29.2 with ComfyUI-Manager **3.41**, reported 159 commits behind upstream.

[...]

PR #653, opened 2026-08-01
fix(manager): invalidate the API dialect cache on restart and on dialect-mismatch failures (#646)

### LINKED ISSUE #646: Invalidate Manager API dialect cache after ComfyUI restarts at same URL
After upgrading ComfyUI-Manager in place from the 3.x custom-node clone to pip comfyui_manager 4.2.2, and restarting ComfyUI at the SAME base URL, download_model kept using the previously cached v2-batch / legacy-UI dialect. The live server reported is_legacy_manager_ui:false and version V4.2.2, but comfyui-mcp still rejected arbitrary URLs as if legacy UI were active.

### Steps to reproduce
1. Start ComfyUI with Manager 3.x or Manager 4 --enable-manager-legacy-ui.
2. Invoke a Manager operation so detectManagerApi caches v2-batch.
3. Restart the same ComfyUI URL with pip Manager 4.2.2 and --enable-manager.
4. Confirm GET /v2/manager/is_legacy_manager_ui returns false and /v2/manager/version returns V4.2.2.
5. Call download_model with a Hugging Face URL.
6. It still routes through runV2BatchTask and returns the legacy-UI arbitrary-URL error.

### Exact error
ComfyUI-Manager API 400 Bad Request for /v2/manager/queue/install_model ... NOTE: this ComfyUI runs Manager v4 ... in LEGACY-UI mode (--enable-manager-legacy-ui).

Live evidence at the same time:
- process args: main.py --listen 127.0.0.1 --port 8188 --enable-manager
- /v2/manager/is_legacy_manager_ui => false
- /v2/manager/version => V4.2.2
- /v2/manager/queue/status => valid empty v4 status

### Fix
Applied locally: attempted once, but the running MCP server could not be reloaded by panel_reload; upstream fix required.

Precise source change: invalidate/re-probe managerApiCache after the connected ComfyUI restarts, or give the dialect cache a short TTL/boot identity. In detectManagerApi, do not return a cached v2-batch/legacy dialect indefinitely solely because the base URL is unchanged. Re-probe /v2/manager/queue/status plus /v2/manager/is_legacy_manager_ui before Manager mutations after reconnect.

Temporary recovery used the exact native Manager v4 endpoint with the required client_id/ui_id ModelMetadata envelope, preserving Manager's queue and target-folder validation.

### Environment
Windows 11; RTX 5070 Ti 16 GB; ComfyUI 0.29.0; Manager 4.2.2; comfyui-mcp 0.48.32; panel 0.11.32.

artokun:
## Root cause

`detectManagerApi()` classifies the connected ComfyUI's ComfyUI-Manager into one of three wire dialects — `v2` (pip Manager 4.x unified task envelope), `v2-batch` (pip Manager under `--enable-manager-legacy-ui`, the bundled 3.x server under `/v2`), `legacy` (the 3.x custom-node Manager) — and cached the verdict **keyed only by base URL, for the whole process lifetime**.

A ComfyUI restart at the **same URL** can change the dialect underneath that key: upgrading the 3.x custom-node clone to pip `comfyui_manager` 4.2.2, or dropping `--enable-manager-legacy-ui`, leaves the URL identical while flipping which endpoints exist. Nothing ever invalidated the entry, so every later Manager call kept speaking the pre-restart dialect. In the report, `download_model` kept routing through `runV2BatchTask` and failed with the legacy-UI arbitrary-URL rejection against a server that answered `is_legacy_manager_ui: false` and `V4.2.2`.

Same class as #528 (stale `/object_info` after a restart), fixed with the same pattern.

## The fix

New leaf module `src/services/manager-api-cache.ts` owns the cached verdict, so all self-healing lives in one place:

1. **Explicit invalidation on the restart lifecycle the repo already has.** `resetManagerApiCache()` is called wherever `resetObjectInfoCache()` already drops live-derived state for a restart, plus the lifecycle points that had no cache reset at all: `process-control` stop / start / supervised-child exit / Manager-reboot-fired / reboot-ready, the panel's managed-restart and Manager-reboot tools, and `comfy_cli_server` stop + launch.

2. **A TTL backstop** (`COMFYUI_MCP_MANAGER_API_TTL_MS`, default 60 s) for the **out-of-band** restart — the reported path, where the user upgraded Manager and restarted ComfyUI themselves so no mcp tool ever observed it. Mirrors the `/object_info` TTL from #528, but measured on a **monotonic** clock so no wall-clock step can extend *or* resurrect a window.

3. **Per-call self-heal** so one stale entry can't wedge every subsequent call. `queueManagerTask` separates **enqueue** from **drain**, and only the enqueue is ever re-sent. Every operation here is a **mutation** and Manager exposes no idempotency key, so a re-send is a genuinely second operation — it therefore happens only when (a) the failure was a **route-level** rejection, 404/405, which proves no handler ran (ComfyUI's catchall answers an unregistered POST with 405), and (b) a fresh probe reports a **different** dialect. An ambiguous 400 (possibly validation, possibly a handler that already acted, possibly a proxy) still refreshes the classification but is **never** re-sent: the caller gets an actionable refusal naming both dialects and stating that the first attempt's effect is unknown. Request bodies are rebuilt per attempt for the dialect being spoken (a git install's params are dialect-specific, not just its route), and the drain uses the dialect that actually enqueued the work.

Probing is bounded on every axis: one shared in-flight detection per (base, epoch); one shared reset+re-detect transaction per base; a cooldown after a "nothing changed" verdict so a persistently-failing endpoint can't buy a probe per call; and, if repeated restarts invalidate every reading, an outright refusal rather than sending a mutation on a dialect known to be stale.

Cache writes are ordered: detections take a ticket before probing and lose to a reset or to any later-started detection, so a slow probe of the pre-restart server can never re-pin itself.

## Codex self-gate (`gpt-5.6-terra`, 6 rounds to PASS)

| Round | Finding | Resolution |
| --- | --- | --- |
| 1 | **P1** `update_all` uses a dedicated route, bypassing the self-heal | extracted `enqueueWithDialectSelfHeal()` and routed both it and `queueManagerTask` through it |
| 1 | **P2** supervisor respawn + timed-out Manager reboot never invalidated | both now reset (the reboot resets as soon as it is confirmed fired, not only on ready) |
| 1 | **P3** test leaked `COMFYUI_MCP_MANAGER_API_TTL_MS` | stashed and restored in `afterAll` |
| 2 | **P2** a wall-clock rollback could *resurrect* an expired entry once wall time caught back up | freshness measured with `performance.now()` |
| 3 | **P1** a git install's params are chosen by dialect, so a retry re-routed the wrong body (a 3.x body at v4 resolves nothing) | task params are a resolver invoked with the dialect each attempt speaks |
| 3 | **P2** `comfy_cli_server` stop/start/restart never invalidated | resets after stop and after launch |
| 4 | **P2** a slower concurrent detection could overwrite a fresher verdict | commit ordering added |
| 5 | **P1** that ordering was by *completion*, so the older probe finishing first won; and `runManagerQueue` re-derived the prefix from the cache, so an operation could enqueue on one generation and drain the other | ordering ticket taken at probe **start**; the drain uses the dialect the enqueue actually spoke |
| 6 | **P3** the cache-derived prefix fallback was now dead | `runManagerQueue(api)` is required; `managerQueuePrefix()` removed |
| review | **P1** every 400 could **re-send** a mutation (a handler that already acted, or a proxy on the response path, also 400s — and Manager has no idempotency key) | only 404/405 may re-send; a 400 refreshes the classification and returns an actionable refusal instead |
| review | **P2** a persistently-400ing endpoint drove a reset+re-probe **per call** | per-base cooldown after a confirmed-unchanged verdict, cleared by every lifecycle reset |
| review | **P2** no single-flight — N callers each fired their own probe round | one shared in-flight detection per (base, epoch) |
| 7 | **P2** `detectManagerApi` still **returned** a reading taken before an invalidation (only the cache write was dropped), so the mutation went out on the dead dialect | re-probes when the epoch moved under it |
| 7 | **P2** concurrent re-checks each reset (bumping the epoch, so the detector couldn't join) and probed independently | the whole reset+re-detect is one shared transaction per base; the cooldown arm is guarded on the post-reset epoch |
| 8 | **P2** `startEpoch` was captured before **detection**, so a mid-detection reset vetoed the #464 v2-batch demotion even when drawn from a successful post-restart batch enqueue | captured after detection, spanning exactly the enqueue (regression test verified to fail against the old ordering) |
| 9 | **P1** on an exhausted invalidation-retry budget the stale reading was returned and used to send a mutation | throws an actionable "ComfyUI kept being restarted, NOTHING was sent" instead |

Final verdict: **PASS**, no findings.

## Tests

`src/__tests__/services/manager-dialect-cache.test.ts` drives one URL whose Manager persona is swapped mid-test:

- restart + explicit invalidation → re-detects and speaks the new dialect;
- within the window → served from cache, no per-call probe;
- out-of-band swap → heals once the TTL lapses;
- a backward wall-clock step neither extends nor revives the window;
- a 400 re-detects but re-sends **nothing** (no enqueue, no queue start) and returns an actionable refusal; the user's reissue then routes correctly on the first attempt;
- a 404/405 route rejection **does** re-send once: a git install's retry carries v4 semantics (repo name + `nightly`, no `files`) rather than the re-routed 3.x body, and `update_all` heals too, starting the queue once on the **live** prefix;
- a 403 neither re-probes nor retries; a persistently-400ing endpoint costs zero probes after the first, and re-checks again once a lifecycle event clears the cooldown;
- five concurrent cold callers share **one** detection; four concurrent failures share **one** reset+re-detect;
- a restart landing mid-detection is neither pinned nor used to enqueue; a restart storm refuses without sending anything; the #464 demotion still pins across a mid-detection reset;
- concurrent detections are ordered by start in both completion orders.

`npm run build` clean; full suite green apart from the pre-existing, environment-dependent `node-dev.test.ts > builtin fallback` (asserts ripgrep is absent — it is installed on this machine; fails identically on `main`).

## Notes

- **Merged with #649** (#424 Manager self-update), which landed the same enqueue/drain split from the other direction. Resolution: `enqueueManagerTask` keeps this branch's `(api, kind, resolveParams, uiId, startEpoch)` signature returning the dialect it spoke, and #649's `updateManagerSelf` now enqueues *through* the self-heal — a 405 only reaches its "this build doesn't support self-update" verdict if it survives a re-probe, and the verdict names the dialect that actually 405'd. A follow-up codex pass on the reconciliation returned PASS (one P3 on that error's dialect label, fixed).

- **Out of scope, filed separately:** the `update_all` **MCP tool** (`src/tools/update-comfyui.ts` → `updateAllCustomNodes()` in `src/services/update-comfyui.ts`) is a separate hardcoded-legacy implementation that has *never* consulted `detectManagerApi` and posts `/manager/queue/update_all` unconditionally, so it misroutes on any v4 host. That is an endpoint-routing gap in a different code path, not a cache-invalidation defect, and it has its own issue.
- **Retarget:** `setComfyuiTarget()` now drops the dialect. The cache is keyed by base URL so a retarget normally just misses, but a local → pod → local round trip can land back on an identical base string whose server was restarted meanwhile — this issue's exact failure mode.
- **Manager shadowing** (a pip `comfyui_manager` coexisting with a 3.x `custom_nodes` clone) is not detectable from the HTTP surface — only one Manager actually loads in the ComfyUI process, so the wire tells you which won, not that two are installed. Diagnosing it needs filesystem inspection and only works for local installs; not attempted here. With this fix the wire dialect is re-probed, so the loaded Manager is always spoken to correctly.

Closes #646

### DIFF
--- a/src/__tests__/services/manager-dialect-cache.test.ts
+++ b/src/__tests__/services/manager-dialect-cache.test.ts
@@ -408,19 +408,37 @@ describe("#646 Manager API dialect cache invalidation", () => {
     expect(getCachedManagerApi(BASE)).toBe("v2-batch");
   });
 
-  it("a slower concurrent detection cannot overwrite a fresher verdict", async () => {
-    // Both probes start after the window lapsed, with NO reset involved (an
-    // out-of-band restart lands between them), so the epoch alone can't order
-    // them: probe A reads the OLD server, probe B reads the new one and commits
-    // first, then A resolves last.
-    const stampA = managerApiCacheStamp();
-    cacheManagerApi(BASE, "v2", managerApiCacheStamp()); // probe B, fresher
-    cacheManagerApi(BASE, "v2-batch", stampA); // probe A, stale, resolves last
+  it("orders concurrent detections by START, so the newest reading wins either way", async () => {
+    // Both probes start after the window lapsed with NO reset involved (an
+    // out-of-band restart lands between them), so the epoch can't order them and
+    // completion order says nothing about which reading is newer. A starts first
+    // and reads the OLD server; B starts second and reads the NEW one.
+
+    // Case 1 — the newer probe finishes FIRST, the older resolves last.
+    let stampA = managerApiCacheStamp();
+    let stampB = managerApiCacheStamp();
+    cacheManagerApi(BASE, "v2", stampB);
+    cacheManagerApi(BASE, "v2-batch", stampA); // stale, must be dropped
     expect(getCachedManagerApi(BASE)).toBe("v2");
 
-    // A verdict backed by a SUCCEEDED enqueue (the #464 demotion) is still
-    // allowed to supersede a probe-derived one — it guards on the epoch only.
+    // Case 2 — the OLDER probe finishes first: its verdict lands, and the
+    // later-started probe must still be able to replace it (this is the order a
+    // write-counter guard gets wrong: it would drop the fresher reading).
+    resetManagerApiCache("test");
+    stampA = managerApiCacheStamp();
+    stampB = managerApiCacheStamp();
+    cacheManagerApi(BASE, "legacy", stampA);
+    expect(getCachedManagerApi(BASE)).toBe("legacy");
+    cacheManagerApi(BASE, "v2", stampB);
+    expect(getCachedManagerApi(BASE)).toBe("v2");
+
+    // A verdict backed by a SUCCEEDED enqueue (the #464 demotion) supersedes a
+    // probe-derived one — it guards on the epoch only …
     cacheManagerApi(BASE, "v2-batch", { epoch: managerApiEpoch() });
+    expect(getCachedManagerApi(BASE)).toBe("v2-batch");
+    // … and it takes a fresh ticket, so a detection that started BEFORE it can
+    // no longer clobber it.
+    cacheManagerApi(BASE, "legacy", stampA);
     expect(getCachedManagerApi(BASE)).toBe("v2-batch");
   });
 
@@ -448,10 +466,11 @@ describe("#646 Manager API dialect cache invalidation", () => {
 
     // The parked detection concluded AFTER the invalidation, so its verdict
     // described a server that may no longer be there and must NOT be pinned over
-    // the reset. The very next detection inside this same operation (the queue
-    // drain resolves the route prefix) therefore has to probe again: two
-    // detections, not one. Without the epoch guard the stale verdict would have
-    // been re-pinned and the second detection served from cache.
-    expect(detections(calls)).toBe(2);
+    // the reset — it left the cache EMPTY. The next operation therefore probes
+    // again; without the epoch guard the stale verdict would have been re-pinned
+    // and this call served from cache with no probe at all.
+    calls.length = 0;
+    await downloadAModel();
+    expect(detections(calls)).toBe(1);
   });
 });
--- a/src/__tests__/services/manager-dialect-cache.test.ts
+++ b/src/__tests__/services/manager-dialect-cache.test.ts
@@ -131,8 +131,12 @@ function stubServer(opts: {
     if (path === "/v2/manager/queue/start") return new Response("", { status: 200 });
     if (path === "/v2/manager/queue/update_all") return new Response("", { status: 200 });
     if (path === "/v2/manager/is_legacy_manager_ui") {
+      // Read the persona BEFORE parking on the gate, so a gated probe answers
+      // with the reading it took when the request was issued — that is what makes
+      // "the server changed while a probe was in flight" testable.
+      const body = { is_legacy_manager_ui: opts.persona() === "v2-batch" };
       if (opts.legacyUiGate) await opts.legacyUiGate();
-      return jsonResponse({ is_legacy_manager_ui: opts.persona() === "v2-batch" });
+      return jsonResponse(body);
     }
 
     if (opts.persona() === "v2-batch") {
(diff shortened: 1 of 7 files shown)


artokun:
Superseded by #670, which contains this branch's full commit stack plus the retarget-pin fixes (both codex-gated PASS). Merging there.

Issue #641, opened 2026-08-01
A backup folder left inside custom_nodes is served as a live web extension and silently shadows the real panel


artokun:
### What happened / root cause

After installing the panel from source (workaround for #639), I left a backup of the previous version at `custom_nodes/.comfyui-agent-panel.bak-0.11.28`. ComfyUI **serves every directory under `custom_nodes` as a web extension — including dot-prefixed ones**. `/api/extensions` then listed the panel twice:

- 55 files under `/extensions/.comfyui-agent-panel.bak-0.11.28/js/...` (0.11.28)
- 57 files under `/extensions/comfyui-agent-panel/js/...` (0.11.32)

`.` sorts before `c`, so the **stale backup won registration** and the user kept running 0.11.28 after updating, restarting, and hard-refreshing. Confirmed by the sidebar wordmark resolving to `http://127.0.0.1:8188/extensions/.comfyui-agent-panel.bak-0.11.28/img/comfyui-mcp-wordmark.svg`.

**The trap is an asymmetry in ComfyUI itself:** dot-prefixed dirs are *skipped* by the Python custom-node loader but *not* skipped by the web-extension server. So a backup that looks safely hidden (invisible to the node loader, hidden in the file browser) is fully live to the frontend — and wins by sort order.

This cost a full diagnostic cycle. Every disk-level check said 0.11.32; only reading `/api/extensions` in the browser revealed the shadow. Version verification that reads the file rather than the served asset cannot catch this.

### Steps to reproduce

1. `cp -r custom_nodes/comfyui-agent-panel custom_nodes/.comfyui-agent-panel.bak-X`
2. Update the real panel to a newer version.
3. Reload the frontend (even hard-refresh, even after a ComfyUI restart).
4. `fetch('/api/extensions')` → both copies listed; the dot-prefixed one loads.
5. `PANEL_VERSION` on disk says new; the running panel is old.

### Exact error (scrubbed)

No error — silent shadowing. Observable only as:

```
/extensions/.comfyui-agent-panel.bak-0.11.28/js/comfyui-mcp-panel.js   <- actually loaded
/extensions/comfyui-agent-panel/js/comfyui-mcp-panel.js                <- ignored
```

### Fix

applied locally: yes (moved the backup outside `custom_nodes`, to `%TEMP%\comfyui-panel-backups\`)
upstream-only: no, for the guard below

Verified after the move, without any ComfyUI restart — `/api/extensions` resolves per request from disk:

| check | before | after |
[...]

Issue #639, opened 2026-08-01
install_panel(update) reports success but does not update the panel on disk (silent no-op via stale ComfyUI-Manager)


artokun:
### What happened / root cause

`install_panel({action:"update"})` returned a **success** payload — `"Queued + updated \"comfyui-agent-panel\" via ComfyUI-Manager."` with `done_count: 2`, `is_processing: false`, `restartRequired: true` — but **nothing on disk changed**. The pack stayed at 0.11.28 while 0.11.32 was current.

This is a **silent failure**: the tool trusted ComfyUI-Manager's queue-drained signal as proof of work done, without verifying the installed version actually moved. An agent (or user) acting on that response would restart ComfyUI expecting 0.11.32 and still be on 0.11.28, with no error anywhere to explain it.

Same environment as #424 — bundled ComfyUI-Manager v3.41, 159 commits behind, whose update path is broken (405 on `/manager/queue/update_all`, task exceptions on single-pack update). Here, though, Manager reported *done* rather than erroring, so the failure surfaced as a false positive instead of a visible error.

### Steps to reproduce

1. ComfyUI Desktop with a stale bundled ComfyUI-Manager (v3.41 here).
2. Panel installed at an older version (0.11.28), latest is 0.11.32.
3. Call `install_panel({action:"update"})` → returns success, `done_count: 2`.
4. `grep PANEL_VERSION custom_nodes/comfyui-agent-panel/web/js/comfyui-mcp-panel.js` → still `0.11.28`.

### Exact error (scrubbed)

No error — that's the bug. Response was:

```json
{
  "action": "update",
  "result": {
    "mechanism": "manager-http",
    "message": "Queued + updated \"comfyui-agent-panel\" via ComfyUI-Manager.",
    "details": { "total_count": 0, "done_count": 2, "in_progress_count": 0,
                 "pending_count": 0, "is_processing": false }
  },
  "restartRequired": true,
  "message": "Panel update queued via ComfyUI-Manager (nightly). RESTART ComfyUI to load the updated panel node."
}
```

Disk before and after: `const PANEL_VERSION = "0.11.28";` — unchanged. Note `total_count: 0` alongside `done_count: 2` is itself incoherent and could serve as a cheap red flag.

### Fix

applied locally: no (worked around by installing from source)
upstream-only: yes

Precise change needed in `install_panel` — **verify, don't assume**:

[...]

PR #657, opened 2026-08-01
feat(skill): agent-driven panel node-pack sync on orchestrator update

artokun:
## Why

The orchestrator ships on npm; the sidebar panel node-pack (`comfyui-agent-panel`) ships on the Comfy Registry. **Updating one does not update the other**, so users end up running a new orchestrator against an old panel — and the failures that produces (a bridge command the panel simply doesn't implement, a documented sidebar feature that isn't there) are confusing and hard to self-diagnose.

Community request: **Discord `#help`, from `seanmcmagic`** — "orchestrator updated, panel didn't; can it offer to sync?" No issue number for this one.

Unblocked by the hardened install/update-verify path that just landed (`0a4d98b`, `00df2c3`, `4c3aed3`, closing #639/#641). This feature is built **on top of** that verified path and inherits its discipline.

## What

A new skill, **`plugin/skills/panel-node-pack-sync/SKILL.md`** (36 → 37 skills), that walks an agent through: detect the version mismatch → check for a pin → warn-or-sync → report the **verified** outcome.

Supporting logic, all behind the existing `install_panel` tool — **no new tool** (`TOOL_NAMES`, `MAX_TOOLS`, the baseline SHA and `tool-vocabulary.json` are untouched, so this doesn't collide with the in-flight consolidation).

### Version pin — new, minimal

There was no pin mechanism in the codebase, so this adds one, following the existing `panel-settings.ts` accessor pattern:

- Stored as `panelPin` in `~/.comfyui-mcp/panel-settings.json`, with a `COMFYUI_MCP_PANEL_PIN` env override (**env > `.env` > json**, matching the rest of the config since `.env` is loaded into `process.env` at boot). `COMFYUI_MCP_PANEL_PIN=off` is an explicit "no pin" escape hatch.
- **Enforced at the mutation choke point** (`runPanelAction`), not in the skill — so `install`/`update`/`reinstall`/`sync` *and* the on-load auto-installer all refuse for every caller, not just the ones that remembered to check.

Owner's three constraints, and where each lives:

| Constraint | Implementation |
|---|---|
| **Respect an explicit pin** | `runPanelAction` throws before queuing anything; `ensurePanelInstalled` skips; `performPanelSync` refuses without calling the mutation at all. |
| **Warn when pinned** | `decision: "pinned-warn"` — names the newer version the orchestrator needs, says they're pinned and *where* the pin lives, and changes nothing. |
| **Allow unset/reset** | `install_panel(action='unpin')` clears it and reports whether a sync can now proceed. An env pin is reported as still in force rather than falsely cleared. |

### The two invariants

1. **Never report a sync that didn't happen.** The mutation goes through `runPanelAction`, which fails closed unless the pack provably moved on disk (#639) and no shadow copy is serving (#641). `performPanelSync` then **re-reads the pack from disk** and reports `verifiedVersion` — the version *observed*, never the one intended. `stillBehind` is tri-state (`null` = the landed version isn't comparable, so the match could **not** be confirmed — deliberately `null` and not `undefined`, which `JSON.stringify` would drop).
2. **Never move a pinned user.** Every "can't tell" resolves to *pinned*: an unparseable settings file, a present-but-malformed pin key, or a pin reader that throws all refuse to move the panel.

### `install_panel` surface

`action` gains `sync`, `pin`, `unpin` (plus optional `version`/`reason`); `status` additionally returns the pin and a `sync` assessment. Existing actions behave as before when nothing is pinned.

## Self-gate

`codex exec -s read-only -m gpt-5.6-terra`, scoped to *"can this ever report a sync that didn't happen, and can it move a pinned user off their pin?"* — **4 rounds, FAIL → FAIL → FAIL → PASS**, 12 findings fixed. The ones worth knowing about:

- **Critical:** `typeof [] === "object"`, so a settings file containing `[]` parsed as valid → read as *unpinned*, **and** a pin written to it was silently dropped (an array expando doesn't survive `JSON.stringify`) while the tool reported "Pinned to X". A user would have believed they were protected and weren't. Arrays/null/scalars are now `unreadable` (= pinned), and set/clear **verify by re-reading**.
- **High:** pin writes and panel mutations weren't serialized, so a pin could commit after an op's final pin check but before ComfyUI-Manager touched disk. Both now share one in-process lock (which independently stops two concurrent ops each reading the other's half-applied state — that would break the #639 before/after proof).
- **High:** `action='pin'` claimed protection even when `COMFYUI_MCP_PANEL_PIN` overrode it. Four resolved outcomes are now reported distinctly, and only the genuine `=off` case says the panel is unprotected.
- **Medium:** a *present but malformed* `panelPin` (`null`, `false`, `0`, a bare string) read as unpinned, and clear keyed on truthiness so a falsy one could never be removed.
- **Medium:** a **failed** shadow scan returned `shadows: []` with the caveat only in prose, so the sync assessment read it as an all-clear. It's now structural (`shadowInspectFailed`) and decides `blocked`.
- **Medium:** the hand-rolled semver screen accepted `01.11.28` / `0.11.28+.`; it now reuses ui-bridge's canonical strict SemVer 2.0.0 grammar, which existed for exactly this reason.

Round 4 verdict: *"No remaining path found that reports a completed sync/update/install without verified disk movement… No remaining path found that moves a currently pinned user."*

## Tests

`src/__tests__/services/panel-sync.test.ts` (new) plus additions to the panel-installer and panel-settings suites — the required decision matrix (mismatch+no pin → sync; mismatch+pin → warn only, **zero** mutations queued; pin cleared → sync proceeds; equal → no-op; verification failure → throws, never a success claim), the choke-point guard, pin-store precedence and every fail-closed path.

## Verification

`npm run build`, `npm run lint`, `npm test` (green — the one `node-dev` failure is pre-existing on this machine: it asserts the builtin search fallback while ripgrep is on `PATH`, and fails identically on `main`), `asset-counts --check`, `check:vocabulary`, `vocab:export --check`, `docs:gen` (no-op), and `smoke-install.mjs`.

Not live-tested against a real ComfyUI-Manager update — the verified path it delegates to was live-tested in #647.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Works on **macOS**, **Linux**, and **Windows**. Auto-detects your ComfyUI instal
 
 **Stuck or have a question? [Join the Discord](https://discord.gg/cW9arBhzCu)** — help, model tips, and release announcements.
 
-**162 MCP tools** | **36 AI skills** (Flux · WAN · LTX 2.3 video · Qwen · Z-Image · Ideogram 4 · ERNIE · ANIMA · model registry · Civitai · node authoring · launch/perf flags) | **55 installer packs** | **11 slash commands** | **4 autonomous agents** | **3 hooks**
+**162 MCP tools** | **37 AI skills** (Flux · WAN · LTX 2.3 video · Qwen · Z-Image · Ideogram 4 · ERNIE · ANIMA · model registry · Civitai · node authoring · launch/perf flags) | **55 installer packs** | **11 slash commands** | **4 autonomous agents** | **3 hooks**
 
 The plugin ships **expert skills that grow with every release** — model-specific generation guides with curated download URLs, workflow recipes, troubleshooting, and custom-node authoring — so Claude knows the right sampler, CFG, resolution, and model files for each architecture without trial and error.
 
--- a/docs/local-vs-comfy-cloud.mdx
+++ b/docs/local-vs-comfy-cloud.mdx
@@ -44,7 +44,7 @@ ElevenLabs without owning a GPU.
 ## What this project offers
 
 `comfyui-mcp` is an MCP server plus a sidebar agent that drives **your** ComfyUI
-install — local, LAN, VPS, or RunPod. 162 MCP tools, 36 model-specific skills,
+install — local, LAN, VPS, or RunPod. 162 MCP tools, 37 model-specific skills,
 and a panel agent that edits your live graph in natural language.
 
 The structural difference is **where the reasoning comes from**. This project is
--- a/docs/plugin.mdx
+++ b/docs/plugin.mdx
@@ -1,6 +1,6 @@
 ---
 title: "Claude Code Plugin"
-description: "comfyui-mcp is a full Claude Code plugin: 36 AI skills, 11 slash commands, 4 autonomous agents, and 3 hooks on top of the 162 MCP tools — expert ComfyUI knowledge that grows with every release."
+description: "comfyui-mcp is a full Claude Code plugin: 37 AI skills, 11 slash commands, 4 autonomous agents, and 3 hooks on top of the 162 MCP tools — expert ComfyUI knowledge that grows with every release."
 icon: "puzzle-piece"
 ---
 
@@ -15,7 +15,7 @@ without trial and error.
 /plugin install comfy
 ```
 
-## 36 AI skills — and growing
+## 37 AI skills — and growing
 
 Skills are curated knowledge documents Claude loads on demand. Each model
 family gets generation parameters, node graphs, curated model download URLs,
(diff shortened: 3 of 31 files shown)


artokun:
Maintainer gate adjudication (final). Twelve rounds of independent review; every concrete finding was fixed: pin check+act atomic under the op lock (incl. update_all/snapshot-restore/generic node tools), embedded git refs honored or refused (never silently nightly), lock completion contract pinned by test, TOCTOU closed by deciding+mutating inside the lock, the workflow-deps panel door peeled off the generic queue (note-only now, with scan-reliability guards), PanelStatus carries scanReliable and can't-tell blocks everywhere (sync/deps/manifest), honest installed reporting, and the pinned-drift warning shows in EVERY state incl. remote/dev-install/blocked.

The two remaining findings are the same async-window class raised each round: queued update_all / deferred snapshot-restore work can still move a panel pinned afterward. Accepted as a documented trade-off and filed as the next issue: a pin governs all FUTURE operations; in-flight Manager work predating the pin warns via the pending-op marker rather than pretending it can be retro-cancelled. The real fix (cancel the queued task or pin-re-check at completion) is tracked in the new issue. CI green.

PR #669, opened 2026-08-02
fix(manager): pin dialect retries to original target

artokun:
## Summary\n- pin Manager dialect probe, recheck, retry, queue drain, and install verification to the ComfyUI target selected at invocation\n- prevent a concurrent panel retarget from sending an in-flight operation to a different ComfyUI\n- add regression coverage for stale-dialect self-heal plus retarget\n\n## Validation\n- 
px tsc --noEmit --pretty false\n- 
pm test -- src/__tests__/services/manager-dialect-cache.test.ts\n- 
pm test (one unrelated baseline failure: 
ode-dev.test.ts > builtin fallback finds matches and skips binary files expects uiltin, received 
ipgrep)\n\nStacked on #653 (ix/manager-dialect-cache-646); merge this after #653.


artokun:
Superseded by #670: this first stacked comparison accidentally included unrelated main commits. #670 is the clean one-commit retarget fix on #653.

Issue #663, opened 2026-08-01
get_image rejects valid MP4 assets instead of saving them to disk


artokun:
### What happened / root cause
The comfyui get_image tool rejects valid video assets even though its documentation and adjacent tool guidance imply non-image outputs can be saved to disk for inspection. The /view request successfully returned content-type video/mp4, but get_image treated any non-image MIME type as IMAGE_NOT_FOUND. This blocks fetching a saved ComfyUI MP4 for last-frame extraction and forces an exact-path filesystem workaround.

### Steps to reproduce
1. Generate a valid MP4 with VHS_VideoCombine.
2. Confirm it exists using list_output_images.
3. Call get_image with filename, subfolder, type:"output", and save_dir.
4. Observe IMAGE_NOT_FOUND while details report mimeType:"video/mp4".

### Exact error (scrubbed)
{"error":"IMAGE_NOT_FOUND","message":"ComfyUI /view did not return an image for \"I2V_HD_FaceLock_00036-audio.mp4\" (output/Eros); got content-type \"video/mp4\".","details":{"filename":"I2V_HD_FaceLock_00036-audio.mp4","type":"output","subfolder":"Eros","mimeType":"video/mp4"}}

### Fix
Applied locally: no. Upstream-only: yes.
Local source/package was not present in the workspace for a safe one-attempt patch. Precise change needed: allow get_image's save-to-disk path to accept video/audio MIME types (or add a general get_media/get_file tool), write response bytes to save_dir using the original extension, and only reject unsupported inline rendering. Keep inline image output limited to image MIME types.

### Environment
OS: Linux
GPU: NVIDIA GeForce RTX 5090, 31 GB VRAM
ComfyUI: 0.19.0 (remote)
Python: 3.12.3
comfyui-mcp: 0.48.21
panel: version unavailable in environment snapshot

artokun:
Resolved by merged #673 (d4aa80d): valid MP4/media saves now pass the full byte-proof gate.

Issue #385, opened 2026-07-29
get_image returns JSON parse error for ComfyUI input image


artokun:
### What happened / root cause
Calling get_image for an MJ Image input slot returned malformed or empty tool output that the MCP client could not parse. Likely the get_image implementation assumes a JSON response where the /view endpoint returns binary image data, or it does not handle a missing input filename cleanly.

### Steps to reproduce
1. Open a workflow with MJ Image configured for folder qwen, select 6.
2. Call get_image with filename 06.png, subfolder qwen, type input, and a Windows save_dir.
3. Observe the tool error.

### Exact error (scrubbed)
Unexpected end of JSON input

### Fix
applied locally: no  upstream-only: yes
Return a structured not-found error when the requested input file does not exist, and handle binary /view responses without attempting JSON parsing.

### Environment
Windows 11 / ComfyUI 0.29.0 / NVIDIA GeForce RTX 5060 Ti 16 GB / comfyui-mcp 0.48.5 / panel version not present in environment snapshot.

artokun:
Fixed in **mcp 0.48.8** (#435): get_image returns a structured not-found for non-image /view payloads (common for type=input refs) instead of a corrupt inline image / JSON parse error. Please update and reopen if it persists.

artokun:
Reopened: this reproduces on the current release, comfyui-mcp 0.50.46 (Windows 11, Node 24.16.0), using a local server whose `/view` returns `404 text/plain "not found"` for `missing.png`. `fetchImage` still emits a JSON `SyntaxError` rather than the `IMAGE_NOT_FOUND` result intended by #435.

The new report identifies why: `client.fetchApi` throws on 4xx before `fetchImage` reaches its `if (!res.ok)` branch. The same unreachable-after-`fetchApi` pattern is also present in the upload and settings paths. Draft #1178 improves the non-JSON diagnostic but does not make the existing endpoint-specific branches reachable.

artokun:
Claiming this. Draft PR: #1187.

The reopen diagnosis is right, and it was mine — I found it while fixing #828 and filed it here rather than widening that PR. Worth stating the scope precisely so nobody re-tests the wrong thing:

`Client.fetchApi` throws for every status outside [200, 400), so it never *returns* a 4xx. #435's `IMAGE_NOT_FOUND` sits behind `if (!res.ok)` in `fetchImage`, which is therefore unreachable — the fix was real but the branch it added has never executed.

[...]

PR #673, opened 2026-08-02
fix(tools): get_image saves valid MP4 assets to disk (#663)

artokun:
Refs #663

## Problem
`get_image` rejected any non-image /view payload as `IMAGE_NOT_FOUND`, so a valid MP4 produced by e.g. VHS_VideoCombine (content-type `video/mp4`) could not be saved to disk — blocking last-frame-extraction workflows and forcing an exact-path filesystem workaround.

## Fix
- `getOutputImage` (src/services/image-management.ts) gains an opt-in `{ allowMedia }` flag that also accepts `video/*` and `audio/*` payloads. The junk-body guards from #385/#483 (empty body, JSON/HTML error pages) still apply, and the default stays image-only so the other consumers (`view_image`, color-analysis, image-convert, storage-upload) are unchanged.
- `get_image` (src/tools/image-management.ts) passes `allowMedia: true`, saves the bytes to `save_dir` under the original filename/extension as before, and only renders the inline image block for `image/*` payloads — video/audio return a text-only result with the saved path and mime type. No new path restrictions.
- Tool description updated; `npm run docs:gen` output committed (docs/tools/assets-images.mdx).

## Tests
- Service (image-management-traversal.test.ts): mp4 rejected by default, resolved with `allowMedia`, JSON error body and empty body still rejected with `allowMedia`.
- Tool (get-image-binary-safe.test.ts): mp4 saved to the requested save_dir under the original name, text-only response (no inline image), handler passes `{ allowMedia: true }`.
- `npx tsc --noEmit` clean; full `npx vitest run` green except the known pre-existing node-dev builtin-fallback/ripgrep environment failure.

### DIFF
--- a/docs/tools/assets-images.mdx
+++ b/docs/tools/assets-images.mdx
@@ -33,7 +33,7 @@ Fetch a registered asset's bytes and return them as an inline image so the agent
 
 ## get_image
 
-Fetch a generated image from ComfyUI and return it as an inline image. Works with remote ComfyUI instances — does not require COMFYUI_PATH. Use get_history first to obtain the filename.
+Fetch a generated image from ComfyUI and return it as an inline image. Video/audio outputs (e.g. a VHS_VideoCombine .mp4) are saved to save_dir with their original extension instead of being rendered inline. Works with remote ComfyUI instances — does not require COMFYUI_PATH. Use get_history first to obtain the filename.
 
 ### Parameters
 
--- a/src/__tests__/services/image-management-traversal.test.ts
+++ b/src/__tests__/services/image-management-traversal.test.ts
@@ -187,6 +187,52 @@ describe("getOutputImage — non-image /view payloads (issue #385)", () => {
   });
 });
 
+describe("getOutputImage — video/audio media (issue #663)", () => {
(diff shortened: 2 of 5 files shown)


artokun:
Maintainer gate adjudication (final round). Two findings from the last review are accepted residuals, by design:

1. **A 16-byte structurally-valid ftyp-only payload passes as mp4** — the box itself is well-formed; a server that truncates mid-asset is broken upstream of us, and we report faithfully what was saved. The junk-body gate exists to stop error pages/empty bodies being saved as media (#385/#483 class), which it now does exhaustively: declared-MIME exact-format matching, ftyp box structure, RIFF/FLAC/Ogg/ID3 structural minimums, cross-format and cross-family rejection, extension claims (media AND image), and a 256-byte floor for signature-only formats.
2. **Signature+floor formats (EBML, bare MPEG/ADTS sync)** — full frame validation needs bitrate/samplerate tables per MPEG version; disproportionate for this gate, documented in the code.

The save path itself is unchanged from the merged #636 lineage (live-config destination + os.path.join drive semantics); this PR only widens what get_image will *accept* (opt-in allowMedia) and then only when bytes, declared type, and filename extension all agree.

Issue #665, opened 2026-08-02
Fix stale LTX-2.3 FP8 checkpoint URL in bundled img2vid pack


artokun:
### What happened / root cause
The bundled `ltx-2.3-img2vid` workflow embeds `https://huggingface.co/Lightricks/LTX-2.3-fp8/resolve/main/ltx-2.3-22b-dev.safetensors` for three checkpoint consumers. Hugging Face returns 404 because the FP8 repo filename is `ltx-2.3-22b-dev-fp8.safetensors`; the non-suffixed `ltx-2.3-22b-dev.safetensors` lives in `Lightricks/LTX-2.3` and is what the pack manifest/default widgets expect.

### Steps to reproduce
1. Read/load `packs/ltx-2.3-img2vid/workflow.json`.
2. Follow or use its embedded checkpoint model URL.
3. Download fails with 404.

### Exact error
`Download failed: 404 Not Found`

### Fix
Applied locally: yes. Upstream-only: no.
Updated the three executable node `properties.models[].url` entries in the img2vid pack from the FP8 repo to `https://huggingface.co/Lightricks/LTX-2.3/resolve/main/ltx-2.3-22b-dev.safetensors`, matching the manifest and widget filename. The Markdown note and equivalent sibling packs should be updated in source generation as well. For the actual 16GB user workflow, the agent intentionally installs an alternate uncensored LTX-2.3 checkpoint instead.

### Environment
Windows 11; ComfyUI 0.29.2; RTX 4080 SUPER 16GB; comfyui-mcp 0.48.22; panel 0.11.32.

artokun:
Resolved by merged #688/#665 (3052321): bundled LTX-2.3 FP8 URLs were corrected.

Issue #664, opened 2026-08-02
Fix SFW CivitAI search selecting mixed-NSFW model versions


artokun:
### What happened / root cause
search_civitai_models with nsfw:false returned model 1403959 and selected version 1666048, then surfaced an explicit adult trained word. The model-level CivitAI flag is nsfw:false, but that selected version has preview nsfwLevels 1,2,8 and mixed trained words. The search path appears to gate only on the model-level nsfw flag, not the selected version's media levels/content.

### Steps to reproduce
Call search_civitai_models with query "subtle camera motion cinematic", types ["LORA"], base_models ["Wan Video"], nsfw false, Highest Rated.

### Exact error (scrubbed)
Wrong SFW output: version 1666048 is returned even though its preview levels include 8 and its trainedWords include explicit adult anatomy/motion language.

### Fix
applied locally: no  upstream-only: yes
The running comfyui-mcp source path could not be located within one bounded search. Precise change needed: when nsfw=false, reject versions whose preview media includes levels above PG-13, and select a clean version when available before serializing trainedWords. For model 1403959, versions 1599906 and 1586994 have only levels 1,2 and contain the relevant camera-motion triggers.
Add a regression test that nsfw:false never serializes trainedWords from a mixed/adult version.

### Environment
Windows 11 / ComfyUI 0.29.2 / RTX 5090 D 32 GB / comfyui-mcp 0.48.23 / panel 0.11.20

artokun:
Resolved by merged #692 (6e2dec3), including the final unclassified-version follow-up.

Issue #515, opened 2026-07-30
Add per-download cancellation for download_model/download_civitai_model


artokun:
### What happened / root cause
A user asked to cancel one specific model download started by `download_civitai_model`. comfyui-mcp exposes `download_status` but no cancel/abort tool. The process-global download registry stores only the settled promise and job metadata; `downloadModel` is called without an AbortSignal/AbortController. Therefore one download cannot be cancelled without restarting the entire MCP server, which also interrupts unrelated downloads.

### Steps to reproduce
1. Start two large local downloads using `download_civitai_model`.
2. Record their returned IDs.
3. Attempt to cancel only one ID.
4. Observe there is no `cancel_download`/`download_cancel` tool or service function.

### Exact error
No thrown error. Missing capability forced an unsafe all-downloads server-respawn workaround.

### Fix
Applied locally: no. Upstream-only: yes for this installed package because a safe fix needs cancellation plumbing through the download job registry and physical stream/cache layers.

Precise change needed:
- Store an `AbortController` per in-flight `DownloadJobEntry`.
- Thread its `signal` through `downloadModel` and the HTTP/cache streaming implementation.
- Add an MCP tool such as `cancel_download({id})` that resolves the public job ID, aborts only that controller, closes the writer, marks the job `cancelled`, preserves the resumable partial/cache, and leaves other jobs running.
- Make cancellation idempotent and test concurrent downloads plus same-destination dedup/adoption.

### Environment
Windows 11; ComfyUI 0.28.3; RTX 5070 Ti Laptop GPU 12 GB; comfyui-mcp 0.48.19; sidebar panel version not present in the supplied environment line.

PR #674, opened 2026-08-02
Fix CivitAI SFW version selection and stale LTX-2.3 pack URL (#664, #665)

artokun:
Two small bundled fixes, one commit each.

## #664 — SFW CivitAI search selecting mixed-NSFW model versions

The `/api/v1/models` search gates only the **model-level** `nsfw` flag, so an `nsfw:false` search could surface a model whose **latest version** fronts mature/explicit previews (`nsfwLevel` 4/8+) and adult `trainedWords` — e.g. model 1403959 selected version 1666048 (preview levels 1,2,8) and serialized its explicit trigger words, while clean versions 1599906/1586994 (levels 1,2) held the relevant triggers.

- `src/services/civitai-resolver.ts`: in SFW mode, `toSearchHit` now prefers the newest version whose preview media all stay at/below PG-13 (`nsfwLevel <= 2`; ladder 1 None · 2 Soft · 4 Mature · 8 Explicit). When no version qualifies, the latest version is still returned for the download handoff, but its `trained_words` are no longer vouched for. `nsfw:true` behavior unchanged (latest version, words included).
- `src/__tests__/services/civitai-resolver.test.ts`: regression tests — clean-version selection over a mixed-NSFW latest, word suppression when NO version is clean, and the unchanged `nsfw:true` path.

Note: the panel-side NSFW-leak fix is tracked separately (panel #515); this PR is the comfyui-mcp side only.

## #665 — Stale LTX-2.3 FP8 checkpoint URL in bundled packs

`packs/ltx-2.3-img2vid/workflow.json` (and sibling `ltx-2.3-txt2vid`) embedded `https://huggingface.co/Lightricks/LTX-2.3-fp8/resolve/main/ltx-2.3-22b-dev.safetensors` for the three checkpoint consumers — Hugging Face returns **404** because the FP8 repo file is `ltx-2.3-22b-dev-fp8.safetensors`, while the non-suffixed file the manifest/install scripts/node widgets expect lives in `Lightricks/LTX-2.3`.

- Updated the three executable node `properties.models[].url` entries plus the Markdown model-links note in both workflows to `https://huggingface.co/Lightricks/LTX-2.3/resolve/main/ltx-2.3-22b-dev.safetensors`.
- HEAD-verified: stale URL → 404; new URL → 302 → CDN (42.98 GB, matches `check-model-urls.mjs` floor). `node scripts/check-model-urls.mjs` passes 5/5 on both packs.

## Verification

- `npx vitest run` — 3160 passed, 1 failed (pre-existing `node-dev` ripgrep-fallback env failure, unrelated)
- `npx tsc --noEmit` — clean
- No tool-surface changes, so docs were not regenerated.

Refs #664, refs #665.


artokun:
Reopening per owner: the close was accidental (suspected duplicate — it isn't; nothing on main covers either half). Both halves passed the codex gate and CI: CivitAI SFW version-selection (#664) + verified LTX-2.3 checkpoint URL (#665).

Issue #486, opened 2026-07-30
panel_ask loses a validated response after tool timeout


artokun:
### What happened / root cause
A `panel_ask` choice card was displayed and the user explicitly validated a choice, but the MCP tool call never returned the selected value. After 300 seconds it failed with `timed out awaiting tools/call after 300s`. The validated response was therefore lost across turns. This appears to be an upstream request-lifetime/state handoff issue: panel responses should be persisted and delivered even when the agent turn/tool wait crosses a continuation boundary.

### Steps to reproduce
1. Call `panel_ask` from the ComfyUI sidebar agent.
2. Let the agent turn end while awaiting the card response.
3. User validates a choice in the card.
4. Resume/wait on the original tool call.
5. Observe timeout instead of the selected value.

### Exact error (scrubbed)
`tool call failed for panel/panel_ask: timed out awaiting tools/call after 300s`

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: persist the card response keyed by call/card id independently of the original MCP request lifetime, and let a resumed agent retrieve the already-submitted value; alternatively return a stable card id immediately and expose a read-result operation.

### Environment
macOS (darwin), MPS 16 GB, ComfyUI 0.29.0, comfyui-mcp 0.48.15, panel 0.11.4.

artokun:
Recurrence on comfyui-mcp 0.49.3 (current) with comfyui-mcp-panel 0.11.32, macOS/MPS, and ComfyUI 0.29.0. Leaving a `panel_ask` card unanswered still ends in `timed out awaiting tools/call after 300s` instead of the expected clean timeout result; the report identifies the remaining timing margin as insufficient. The panel installation is behind the current 0.11.38 release; update it before retesting.

PR #683, opened 2026-08-02
fix(ui-bridge): reuse the original rid on mutation retry + surface late completions (#517)

artokun:
Orchestrator half of the duplicate-mutation fix for #517, complementing the panel-side command ledger in artokun/comfyui-mcp-panel#521.

## Problem

`dispatch()` minted a **fresh `randomUUID()` rid per attempt**, and a late reply to a timed-out rid was silently dropped (`pending.get(rid)` miss, except the #486 ask_user buffer). So: bridge reply-timeout → the agent re-issues the same mutation → the panel applies it **twice** (duplicate/orphan nodes), and the first attempt's late completion vanishes.

## Fix (correlation only — no timeout/ack budgets touched)

- **Stable rid per logical command.** `SendCtx` now carries `rid`, minted once in `send()` and reused verbatim by every re-dispatch (the existing idempotent read-resume path included). Genuinely new commands still mint fresh rids.
- **Rid reuse on mutation retry.** A mutating command whose attempt ends ambiguously (reply-timeout, or an OUTCOME-UNKNOWN mid-command drop) is recorded in a bounded (64), TTL'd (5 min) ledger keyed by a canonical fingerprint of the logical command (canonical tab id + cmd args; bridge-owned `rid`/`workflow_uuid` excluded). A retry of the same mutation **adopts the original rid**, so the panel's #521 ledger answers it with the ORIGINAL reply instead of applying the mutation a second time. Miss/expiry fails open into the pre-#517 behavior.
- **Truthful late completion.** A late reply for a timed-out mutation is logged and buffered; a retry within the TTL is answered with that outcome directly — *"applied after timeout"*, no frame re-sent at all (the panel ledger's replay semantics, one hop earlier; also covers pre-#521 panels when the reply did arrive). ask_user buffering (#486) is untouched and still runs first.
- Records are cleared once any reply for the rid reaches a caller (a later identical command is genuinely new), purged by `dropQueuedDeliveries` on an unproven workflow replacement, and cleared on `stop()`.

## Panel-ledger contract assumptions

- The panel dedupes purely on `rid` (all commands, not just mutations) and answers a duplicate with the ORIGINAL reply, awaiting it if the first execution is still in flight — including `ok:false` replies (a failed command's replay re-sends the same error without re-executing; mirrored bridge-side by the buffered-error path).
- The ledger is module-scoped, so it survives a socket supersede; a full page reload wipes it. In the wipe case a correlated retry simply re-executes — identical to today's behavior (fail-open).
- Replaying the same rid on a reconnected socket is valid and reaches the same ledger.

## Tests

`src/__tests__/services/ui-bridge.test.ts` (+7): retried mutation reuses the original rid and resolves with the deduped reply; late success **and** late error replies answer the retry with no re-dispatch (and are surfaced via `logger.info`); fresh rids after a success and for different args; OUTCOME-UNKNOWN disconnect correlates the post-reconnect retry; the idempotent read-resume path still resolves the caller, replaying the same rid.

- `npx vitest run src/__tests__/services/ui-bridge.test.ts` — 128/128 pass
- Full suite `npm test` — 221 files, 3335 passed / 2 skipped
- `npx tsc --noEmit` — clean

### DIFF
--- a/src/__tests__/services/ui-bridge.test.ts
+++ b/src/__tests__/services/ui-bridge.test.ts
@@ -2593,4 +2593,197 @@ describe("UiBridge (mutation retry dedupe — #517)", () => {
     await expect(promise).resolves.toMatchObject({ errors: [] });
     a2.close();
   });
+
+  // Gate finding 1 — the fingerprint must match the frame's ACTUAL JSON wire
+  // semantics: a key JSON drops (undefined value) is absent from the frame and
+  // must not differentiate the fingerprint, while an explicit null IS on the
+  // wire and must stay distinct (a different command, never an rid adoption).
+  it("fingerprint follows JSON wire semantics: undefined ≡ missing key, explicit null stays distinct", async () => {
+    const sock = await connectPanel("tab-dedupe-8");
+    await vi.waitFor(() => expect(bridge.tabs().some((t) => t.tab_id === "tab-dedupe-8")).toBe(true));
+    const frames = collectFrames(sock);
+    // Attempt 1 carries an undefined-valued arg — JSON.stringify DROPS it from the
+    // frame, so the wire command is exactly {cmd, type}.
+    await expect(
+      bridge.send(
+        { cmd: "graph_add_node", type: "KSampler", extra: undefined },
+        { tabId: "tab-dedupe-8", timeoutMs: 60 },
+      ),
+    ).rejects.toThrow(/did not reply/);
+    const rid1 = frames[0].rid as string;
+    expect(frames[0]).not.toHaveProperty("extra"); // confirm the key never hit the wire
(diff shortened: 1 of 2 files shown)


artokun:
Maintainer gate adjudication (final round). Every gate finding across three rounds was fixed: rid reuse on re-dispatch, late-reply buffering with truthful 'applied after timeout', single-slot pending serialization, in-flight eviction protection, migration-aware tab identity (#570 uuid), JSON-wire canonicalization (undefined≡missing, null distinct, array holes, toJSON with the property key, boxed primitives unwrapped), caller-rid override, and consumed-reply re-answer. The boxed-primitive class cannot occur from MCP-JSON callers but is guarded anyway as cheap insurance. CI green on the final head.

Issue #689, opened 2026-08-02
Pin write during a pending update_all/snapshot-restore window only warns — queued Manager work can still move the panel


artokun:
Filed from the #657 gate (raised consistently across rounds; adjudicated for merge as documented trade-off).

Shape: `update_all` (update-comfyui.ts) and deferred snapshot restore (node-snapshots.ts) queue Manager work and release the panel op lock — correctly, since holding it across an asynchronous Manager queue would wedge it. A pin written afterward is honored for all FUTURE operations, and the pending-op marker (panel-pin-guard.ts) warns that the queued change may still land. But the already-queued work can still move the now-pinned panel.

The owner's constraint is 'never auto-update over an explicit pin'. True closure needs one of:
1. On pin write during an active pending-op window, attempt to CANCEL the queued Manager task (needs a Manager queue-delete/clear route for the specific task — check what the dialect exposes) and report what happened;
2. Failing a cancel route, the completion/restart path should re-check the pin before applying/reporting and leave the panel untouched when pinned (truthful 'not applied — pinned' outcome).

Until then: the pin governs all future ops, and the user is warned about the in-flight one — documented in panel-pin-guard.ts and the sync skill.

Refs #657, install-panel.ts:124, update-comfyui.ts, node-snapshots.ts.

artokun:
Maintainer investigation: no safe MCP-only fix exists on current Manager APIs. Both legacy and v4 expose only whole-queue reset (which discards unrelated queued work and cannot stop an active task); snapshot restore is server-side/deferred to arbitrary restart with no cancellable task or completion hook. Existing marker + warning is the only honest behavior. Resolution requires Manager per-task cancel/unschedule APIs or Manager-side pin enforcement.

artokun:
Reconfirmed by #702 gate: whole-queue reset is unsafe (drops unrelated work) and cannot prove the panel task did not win the status/reset race. The experimental cancellation PR was closed; existing warning/marker behavior remains the only honest MCP-only approach until Manager exposes per-task cancellation or pin enforcement.

artokun:
[...]

PR #702, opened 2026-08-02
fix(panel): pin write cancels pending update_all / snapshot restore (#689)

artokun:
Refs #689. Filed from the #657 gate: a pin written during a pending `update_all` / snapshot-restore window only **warned** — the queued ComfyUI-Manager work could still move the now-pinned panel, violating the owner's constraint *"never auto-update over an explicit pin"*.

## What changed

The pin-write critical section (`install_panel(action='pin')`) now attempts to **cancel** each active pending-op marker, inside the same lock, before reporting:

- **`update-all`** → `POST <prefix>/manager/queue/reset` via `resetQueue(base)` (now takes an optional base, `ManagerFetchOptions.base` convention). The reset is aimed at the server the marker was queued on — captured in the marker at enqueue time — so a retarget between enqueue and pin can't send it to the wrong ComfyUI (or wipe an innocent queue). A marker with no recorded base falls back to the current target and says so.
- **`snapshot-restore`** → deletes `<manager files>/startup-scripts/restore-snapshot.json` before the next ComfyUI restart consumes it — the exact file `POST /snapshot/restore` writes and `prestartup_script.py` applies (`if os.path.exists(...)`), verified against ComfyUI-Manager 3.x `main` and the v4 pip package. Local installs only; remote mode reports **"cannot cancel — remote host"** and names the manual remedy.

## Provably honest reporting — never "the panel was saved"

Queue counts are read **before and after** the reset (v4 `pending_count`; 3.x `total−done−in_progress`) and the pin result's new `pendingOpCancellations` field names the numbers:

| outcome | meaning | marker |
|---|---|---|
| `cancelled` | reset dropped N pending tasks, none remain, nothing running (shared-queue blast radius stated) | cleared |
| `partially-cancelled` | pending dropped, but task(s) already running — in-flight work cannot be cancelled | kept + warning |
| `already-drained` | queue empty+idle / no restore file — nothing left to cancel; the panel may **already** have moved | cleared |
| `already-running` | nothing pending, worker busy — no reset even sent; cannot cancel | kept + warning |
| `could-not-verify` | state unreadable, reset/delete failed, incoherent counts (#639 signature), or marker itself wouldn't clear — a blind reset is never sent | kept + warning |
| `cannot-cancel-remote` | deferred restore file lives on the ComfyUI host | kept + warning |

Markers are cleared via the new `clearPanelPendingOp` (exact `kind`+`queuedAt` match, read-back verified, fails closed) **only** for provably-cancelled / proven-drained work. The pin itself always lands; refusal semantics are unchanged.

`PanelPendingOp` now also carries `base` + `uiId` of the update_all enqueue attempt that landed (v4 derives per-pack task ids as `${ui_id}_${pack}`), so a v4 host can later answer "did the panel's task already run" via `/v2/manager/queue/history` (capture only — no history query yet).

## Tests

- New `panel-pin-cancel.test.ts` (15 tests) drives the real pin tool handler: cancel on the marker's server after a retarget, reset failure, unverifiable state (nothing sent), in-flight, partial, already-drained, legacy-3.x arithmetic + unprefixed reset route, incoherent counts, local/remote/not-scheduled snapshot cases, no-markers (zero Manager calls), and base/ui_id capture at enqueue.
- `panel-pin-guard.test.ts` gains `clearPanelPendingOp` + marker round-trip tests.
- Full suite: **229 files / 3741 tests passed** (2 skipped); `tsc --noEmit` clean. No tool descriptions changed, so no docs regen.

### DIFF
--- a/src/__tests__/services/panel-pin-cancel.test.ts
+++ b/src/__tests__/services/panel-pin-cancel.test.ts
@@ -484,8 +484,38 @@ describe("pin write cancels a pending update_all (#689)", () => {
     expect(report.note as string).toMatch(/WARNING — a panel-affecting operation is still pending/);
   });
 
-  it("partial: the dropped count is DERIVED from before/after reads, never asserted", async () => {
-    recordUpdateAllMarker({ base: ORIG, uiId: "ui-5" });
+  it("partial with a concurrent RE-FILL (pending 2 → 3): drop UNPROVEN, never a 'dropped' claim", async () => {
+    recordUpdateAllMarker({ base: ORIG, uiId: "ui-5b" });
+    const mine: QueueState = { pending: 2, inProgress: 0, done: 1 };
+    v4Persona(
+      { pending: 2, inProgress: 0, done: 3 },
+      {
+        mine,
+        onReset: () => {
+          mine.inProgress = 1; // a task dequeued mid-reset…
+          mine.pending = 3; // …and a concurrent enqueue added MORE than was dropped
+        },
+      },
+    );
+
+    const report = await writePin();
+
+    expect(resetPosts()).toHaveLength(1);
+    const [cancel] = cancelReports(report);
+    expect(cancel.outcome).toBe("partially-cancelled");
+    expect(cancel.markerCleared).toBe(false);
+    expect(cancel.pendingBefore).toBe(2);
+    expect(cancel.pendingAfter).toBe(3);
+    // mineAfter.pending > mine.pending: a "dropped" count would be zero or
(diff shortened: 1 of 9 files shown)


artokun:
NO-SHIP: Manager reset_queue discards unrelated pending work; counts cannot prove the panel task did not run between status/reset, so cancellation can be falsely reported; Ubuntu CI also fails due same-millisecond pending-marker identity collision. Current Manager APIs lack safe per-task cancel/atomic pin enforcement. Preserve the existing warning/marker contract until upstream support exists.

PR #699, opened 2026-08-02
0.49.0 slice: consolidate queue/jobs (8→1)

artokun:
## Summary

Fourth 0.49.0 tool-consolidation slice (after #644 bisect, #658 snapshots/batch/apps, #684 comfy_cli): the eight queue/jobs tools fold into ONE action-parameterized **`queue`** tool, same flat-enum mechanics as the merged templates.

| retired | replacement |
|---|---|
| `get_queue` | `queue` (action:"list") |
| `get_queued_workflow` | `queue` (action:"get_workflow") |
| `move_queued_job` | `queue` (action:"move") |
| `edit_queued_job` | `queue` (action:"edit") |
| `get_job_status` | `queue` (action:"status") |
| `cancel_job` | `queue` (action:"cancel") |
| `cancel_queued_job` | `queue` (action:"cancel_queued") |
| `clear_queue` | `queue` (action:"clear") |

- **Same services, same return shapes.** Each action calls the identical `queue-manager` function with identical arguments; JSON for query actions, exact prose for `cancel_queued`/`clear`, escalation message + `isError` on a wedged `cancel`.
- **Flat `z.enum` action** (never `discriminatedUnion` — the SDK renders that as zero visible params). Only `action` is schema-required; per-action presence (`prompt_id`, `position`) is enforced in the handler and names the missing field. Guards test **absence, not falsiness** — `prompt_id:""` still reaches the service's own error path, as before.
- **Ratchet artifacts:** TOOL_NAMES 8→1 in the same registration slot; +8 DEAD_NAMES (`since:"0.49.0"`); MAX_TOOLS 162→155 (== TOOL_NAMES.length); `queue` appended to tool-surface.txt; BASELINE_SHA256 → `5f4e68f4…` (LF bytes); tool-vocabulary.json regenerated (155 core / 35 dead).
- **Whitelist:** the orchestrator `call_tool` whitelist collapses `cancel_job` → `queue`; the comment now states the risk posture over the whole tool (the old entry already allowed wiping all pending jobs via `clear_pending:true`).
- **Prose rewritten, not exempted** — every model-facing mention (tool descriptions/messages, panel + headless hints, 5 skills, 2 slash commands, monitor-progress, arena scenarios, README, ROADMAP, concepts/configuration) now uses the action form. The only exemptions: 5 per-occurrence `allowedIn` entries for the dated tdqs-case-study post, whose dead-name mentions are the audit's verbatim evidence.
[...]

PR #703, opened 2026-08-02
0.49.0 slice: consolidate workspace (3→1)

artokun:
Fold the workspace family into one action-parameterized `workspace` tool, following the pattern slices 1-5 proved for bisect_* (#644), snapshots/batch/apps (#658), comfy_cli_* (#684), queue/jobs (#699), and model-metadata (#701):

  get_workspace         -> workspace (action:"get")
  set_default_workspace -> workspace (action:"set_default")
  list_workspaces       -> workspace (action:"list")

Pure surface consolidation. The workspace-env services are untouched: each action calls the identical function the old tool called, with the identical arguments, and returns the identical JSON content block. `get_environment`, which shares the registration module, is untouched.

Shape is a FLAT `{ action: z.enum([...]) }` plus the union of the per-action parameters (NOT z.discriminatedUnion, which the MCP SDK renders as zero visible params). `action` is the only schema-required field; every VALUE constraint stays at the zod layer (`path` keeps its `.min(1)`, so an empty string fails exactly as before), and per-action PRESENCE (`path` for set_default) is enforced in the handler, which names the missing field — the one deliberate behavioural difference a flat enum permits. The guard tests ABSENCE, not falsiness: an empty path still reaches the service's own non-empty validation, as it did before.

Ratchet artifacts:
- TOOL_NAMES: 3 names removed, `workspace` added in the same registration slot (order is asserted by registry-surface.test.ts)
- DEAD_NAMES: +3 entries `{since:"0.49.0", replacement:'workspace (action:"...")'}`
- MAX_TOOLS 153 -> 151 (net -2), still == TOOL_NAMES.length
- docs/design/tool-surface.txt: append `workspace` (the 3 old names stay as history — deleting them is what disables the ratchet)
- BASELINE_SHA256 -> 72f079e2... (LF bytes, matches CI)
- tool-vocabulary.json regenerated (core 151, dead 41)
- asset-count prose 153 -> 151 (README x2, plugin.mdx, local-vs-comfy-cloud.mdx, local-llms.mdx); gen-tool-docs category + regenerated install-environment.mdx (12 -> 10 tools) and models.mdx (list_extra_paths description)

Prose rewritten rather than exempted: every live mention of the old names — the extra-paths / model-resolver / node-authoring / node-dev / node-verify / output-dir error strings that tell the caller to save a default workspace, the `list_extra_paths` tool description, and the surrounding code comments — now names `workspace` with the action form. No allowedIn exemptions needed: the only other mentions are the CHANGELOG (HISTORICAL) and the retirement baseline itself.

The set_default action carries the #701 lesson forward: `path` is per-action-required, so the description shows the full call shape (`{action:"set_default", path:"/opt/ComfyUI"}`) and gen-tool-docs gets an EXAMPLE_ARG_OVERRIDES entry — the generated skeleton only carries schema-required fields.

The orchestrator's call_tool whitelist is untouched: none of the three retired names was whitelisted, so no CALL_TOOL_ACTION_WHITELIST scope was needed (unlike slice 4's queue, whose cancel entry required action-scoping).

Gates green: build, tsc --noEmit, check:vocabulary (1034 files, 41 dead, no live refs), vocab:export --check, asset-counts --check, docs:gen (no-op on re-run), smoke-install, full vitest (229 files, 3742 passed, 2 skipped).


artokun:
Holding merge: GitHub currently reports DIRTY against main after #699/#701. Please rebase onto current main and rerun the full CI/vocabulary gates; the consolidated vocabulary needs a clean post-merge baseline before it can be merged.

Issue #694, opened 2026-08-02
Mutation retry needs an explicit caller retry-identity (the #683 revert's unsolved half)


artokun:
Follows #687 (revert of #683).

Problem: after a bridge reply-timeout, a caller re-issuing a mutation gets a fresh rid. Two live consequences:
1. Double-apply is possible when the original timed-out command applied late (the #517 class — panel-side ledger from #521 alone can't help, since rids are never reused today).
2. A timed-out-but-APPLIED mutation is silently invisible to the caller (late replies are dropped except ask_user).

The #683 approach (bridge infers 'retry' from an args fingerprint) was correctly reverted: a timed-out mutation and a genuinely NEW identical mutation are indistinguishable at the bridge API boundary, so inferred retry can suppress a real repeat.

The workable shape, per #687's rationale: the CALLER supplies retry identity — e.g. an optional `retry_of` (rid or client-supplied idempotency key) on bridge commands. First dispatch of a logical operation mints/carries the key; a retry declares it; the panel dedupe ledger (#521, already merged) can then safely answer duplicates with the original outcome, and late completions surface truthfully as 'applied after timeout'. Bridge must never infer retry from args.

Also still open from the #508 verification: the session-epoch handshake (an orchestrator restart at the same URL within the replay window can inherit the previous session's non-sensitive journal) — worth folding into the same design (epoch stamped into every frame).

Refs #517, #521, #683, #687, #508.

artokun:
Shipped in #704 (orchestrator) + panel #543 (merged earlier), across six gate rounds:

- **retry_of** (explicit caller retry identity): mutating panel tools accept it; a dispatched reply-timeout or OUTCOME-UNKNOWN disconnect names the attempt's rid as a retry token; minting requires retry-map admission + typed markers only; reads never mint or carry it (find_nodes/canvas/screenshot/list_subgraphs excluded); caller cmd.rid can never override the bridge rid.
- **Panel**: retries ({rid: fresh, retry_of}) dedupe via the #521 ledger — fingerprint strips retry_of, fallback lookup with rid-rewrite, unknown token executes fresh.
- **Session epoch**: per-process UUID on the models handshake; journal replay and summaries filter by it; one captured instant serves both (no boundary disagreement).

[...]

Issue #543, opened 2026-07-31
Context meter denominator ratchets up: switching to a smaller-window model under-reports context fill (5x)


artokun:
## Summary

The panel's context-usage ring scores fill against a denominator that only ever ratchets **upward**. Because the model can change mid-session (`setModel` is live — no session restart), switching to a model with a *smaller* context window leaves the denominator pinned at the largest window any model in the session ever reported. The meter then silently under-reports how full the context is, right up until the real limit is hit.

Fix applied in the working tree, with regression tests. Details below.

## Root cause

`src/orchestrator/panel-agent.ts`, `result` event handler (was line 997):

```ts
if (ev.contextWindow && ev.contextWindow > this.contextWindow) {
  this.contextWindow = ev.contextWindow;
}
```

The `>` guard makes this a max-ratchet. But the field's own doc comment (line 272) reads *"Context window for the **active** model, cached from result.modelUsage"* — the ratchet contradicts the stated intent.

It is genuinely reachable, not theoretical: the composer has a live model picker, and `setModel` (line 430) is explicitly documented as *"setModel is live: no session restart, the next turn uses it."* So a user switching from a 1M-context model down to a 200k one keeps the 1M denominator forever.

**Impact:** at 100k tokens used, the ring shows **10%** when the true figure is **50%**. A 5x under-report, in the exact direction that matters — it makes a nearly-full context look comfortable.

There was a second, narrower window of wrongness: between the model switch and the *next* `result` event, the cache still held the outgoing model's value, so the entire next turn was scored against the wrong denominator.

## Fix

1. **Track the latest reported window, not the maximum.** Drop the `>` comparison; the `ev.contextWindow &&` guard already rejects `0`/`undefined`.
2. **Clear the cache on a model switch.** `setModel` now resets `contextWindow = 0`. `reportStatus` omits `contextPct` while the cache is `0`, so the panel holds its previous reading rather than displaying a knowingly-wrong one, and the next `result` refreshes it for the incoming model. `used` (the raw token count) is still reported throughout.

## Verification

New test file `src/__tests__/orchestrator/context-window-denominator.test.ts`, built on the existing `effort-carry.test.ts` harness pattern (fake `AgentBackend` + `PanelAgentManager`):

- *"a SMALLER window reported later replaces the larger one (no max-ratchet)"* — turn 1 reports a 1M window, turn 2 reports 200k; asserts the final `contextPct` is `used/200_000`, not `used/1_000_000`.
[...]

PR #704, opened 2026-08-02
feat(panel): explicit retry identity (retry_of) for outcome-unknown mutations (#694)

artokun:
Orchestrator half of #694. The panel half lands separately in comfyui-mcp-panel; the wire contract is shared:

- `retry_of?: string` rides on mutating command frames, verbatim.
- Outcome-unknown mutating failures name the dispatched attempt's rid in the error text plus one line: `To retry this exact mutation, re-issue identical args plus retry_of:"<rid>"; otherwise call normally.`
- The `models` push frame carries a per-process `epoch`.

## What changed

**Retry token on OUTCOME-UNKNOWN mutating failures** (`src/orchestrator/panel-tools.ts`, `makePanelToolCtx`'s `call`)
When a MUTATING panel command fails with a post-write reply timeout or a mid-command disconnect — the bridge's only two `dispatched:true` rejections — the error text now names the dispatched attempt's rid (captured via the existing `onDispatchedRid` observer, same pattern as `workflow_open`) as the caller's explicit retry token. Pre-write refusals (`dispatched:false`) mint NO token (nothing was sent — nothing to dedupe), reads never mint one (`requiresWorkflowStampEnforcement` gate), and genuine executor `ok:false` errors carry none (a definite outcome). The bridge never reuses rids and never infers retries (#683/#687 lesson).

**`retry_of` threading** (`buildPanelToolDefs`)
Every MUTATING panel tool — `graph_*` outside `BRIDGE_READONLY_CMDS` plus the four workflow mutators (`workflow_save`/`save_as`/`rename`/`close`), i.e. exactly the `requiresWorkflowStampEnforcement` surface — accepts `retry_of: z.string().optional()` and forwards it, untouched, on its command frame (explicit `RETRY_TOKEN_CMD_BY_TOOL` map + a per-def wrapper that attaches the token to mutating commands only, so a read probe inside the same handler stays clean; `panel_screenshot`'s direct-send path threads it by hand). No signature changes; `retry_of` is opaque caller data to the bridge — never stamped or stripped (contrast `workflow_uuid`, bridge-owned). Navigation (`workflow_open`/`workflow_new`) and reads are excluded. A few view/list commands (`graph_find_nodes`, `graph_list_subgraphs`, `graph_screenshot`, …) sit outside `BRIDGE_READONLY_CMDS` and therefore accept the token per the design's mutation-surface formula — harmless: optional, opaque, and deduping an idempotent read is a no-op.

**Hardening** (`src/services/ui-bridge.ts`)
`frame = { rid, ...cmd }` let a caller-supplied `cmd.rid` override the bridge-minted rid, silently breaking reply correlation (`pending` is keyed on the minted rid). Now `{ ...cmd, rid }` — the bridge-owned rid always wins.

**Session epoch** (`src/services/ui-bridge.ts`, `src/orchestrator/index.ts`)
`SESSION_EPOCH` — one `randomUUID` per orchestrator process, minted at module load near `startUiBridge` — is stamped ONLY on the `models` push frame (via an exported pure `buildModelsPushFrame`, also the test seam); no per-frame stamping. The panel can scope its retry dedupe cache to the process that minted the rids, so a restarted orchestrator's tokens never collide with a prior process's.

## Tests

New `src/__tests__/orchestrator/panel-retry-identity.test.ts` (20 tests): (a) mutating reply-timeout AND mid-command disconnect carry the dispatched rid as token (incl. `workflow_save`); (b) `dispatched:false` refusal mints none; (c) executor `ok:false` carries none; (e) reads never emit tokens (timeout, drop phrasing, and the realistic park-expired path); (f) zod accepts optional `retry_of` on every mapped mutating tool, reads lack the field, nothing requires it, map integrity vs `requiresWorkflowStampEnforcement`; (h) two models frames in one process share the same UUID epoch. Plus defensive no-rid-no-token.

`src/__tests__/services/ui-bridge.test.ts`: (d) `retry_of` reaches the wire verbatim while `rid` stays a fresh bridge UUID; (g) a caller `cmd.rid` never wins (and the send still resolves — correlation intact).

Full suite: 229 files / 3743 passed / 2 skipped. `npx tsc --noEmit` clean. `check:vocabulary` clean. `npm run docs:gen` produced no real changes (panel tool schemas are outside its surface; only line-ending churn, reverted).

Refs #694

### DIFF
--- a/src/__tests__/orchestrator/panel-retry-identity.test.ts
+++ b/src/__tests__/orchestrator/panel-retry-identity.test.ts
@@ -1,0 +1,307 @@
+// #694 — EXPLICIT caller retry identity for MUTATING panel commands.
+//
+// When a mutating panel command fails OUTCOME-UNKNOWN (a post-write reply
+// timeout or a mid-command disconnect — the bridge's two dispatched:true
+// rejections), the error text must name the dispatched attempt's rid as the
+// caller's retry token: re-issue identical args plus retry_of:"<rid>". A
+// pre-write refusal (dispatched:false), a genuine executor ok:false error, and
+// any READ mint NO token. retry_of is opaque caller data: the tools accept it,
+// the handlers forward it to the cmd bag, and the bridge ships it to the wire
+// untouched (see ui-bridge.test.ts for the wire-level assertions).
+//
+// The ctx-under-test is the REAL makePanelToolCtx (the token logic lives in its
+// `call`); only the bridge is faked, mirroring the reconnectingBridge harness in
+// panel-session-rebind-residuals.test.ts.
+
+import { describe, expect, it, vi } from "vitest";
+import { z } from "zod";
+import {
+  buildPanelToolDefs,
+  makePanelToolCtx,
+  RETRY_TOKEN_CMD_BY_TOOL,
+  type PanelToolCtx,
+  type ToolResult,
+} from "../../orchestrator/panel-tools.js";
+import {
+  markDispatched,
+  markReplyTimeout,
+  requiresWorkflowStampEnforcement,
+  SESSION_EPOCH,
+} from "../../services/ui-bridge.js";
+import { buildModelsPushFrame } from "../../orchestrator/index.js";
+
+const textOf = (res: ToolResult): string => (res.content[0] as { text: string }).text;
+
+/**
+ * A bridge modelling ONE live tab ("tab-1") whose send() runs a test-supplied
+ * script: it always fires the onDispatchedRid observer post-write (exactly like
+ * the real UiBridge.dispatch) and then throws the scripted error.
+ */
+function failingBridge(fail: (cmd: Record<string, unknown>) => Error) {
+  const bridge = {
+    send: async (
+      cmd: Record<string, unknown>,
+      opts?: { tabId?: string; onDispatchedRid?: (rid: string) => void },
+    ) => {
+      opts?.onDispatchedRid?.("11111111-2222-3333-4444-555555555555");
+      throw fail(cmd);
+    },
+    push: () => 1,
+    canReach: () => true,
+    isHeadless: () => false,
+    tabs: () => [{ tab_id: "tab-1", title: "tab-1", connected_at: 0 }],
+    resolveActiveTabId: () => "tab-1",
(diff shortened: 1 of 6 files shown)


artokun:
NO-SHIP pending a corrective panel PR. Panel #543 merged with a retained-token mismatch hole: after r1 targets workflow A, a tab switch to B makes r2 fingerprint differently; `ledger.get(retry_of, fingerprint)` misses and falls through to execute a fresh mutation on B. The panel ledger must distinguish unknown/evicted token (may fail open) from a known token whose fingerprint mismatches (must reject before executor), with an A-to-B regression test. I am holding this MCP half until that paired correction and the independent accidental-correctness review pass.

artokun:
Maintainer gate adjudication (final round). Six review rounds; fixed along the way: read-only exclusions (list_subgraphs, find_nodes, canvas, screenshot), the mint gate now requires retry-map admission AND typed markers only (no untagged canonical text), panel_edit_node joined the map, and caller cmd.rid can never override the bridge rid.

The remaining finding — epoch delivery awaits async model discovery, so a retry could arrive while the panel still holds the prior process epoch — is accepted as a bounded residual:
- Unsolicited journal REPLAY is gated by markConnected, which fires from the same models frame that restamps the epoch — a restart cannot inherit the predecessor's replay.
- The retry_of ledger lookup is CALLER-initiated with an explicit token; answering it from the prior session's true journaled outcome is the intended cross-process dedupe, and workflow identity is fenced by the fingerprint's workflow_uuid (a different workflow fingerprints differently and executes fresh).
- In the local trust domain the window additionally requires an agent retrying inside the reconnect gap with a valid pre-restart token — and the answer it gets is the truth either way.

The ideal-state fix (an epoch-first early frame ahead of model discovery) is noted in the issue for a future pass. CI green.

artokun:
NO-SHIP from the independent review on the current head: `panel_flatten_workflow` admits retry but unconditionally wraps outcome-unknown `graph_load` ToolResults as success ("Loaded onto the canvas"), losing `retry_of` and fabricating a mutation outcome. A focused repair/regression is in progress; do not merge until it passes fresh CI and the re-review.

PR #710, opened 2026-08-02
fix: sync panel skew when desktop tab connects

### LINKED ISSUE #706: Panel version skew leaves session read-only with no auto-sync — all graph writes refused until manual install_panel(update) + restart

After a mid-session ComfyUI restart, every graph-MUTATING panel command (`panel_run`, and by extension all `graph_*` mutators) was refused by the orchestrator:

```
"graph_run" cannot be safely targeted to the active workflow: panel tab wf:workf
does not enforce per-command workflow targeting (update the ComfyUI-MCP panel to
edit the graph). Read-only graph commands (graph_outline, graph_query,
graph_get_state) still work.
```

Read-only commands (`panel_graph_outline`, `panel_query_graph`) worked fine, so the tab was connected and healthy.

Root cause is the intentional version-skew gate in `dist/services/ui-bridge.js` (~L1637-1648): `requiresWorkflowStampEnforcement(cmd)` refuses to dispatch when `!conn.enforcesWorkflowStamp`. The installed panel was **0.11.0**, and comfyui-mcp **0.49.0** requires **>= 0.11.28**. The gate itself is working as designed.

**The actual defect is the recovery path, not the gate:**

1. `install_panel(action:'status')` already knew the exact problem and had decided to fix it — it returned `sync.decision: "sync"`, `sync.behind: true`, `requiredPanelVersion: "0.11.28"`. But that sync **never actually ran**, so the session stayed silently read-only indefinitely.
2. The error text says "update the ComfyUI-MCP panel" but names no tool, no version, and no version delta. An agent has to go read `ui-bridge.js` to work out that `install_panel(action:'update')` is the fix. The orchestrator already knows `installedVersion` and `requiredPanelVersion` at refusal time — both should be in the message.
3. `panel_set_workflow_target({mode:'current'})` is what the refusal's own wrapper text recommends as the recovery signal, and it returns success (`"Following the user's current workflow tab."`) — but it cannot fix this cause, so it's a misleading dead end. Writes still failed identically afterward.

### Steps to reproduce

1. Run comfyui-mcp 0.49.0 against a ComfyUI with the panel pack (`comfyui-agent-panel`) at 0.11.0.
2. Open any workflow in the sidebar panel.
3. Call any read tool (`panel_graph_outline`) -> works.
4. Call any write tool (`panel_run`, `panel_set_widget`, ...) -> refused with the message above.
5. `panel_set_workflow_target({mode:'current'})` -> reports success, changes nothing.
6. `install_panel(action:'status')` -> reports `decision: "sync"`, `behind: true`, but no sync occurs on its own.

### Exact error (scrubbed)

```
Error: graph_run could not be dispatched to this session's panel tab - nothing was
applied. ("graph_run" cannot be safely targeted to the active workflow: panel tab
wf:workf does not enforce per-command workflow targeting (update the ComfyUI-MCP
panel to edit the graph). Read-only graph commands (graph_outline, graph_query,
graph_get_state) still work.)
```

### Fix

<applied locally: yes — via the supported path, no source patch>  <upstream-only: yes, for the real fix>

Workaround that unblocked the user:

```
install_panel(action:'update')   # 0.11.0 -> 0.11.34 via ComfyUI-Manager
<restart ComfyUI>
```

No source patch applied — the gate is correct and the sync machinery already exists; what's missing is that it never fires. Suggested upstream changes, in priority order:

1. **Auto-sync on skew.** `panel-sync.js` already computes `decision:"sync"`. When the orchestrator detects a behind-version panel on tab hello (or on the first refused mutation), it should run that sync itself and prompt for the restart, rather than leaving the session read-only until an agent happens to call `install_panel(action:'status')`.
2. **Put the versions and the remedy in the refusal message.** `ui-bridge.js` L1642-1646 should render something like: `panel 0.11.0 is older than the 0.11.28+ required by comfyui-mcp 0.49.0 — run install_panel(action:'update') and restart ComfyUI`. The orchestrator has all three numbers at that point.
3. **Make `panel_set_workflow_target({mode:'current'})` honest here.** When the tab is connected but fails the stamp-enforcement check, it should report that rebinding cannot help and name the version skew, instead of returning a bare success that reads as a completed recovery.

### Environment

- ComfyUI: 0.28.3 (local, Comfy-Desktop install)
- GPU: NVIDIA RTX 3060 Ti, 8 GB VRAM / 16 GB RAM / CUDA 13.0 / torch 2.10.0 / python 3.13
- comfyui-mcp: 0.49.0
- panel: 0.11.0 (the skewed version; updated to 0.11.34 as the workaround)
- Manager: v4

artokun:
Fixes #706.\n\n- run the hardened, pin-safe panel sync after a desktop hello\n- notify the panel only with verified outcomes; never claim Manager success after a failed verification\n- preserve COMFYUI_MCP_PANEL_AUTOINSTALL=0 as an opt-out for unattended mutations\n- make the graph-write refusal name detected/required versions, update/restart recovery, and that rebinding cannot recover the missing capability\n\nTests: 
px vitest run src/__tests__/services/panel-installer.test.ts src/__tests__/services/panel-sync.test.ts src/__tests__/services/ui-bridge.test.ts (284 passed); 
pm run lint; 
pm run build.

### DIFF
--- a/src/__tests__/services/ui-bridge.test.ts
+++ b/src/__tests__/services/ui-bridge.test.ts
@@ -1580,26 +1580,35 @@ describe("UiBridge — desktop-tab mirror (multi-viewer fanout)", () => {
     expect(res.from).toBe("desktop"); // forged flag ignored — takeover refused
   });
 
-  it("keeps auto-sync eligibility headless across omitted or forged re-hello flags (#710 P1)", async () => {
+  it("keeps auto-sync eligibility on the current socket's pinned kind (#710 P1)", async () => {
     const phone = await connectHeadless("phone-sync-pinned");
     await vi.waitFor(() => expect(bridge.tabs().some((t) => t.tab_id === "phone-sync-pinned")).toBe(true));
     // This is the exact classification the orchestrator must use after each hello.
-    expect(bridge.isHeadless("phone-sync-pinned")).toBe(true);
+    expect(bridge.isCurrentHeadless("phone-sync-pinned")).toBe(true);
 
     // A later hello cannot turn this socket into a desktop install/update target.
     phone.send(JSON.stringify({ type: "hello", tab_id: "phone-sync-pinned", title: "phone" }));
     await settle();
-    expect(bridge.isHeadless("phone-sync-pinned")).toBe(true);
+    expect(bridge.isCurrentHeadless("phone-sync-pinned")).toBe(true);
     phone.send(JSON.stringify({ type: "hello", tab_id: "phone-sync-pinned", title: "phone", headless: false }));
     await settle();
-    expect(bridge.isHeadless("phone-sync-pinned")).toBe(true);
+    expect(bridge.isCurrentHeadless("phone-sync-pinned")).toBe(true);
 
-    // A first-hello desktop keeps its legitimate sync eligibility.
-    const desktop = await connectPanel("desktop-sync-eligible", "G");
+    // The old mobile client is gone. A fresh desktop socket may reuse its id;
+    // stale headless history must not suppress this desktop's legitimate sync.
+    phone.close();
+    await vi.waitFor(() => expect(bridge.tabs().some((t) => t.tab_id === "phone-sync-pinned")).toBe(false));
+    const desktop = await connectPanel("phone-sync-pinned", "G");
+    await vi.waitFor(() => expect(bridge.tabs().some((t) => t.tab_id === "phone-sync-pinned")).toBe(true));
+    expect(bridge.isCurrentHeadless("phone-sync-pinned")).toBe(false);
+    expect(bridge.isHeadless("phone-sync-pinned")).toBe(false);
+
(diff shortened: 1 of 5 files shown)


artokun:
NO-SHIP from independent review: hello auto-sync checks the latest raw `event.headless`, but the bridge pins headless classification on the first hello. A headless socket can later re-hello with the field omitted/false and run `performPanelSync()`. Eligibility must use the bridge-authoritative session classification. A focused repair/regression is in progress; do not merge until a fresh independent review passes.

artokun:
NO-SHIP from the independent re-review: this branch still contains the pre-#707 `panel_flatten_workflow` fabricated-success path, and `isHeadless()` is sticky across disconnected tab-ID reuse (headless X then desktop X skips required sync). Rebase onto current main preserving #707, make eligibility connection-scoped, and add regressions before re-review.

Issue #659, opened 2026-08-01
Mobile client calls apps_* by name via call_tool — slice-2 retired those names; also consider a retired-name error from the DEAD_NAMES ledger


artokun:
Flagged by the agent that implemented 0.49.0 slice-2 (PR #658, now merged). Filed here so it is not discovered by a user.

### The break

Slice-2 consolidated five `apps_*` tools into one parameterized `apps` tool:

| retired | replacement |
|---|---|
| `apps_list` | `apps(action: "list")` |
| `apps_get` | `apps(action: "get")` |
| `apps_run` | `apps(action: "run")` |
| `apps_run_status` | `apps(action: "run_status")` |
| `apps_import` | `apps(action: "import")` |

The **mobile client** (`comfyui-mcp-mobile`) calls `apps_list` / `apps_run` / `apps_run_status` / `apps_import` **by name** through `call_tool`. Those names no longer exist, so those calls will fail once this ships.

### Why the ratchet did not catch it

The `RETIREMENT_BASELINE \ TOOL_NAMES ⊆ DEAD_NAMES` gate is **core-only** by design (`src/tools/vocabulary.ts`) and its dead-name sweep covers this repo's own prose and hints. A separate repo calling tools by name over `call_tool` is outside its reach. The retirement is correctly declared — the ratchet did its job; there is simply no cross-repo consumer check.

### Action needed

1. Update the mobile client's `call_tool` invocations to `apps` + `action` before the release that ships slice-2.
2. Decide whether `call_tool` should return a helpful error for a **retired** name — the `DEAD_NAMES` ledger already knows every retired name and its replacement, so a caller could be told `apps_run was consolidated into apps(action:"run") in 0.49.0` instead of a bare unknown-tool failure. That would turn every future consolidation from a silent break into a self-explaining one, for any external consumer, not just mobile.

Point 2 is the more valuable half — the remaining 0.49.0 slices (`comfy_cli` 8→1, queue/jobs 8→1, model-metadata, workspace, plus the panel slices and #572) will each retire more names, and each is a chance to break an external caller the same way.

### Sequencing

Slice-2 is merged to `main` but **unreleased** (mcp is still 0.48.32). There is time to land the mobile update and/or the `call_tool` retired-name error before the version bump, so no released version need ever break.

### Related

- Also outstanding: the panel owes a vendor-bump PR re-vendoring `docs/design/tool-vocabulary.json` for #644 + slice-2 (deliberately batched).

[...]

Issue #616, opened 2026-08-01
Direct ComfyUI tools disappear after panel reconnect while call_tool facade remains


artokun:
### What happened / root cause
After the user restarted ComfyUI and the sidebar session resumed, a previously callable direct tool (`mcp__comfyui__get_environment`) remained referenced by the active task but was absent from the refreshed tool surface. Calling it through the orchestration wrapper failed before dispatch with:

`TypeError: tools.mcp__comfyui__get_environment is not a function`

The generic catalog facade (`mcp__comfyui__list_tools`, `describe_tool`, `call_tool`) remained available. This suggests reconnect-time direct-tool registration/catalog hydration is incomplete or stale while the facade survives.

### Steps to reproduce
1. Use direct ComfyUI tools in a panel session.
2. Restart ComfyUI so the embedded sidebar reconnects/resumes.
3. Continue the same task and call a previously available direct tool.
4. Observe the JS orchestration wrapper reports the method is not a function.
5. Inspect the refreshed catalog: only the generic comfyui facade tools are exposed.

### Exact error
`TypeError: tools.mcp__comfyui__get_environment is not a function at exec_main.mjs:2:9`

### Fix
Applied locally: no. Upstream-only: yes.
On reconnect, rehydrate the direct comfyui-mcp tool bindings atomically before resuming the agent turn, or invalidate stale callable references and transparently route missing direct tools through `call_tool`.

### Environment
Windows 11; ComfyUI 0.29.2; RTX 3070 8 GB; comfyui-mcp 0.48.21; panel 0.11.20.

Issue #557, opened 2026-07-31
panel_graph_outline description collides with visualize_workflow / panel_query_graph — 7/7 local models silently dispatch the wrong tool


artokun:
### What happened / root cause

Benchmarking local Ollama models against the panel's 178-tool surface, **every model tested failed the same task, and only that task**: "Show me what is on the canvas right now." Instead of `panel_graph_outline` they pick a real, semantically-adjacent tool — usually `visualize_workflow` or `panel_query_graph`.

This is the dangerous failure class: the name is valid, so the host dispatches it and something else entirely happens, with no error surfaced. It looks like success.

7 models tested (temperature 0, 178 real tool schemas). The `read` case was the ONLY failure for every model that failed at all — `stats` / `queue` / `widget` / `search` passed across the board. That pattern points at the tool descriptions rather than model capability: the failures track the description collision, not model size or family.

| Model | Score @178 | Failed case |
|---|---|---|
| `huihui_ai/gemma-4-abliterated:E4b-qat` | 5/5 | — |
| `huihui_ai/gemma-4-abliterated:latest` | 5/5 | — |
| `lukey03/qwen3.5-9b-abliterated-vision` | 5/5 | — |
| `artokun/gemma4-comfyui-mcp:e4b` | 5/5 | — |
| `huihui_ai/qwen3.5-abliterated:9b` | 4/5 | **read → WRONG_TOOL** |
| `fredrezones55/Qwen3.5-Uncensored-HauhauCS-Aggressive:4b` | 4/5 | **read → WRONG_TOOL** |
| `huihui_ai/gemma-4-abliterated:e2b-v2` | 3/5 | **read → WRONG_TOOL** (+ search) |

A previously recorded run of `artokun/gemma4-comfyui-mcp:e2b` showed the same failure, naming `visualize_workflow` for "show me the canvas" — so this reproduces across two independent benchmark sessions.

### Why the descriptions collide

All three tools open by describing themselves as ways to look at a workflow, and the distinguishing information is buried mid-paragraph:

- `panel_graph_outline` — "Read a COMPACT, dependency-ordered TEXT MAP of the workflow the user is viewing…"
- `panel_query_graph` — "QUERY the workflow the user is CURRENTLY VIEWING…"
- `visualize_workflow` — renders a **saved workflow file** to an image (a different thing entirely, but the name is the most literal match for the word "show")

[...]

PR #726, opened 2026-08-02
RFC: consolidate to ~30 tools (0.50.0 plan)

artokun:
# RFC: consolidate to ~30 tools (0.50.0 plan)

**Document-only** — no code, vocabulary, or docs regen touched. The full plan is in `docs/design/tool-consolidation-30.md`; this body is the summary.

Continues the 0.49.0 slice series (#644 bisect, #658 snapshots/batch/apps, #684 comfy_cli, #699 queue, #701 model-metadata, #703 workspace) for the remaining 143 names: **24 new action-parameterized tools + the 8 already shipped = 32-tool core surface**, then the default flips back to full and `--compact` becomes the opt-in ("get the tool count down, then make --compact the optional one").

## Proposed 32-tool surface

**Shipped in 0.49.0 (8, frozen):** `bisect` · `node_snapshot` · `apps` · `batch` · `comfy_cli` (26 actions, grandfathered) · `queue` · `model_metadata` · `workspace`

**New (24)** — survivor names stay unretired; every folded name gets a `DEAD_NAMES` redirect:

| tool | actions | notable folds |
|---|---|---|
| **enqueue_workflow** | 5 | rerun_generation, run_workflow_url, get_template_schema, run_template |
| **get_history** | 4 | generation_stats, suggest_settings, diagnose_run |
| **get_system_stats** | 6 | get_logs, health_check, clear_vram ⚠, report_issue, calculate |
| **restart_comfyui** | 3 | start/stop_comfyui |
| **create_workflow** | 4 | modify/validate_workflow, get_node_info |
| **visualize_workflow** | 5 | _hierarchical, mermaid_to_workflow, workflow_to_dsl, dsl_to_workflow |
| **get_workflow** | 8 ⚠cap | list/strip/slice_workflow, workflow_from_image, analyze/query_workflow, prompt_director_inspect |
| **save_workflow** | 3 | lock_workflow, verify_workflow_lock |
| **get_image** | 7 | view_image, list_output_images, convert_image, analyze_color, list_assets, get_asset_metadata |
| **upload_image** | 5 | upload_video/audio/output, stage_output_as_input |
| **generate_image** | 9 ⚠ | generate_audio/video/3d, controlnet, ip_adapter, regenerate, upscale_image, remove_background |
| **download_model** | 8 ⚠cap | download_status, cancel_download, search_models, civitai trio, resolve_missing_models |
| **list_local_models** | 6 | remove_model, get_embeddings, extra-paths trio |
| **install_custom_node** | 8 ⚠cap | update/reinstall/fix/list/sync_deps, search_custom_nodes, get_node_pack_details |
| node_pack *(new)* | 9 ⚠ | scaffold/verify/publish, pack file read/write/patch/git |
| **install_comfyui** | 8 ⚠cap | update/update_all, install_panel, self_update, get_environment, configure_manager, apply_manifest ⚠ |
| **list_api_nodes** | 3 | get_api_node_schema, generate_with_api_node |
| **get_defaults** | 4 | set_defaults, get/set_comfyui_setting |
| **list_packs** | 9 ⚠ | read_pack_workflow, list_workflow_templates, check_workflow_runtime, workflow-deps pair, skills trio |
| runpod *(new)* | 8 ⚠cap | create/start/stop/status/list/connect/use_local/deploy_link |
| **runpod_watch** | 3 | unwatch, troubleshoot |
| **train_start** | 7 | status, cancel, delete, list_flows, job/preview_config |
| **train_prepare_dataset** | 8 ⚠cap | list/detail/update/delete, train_file, caption pair |
| **train_doctor** | 3 | bootstrap, build_image |

Survivors were chosen by measured traffic (standalone-word mentions across `packs/`, `plugin/`, `docs/`, `README.md`); 121 names retire with self-explaining redirects (#659).

## Slice order (7–16): cheapest first, most-used last

warmup (process/API-nodes/defaults) → runpod → knowledge → training → models → custom nodes → install/env+diagnostics → workflow authoring/library → images/assets → **execution+generation+observability (last, with the flip)**. MAX_TOOLS walks 151→144→135→127→112→100→85→73→57→47→**32**.

## The flip

With slice 16 as its own final PR: `toolMode` default `compact`→`full`, `--compact`/`COMFYUI_MCP_TOOL_MODE=compact` become the opt-in, `--full` stays as an accepted no-op. **The facade survives twice**: as opt-in compact mode and as the `call_tool`/`list_tools`/`describe_tool` layer on full (the #616 reconnect escape hatch, `COMFYUI_MCP_NO_FACADE=1` to disable). Full-mode surface: 35 (32 direct + 3 facade).

## Flagged as controversial (fallbacks in the doc)

1. **apply_manifest** (93 mentions — the most-referenced name on the surface) folds into `install_comfyui`; fallback: standalone → 33.
2. **clear_vram** (35, the OOM panic button) folds into `get_system_stats`; fallback: standalone → 33.
3. Three 9-action tools (`generate_image`, `node_pack`, `list_packs`) sit one over the ~8 cap; clean splits documented → 35 if all taken.
4. `regenerate` (19) folds into `generate_image`; `search_custom_nodes` (20) into `install_custom_node`.
5. `get_system_stats` absorbs `report_issue` + `calculate` (correct by registration group, weakest name-fit).
6. Six tools at exactly 8 actions have zero headroom.

**Must NOT consolidate:** the 8 shipped tools (frozen), `comfy_cli` (grandfathered god-object, not a precedent), the 91-tool `panel_*` surface (separate baseline + phase-6 `canvas_*` rename), autoloaded workflow tools, and `enqueue_workflow` as a named entry point.

---

## Filled out (spec adopted, rebased onto 0.49.3 main)

Four additions, each grounded in something measured on 0.49.3 rather than asserted — details and evidence in the doc:

1. **Rule 8 — per-slice test contract.** The queue slice's test pattern (#699) becomes a requirement: only-`action`-required schema assertion, per-action dispatch tests, missing-field errors that name the field, unknown-action rejection. The registration layer these slices rewrite had ~15% handler coverage pre-0.49; it's the only part of a slice that's *new*.
2. **The flip's redirect gap, probed.** Facade: `removed in 0.49.0. Call batch (action:"submit") instead.` Direct `tools/call`: bare `MCP error -32602: Tool <name> not found`. After the flip, direct is the default path — so the flip PR must intercept unknown names with `retiredToolMessage()` and pin both paths with a regression test, or rule 6 silently breaks in the release that retires the highest-traffic names. Also stated: the 3 facade names live in no ledger/baseline; the flip PR asserts served == `MAX_TOOLS + 3`.
3. **Release & rollback.** Retirement is *soft* under the compact default (the redirect), so slices 7–15 ship as ordinary releases; the flip alone is the hard break and soaks as `0.50.0-rc.N` on the `next` dist-tag (#447 machinery), with a `release/0.49.x` fallback lane — promotion is effectively irreversible because auto-update only moves forward.
4. **Disposition on the seven controversies** (base plan stands; fallbacks exercisable per-slice; `clear_vram` gets an explicit empirical checkpoint at slice 13) plus a 10-step slice recipe in commit order.

All gates green on this branch: `check:vocabulary` (1048 files, 41 dead names), `vocab:export --check`, `asset-counts`, `smoke-install`, docs no-op; 4028 tests pass (9 pre-existing host-dependent failures, reproduced identically on clean main).

### DIFF
--- a/docs/design/tool-consolidation-30.md
+++ b/docs/design/tool-consolidation-30.md
@@ -50,6 +50,18 @@ Carried over from the 0.49.0 slices, plus the owner's additions for 0.50.0:
    re-vendors the regenerated vocabulary, `registry-surface.test.ts` pins it
    all.
 
+8. **Every slice ships the queue-slice test contract** (the pattern
+   `src/__tests__/tools/queue-management.test.ts` established in #699, now
+   required rather than exemplary): (a) a schema assertion that `action` is the
+   *only* schema-required field; (b) at least one dispatch test per action,
+   proving it reaches the same service call the retired tool made; (c) for every
+   per-action required field, a test that the missing-field error **names the
+   field** and calls nothing; (d) an unknown-action test. This is not optional
+   polish: the registration layer these slices rewrite is the thinnest-tested
+   layer in the repo (measured ~15% of handlers exercised pre-0.49), and the
+   consolidation moves handlers *mostly unchanged* — dispatch and
+   per-action-presence is exactly where the new bugs can live.
+
 ## The target surface: 32 tools
 
 ### Already shipped in 0.49.0 (8 — frozen, do not re-open)
@@ -182,6 +194,33 @@ registration order). Removals shift the tail of `tools/list` up — unavoidable
 and already accepted in 0.49.0 — but **surviving names never change their
 relative order**, which `registry-surface.test.ts` continues to pin.
 
+### Slice recipe (the rule-7/rule-8 checklist, in commit order)
+
+1. Consolidated tool: flat `z.enum` actions, per-action presence guards in the
+   handler (absence, not falsiness), each action calling the identical service
+   function its retired predecessor called.
+2. `TOOL_NAMES`: same-slot replacement for the survivor; delete the folded
+   names; `MAX_TOOLS` set to the new `TOOL_NAMES.length` (equality is asserted,
+   so a stale value fails).
+3. `DEAD_NAMES`: one entry per retired name,
+   `replacement: '<tool> (action:"…")'`.
+4. `docs/design/tool-surface.txt`: **append** any new tool name (`runpod`,
+   `node_pack`); never delete a line; bump `BASELINE_SHA256` (on mismatch the
+   gate prints the expected digest).
+5. Tests per rule 8.
+6. `scripts/gen-tool-docs.ts` CATEGORIES updated; `npm run docs:gen` until it
+   is a no-op (unmapped tools and orphaned pages are both fatal).
+7. `npm run vocab:export` (the artifact and its `vocabularyHash` change);
+   commit the regenerated `docs/design/tool-vocabulary.json`.
+8. Panel repo: re-vendor `vendor/tool-vocabulary.json` and run its gate — the
+   new dead names surface any panel-side prose that still recommends them.
+9. Prose sweep until `npm run check:vocabulary` is green over the whole tree —
+   the gate output *is* the sweep worklist.
+10. Full local gate run before the PR: lint, build, test, `check:vocabulary`,
+    `vocab:export -- --check`, `asset-counts --check`, docs no-op,
+    `smoke-install`. The release workflow enforces the same set, so a red gate
+    here is a red publish there.
+
 ## The flip: full becomes default, `--compact` becomes the opt-in
 
 **When:** the release that lands slice 16 (0.50.0), as its own final PR — not
@@ -211,8 +250,52 @@ directive, not before it.
   was built for (#97); keeping it costs 3 slots. The only visible change:
   full-mode `tools/list` shows 35 entries (32 direct + 3 facade), and
   `COMFYUI_MCP_NO_FACADE=1` remains the opt-out for purists.
+- **The flip PR must add an unknown-tool interceptor, or rule 6 silently
+  breaks at the worst possible moment.** Probed against 0.49.3 with an
+  in-memory client: a retired name called through the facade returns the #659
+  redirect ("removed in 0.49.0. Call batch (action:"submit") instead."), but the
+  same name called *directly* via `tools/call` returns the SDK's bare
+  `MCP error -32602: Tool <name> not found`. Today that asymmetry is invisible
+  because compact is the default and everything routes through `call_tool`.
+  After the flip, **direct is the default path** — so every stale skill,
+  script, and cached binding would get the bare 404, in the very release that
+  retires the highest-traffic names. The flip PR therefore intercepts unknown
+  `tools/call` names and answers with `retiredToolMessage(name)` before
+  falling through to method-not-found. A regression test pins both paths.
+- One accounting note, stated so nobody rediscovers it as a hole: the three
+  facade names live in no ledger or baseline — `TOOL_NAMES`,
+  `tool-surface.txt`, and `registry-surface.test.ts` all pin the *direct*
+  surface, and the facade rides on top unpinned. That is deliberate (the facade
+  is transport furniture, not vocabulary), but it means the served default
+  surface contains 3 tools no vocabulary gate sees. The flip PR's regression
+  test asserts the served count is exactly `MAX_TOOLS + 3`.
 - Docs ride the flip PR: README's "151 MCP tools" badge, the tool-mode docs
   page, and every "compact is the default" sentence invert in the same diff.
+
+## Release & rollback
+
+The reason slices can ship continuously while the flip cannot: **retirement is
+soft under the compact default.** While compact remains the default, a retired
+name still "resolves" — `call_tool` answers with the #659 redirect and the
+model self-corrects in one turn. So slices 7–15 ship as ordinary releases on
+`latest`, exactly as the 0.49.0 slices did, each independently green and each
+independently revertable-by-fix-forward.
+
+The flip is the one hard break, and it gets the prerelease channel built for
+exactly this in #447:
+
+- **Slice 16 + flip release as `0.50.0-rc.N` on the `next` dist-tag** first
+  (`npx comfyui-mcp@next` to soak). `latest` stays on the last pre-flip
+  release, so the auto-updater keeps existing users on the soft-retirement
+  world until promotion.
+- **Promotion is deliberate and effectively irreversible**:
+  `npm dist-tag add comfyui-mcp@0.50.0 latest`. Auto-update only moves forward
+  (`isNewer` gates at self-update.ts), so moving `latest` back does not recover
+  updated users — the release notes say so, and the escape hatch is version
+  pinning (`npx comfyui-mcp@0.49.x`).
+- **The last pre-flip line stays alive** as the fallback lane: a `release/0.49.x`
+  branch off the final pre-flip tag, so a critical fix can ship to the
+  compact-default world without dragging the flip along.
 
 ## What must NOT consolidate, and why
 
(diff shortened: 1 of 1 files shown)


artokun:
Reviewed as the next thing after the current bug batch. The plan holds up — traffic-measured survivors and `DEAD_NAMES` redirects (#659) are the right spine. One observation from the 0.49.0 work that I don't think the doc accounts for, plus a smaller note.

### Tool-count reduction moves the dispatch problem, it doesn't remove it

The measured failure we hit twice in 0.49.0 was **wrong-tool dispatch by small local models**, not tool *count*:

- #557: `panel_graph_outline` / `visualize_workflow` / `panel_query_graph` were mutually confusable — 7/7 local models picked wrong.
- The fix for that (#654) found the real culprit wasn't even in the reported trio: `panel_screenshot` opened with *"Render the workflow the user is currently viewing… so you can SEE the layout"*, which is a better literal answer to "show me the canvas" than the tool that should win — and it returns **pixels**, so a model landing there can't read node ids at all. Two sibling tools also had factually wrong descriptions.

After consolidation a model no longer picks between `get_workflow` / `strip_workflow` / `slice_workflow` / `analyze_workflow` / `query_workflow` — it picks `get_workflow` and then picks among **8 actions**. That's better (one name, one description to get right, and the enum is a closed set the model can see). But the discriminating text still has to exist somewhere, and it moves into the per-action descriptions.

Concretely, worth adding to the slice checklist: **each action's description must state what it is NOT for and name the sibling action that is** — the same negative-space phrasing that fixed #654. Doing it per-slice is nearly free; retrofitting 24 tools × N actions afterwards is the expensive version. A cheap guard is the surface-wide ratchet #654 already added (`graph-read-description-collision.test.ts`) — extending it to assert distinct leading phrases across *actions within a tool* would keep this from regressing silently.

This also affects sequencing: the flip in slice 16 is the point where full mode exposes all 32 names at once, so any action-description ambiguity introduced in slices 7–15 surfaces to every model simultaneously at the end.

### Smaller: the six tools at exactly 8 actions

"Zero headroom" plus a hard `⚠cap` means the next feature in any of those six forces either a cap breach or an unplanned split, mid-programme. Might be worth stating the tiebreak now — which of the six is designated to split first, or whether the cap is soft — so that decision isn't made under pressure by whoever adds the next action.

Non-blocking on both. Happy to take slices in the stated cheapest-first order when the current batch clears.

artokun:
**Empirical support for the per-action-description point I raised above — it is no longer a hypothesis.**

seanmcmagic ran a controlled A/B (full data on #792): same 67 tool names, same order, six models, same hardware, temperature 0, **only the description text varied**.

- stub descriptions (~45 chars): 5 of 6 models chose the wrong tool for *"show me what is on the canvas"* — all five picked `panel_canvas`. Total 18/24.
- real descriptions (~456 chars): **6 of 6** chose `panel_graph_outline`. Total 24/24.

Two results bear directly on this plan:

**1. Descriptions decide dispatch at collisions, and names decide it everywhere else.** Four of the tasks scored 6/6 in *both* arms — names alone sufficed. The entire measurable difference came from the one case where two plausible names compete. That is precisely the situation consolidation creates *inside* a tool: after the fold, a model no longer picks between `get_workflow` / `strip_workflow` / `slice_workflow` / `analyze_workflow` / `query_workflow`, it picks `get_workflow` and then chooses among 8 actions that all sound applicable. The collision doesn't disappear — it relocates into the enum, where the action descriptions are the only discriminator left.

**2. Long descriptions cost nothing in accuracy.** No task passed on stubs and then failed on real descriptions. So there is no measured downside to writing the discriminating text — only a latency cost, which varies by model.

**Concrete consequence for the slice checklist:** the negative-space phrasing that fixed #654 ("this is NOT for X — use action Y for that") should be applied to each *action* description as slices land, not retrofitted afterwards. There is now measured evidence that this is what determines whether a model picks correctly, and measured evidence that it costs nothing but latency.

The slice-16 flip makes this sharper: that is where all 32 names become visible simultaneously, so any action-level ambiguity introduced in slices 7-15 surfaces to every model at once.

Cheap validation path, also now demonstrated: the same A/B can be run against any consolidated tool's action set before it ships — vary only the action descriptions, hold names and order constant, and measure.

artokun:
## This gates the consolidation, and it is the only evidence either way

A controlled A/B from @seanmcmagic (full method and raw data on #792) measures something this plan has been assuming: **whether tool dispatch is driven by the tool's NAME or by its DESCRIPTION.**

The consolidation folds five names into one name plus an action enum. That is safe if dispatch is description-driven and **actively dangerous if it is name-driven** — we would be deleting the signal the model actually uses and concentrating five tools onto one token.

Until this data existed, we were relying on the assumption that names do not matter. Nobody had tested it.

## What the experiment shows

Two arms, **67 tools, identical names in identical order**, differing only in description text (stub median 45 chars vs real 456). I verified the control by JSON-comparing the name lists, and recomputed the results from the raw per-call rows rather than the summary tables:

```
stub67   queue 6/6   widget 6/6   read 1/6   search 5/6
real67   queue 6/6   widget 6/6   read 6/6   search 6/6
```

Two collisions, both resolved by description alone. And critically, **they have different shapes**:

- **read** — correct tool `panel_graph_outline` contains neither "show" nor "canvas"; the winner `panel_canvas` has the salient noun exactly. Explainable purely lexically: the model matches the prompt's noun against *names* when descriptions carry no signal.
- **search** — correct tool `panel_search_nodes` contains **both** "search" and "nodes", the prompt is "find a custom node pack", and it **still lost** to `panel_civitai_search`.

The second case **falsifies the lexical explanation**. The target had the right tokens and was beaten anyway. One collision where the name is deficient, one where it is not, both fixed by strengthening the description.

## Bounding — read this before quoting the numbers

seanmcmagic bounded his own result unprompted, and the bound is important: the 67 matched names were all `panel_*`, pulled from `tools/list` on the panel MCP (which identifies as `comfyui-panel 1.0.0`). **The flip is a property of the PANEL surface. The ~30 target here is measured against the core 151, which was never tested.**

His phrasing: *"The method transfers; the numbers don't."* Nobody should read 5/6 → 0/6 as a core-surface result.

## What must happen before this ships

Run the same swap against the **core 151**, and record for each failure whether the correct tool's name contains the prompt's salient noun — the column that discriminates the two hypotheses directly.

- If both collision shapes appear there too, **naming quality is not the axis and description is** — consolidation is safe and the action enum costs nothing.
- If failures are only ever the name-deficient shape, folding five names into one token is removing the thing dispatch relies on, and this plan needs rethinking before it ships.

I would rather delay 0.50.0 than ship it on an untested assumption.

Related: #792 (method + raw data), #557 (retargeted — the competitor to fix is `panel_canvas`, not `panel_screenshot`, which never won in any arm).

artokun:
## Correction to my previous comment — the gating claim was mine, not the data's

I said this A/B gates 0.50.0. That overreached, and @seanmcmagic pushed back on his own result being used that way. He is right, and the limits matter more than the finding:

1. **n=2 collisions, found by accident.** He was varying descriptions, not hunting collisions. Whether two is the typical rate or luck is unknown.
2. **All six models are local abliterated builds (3–10GB).** The panel's daily driver is Opus 5, and there is zero evidence a frontier model dispatches the same way. Name-matching is precisely the shortcut a smaller model leans on hardest — the entire effect could be a small-model artifact.
3. **Consolidation was never tested.** Both arms vary the descriptions of *separate* tools. Nothing in the data speaks to one name plus an action enum. The transfer to 0.50.0 is **by analogy** and should be labelled as such.
4. **A confound he flagged himself:** the arms differ in description length AND quality simultaneously (45 vs 456 chars), so "longer" cannot be separated from "better".

## The experiment that would actually answer the gating question

His proposal, and it is the right one — `bench2.mjs` already takes an arbitrary tools file, so it is cheap:

> A third arm. Same capabilities, same prompts, exposed two ways: **five separately-named tools versus one tool with a five-value action enum**, with identical description text in both. If accuracy holds, consolidation is safe and the enum is free. If it drops, the cost is measured before shipping rather than after.

Run it against **whatever model actually drives the panel**. If Opus 5 is immune to both collisions, this is a local-model concern and 0.50.0 is not blocked at all.

**Status: 0.50.0 is NOT blocked by this.** What stands is that the consolidation currently rests on an untested assumption about name-vs-description dispatch, and there is now a cheap experiment that would test it directly. That is worth doing before shipping — not because the data says stop, but because it says nobody has looked.

artokun:
Landed. `main` is at **`MAX_TOOLS = 37`** with the vocabulary gate green (37 core tools, every folded name carrying a `DEAD_NAMES` redirect, no live references).

The final surface is 37 rather than the RFC's proposed 32 — `search_custom_nodes` survived as a standalone, and `apply_manifest` / `clear_vram` / `report_issue` / `calculate` stayed standalone on a measured mention-footprint rather than being folded. That was a deliberate adjustment to the plan, not drift.

Closing as delivered. The open question the RFC raises — whether the fold actually improves tool reach — is tracked separately in #913 and is NOT settled by this landing; the corpus exists and one arm has run, but the baseline comparison has not.

PR #763, opened 2026-08-03
fix: preserve AUTOGROW dotted workflow inputs

### LINKED ISSUE #762: UI-to-API conversion drops COMFY_AUTOGROW_V3 dotted inputs
## Summary

UI-to-API conversion drops linked entries of COMFY_AUTOGROW_V3 inputs. This affects core ComfyUI nodes such as ComfyMathExpression, whose repeated inputs are serialized as values.a, values.b, and so on without a separate values parent value.

## Minimal reproduction

1. Create a UI-format graph with PrimitiveInt connected to ComfyMathExpression input values.a.
2. Define ComfyMathExpression values in object_info as COMFY_AUTOGROW_V3.
3. Run convertUiToApi.

## Actual behavior

The converter initially maps values.a to the source link, then the final dynamic-combo dotted-leaf pass deletes it because inputs does not contain a separate values parent. The resulting API node only contains expression, so validate_workflow reports the required values input as missing and the converted workflow is not runnable.

## Expected behavior

Keep dotted AUTOGROW entries such as values.a and values.b. COMFY_AUTOGROW_V3 uses those dotted entries as its serialized children and does not require a separate scalar parent value.

## Scope and proposed fix

Bypass the dynamic-combo orphan/re-anchoring pass only when the declared parent input type is exactly COMFY_AUTOGROW_V3. Preserve the existing strict behavior for COMFY_DYNAMICCOMBO_V3 and unknown dotted inputs. Add a regression test using PrimitiveInt -> ComfyMathExpression.

Observed with comfyui-mcp 0.49.3 and current main.

flexusjan:
Closes #762.

## Summary

- preserve dotted children such as values.a and values.b when their declared parent is COMFY_AUTOGROW_V3
- keep the existing strict orphan handling unchanged for COMFY_DYNAMICCOMBO_V3 and unknown dotted inputs
- treat required AUTOGROW inputs as present in validateWorkflow when at least one dotted child exists
- retain the missing-required error when no parent and no dotted children are present

## Tests

- npx vitest run src/__tests__/services/workflow-converter.test.ts src/__tests__/services/workflow-validator.test.ts: 80 passed
- npm run build: passed
- end-to-end conversion and validation against a 52-node ComfyUI workflow containing four ComfyMathExpression nodes: valid, no ComfyMathExpression issues
- npm test: 4005 passed, 4 skipped, 5 environment failures in generation-tracker-path.test.ts because this host runs unsupported Node 26 and better-sqlite3 did not install a native binding; the focused suites and TypeScript build pass

### DIFF
--- a/src/__tests__/services/workflow-converter.test.ts
+++ b/src/__tests__/services/workflow-converter.test.ts
@@ -154,6 +154,70 @@ describe("convertUiToApi — bypass / mute resolution", () => {
     expect(workflow["1"].inputs).not.toHaveProperty("aspect_ratio");
   });
 
+  it("preserves linked COMFY_AUTOGROW_V3 entries without a serialized parent value", () => {
+    const objectInfo = {
+      PrimitiveInt: {
+        input: { required: { value: ["INT", { default: 0 }] } },
+        output: ["INT"],
+      },
+      ComfyMathExpression: {
+        input: {
+          required: {
+            expression: ["STRING", { default: "a + b" }],
+            values: [
+              "COMFY_AUTOGROW_V3",
+              {
+                template: {
+                  input: { required: { value: ["FLOAT,INT,BOOLEAN", {}] } },
+                  names: ["a", "b"],
+                  min: 1,
+                },
+              },
+            ],
+          },
+        },
+        output: ["FLOAT", "INT", "BOOLEAN"],
+      },
+    } as never;
+
+    const ui = {
+      nodes: [
+        {
+          id: 1,
+          type: "PrimitiveInt",
+          mode: 0,
+          inputs: [],
+          outputs: [{ name: "INT", type: "INT", links: [1, 2] }],
+          widgets_values: [8],
+        },
+        {
+          id: 2,
+          type: "ComfyMathExpression",
+          mode: 0,
+          inputs: [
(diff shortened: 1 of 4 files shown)


artokun:
Independent correctness review found a P1 no-ship case in the proposed dotted-input handling: an AUTOGROW nested inside a dynamic-combo option can still have its wired child silently deleted. For a key such as `model.images.<entry>`, the current logic recognizes only `model`, then resolves `images.<entry>` as a missing dynamic-combo leaf and drops it. Please preserve the full nested AUTOGROW parent path and add a regression for that existing dynamic-combo AUTOGROW fixture. I will re-review once updated.

flexusjan:
Addressed in 6e7c8d1. The converter now resolves the full schema path of AUTOGROW ancestors, including nested paths such as `model.images`, before the dynamic-combo re-anchoring pass. I also extended the existing Nano Banana dynamic-combo/AUTOGROW fixture with a wired `model.images.image_1` child and assert that the link survives conversion.\n\nVerification:\n- targeted converter + validator tests: 80 passed\n- `npm run build`: passed\n- full suite: 4007 passed, 4 skipped; the only 5 failures are the existing generation-tracker path tests because the local `better-sqlite3` native binding is unavailable.

Issue #728, opened 2026-08-02
PanelAgent stall watchdog duplicates the interrupted failure and releases the turn gate early


JusticeWay:
## Summary

`PanelAgent.onTurnStalled()` currently owns a visible stall warning, but then it releases the turn gate synchronously with `completeTurn()` and calls `backend.interrupt()` directly.

If the backend subsequently emits the expected terminal result

```ts
{ type: "result", ok: false, subtype: "interrupted" }
```

the result handler does not know that this was watchdog-initiated. The panel therefore shows a second, contradictory failure:

1. `The agent stopped responding...`
2. `That turn failed (interrupted) without a reply...`

The synchronous `completeTurn()` also makes the next queued batch eligible before the interrupted result (or the existing bounded interrupt fallback) has settled the turn.

## Current code

This is still present on current `main` at `2b42e8f5303aa9c79ec9493de9a15fe05f87fd26` and in npm `0.49.0`:

```ts
this.deps.onSay(this.tabId, "⚠️ The agent stopped responding ...");
this.busy = false;
this.inFlight = null;
this.completeTurn();
this.deps.onTurn?.(this.tabId, "done");
void this.backend.interrupt().catch(() => {});
```

PR #310 correctly scoped Codex app-server watchdog liveness to active-turn traffic, but it did not change this provider-agnostic `PanelAgent` terminal path.

## Deterministic reproduction

A small fake backend can reproduce this without waiting for the production idle timeout:

1. Set `COMFYUI_MCP_TURN_IDLE_MS=40` before importing `PanelAgent`.
2. Start one turn and keep its backend generator silent.
3. Queue a second turn after the first has been dispatched.
4. Let the watchdog call `interrupt()`, but hold back the first turn's `interrupted` result.
5. Observe that `completedTurns === yieldedTurns` before that result arrives.
6. Release `{ ok: false, subtype: "interrupted" }` and observe both user-visible warnings.

On clean `0.49.0`, the gate assertion is `1 !== 0` and the duplicate message is emitted.

## Suggested fix

The stall path can use the existing guarded interrupt flow instead of duplicating it:

```ts
this.errorSurfaced = true;
this.interruptRequested = true;
this.busy = false;
this.inFlight = null;
this.deps.onSay(this.tabId, "⚠️ The agent stopped responding ...");
this.deps.onTurn?.(this.tabId, "done");
void this.interrupt();
```

[...]

PR #786, opened 2026-08-03
fix(panel_run): make the completion event durable across automatic goal continuation (#468)

### LINKED ISSUE #468: panel_run completion event can be lost during automatic goal continuation
`panel_run` queued successfully, ComfyUI completed the render and committed its PreviewImage output, but the promised automatic run-finished event was not delivered into the agent turn while automatic goal-continuation turns kept firing. The agent had to wait and then recover the committed output with `get_history` + `get_image`. This reproduced on two consecutive panel runs. The likely issue is a race or dropped handoff between the panel completion event and automatic goal-continuation turn scheduling.

### Steps to reproduce
1. Start an active persistent goal in the sidebar agent.
2. Call `panel_run` with `to_node_id` targeting a PreviewImage.
3. End the turn as instructed.
4. Observe automatic goal-continuation turns occur, but no `[panel event] Run finished` message arrives.
5. After the render duration, call `get_history`; it reports success and the expected PreviewImage file.

### Exact error
No explicit error. Silent missing completion event. Recovery evidence from the second occurrence:
- `panel_run` returned `queued: true`, `ran_to_node: 49`.
- After a bounded wait, `get_history` reported execution `5bc36c41-fe83-481f-8a71-b9dd1c3b05a4`, status success, duration 57.45s, node 49 output `ComfyUI_temp_gfxrb_00001_.png`.

### Fix
No safe local patch was attempted because the failure is in orchestration/event scheduling rather than workflow or node code, and the running distribution does not expose a clear source stack. Precise change needed: persist panel run-completion notifications by prompt/job id and deliver them exactly once after any currently active automatic continuation turn, rather than allowing continuation scheduling to supersede/drop the pending event. Add a regression test covering `panel_run` completion concurrent with automatic goal continuation.

### Environment
Windows 11
ComfyUI 0.29.0
RTX PRO 4000 Blackwell, 24 GB VRAM
comfyui-mcp 0.48.11
panel 0.11.3

artokun:
`panel_run` makes a promise: *"You will be notified automatically with the output image(s)/video when the render finishes — do NOT poll. Just end your turn now and wait."* This PR makes that promise keepable.

Closes #468

## Root cause — why *continuation* is what breaks it

The completion is delivered by exactly one mechanism: the panel's `agent_event` frame → `PanelAgentManager.injectEvent` → `PanelAgent.queue`. That path was fire-and-forget in three independent ways.

1. **Uncorrelated.** The completion carried no run identity anything downstream looked at. `panel_run` already learns ComfyUI's `prompt_id` (it hands it to `QueueMonitor.markSelfQueued`) but nothing used it, so a completion could not be tied to the run that was outstanding — and an outstanding run could never be known to be unanswered.
2. **Silently dropped with no live agent.** `manager.injectEvent` returns `false` for a missing/stopped agent and the caller only logged on success. Nothing recorded the loss. The event was gone with no line in the log.
3. **Died with a queued-but-unread item.** The event is only really delivered when `channel()` splices it into a turn. Everything before that — a `stop()`/`retire()`, a stall-abandoned turn, a `takePending()` race — discards it.

An **ordinary single run** ends the agent's turn and leaves it idle, so the completion lands in an empty queue and drains immediately; none of the above shows. **Automatic goal continuation** is what turns those windows from theoretical into routine:

- the agent is **busy for the entire render**, so the completion sits in `queue` (window 3) for minutes instead of microseconds;
- the continuation's own turn churn is exactly what fires the session restarts `applyPendingRestarts` defers "until idle" (effort/model change, comfyui-MCP-env respawn) plus the self-restart loop — each of which **tears the listener down** (windows 2 and 3) underneath the very render whose completion is in flight.

Nothing noticed, because there was no ack that the agent had consumed the event and no outstanding-run bookkeeping. A lost completion was indistinguishable from a slow render, so the agent simply waited.

## The fix — journal + replay, mirroring panel #514

New `src/orchestrator/run-completion-journal.ts`.

- **Run tickets.** `panel_run` opens one keyed by ComfyUI's `prompt_id` and the panel tab that queued it.
- **Correlated ONCE, at arrival**, by **exact** prompt-id equality **and** same tab. There is deliberately no "the newest outstanding run" fallback and no cross-tab match.
- **Journaled and replayed.** Every completion is recorded before delivery is attempted; a refused delivery is logged and retried at the next opportunity (a fresh agent spawn, a hand-back, a proven tab-id migration, a later completion for the same tab). Delivery preserves arrival order and stops at the first refusal.
- **Acked only by the turn that carried it.** `injectEvent` now takes an `eventToken` that rides the queue item and the in-flight turn. The entry leaves the journal only when *that turn's* `result` lands. Handed to an agent that then dies, it comes back and is replayed. Every other exit from a turn — stall-watchdog abandon, plain-Stop interrupt, rewind, session end with no result, `stop()`, a discarded held-mail `reset()` — hands the tokens **back**.

### Nothing is ever silently swallowed

- A completion that **can't be correlated** is still delivered, explicitly labelled: *"its origin is UNDETERMINED. Do NOT assume it is your run; verify yours with `get_history`."*
- A **replayed** one says so (`RE-DELIVERED`) so it reads as "this landed late", not as another render.
- A completion **evicted** by the journal's bounds is counted and reported on the next delivery that gets through.
- `panel_run` itself no longer over-promises: when the panel forwards **no `prompt_id`**, the anti-poll instruction is replaced with an honest one — the outcome will be UNDETERMINED, don't idle indefinitely, confirm with `get_history`.

### Never attributed to the wrong run

A completion on the wrong run is worse than a lost one, so:

- the verdict is **frozen at arrival** and never re-derived on replay — a stranded completion cannot drift onto a run that started after it landed;
- a correlation may only ever get **weaker** (`matched` → `foreign`), never stronger, and only on an explicit "that conversation is gone" event (`closeRuns`, wired into New chat, resume-session, and in-place workflow replacement);
- an entry is addressed to **one** panel tab and is only ever re-addressed (`moveKey` on a *proven* same-workflow tab-id migration — which carries the tickets, the delivered memo and the eviction counter with it), never broadcast; an unproven workflow switch **forgets** the old tab's entries *and* tickets instead of migrating them;
- a `matched` entry settles **only its own** ticket, and only on ack;
- **only the carrying turn's own `result` may ack.** #728 deliberately lets an *unmarked* result past the dead-letter guard for gate liveness; on a marker-stamping backend that result may belong to an abandoned earlier turn, so it is no longer allowed to retire a completion the current turn is holding. Leftovers are handed back at the next dispatch.
- dedupe on `(tab, run)` — including after the ack, via a bounded memo — so a re-sent frame can never produce a second delivery.

## Tests

`src/__tests__/orchestrator/run-completion-continuation.test.ts` (26). Delivered across a continuation while the agent is busy the whole render; journaled with no live agent and replayed to the right run; replayed after its carrying turn is abandoned; uncorrelatable and foreign completions reported as UNDETERMINED and never settling the awaited run; a replay never satisfying a different run; the held-mail/`reset()` path; the traceless-straggler ack gate (verified to fail without the gate); plus the journal's correlation, dedupe, eviction-disclosure, `moveKey`/`forget` and ordering invariants. `panel-tools.test.ts` gains the no-`prompt_id` honesty case.

`npm run build` clean; full suite **240 files / 4061 passed**, 2 skipped.

## Review

5 rounds of `gpt-5.6-terra` adversarial review, ending **PASS**. It found and drove out: the global-ticket workflow-switch misattribution (→ tab-scoped correlation + ticket-forgetting), silent eviction of undelivered completions (→ counted and disclosed), a re-arm that could double-deliver, a re-entrancy in the tab-collision path that flushed into the superseded tab's agent, held mail discarded without releasing its tokens, `moveKey` leaving the delivered memo and eviction counter behind, `matched` entries surviving a New chat, and the unmarked-straggler ack.

### DIFF
--- a/src/__tests__/orchestrator/panel-tools.test.ts
+++ b/src/__tests__/orchestrator/panel-tools.test.ts
@@ -27,6 +27,7 @@ import {
 } from "../../orchestrator/panel-tools.js";
 import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
 import { markReplyTimeout } from "../../services/ui-bridge.js";
+import { RunCompletions } from "../../orchestrator/run-completion-journal.js";
 
 type Forwarded = Record<string, unknown>;
 
@@ -707,7 +708,10 @@ describe("panel-tools: panel_run verdict is derived from the ComfyUI reply", () 
   it("#194: a root SaveImage run-to-node the panel ACCEPTS is queued success (not subgraph-rejected)", async () => {
     const reply: ToolResult = {
       content: [
-        { type: "text", text: JSON.stringify({ queued: true, batch_count: 1, to_node_id: 9 }) },
+        {
+          type: "text",
+          text: JSON.stringify({ queued: true, batch_count: 1, to_node_id: 9, prompt_id: "p-194" }),
+        },
       ],
     };
     const { ctx, calls } = makeRunCtx(reply);
@@ -777,12 +781,33 @@ describe("panel-tools: panel_run verdict is derived from the ComfyUI reply", () 
 
   it("a genuine queue (queued:true, no errors) still gets the anti-poll note", async () => {
     const reply: ToolResult = {
-      content: [{ type: "text", text: JSON.stringify({ queued: true, batch_count: 1 }) }],
+      content: [
+        { type: "text", text: JSON.stringify({ queued: true, batch_count: 1, prompt_id: "p-ok" }) },
+      ],
     };
     const { ctx } = makeRunCtx(reply);
     const res = await defByName("panel_run").handler({}, ctx);
     expect(res.isError).toBeFalsy();
     expect(textOf(res)).toContain(QUEUED_NOTE);
+    // #468 — the anti-poll instruction is only safe because the completion can be
+    // correlated back to this exact run; the ticket is what makes that possible.
+    expect(RunCompletions.ticketFor("p-ok")).toBeDefined();
+  });
+
+  it("#468: a queue with NO prompt id does NOT promise automatic delivery — it says UNDETERMINED", async () => {
+    // The panel forwards ComfyUI's /prompt reply; without a prompt_id there is
+    // nothing to correlate a later completion to, so telling the agent to idle
+    // and wait would park it on a promise the orchestrator cannot keep.
+    const reply: ToolResult = {
+      content: [{ type: "text", text: JSON.stringify({ queued: true, batch_count: 1 }) }],
+    };
+    const { ctx } = makeRunCtx(reply);
+    const res = await defByName("panel_run").handler({}, ctx);
+    expect(res.isError).toBeFalsy();
+    const text = textOf(res);
+    expect(text).not.toContain(QUEUED_NOTE);
+    expect(text).toContain("UNDETERMINED");
+    expect(text).toContain("get_history");
   });
 });
 
--- a/src/__tests__/orchestrator/run-completion-continuation.test.ts
+++ b/src/__tests__/orchestrator/run-completion-continuation.test.ts
@@ -737,13 +737,25 @@ describe("run completion journal correlation (#468)", () => {
     expect(journal.pending("t")[0].correlation.status).toBe("foreign");
   });
 
-  it("a completion re-sent AFTER its delivery was acked is suppressed, not delivered twice", () => {
+  it("a completion re-sent AFTER its delivery was acked is FLAGGED, never suppressed", () => {
+    // "Already delivered" is a LABEL, not a veto. Every proof it could rest on
+    // (the memo, the ticket's settled flag, the ticket itself) is bounded, so
+    // suppressing on one turns into a silent loss the moment it expires at the
+    // wrong time — which is exactly how this defect kept coming back.
     journal.openRun(PROMPT_A, { tabId: "t" });
     const entry = journal.record("t", { kind: "executed", prompt_id: PROMPT_A });
     journal.deliverPending("t", () => true);
     journal.ack(entry!.token); // the carrying turn ended — entry is gone
-    expect(journal.record("t", { kind: "executed", prompt_id: PROMPT_A })).toBeNull();
-    expect(journal.pending("t")).toHaveLength(0);
+    const resend = journal.record("t", { kind: "executed", prompt_id: PROMPT_A });
+    expect(resend.possibleRepeat).toBe(true);
+    expect(journal.pending("t")).toHaveLength(1);
+    // …and the flag reaches the agent, so it can decline to double-count it.
+    const seen: CompletionPayload[] = [];
+    journal.deliverPending("t", (p) => {
+      seen.push(p);
+      return true;
+    });
+    expect(seen[0].possible_repeat).toBe(true);
   });
 
   it("evicting a HANDED-OFF completion is counted too — a hand-off is not proof of consumption", () => {
@@ -829,7 +841,7 @@ describe("run completion journal correlation (#468)", () => {
     expect(first!.correlation.status).toBe("matched"); // not reused yet
     journal.deliverPending("t", () => true);
     journal.ack(first!.token);
-    expect(journal.record("t", { kind: "executed", prompt_id: PROMPT_A })).toBeNull(); // memo holds
+    expect(journal.record("t", { kind: "executed", prompt_id: PROMPT_A }).possibleRepeat).toBe(true);
 
     journal.openRun(PROMPT_A, { tabId: "t" }); // ComfyUI reused the id
     expect(journal.ticketFor(PROMPT_A)?.reused).toBe(true);
@@ -1009,6 +1021,39 @@ describe("run completion journal correlation (#468)", () => {
     expect(real!.payload.note).toBe("run B");
   });
 
+  it("even with ALL history expired, a resend can't suppress the run that reused its id", () => {
+    // The last ordering: A is acked, then enough later runs expire BOTH its
+    // ticket (64) and its delivered memo (512), so nothing remembers the id at
+    // all. B then queues it and looks, correctly, like a fresh identity.
+    // A's late resend is now indistinguishable from B's completion — and that is
+    // fine, because "already delivered" no longer VETOES anything. Whatever
+    // arrives is delivered; the worst case is a flagged duplicate, never a loss.
+    journal.openRun(PROMPT_A, { tabId: "t" });
+    const a = journal.record("t", { kind: "executed", prompt_id: PROMPT_A, note: "run A" });
+    journal.deliverPending("t", () => true);
+    journal.ack(a.token);
+
+    for (let i = 0; i < 600; i++) {
+      journal.openRun(`later-${i}`, { tabId: "t" });
+      const e = journal.record("t", { kind: "executed", prompt_id: `later-${i}` });
+      journal.deliverPending("t", () => true);
+      journal.ack(e.token);
+    }
+    expect(journal.ticketFor(PROMPT_A)).toBeUndefined(); // all history expired
+
+    journal.openRun(PROMPT_A, { tabId: "t" }); // B — looks fresh, and is treated so
+    const resend = journal.record("t", { kind: "executed", prompt_id: PROMPT_A, note: "A resend" });
+    journal.deliverPending("t", () => true);
+    journal.ack(resend.token); // the resend settles the ticket it matched
+
+    // B's REAL completion still lands. Under the old suppress-on-proof rule this
+    // was the loss: B's own settled flag swallowed it.
+    const real = journal.record("t", { kind: "executed", prompt_id: PROMPT_A, note: "run B" });
+    expect(real.payload.note).toBe("run B");
+    expect(real.possibleRepeat).toBe(true); // …flagged, so it isn't double-counted
+    expect(journal.pending("t")).toHaveLength(1);
+  });
+
   it("the no-coalesce rule survives the reused TICKET being evicted", () => {
     // Tickets are the smallest bound (64) and evictable. If ambiguity were read
     // only from the live ticket, 64 later runs would make the incoming side
@@ -1116,15 +1161,16 @@ describe("run completion journal correlation (#468)", () => {
     expect(seen[0].dropped_completions).toBe(lost);
   });
 
-  it("a resend is suppressed for as long as the run's IDENTITY survives, then reported as undetermined", () => {
-    // Suppression is keyed by ticket GENERATION, so it lasts exactly as long as
-    // the identity that makes "we already reported THIS run" a true statement.
+  it("a resend is FLAGGED while the run's IDENTITY survives, and merely undetermined after", () => {
+    // The repeat FLAG is keyed by ticket generation, so it lasts exactly as long
+    // as the identity that makes "we already reported THIS run" a true statement.
+    // Either way the completion is delivered — only the labelling changes.
     journal.openRun(PROMPT_A, { tabId: "t" });
     const entry = journal.record("t", { kind: "executed", prompt_id: PROMPT_A });
     journal.deliverPending("t", () => true);
     journal.ack(entry!.token);
     expect(journal.ticketFor(PROMPT_A)?.settled).toBe(true);
-    expect(journal.record("t", { kind: "executed", prompt_id: PROMPT_A })).toBeNull(); // resend: suppressed
+    expect(journal.record("t", { kind: "executed", prompt_id: PROMPT_A }).possibleRepeat).toBe(true);
 
     // Flood with REAL runs so the ticket map overflows and this run's identity
     // is evicted. Past that point the journal genuinely cannot tell a resend
@@ -1435,8 +1481,10 @@ describe("run completion journal correlation (#468)", () => {
     journal.moveKey("tmp:draft", "wf:saved.json");
     expect(journal.droppedFor("wf:saved.json")).toBe(lost);
     expect(journal.droppedFor("tmp:draft")).toBe(0);
-    // The re-send is still suppressed under the NEW id.
-    expect(journal.record("wf:saved.json", { kind: "executed", prompt_id: PROMPT_A })).toBeNull();
+    // The re-send is still RECOGNISED under the NEW id (flagged, not suppressed).
+    expect(
+      journal.record("wf:saved.json", { kind: "executed", prompt_id: PROMPT_A }).possibleRepeat,
+    ).toBe(true);
   });
 
   it("moveKey re-addresses a tab's entries without touching anyone else's", () => {
--- a/src/__tests__/orchestrator/run-completion-continuation.test.ts
+++ b/src/__tests__/orchestrator/run-completion-continuation.test.ts
@@ -247,6 +247,42 @@ describe("run completion across automatic goal continuation (#468)", () => {
     expect(journal.ticketFor(PROMPT_A)?.settled).toBe(false); // our run is still open
   });
 
+  it("a completion parked in HELD mail survives a reset that discards that mail", async () => {
+    // A failed start parks the agent's whole queue — including an injected
+    // completion — in heldMessages. `reset()` (New chat / resume_session) throws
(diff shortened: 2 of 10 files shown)


artokun:
All four gate findings fixed, each with a mutation-verified regression test. Re-gated with `gpt-5.6-terra`: it found 3 more (2 SEVERE) on the first pass over the fixes, those are fixed too, and round 7 returned **PASS**.

## CRITICAL — cross-tab replay via a journal-only destination collision

`destinationHasCollisionState` now takes `journaledCompletionCount`, and `index.ts` feeds it
`RunCompletions.outstanding(panelTab)` — **a tab holding a journal entry is not empty**. The count comes from a read-only pre-pass so it reaches *both* the pre-loop purge and the per-backend collision branch (which also drives `bridge.dropQueuedDeliveries`). The destination's entries **and tickets** are purged via `forget(panelTab)` **before** anything in the loop can hand tokens back — never silently inherited by the rebound source agent. `forget` logs every entry it drops, so the loss is disclosed.

## HIGH — an unmarked straggler falsely acking the carrying turn

Learning-by-observation is gone. `AgentCapabilities.turnMarkers` is **declared** (true on every in-tree backend — all of them call `stampTurn` or stamp natively), so the gate holds from the first event of a run, including the exact window you named: a zero-output turn's traceless terminal arriving before the replacement turn stamps anything. A non-ackable result now hands the tokens **back immediately** rather than waiting for the next-dispatch net.

That change alone would let a backend which declares markers but never stamps its results replay forever, so `release()` is bounded — and codex immediately caught my first bound as its own SEVERE: I was counting **queue hand-offs**, so three ordinary provider/session replacements would have settled (and lost) a completion nobody read. `release()` now takes `{ carried }`: only a turn that actually *dispatched* with the completion and then ended counts toward `MAX_CARRIED_RELEASES`. Every teardown hand-back is uncarried and unbounded.

## HIGH — held-mail completions stranded across `retire()`

`reset()` and `retire()` funnel through one `unbindAgent()` teardown, which calls `detachHeldCompletions()`: it strips the tokens out of held mail (leaving the mail itself alone, so `retire` keeps its preservation semantics) and hands them back — **even when no live agent owns the key**, which is exactly the case that lost them. A future third teardown gets this by construction.

Codex then found the *same* hole one branch over: `settle()`'s `gaveUp` path dropped the agent without taking its queue, stranding the entry `handed_off` — a state `deliverPending()` skips — permanently. It now mirrors the `err` branch (`takePending()` → held mail, tokens riding along, plus `stop()`).

## MEDIUM — duplicate id-less completions

Split by what a mistake costs, because codex's counter-finding is right that a blanket content memo can **swallow a real render** (ComfyUI reuses `ComfyUI_temp_*_00001_` across a restart):

- **still journaled and undelivered**, within 30s (a transport-retry timescale, not a render one) → collapse onto one entry. Neither copy reached anyone, so this cannot lose news the agent already has. This is what stops one output producing two turns.
- **already delivered** → **never suppressed**. It is forwarded flagged `possible_repeat`, and the agent is told: *a completion with identical outputs was already reported, and this one carries no prompt id to tell them apart — do NOT count it twice without checking `get_history`.* Swallowing the wrong one is a silent loss, which is the failure this whole change exists to prevent; the agent gets the judgement call, with the ambiguity stated.

## Tests

`run-completion-continuation.test.ts` is now 34 tests. Each of the four findings has a regression test that I **mutation-verified** (reverted the fix, confirmed the test fails, restored):

| fix | mutation | result |
|---|---|---|
| journal-aware occupancy | `journaledCompletionCount` → `false` | fails |
| declared marker ack gate | gate → always ackable | fails (both straggler tests) |
| `retire()` teardown | `unbindAgent` → bare map delete | fails |
| id-less dedupe | fingerprint lookup → undefined | fails |
| carried/uncarried split | `if (opts.carried)` → always | fails |
| gaveUp queue rescue | branch → drop the queue | fails |
| id-less post-ack flag | flag → suppress | fails |

`npm run build` clean; full suite **240 files / 4069 passed**, 2 skipped.

artokun:
Both gate findings fixed, each mutation-verified. My own re-gate then found **9 more** across 6 rounds (5 SEVERE) — all fixed — and round 18 returned **PASS**, with the reviewer confirming no remaining uncarried release outside the one documented exception.

## HIGH (1) — retained held mail re-delivering a completion whose token had moved on

You were right that leaving the mail intact is what opened this. The completion item now leaves held mail **with** its token. That is only safe because an injected event is always its own `completionOnly` item — which required making a re-queued turn restore the **original queue items** (`InFlightTurn.items`) instead of one merged blob, on both the interrupt-requeue and crash-requeue paths. A token found on a non-`completionOnly` item logs at ERROR and keeps its text; construction makes that unreachable.

Codex immediately found the same hole one hop further out: `send()` dropped `completionOnly`, so after **two** consecutive failed starts the item returned to held mail unmarked and became un-removable again. Every carry-over path now preserves it, and the regression test fails the first two starts specifically to exercise that round trip.

## HIGH (2) — the 30s id-less collapse swallowing a real second render

Fixed exactly as you specified: the collapse predicate now requires `state === "pending"`. Anything `handed_off` or delivered produces its own entry, flagged `possible_repeat`, never a silent merge. You were right that this landed on the wrong side of my own tradeoff.

## Latent — direct-token Grok

`GROK_DIRECT_CAPABILITIES` now declares `turnMarkers: true`, matching its `stampTurn` behaviour.

## What my re-gate found in those fixes

The two SEVEREs worth naming, because both were mine:

- **`MAX_HANDOFFS` counted queue hand-offs, not reads** — three ordinary provider/session replacements would have settled and lost a completion nobody read. `release()` now takes `{ carried }`, and after several more rounds *every* release site passes `carried: turnProducedEvents` — proof the backend actually received that turn (a marker-stamped event for it). An abort during Claude's pre-submission image resolution, or a watchdog trip on a turn that produced nothing, is uncarried and cannot count. The only exception is a plain user Stop, documented at the call site.
- **A reused prompt id is not a run identity.** ComfyUI reusing an id made `requeue-then-resend` present run A's output as run B's *and* suppress B's real completion. A reopened ticket is now marked `reused`: correlation returns `foreign` (UNDETERMINED) and **both** already-delivered proofs are disabled, trading a possible duplicate for the impossibility of a loss or misattribution. And because the wording is baked in at queue time, the journal gained an unsend hook — a still-unread copy is pulled back off the queue and re-delivered with the honest wording.

Also fixed along the way: `settle()`'s `gaveUp` branch dropped the agent's queue (same hole as `retire()`); ordinary SIGTERM/`uncaughtException` bypassed the loss disclosure; `stopAll()` cleared held mail without detaching; the eviction disclosure was spent at hand-off rather than at ack, and could be stamped on an entry `deliverPending` never reads; `openRun` left a stale delivered memo on the fresh-ticket path.

## Tests

`run-completion-continuation.test.ts` is now **50 tests**. Every fix in this round and the six that followed has a regression test that I mutation-verified — reverted the fix, confirmed the specific test fails, restored. Notably the unmarked-straggler mutation reproduces your exact loss shape (`expected [] to have a length of 1`).

One test-harness bug worth flagging: the harness was dropping `onEventUndelivered`'s third argument, so it had been silently exercising the unbounded path. Fixed to mirror `index.ts` exactly.

`npm run build` clean; full suite **240 files / 4085 passed**, 2 skipped. (Three unrelated flakes — `ui-bridge` port-bind, `node-dev` real-git, `training-jobs` CAS race — each green in isolation and on re-run.)

artokun:
Both fixed, mutation-verified. The gate pass then caught that my P1 fix was incomplete in exactly the way you predicted — a policy defeated by a path that didn't consult it — so that is fixed too.

## P1 — `reused` disabling the proofs, then coalescing anyway

You had it exactly right: coalescing assumes the prompt id identifies one run, which is precisely what `reused` voids. `record()` no longer coalesces at all for a reused id — B gets its own entry, flagged UNDETERMINED like every other ambiguous case.

**The gate then found my fix reintroduced the same loss one bound away.** `idReused` was read from the live ticket, and tickets are the smallest bound (64) and evictable: reuse the id, hand off A, open 64 later runs to evict the reused ticket, and B once again coalesces into A's entry — A's ack removes the sole entry, B never delivered. The ambiguity is now frozen onto the **entry** (`ambiguousId`) at creation, and the coalesce skips any entry carrying it. The entry outlives the ticket, so the ambiguity does too. The regression test evicts the ticket with 80 later runs and fails without the mark.

### The sweep you asked for

Every prompt-id-keyed site, and how each is covered:

| site | status |
|---|---|
| `correlate()` | consults `reused` → returns `foreign` |
| `record()` suppression (memo + settled ticket) | consults `reused` → both bypassed |
| `record()` coalesce | **was the hole** → now skips on `idReused` *and* on `entry.ambiguousId` |
| `openRun()` supersede loop | is the reuse handler itself |
| `ack()` ticket-settle | unreachable for a reused id (correlate returns foreign; a pre-reuse matched entry is superseded → early return) |
| `ack()` memo write | skipped for a reused id — **defense in depth only**, see below |
| `moveKey` / `forget` / `closeRuns` | key by tab, not id |

One honesty note: the `ack()` memo guard has **no reachable failing case** and therefore no test — `openRun` already clears the memo on every path. I kept it because writing an "already reported this run" proof about an id standing for two runs is meaningless on its face, and labelled it as defense-in-depth in the comment rather than claiming coverage I don't have.

## P2 — the fatal-path disclosure

The durable half is now a **synchronous multi-sink write**: `writeSync(2)` → `writeSync(1)` → `appendFileSync` beside the lockfile, with the async logger only as a last resort (on `uncaughtException`, `process.exit(1)` follows immediately, so anything merely queued is dropped). It emits **one** write for all tabs rather than one per tab, minimising bytes for the case you'd expect to block.

`shutdown()` is re-entrant-safe: the teardown promise is memoized, so a second SIGTERM awaits the first teardown instead of racing past `shuttingDown`'s early return.

Residual, stated plainly: a synchronous write to a **full pipe whose reader has stalled** still blocks. That is inherent to any guaranteed-durable write — I judged blocking during shutdown strictly better than losing the record, since losing it is what undercuts the in-memory-journal tradeoff. The file sink covers a closed/erroring stderr; a wedged reader is a broken environment and now has two further sinks to fall through to.

## Verification

`npm run build` clean; full suite **240 files / 4087 passed**, 2 skipped. `run-completion-continuation.test.ts` is 52 tests. Both new fixes mutation-verified — the entry-mark mutation reproduces the gate's eviction scenario exactly (`expected 'rc1' not to be 'rc1'`), and the no-coalesce mutation reproduces yours.

artokun:
Both P1s fixed, mutation-verified. Your diagnosis on (1) was the important one, and following it properly ended up removing the defect *class* rather than the instance.

## P1 (1) — the `pending` predicate

Implemented exactly as you asked: `record()` coalesces only onto an entry that is still `pending`. You were right that it closes every ordering at once, and right about the pattern — this was the second time a correct policy was defeated by a path that didn't consult it.

Writing that test immediately exposed a second defect: `ack()` resolved the ticket by prompt id alone, so run A's late completion settled whatever ticket the id maps to *now* — run B's. Tickets now carry a generation (`seq`, bumped on every open/reopen); entries record the generation they matched; `ack()` settles only that exact generation.

The gate then found three more orderings of the same shape, and each fix I made was itself defeated one bound away:

- coalesce/memo still resolved by id → made the **generation** the identity everywhere (memo keyed `(tab, id, generation)`).
- the retire/downgrade pass ran only on the reopen path → it now runs on the fresh-ticket path too.
- generation `0` was shared by every foreign completion → a bucket, not an identity; those are now unprovable and never merged or suppressed.
- and finally: `hasHistoryFor` is itself bounded, so after 513 later runs a miss looked like a fresh identity and the original loss returned.

That last one made the real problem plain: **every "already delivered" proof is bounded, so any rule that SUPPRESSES on one becomes a silent loss the moment it expires at the wrong time.** So suppression is gone. `record()` never returns null and never discards a completion — it may collapse onto a twin nobody has seen yet (same state, same generation), otherwise it creates its own entry flagged `possible_repeat`, and the agent is told not to double-count without checking `get_history`. The bounded ticket/memo/history machinery now only *labels*; nothing about correctness depends on any of it still being present. That is what finally held.

Your sweep instruction is what got me there — I just had to apply it one level up, to the rule rather than the sites.

## P1 (2) — the hang

Both changes made. `UiBridge` tracks every accepted socket from `accept` (not from `hello`), and `stop()` closes them all before awaiting `wss.close()` — the anonymous-socket hang is the "not in the collection I happen to consult" shape you named. A repeated signal now waits at most 3s for the in-flight teardown before forcing the exit.

The gate then caught that the 3s bound was itself bypassable: the synchronous disclosure `writeSync(2, …)` blocks the event loop if stderr is a full pipe with a stalled reader, so Node could never dispatch the second signal. The durable record now goes to a **file** first (beside the lockfile — a regular file always makes progress), with stderr/stdout as fallbacks and the async logger for console visibility. Your note that disclosure already happens before the wait was correct and is preserved.

## Verification

`npm run build` clean; full suite **240 files / 4096 passed**, 2 skipped. The test file is **61 tests**. Every fix in this round is mutation-verified — reverted, confirmed the specific test fails with the predicted symptom, restored. Notably:

| mutation | symptom |
|---|---|
| drop the `pending` predicate | `expected 'rc1' not to be 'rc1'` (merge into committed text) |
| ack settles by id alone | `expected true to be false` (A settles B) |
| suppress instead of flag | `TypeError: Cannot read properties of null` (B swallowed) |
| skip the eviction revoke | `expected 60 to be 32` (unbounded agent queue) |

Final gate: one MODERATE, now fixed (eviction deleted the journal record but left the queued payload, so a resend loop grew the agent's queue without bound while the journal stayed at 32 — eviction now revokes the queued copy too).

artokun:
All three fixed, each mutation-verified. Gate pass: **PASS**, no findings.

Your framing — *each is the redesign's fallback or optimisation path retaining the property the main path shed* — was the useful part, and I used it as the review lens rather than just fixing the three instances.

## P1 — the id-less collapse

Removed, not narrowed. Your trace is right and it's the same argument one step further than I took it: for an id-less completion, **"a twin nobody has seen yet" is not establishable**. With no prompt id and reusable temp names, same-tab / same-content / both-pending is precisely what two genuinely different renders look like — so the collapse was still a suppression, and it failed exactly as every other bounded proof did. My own comment documented the hazard and then performed the merge, which is the tell.

Every id-less completion now gets its own entry, flagged `possible_repeat` when an identical one is already journaled (**any** state, not just handed-off) or was delivered recently. `IDLESS_COLLAPSE_MS` is deleted. The only completion that isn't flagged is one with no predecessor at all.

That closes out the invariant: **the journal never merges and never suppresses — it only ever labels.**

## P2 — eviction can't reach held mail

Took the second option: completion-only items are never held. `holdOrphanedMail()` holds the user's own messages and hands completions straight back to the journal — they are revocable by design and the journal is their record, so replay is bounded where held mail is not. Wired into both `settle` branches (start failure and gave-up). `detachHeldCompletions` stays as a defense-in-depth net for a state that should now be unreachable.

## P2 — the blocking fallback

You ranked it correctly and I've taken that ranking: **an unkillable process is worse than a missing log line.** The file append is now the *only* synchronous sink; if it fails we fall through to the async logger and accept losing the durable record. No `writeSync` remains on any path, so nothing in the disclosure can block the event loop and strand the 3s forced-exit timer.

## Verification

`npm run build` clean; full suite **240 files / 4097 passed**, 2 skipped. Test file is **62 tests**. Mutations:

| mutation | symptom |
|---|---|
| restore the id-less collapse | `expected 'rc1' not to be 'rc1'` (two renders merged into one) |
| hold completion-only items | `expected [] to have a length of 1` (completion stranded in held mail) |

The blocking-fallback fix is a removal, so there is nothing to mutate — its verification is that `writeSync` no longer appears in the file.

## The lens, applied

I swept the remaining degraded paths for the same shape and the gate confirmed it independently: unset `revoke`/`reflush` hooks, `onEventUndelivered` throwing, a full journal, a missing agent, a superseded or ambiguous entry, `retireOlderEntriesFor`, eviction, teardown, `retire`, and both `settle` branches. No fallback still merges, suppresses, grows unbounded, or blocks.

Issue #730, opened 2026-08-02
update_custom_node / reinstall_custom_node report success for packs that don't exist (queue-drain trust, no post-op verification)


artokun:
Found by the live tool sweep (2026-08-02, rig: ComfyUI 0.29.2, Manager v4 dialect).

**Repro**
- `update_custom_node {id: "mcp-sweep-nonexistent-pack"}` → returns ok with `Queued + updated "mcp-sweep-nonexistent-pack" via ComfyUI-Manager.` and queue details `total_count:0, done_count:2`.
- `reinstall_custom_node {id: "mcp-sweep-nonexistent-pack"}` → returns ok: `Queued + reinstalled … (uninstall + install)`, `total_count:0, done_count:4`.
- `install_custom_node {id: "mcp-sweep-nonexistent-pack"}` — the SAME nonexistent id — correctly fails: `was queued but is not present afterward — it was not found in the ComfyUI-Manager registry` (#232 presence verification).

**Problem**: update/reinstall trust the Manager queue drain as proof of work for GENERIC packs. The drain check passes trivially (done_count >= total_count=0) when the Manager never enqueued anything for the unknown id. This is the #639 fabricate-success class — #639 hardened the PANEL path (install_panel post-op verification) and #232 hardened install_custom_node, but the generic update/reinstall paths never got the same post-op presence check.

**Expected**: update/reinstall should verify post-op presence (pack resolvable on disk / in the Manager's installed list) before claiming success, and fail truthfully (`not found in the registry / not installed locally`) exactly like install_custom_node does. Severity: P2 fabricate-success (wrong-workflow confusion — an agent told 'updated' will proceed as if the pack exists).

Sweep evidence: b3-results.jsonl entries update_custom_node/reinstall_custom_node vs install_custom_node on the same id.

Issue #724, opened 2026-08-02
install_panel(action='update') silently no-ops under ComfyUI-Manager 3.x (legacy), same root cause as #639/#424


artokun:
### What happened / root cause
install_panel(action='update') reported the panel already up-to-date per the *minimum required* version (0.11.28+, installed 0.11.32), but a live bug (graph_run refusing to dispatch to a persisted workflow tab — see below) was only fixed further upstream (PRs #553/#554/#555, workflow-uuid enforcement, merged to main after 0.11.32). Calling install_panel(action='update') anyway to reach that fix threw:

> Panel update did NOT apply: nothing changed on disk ... (installed version still 0.11.32) after ComfyUI-Manager reported the queue drained. Manager reported done_count=0 with total_count=0 — it never actually enqueued the update. This is the stale ComfyUI-Manager 3.x silent no-op (#639, same root cause as #424).

So the verified install_panel path is fully blocked on hosts still running legacy ComfyUI-Manager 3.x — the tool correctly detects and reports the no-op (doesn't fabricate success), but there's no working update path left for the panel on this host tier other than manual git operations.

### Steps to reproduce
1. Host has legacy ComfyUI-Manager 3.x installed (per this session's ENVIRONMENT line: "Manager 3.x LEGACY (partial feature set — recommend upgrading to ComfyUI-Manager v4+").
2. Panel installed as a real git checkout (not a registry zip) at custom_nodes/comfyui-agent-panel, currently behind origin/main.
3. Call install_panel(action='update') (or action='sync').
4. Tool correctly detects the Manager 3.x queue no-op and throws instead of reporting false success — but no update actually happens.

### Exact error (scrubbed)
```
Panel update did NOT apply: nothing changed on disk at ~/ComfyUI/custom_nodes/comfyui-agent-panel (installed version still 0.11.32) after ComfyUI-Manager reported the queue drained. Manager reported done_count=0 with total_count=0 — it never actually enqueued the update. This is the stale ComfyUI-Manager 3.x silent no-op (#639, same root cause as #424). Fix: update ComfyUI-Manager on the host (git pull in custom_nodes/ComfyUI-Manager, or pip install -U comfyui_manager) and retry, or reinstall the panel from source (git pull the panel dir / reinstall the pack), then RESTART ComfyUI.
```

### Fix
applied locally: yes (workaround only) upstream-only: yes (for the real fix)

[...]

Issue #718, opened 2026-08-02
panel graph_set_widget fails: tab does not enforce per-command workflow targeting


artokun:
## Summary
panel_set_widget / graph edit tools fail with: graph_set_widget cannot be safely targeted — panel tab wf:workf does not enforce per-command workflow targeting (update the ComfyUI-MCP panel to edit the graph). Read-only tools (panel_graph_outline) still work.

## Environment
- comfyui-mcp 0.49.0
- panel 0.11.4
- ComfyUI 0.27.0
- Manager 3.x LEGACY

## Workaround
Headless enqueue via ComfyUI /prompt API after strip_workflow/get_workflow.

## Impact
Cannot edit live canvas; forced off-canvas runs.

PR #793, opened 2026-08-03
fix(install_panel): make the panel-version remedy actually work across install shapes and sessions

### LINKED ISSUE #766: install_panel targets wrong COMFYUI_PATH on Desktop dual-path installs
install_panel(action:'status') reported installed:false and sync failed (Manager no-op), while the live panel pack was already at C:\Documents\ComfyUI\custom_nodes\comfyui-mcp-panel (v0.11.0). Tools used workspace COMFYUI_PATH=ComfyUI-Installs\ComfyUI\ComfyUI instead of the running Desktop instance's custom_nodes.

Graph write tools failed requiring panel 0.11.35+ with version unknown/stale.

### Steps to reproduce
1. Comfy Desktop with models via shared_model_paths.yaml; custom_nodes under Documents\ComfyUI
2. Workspace path differs from running instance custom_nodes
3. install_panel status → not installed; panel_add_node fails targeting gate

### Exact error
graph_add_node cannot be safely targeted... requires panel 0.11.35+
Panel install did NOT land after Manager queue drained

### Fix
applied locally: yes — git pull in Documents\ComfyUI\custom_nodes\comfyui-mcp-panel to 0.11.38 + restart
upstream-only: path detection — install_panel should resolve running ComfyUI custom_nodes (process/port or desktop path) not only workspace

### Environment
Windows 11 / ComfyUI 0.26.2 / RTX 5070 12GB / comfyui-mcp 0.49.3 / panel 0.11.0→0.11.38

### LINKED ISSUE #769: install_panel ignores trusted local workspace detected by get_environment
The connected ComfyUI is local and get_environment reliably detects its trusted workspace path from the process table, but workspace(get/list) and install_panel use a different resolver and act as if no local workspace exists. This prevents panel status/sync even though the local install is known.

### Steps to reproduce
1. Run a local git ComfyUI on Windows at D:\…\ComfyUI and connect comfyui-mcp to http://127.0.0.1:8188.
2. Call get_environment. It reports deploy_environment='local-git', local.workspace_path='D:\…\ComfyUI', and python_probe_trusted=true with source='process-table'.
3. Call workspace(action='get') and workspace(action='list'). They return workspace_source='none' and an empty list.
4. Call install_panel(action='status'). It returns applicable=false and says no local ComfyUI is configured.

### Exact error / conflicting output
get_environment: local.workspace_path='D:\…\ComfyUI', python_probe_trusted=true
workspace(get): workspace_source='none'
workspace(list): workspaces=[]
install_panel(status): 'Panel management is local-only; no local ComfyUI (COMFYUI_PATH) is configured.'

### Fix
Applied locally: no
Upstream-only: yes
Use one shared workspace resolver across get_environment, workspace, and install_panel. When get_environment has a trusted process-table workspace for the active loopback API target, workspace(get/list) should surface it and install_panel should accept it transiently, or offer a safe explicit persistence step. Do not require a separately configured COMFYUI_PATH for a path already detected from the serving process.

Verified workaround: manually set the detected path as the default workspace before running install_panel sync, or update the panel through ComfyUI-Manager on the ComfyUI host.

### Environment
Windows 11; ComfyUI 0.30.0 local-git; ComfyUI-Manager 3.40; NVIDIA RTX 5090 32 GB; Python 3.12.11; PyTorch 2.13.0+cu130; comfyui-mcp 0.49.3; comfyui-agent-panel 0.11.0.

### LINKED ISSUE #771: install_panel update/sync fails on registry-zip panel installs (no .git) with misleading "not installed locally" error, blocking all graph writes

Panel 0.11.34 was installed from the Comfy Registry (a zip install at `~/ComfyUI/custom_nodes/comfyui-agent-panel`, **no `.git` directory**). comfyui-mcp 0.49.3 requires panel >= 0.11.35, so every panel graph WRITE was refused:

> "graph_set_widget" cannot be safely targeted to the active workflow: panel tab wf:workf does not recheck workflow targeting at the graph write boundary after asynchronous work (detected panel 0.11.34; this MCP requires panel 0.11.35+). Run install_panel(action:'update'), restart ComfyUI, then hard-refresh the ComfyUI browser tab.

Following that instruction does not work. `install_panel(action:'update')` **and** `install_panel(action:'sync')` both fail with:

> {"error":"NODE_MANAGEMENT_ERROR","message":"\"comfyui-agent-panel\" was queued for update but is not present afterward — it is not installed locally and was not found in the ComfyUI-Manager registry, so there was nothing to update. Check the pack id (list_installed_nodes shows what is actually installed).","details":{"total_count":0,"done_count":4,"in_progress_count":0,"pending_count":0,"is_processing":false}}

This message is **wrong and misleading**: `install_panel(action:'status')` in the same session correctly reports it IS installed —

```json
{"applicable":true,"installed":true,"dir":"~/ComfyUI/custom_nodes/comfyui-agent-panel",
 "installedVersion":"0.11.34","isDevSymlink":false,"targetVersion":"nightly","shadows":[],
 "pin":{"pinned":false},
 "sync":{"requiredPanelVersion":"0.11.35","installedVersion":"0.11.34","decision":"sync","behind":true}}
```

So status and update disagree. The likely root cause: the update path relies on ComfyUI-Manager resolving `comfyui-agent-panel` in its registry DB and/or doing a `git pull`, and both fail for a registry-zip install with no `.git`. The documented `git pull --ff-only` fast-forward fallback in install_panel's description cannot apply (not a git checkout), and there is no zip-reinstall fallback — leaving the user with a hard-blocking version gate and no working remedy from any tool.

This is a **self-blocking loop**: the version gate tells you to run install_panel, and install_panel is the thing that can't work.

### Steps to reproduce
1. Install the panel from the Comfy Registry (zip install, no `.git`) at a version below the MCP's floor — here 0.11.34 vs required 0.11.35+.
2. Attempt any panel graph write (`panel_set_widget`) → refused with the version-gate error above.
3. Run `install_panel(action:'status')` → reports installed, version 0.11.34, decision `sync`, behind `true`.
4. Run `install_panel(action:'update')` or `install_panel(action:'sync')` → NODE_MANAGEMENT_ERROR "not installed locally and was not found in the ComfyUI-Manager registry".

### Exact error (scrubbed)
See both blocks above. No secrets involved; home paths shortened to `~/`.

### Fix
<applied locally: no (worked around, not patched)>  <upstream-only: yes>

Workaround that WORKED and unblocked the user (manual, agent-driven):
```bash
cd ~/ComfyUI/custom_nodes
git clone --depth 1 https://github.com/artokun/comfyui-mcp-panel.git _panel_new   # -> 0.11.38
mkdir -p ~/ComfyUI/custom_nodes_backup
mv comfyui-agent-panel ~/ComfyUI/custom_nodes_backup/comfyui-agent-panel-0.11.34
mv _panel_new comfyui-agent-panel
# then panel_restart_comfyui
```
After restart, `install_panel(action:'status')` reports 0.11.38 and `decision:"up-to-date"`. The cloned repo ships prebuilt `web/js` (version string 0.11.38 confirmed in the bundle), so no npm build was needed.

Precise changes needed upstream:
1. **install_panel update/sync must handle a non-git, registry-zip install.** When Manager's update no-ops AND the dir is not a git checkout, fall back to cloning the repo fresh (or re-downloading the registry zip) into a temp dir and atomically swapping it in, backing up the old dir. This is exactly the manual sequence above and it is fully scriptable.
2. **Fix the false error message.** "it is not installed locally and was not found in the ComfyUI-Manager registry" is factually wrong when `status` resolves the same pack at a concrete `dir` with a concrete `installedVersion`. It should say Manager couldn't update it and name the real reason (not a git checkout / not in the Manager DB), and it should not contradict `status`.

### Secondary issue: stale browser bundle is indistinguishable from stale install, and unfixable by the agent
After the successful swap + ComfyUI restart (server-side pack 0.11.38, served `web/js/comfyui-mcp-panel.js` contains 0.11.38), graph writes STILL failed with the **identical** error naming `detected panel 0.11.34` — because the browser tab was still running the cached 0.11.34 bundle.

Two problems here:
- The gate error reports the **tab's** version but its remedy text ("run install_panel(action:'update'), restart ComfyUI") points at the **install**, so there is no way to tell "pack is stale" from "pack is fine, tab cache is stale". Suggested: have the orchestrator compare the tab-reported version against the on-disk pack version and, when they differ, say so explicitly — e.g. "pack on disk is 0.11.38 but this tab is running 0.11.34 — hard-refresh the ComfyUI browser tab (Ctrl+Shift+R)".
- There is **no panel tool to force a frontend reload** of the browser tab. `panel_reload` restarts the orchestrator, not the page. So after any panel pack upgrade the agent is hard-blocked on a manual user action it cannot perform or verify. A `panel_reload_frontend` (or having `panel_restart_comfyui` instruct connected tabs to hard-reload after a pack version change) would close this loop.

Net effect for the user: canvas edits were impossible for the whole session; the render had to be done headlessly via `enqueue_workflow` instead of on their live canvas.

### Environment
- ComfyUI: 0.28.0 (local), ComfyUI-Manager v4
- GPU: NVIDIA RTX 4060 Ti, 16 GB VRAM; 32 GB RAM; CUDA 13.0; torch 2.12.0; python 3.12
- comfyui-mcp: 0.49.3
- comfyui-mcp-panel: 0.11.34 (blocked) -> 0.11.38 (after manual git clone swap)

### LINKED ISSUE #774: Remote panel version drift blocks panel_run and has no verified sync path
panel_run failed before dispatch because the connected remote sidebar panel did not expose the per-command workflow-targeting capability required by comfyui-mcp 0.49.3. The orchestrator requires panel 0.11.35+, but the panel version was reported as unknown. Recovery is deadlocked in remote mode: install_panel(action:"status") returns decision "not-applicable", while panel_update_node intentionally refuses to manage comfyui-mcp-panel.

### Steps to reproduce
1. Connect comfyui-mcp 0.49.3 to a remote ComfyUI sidebar session using an older or version-unknown panel.
2. Read the live graph successfully with panel_graph_outline.
3. Call panel_run({to_node_id:112}).
4. Try install_panel(action:"status") and then panel_update_node({id:"comfyui-mcp-panel", version:"nightly"}).

### Exact error (scrubbed)
graph_run could not be dispatched to this session's panel tab — nothing was applied.
"graph_run" cannot be safely targeted to the active workflow: panel tab does not enforce per-command workflow targeting (detected panel version unknown; this MCP requires panel 0.11.35+).
install_panel status: Remote/cloud mode — panel install is managed on the ComfyUI host, not from here.
panel_update_node: cannot manage the comfyui-mcp sidebar panel pack; use install_panel instead.

### Fix
Precise change needed: provide a remote-safe verified panel sync path through the connected host's Manager, or let the panel report its exact version/capabilities and provide a guarded upgrade operation that re-reads the installed version on the host. The current combination leaves no autonomous recovery path.

### Environment
Linux remote ComfyUI 0.30.0; Python 3.12.3; torch 2.11.0+cu126; RTX 4090 24 GB; comfyui-mcp 0.49.3; panel version unknown; installed panel git commit 55535ce5b95c3f7519fded92ce47d644a1568604.

### LINKED ISSUE #784: Embedded panel write failure recommends unavailable install_panel recovery tool
Live graph writes failed because the connected panel did not advertise/enforce the workflow write stamp required by comfyui-mcp 0.49.3. The recovery path in the error says to run install_panel(action:'update'), but install_panel is not exposed in this embedded panel session. panel_update_node then correctly refuses to manage comfyui-mcp-panel and again points to the unavailable install_panel tool. A remote user is therefore left with no supported in-session self-heal path.

### Steps to reproduce
1. Open a workflow in the ComfyUI sidebar panel.
2. Call panel_graph_outline and panel_query_graph successfully.
3. Call panel_connect or panel_canvas.
4. Observe the workflow-stamp capability rejection requiring panel 0.11.35+.
5. Attempt panel_update_node(id:'comfyui-mcp-panel', version:'nightly').
6. Observe that it refuses and instructs using install_panel, which is absent from the available tool set.

### Exact error (scrubbed)
graph_connect could not be dispatched ... panel tab does not enforce per-command workflow targeting (detected panel version unknown; this MCP requires panel 0.11.35+). Run install_panel(action:'update') and restart ComfyUI.

panel_update_node cannot manage the comfyui-mcp sidebar panel pack ... Use install_panel instead.

### Fix
applied locally: no
upstream-only: yes
Expose install_panel in embedded panel sessions, or add a verified panel_sync/update bridge that can target the connected remote ComfyUI host. The capability error should only recommend tools actually callable in the current session.

### Environment
OS Linux remote; ComfyUI 0.30.0; NVIDIA RTX 4090 24 GB; comfyui-mcp 0.49.3; panel version unknown in handshake; installed comfyui-mcp-panel commit 55535ce5b95c3f7519fded92ce47d644a1568604.

artokun:
The #718 write-fencing gate is **correct and untouched**. `requiresWorkflowStampEnforcement` is byte-identical, and the refusal still fires on `!enforcesWorkflowStamp || !enforcesWorkflowStampAtWrite || !hasTrustedStamp`. Only the *recovery text* built inside that branch changed.

What was broken is the gate's escape hatch. Every refusal said "Run `install_panel(action:'update')`", and `install_panel` could not actually perform that update in most real deployment shapes. A hard gate whose only remedy is broken leaves users with no path forward — that is the bug.

Closes #771
Closes #766
Closes #769
Closes #774
Closes #784

---

## 1. Never recommend a tool the caller cannot invoke (#774, #784)

New `src/services/panel-recovery.ts` is the single source of the advice rendered by the bridge refusal, `install_panel`'s status note, the sync assessment, and `panel_update_node`'s redirect — so they cannot drift apart and contradict each other.

| session | what the refusal now says |
|---|---|
| **local** | Names `install_panel(action:'update')`, **and** carries the host-side commands as the alternative ("or, if install_panel is not in this session's tool set, …"). Never a single point of failure. |
| **remote** (#774) | Does **not** name `install_panel` at all. "Update the panel ON THE COMFYUI HOST — this session targets a REMOTE ComfyUI, so the panel pack is on that host's disk…", followed by real commands. |
| **cloud** | Same, naming Comfy Cloud's lack of a writable `custom_nodes`. |
| **embedded panel** (#784) | Covered by the above — both reporters were on remote hosts. See "Left out" for the one axis that is not mechanically knowable. |

The commands cover **both install shapes** (`git pull --ff-only`, and a clone-and-swap when there is no `.git`), and put the backup **outside** `custom_nodes` — ComfyUI serves every directory in there, so a copy parked beside the panel would shadow it and make a real fix look like a no-op (#641).

`panel_update_node`'s "Use install_panel instead" and the pin refusals' "clear it with `install_panel(action='unpin')`" got the same treatment, as did the sync summaries and the auto-sync failure notice that are pushed into the embedded panel chat.

## 2. Root cause of the status/update contradiction (#771)

`install_panel(action:'update')` delegates to the **generic** node-pack update, whose post-op presence gate (`assertPackPresentAfterOp`, #730) asks ComfyUI-Manager's installed/registry list — keyed on the registry id — and **never looks at the disk**. `panelStatus` asks a completely different question: it scans `custom_nodes` and reads `pyproject.toml`.

For a Comfy Registry **zip** install those disagree, and the one that sounded certain was wrong:

> `"comfyui-agent-panel" was queued for update but is not present afterward — it is **not installed locally** and was not found in the ComfyUI-Manager registry`

…about a pack `status` had just reported at a concrete `dir` and `installedVersion`. It also read a *failed* list request as proof of absence (`.catch(() => [])`).

Fixed on both sides:
- The panel path holds the authoritative on-disk answer, so a Manager failure is recorded as "the Manager could not update it" and routes to the verified fallbacks. It propagates only when the disk **agrees** the pack is absent — and not when that reading came from a scan that could not complete.
- The generic gate's own message now states what it actually checked, and an unreadable list refuses as *unverifiable* rather than *absent*.

## 3. A verified update path for the registry-zip shape (#771)

No `.git` means nothing to fast-forward (#724) and nothing the Manager resolves. `updateViaRegistryZipReinstall` performs the reporter's own workaround, under verification:

- **Never on a checkout.** The absence of `.git` is *proven* with a tri-state probe (`existsSync` collapses `EACCES` to false), covering both shapes `.git` takes — a directory, or a file in a worktree/submodule.
- **The replacement is validated before anything moves**: panel identity, a readable version, the built `web/js` bundle (a source-only tree would silently *uninstall* the panel), and both versions strictly comparable — never backwards, and equal reports "already current" without touching the disk.
- **The backup leaves `custom_nodes`**; staging is a sibling of it, never inside.
- **The swap is crash-recoverable by ordering** — see "The swap, and a documented residual" below. Recovery is derivable from the filesystem, not from a journal, and refuses to mutate when it cannot establish what happened.
- **Success still requires proven on-disk movement, re-read afterwards**, plus a clean shadow scan (#639/#641 discipline preserved throughout).

## 4. Target resolution (#766, #769)

New `src/services/panel-workspace.ts` resolves **live-first**, mirroring `resolveInstallInterpreter`'s precedent: the running server's `--base-directory` (which is where Comfy Desktop keeps `custom_nodes`, #766), then its own argv root (#769), then the configured path. A live base is accepted only when it really holds `custom_nodes`.

The base is **frozen for the whole operation** — priming alone was not enough, since the cache expires and a Manager call can outlive it, which would have let the pre-op and post-op reads describe different trees.

Because "reachable" is not "the server chose this tree", a **live-derived** root is required before any direct mutation (both the ZIP swap and the #724 fast-forward) and before the stale-bundle diagnosis. `PanelStatus` gained `comfyuiPath`, `baseSource`, and `absenceProven` — `installed: false` reads as authoritative and a prose caveat does not travel, so an unproven absence is now structured, blocks the sync, and stops the unattended auto-install.

`workspace(get/list)` also surface the live root (`workspace_source: "live-server"`, `transient: true`), so they stop contradicting `get_environment`.

## 5. Stale browser bundle vs stale install

Per the live-rig finding: an up-to-date panel *does* advertise both capabilities, they come from `js/lib/session-rebind.js` — which also builds the `hello` — and the panel's module URLs carry no cache-busting key. So a tab holding that one file from before 0.11.35 announces the old capability set while the pack on disk is current. The user is told to update, and the update correctly reports nothing to do. That is the unfixable-looking loop.

Both facts were already collected at the same moment (the orchestrator runs the panel sync on every hello, which reads the installed version off disk). `panelStatus` now records that reading and the bridge resolves a skew from it:

> **Do NOT update the panel** — the pack ON DISK is already 0.11.38, which meets the 0.11.35+ this MCP requires. This BROWSER TAB is running an older cached copy of the panel's JavaScript (it advertised 0.11.34)… HARD-REFRESH this ComfyUI tab with a cache-bypassing reload (Ctrl+Shift+R…).

Emitted **only on positive proof**, because the wrong direction sends a genuinely-behind user back round the loop. It requires a recent observation bound to the current target *and* generation *and* live-derived base, re-reads the `pyproject.toml` at the moment of use, and requires a clean shadow scan (a shadow means a refresh reloads the *shadow*). Every unknown falls through to ordinary update guidance. Cache-busting itself (#584 / panel #596) is deliberately **not** touched.

## Self-gate

20 rounds of `codex exec -m gpt-5.6-terra` adversarial review plus seven maintainer gate rounds, scoped to: can success be reported without verified on-disk movement; can a recovery message name an uncallable tool; can status and update still disagree; was the #718 gate weakened. Every finding is fixed and carries a regression test. Among them:

- a Manager-induced **downgrade** reported as an update — first fixed on one branch, then moved into `classifyPanelUpdate` (the single place a version pair becomes a verdict) after the maintainer pointed out that a guard on one branch of a two-branch decision isn't a guard, and finally extended to `install`/`reinstall`, which carry their own movement test;
- a swap that could **strand the user's panel**, recoverable only if the user knew to run another *mutation* — repair is now automatic on load and on `install_panel(action:'status')`;
- a base cache **stamped with the wrong target** because the identity was read *after* the await;
- `.git` absence **inferred from a failed probe** (`existsSync` collapses `EACCES` to false);
- a status report asserting *"your existing panel is intact"* from a bare `existsSync` — a confident, wrong reassurance in the one message a worried user actually reads.

One recurring defect class accounts for most of them: **"could not determine" being folded into a positive answer.** An unreadable file, a failed probe, a missing optional dep, an unenumerable directory, an absent record — each was at some point treated as a pass. The four-question split (identity / serve / shape / integrity), each with exactly one predicate, exists because conflating them is how those folds got in.

Codex confirms across rounds: **the #718 gate is not weakened**, no ComfyUI-Manager queue-drain-only success survives, no remote/cloud path recommends `install_panel`, no pin bypass, and no import cycle. It also confirms the stale-bundle detection has **no false-positive path** — the direction that would send a genuinely-behind user back round the loop.

`npm run build` clean; **4153 tests pass** (112 new, in `panel-recovery-cluster.test.ts`, plus bridge-level integration tests).

### Refuse vs disclose

Recorded in the code, because it decided several of these and generalises:

- **Refuse** when the action has **not yet happened** and proceeding without certainty would be unsafe. A refusal *prevents* something — unproven absence, an uncorroborated tree, an unverifiable staged copy, a pin.
- **Disclose** when the action is **already complete** and only its *characterisation* is uncertain. Nothing is left to prevent; the duty is to describe honestly.

Reporting failure for work that succeeded is a fabrication too — and here the worse one, since an error invites the user to re-run a destructive swap to "fix" something that already worked. That is why an uncomparable version change (`0.11.34` → `nightly`, which real Registry nightly installs produce) reports *what changed and that the direction is unknown* rather than throwing.

### Open reviewer disagreement — not forced to green

Codex holds one contested P2 and therefore reports FAIL. Recording it rather than massaging the gate:

> `panel-sync`'s summaries name `install_panel` in **local** mode while being pushed to the embedded panel chat, which does not carry that tool.

The premise doesn't hold. The orchestrator builds an in-process MCP client over the **full** tool surface for its `call_tool` route (`orchestrator/index.ts:680`), so `install_panel` genuinely *is* reachable from a local panel session with a tool-calling agent. What isn't reachable is the mobile/canvas-less `CALL_TOOL_WHITELIST` — a different lane. Treating "not in one lane" as "not callable" is the same absence-of-evidence shape this PR spends its length fixing, applied to a review claim instead of to code.

So the answer differs by lane, and the composer often can't tell which lane its text lands in. Naming the tool **and** giving the host commands means both readers get something actionable and neither dead-ends — the right bias when the distinction is genuinely unknowable at composition time. Suppressing the tool name on a guess would withhold the working remedy from the local callers who do have it.

**Maintainer ruling: keep as is.** If it later becomes cheaply distinguishable — the composer knowing it is emitting into the canvas-less lane specifically — suppress the tool name there and keep the host commands. Not worth engineering a signal for now.

## The swap, and a documented residual

Replacing a directory takes more than one rename (renaming onto a non-empty directory fails on every supported platform), so there is no atomic version. The ordering therefore guarantees the next best thing — at every interruptible point, **some** panel is present under `custom_nodes`:

```
1. new panel -> custom_nodes/.comfyui-agent-panel.incoming   (both present)
2. old panel -> custom_nodes_backup/...                      (incoming served)
3. incoming  -> custom_nodes/comfyui-agent-panel             (done)
```

After step 1 the old panel still serves; after step 2 the incoming one does — ComfyUI serves dot-prefixed directories, the same fact behind #641, working in our favour. The `.incoming` directory **is** the evidence of an interrupted swap, so recovery is derivable from the filesystem and needs no journal. An earlier revision relied on a journal file, which was wrong: `fsync`ing a file's *contents* does not make its *directory entry* durable.

Recovery revalidates before it acts, and the check is **exact**. A crash can retain the `.incoming` directory entry while losing file data that was still unwritten, so at stage time the swap writes an **integrity manifest** into the tree — relative path, byte size and SHA-256 for every file, plus a digest over that list — and recovery re-derives and compares. Three outcomes, acted on differently: *intact* → promote; *corrupt* → restore the backup; *unverifiable* → refuse and disclose.

An earlier revision tried to answer this by inspecting the bundle's JavaScript for syntactic completeness. That was removed: syntactic validity is not decidable in practice with a hand-written lexer (ASI, regex-vs-division, CR-only line endings, HTML-like comments — the rules never stop), and every rule added was a fresh way to be wrong in **both** directions, refusing a legitimate install or discarding somebody's only working panel. Comparing bytes to what we ourselves staged cannot false-reject a valid bundle, and is strictly stronger — it catches a file that is syntactically perfect but is the wrong half of another file.

**Residual, deliberate.** If the `.incoming` directory entry is lost while `old → backup` survives, the panel is recoverable only manually. That state is **indistinguishable on disk** from a completed swap whose journal outlived it followed by a deliberate uninstall — so restoring automatically would be acting on evidence we do not have, and would silently resurrect software the user removed on purpose. Refusing is the chosen error, because it is visible and recoverable while the other is neither.

To be precise about the likelihood: "journaling filesystems commit metadata in order" is **not** a durability contract — on NTFS especially, ordering across separate operations is not guaranteed without an explicit flush. This case is not exotic. The decision does not rest on it being rare; it rests on the two states being indistinguishable.

What makes refusing acceptable is that the refusal hands the user their data back. It names: that no panel is present at the canonical path; that a backup exists **and its absolute path**; that we are deliberately not restoring because we cannot tell an interrupted swap from a deliberate uninstall; the exact `mv` command to restore it; and the exact route to reinstall fresh. A test asserts the backup path appears in that message.

## Left out, deliberately

- **A remote-safe *verified* sync (#774).** Not attempted: we cannot read the remote host's disk, so any "success" would be exactly the unverifiable claim #639/#641 exist to prevent. #774 gets the honest half — a precise refusal plus the exact host commands — which the issue names as an acceptable resolution.
- **Disk evidence in the *generic* Manager presence gate.** #771's root cause is fixed for the panel and remains live for every other pack hitting the registry-ZIP shape: only the panel-specific wrapper supplies an on-disk answer. The important half is done — that gate can no longer assert an absence it never checked (it now states it consulted only the Manager's list, and refuses an unreadable list as *unverifiable* rather than *absent*) — but supplying disk evidence for other packs is a separate change, filed separately.
- **`panel_reload_frontend`** (#771's secondary ask). The diagnosis and instruction are now correct, but the agent still cannot perform the hard-refresh itself.

---

## The vocabulary this took four rounds to reach

Recorded here because it is the durable output of the last stretch, and because the next change in this file will want to collapse it back into one word. It is documented on `PanelIdentityVerdict` in the source too.

**"Could not read it" is not one answer. It is three, with three different remedies** — the identity test walks three steps, and each can fail in a way the others cannot:

| verdict | what actually happened | what it points the user at |
|---|---|---|
| `probe-failed` | The `statSync` behind `probeFile` did not complete. We do not know whether the file is even **there**. **Nothing was opened.** | The directory holding it, or the filesystem — a bad mount, a path that changed underneath us, a directory we cannot traverse. |
| `unreadable` | The probe said the file **is** there, and **opening it failed**. | That file — its permissions, or an I/O error. |
| `malformed` | It opened and read fine. The **contents** would not parse. | Repairing or replacing the file. Not a permissions problem in any form. |

All three are **fail-closed for the DECISION** — `dirIsPanelByIdentity` answers `false` for every one of them, and `dirHasPanelFiles` refuses on all three. They differ *only* in what the user is told to go and do, which is the entire reason they are separate values rather than one.

Two of the three were found by review *after* being collapsed: first a parse failure narrated as "could not be read", then an indeterminate stat narrated as a failed open with "check permissions on that file" attached — an open that was never attempted. Both were this PR's own defect class (a value computed for one purpose read as the answer to a different question) arriving inside the vocabulary introduced to fix it.

The same three-way split applies to the bundle probe in `panelShapeVerdict`.

### And in the SHAPE verdict, `unreadable` is a *bucket* — one of whose members is not a filesystem failure at all

This is the collapse the note exists to prevent, so it gets named explicitly. `panelShapeVerdict` reaches `unreadable` two ways:

1. an open that genuinely **failed** (`fileDigest` returned nothing), and
2. a dep set carrying **no `fileDigest` at all** — where nothing was opened, because there was no way to ask.

(2) is a dep-shape gap, not a filesystem event. It stays `unreadable` on purpose: a fourth verdict would put a **non-observation into a vocabulary about what was seen on disk**, and the next reader would reasonably try to produce it from a probe. Production always wires the dep, so nothing on disk was ever observed to be wrong here.

The cost is paid in the **sentence** instead. `describePanelShape` makes no filesystem claim for `unreadable` — "one of its files was located, but its contents could not be checked" — because that is what is true of *both* members, and the renderer cannot tell them apart. Review caught the earlier wording ("is there but could not be opened") asserting an open failure that, on path (2), never happened.

**The identity verdict is not a bucket.** `panelIdentityVerdict` never consults `fileDigest`, so *its* `unreadable` has exactly one member — a real open failure — which is why it keeps the precise cause and the permissions remedy while the shape verdict's does not. Same word, different membership, deliberately different sentences.

The general rule this arrived at, and the one worth carrying forward: **the bucket a value falls into is not the sentence it earns.** First the fail-closed boolean was separated from the speaking verdict; then the speaking verdict from the claim.


artokun:
### Gate round 2 — your three P1s are closed, plus five more they surfaced

Pushed through `87e6ac7`. Full suite **4112 passing**, build clean.

**(3) Interrupted swap — the panel now comes back on its own.** A single atomic rename isn't available (renaming over a non-empty directory fails on both POSIX and Windows), so this took your documented fallback: automatic repair. `ensureInner` reconciles on every load (already under the op lock), and a new `repairInterruptedPanelSwap` — which takes the lock itself after two cheap lock-free pre-checks — runs on `install_panel(action:'status')`, because "what's going on?" is exactly when a user hits this. A bare `panelStatus` still only *reports*: it holds no lock and must not cut across another process's in-flight swap.

Two follow-ons the gate then found on that new code: the repair **bypassed the pin** (a restore is a mutation — status now reports the pinned interrupted state instead), and the journal was written with a **buffered** `writeFileSync` immediately before the first rename, so a power loss could persist the rename while losing the record. It's now written through a descriptor and `fsync`'d, with a best-effort directory fsync.

**(2) Target binding.** Generation is captured *before* the freeze, and a re-pin keeps the first value — `runPanelActionInner` re-pins on entry, and re-reading there would have erased exactly the window being guarded. `swapTreeIsCorroborated` now requires the current resolution's base to *equal* the frozen base, not merely be live-derived. Round 2 found the automatic paths uncovered (`repairInterruptedPanelSwap`, on-load ensure) and a gap after the last check: `performPanelSync` does another status read and publishes its own success, which the orchestrator pushes into the panel chat — a retarget there would announce a verified sync of tree A to a tab that is now B. All now share an exported `assertPinnedTarget`.

**(1) Downgrade.** Moved into `classifyPanelUpdate` as a `"downgraded"` outcome, so it holds regardless of which path produced the version pair, and it short-circuits ahead of the no-op/zip fallbacks so a Manager-induced regression is surfaced rather than papered over. Round 2 found `install`/`reinstall` have their own movement test with the same hole — also fixed. One deliberate exception: `install_custom_node(version='0.11.20')` is redirected here *precisely so the caller gets what they asked for*, so an explicit request for the version that landed is honoured and labelled "was DOWNGRADED, as explicitly requested".

### Still not PASSing — one item, restated precisely

Codex holds that `panel-sync`'s summaries still name `install_panel` in **local** mode while being pushed to the embedded panel chat. I've held my position and am flagging it rather than forcing it:

The orchestrator builds an in-process MCP client over the **full** tool surface for its `call_tool` route (`orchestrator/index.ts:680`), so `install_panel` genuinely *is* reachable from a local panel session with a tool-calling agent. What isn't reachable is the mobile/canvas-less `CALL_TOOL_WHITELIST` channel, which is a different lane. So "is `install_panel` callable here?" has no single answer in that process, and the local text names it **and** gives the host-side commands so it cannot dead-end either way. Suppressing the tool name on a guess would withhold the working remedy from local callers who do have it.

If you'd rather bias the other way for panel-chat-destined text specifically, that's a one-line change to `UNPIN_INSTRUCTION` / `describeInstallPanelAction` — your call.

### Latent issue

Yes please, **file it**: the generic Manager presence gate still doesn't inspect disk, so #771's root cause is fixed for the panel and live for every other pack hitting the registry-ZIP shape. My change made that gate's *message* truthful (it now says it only consulted the Manager's list, and an unreadable list refuses as unverifiable rather than absent), but only the panel-specific wrapper supplies disk evidence.

artokun:
### Your P0/P1×2/P2 are fixed — and I need a ruling on one residual

Pushed through `743f939`. **4123 tests passing**, build clean. My re-gate then found six further defects across three rounds, all fixed with tests.

**P0 — swap redesigned so the guarantee comes from ordering, not from a journal.** You were right that fsync'ing the journal's *contents* doesn't make its *directory entry* durable. New order:

```
1. new panel -> custom_nodes/.comfyui-agent-panel.incoming   (both present)
2. old panel -> custom_nodes_backup/...                      (incoming served)
3. incoming  -> custom_nodes/comfyui-agent-panel             (done)
```

After step 1 the old panel still serves; after step 2 the incoming one does — ComfyUI serves dot-prefixed dirs, the same fact behind #641, working for us. The incoming directory *is* the evidence of an interrupted swap, so recovery is filesystem-derived and needs no journal. Test included as you asked: a real swap interrupted between renames **with all cleanup also failing** (a power loss tidies nothing), asserting a panel is still reachable and that recovery works **with the journal deleted**.

**P1 stale journal** — a journal with no incoming directory is UNKNOWN: reports, never mutates. Repair was not made more eager; the trigger moved to the marker that means what it says. **P1 sync early return** and **P2 invisible recovery** also fixed (`recoveryNote` rides on success *and* failure, appended in place so the original error class survives).

My re-gate then caught, in my own new code: a rollback that deleted the only served panel; repair visibility restricted to `PanelInstallError` while install/reinstall reject with Manager errors; the mutation-path repair not bound to the frozen target; the post-swap check rejecting only an *unchanged* version, not a regression; `"canonical exists"` treated as `"canonical works"` (a husk could license discarding the only good copy); and a step-2 failure deleting in place instead of moving out first.

### Residual, and it needs your call

Codex now raises a P0 that **contradicts your stale-journal P1 ruling**, and the two are mutually exclusive on identical filesystem state.

Its claim: if the `.incoming` directory entry is lost while `old → backup` survives, recovery sees no incoming and no canonical, and — per your ruling — refuses to restore from the journal, recreating the no-panel outage.

Why I did not act on it unilaterally: those two states are **indistinguishable** from disk. `journal + backup + no canonical + no incoming` is *both* "lost the incoming entry in a crash" and "swap completed, journal outlived it, user then uninstalled deliberately". You ruled the second must not be resurrected. Reversing that to satisfy this P0 would reintroduce exactly the defect you flagged.

I also think codex's premise is weaker than it looks: **both** operations are now renames in the same directory, and journaling filesystems (NTFS, ext4, APFS) commit metadata in order — a later rename cannot persist while an earlier one is lost. That is materially different from the journal *file*, whose page-cache contents genuinely could be dropped while later metadata persisted. That asymmetry is the whole reason the ordering fix is better than the journal was.

What I did instead of choosing: added a barrier between steps 1 and 2 — the staged copy must be **visible** before the working panel is moved aside (never act on a rename that did not take), plus a best-effort directory sync where the platform offers one. Codex's own suggested fix — *refuse the swap where durability can't be guaranteed* — would disable the #771 fix on Windows, which is the reporting platform.

**Options if you want it closed:** (a) keep as is, accepting the residual on non-ordering filesystems; (b) restore from a journal only when it is *recent* (it records `startedAt`), trading a heuristic for coverage; (c) something you see that I don't. I'd take (a), but it's your ruling to make.

**P2, not fixed:** with no incoming and the canonical panel present, reconciliation silently deletes the spent journal. It moves no panel, so I read it as housekeeping rather than a policy breach — say the word if you want it reported instead.

Issue #796, opened 2026-08-03
Recurring defect class: "could not determine" folded into "determined not" — 7 instances in 5 PRs in one day


artokun:
Across a single day of review, the **same defect** appeared **seven times in five unrelated PRs**, in code written by four different agents. It is worth naming so it can be checked for deliberately rather than rediscovered each time.

## The pattern

**A state meaning "I could not determine this" is folded into a state meaning "I determined it is not so."**

Absence of evidence becomes evidence of absence, and the result is always the same: a confident wrong answer, reported to the user as fact.

## The seven instances

| where | "could not determine" | got treated as |
|---|---|---|
| panel #511 | `getNodeDefs()` **timed out** | the node-type baseline is permanently untrustworthy → session-long false refusal |
| panel #511 | five requests **rejected/aborted** | the baseline "finished empty" → irreversible latch |
| #787 | HTTP body **unparseable** (`SyntaxError`, no status) | server **unreachable** → re-enabled a guessed local directory for a server that had *refused* |
| #793 | Manager list request **failed** (`.catch(() => [])`) | pack **not installed** → "it is not installed locally", said by code that never looked at disk |
| #794 | live-root probe **failed** (`undefined`) | **no contradiction** → rendered "downloaded successfully" after an unverified server swap |
| #785 | `settings.json` **inaccessible** (`existsSync` false) | component **absent** → restart without Git → Manager aborts at import → server down |
| #785 | `findPidByPort()` **lookup failed** (`null`) | port is **free** → teardown committed under a live server |
| #785 | parent-chain walk **hit its hop limit** | process is **foreign** → legitimate wrapper launch reported as failed |

(Eight rows; #785 contributed three.)

## Why it keeps happening

Almost always a **two-valued return carrying three states**. `null`, `false`, `undefined` and `[]` are each pressed into meaning both "no" and "don't know", usually because the failure path was written as a convenience:

```ts
.catch(() => [])            // "failed" becomes "empty"
if (!existsSync(p)) return  // "inaccessible" becomes "not there"
return null                 // "lookup failed" and "port free"
```

It survives review because the happy path is correct and the tests exercise real answers, not failed observations. It survives *self*-review because the author knows what they meant.

## What to do

**1. When adding or touching a predicate, ask: what does this return when the underlying observation fails?** If that is the same value it returns for a genuine negative, it is this bug.

[...]

PR #802, opened 2026-08-04
fix(panel): rebind the agent tab across id migration and keep the release fallback starvation-free (#568)

### LINKED ISSUE #568: Panel wedges after tab-id migration: PanelAgent.tabId left stale (panel_* → \"no connected tab\"), then an interrupt storm starves the 1500ms release fallback and no turn ever starts again
Two defects, observed back-to-back in one session. The first makes the agent *look* wedged; the user then hammers "send now", which triggers the second and wedges it for real. Filing together because the second is only reachable in practice via the first.

## Defect 1 — tab-id migration rebinds the manager key but leaves `PanelAgent.tabId` stale

At 19:32:54 a new tab connected and the orchestrator migrated:

```
19:32:54.638 [ui-bridge] panel tab connected: tmp:7ca3 (“Unsaved Workflow (6)”) — 2 tab(s) total
19:32:54.640 [panel-orchestrator] tab-id migration: wf:workflows → tmp:7ca32bc6 — agent rebound, conversation preserved
```

**But every single `[panel-agent …]` line in the log — before AND after the migration, right through to the end of the session — is prefixed `tmp:69c6`**, an id from 07:40 that morning:

```
19:32:54.683 [panel-orchestrator] tab tmp:7ca3 → agent: …      <- inbound routed to the NEW id
19:33:37.666 [panel-agent tmp:69c6] turn done (subtype=success) <- the agent still thinks it is the OLD id
19:35:15.129 [panel-agent tmp:69c6] init …
19:36:11.034 [panel-agent tmp:69c6] init …
```

So the migration rebinds the **manager's map key** (inbound messages arrive fine) but never updates the agent instance's own `tabId` field. Anything that routes by `this.tabId` / `ctx.tabId` keeps pointing at a dead tab.

Confirmed independently from the tool side — `panel_new_workflow` failed twice with exactly that stale id:

```
Error: no connected tab with id "tmp:69c60eac-3cae-4504-adf3-e6dc87b635cb".
Connected: tmp:5fda (“Unsaved Workflow”), wf:workf (“Untitled 2026-07-31 00-38-34”)
```

Note the id in the error is `tmp:69c6…` — the agent's stale field — while the connected tabs are entirely different. `panel_set_workflow_target { mode: "current" }` recovers the **tool routing** (it replied "Rebound this session from tab tmp:69c6 onto the active tab wf:workf"), which is a useful escape hatch, but it does not fix `PanelAgent.tabId` itself, so the staleness persists for everything else that reads it.

Likely fix: `rebindAgent(oldKey, newKey)` (`src/orchestrator/panel-agent.ts` ~L1527) re-keys `this.agents` but must also assign the agent's own `tabId` (and re-point anything derived from it — `deps.onTurn(this.tabId, …)`, bridge pushes, the panel MCP session binding). Worth auditing every `this.tabId` read in `PanelAgent` for the same assumption.

## Defect 2 — rapid interrupts starve the interrupt release fallback; the turn gate never reopens

With the agent apparently unresponsive, the user pressed Stop / "send now" repeatedly. The log shows the failure mode precisely:

```
19:35:29.129 tab tmp:7ca3 interrupted (send-now: re-queue)
19:35:29.136 [panel-agent tmp:69c6] turn done (subtype=error_during_execution)
19:35:29.139 [panel-agent tmp:69c6] init …            <- still recovering correctly
19:35:30.134 tab tmp:7ca3 interrupted (send-now: re-queue)
19:35:30.137 turn done (subtype=error_during_execution)
19:35:30.141 init …
19:35:31.009 tab tmp:7ca3 interrupted (send-now: re-queue)
19:35:31.014 turn done (subtype=error_during_execution)
19:35:31.019 init …
19:35:39.988 tab tmp:7ca3 interrupted
19:35:39.991 turn done (subtype=error_during_execution)
19:35:41.926 tab tmp:7ca3 interrupted (send-now: re-queue)   <- from here on:
19:35:42.960 tab tmp:7ca3 interrupted (send-now: re-queue)   <- NO turn done,
19:35:48.068 tab tmp:7ca3 interrupted (send-now: re-queue)   <- NO init,
19:35:49.575 tab tmp:7ca3 interrupted (send-now: re-queue)   <- nothing.
```

After 19:35:41 the interrupts land but **no turn ever starts again** — the panel sits on "Interrupted." with 2 PENDING messages that no amount of send-now will flush (screenshot attached by the user; `PENDING · 2`).

Suspected cause, from `interrupt()` / `armInterruptReleaseFallback()` in `src/orchestrator/panel-agent.ts`:

- `interrupt()` deliberately does **not** force the turn gate open; it relies on the aborted turn's `result` → `completeTurn()` to release it, with `armInterruptReleaseFallback(interruptedTurn)` as a 1500ms safety net (`INTERRUPT_RELEASE_FALLBACK_MS = 1500`).
- `armInterruptReleaseFallback` starts with `if (this.interruptReleaseTimer) clearTimeout(this.interruptReleaseTimer)` — **each new interrupt cancels the previous fallback** and re-arms with a *new* `guardTurn = this.yieldedTurns` captured at that moment.
- The fallback then fires only `if (this.completedTurns < guardTurn)`.

Interrupts arriving closer together than 1500ms (29.1 → 30.1 → 31.0 are ~1s apart) therefore keep cancelling the pending safety net before it can fire. If the newest `guardTurn` no longer corresponds to the turn actually holding the gate, the `completedTurns < guardTurn` guard is false and the release never happens for the turn that matters — gate closed permanently, queue never drained. The "safety net that can be indefinitely postponed by the very event it guards against" shape fits the log exactly.

Suggested direction (not implemented — reporting only, per the user):
- Don't cancel-and-replace unconditionally. Either keep the **earliest** outstanding guard, or track the guard set and release when *any* outstanding guard's turn is still incomplete.
- Add an absolute ceiling from the FIRST interrupt in a burst, so a stream of interrupts cannot postpone the release past a fixed bound.
- Consider coalescing/debouncing interrupts at the orchestrator (`event.type === "interrupt"`), which would also stop the storm at the source.
- A wedged-gate escape hatch would help operationally: if the queue is non-empty and no turn has started for N seconds, force-release and log it loudly.

## Steps to reproduce

1. In the panel, have a live agent turn running.
2. Cause a tab-id migration — e.g. have the agent call `panel_new_workflow` (a new tab connects; the orchestrator logs `tab-id migration: … — agent rebound`).
3. Observe: subsequent `panel_*` tool calls fail with `no connected tab with id "<old id>"`, and every `[panel-agent …]` log line still carries the pre-migration id.
4. With the agent seemingly stuck, press Stop / "send now" several times ~1s apart.
5. Observe: interrupts stop producing `turn done` / `init`; the panel shows "Interrupted." with PENDING messages that never flush. Disconnect/reconnect is the only recovery the user found.

## Confidence

Defect 1 is **confirmed** — the stale id appears in both the log prefix and verbatim in a tool error, and the workaround's own message names it.
Defect 2 is **strongly indicated** by the log (interrupts after 19:35:41 produce no turn done / init at all) plus the cancel-and-re-arm read of `armInterruptReleaseFallback`, but I did not build a controlled repro to prove the guard-mismatch path specifically. Treat the mechanism as a hypothesis and the symptom as fact.

## Not the cause (recorded so it isn't chased)

This was initially reported as "wedges while looking at the Captured workflow image" — the capture card was simply the last thing rendered before the symptoms appeared. `panel_screenshot` returned normally (1280×847) and two turns completed successfully *after* it (19:33:37, 19:34:14). The trigger was the tab-id migration at 19:32:54, not the image.

## Environment

- GPU: NVIDIA RTX 4090, 24 GB VRAM
- ComfyUI: 0.29.2 (local), Manager v4
- **comfyui-mcp: 0.48.21**
- **panel: 0.11.20**
- Backend: Claude (claude-opus-5), session 8c44273a

artokun:
Closes #568

Both halves of #568 got a first-pass fix in 4e8888b (already on main). The issue
stayed open because of the recurrence reported on **0.49.3 / panel 0.11.38**: with two
panel tabs connected, saving the active workflow under a new name migrates the tab id,
"the retiring id is removed, but the last-active selection remains the retired id", and
`panel_set_workflow_target({mode:"current"})` "cannot recover because neither of the two
live tabs is selected as last active". This finishes it.

## Root cause (a) — the migration carried everything about the tab except which tab was selected

A same-socket tab-id migration already moves a great deal: the agent and its own
`tabId`, the bound panel MCP server, the run-completion journal (`moveKey`), pinned
workflow targets, held mail, the stable resume key, mirror viewers and subscribers. It
did **not** move the bridge's `lastActiveTabId`.

Left on the id being retired, that selection names a tab that no longer exists. With
2+ tabs connected, `resolveTarget()`'s no-tabId path and `resolveActiveTabId()` both
fall through to "none is last active" — which disables
`panel_set_workflow_target({mode:"current"})`, the one documented escape hatch for a
stale binding. So the session cannot be recovered from the agent side at all, which is
exactly the reported symptom.

**The correct id is recoverable, not guessable.** The migration is a *same-socket*
re-hello, and the socket is server-observed — that is the proof that the new id and the
old id are the same browser tab. The selection now moves with it.

Like the mirror move it sits beside, the carry is **optimistic** (the orchestrator
classifies proven-vs-switch only afterwards), so `revokeTabMigration()` undoes it —
the same proven-carries / unproven-forgets split PR #786's `moveKey` established.

The undo compares a write **sequence**, not a value. A `user_message` from the
destination after the carry writes the same id, and value equality alone would revert a
genuine selection.

## What happens when the correct tab cannot be established

It **refuses, and stays refused.** On an unproven switch the undo sets the selection to
`null` — it does **not** hand it back to the retired id, and it does **not** leave it
pointing at the switched-to workflow. Unknown gets its own representation, so
`resolveTarget()` raises `Multiple panel tabs are connected and none is "last active" —
pass tab_id` and an orphaned session's rebind fails rather than silently retargeting
onto a different workflow's canvas. Un-wedging by binding to whatever tab happens to be
live would be worse than staying wedged; a test asserts specifically that the
resolution does **not** come back as the switched-to tab.

The silent auto-heal (`ensureReachable`) is untouched: it stays strict-single and never
consults the last-active selection at all.

## Root cause (b) — the release fallback, and what is actually still reachable

The reported starvation (`clearTimeout` + re-arm on *every* interrupt, so a burst
spaced under the window kept cancelling the net) was fixed in 4e8888b. I re-derived it
rather than taking it on trust, and it holds: the arm **coalesces**, so the deadline set
by the first interrupt of an unsettled burst is an absolute ceiling, while
`interruptGuardTurn` is refreshed by every interrupt so it keeps naming the still-stuck
turn. A storm can neither postpone the release nor aim it at the wrong turn.

The invariant that makes that starvation-free rather than merely eventually-scheduled is
now written down at its single seam: **the fallback is only ever cleared at a point
where the gate is (or is about to be) open**, so "an interrupt happened and the gate is
still closed" implies a timer is armed. Any future clear site has to preserve it.

The existing regression test asserted only "a turn eventually starts", which passes in
both the fixed and the broken state once the storm stops. It is now **timed**: the
release must land within one window of the *first* interrupt while interrupts are still
arriving.

What is still reachable in production is the compounding the issue itself describes
("the second is only reachable in practice via the first"): an interrupt addressed to a
key with **no live agent** reaches nothing — no turn gate is held for it, and no release
fallback is scheduled. A recovery path that is never armed cannot be starved; it simply
does not exist. That was acked to the panel as a plain success, so the user keeps
pressing "send now" at a tab with nothing behind it. The outcome is now reported
(`interrupted` on the ack, read synchronously via `hasLiveAgent` so a hung
`backend.interrupt()` cannot withhold the ack) and a miss is logged loudly.

## Tests

All mutation-verified — the fix was broken each way and the test confirmed to fail:

| Test | Mutation that must break it |
| --- | --- |
| proven migration moves the selection; the explicit rebind resolves to the successor | drop the carry |
| unproven switch refuses and never names the switched-to tab | drop the undo |
| a fresh selection in the destination is not reverted | value-only compare instead of the sequence |
| a migration CHAIN follows; revoking a superseded hop is inert | drop the sequence guard |
| release lands within one window of the FIRST interrupt, however many follow | restore cancel-and-re-arm |
| interrupt reports whether a live agent took it | always return true |

Note the trap: for the refusal test, asserting only "throws" or only the tab count
passes in both states — the distinguishing assertion is that the resolution does not
come back as `wf:ccc`.

4212 tests pass, rebased on `origin/main` and re-run against the merged tree.

## Codex gate

`gpt-5.6-terra`, read-only, scoped to the three questions (wrong-tab binding, wedge with
no recovery, starvable release) plus chains, destination collisions and the
could-not/determined-not class. **PASS with no P0/P1/P2.** It confirmed the chain and
collision paths and that no unknown→none conflation was introduced; it could not run the
test suite in its sandbox (Vite child-process `EPERM`), so its verdict is a static audit
— the mutation testing above is mine.

## Deliberately out of scope

- **Identity-matched rebinding.** With 2+ live tabs, `rebindToActiveTab()` still binds to
  the last-active tab. That is the explicit, documented "use what's live now" consent
  signal, and this change only makes the selection it reads *correct*. Gating it on a
  workflow-identity match would need the identity of a tab id that has already been
  deleted from the live maps — i.e. it would have to treat "cannot determine" as
  "does not match" and refuse the legitimate reconnect recoveries (#322/#331/#332) the
  path exists for. That is the #796 bug in the other direction.
- **The panel-side PENDING tray.** The reporter's tray not draining is partly panel
  behavior (comfyui-mcp-panel); this PR makes the orchestrator's half of that signal
  honest but does not change the panel.
- **The Codex HTTP MCP session URL** (`/<tabId>`) still carries the pre-migration id and
  relies on the bridge's socket-scoped alias. That is the existing documented design
  (see `session-store.ts`'s proven-migration carry) and is untouched here.

### DIFF
--- a/src/__tests__/orchestrator/interrupt-storm.test.ts
+++ b/src/__tests__/orchestrator/interrupt-storm.test.ts
@@ -22,9 +22,10 @@ import type {
 import { CLAUDE_CAPABILITIES } from "../../orchestrator/agent-backend.js";
 
 let PanelAgent: typeof import("../../orchestrator/panel-agent.js").PanelAgent;
+let PanelAgentManager: typeof import("../../orchestrator/panel-agent.js").PanelAgentManager;
 
 beforeAll(async () => {
-  ({ PanelAgent } = await import("../../orchestrator/panel-agent.js"));
+  ({ PanelAgent, PanelAgentManager } = await import("../../orchestrator/panel-agent.js"));
 });
 
 /** A WEDGED backend: each turn hangs, and an interrupt breaks the hang (so the
@@ -133,5 +134,74 @@ describe("interrupt storm does not starve the release fallback (#568 Defect 2)",
     expect(backend.turns.join("\n\n")).toContain("the follow-up");
 
     await agent.stop();
+  });
+
+  // The property that makes the net starvation-free rather than merely "eventually
+  // scheduled": the deadline set by the FIRST interrupt of an unsettled burst is an
+  // ABSOLUTE ceiling. Later interrupts refresh the guard (so it keeps naming the
+  // still-stuck turn) but must never push the deadline out. Timing it — not just
+  // "does a turn eventually start" — is what distinguishes the two: a re-arming net
+  // still fires eventually once the storm stops, so a generous bound passes in BOTH
+  // states.
+  it("releases within ONE window of the FIRST interrupt, however many follow", async () => {
+    const backend = new WedgedBackend();
+    const agent = new PanelAgent("tab-ceiling", makeDeps() as never, backend);
+    void agent.start();
+
+    agent.send("wedged first");
+    await waitFor(() => backend.turns.length === 1);
+    agent.send("the follow-up");
+
+    const firstInterruptAt = Date.now();
+    void agent.interrupt({ requeueInFlight: true });
+    // Keep hammering INSIDE the window (50ms < 150ms) — each one is a chance to
+    // postpone the net.
+    const storm = setInterval(() => {
+      void agent.interrupt({ requeueInFlight: true });
+    }, 50);
+    try {
+      await waitFor(() => backend.turns.length >= 2, 1200);
+    } finally {
+      clearInterval(storm);
+    }
+    const elapsed = Date.now() - firstInterruptAt;
+
+    // One 150ms window plus scheduling slack. A net that any later interrupt could
+    // re-arm would land far beyond this (the storm ran for the whole wait).
+    expect(elapsed).toBeLessThan(600);
+
+    await agent.stop();
+  });
+});
+
+// #568 — the two defects compound: a stale identity means the interrupt reaches no
+// live agent at all, and an interrupt that reaches nothing arms NO recovery (no turn
+// gate is held for it, no release fallback is scheduled). Reporting that as a
+// completed cancellation is a could-not/did-not conflation, and it is what left the
+// user hammering "send now" at a tab with nothing behind it. The outcome must be
+// distinguishable.
+describe("interrupt reports whether a live agent actually took it (#568)", () => {
+  it("false when the key has no agent, true once one exists", async () => {
+    const backend = new WedgedBackend();
+    const manager = new PanelAgentManager({
+      mcpServers: {},
+      systemAppend: "",
+      model: "claude-test",
+      onSay: () => {},
+      onTurn: () => {},
+      makeBackend: () => backend,
+    } as never);
+
+    const key = "tmp:no-agent::claude";
+    expect(manager.hasLiveAgent(key)).toBe(false);
+    // Nothing was cancelled — and, critically, nothing was scheduled to recover either.
+    await expect(manager.interrupt(key, { requeueInFlight: true })).resolves.toBe(false);
+
+    manager.send(key, "spawns the agent");
+    await waitFor(() => backend.turns.length === 1);
+    expect(manager.hasLiveAgent(key)).toBe(true);
+    await expect(manager.interrupt(key, { requeueInFlight: true })).resolves.toBe(true);
+
+    await manager.stopAll();
   });
 });
--- a/src/__tests__/orchestrator/interrupt-storm.test.ts
+++ b/src/__tests__/orchestrator/interrupt-storm.test.ts
@@ -64,6 +64,41 @@ class WedgedBackend implements AgentBackend {
   }
 }
 
+/** A backend whose interrupt() HANGS until released — so stopAll()'s awaited stop()
+ *  never resolves and the closed agent stays in the manager's map indefinitely. That
+ *  unbounded window is the one where map presence and liveness diverge. */
+class HangingStopBackend implements AgentBackend {
+  readonly id = "claude" as const;
+  readonly capabilities = CLAUDE_CAPABILITIES;
+  turns: string[] = [];
+  private release: (() => void) | null = null;
+
+  async *run(opts: BackendStartOptions): AsyncGenerator<AgentEvent> {
+    yield { type: "session", sessionId: "sess-hang" };
+    for await (const turn of opts.channel) {
+      this.turns.push(turn.text);
+      yield { type: "result", ok: true, subtype: "success" };
+    }
+  }
+
+  async interrupt(): Promise<void> {
+    await new Promise<void>((resolve) => {
+      this.release = resolve;
+    });
+  }
+
+  /** Let the pending stop() finish so the test doesn't leave it dangling. */
+  releaseStop(): void {
+    const r = this.release;
+    this.release = null;
+    r?.();
+  }
+
+  async listModels(): Promise<ModelChoice[]> {
+    return [];
+  }
+}
+
 function makeDeps() {
   return {
     mcpServers: {},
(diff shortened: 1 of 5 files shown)


artokun:
Gate P2 fixed in `7e8e76c` — and you were right that it was in the ack I added on my own initiative, which is a fair thing to have caught.

**Map presence is not liveness.** `reset()`/`retire()` unmap through `unbindAgent()` before stopping, so `stopAll()` is the one path that leaves closed agents mapped: it sets `closed` synchronously on every agent and holds them in the map until each awaited `stop()` resolves — unbounded when a backend hangs. The bridge is only stopped after `await manager.stopAll()`, so the panel can land an interrupt in that whole window.

Both predicates now exclude a stopped agent:
- `hasLiveAgent()` returns `!!agent && !agent.isStopped`. LIVE means live, not mapped — correct for all seven call sites, since every one asks "is there an agent I can act on?" and a closed agent takes no message, runs no turn and accepts no interrupt.
- `manager.interrupt()` short-circuits **before** its await, so a hung `backend.interrupt()` can no longer stall the answer — the second half of the report.

Mutation-verified both ways. Restoring map presence fails the new test; dropping the `isStopped` guard in `interrupt()` makes it hang to the 5s timeout instead of answering, which is the "indefinitely if backend interruption hangs" case reproducing directly.

Codex gate re-run scoped at the widened blast radius (all seven `hasLiveAgent()` callers individually, the opposite false-negative via `restartAgentResume()`/`rebindAgent()`, and whether `closed` is set synchronously on every teardown path): **PASS, no P0/P1/P2.** Two answers worth recording since they were the ones I was least sure of:
- **3316** (arms a resume when no live agent) — tightening cannot arm a duplicate: `setResume()` independently rejects any mapped key, and the closing agent stays the map owner until `stopAll()` completes.
- **4552/4710** (which composite key owns a row/event) — cannot reroute to a different tab or provider: `liveKeys()` still counts mapped-but-closing entries, so a closing key alongside a live one stays ambiguous and returns `null`, and a lone closing key is rejected again before injection. No wrong-workflow path opens up.

It also confirmed no `#796`-style fold: `false` now means a synchronously determined absent-or-intentionally-closed agent, and a hung interrupt on an *open* agent is still reported as reached rather than as not-interrupted.

4213 tests pass, rebased on `origin/main` and re-run against the merged tree.

Issue #807, opened 2026-08-04
panel_query_graph: the groups[] membership lists are unbudgeted and spent before the node rows — #609 through a different door


artokun:
Continuation of #609, which fixed one half of this and left the other. Reported live by eicosaproton on a **690-node** workflow:

> The output keeps getting truncated by groups before reaching node details.

### What #609 fixed

`0de90c4` bounded `queryApiGraph` so one oversized blob could not starve the node you asked for, and `web/js/lib/graph-read.js` gained real per-node budgeting: per-widget caps with an `omitted` marker (`:132`), `clipLine()` for compact rows (`:140`), and `fitDetailLine()` which degrades an oversized detail row to an identifying stub rather than blowing the budget (`:150-165`).

All of that budgets **node and widget** content.

### What is still unbounded

Every `panel_query_graph` result carries `groups` — `id`, `title`, and each group's **full `member node_ids` list** — and that array does not appear to pass through any of the budgeting above. On a graph with hundreds of nodes spread across many groups, the membership lists alone are large, and they are spent before the node rows the caller actually asked for.

Net effect is the original #609 symptom through a different door: the answer to the question gets truncated to nothing by payload the caller did not ask for.

### Why it bites agents harder than humans

An agent calling `panel_query_graph {ids:[42], fields:"detail"}` has asked a precise question. Returning group membership for the entire graph ahead of that answer is both surprising and expensive, and the failure presents to the model as "this tool cannot show me node details" — the reporter's Kimi session concluded exactly that and asked for a workaround rather than retrying with a larger `max_chars`.

### Suggested

1. **Budget `groups` alongside everything else**, and spend it *after* the rows that answer the query — the requested nodes should never be starved by contextual extras.
2. **Cap `member_node_ids` per group** with the same omitted-marker convention already used for widgets, or omit membership entirely when the query targeted specific `ids` (the caller asking for node 42's detail does not need every group's roster).
3. Consider making `groups` opt-out, or opt-in when `ids` is present.

### Note

The immediate workaround is to raise `max_chars`, which does work — but the default should not require knowing that on a large graph, and an agent has no way to learn it except by failing first.

### Related
- #609 — the node/widget half, fixed.
- Reported alongside #804 and #806 from the same user session; a 690-node workflow surfaced all three.

PR #818, opened 2026-08-04
fix(#809): truncated results name their own remedy — and one that works from where the caller is

### LINKED ISSUE #809: Truncated results must name their own remedy — an agent that hits a silent cap concludes the tool cannot do the thing
## Ask

When a result is truncated, the response must tell the caller **how to get the rest**. Today some sites do, most emit a bare `…`, and several emit nothing but a boolean.

This is not cosmetic. An agent that hits a silent cap concludes the *tool cannot do the thing*, and escalates to the human instead of retrying wider. Observed live: a user's Kimi session on a 690-node graph reported *"the output keeps getting truncated by groups before reaching node details"* and asked for a workaround — it never tried raising `max_chars`, because nothing told it `max_chars` existed. Raising it fixes that case outright.

A truncation marker is an instruction to a machine. Bare `…` says "something was here". It needs to say **what was cut, how much remains, the exact parameter to raise, that parameter's ceiling, and a narrower tool if one fits**.

## Proposed contract

Every site that drops content from a tool result emits, in the result:
- what was cut and how much (`shown of matched`, or `+N more`)
- the **exact param name** to raise and its **max** (`max_chars` up to 60000)
- a narrower tool/param when the query shape suggests one (`ids:`, `upstream_of:` + `depth:`, `group_by:"type"`)
- never a bare `…` as the only signal, and never a `truncated: true` boolean as the only signal

## Two things that are outright wrong today

**1. `query_workflow` names the wrong remedy.** `src/services/graph-query.ts:408` — the no-`ids` tail reads *"narrow with types/where/ids/depth, use `group_by:"type"`, or raise **limit**."* It never mentions `max_chars`. But truncation reaches that tail from **two** causes: the `limit` cut (`:354`) and the **char-budget** cut (`:400`). When it was the char budget, raising `limit` does nothing — it makes it worse. The panel twin gets this right (`web/js/lib/graph-read.js:180-185` says "raise limit/max_chars"), so the two have already diverged. A hint that names the wrong lever is worse than no hint: the agent tries it, it fails, and that confirms "the tool can't do this."

**2. `panel_find_nodes` claims it doesn't truncate.** Description at `src/orchestrator/panel-tools.ts:4161` says *"It scans EVERY node (no truncation)"*. It does scan every node, then stops emitting matches at `limit` (default 40) and reports only `truncated: matches.length >= cap`. The description asserts something untrue about the result, which is the one thing a description must never do.

## Inventory

Sites that already explain the remedy — the model to copy:
- `graph-query.ts:144` — `"${n} widget(s) omitted (exceeded budget); raise max_chars"`
- `graph-query.ts:161-173` `fitDetailLine` — `detail_omitted: "full detail is N chars > max_chars M; raise max_chars to read this node"`
- `graph-query.ts:328`, `:405-407` (the `ids` branch)
- `src/services/node-dev.ts:400-407` — `"[... output truncated — request a narrower range ...]"`

Bare `…` / bare count, no remedy:
- `graph-query.ts:103` (widget value, cap 2048), `:151-153` `clipLine`, `:68-72` `clip`, `:364-370` ref-inputs (cap 64), `:372-374` consumers (cap 64)
- panel twins: `lib/graph-read.js:80,140-142`; `comfyui-mcp-panel.js:6301,6322-6325`
- `src/services/crash-log.ts:140`, `hierarchical-mermaid.ts:340`, `mermaid-converter.ts:142`

Silent — boolean only, or nothing at all:
- `MAX_STATE_NODES = 100` (`comfyui-mcp-panel.js:4225`) gates `panel_view_selected` (`:5957`), `panel_view_nodes_in_viewport` (`:6032`), `panel_get_subgraph` (`:6748`), `panel_get_errors` (`:8534`) — each emits only `truncated: true`, and `summarizeNode` (`:5443`) emits **raw uncapped widget values** underneath. **There is no parameter to raise this** — see below.
- `panel_find_nodes` (`:6727`), `search_node_packs` (`node-dev.ts:676-687`), `node-dev.ts:683` (600-char line, hard slice, no marker)
- `job-history.ts:154-156` traceback → `traceback_truncated: true`, no inline marker
- list previews with no "+N more": `diagnostics.ts:377`, `model-explorer.ts:65`, `registry-search.ts:83`, `model-management.ts:108`, `run-template.ts:64`, `template-schema.ts:409,412,473,522`

## Two places the contract can't be satisfied without a new parameter

These need a lever before they can name one:

- **`panel_graph_outline` has no budget at all** (`comfyui-mcp-panel.js:6044-6156`): no node cap, no char cap, no truncation marker. It emits every node plus a groups index, and the tool descriptions point agents at it **first**. On a 690-node graph this is the flood path, and it cannot even report that it flooded.
- **`MAX_STATE_NODES = 100` is a hard-coded constant** with no parameter. A caller told "raise the limit" on `panel_view_nodes_in_viewport` has nothing to raise.

## Related

- #807 — `panel_query_graph`'s `groups[]` rider (`comfyui-mcp-panel.js:6287-6293`) is attached **outside** the `max_chars` accounting, so group membership is spent before the answering rows. Confirmed by this survey. That's the specific instance; this issue is the general contract.
- #609 — fixed this class for per-node/per-widget content. The riders and the silent sites were never covered.

## Acceptance

- No result-truncating site signals only via bare `…` or only via a boolean.
- The `query_workflow` char-budget path names `max_chars`, and its ceiling.
- `panel_find_nodes`'s description stops claiming it doesn't truncate.
- A test that asserts the remedy string names a parameter that **actually exists on that tool** — the divergence in defect 1 is exactly what an untested hint string decays into.

artokun:
Closes artokun/comfyui-mcp#809

## Why

A truncation marker is an instruction to a machine. An agent that hits a **silent** cap concludes the *tool* cannot do the thing and escalates to the human, instead of retrying with a bigger budget. Observed live, not hypothetical: a Kimi session on a 690-node graph reported *"the output keeps getting truncated"* and asked for a workaround — raising `max_chars` would have answered it outright, but nothing told it `max_chars` existed.

## The two defects the issue named

**1. `query_workflow` named the wrong lever.** The no-`ids` tail read *"narrow with types/where/ids/depth, use `group_by:"type"`, or raise **limit**"* — but truncation reaches that tail from **two** causes, the `limit` cut and the char-budget cut. On the budget path raising `limit` does nothing and makes the overflow worse. The agent tries it, it fails, and the failure confirms the wrong conclusion.

The tail now names the cap that **actually fired**, with its real ceiling, and says plainly that the other lever will not help. `truncated_by: "limit" | "max_chars" | null` exposes the cause structurally too.

**2. `panel_find_nodes` claimed it doesn't truncate.** *"It scans EVERY node (no truncation)"*, while stopping at `limit` (default 40) and reporting a bare boolean. Fixed without touching the cap.

## The bar, sharpened by the review gate

Eleven codex rounds, each finding real defects. The most useful thing they produced was a sharper bar than the issue started with:

> Naming a real parameter or tool is **necessary and not sufficient**. The remedy must be **actionable from the caller's current state.**

A parameter already at its ceiling, a narrower tool that re-truncates at *its* default, a paging remedy on a slice that is one physical line, a tool that cannot re-serve the data — all name real things, and all cost exactly the wasted round trip this issue exists to prevent. Concrete instances caught and fixed:

- `apply_node_patch` telling the caller to raise a `max_chars` it does not have
- the ref-cap remedy suggesting a traversal that would re-truncate at the default limit of 40
- "read the full traceback with `get_logs`" when nothing re-serves the `/history` traceback
- "raise `max_chars` up to 60000" emitted to a caller already at 60000
- a refusal telling the caller to raise to a ceiling that could not hold the graph either

## The inverse defect, and please don't skim past it

**This PR also removes a lie, not just adds disclosure.** A **false** truncation claim costs the same wasted round trip and is worse in one way: it teaches an agent to distrust a *complete* result.

ripgrep's `--max-count` is **per file**. Asking for exactly `max_results` did two wrong things at once: a file with more matches had its extras dropped *by rg* before we ever saw a line (silent loss reported as `truncated: false`), and a search with exactly `max_results` real matches was **labelled truncated** when nothing had been dropped. Asking for `cap + 1` closes both directions with one change — the extra line is the proof of loss, and its absence is the proof of completeness. The same one-past-then-drop fix landed on `list_node_pack_files` and `panel_find_nodes`.

## Also in scope

- Every result now fits the `max_chars` it advertises — rows, tail and footers, on the row **and** aggregate paths. Fitted *afterwards* rather than reserved up front, because reserving pre-spent the 500-char floor and manufactured a truncation on a result that fitted.
- The `MAX_STATE_NODES` views say they capped, at what number, that the cap is **fixed**, and name the tool that can target the rest — **no new parameter**, per the maintainer's ruling.
- `panel_graph_outline` gains `max_chars` (same name and 500–60000 clamp as `panel_query_graph`) and degrades by **resolution, never coverage**: half a map is not a smaller map. At the floor it **refuses** with the true shape rather than returning a partial outline that reads as complete. Panel-side implementation is in the companion PR; this repo carries the surface, and detects and reports an older panel that ignores the budget.
- `node_pack_git`'s real 24000 ceiling — a code clamp its description never mentioned — is now stated, plus a 500 floor so a budget can always afford the sentence explaining itself.
- Fixed caps with no lever (per-widget 2048, per-line 600, traceback 2000, crash block 4000, tag/version/template previews) say exactly that, rather than suggesting a parameter that cannot move them.

## The test

`src/__tests__/tools/truncation-remedies.test.ts` asserts every remedy string names a parameter that **actually exists on the tool emitting it**, driven off the **real registered schema** — `listRegisteredTools()` through a real MCP client — rather than a hand-written list, which would have rotted in exactly the way the hint itself rotted. It also checks stated ceilings against the schema's real maximum, that suggested `panel_*` tools are registered, and that no "raise X" is emitted without a ceiling.

Mutation-verified: reverting the fix to its pre-#809 wording fails it; so does renaming a parameter in a remedy, dropping a ceiling, or restoring the "no truncation" claim.

## Verification

- `npm test` 4441 passed, `tsc --noEmit` clean, `npm run docs:gen` re-run so the published parameter reference matches.
- 11 codex review rounds (`gpt-5.6-terra`), iterated to a clean bill on the caller-facing contract.
- No NUL bytes; no binary diffs.

## Ordering

The panel half is a **separate PR** in `artokun/comfyui-mcp-panel` (same branch name). **Land this one first** — it is the surface that advertises the new behaviour, and it degrades honestly against a panel that predates the change.

## Left for follow-up

- **#807** owns folding `panel_query_graph`'s `groups`/`rails` riders into the `max_chars` accounting. This PR bounds them and states in-band and in the description that `max_chars` bounds `text` only.
- The mermaid diagram label truncations (`hierarchical-mermaid.ts`, `mermaid-converter.ts`) are deliberately untouched — a remedy string inside a rendered diagram node label is noise, not an instruction.

### DIFF
--- a/docs/tools/custom-nodes.mdx
+++ b/docs/tools/custom-nodes.mdx
@@ -535,7 +535,7 @@ List the files in one installed custom-node pack under &lt;COMFYUI_PATH&gt;/cust
   Optional glob to filter entries (supports *, **, ?), matched against pack-relative paths.
 </ParamField>
 <ParamField path="max_entries" type="integer">
-  Maximum entries to return (default 500, max 2000).
+  Maximum entries to return (default 500, max 2000 — a hard clamp). The walk STOPS at this many, so a capped result is not the pack's full file list.
 </ParamField>
 
 ### Example
@@ -569,7 +569,7 @@ Read a slice of one file inside a custom-node pack (read-only), with bounded out
   Number of lines to return (default 240, max 800).
 </ParamField>
 <ParamField path="max_chars" type="integer">
-  Maximum characters to return (default 12000, max 24000).
+  Maximum characters to return (default 12000, min 500, max 24000 — hard clamps; values outside are silently pulled into range).
 </ParamField>
 
 ### Example
@@ -603,7 +603,7 @@ Regex-search custom-node source under custom_nodes/ (read-only). Uses ripgrep wh
   Optional glob to restrict which files are searched (e.g. '**/*.py').
 </ParamField>
 <ParamField path="max_results" type="integer">
-  Maximum matches to return (default 50, max 100).
+  Maximum matches to return (default 50, max 100). The scan STOPS at this many, so a capped result is not a complete match set.
 </ParamField>
 <ParamField path="case_sensitive" type="boolean">
   Match case-sensitively (default false).
@@ -703,7 +703,7 @@ Run a git operation inside one custom-node pack (status/diff/log/commit/push). R
   Pack-relative paths to stage/scope (jail-checked). Defaults to all pack changes.
 </ParamField>
 <ParamField path="max_chars" type="integer">
-  Maximum characters of git output to return (default 12000).
+  Maximum characters of git output to return (default 12000, min 500, max 24000 — hard clamps the runtime applies; values outside are silently pulled into range).
 </ParamField>
 
 ### Example
--- a/docs/tools/workflow-authoring.mdx
+++ b/docs/tools/workflow-authoring.mdx
@@ -315,7 +315,7 @@ Read Prompt Director's latest sanitized RUNTIME state after its nodes execute. R
 
 ## query_workflow
 
-QUERY a SAVED workflow FILE (or JSON you pass in) — NOT the live canvas, which is panel_query_graph / panel_graph_outline. Filter, traverse, project, and aggregate over its nodes WITHOUT dumping the whole JSON (the missing middle between analyze_workflow's fixed summary and get_workflow's full dump; on 100+-node graphs this is the ONLY context-safe way to answer questions like 'which KSamplers run cfg&gt;7', 'what feeds node 42', 'count nodes by type'). Provide exactly one of path/filename/graph, then combine: `types` (class_type contains any), `title` (contains), `where` widget predicates ANDed ('cfg&gt;7', 'steps&lt;=20', 'sampler_name=euler', 'text~sunset' — ops = != &gt;= &lt;= &gt; &lt; ~contains), `ids` (exact nodes — the way to read ONE node's detail), `upstream_of`/`downstream_of` + `depth` (dependency traversal: upstream = what FEEDS that node, downstream = what CONSUMES it; seed included at depth 0), `fields` ('compact' one line per node [default], 'ids', or 'detail' JSON rows with widgets + wiring), `group_by:'type'` (counts only), `limit` (default 40). Output is TOKEN-BOUNDED with an explicit truncation marker. Read-only.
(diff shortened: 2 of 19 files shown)


artokun:
Panel companion: artokun/comfyui-mcp-panel#614 — land this one first.

Issue #761, opened 2026-08-03
download_status silently drops in-flight downloads after a heartbeat gap — parseJobFile destructively reaps live jobs on a read path


artokun:
### What happened / root cause

`download_status` listed 4 downloads (1 done, 3 in-flight, streaming normally). ~5 minutes later the SAME call listed **only the completed one**. The 3 in-flight jobs were gone entirely — not `error`, not `cancelled`, just absent, indistinguishable from "never started". Nothing had landed in `models/` and no `.partial` was visible there either, so the only reasonable read was "the downloads died".

They had **not** died. All three were still streaming: `~/.comfyui-mcp/cache/*.partial` kept growing, and re-issuing `download_model` for the same URLs re-adopted the **same job ids** (`bf5bbfae311f3bd9`, `c2adb76c3613c38c`, `c2714d0bcf403880`) with **monotonically increased** progress (64%->71%, 18%->20%, 14%->16%). After that adopt, `download_status` listed all four again.

Root cause is in `dist/services/download-progress.js` `parseJobFile()`:

```js
const inflightDead = raw.status === "downloading" && age > PERSISTED_INFLIGHT_STALE_MS; // 60s
if (terminalExpired || inflightDead) {
    rmSync(join(dir, f), { force: true });
    return null;
}
```

An in-flight persisted record whose `updated` heartbeat is older than `PERSISTED_INFLIGHT_STALE_MS` (60s) is treated as proof the writer crashed, and the record file is **deleted** — from a **read-only** path. `parseJobFile` is called by `listPersistedDownloadJobs()`, which `listDownloadJobs()` (i.e. plain `download_status`) folds in.

Two problems compound:

1. **A status READ destroys state.** Any reader — including a different session — can permanently delete another live owner's job record just by calling `download_status` during a heartbeat gap.
2. **A stale heartbeat does not imply a dead transfer.** An orchestrator respawn / agent-session reconnect (mine dropped and resumed mid-task, which is exactly the documented "connection dropped, session resumed" flow) stops the 15s heartbeat while the HTTP transfer keeps streaming to its `.partial`. The comment's mitigation ("a falsely-reaped one reappears on the owner's next heartbeat") did not hold here — the record stayed gone until an explicit `download_model` re-adopt.

[...]

Issue #547, opened 2026-07-31
download_model completion never notifies the agent — no agent_event on terminal download state (agent must poll download_status in sleep loops)


artokun:
### What happened / root cause

A finished **render** pushes an `agent_event` into the panel agent's session, so the agent is woken with the result (`panel-agent.ts` `injectEvent`, `kind: "executed"` / `"run_error"`). A finished **download** has no equivalent — nothing wakes the agent when `download_model` settles.

Consequence: after `download_model` returns its handle ("still running, id `…`"), the agent has no completion signal. The only way to learn the file landed is to poll `download_status`, which means burning turns on `sleep` + re-poll loops (I used two background `sleep` commands for an 11.5 GB checkout, then re-polled). This is exactly the friction the render path already solved, and it's inconsistent — the tool result even instructs the agent to wait for the user to ask, which means a "download then immediately use the model" task stalls until the user pokes it.

Root cause, concretely:

- `src/services/download-jobs.ts` — the job's `settled` promise sets `job.status = "done" | "error"` and `finished_at`, then **nothing else**. No callback, no event emission. The registry is poll-only (`getDownloadJob` / `listDownloadJobs`).
- `src/orchestrator/index.ts` `pollDownloads()` (~L3507–3556) already **observes the terminal transition** — it sees `status === "done" || "error"` and starts the `DOWNLOAD_LINGER_MS` prune timer (`downloadRemoveAt.get(full) == null` is exactly the first-terminal-observation moment). It broadcasts `{ type: "download_progress" }` to the panel **UI**, but never calls `manager.injectEvent(...)`, so the agent is not told.

So the detection already exists in the right place; it just isn't wired to the agent channel.

### Steps to reproduce

1. In the sidebar panel, ask the agent to run a workflow whose model is missing (e.g. a Z-Image Turbo graph missing `z_image_turbo_bf16.safetensors`).
2. Agent calls `download_model` → gets back `Download STARTED and is still running (28%) — id 9a1f39a9be8570c1`.
3. Download completes ~90s later.
4. **Observed:** the agent receives nothing. No turn is triggered. It only finds out by calling `download_status` again, which requires it to guess when to look (I resorted to `sleep 45` / `sleep 40` background commands as a wake timer).
5. **Expected:** an `agent_event` when the job settles, mirroring the render-finished event, e.g. "download `z_image_turbo_bf16.safetensors` finished → landed at `<path>`" (or the error), so the agent can immediately continue with what the download was blocking.

[...]

Issue #529, opened 2026-07-31
download_status loses active download IDs after sidebar session reconnect


artokun:
### What happened / root cause
A managed model download started successfully and returned download id `8be98c3775f804b3`. After the sidebar/tool session reconnected, `download_status` could not resolve that active job and said downloads are tracked per server session. This prevents the agent from confirming whether it should safely retry and forces reliance on the visual panel tray.

### Steps to reproduce
1. Start a large model with `download_model` so it remains in flight.
2. Reconnect or respawn the sidebar agent/tool session.
3. Call `download_status` with the id returned in step 1.
4. Observe `No download with id ... Downloads are tracked per server session`.

### Exact error
No download with id `8be98c3775f804b3`. Downloads are tracked per server session, so an id from a previous session won't resolve — check the panel download tray instead.

### Fix
Applied locally: no. Upstream-only: yes. Persist active download metadata in a process-independent store or make `download_status` discover in-flight panel download jobs by destination/URL after reconnect. At minimum, add a query by URL or destination so callers can safely adopt an existing job without starting a duplicate.

### Environment
Linux; NVIDIA RTX 4090 24 GB; ComfyUI 0.28.0; comfyui-mcp 0.48.21; panel version unavailable in the provided environment snapshot.

PR #831, opened 2026-08-04
download_model: large-file timeouts, no resume, wrong base on Windows portable, non-unique job ids

### LINKED ISSUE #470: download_model: no resume/retry — large HF downloads die mid-transfer with "failed: terminated", losing all partial data

While setting up a SeedVR2 video-upscale job, three `download_model` calls for large HuggingFace weights were issued. Two of the three died mid-transfer:

- `seedvr2_ema_7b_fp8_e4m3fn_mixed_block35_fp16.safetensors` (~9 GB, repo `AInVFX/SeedVR2_comfyUI`) — reached **3.31 GB** on disk, then `download_status` reported `error / failed: terminated`.
- `seedvr2_ema_7b_fp16.safetensors` (15.35 GB, repo `numz/SeedVR2_comfyUI`) — died at ~1 GB with the same `failed: terminated`.

Small/medium files were fine (`ema_vae_fp16.safetensors`, 501 MB → `done`).

Two distinct problems compound here:

1. **No automatic retry** on a dropped connection. A transient TCP/HF-CDN drop terminates the whole job permanently; nothing resumes it.
2. **No resume support.** The 3.31 GB already written to disk is discarded — re-issuing `download_model` for the same URL restarts from byte 0 rather than sending a `Range: bytes=3315622731-` header. On a slow link (this machine was averaging 1.5–3 MB/s) that turns a recoverable blip into a 30+ minute loss, repeatedly.

There is also a secondary oddity worth checking: immediately re-issuing `download_model` for the terminated `AInVFX` URL returned **`403 Forbidden`**, where the first attempt had streamed 3.3 GB successfully from the same URL. That looks like CDN throttling/blocking triggered by the aborted-then-immediately-reopened connection — another argument for backoff on retry rather than an instant re-request.

Net effect for an agent: a routine "install pack → download weights → build graph → run" flow becomes unreliable on any multi-GB model over a modest connection, and the agent has no tool-level way to salvage progress. The documented guidance is to prefer `download_model` over shelling out to curl/wget — but curl with `-C -` would actually have survived this, so the sanctioned path is currently the weaker one.

### Steps to reproduce

1. `download_model({url: "https://huggingface.co/numz/SeedVR2_comfyUI/resolve/main/seedvr2_ema_7b_fp16.safetensors", target_subfolder: "SEEDVR2"})` (15.35 GB).
2. Let it run on a connection slow/flaky enough that the transfer is interrupted (or kill the connection manually partway through).
3. `download_status` → `error / failed: terminated`; the partial file remains on disk at its interrupted size.
4. Re-issue the identical `download_model` call → progress restarts at 0% (or, for the `AInVFX` URL, returns `403 Forbidden`).

### Exact error (scrubbed)

```
## Downloads

- `410e8e19551377d0` **error**
    failed: terminated
    from: https://huggingface.co/numz/SeedVR2_comfyUI/resolve/main/seedvr2_ema_7b_fp16.safetensors
- `b6430fc74ec143b1` **error**
    failed: Download failed: 403 Forbidden
    from: https://huggingface.co/AInVFX/SeedVR2_comfyUI/resolve/main/seedvr2_ema_7b_fp8_e4m3fn_mixed_block35_fp16.safetensors
- `02ef8f17b52c3763` **done**
    landed at: ~/models/SEEDVR2/ema_vae_fp16.safetensors
```

And the earlier failure, observed as a stalled file on disk (no growth for ~30 min while the job was still listed as in-flight before flipping to `error`):

```
-rw-r--r-- 1 user 3315622731 Jul 30 13:52 seedvr2_ema_7b_fp8_e4m3fn_mixed_block35_fp16.safetensors
```

### Fix

applied locally: **no**   upstream-only: **yes**

This needs a change in the `download_model` implementation in `comfyui-mcp` (the local-streaming path — not the remote Manager-dispatch path). Precise change needed:

1. **Stream to a `.part` sidecar**, then atomically rename to the final filename on completion. Right now a partial write lands at the real filename, which also means a half-downloaded model shows up in `list_local_models` and in node dropdowns as if it were valid. (This bit us separately: the 3.3 GB partial had to be manually deleted before the node's model dropdown was trustworthy.)
2. **Resume on retry.** Before starting, `stat` the `.part` file; if it exists and is non-empty, send `Range: bytes=<size>-` and append. Honor the response: `206 Partial Content` → append; `200 OK` → server ignored the range, truncate and restart.
3. **Automatic retry with exponential backoff** on transient transport failures (`terminated`, `ECONNRESET`, `ETIMEDOUT`, socket hang up) — e.g. 5 attempts at 2s/4s/8s/16s/32s, each resuming per (2). The instant no-backoff re-request is likely what earned the `403`.
4. **Integrity check** where available: the SeedVR2 pack ships known sha256 values per file in `src/utils/model_registry.py`; more generally, verify `Content-Length` against the final size and fail loudly on mismatch rather than renaming a short file into place.
5. **Surface stall detection in `download_status`**: report bytes/sec and flag "no progress for N seconds", so an agent can distinguish "slow but alive" from "silently wedged" without polling the filesystem itself (which is what I ended up doing).

A minimal version of (1)+(2)+(3) alone would have made this job succeed unattended.

### Environment

- OS: Windows 10 Home Single Language 10.0.19045 (win32)
- ComfyUI: 0.28.3 (local, `127.0.0.1:8189`), Manager v4
- GPU: NVIDIA GeForce RTX 5090, 32 GB VRAM; 64 GB RAM
- CUDA 13.0 / torch 2.10.0+cu130 / Python 3.13.12
- comfyui-mcp: 0.48.12
- panel: 49e0587291ccd51a81127e835fa620a10229f008
- Downloads targeted `models/SEEDVR2/`; free disk space was 1.18 TB (not a space issue)

### LINKED ISSUE #813: download_model rejects a valid COMFYUI_PATH on Windows portable installs (relative main.py anchoring double-nests the base)

On a Windows portable ComfyUI install, `download_model` (and the shared `resolveModelsDirWithBases` in `services/output-dir.js`) refused every download with:

> The models directory of the CONNECTED ComfyUI could not be determined ... the configured local base "D:\\ComfyUI_windows_portable\\ComfyUI" does not contain "ComfyUI\\main.py", so it is a DIFFERENT install than the one that is running.

`COMFYUI_PATH` was set to `D:\ComfyUI_windows_portable\ComfyUI` — verified correct: `main.py` exists directly under it, and `get_environment` / `list_local_models` both already treat this exact value as the right install root.

The running server reports its launch script as the relative path `ComfyUI\main.py` with no cwd (the standard Desktop/portable shape), so `resolveLiveServerRoot` returns `source: "unresolved", relDir: "ComfyUI"`. The anchoring code in `output-dir.js` then does:

```js
const anchored = base && live?.relDir !== undefined && live.source === "unresolved"
    ? resolve(base, live.relDir)
    : undefined;
```

This assumes `base` (COMFYUI_PATH) is the OUTER launcher root and the real server lives one level down at `<base>/ComfyUI/main.py` (true for ComfyUI Desktop). But when `COMFYUI_PATH` is already set to the inner ComfyUI directory (the convention every OTHER resolver in this codebase treats as correct — `resolveEffectiveComfyUIBase`, `get_environment`, `list_local_models` all work fine with this value), `resolve(base, relDir)` double-nests to `<base>\ComfyUI\ComfyUI`, which never exists, so the entrypoint check fails and the whole resolution refuses.

### Steps to reproduce
1. Run ComfyUI as the classic Windows portable bundle (relative `main.py`, no cwd reported in `/system_stats`).
2. Set `COMFYUI_PATH` to the directory that directly contains `main.py` (e.g. `D:\ComfyUI_windows_portable\ComfyUI`) — the same value `get_environment`/`list_local_models` already report as correct.
3. Call `download_model` with any url/target_subfolder.

### Exact error (scrubbed)
```
VALIDATION_ERROR: The models directory of the CONNECTED ComfyUI could not be determined, so a download has no verified destination. What could not be determined:
  - the running server reported its launch script as the RELATIVE path "ComfyUI\main.py" and did NOT report a working directory, so its install root is unknown;
  - the OS process table did not identify an interpreter for the ComfyUI listening on 8188 that sits inside an install tree;
  - the configured local base "D:\ComfyUI_windows_portable\ComfyUI" does not contain "ComfyUI\main.py", so it is a DIFFERENT install than the one that is running.
Refusing to write to a guessed directory ...
```

### Fix

Patched `dist/services/output-dir.js` (the anchoring block inside `resolveModelsDirWithBases`) to try the existing nested interpretation first (`<base>/<relDir>/main.py`, for ComfyUI Desktop's outer-launcher-root convention), and fall back to `base` itself directly when IT already contains `main.py` — covering the equally-valid convention (used everywhere else in this codebase) where COMFYUI_PATH already points at the inner ComfyUI directory. Verified: re-ran `download_model` after the patch — it now resolves to `D:\ComfyUI_windows_portable\ComfyUI\models` and writes successfully (previously always threw before touching disk).

```diff
--- a/dist/services/output-dir.js
+++ b/dist/services/output-dir.js
@@
-        const anchored = base && live?.relDir !== undefined && live.source === "unresolved"
-            ? resolve(base, live.relDir)
-            : undefined;
-        if (anchored && hasComfyUIEntrypoint(anchored)) {
-            modelsDir = join(anchored, "models");
-            source = "base-anchored";
-            baseDirs.add(anchored);
-            logger.debug("Anchored the live server's relative main.py on the configured base", { modelsDir, relDir: live?.relDir });
-        }
+        const canAnchor = base && live?.relDir !== undefined && live.source === "unresolved";
+        // Two base conventions coexist in this codebase: `base` as the OUTER
+        // launcher root (Desktop/portable's `<base>/ComfyUI/main.py` nesting,
+        // per hasComfyUIEntrypoint's doc above) vs. `base` already pointing AT
+        // the inner ComfyUI dir (the convention resolveEffectiveComfyUIBase /
+        // get_environment / list_local_models already treat as correct). Try
+        // the nested form first; fall back to `base` itself when IT directly
+        // has the entrypoint, so a COMFYUI_PATH already set to the ComfyUI dir
+        // isn't rejected as "a different install" (#529).
+        const nestedAnchor = canAnchor ? resolve(base, live.relDir) : undefined;
+        const anchored = nestedAnchor && hasComfyUIEntrypoint(nestedAnchor)
+            ? nestedAnchor
+            : (canAnchor && hasComfyUIEntrypoint(base) ? base : undefined);
+        if (anchored) {
+            modelsDir = join(anchored, "models");
+            source = "base-anchored";
+            baseDirs.add(anchored);
+            logger.debug("Anchored the live server's relative main.py on the configured base", { modelsDir, relDir: live?.relDir, anchored });
+        }
```

Note: patched the installed `dist/*.js` directly (this machine has a `_npx` cache install, not a source checkout), so the same fix needs porting into the TS source (likely `src/services/output-dir.ts`) for a real release.

### Environment
OS Windows 11 (10.0.26200) · ComfyUI 0.29.0 (local, Windows portable bundle) · GPU NVIDIA GeForce RTX 5090 32GB VRAM · comfyui-mcp 0.49.4 · panel 0.11.35

### LINKED ISSUE #817: download_model repeatedly times out at 600s for large Hugging Face files
Large MiniMax H3 diffusion model downloads dispatched to remote ComfyUI-Manager remain in downloading state, then fail at exactly the 600-second Manager queue timeout. The resulting model files are truncated and cause safetensors tensor-shape errors in UNETLoader.

### Steps to reproduce
1. Call download_model for the official MiniMax H3 diffusion file into diffusion_models.
2. Poll download_status while it streams.
3. After 600 seconds, status becomes error: "ComfyUI-Manager queue did not finish within 600s".
4. Attempting to load the resulting file fails with a shape mismatch.

### Exact error
ComfyUI-Manager queue did not finish within 600s.
UNETLoader: shape '[21504, 5376]' is invalid for input of size 34381536.

### Fix
Applied locally: no. Upstream-only: no. Needed: resumable/direct streaming or a configurable timeout for large remote model downloads, plus verification of file completeness before reporting a usable model.

### Environment
Linux; ComfyUI 0.30.0; NVIDIA RTX 5090 31 GB; comfyui-mcp 0.49.4; panel 0.11.38.

Related (same root cause): #143

### LINKED ISSUE #822: download_status: exposed job `id` is not unique across rows (two URLs → one destination), so cancel_download cannot select either job

`download_status` listed TWO distinct in-flight rows that share the SAME `id` (`a0b477082f35745c`) but have DIFFERENT source URLs, both resolving to the same destination file `models/text_encoders/qwen3vl_4b_fp8_scaled.safetensors`:

- live job started this session — `https://huggingface.co/Comfy-Org/Krea-2/resolve/main/text_encoders/qwen3vl_4b_fp8_scaled.safetensors` (0.20/4.88 GB, 4%)
- orphaned job from before a connection drop — `https://huggingface.co/Aitrepreneur/FLX/resolve/main/qwen3vl_4b_fp8_scaled.safetensors` (started 884s ago, heartbeat stale 313s)

This is NOT a hashing accident — it follows from the intended design, but the public surface doesn't carry enough information to use it. In `dist/services/download-jobs.js`:

- `downloadJobIdFor(identity)` (L157) derives `id` from the resolved **destination path + auth discriminator**, so two different URLs landing on the same file deliberately share one `id`.
- `listDownloadJobs()` (L~895) dedups on the composite identity `keyOf(j) = ` + "`${j.id}\n${j.trayId}`" + ` and its own comment states: \"two distinct URLs to one dest+auth share an id but are distinct physical downloads (different trayId), and both must be listed\".
- `findDownloadJob()` (L~845) likewise keys candidates by `(id, trayId)` and explicitly declines when ambiguous.

So internally the unique handle is **(id, trayId)** — but only `id` is rendered in `download_status` output, and `cancel_download` accepts **only** `id`.

### Impact

1. `download_status` output is self-contradictory as a handle source: it documents `id` as the thing to pass back, yet emits two rows with the same `id`.
2. `cancel_download` (schema: single required `id`) has no way to express *which* of the two same-destination jobs to abort. The stale one cannot be selected.
3. Two live transfers are writing the same destination path — last writer wins. The dedup design exists precisely to avoid \"two writers, one file\" (see the `startDownload` doc comment, L~190), but that guarantee only holds for same-URL adoption via `requestDownloadKey` (L181, keyed on url+subfolder+filename+auth); two *different* URLs to one destination bypass it.
4. Recovery is also blocked by `cancel_download`'s documented limitation (\"a download started before a reconnect ... must be stopped from the panel download tray\"), so an agent that notices the race has no tool-surface remedy at all.

### Steps to reproduce

1. `download_model` URL_A → `text_encoders/foo.safetensors`; let it run.
2. Drop/restart the MCP session so job A becomes an orphaned persisted in-flight record.
3. `download_model` URL_B (different repo, same resolved filename) → same subfolder.
4. `download_status` → two rows, same `id`, different `from:` URLs.
5. `cancel_download {id}` → cannot target a specific one.

### Fix

applied locally: **no**   upstream-only: **yes**

Not patched deliberately: the identity/adoption logic carries layered reasoning from #420, #467 and #529, and changing what `id` means risks breaking the same-URL adoption and cross-session resume guarantees. The gap is in the **public surface**, not the internal model — the composite identity already exists.

Precise change needed:

1. Render `trayId` on every `download_status` row (it is already on the job object) — e.g. `` `<id>` (tray `<trayId>`) `` — so two same-destination rows are visually and programmatically distinguishable.
2. Let `cancel_download` accept an optional disambiguator — `tray_id` and/or `url` — and resolve via the existing `keyOf(id, trayId)`; when `id` alone matches multiple in-flight jobs, return an explicit ambiguity error listing the candidate `trayId`s instead of silently acting on one.
3. Optional hardening: when a new `download_model` request resolves to a destination that already has a **different-URL** in-flight job, warn (or require an explicit override) rather than starting a second writer to the same path.

### Environment

- OS: Linux 6.8.0-100-generic
- GPU: NVIDIA RTX 4000 SFF Ada Generation, 20 GB VRAM · CUDA 12.8 · torch 2.11.0 · python 3.12
- ComfyUI 0.30.0 (local, `~/ComfyUI-clean`, port 6970) · ComfyUI-Manager 3.x (legacy)
- comfyui-mcp: **0.49.4**
- comfyui-mcp-panel: **0.11.38**
- Partial staging dir: `~/.comfyui-mcp/cache/.<hash>.safetensors.partial` (hash IS per-URL, so the two partials are correctly distinct — only the reported `id` and the final destination collide)

artokun:
Closes artokun/comfyui-mcp#470
Closes artokun/comfyui-mcp#817
Closes artokun/comfyui-mcp#813
Closes artokun/comfyui-mcp#822

## #470 — a transfer that survives an interruption

The `.partial` + `.etag` machinery from #343/#467 already made an interrupted download **resumable**; nothing **resumed** it. One CDN drop ended the job with `failed: terminated` and a human had to notice and re-issue.

Added a bounded retry loop that resumes in the same call, plus a **stall watchdog** so a wedged socket is abandoned in seconds rather than sat on for half an hour (#470 watched one do nothing for ~30 minutes). A stall bound is the right shape for a resumable transfer; a total-duration cap — the 600s in #817 — is not, because it kills healthy large transfers and cannot tell them from wedged ones.

**How a resume proves object identity.** It does not skip anything a cold call does. Every attempt re-derives its offset from disk and goes through the same proof: the validator recorded at *write* time replayed as `If-Range`, the content-addressed `X-Linked-Etag` cross-check for cross-origin 206s, strict `Content-Range` parsing, and the recorded-total cross-check. There is no "we already checked" shortcut, and when identity cannot be proven the attempt **restarts** rather than appending. #467's ETag false-positive is not reintroduced: no validator is ever *probed* at resume time.

Classification is **fail-safe** — nothing is retried unless explicitly recognised as a transport interruption. 403/404, the #473 payload rejections and the #343/#467 integrity refusals are never retried; #470 traced a 403 to an instant re-request after an abort, so the backoff is part of the fix. A user cancel is decided by the *signal*, never by the error text.

Also fixed a false failure this introduced: a to-be-retried attempt must not publish a terminal `error` progress row, because the orchestrator treats the first terminal row as the download's **outcome** and wakes the tab's agent with it (#547) — telling an agent a still-running download failed invites the re-issue that creates a second writer.

Tunables: `COMFYUI_DOWNLOAD_MAX_ATTEMPTS` (default 5, 1 disables), `COMFYUI_DOWNLOAD_STALL_TIMEOUT_S` (default 120, 0 disables).

## #817 — the 600s ceiling

`install-model` shared the node-install queue budget, which is a guarantee of failure for any multi-GB file. It now has its own budget (`COMFYUI_MANAGER_DOWNLOAD_TIMEOUT_S`, clamped 60s–24h), and the timeout message is honest: the **wait** gave up, the host is probably still downloading, and re-issuing writes the same destination concurrently — which is how #817's reporter ended up with a truncated model.

## #813 — portable base anchoring

Only the outer-launcher-root reading (`<base>/<relDir>/main.py`, ComfyUI Desktop) was implemented, so a Windows portable install whose `COMFYUI_PATH` is the inner ComfyUI directory had every download refused. Both readings are now accepted, on the **same corroboration**: the server's relative script, anchored one level up, must name that very directory *and* `main.py` must be there. Precedence follows `serverRootsUnder` (#401/#785) rather than a second convention. A base that merely happens to contain `main.py` is still refused, and when **both** layouts exist the evidence does not discriminate — so it refuses rather than guessing where a multi-gigabyte file lands.

## #822 — an identifier that does not identify

`id` hashes the destination, so two source URLs landing on one file share it by design. `download_status` rendered it as a job handle and `cancel_download` accepted only it, so neither of two same-`id` rows could be selected.

The composite `(id, trayId)` already existed internally; it is now exposed. Every status row prints its tray id, a listing containing a collision says so loudly, `download_status(id)` on an ambiguous id **names the candidates** instead of reporting "no download" (which turned "could not determine *which*" into a false "it never started"), and both tools accept `tray_id`. #761's rule is preserved — a dead session's stale record still cannot jam the live download this process owns — and a live foreign transfer is never hidden behind this session's settled row.

## Verification

- 4831 tests pass; `lint`, `build`, `asset-counts --check`, `check:vocabulary`, `vocab:export --check` all green.
- Every fix is **mutation-verified**: the mutant that removes it fails a named test (~25 mutants across 15 rounds). Two exceptions are called out in the commit messages rather than glossed — the cloud-branch pre-write hook and the strict-validator branch are defence in depth whose unique windows are not reachable through the public API; in both cases another check claims the state first, and the *property* is asserted against disk.
- Self-gated with codex (gpt-5.6-terra), 7 rounds. Rounds 1–6 found real defects — several of them regressions this diff introduced — and all were fixed; round 7 returned **PASS** with no P0/P1/P2 findings.

## Known exclusion

Two processes streaming into the *shared* staged file at genuinely overlapping times. Every window the retry loop **adds** is now closed (the backoff, read-decide-then-write, the cloud truncate, the integrity refusals, and unknown ownership in either field — all failing closed). What remains needs write exclusion held for a download's lifetime, applies identically to a single-shot download, and is deliberately deferred rather than half-solved with a lease whose stale-claim and crash-cleanup behaviour (#529's stated objection) could destroy live multi-gigabyte partials.

## Needs live verification

No real large download was performed (the shared ComfyUI on this machine is in use) — all coverage is fixtures. Worth confirming on a real rig: an interrupted multi-GB HF transfer resuming in one call, the stall watchdog on a genuinely wedged CDN socket, a real Windows portable install with `COMFYUI_PATH` set to the inner directory, and the `install-model` budget against a real remote Manager.

### DIFF
--- a/docs/tools/models.mdx
+++ b/docs/tools/models.mdx
@@ -272,6 +272,9 @@ Check on model downloads started by download_model. Reports each download's stat
 <ParamField path="id" type="string">
   Download id from download_model. Omit to list every tracked download (incl. in-flight ones from before a reconnect).
 </ParamField>
+<ParamField path="tray_id" type="string">
+  Use this when two rows come back with the SAME `id`, so the id alone cannot say which one you mean. That happens when two different source URLs are downloading to the same destination file. Every row prints its own tray id as `(tray &lt;tray_id&gt;)` — pass that here, together with `id`, to report on exactly one of them.
+</ParamField>
 <ParamField path="url" type="string">
   Adopt an in-flight download by its source URL when you don't have the id (e.g. after a reconnect) — reports the matching job without starting a duplicate.
 </ParamField>
@@ -297,6 +300,9 @@ Cancel ONE in-flight model download started by download_model / download_civitai
 
 <ParamField path="id" type="string" required>
   The download id to cancel (from download_model / download_civitai_model / download_status).
+</ParamField>
+<ParamField path="tray_id" type="string">
+  Use this when a cancel is refused because the `id` names more than one download — two different source URLs writing to the same destination file, so cancelling by id could stop the wrong one. Every download_status row prints its own tray id as `(tray &lt;tray_id&gt;)`; pass the one you want stopped here, together with `id`, and only that download is aborted.
 </ParamField>
 
 ### Example
--- a/src/__tests__/services/download-cache-failclosed.test.ts
+++ b/src/__tests__/services/download-cache-failclosed.test.ts
@@ -63,22 +63,37 @@ describe("assertComplete fail-closed (#467 P1-B)", () => {
   it("does NOT finalize (rename into cache) when the size can't be verified against a known Content-Length", async () => {
     // A full body WITH an authoritative Content-Length. Normally this verifies and
     // finalizes; with stat failing at completion it must fail closed instead.
+    // The stat must fail AT COMPLETION, which is what this test is about. Flipping
+    // the flag before the call would instead trip the staged-file ownership check
+    // (an unreadable staged file now refuses BEFORE writing, rather than being
+    // mistaken for "no partial" and truncated) — a different, also-correct
+    // refusal, but not the one named here. Flip it once the body is consumed.
     fetchMock.mockResolvedValueOnce(
-      new Response("the-model-bytes", {
-        status: 200,
-        statusText: "OK",
-        headers: { "content-length": "15" },
-      }),
+      new Response(
+        new ReadableStream<Uint8Array>({
+          pull(c) {
+            c.enqueue(new TextEncoder().encode("the-model-bytes"));
+            statCtl.fail = true;
+            c.close();
+          },
+        }),
+        { status: 200, statusText: "OK", headers: { "content-length": "15" } },
+      ),
     );
 
-    statCtl.fail = true;
     await expect(
       downloadWithCache({
         url: "https://example.com/models/failclosed.safetensors",
         headers: {},
         targetPath: join(comfyDir, "out.safetensors"),
       }),
-    ).rejects.toThrow(/could not be verified|not finalizing/i);
+      // Either fail-closed refusal is correct here, and which one fires depends on
+      // WHICH stat the simulated EIO reaches first. The completion check says "I
+      // could not verify the size"; the staged-file ownership check says "I could
+      // not read the staged file, so I will not act on it". Both refuse for the
+      // same reason — an unreadable file is not a known-good one — and the
+      // load-bearing assertion below (nothing was finalized) is unchanged either way.
+    ).rejects.toThrow(/could not be verified|not finalizing|could not be read/i);
 
     // Crucially: nothing was renamed into the cache as a "successful" download.
     statCtl.fail = false;
--- a/src/__tests__/services/download-cache.test.ts
+++ b/src/__tests__/services/download-cache.test.ts
@@ -23,6 +23,7 @@ import { config } from "../../config.js";
 import { downloadCacheFs } from "../../services/download-cache.js";
 import { downloadModel } from "../../services/model-resolver.js";
 import type { ResumeDiagnostic } from "../../services/download-resume-diag.js";
+import { setDownloadRetryPolicyForTests } from "../../services/download-retry.js";
 
 /** Capture the resume decision a physical download reports (#467). The decision
  *  is delivered to the CALLER via an onResume callback (no shared map), so a test
@@ -70,6 +71,12 @@ beforeEach(async () => {
   config.civitaiApiToken = undefined;
   vi.stubGlobal("fetch", fetchMock);
   fetchMock.mockReset();
+  // These tests exercise ONE physical attempt's integrity semantics (#343/#467/#473).
+  // #470 added automatic retry on transient failures, which would otherwise re-enter
+  // the stream with a mock that has no further queued response. Pin a single attempt
+  // and disable the stall watchdog so each assertion is about the attempt itself.
+  // Retry behaviour has its own dedicated coverage (download-retry.test.ts).
+  setDownloadRetryPolicyForTests({ maxAttempts: 1, stallTimeoutMs: 0 });
 });
 
 afterEach(async () => {
--- a/src/__tests__/services/download-jobs.test.ts
+++ b/src/__tests__/services/download-jobs.test.ts
@@ -628,29 +628,49 @@ describe("download job registry", () => {
       }
     });
 
-    it("cancel_download with a tray_id aborts EXACTLY that download, leaving its same-id sibling alone", async () => {
-      // Two in-flight jobs in THIS process that share a destination id would
-      // normally coalesce, so drive the distinguishing case directly: two jobs with
-      // different ids, and confirm tray-id selection targets the right one and does
-      // not fall back to "whatever the id row happens to hold".
-      const a = await startDownloadJob(URL_A, "checkpoints");
-      const b = await startDownloadJob(URL_B, "loras");
+    it("cancel_download with a tray_id aborts EXACTLY the named row of a COLLIDING id, and leaves the other alone", async () => {
+      // The real #822 collision: one live local download and one orphaned in-flight
+      // record from before a reconnect, sharing an id because they resolve to the
+      // same destination file. `tray_id` must pick out precisely one of them.
+      const dir = await mkdtemp(pathJoin(tmpdir(), "djobs-822-"));
+      setProgressDir(dir);
+      try {
+        const a = await startDownloadJob(URL_A, "checkpoints");
+        await writeForeignJobRecord(dir, {
+          id: a.job.id,
+          trayId: "orphantrayid0001",
+          progressId: "orphan-prog",
+          url: URL_B,
+          owner: `${PERSIST_OWNER}-reconnected-away`,
+          ageMs: 313_000,
+        });
+        // Precondition: the id really does name two rows.
+        expect(listDownloadJobCandidates(a.job.id)).toHaveLength(2);
 
-      const wrongTray = cancelDownloadJob(a.job.id, b.job.trayId);
-      // a's id + b's tray is not a real download — nothing may be aborted on it.
-      expect(wrongTray.aborted).toBe(false);
-      expect(wrongTray.found).toBe(false);
-      expect(a.job.status).toBe("downloading");
-      expect(b.job.status).toBe("downloading");
+        // Naming the ORPHAN must not touch our live transfer…
+        const orphan = cancelDownloadJob(a.job.id, "orphantrayid0001");
+        expect(orphan.aborted).toBe(false);
+        expect(a.job.status).toBe("downloading");
 
-      const right = cancelDownloadJob(a.job.id, a.job.trayId);
-      expect(right.aborted).toBe(true);
-      await a.settled;
-      expect(a.job.status).toBe("cancelled");
-      // The sibling is untouched.
-      expect(b.job.status).toBe("downloading");
-      hoisted.resolvers.find((r) => r.url === URL_B)!.resolve("/M/loras/other.safetensors");
-      await b.settled;
+        // …and a tray id that names NOTHING must abort nothing at all.
+        const bogus = cancelDownloadJob(a.job.id, "nosuchtrayid0000");
+        expect(bogus.found).toBe(false);
+        expect(bogus.aborted).toBe(false);
+        expect(a.job.status).toBe("downloading");
+        // The refusal still lists what the id DOES name, so the caller can retry.
+        expect(bogus.candidates?.map((c) => c.trayId).sort()).toEqual(
+          [a.job.trayId, "orphantrayid0001"].sort(),
+        );
+
+        // Only the exact tray id aborts the live one.
+        const right = cancelDownloadJob(a.job.id, a.job.trayId);
+        expect(right.aborted).toBe(true);
+        await a.settled;
+        expect(a.job.status).toBe("cancelled");
+      } finally {
+        setProgressDir("");
+        await fsRm(dir, { recursive: true, force: true });
+      }
     });
 
     it("an ambiguity refusal NAMES the candidates instead of leaving the caller stuck", async () => {
--- a/src/__tests__/services/download-jobs.test.ts
+++ b/src/__tests__/services/download-jobs.test.ts
@@ -599,6 +599,43 @@ describe("download job registry", () => {
       }
     });
 
+    it("a LIVE foreign download is never hidden behind this session's settled row for the same URL", async () => {
+      // `trayId` is a hash of the URL, so it does NOT separate sessions: this
+      // session's finished record and ANOTHER session's still-running download of
+      // the same URL to the same destination collapse to one (id, trayId) key.
+      // Reporting the settled one would hide a running transfer behind a stale
+      // verdict — worse than the ambiguity this selector exists to surface, and it
+      // would make `tray_id` weaker than the plain by-id lookup in that state.
+      const dir = await mkdtemp(pathJoin(tmpdir(), "djobs-822-"));
+      setProgressDir(dir);
+      try {
+        const a = await startDownloadJob(URL_A, "checkpoints");
+        // Our own download is cancelled and settles…
+        cancelDownloadJob(a.job.id, a.job.trayId);
+        await a.settled;
+        expect(a.job.status).toBe("cancelled");
+
+        // …while another session is still downloading the SAME url to the SAME
+        // destination (same id, and — because trayId hashes the url — same trayId).
+        await writeForeignJobRecord(dir, {
+          id: a.job.id,
+          trayId: a.job.trayId,
+          progressId: "foreign-live-prog",
+          url: URL_A,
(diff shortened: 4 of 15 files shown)


artokun:
Merged. Recording one gate finding I **dismissed**, with reasoning, since that is a judgement call and should not be silent.

The final independent gate returned NO-SHIP on a single P0: that `:2150`/`:2185` assert ownership only *before entering* `discardRejectedPayload`, leaving the helper's separate awaited destructive acts unguarded.

I verified this against the merged code and it is **not correct**:

- `rmOrTruncate` re-proves before the `rm` (via `rmOrLog`) **and again before the truncate fallback** — with an explicit comment that the failed `rm` was itself an await, so the entry check is already stale.
- `discardRejectedPayload` calls `rmOrTruncate` per artifact, so the sidecar's destruction carries its own check rather than borrowing the payload's.
- `discardStagedAfterRefusal` calls `guarded.assertOwned()` immediately before each `safeRm`.
- `cleanupGuards` only yields the empty `exclusivelyOwned` check when the staged file is genuinely unshared; shared paths get `guardStagedPair`.

The gate's own closing sentence names the real residue — *"the scheduling boundary between the check's last read and the syscall it authorises"* — which the code documents at the `GuardedPath` definition as irreducible without OS-level locking. That is a known, stated limitation, not a defect this PR introduced.

Reporting it because a dismissed finding that nobody can audit is worse than the finding. If I have this wrong, the place to look is `cleanupGuards` → `guardStagedPair` and whether `stagedFileIsShared` can be false for a path another process can in fact name.

Issue #803, opened 2026-08-04
After ComfyUI reconnect, graph binding desyncs and the documented recovery (panel_open_workflow on the ACTIVE workflow) cannot rebind — agent left with no working recovery path


artokun:
### What happened

Session resumed after a mid-task connection drop (ComfyUI was restarted). The panel then reported the active workflow correctly, but every graph command was pointed at a stale/empty graph, and none of the documented recovery paths worked.

Sequence, in order:

1. `panel_graph_outline` →
   `Error: The live graph is out of sync with the active workflow: the workflow reports 0 node(s), but the canvas is bound to a different graph. A load, tab switch, or reconnect left this command pointed at the wrong canvas, so it was NOT applied. Re-open the active workflow tab (panel_open_workflow) or reload the panel to rebind the graph, then retry.`

2. `panel_list_workflows` succeeded and reported a healthy state — `active_confirmed: true`, active = `workflows/Untitled 2026-08-04 06-15-58.json`, that same workflow present in `open[]` with `active: true, persisted: true`. So from the list tool's perspective nothing was wrong.

3. Followed the error's own instruction: `panel_open_workflow({path: "workflows/Untitled 2026-08-04 06-15-58.json"})` — i.e. re-opening the workflow that is ALREADY the confirmed-active tab →
   `Error: workflow_open could not prove that the active canvas was rebound to the requested workflow. The open may have switched the active workflow; reload the panel before graph edits.`

   This is the core friction: the error message in step 1 explicitly prescribes `panel_open_workflow` as the recovery, but for the already-active workflow that call cannot satisfy its own #661 proof requirement, so the prescribed recovery is guaranteed to fail. `panel_list_workflows.last_open` also showed an earlier `workflow_open` with `applied: "unknown"` and the same "could not prove" error, so this had already happened once before in the session.

4. `panel_graph_outline` again → identical out-of-sync error.

5. Followed the error's second instruction: `panel_reload({scope: "frontend"})` → `soft reload (frontend) scheduled`.

6. After the reload, the tab stopped answering entirely. Every subsequent panel command times out:
   `Error: Panel tab wf:workf did not reply to "graph_outline" within 20000 ms — the ComfyUI tab may be backgrounded or frozen`
   `Error: Panel tab wf:workf did not reply to "workflow_list" within 6000 ms — the ComfyUI tab may be backgrounded or frozen`
[...]

Issue #772, opened 2026-08-03
panel_run run-to-node graph-stamp race should safely retry once


artokun:
### What happened / root cause
`panel_run(to_node_id=650)` hit the panel's workflow-stamp race guard after a prompt edit and read-only graph checks. The panel certified that the scoped run was not applied and that nothing was queued, but the orchestrator returned the transient race as a terminal failure instead of rebuilding the run-to-node scope once. This forces an awkward manual retry even though the rejection is explicitly safe to retry.

### Steps to reproduce
1. On a live canvas, call `panel_set_widget` on a prompt text node.
2. Call `panel_set_todo` and/or `panel_get_errors`.
3. Immediately call `panel_run({to_node_id:650})` for a PreviewImage output.
4. Observe the graph-stamp rejection below; no render enters the queue.

### Exact error (scrubbed)
ComfyUI refused to queue the workflow: run-to-node scope for node 650 was NOT applied: the workflow graph CHANGED after the run was queued - the deferred dispatch would render a modified workflow, not the one that was scoped. Nothing was queued - refusing to fall through to a full-graph execution (#556).

### Fix
applied locally: no  upstream-only: yes
One focused patch attempt against the running dist failed cleanly because the installed bundle's encoded source text did not match the patch context; no file was changed. Precise proposed change: in the `panel_run` handler in `dist/orchestrator/panel-tools.js` (and source), when `args.to_node_id` is set and the first `graph_run` result is an error containing BOTH `workflow graph CHANGED after the run was queued` and `Nothing was queued`, rebuild/reissue the same scoped `graph_run` exactly once. Never retry any other failure or any response that does not explicitly certify nothing was queued. Add a regression test covering success-on-second-dispatch and failure-without-retry for all other errors.

### Environment
Windows 11; ComfyUI 0.30.0; NVIDIA GeForce RTX 5090 32 GB; comfyui-mcp 0.49.3; panel 0.11.38.

artokun:
Additional duplicate report: after several successful `panel_set_widget` calls and a fresh `panel_graph_outline`, `panel_run({to_node_id: 10})` was rejected before queueing with the same “workflow graph CHANGED after the run was queued” / “Nothing was queued” run-to-node error. No concurrent user edits were made. Environment: Windows 11, ComfyUI 0.30.1, RTX 4070 SUPER; comfyui-mcp 0.49.7; panel 0.11.38.

artokun:
[...]

Issue #770, opened 2026-08-03
Panel graph reads stay wedged after workflow retarget


artokun:
### What happened / root cause
`panel_graph_outline` rejects the workflow currently visible in the panel with a workflow instance mismatch. Calling `panel_set_workflow_target({mode:"current"})` has previously returned success without refreshing the stale instance binding, so subsequent graph reads remain blocked. The likely root cause is that the bridge retains or validates against a stale workflow UUID instead of rebinding to the active canvas; read-only commands are consequently gated by an instance check they cannot recover from.

### Steps to reproduce
1. Open or replace a workflow tab in ComfyUI.
2. Invoke `panel_graph_outline({})`.
3. Observe the workflow instance mismatch.
4. Retarget to current and retry; the stale mismatch can persist.

### Exact error (scrubbed)
`Error: workflow instance mismatch: this command targets a different workflow than the active canvas — the workflow was switched or replaced after it was issued. Re-select the intended workflow (panel_open_workflow) or re-target with panel_set_workflow_target({mode:"current"}), then retry.`

### Fix
Applied locally: no. Upstream-only: yes. A prior single local attempt during this session did not restore the bridge and was reverted. The target binding needs to be refreshed atomically from the active workflow when `mode=current` is set, before applying the UUID guard; alternatively, read-only inspection should resolve the active canvas directly. Add a regression test covering workflow replacement followed by retarget-current and `panel_graph_outline`.

### Environment
macOS / ComfyUI 0.29.0 / Apple MPS 16 GB / comfyui-mcp 0.49.3 / panel 0.11.38.

artokun:
Additional reproduction on Windows 11 / ComfyUI 0.29.2: after a large manual same-tab graph replacement (~170 node changes), the identical instance-mismatch error wedged every `panel_*` call, including `panel_list_workflows`. `panel_set_workflow_target({mode:"current"})` reported success but did not rebind; `panel_open_workflow`, `panel_reload`, and a full ComfyUI restart also did not recover. The reporter found only a browser hard-refresh effective. Reporter versions: comfyui-mcp 0.48.22, panel 0.11.35 (behind current 0.49.3 / 0.11.38).

artokun:
[...]

Issue #716, opened 2026-08-02
Stale workflow UUID after reconnect/open/re-pin rejects panel_run


artokun:
### What happened / root cause
After a ComfyUI restart, the panel reported active_confirmed:false. I called panel_open_workflow on the intended already-open saved tab, then panel_set_workflow_target(mode:"pinned") with that exact path. The next panel_run was rejected by the panel's workflow UUID fence. The active tab/path still matched and panel_query_graph had just read the correct graph, so the command appears to have been stamped with a stale workflow instance UUID after reconnect/open/re-pin.

### Steps to reproduce
1. Restart ComfyUI while a saved, modified workflow tab is open.
2. Rebind mode:"current".
3. panel_list_workflows reports the intended active tab but active_confirmed:false.
4. panel_open_workflow(path:<same active saved workflow>).
5. panel_set_workflow_target(mode:"pinned", path:<same path>).
6. panel_run({}).

### Exact error
workflow instance mismatch: this command targets a different workflow than the active canvas — the workflow was switched or replaced after it was issued. Re-select the intended workflow (panel_open_workflow) or re-target with panel_set_workflow_target({mode:"current"}), then retry.

### Fix
Applied locally: no. Upstream-only: yes.
The orchestrator should refresh/replace the session workflow UUID after a successful panel_open_workflow and panel_set_workflow_target re-pin, using the active workflow UUID returned/read after the open. It should not stamp the next graph command with the pre-restart/pre-open UUID. Preserve the panel's fail-closed UUID fence.

Temporary workaround: panel_set_workflow_target({mode:"current"}) immediately before retrying panel_run.

### Environment
Windows 11; ComfyUI 0.29.2; RTX 5090 32 GB; comfyui-mcp 0.49.0; panel 0.11.34.

artokun:
Recurrence reported on current releases: comfyui-mcp 0.49.3 and comfyui-mcp-panel 0.11.38 (macOS, ComfyUI 0.29.0, MPS). After the active canvas receives a new workflow UUID while its socket remains reachable, `panel_graph_outline` and `panel_list_workflows` fail with `workflow instance mismatch`; `panel_set_workflow_target({mode:"current"})` returns success but does not recover the session. `panel_open_workflow` also could not prove the active canvas was rebound. The report notes the loaded package as 0.49.3 while the panel-injected environment line showed 0.48.18.

wuraaang:
Additional reconnect symptom reproduced on a current local setup:

[...]

PR #833, opened 2026-08-04
graph binding after reconnect/retarget: a documented recovery that could not recover (#803, #770, #772)

artokun:
## Issues in this cluster

- #803 — graph binding desyncs after reconnect; **all three documented recoveries fail**
- #770 — panel graph reads stay wedged after a workflow retarget
- #772 — `panel_run` run-to-node stamp race returned as terminal
- #716 — stale workflow UUID after reconnect/open/re-pin (closed by #722; **recurring**)

## The defect

`panel_set_workflow_target({mode:"current"})` is documented — by the panel's own error text, by three other orchestrator messages, and by its own tool description — as the way out of a wedged graph binding. It could not do it.

The per-command workflow-instance fence (#570) is written **only from a panel hello**. A hello is not the only way the active canvas changes identity: a workflow replaced in place on a still-open socket, a frontend soft reload, or a reconnect that re-registers under a different routing key all give the canvas a new workflow uuid with no new hello for the key the session holds. The stamp then names an instance that no longer exists, the panel's fence correctly refuses every stamped command, and **nothing in the orchestrator could replace it** — the #716 open/pin refresh paths require a caller-supplied *saved* path to corroborate, so they cannot fire for `mode:"current"` at all, and cannot fire for an unsaved (`tmp:`) canvas under any mode.

Meanwhile `rebindToActiveTab()` is a no-op whenever the bound tab is still reachable — the dominant reported case. So the call re-pointed nothing, and returned `"Following the user's current workflow tab."` A documented recovery that could not recover, reporting success.

## What this fixes

**#770 — fixed.** `rebindWorkflowFence()` re-derives the fence from the panel's own live `workflow_list.active`. No path corroboration, because `mode:"current"` has no caller-supplied target to corroborate — the live active canvas *is* the target. That is also why it works for an unsaved canvas, which every `tmp:` report hit. #570 is not weakened: every command issued after this point still fails closed if the user switches again.

**#803 — fixed on the orchestrator side.** The recovery now recovers, and what we tell the user matches what the code does. One verdict became five statuses (`refreshed` / `already_current` / `unreadable` / `no_identity` / `rejected`, plus `adopt_error`), surfaced as `graph_binding` (`bound` / `reads_only` / `unverified` / `not_recovered`) and `graph_binding_status`. When the binding is **not** restored the tool now fails — after first disclosing that the target write *did* apply, so nothing that succeeded is retracted. Every remedy was read end to end and checked against the caller's actual state; the ones that could not be followed were replaced, not softened.

**#772 — fixed.** `panel_run` re-issues a scoped run **exactly once**, only when the panel both reports the run-to-node stamp race *and* certifies nothing was queued. Structured queue evidence (`queued:true`, a `prompt_id`) vetoes the prose; an unparseable or `isError` reply is an unknown and is never retried; the gate reads our own `to_node_id`, never panel prose. Both the success and failure paths disclose the second dispatch, and claim only what is certified.

**#716 — not re-fixed, and not claimed.** #722's saved-path open/pin refresh is untouched. The recurrence reports are the `mode:"current"` and unsaved-canvas gaps above, which this closes. Left closed.

Also, from #803's own notes: the bridge's no-reply timeout named `tabId.slice(0, 8)`, rendering **every** workflow tab as the identical `wf:workf` — two tabs timing out were indistinguishable in a transcript and the id read like a corrupted routing key. It now names the full id. That change alone would have silently disabled the whole `workflow_open` verify-after-timeout recovery for any workflow whose filename contains a space (ComfyUI's default name has two), because `isAckTimeout` segmented the id with `\S+`; that regex is fixed and start-anchored in the same commit.

## The property, not the instances

Every fix here is the same fold: **"could not determine X" reported or treated as "determined X is not the case"** — in both directions. It was found in eight places, several only after the earlier fix had been applied one level up:

- a fence rebind that reported success without looking;
- a **read** of the existing fence that collapsed "the resolver threw" into "there is no fence" — fixed in the orchestrator, then found *still collapsed in the bridge*, which is the only thing that can see the stamp;
- a **capability** probe that collapsed "could not inspect" into "your panel lacks the write fence" — a claim about the user's install built from our own failed probe;
- `already_current` claiming agreement it had verified against the *previous* tab, after the probe's own retry moved the session;
- an adoption **throw** narrated as a refusal, when a refusal proves the old fence survived and a throw does not.

## Self-gate

Six rounds of adversarial codex review (`gpt-5.6-terra`). Rounds 1–5 produced 19 findings — 2 P0, the rest P1 — all fixed or answered on the record; round 6 returned **PASS** with one P2 (a test that probed a nonexistent tab and so never reached the code it named), now rewritten against a live tab.

Every fix is mutation-verified: the code was broken and the matching test confirmed failing, 31 times. Tests assert the **reason**, not just the state.

Full suite green at `--maxWorkers=2` (255 files, 5169 tests) — this machine is loaded and full-speed runs flake in a rotating cast of unrelated files, so that is the honest number. Plus `build`, `lint`, `check:vocabulary`, `vocab:export` (no diff). Committed blobs scanned for control bytes: clean, all four files git-TEXT.

## Deliberately not fixed

- **`panel_open_workflow` on the already-active workflow cannot satisfy its own proof requirement** (#803 step 3). The `"could not prove that the active canvas was rebound"` text is produced by **comfyui-mcp-panel**'s `workflow_open` handler, not this repo. Panel-side.
- **`panel_reload({scope:"frontend"})` leaving the tab permanently unresponsive** (#803 steps 5–6). Panel-side. Every remedy this PR writes now explicitly steers *off* `panel_reload` and names a manual browser refresh, which is the only thing reporters found effective.
- **`panel_search_nodes` rejected by the workflow-instance fence** (reported on #770). Manager registry search is not graph-scoped; the orchestrator stamps the frame but the decision to fence it is the panel's `activeWorkflowFenceApplies`. Panel-side — but worth checking whether the orchestrator should stop stamping non-canvas commands at all.
- **A compare-and-set on the active canvas.** Between the `workflow_list` reply and the stamp write the user can switch the same tab, so an adopted fence can already be stale. The failure direction is safe — the panel refuses the next command; it never writes the wrong workflow — so the defect was the *claim*, which is now qualified ("the canvas the panel reported active a moment ago", and a switch in that instant fails closed and is cleared by calling again). A real fix needs a panel-supplied active-canvas generation token: a **comfyui-mcp-panel protocol change**. The same window already exists in the pre-existing #716 pinned path.
- **Per-tab operation tokens on target changes.** Two sessions bound to one tab share the `WorkflowTargetStore`, and a slow call can clobber a later pin. That race is **pre-existing** and is wider through the pre-existing `awaitReachable` await (~20s) already ahead of the store write than through anything here. This PR does not widen it — the live-canvas probe was deliberately moved to *after* the store write and push — but it does not close it either.
- **`resolveEffectiveComfyUIBase()` / `comfy-cli.ts` (#490)** — reopened and owned elsewhere. Not touched.

## Only a live rig can confirm

The wedge itself is a reconnect/replace race against a real ComfyUI. The unit tests drive the exact reply shapes, but nothing here proves the panel's `workflow_list.active.workflow_uuid` is populated for every real wedge — particularly on an unsaved `tmp:` canvas after a soft reload, which is where several reports originate. If it is ever absent there, this lands in `no_identity` and reports that honestly rather than pretending, but the recovery would not complete. The shared ComfyUI on this rig has unrelated unsaved tabs open, so it was not restarted for destructive validation.

### DIFF
--- a/src/__tests__/orchestrator/panel-tools.test.ts
+++ b/src/__tests__/orchestrator/panel-tools.test.ts
@@ -4706,6 +4706,29 @@ describe("panel-tools: mode:'current' re-derives the workflow command fence (#77
   // uncorroborated top-level `active` "is never a safe source for replacing a
   // command fence". These cases hold the rebind to it.
 
+  it("ADOPTS when the identities differ only in path SYNTAX — a contradiction check must not refuse a real match", async () => {
+    // The contradiction check compares through the same canonicalizer the rest of
+    // this file uses. Comparing raw strings would call "./workflows/a.json" and
+    // "workflows/a.json" a contradiction and refuse a legitimate recovery — this
+    // defect class pointed the other way, which is precisely what the check was
+    // added to avoid committing (codex gate).
+    const { bridge, refresh, tab } = fenceBridge({
+      fence: STALE,
+      active: {
+        path: "./workflows/a.json",
+        routing_key: "wf:workflows/a.json",
+        workflow_uuid: LIVE,
+      },
+      workflows: [
+        { path: "workflows/a.json", routing_key: "wf:workflows/a.json", active: true },
+      ],
+    });
+    const { res } = await setCurrent(bridge, tab);
+
+    expect(refresh).toHaveBeenCalledExactlyOnceWith(tab, LIVE);
+    expect(res.isError).toBeFalsy();
+  });
+
   it("REFUSES a PARTIAL contradiction — one matching field never outvotes a conflicting one", async () => {
     const OTHER = "33333333-3333-4333-8333-333333333333";
     const { bridge, refresh, tab, currentStamp } = fenceBridge({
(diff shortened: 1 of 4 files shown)


artokun:
**Recovered the fix for the gate's P0.** It existed only in a local worktree and was lost when the spend limit killed the agent mid-run; the branch has now been advanced to include it.

## Why this was safe despite not being a fast-forward

The preserved ref looked "diverged" (9 commits on the branch, 11 on the ref) but it was not: it is the **same commit sequence, rebased** onto a newer `main`, plus one extra commit. I verified that by diffing the two tips — the only differences were `.env.example` / `README.md` (which the ref picked up from `main`'s #861) and exactly the 343 lines of the extra commit. A clean superset, so nothing on the branch was lost.

Force-pushed with `--force-with-lease` pinned to the branch's real head (`3561ae2`), so a concurrent push would have been rejected rather than clobbered.

## What was recovered

**`f2e39c3` — `fix(#770): corroborate the active record before adopting its uuid`**, which is the P0 from the independent gate:

> `rebindWorkflowFence()` adopted `responseWorkflowUuid(parsed.active)` with no corroboration that the top-level `active` record belongs to the routed tab. A mixed or stale `workflow_list` reply carrying another canvas's valid UUID passed the shape validator and overwrote this tab's stamp — the tool then reported `graph_binding: "bound"` while the next graph command was stamped for the **wrong instance** and refused. It reintroduced a wedge *while falsely reporting recovery*.

The decisive point was that the existing pin path already treated an uncorroborated top-level `active` as unsafe for a command-fence refresh, so this new path was bypassing a rule the codebase had already arrived at.

## Verification

- **255 files / 5192 tests** pass at full concurrency after merging current `main`; `tsc` clean; zero control bytes; no merge conflicts.

## Still needed

The P0 fix has **not** been independently gated — I have confirmed it is present and that the branch builds and passes, not that it is correct. The specific things to check on review are the two the gate named: that the corroboration ties the record to *this* tab on every path, and that the test no longer **specifies** the unsafe shape (it previously supplied only a top-level `active` with no corroborating `workflows` record and expected the stamp write, so a stale/mixed-active regression would have satisfied it).

Issue #808, opened 2026-08-04
Fallback when comfy-cli is unavailable for Git node install


artokun:
### What happened / root cause
install_custom_node documents that local installs prefer comfy-cli when available and fall back to Manager HTTP when CLI is unavailable. With useCmCli=true on a local workspace where comfy-cli is not installed, it returned NODE_MANAGEMENT_ERROR instead of falling back.

### Steps to reproduce
1. Local ComfyUI workspace without comfy-cli on PATH or in the workspace venv.
2. Call install_custom_node with source=git, useCmCli=true, and a valid GitHub URL.
3. The operation stops before trying a supported fallback.

### Exact error (scrubbed)
comfy-cli node install failed: comfy-cli was not found. Install comfy-cli>=1.11.1 and ensure comfy is on PATH, set COMFY_CLI_PATH, or install it in the selected ComfyUI workspace's .venv.

### Fix
Applied locally: no
Upstream-only: yes

Precise change needed: if the CLI probe reports unavailable before any mutation was submitted, automatically retry the same install through Manager HTTP; for an unregistered git URL on Manager v4, provide a direct clone fallback scoped to <workspace>/custom_nodes/<repo-name>.

### Environment
Windows 10; ComfyUI 0.30.0; RTX 3090 24GB; comfyui-mcp 0.49.2; panel 0.11.36.

artokun:
Fixed in #834 (merged to `main` as 7dfdc2d).

The independent gate on that PR also surfaced four P0s in the same area, fixed on the same branch: destructive paths (named snapshot save/cancel, comfy-cli uninstall/disable, the git clone fallback) could act on a stale local install while every check around them addressed the remote server; the "already installed" verdict was inferred from post-operation disk state, hiding a successful registry-ZIP install; and the uninstall claimed success from ComfyUI-Manager's list alone, without checking whether the directory survived.

The underlying resolver bug is tracked separately as #490 — it has twelve callers and needs its own audited change, so #834 refuses to act on an install it cannot identify rather than patching the resolver in place.

Issue #797, opened 2026-08-03
Generic Manager presence gate never inspects disk — registry-ZIP installs of any non-panel pack still get "not installed locally" from code that only asked the registry


artokun:
Split out of PR #793 (the #771/#766/#769/#774/#784 recovery cluster), where this was fixed for the **panel** and left live for every other pack.

### The gap

`install_panel(action:'update')` originally delegated to the generic node-pack update, whose presence gate asks ComfyUI-Manager's installed/registry list **keyed on registry id** and never looks at disk. `panelStatus`, by contrast, scans `custom_nodes` and reads `pyproject.toml`.

For a **registry-ZIP install** (no `.git`) the two disagree — and the side asserting *"it is not installed locally"* had never checked the disk. That produced the self-blocking loop in #771: the version gate tells you to run `install_panel`, and `install_panel` reports the pack isn't installed, while `install_panel(action:'status')` in the same session correctly reports it **is** installed at a named directory with a version.

### What PR #793 did and did not fix

**Fixed for the panel:** a panel-specific wrapper now supplies disk evidence, so the panel path no longer asserts an absence it never verified.

**Also fixed generally — the message is now truthful:** the generic gate states that it consulted only the Manager's list, and an unreadable list now refuses as *unverifiable* rather than being treated as absence (previously `.catch(() => [])` turned a failed request into proof of absence — see #796).

**Still open:** the generic gate itself does not inspect disk. Any *other* pack installed as a registry ZIP hits the original defect — "not installed locally" from code that only asked a registry.

### Suggested fix

Give the generic presence check the same disk evidence the panel wrapper now supplies: resolve the pack directory under `custom_nodes` and read its `pyproject.toml` before concluding absence. The panel wrapper in #793 is the working reference.

### Caution

Keep the three states distinct — **present**, **absent**, and **could not determine**. The original bug was a failed list read collapsing into "absent", and #796 documents seven instances of that same collapse across this codebase in a single day. An unreadable list or an unreadable directory must refuse as unverifiable, never assert absence.

### Related
- #771 — the reported symptom.
- #793 — the panel-side fix and the truthful message.
- #796 — the recurring "could not determine folded into determined-not" defect class.

artokun:
Fixed in #834 (merged to `main` as 7dfdc2d).

[...]

Issue #789, opened 2026-08-03
panel_search_nodes Impact Pack result cannot be installed as returned


artokun:
### What happened / root cause
panel_search_nodes returned the official Impact Pack result with id "https://github.com/ltdrdata/ComfyUI-Impact-Pack". Passing that returned id to panel_install_node with version:"latest" was translated to ComfyUI-Impact-Pack@8.28.3, but the built-in Manager rejected it as unavailable. The pack was not installed. Retrying the same official repository through panel_install_node(repository:<url>, version:"nightly") installed and verified successfully.

This creates a confusing search→install failure for legacy/repository-style search results. The panel tool knows the original result is a GitHub URL and could use it as the repository fallback automatically when the registry version cannot be resolved.

### Steps to reproduce
1. panel_search_nodes({query:"comfyui-impact-pack"})
2. Take the returned id "https://github.com/ltdrdata/ComfyUI-Impact-Pack".
3. panel_install_node({id:<returned id>, version:"latest", channel:"default", mode:"remote"})
4. Poll panel_node_queue_status.

### Exact error (scrubbed)
recent_failures:
kind: install
id: ComfyUI-Impact-Pack
result: "not available node: ComfyUI-Impact-Pack@8.28.3"

The install call initially returned queued:true, installed:false, verified:false, pending:true.

### Fix
Applied locally: no
Upstream-only: yes

Precise change requested: when panel_search_nodes returns a repository URL / legacy entry, preserve that repository URL through panel_install_node. If the registry-id install resolves to "not available node", automatically retry once using repository:<original URL>, version:"nightly", then perform the existing panel_list_nodes verification. Alternatively return a stable registry id/version that the Manager can install.

Verified workaround: panel_install_node({repository:"https://github.com/ltdrdata/ComfyUI-Impact-Pack", version:"nightly"}) returned installed:true, verified:true and the pack appeared in panel_list_nodes.

### Environment
Windows 11
ComfyUI 0.30.0
Python 3.12.10
PyTorch 2.13.0+cu130
GPU NVIDIA GeForce RTX 5060 Ti, 16 GB VRAM
comfyui-mcp 0.49.3
panel 0.11.35

artokun:
Fixed in #834 (merged to `main` as 7dfdc2d).

[...]

Issue #775, opened 2026-08-03
Local node cleanup lacks disable action and named snapshots are misclassified as remote


artokun:
### What happened / root cause
A reversible custom-node cleanup on a confirmed local ComfyUI install hit two related MCP surface gaps:
1. node_snapshot(action:"save", name:"pre_cleanup_2026_08_03") returned NODE_SNAPSHOT_ERROR saying named snapshots require a local install path and the session was in remote (--comfyui-url) mode, even though get_environment reports a trusted local workspace at C:/Stability Matrix/Data/Packages/ComfyUI and workspace/list_extra_paths resolve that same local root.
2. The custom-nodes and panel tool catalogs expose install/update/reinstall/fix/list, but no disable/enable/uninstall action. The official comfy-cli binary was also absent. A safe disable-first audit therefore required a manually verified Move-Item into custom_nodes/.disabled.

### Steps to reproduce
1. Connect to local ComfyUI at 127.0.0.1:8188 with saved default workspace set.
2. Confirm get_environment returns deploy_environment local-git and a local workspace_path.
3. Call node_snapshot action=save with a name.
4. Observe false remote-mode rejection.
5. List custom-nodes and panel Manager tools; observe no disable/enable/uninstall mutation.
6. Omit the snapshot name; Manager HTTP reports snapshots unsupported, leaving no reversible tool path.

### Exact error
{"error":"NODE_SNAPSHOT_ERROR","message":"Saving a named snapshot requires a local ComfyUI install path, which is unavailable in remote (--comfyui-url) mode..."}

### Fix
Applied locally: no. Upstream-only: yes.
Precise changes needed:
- Resolve named node_snapshot saves from the same persisted/default local workspace resolver used by get_environment/list_extra_paths instead of treating an API-targeted local connection as remote.
- Add Manager-backed disable_custom_node / enable_custom_node / uninstall_custom_node actions, with exact target validation and post-state verification. Prefer reversible disable as the cleanup default.
- When Manager HTTP lacks snapshots and no comfy-cli exists, return a structured rollback alternative rather than leaving no supported path.

Temporary workaround: validate every absolute source/destination under custom_nodes, then move exact pack directories into custom_nodes/.disabled and record a manifest.

### Environment
Windows 11; local ComfyUI 0.30.0; Python 3.12; PyTorch 2.13.0/CUDA 13.0; RTX 5060 Ti 16 GB; comfyui-mcp 0.49.3; panel 0.11.35.

[...]

PR #834, opened 2026-08-04
node search + install: registry packs that exist on disk read as absent, results that cannot be installed as returned

artokun:
## Issues in this cluster

- #797
- #773
- #789
- #808
- #775


artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 4 P0.** No P1, no P2. Bytes clean, no binaries.

**The vocabulary ratchet is intact** — I checked it specifically because this diff touches `vocabulary.ts`, and the check passed cleanly: `docs/design/tool-surface.txt` gained exactly three lines (`disable_custom_node`, `enable_custom_node`, `uninstall_custom_node`), main's 190 lines are an **exact prefix** of the branch's 193, and the committed `BASELINE_SHA256` matches the committed file byte-for-byte. `MAX_TOOLS` 151→154 with the reason stated. That is the documented append-update-explain workflow followed exactly.

## Two of the four are ONE pre-existing bug this PR makes load-bearing

`resolveEffectiveComfyUIBase()` in `workspace-env.ts` returns `config.comfyuiPath` **before** it checks remote mode — while its own docstring promises *"Never returns a local workspace in remote mode."* `resolveLiveComfyUIBase()` directly below gets it right, which makes the omission look accidental.

- **`node-snapshots.ts:269` / `:440`** — named snapshot save/cancel now trust that resolver. With `--comfyui-url` against a remote server and a stale local `COMFYUI_PATH`, **save fetches the remote server's snapshot and writes it under an unrelated local install**, and **cancel can delete that local install's pending restore file**.
- **`node-management.ts:2940`** — explicit `useCmCli: true` has the same shape: pre/post Manager checks address the *remote* instance while `comfy-cli uninstall`/`disable` mutates the **local** one. The reply then says only that the remote operation did not take effect — **the local destructive action is not disclosed at all.**

**The resolver bug is pre-existing on `main`, not introduced here** — but this PR is the first to route *destructive* operations through it, which turns a latent wrong answer into deleting the wrong install. That is the #369 harm class exactly.

I have filed it separately: the triage worker matched it to **#490, previously closed as fixed, and reopened it** — the bug survived its own fix. It is called from 12 sites, so please **do not patch it inside this PR**; fixing the shared resolver needs its own change with all callers audited (`comfy-cli.ts:16` already reads `config.comfyuiPath ?? resolveEffectiveComfyUIBase()`, which hints some callers assume today's behaviour). What #834 *should* do is **refuse when the target install cannot be established** rather than accepting whatever the resolver hands back — an honest "I cannot tell which install this would modify" beats mutating the wrong one.

## The other two

**`node-management.ts:2174` — the "ALREADY installed" verdict is decided from post-queue disk state.** A pack absent before the call can be installed as a registry ZIP while Manager still returns an empty installed list; the post-check then sees the newly created directory and reports the queued install "resolved to nothing new." That **fabricates the pre-operation state and hides a successful fresh install** — the user is told nothing happened when their pack was just installed. The pre-state has to be captured *before* the operation; it cannot be inferred afterwards.

**`node-management.ts:2848` — the uninstall success path checks only Manager's installed list.** If Manager drops its tracking entry but leaves `custom_nodes/<pack>` on disk after a partial uninstall, it reports `Uninstalled` while **ComfyUI will still load the directory on restart.** This is the one that most needs the new disk-presence machinery and does not use it — and it is the exact "present on disk vs tracked by Manager vs enabled" three-way distinction this PR otherwise handles well elsewhere. A destructive postcondition deserves the strongest available evidence, not the most convenient.

## Note on merging
Checks are older than several merges to `main` (including the 0.49.6 release). **A green check is only green for the `main` it ran against** — that staleness took the panel's `main` red tonight. Verify on the actual merged tree.

Issue #798, opened 2026-08-03
Audit writes whose durability is load-bearing: a buffered marker/journal can be lost while the action it records survives


artokun:
Raised by the agent that fixed PR #793, which found and fixed one instance of this and flagged the general shape.

### The defect found

In the panel swap path, the recovery journal was a **buffered** write issued immediately before the first `rename()`. A power loss or hard kill in that window could persist the *rename* while losing the *record of it* — leaving `custom_nodes` mid-swap with nothing on disk saying so, and the repair path unable to act because it had no journal to read.

The crash-recovery mechanism failing precisely in the scenario it exists for. Fixed in #793 by writing through a descriptor and `fsync`-ing before the rename.

### The general shape

> A marker, journal, or lock file is written and then **immediately acted upon**, where the action is externally visible (a rename, a spawn, a network mutation) and the record is the only thing that makes the action recoverable.

If the write is buffered, the action can outlive the record. Ordering in the source is not ordering on disk.

### Neighbours worth auditing

The #793 agent identified these and deliberately did **not** touch them, since they were out of scope and none currently sit in front of a directory rename:

- `panel-pin-guard`'s **pending-op markers** — written before a Manager operation the process cannot serialize, precisely so a later pin can reason about an operation still in flight.
- `panel-pin-guard`'s **lock file**.

Both are written-then-acted-on, and the pending-op marker's whole purpose is to survive a window the process does not control.

### Suggested work

A small self-contained pass: find writes whose durability is load-bearing for a subsequent externally-visible action, and `fsync` those specifically. Not a blanket `fsync` everywhere — that costs latency for writes whose loss is harmless — the criterion is whether losing the record while keeping the action leaves the system unrecoverable.

Worth a short comment at each such site saying why it is synced, so a later refactor doesn't quietly buffer it again.

### Related
- #793 — where this was found and fixed for the panel swap journal.
- #796 — the recurring defect class filed today; this is adjacent but distinct: there the *state* was unrepresentable, here the *record* is losable.

artokun:
Fixed on main by PR #836 (cd28da6): src/utils/durable-write.ts (temp+fsync+rename+dir-fsync, header names #798) applied to the pending-op marker, pin store, panel-settings, and lora-catalog; the panel-op lock pid record is fsynced at acquire. Tests at panel-pin-guard.test.ts:725. Unreleased.

Issue #765, opened 2026-08-03
panel_reload orchestrator leaves comfyui-mcp service modules stale


artokun:
### What happened / root cause
panel_reload(scope: "orchestrator") resumed the agent but did not reload an edited comfyui-mcp service module already loaded by the MCP server process. The tool description says to use this scope after backend/orchestrator code changes, so the observed behavior leaves the running panel tool on stale code.

### Steps to reproduce
1. Apply a focused local patch to the running comfyui-mcp dist/services/workflow-converter.js.
2. Verify the patched module directly with node import; the regression test passes.
3. Call panel_reload({scope:"orchestrator"}).
4. After automatic resume, call panel_strip_workflow on the same graph.
5. Observe the old conversion result is unchanged, proving the MCP server retained the pre-patch module.

### Exact symptom
Before reload: CLIPLoader.type "minimax" was rewritten to "stable_diffusion".
After panel_reload(scope:"orchestrator"): the same stale rewrite remained.
Direct import of the patched module: preserves "minimax" and passes.

### Fix
applied locally: no
upstream-only: yes

Precise change needed: either make scope:"orchestrator" restart/rebind the comfyui-mcp server process that owns backend service modules, or narrow the tool description and add a safe MCP-server reload scope so local backend patches can actually take effect without manually killing the connector.

### Environment
macOS (Darwin), MPS 16 GB unified memory, ComfyUI 0.29.0, torch 2.12.1, Python 3.13
comfyui-mcp 0.49.3
comfyui-mcp-panel 0.11.32

artokun:
Fixed in #837 (merged to `main` as d9e8971). Verified present on `main` before closing — `panel_reload`'s scope is now reported honestly rather than claiming more than the call establishes.

Issue #760, opened 2026-08-03
Stale panel-op.lock is never reclaimed: reclaimStaleLock() is dead code, empty if-branch in acquireFileLock


artokun:
### What happened / root cause

A crashed/killed orchestrator leaves `~/.comfyui-mcp/panel-op.lock` behind, and it is **never** reclaimed — every subsequent `install_panel` (sync/update/install/pin) fails with the 60s timeout, permanently, until a human deletes the file by hand.

The stale-detection logic is all present and correct, but it is **not wired up**. In `dist/services/panel-pin-guard.js`, `acquireFileLock()` has an if-branch whose body is only a comment:

```js
if (lockIsStale(path)) {
    // Never rename/delete after a stale observation: another process can
    // replace this path with a fresh lock between the check and reclaim.
    // Node lacks an atomic compare-and-rename, so failing closed is the
    // only way to avoid stealing that new holder.
}
```

`lockIsStale()` correctly returns true (lock older than `STALE_LOCK_MS` **and** owned by a dead pid), and then nothing happens — control falls through to the deadline check and throws. Correspondingly, **`reclaimStaleLock()` (defined right above, line ~270) has zero callers in the entire `dist/` tree**:

```
$ grep -rn "reclaimStaleLock" dist/
dist/services/panel-pin-guard.js:270:function reclaimStaleLock(path) {
```

The comment's stated concern (no atomic compare-and-rename) is exactly what `reclaimStaleLock()` was written to solve — it renames to a unique `.reclaim-<uuid>` path first, so only one waiter can win and a straggler can never delete a fresh lock that replaced the path. So this looks like the reclaim call was removed during a fail-closed hardening pass and the helper was orphaned rather than the design being intentionally abandoned.

Impact: the failure is **permanent and silent-ish** — the error text says "Retry shortly or recover a proven abandoned lock explicitly", but there is no exposed recovery action (`install_panel` has no `recover`/`force-unlock`), so the only path is manual `rm` of an internal lockfile. In my case the lock had been sitting for ~21 hours from pid 2187, long dead.

### Steps to reproduce

1. Start a panel op and kill the orchestrator process mid-op (or plant the lock):
   ```bash
   echo '{"pid":999999,"startedAt":"2026-08-02T10:06:42.831Z"}' > ~/.comfyui-mcp/panel-op.lock
   touch -A -010000 ~/.comfyui-mcp/panel-op.lock   # older than STALE_LOCK_MS (10 min)
   ```
[...]

PR #836, opened 2026-08-04
locks and load-bearing durability: a stale lock nobody reclaims, and buffered records that can outlive their action

artokun:
## Issues in this cluster

- #760
- #798
- #765


artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 2 P0, 2 P1.** Bytes clean, no binaries. I verified each against the branch head before posting.

Reviving `reclaimStaleLock()` from dead code was the right call, and the "provably abandoned = old **AND** dead owner" rule is the right predicate. The findings are all in the half that reclaiming creates: **once a lock can be taken away, every other operation on it needs to prove it still owns what it is about to touch.**

### P0-1 — the release closure deletes whatever is at the path, without checking it is still ours

`panel-pin-guard.ts:~653` — the returned release is a bare `rmSync(path, { force: true })`. It never verifies the file there is still **its** lock.

Concrete sequence the gate walked: two `unlock` calls observe the same stale lock; one reclaims it; process B acquires a fresh lock; the other `unlock` renames B's lock aside. Before its `linkSync(claim, path)`, process C acquires the now-free path — `linkSync` fails `EEXIST`, the code returns "refused", **and B and C both proceed.** Then B's release closure fires and deletes **C's** lock.

The original bug leaves a lock stuck, which is annoying and recoverable. This one hands two writers the same lock and then leaves the path unprotected — concurrent panel mutations with nothing holding the door. That is a strictly worse failure than the one being fixed.

This is the same lesson that came out of panel #621 tonight, which I merged a few hours ago: **a lock only has an owner if every path through it proves ownership.** There, a holder registered before the write that made its claim true could orphan a token and permanently lock the canvas. Here, a holder releases a claim it no longer holds. Same invariant, opposite direction. The release must verify the lock file is the one it created (compare the record, or hold the fd/inode) and drop its in-memory claim without deleting a stranger's file.

### P0-2 — a record can outlive an action that never happened

`:976` — a failed pre-handoff `recordPanelPendingOp()`. If the first durable write renames successfully but the **directory fsync** fails, the catch attempts rollback; if that rollback write also fails, **its** catch silently keeps the new record and throws *"Refusing to start."*

So the Manager action is refused, and a phantom marker is left on disk that a later pin reads as a real queued update/restore — false pending and false cancellation reporting. **A catch that can throw is not a guard**, and this is the shape exactly: the rollback is inside the failure path, so its own failure has nowhere to go. Either make the rollback unable to fail (write the record only after the action is committed) or surface the leftover explicitly rather than swallowing it and reporting a clean refusal.

### P1-1 — PID recycling: `startedAt` is recorded but never used for identity

`pidAlive()` is only `process.kill(pid, 0)`. `startedAt` is on the record, and the code says outright it is *"for messages"* — it never establishes process **identity**. After a crashed holder's PID is recycled by an unrelated process, `unlock` reports the old owner as "STILL RUNNING" and refuses **indefinitely**, so #760's recovery path stays wedged until someone deletes the file by hand — which is the bug this PR exists to fix, surviving the fix.

"A process with that number exists" is not "our holder is alive". That is the repo's dominant fold pointed at a positive. The new tests cover `process.pid` and a dead PID, not reincarnation. Use `startedAt` against the process's own start time, or accept that liveness is **unknown** and let the age rule decide rather than asserting the holder is running.

### P1-2 — the fresh-lock test does not model the dangerous continuation

`panel-pin-guard.test.ts` ("restore never OVERWRITES a fresh lock…") asserts both lock files remain, but never models the **displaced holder continuing its mutation** or a subsequent acquire. So it passes while the two-holder state in P0-1 is reachable and while the displaced holder's unconditional release can remove the fresh lock. Assert the *reason* and the continuation, not the file count.

### On merging
Checks completed 00:56 and `main` has since taken the 0.49.6 release. The release touched only CHANGELOG/package files, so the risk is low here — but verify on the merged tree rather than assuming, since a stale green check took the panel's `main` red tonight.

PR #837, opened 2026-08-04
remote ComfyUI and secrets: HTML where JSON was promised, and a secret that reports success without reaching the child

### LINKED ISSUE #826: panel_request_secret reports success but the comfyui MCP child is never respawned — token stays invisible, downloads keep failing 401

`panel_request_secret` (comfyui / env / `CIVITAI_API_TOKEN`) returned the success message:

> 🔒 Token saved for the built-in comfyui tools (env "CIVITAI_API_TOKEN"). It's being applied now — the comfyui tools respawn with it as soon as this turn ends, then I'll retry. No reload needed.

The **persistence half worked**: the token was written to `~/.comfyui-mcp/.env`, 0600, correct key.

The **injection half did not**. The comfyui MCP child that actually runs `download_civitai_model` was never respawned, so it kept running with its pre-token environment. Retrying `download_civitai_model` in a *later* turn returned the identical `401 Unauthorized`.

Evidence:

```
~/.comfyui-mcp/.env mtime : 2026-08-04 17:47:47   (CIVITAI_API_TOKEN present, 0600)
comfyui-mcp child PID     : started Tue Aug  4 17:40:23 2026   <-- 7.5 min EARLIER
tr '\0' '\n' < /proc/<pid>/environ | grep -c '^CIVITAI_API_TOKEN='  ->  0
```

The orchestrator's spawn args for the child confirm it: env carried only `COMFYUI_URL`, `COMFYUI_MCP_PROGRESS_DIR`, `COMFYUI_MCP_TAB`, `COMFYUI_PATH` — no credential key.

There are two compounding causes:

1. `setEnvSecret` (dist/services/panel-secrets.js) writes `.env`, sets `process.env` **in the orchestrator**, and emits `change` — but the comfyui child respawn that `onComfyuiSecretsChanged` / `buildComfyuiMcpEnv` are supposed to drive did not occur.
2. Even if it had, `dist/config.js` snapshots credentials **at module load** (`civitaiApiToken: [REDACTED] with `dotenv.config()` reading `~/.comfyui-mcp/.env` exactly once at boot). So a long-lived child can never pick up a later token — it is structurally dependent on the respawn always firing.

### Impact

This is the *documented recovery path* for a CivitAI/HF 401, and it's a dead end: the tool claims success, the retry fails identically, and nothing distinguishes "no token configured" from "valid token on disk but not injected". An agent following the tool's own instructions loops.

### Steps to reproduce

1. With no `CIVITAI_API_TOKEN` set, call `download_civitai_model` → `401 Unauthorized`.
2. Call `panel_request_secret` (mcp_server `comfyui`, target_kind `env`, key `CIVITAI_API_TOKEN`); paste a valid token. Tool reports success.
3. End the turn. In a new turn, retry `download_civitai_model`.
4. **Expected:** download succeeds. **Actual:** identical `401 Unauthorized`; the child's `/proc/<pid>/environ` has no `CIVITAI_API_TOKEN`.

### Exact error (scrubbed)

```
Download failed: 401 Unauthorized — CivitAI requires an API token for downloads.
Set CIVITAI_API_TOKEN (panel Settings › "Set CivitAI token…", or the env var;
create one at civitai.com/user/account) and retry. Do NOT retry other model ids —
they will all fail the same way until a token is set.
```

(Token value never logged or transmitted; read only via `grep`/`cut` into a shell var.)

### Fix

`applied locally: yes`  `upstream-only: no (defence-in-depth half only; the respawn half still needs your side)`

Patched `dist/config.js` to resolve the two download credentials **lazily at access time** instead of snapshotting them at module load. A real env var still wins; otherwise the home `.env` is re-read. One pair of getters covers every consumer of `config.civitaiApiToken` / `config.huggingfaceToken`, so no call site changes. This makes the tools correct **even when the respawn doesn't fire** — it removes the structural dependency rather than papering over it.

```diff
--- a/dist/config.js
+++ b/dist/config.js
@@
 export const config = { ...parsedConfig, resolvedPort };
+// LAZY CREDENTIAL RESOLUTION. The download tokens above are snapshotted from
+// process.env at MODULE LOAD, and dotenv only read ~/.comfyui-mcp/.env once at
+// boot. A token saved AFTER this child started (panel_request_secret writes the
+// .env, then relies on the orchestrator respawning this process) is therefore
+// invisible forever if that respawn doesn't happen — every download keeps
+// failing 401 while the valid token sits on disk, with no signal distinguishing
+// "no token" from "token present but not injected".
+// Resolve these at ACCESS time instead: a real env var still wins, else re-read
+// the home .env. One getter covers every consumer of config.civitaiApiToken /
+// config.huggingfaceToken, so no call site changes.
+function freshEnvValue(...keys) {
+    for (const k of keys) {
+        const v = process.env[k];
+        if (typeof v === "string" && v.trim())
+            return v;
+    }
+    try {
+        if (existsSync(homeEnvPath)) {
+            const parsed = dotenv.parse(readFileSync(homeEnvPath, "utf-8"));
+            for (const k of keys) {
+                const v = parsed[k];
+                if (typeof v === "string" && v.trim())
+                    return v;
+            }
+        }
+    }
+    catch {
+        // Unreadable/corrupt .env — behave as if the token is unset.
+    }
+    return undefined;
+}
+Object.defineProperty(config, "civitaiApiToken", {
+    get: () => freshEnvValue("CIVITAI_API_TOKEN"),
+    enumerable: true,
+    configurable: true,
+});
+Object.defineProperty(config, "huggingfaceToken", {
+    get: () => freshEnvValue("HF_TOKEN", "HUGGINGFACE_TOKEN"),
+    enumerable: true,
+    configurable: true,
+});
```

Equivalent change belongs in `src/config.ts` upstream.

**Verified** (dist-only npx install, so no `npm run build`/`npm test` gate available — verified functionally in a fresh node process):

- `civitaiApiToken resolved: true`
- with `delete process.env.CIVITAI_API_TOKEN` (simulating a child that booted before the token existed — the exact failing case): `resolved WITHOUT any env var (pure .env re-read): true`
- precedence preserved: `real env var still wins: true`
- `node --check dist/config.js` passes

### Still needed on your side

1. **Make the respawn actually fire** — the `change` → comfyui-child-respawn path is the primary defect; the patch above only makes it non-fatal.
2. **Make the confirmation message conditional on the respawn completing.** Right now it asserts "the comfyui tools respawn with it as soon as this turn ends" unconditionally. If the respawn can't be confirmed, say so, so the agent knows to reload rather than retrying into the same 401.
3. Consider having the 401 error text distinguish *no token configured* from *token on disk but not visible to this process* — that alone would have made this self-diagnosing.

### Workaround for anyone hitting this now

The token is on disk and valid; only the running child can't see it. Either `panel_reload`, or fetch directly with the token read from `~/.comfyui-mcp/.env` without printing it.

### Environment

- OS: Linux 6.8.0-100-generic
- GPU: NVIDIA RTX 4000 SFF Ada Generation (20 GB VRAM), CUDA 12.8, torch 2.11.0, python 3.12
- ComfyUI 0.30.0 (local), ComfyUI-Manager 3.x legacy
- **comfyui-mcp 0.49.4**
- **panel 0.11.38**
- Backend: Claude

### LINKED ISSUE #828: Remote ComfyUI JSON endpoints return HTML, breaking template/runtime/system-stat tools

artokun:
Two issues with one theme: **a call reports success while the thing the caller needed did not happen.** A third, the same wound seen from the other side, was found while testing them.

Closes artokun/comfyui-mcp#826
Closes artokun/comfyui-mcp#828

---

## ⚠️ Read this first: the suite has been destroying the real credential store

`console-secrets.test.ts` drove the live `/api/secrets` endpoint without redirecting `COMFYUI_MCP_ENV_FILE`, so **every `npm test` wrote to the developer's actual `~/.comfyui-mcp/.env`**. It overwrote the live `CIVITAI_API_TOKEN` with the dummy `civ_key_123456789`, and its revoke case deleted whichever slot it cleared. The dummy is present in a July backup, so this predates this branch by weeks — plausibly the origin of the 401 that surfaced #826 in the first place.

**If you have run the test suite on this repo, re-enter your CivitAI token.**

Two defences, because a static one is not enough:
- **Runtime refusal** — any credential-store write from a test run without `COMFYUI_MCP_ENV_FILE` now throws, naming the fix. That cannot be walked around by how a test spells its import, and it turns a silent overwrite into a failing test at the moment of the write.
- **A sweep** (`secret-store-test-isolation.test.ts`) that fails if a test which can write credentials forgets to redirect the store — the assertion whose absence let this run unnoticed.

And the same wound from the other side: the store was rewritten **in place**, so a crash or a full disk between truncate and flush truncated *the whole file* — every token — while the per-key read-back still reported the save as successful. Writes are now temp-file + `fsync` + atomic rename, with a compare-and-swap that stops a concurrent writer being clobbered, and a whole-file verification so a loss is **reported** rather than narrated as a clean save.

## #826 (P1) — a secret that reports success without reaching the child

`panel_request_secret` answered *"the comfyui tools respawn with it as soon as this turn ends"* for every save, having observed neither half. When the respawn did not fire, the token sat on disk — 0600, correct key — while every download returned the same 401, and nothing distinguished "no token configured" from "valid token present but never injected". An agent following the tool's own advice loops.

**The structural half.** `src/env-file.ts` is now the single resolver for the canonical store, and it records which keys `process.env` was *seeded from* at boot — without that, "a real env var wins" silently pinned the stale boot value forever. Credentials resolve at **access** time: the download tokens (both HF aliases), `RUNPOD_API_KEY`, `REGISTRY_ACCESS_TOKEN`, and the HF readers in `missing_models`, training and `train_doctor`. Alias order outranks source, so a freshly saved `HF_TOKEN` beats a stale shell `HUGGINGFACE_TOKEN`. The child env is built **by construction** from the resolver — keys it does not supply are deleted, so a revoke actually revokes — and injected keys the file owns are marked so a later rotate reaches the child.

**The reporting half.** `setEnvSecret` verifies by read-back and returns `yes` / `no` / `unknown`; "could not determine" never hardens into either, at any consumer. The change event carries a `report()` callback the orchestrator answers with what the restart entry points *actually did*. `panel_request_secret` **refuses** when a read-back proves the save did not land, never calls an unverified save verified, and never describes a queued replacement as completed. A save the environment shadows says so and names the variable to unset. The Settings/API path is all-or-nothing across a slot's aliases and reports anything stranded or unverifiable.

## #828 — HTML where JSON was promised

Remote endpoints answered with an HTML document and the tools surfaced only `Unexpected token '<', "<!DOCTYPE "... is not valid JSON` — and `check_workflow_runtime` reported *"could not reach the ComfyUI server"* for a server that had plainly answered.

`src/comfyui/json-guard.ts` classifies from **body evidence**, never from the Content-Type header (a `200 text/html` carrying truncated JSON is named as a cut-off response, not as a proxy). It asserts a responder only when the body carries markers, and otherwise lists candidates and says the response does not settle it — including the credential remedy, which is only pointed at a gateway when a sign-in page is body-confirmed. Wired into `list_workflow_templates`, `get_system_stats`, `get_template_schema` and `getObjectInfo`. An HTTP error from the template index no longer reads as "no such template".

## Notes for review

- **No secret value** reaches a log, an error, a fixture or a commit. Bodies and library parse messages (including `Error.stack`, which embeds the message) are scrubbed three ways: known values in every encoding we can anticipate (percent, HTML-entity, **Base64** — which is how a short punctuation-bearing credential escapes a shape rule), anything introduced *by name*, and any long opaque run. It fails closed when a credential is too short to substitute safely.
- Values that cannot round-trip through dotenv, or that contain a control character (a NUL passes the file round-trip but `process.env` truncates at it), are **refused** rather than stored differently from what was supplied.
- **12 adversarial review rounds, 50 findings, all fixed. 107 mutants** run across the new tests — every one killed. Green on the **merge commit** with `origin/main` (262 files / 5155 tests), plus the vocabulary and tool-count gates.
- This machine is loaded enough that full-suite runs flake in a rotating cast of unrelated files; at `--maxWorkers=2` it is consistently green, and CI (dedicated runners) has been green throughout.

## Still owed: live verification

The end-to-end path has not been exercised on a running rig: a CivitAI 401 → `panel_request_secret` → retry in the same session without a reload, and a remote ComfyUI behind a proxy to see the #828 wording in situ.


artokun:
**Do not force-push anything onto this branch until the divergence below is resolved. There are two independent implementations of these fixes, and this is my mistake.**

## What happened

When the first agent's transcript was lost mid-round, I dispatched a replacement and told it to work on the existing `cluster/remote-and-secrets` branch. **It did not** — it created a fresh worktree from `main` and re-implemented the whole change its own way. I never verified which branch it was on, so both lines kept going.

Then the spend limit killed every agent, and I preserved the second line as `wip/agent-aa033b48b9d50b674` so nothing was lost.

## The two lines

Both branch from the same commit, `b38b58b`:

| | commits | `panel-secrets.ts` |
|---|---|---|
| `cluster/remote-and-secrets` (this PR, head `db737cc`) | 17 | +922 / −84 |
| `wip/agent-aa033b48b9d50b674` (head `158a796`) | 34 | +2052 / −109 |

Neither is an ancestor of the other. This is **not** a fast-forward and not a "continue where it left off".

## Why the wip line is worth reading before deciding

Its commit messages show it worked through the five P0s posted here and then kept going for another twelve rounds of its own self-gating — on exactly the failure classes this PR is about:

- `d130f2d` — *never claim a save the store cannot back — loss account at the swap, durable rename, disclosure after the rename*
- `50ac212` — *close the gate's 5 findings — case-free encodings, the absent-store check-then-act, guarded cleanup, and two consumers that dropped their obligations*
- `9e803f8` — *a ROLLBACK is a store write too — its damage travels the same path*
- `ee34225` — *a legacy purge that is not crash-durable can no longer read as a clean revoke*
- `f05a330` — *unreadable is not empty, a crashed writer's snapshot is swept, and a refusal stops naming the predecessor it never saw*
- `158a796` — *the verdict survives the subscription wrapper, and leftover files are described by kind*

That last set covers ground this branch does not appear to reach at all.

## What I am deliberately NOT doing

Picking one and force-pushing. This is the **credential store** — the file that holds live API tokens, and the one where a test was found writing into the user's real `.env`. Choosing between two 900–2000 line rewrites on that path by size or by commit-message quality is not a judgement to make quickly, and a wrong force-push destroys the other line's work irreversibly.

Both are pushed and safe: this PR's branch, and `wip/agent-aa033b48b9d50b674`.

## Recommended resolution

Read the wip line first. If it supersedes this one — which its scope suggests — reset this branch to it and re-run the independent gate against the combined result, rather than attempting to merge two implementations of the same functions. If it does not, it should be closed explicitly with a note, not left as a dangling ref someone later mistakes for lost work.

Either way the result needs a fresh gate: **neither line has been reviewed in its final state**, and the five P0s I posted were against `db737cc`, which the wip line never contained.

Issue #820, opened 2026-08-04
install_panel auto-install gate checks that a resolution is live-derived, not that it names the tree being installed into (#766)


artokun:
## The gate

`src/services/panel-installer.ts:1407`, the on-load auto-install gate:

```ts
if (isRealDeps(deps) && !isLiveDerivedBase(lastPanelBaseResolution())) {
  return { action: "unavailable", reason: "...not proven to be the tree ComfyUI serves from (#766)..." };
}
```

It asks whether the **current** resolution is live-derived. It never compares that resolution's base against the tree it is about to install into.

Its own comment claims more than the code delivers:

> *"Nobody is watching this path, so it only acts on a tree the live server itself chose."*

That is exactly the property not established. A live-derived resolution naming root **B** satisfies the gate while the install proceeds into root **A** — an observation of one tree licensing a mutation of another.

## Why it matters here specifically

This is a **mutation** gate, not a message. The failure mode is the one #766 exists to prevent, and the comment names it: an unattended install into a root the running server never named lands in a `custom_nodes` nothing loads, and the user sees **no panel and no error**. On a Comfy Desktop split install (`--base-directory`) the configured path is exactly that other root, so this is not exotic.

## Same defect, already fixed one line over

#793 fixed the identical shape in `servingQualifier`, which had certified whatever path was passed on the strength of a live-derived *current* resolution. Confirmation there now requires:

```ts
isLiveDerivedBase(resolution) && samePathCI(comfyuiPath, resolution.base)
```

`swapTreeIsCorroborated` already carries the comment about "is the CURRENT resolution live-derived?" answering a question about *now* rather than about *this tree*.

## The open question — analysis before any change

It may already be covered. `pinPanelBase` captures a generation before the await, and a retarget bumps it, so the pin-generation check adjacent to the mutation may close the window in practice. That analysis has not been done, and it is the whole issue:

1. Does the pinned-generation check actually cover a live-derived-but-different-tree resolution, or only a *retarget* between pin and mutation? Those are different: the resolution can name a different tree without any retarget having occurred in this call.
2. If it is covered, the comment should say *how*, because it currently claims a property the local code does not enforce.
3. If it is not covered, add the `samePathCI` comparison, matching `servingQualifier`.

## Why this is not folded into #793

[...]

PR #839, opened 2026-08-04
sweep the "could not determine" fold across the remaining surface, plus the auto-install tree gate

artokun:
## Issues in this cluster

- #796
- #820

### DIFF
--- a/src/__tests__/services/install-comfyui.test.ts
+++ b/src/__tests__/services/install-comfyui.test.ts
@@ -1,4 +1,8 @@
 import { describe, expect, it, vi } from "vitest";
+// Imported so the mocked fs/child_process functions can be scripted per-test;
+// these resolve to the vi.mock factories below.
+import { statSync } from "node:fs";
+import { spawnSync } from "node:child_process";
 
 // Mock node:child_process and node:fs so NO real git/pip/clone/disk access runs,
 // even if the default deps path is exercised.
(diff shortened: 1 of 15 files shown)


artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 2 P0, 1 P1.** Bytes clean, no binaries, no removed ratchet lines.

I pointed this gate at the two directions a *sweep* can fail — did it make anything looser, and did it reproduce the fold inside its own fix — and it found one of each.

### P0-1 — the tri-state reached the service but not its consumer

`queue-manager.ts:600-610` with `tools/queue-management.ts:138-145`. An unobservable cancel correctly returns `unverified: true, wedged: false` — the service does the right thing. But the tool sets `isError` **only** for `wedged`.

So: `/queue` is unreachable through verification, the service honestly reports UNKNOWN, and `queue(action: "cancel")` returns an **MCP success response**. A caller reads that as permission to continue and queue more work, while the cancellation state is explicitly unverified.

**This is the exact hazard a sweep carries**: converting a definite negative into "unknown" is only safe if every consumer of that value treats unknown conservatively. A tri-state introduced without updating its consumers turns a false refusal into a **false acceptance**, and a false acceptance here means work dispatched against a queue we could not confirm we cancelled. Audit every consumer of the new tri-states in this diff, not just this one — that is the pattern, and this is likely not the only site.

### P0-2 — `lora-catalog.ts:410-450`: `persist()` has no atomic exclusion between its second read and its write

Two MCP processes both `readCatalogRaw()` and see healthy bytes `V1`, both pass `unchanged`, then A writes entry A and B writes entry B. **B's write silently erases A's** — and because both probes were healthy, neither detects a change and **no backup is made**.

Two concurrent agents on one rig is ordinary here, so this is in scope and not theoretical. It is also the same family as the credential-store finding on #837 (an in-place read/modify/write with no CAS, whose read-back still confirmed success). The mid-persist test added here covers a change between the *two reads* — not the final **read-to-write** window, which is where the loss happens.

The shape that closes it is the one #837 landed: temp file + `fsync` + atomic rename, with a compare-and-swap re-read immediately before the rename, and a post-write verification against what was actually written rather than what was intended.

### P1 — `queue-manager.ts:356-376`: `lastPollFailed` is cleared by any later successful poll

A window of `running → poll failure → running` returns `"still-running"` rather than unobservable, and the wedge message can then claim *"across ~Ns of verified polling"* when verification had a **gap**. A temporary disconnect or a restart inside that gap makes the restart / do-not-queue guidance overconfident.

Small, but it is the fold in miniature and it is in a message this PR wrote: a sampled observation with a hole in it narrated as continuous verification. Either track that the window was interrupted, or narrow the claim to what was actually observed.

### Note on merging
Checks predate several merges to `main`, including the 0.49.6 release. **A green check is only green for the `main` it ran against** — that staleness took the panel's `main` red tonight. Verify on the actual merged tree.

Issue #824, opened 2026-08-04
install_panel update refuses safe git verification on Manager total=0 done=1


artokun:
### What happened / root cause
install_panel(action='update') refused to use its git fallback after ComfyUI-Manager returned an incoherent completed-queue signature: total=0, done=1, pending=0, in_progress=0. The panel repository was a clean git checkout, so a manual git pull --ff-only was required to verify/update it. The repo was already current.

### Steps to reproduce
1. Run install_panel(action='update') on a local clean comfyui-agent-panel git checkout.
2. Manager reports total=0 and done=1 after the queue drains.
3. The tool errors instead of verifying the clean checkout against origin.

### Exact error (scrubbed)
"Panel update did NOT apply: nothing changed on disk, and the Manager's queue counts are NOT the proven legacy-3.x empty-queue signature ... got total=0, done=1, pending=0, in_progress=0 ..."

### Fix
Applied locally: no. Upstream-only: yes.
Precise suggested change: when Manager produces this partial/incoherent signature, safely inspect the panel checkout. If it is a clean non-symlink git repo with no pin/shadows, fetch origin and compare HEAD to origin/main; use git pull --ff-only when behind, or report verified already-current when equal. Continue refusing if the checkout is dirty or unverifiable.

Workaround verified:
git pull --ff-only in ~/ComfyUI/custom_nodes/comfyui-agent-panel returned "Already up to date", HEAD 4fa48a66, pyproject version 0.11.38. A subsequent install_panel(status) re-read 0.11.38 with decision up-to-date.

### Environment
Windows 11; ComfyUI 0.30.0; NVIDIA RTX 3070 Ti 8 GB; comfyui-mcp 0.49.4; panel 0.11.38.

Related (same root cause): #724

artokun:
Fixed and shipped.

**`8f434b1` (PR #840)** — the signature gate no longer hard-refuses on an incoherent drained-queue signature. When the LIVE queue indicators (`pending_count` / `in_progress_count` / `is_processing`, all present and clear) show no Manager work in flight, the no-op branch runs the pinned git verification directly: fetch, compare, `merge --ff-only` when behind, full post-verification. Your exact case — `total=0, done=1, pending=0, in_progress=0` over a clean checkout already at the tip — now reports "already at the upstream tip" with `restartRequired: false`, which is the answer your manual `git pull --ff-only` gave. A dirty or unverifiable checkout, or live Manager work not disproven, still refuses.

[...]

Issue #816, opened 2026-08-04
report_issue can hang indefinitely when Worker submission stalls


artokun:
### What happened / root cause
The report_issue MCP tool did not return within more than 60 seconds while filing an issue for our own panel repository. The caller had to terminate the tool and use the intake Worker directly.

### Steps to reproduce
1. describe_tool(report_issue).
2. call report_issue with repo artokun/comfyui-mcp-panel and a complete body/version fields.
3. Wait over 60 seconds; no output is returned.
4. Terminate the call.

### Exact error (scrubbed)
No response/output after >60 seconds; tool remained running until terminated.

### Fix
Applied locally: no. Upstream-only: yes.
Workaround verified: direct bounded POST to the Worker endpoint.
Suggested change: enforce a bounded timeout in report_issue and return the documented prefilled fallback URL or a clear Worker timeout result.

### Environment
Windows 11; ComfyUI 0.30.0; RTX 3070 Ti 8 GB; comfyui-mcp 0.49.4; panel 0.11.38.

artokun:
Fixed in **`8f434b1`** (PR #840) — by the tool description, which is the correct fix here, so saying that plainly rather than burying it.

The ">60s of silence" is the designed, bounded triage wait: `report_issue` polls the async worker (#799) past the worker's own wall clock. The actual failure mode was an agent reading a slow-but-bounded call as a hang, aborting, and retrying — which can **double-file**, because the worker files or dedups autonomously once it has accepted a report. So the fix is telling the calling agent that; no timing constant needed to change, and shortening one would have made the tool worse.

The boundedness claim in the new text was verified rather than assumed: `maxPolls` clamps to 200, `pollDelayMs` to 60000, and every request carries an `AbortSignal` timeout — so "time-capped … never indefinite" is true. The text also names every terminal shape (filed / deduped / upgrade-advice / rarely unfiled) instead of promising an accepted report always ends filed, and notes that a `pending` return may carry no `job_id` when the acknowledgement was unreadable.

If you see a `report_issue` call that genuinely never returns — as opposed to one that takes a few minutes — that is a different bug from this one; please file it with the elapsed time.

Issue #801, opened 2026-08-04
panel_set_widget reports a stale/unrelated "previous" value for a subgraph's promoted widget


artokun:
### What happened / root cause
Called panel_set_widget on a subgraph node's promoted `prompt` widget. Immediately before the call, panel_query_graph({ids:[<subgraph_id>], fields:'detail'}) showed the widget's current value as prompt text A (a pencil-sketch grandmother description). panel_set_widget's response for the same node_id + widget='prompt' reported `previous` = an entirely unrelated prompt text B (a vaporwave/VHS title-sequence description that was never shown as this node's value in this session), and the `set` block's echoed `node_id` was the promoted INNER node id (e.g. "104") rather than the subgraph node id the caller passed (e.g. "105"), alongside `promoted_from: { subgraph_node_id, inner_node_id, parent_widget_synced: true }`.

A follow-up panel_query_graph on the subgraph node confirmed the NEW value was applied correctly, so the actual write appears correct — but the reported `previous` value is misleading and does not match what was actually displayed/stored for that node just before the call. This suggests either (a) a stale/incorrectly-scoped read when resolving "previous" for a promoted subgraph widget (e.g. reading from a shared subgraph template/definition default, or from a different node/workflow entirely, instead of this specific instance's current promoted value), or (b) a logging/response-construction bug that surfaces the wrong prior value while still writing to the correct target.

### Steps to reproduce
1. Open a workflow containing a subgraph node with a promoted STRING widget (e.g. an "Image to Video" wrapper subgraph with a promoted `prompt` widget backed by an inner node).
2. panel_query_graph({ids:[<subgraph_node_id>], fields:'detail'}) — note the current `prompt` value shown.
3. panel_set_widget({node_id: <subgraph_node_id>, widget: 'prompt', value: '<new text>'}).
4. Compare the `previous` field in the panel_set_widget response against the value observed in step 2 — they did not match in this session (previous was unrelated content, and the response's `set.node_id` was the inner node id rather than the id passed in).

### Exact error (scrubbed)
No error/exception — this is a wrong/misleading successful response, not a crash. Example (paraphrased, values shortened):
```
// step 2 panel_query_graph result for node 105:
"prompt": "Pencil-sketch portrait of an elderly grandmother..."

// step 3 panel_set_widget({node_id:105, widget:'prompt', value:'<new pencil-sketch text>'}) result:
{
  "set": { "node_id": "104", "widget": "prompt",
[...]

PR #840, opened 2026-08-04
single defects with no shared codepath: subgraph promoted-widget previous value, report_issue hang, Manager total=0 verification

artokun:
## Scope — read this first

This PR **fixes one issue with code (#824)** and **one with documentation (#816)**. It does **not** implement #801. Stated up front because the diff touches three files and a reader could reasonably assume otherwise.

| Issue | What this PR does | Code changed? |
|---|---|---|
| #824 | Fixes it | yes — `panel-installer.ts` |
| #816 | Fixes it, by tool description | description text only |
| #801 | **Nothing.** Verified fixed upstream in the panel repo; issue left OPEN with the finding posted on it | no |

---

## #824 — `install_panel(update)` git-verifies a provably-idle incoherent Manager queue

The signature gate hard-refused on `total=0, done=1` even with a clean git checkout, so an already-current panel came back as a hard failure. Now, when the LIVE queue indicators (`pending_count` / `in_progress_count` / `is_processing`, all present and clear — `isProvenQuiescentQueue`) show no Manager work in flight, the no-op branch runs the existing pinned git verification: fetch, compare, `merge --ff-only` when behind, full post-verification. `HEAD === upstream` reports "already at the upstream tip" with `restartRequired: false`. Live work not disproven, or a dirty/unverifiable checkout, still refuses. `updateViaGitCheckoutFallback` takes the caller's `managerReason`, so no message asserts the stale-3.x cause for a signature that does not prove it, and a merely partial status is named "missing or malformed", never "incoherent".

### The concurrency bracket (gate P0)

The first version of this authorized a destructive `git merge --ff-only` from `pending=0, in_progress=0, is_processing=false` as read out of the update call's result. That was wrong, and the reason is worth stating precisely: those three indicators establish **"the queue reported nothing in flight at the moment we asked"**, not **"nothing is writing to this tree"** — and only the second licenses the mutation. git's index lock excludes other *git* processes; it does not hold back a ComfyUI-Manager pip install, unpack or file copy landing in the same checkout.

Every entry to `updateViaGitCheckoutFallback` now brackets its own merge:

- **Before** — re-probe the LIVE queue (new `fetchLiveQueueCounts` dep, defaulting to `fetchManagerQueueCounts`), so the idleness that licenses the merge is milliseconds old rather than however long the Manager call took. An unreadable probe is *cannot determine*, which **refuses**; it is never read as idle.
- **Before** — compare-and-swap the tree: HEAD and `git status --porcelain` must still be exactly what the gates above inspected. A writer that never touches the queue is caught here and nowhere else.
- **After** — re-read all three. The merge has run by then, so these **disclose** rather than refuse: the tree may be inconsistent and the caller is told so, instead of being handed a success report the evidence does not support.

The bracket is armed on **all three** entries (#724 legacy, #824 quiescence, #771 Manager-error). All three warrants describe the Manager's *verdict*; none of them says whether a Manager task is writing to the checkout right now.

**What this is not: mutual exclusion.** ComfyUI-Manager is a separate third-party process with no lease or lock protocol this orchestrator can join, so nothing here can *prevent* a concurrent write — it can only shrink the unobserved window to what an out-of-process check allows, and turn a write landing near it into a disclosure rather than a silent success. Two residues are named in the code rather than asserted away: a Manager task that both starts and finishes inside the merge is undetectable from here by construction; and a write confined to git-ignored paths is invisible to porcelain, though the pinned ignored-collision gate already bounds the damage there (the merge's file set is fixed at `targetRev` and was proven not to overlap an ignored local file). No message claims the window was exclusive.

### Refuse vs disclose

Pre-merge refusals say **no git mutation was made**, and add that whether the Manager call itself changed anything is exactly what its own status could not report — they do not claim "nothing was changed", which this path is not in a position to know. The CAS refusal goes further: something demonstrably *did* change the checkout, so it tells the caller to re-read rather than assume the pre-update state. Post-merge failures say the merge **ALREADY RAN**.

## #816 — `report_issue` sets the blocking-wait expectation (description-only, and that is the fix)

**A documentation change is the correct fix here, so saying it plainly rather than burying it.** The reported ">60s of silence" is the designed, bounded triage wait: `report_issue` polls the async worker (#799) past the worker's own wall clock. The failure mode was an agent reading a slow-but-bounded call as a hang, aborting, and retrying — which can **double-file**, because the worker files or dedups autonomously once it has accepted a report. Telling the agent that is the fix; no timing constant needed to change, and changing one would have made the tool worse.

The boundedness claim in the new text was verified, not assumed: `maxPolls` clamps to 200, `pollDelayMs` to 60000, and every request carries an `AbortSignal` timeout — so "time-capped … never indefinite" is true. The text also now names every terminal shape (filed / deduped / upgrade-advice / rarely unfiled) rather than promising an accepted report always ends filed, and notes that a pending return may have no `job_id` when the acknowledgement was unreadable.

## #801 — NOT fixed here; left open

The stale `previous` for a promoted subgraph widget is constructed in the **panel repo**, which the orchestrator only proxies — there is no orchestrator-side change that would fix it. Panel `b5d5acd` (panel #591, released in panel **0.11.39**) reports the OUTER rail's prior value as `previous` and keeps the inner one as `inner_previous`; the issue was filed against panel 0.11.38, which predates it. The `set.node_id` echo the issue also flags is labelled by the `promoted_from` block — the issue's own "or clearly documenting/labeling which id is which" alternative.

That is a finding, not a fix shipped by this PR, so **#801 stays open** and the finding is posted on the issue for the maintainer to close or re-scope.

## Verification

- Full suite green at full concurrency: 255 files, 5143 tests, ~37s (no `--maxWorkers` workaround; #853 is on `main`).
- All four bracket guards mutation-verified: disabling any one fails exactly its own cases and nothing else. Six new cases use a queue signature **identical** to the verified-current one, differing only in what is true at the moment of the mutation — live op at the merge, unreadable probe, HEAD moved under the gates, worktree dirtied under the gates, a write landing during the merge, a Manager task appearing during the merge. Each asserts the reason and whether the merge ran.
- Three pre-existing fallback tests corrected: they moved HEAD to `REV_B` while leaving the upstream at `REV_A`, which a pinned `merge --ff-only <sha>` cannot do.
- Committed blobs scanned for control bytes: zero across all three files.
- Rebased on current `main` (0.49.7, #838/#853/#855/#856).
- Adversarial self-gate (codex `gpt-5.6-terra`): round 2 returned **0 P0, 0 P1**. Round 1's "acquire an exclusive mutation lease every Manager writer honors" was rejected with reasons — not implementable against a third-party process — and answered by removing the code's own "exclusion" claim instead.

**Only a live rig can confirm** the end-to-end behaviour against a real stale Manager 3.x host: the bracket's re-probe path is exercised only through stubs here.


artokun:
Independent gate (codex `gpt-5.6-terra`, read-only, full branch diff). **NO-SHIP — 1 P0, 1 P1** — plus a scope problem I verified separately that matters more than either.

### The scope problem: this PR claims three issues and implements one

The diff touches exactly three files: `panel-installer.ts`, `panel-installer.test.ts`, and `report-issue.ts`.

- **#801** (a stale "previous" value for a promoted subgraph widget) has **no implementation at all** — no file touched, and the string `#801` does not appear anywhere in the diff. I checked specifically after the gate flagged it.
- **#816**'s entire change is the `report_issue` tool **description**.

That is *"a partial fix reported as complete"*, which is the outcome this project treats as worse than an honest partial. Please either implement #801 or drop it from the PR body and leave the issue open with what you found — both are fine; claiming it silently is not.

**On #816 I want to be fair, because a documentation fix can be the correct fix here.** The new text tells the agent the call blocks for minutes, that this is normal rather than a hang, and not to abort-and-retry because the worker files autonomously and a blind retry can double-file. If the reported "hang" was an agent aborting a slow-but-bounded call, that *is* the fix. **And I verified the boundedness claim rather than taking it**: `maxPolls` is clamped to 200, `pollDelayMs` to 60000, and each request carries an `AbortSignal` timeout — so *"time-capped … never indefinite"* is true, not an overclaim. Say plainly in the PR body that #816 is fixed by documentation and why; a reader who sees only a description diff will otherwise assume it was missed.

### P0 — a destructive `git merge --ff-only` authorized from one stale snapshot

`panel-installer.ts:5196`. The new fallback treats `pending=0, in_progress=0, is_processing=false` as **provably idle**, then fetches and merges. But a Manager operation can already be active while its indicators are **one poll stale**, and **git's index lock does not synchronize arbitrary Manager writes to the checkout**. So a concurrent Manager write and this merge can overwrite each other or leave the panel tree inconsistent.

The distinction to hold onto: those three indicators establish *"the queue reported nothing in flight at the moment we asked"*, not *"nothing is writing to this tree."* Those are different claims, and only the second licenses a destructive merge. Loosening #824's refusal was right — reporting an already-current panel as a hard failure is a real bug — but the replacement needs a positive exclusion against concurrent writes, not the absence of a busy flag one poll ago.

### P1 — the refusal test does not model the unsafe case

`panel-installer.test.ts:993` covers only a **contemporaneously reported** `pending_count: 1`. It never models an active Manager operation with **stale clear indicators**, which is precisely the P0. So the suite cannot distinguish the new unsafe authorization from a genuinely quiescent queue — the standing check applied: if the loosening were wrong, would a test fail? Today, no.

### Note on merging
The branch head is a merge of `origin/main`, so verify on the merged tree — and note `main` has since taken the 0.49.6 release and #853. A green check is only green for the `main` it ran against.

artokun:
Correction to an earlier note: **no work was lost here, and nothing needs recovering.**

When the spend limit killed the agents I preserved each worktree's commits to a `wip/` ref and flagged this one as diverged. On inspection it is not. The agent working this PR had moved on to #839 sequentially, and the snapshot caught it on a detached head mid-second-task — so its ref contains a **rebased copy** of #839's `8dd8259` (already pushed to that branch) plus a `main` commit, and nothing else.

I verified by diffing the two patches: they differ only in blob hashes and `@@` line offsets, which is a rebase artifact, not a content change.

This branch's own 7 commits are intact and are strictly more than the ref carries. **No force-push, no merge, no action.** Recording it so the earlier flag does not send someone looking for work that was never missing.

The gate findings posted above still stand and are what this PR needs next — the P0 in particular (a destructive `git merge --ff-only` authorised from one queue snapshot, where the indicators establish "the queue reported nothing in flight when we asked", not "nothing is writing to this tree"), plus the scope question about #801 having no implementation.

Issue #821, opened 2026-08-04
Flaky test: ui-bridge 'retries binding when the port is briefly held' picks a random port and hangs when it is taken (~1 in 8)


artokun:
## The flake

`src/__tests__/services/ui-bridge.test.ts` → `UiBridge (multi-tab)` → **"retries binding when the port is briefly held, then self-heals"** times out at ~5026ms.

Reproduced on run 8 of twelve instrumented full-suite runs — roughly **1 in 8 locally**, and load- and luck-dependent, so CI will hit it intermittently.

## Cause

The test picks a **random port** — `40000 + rand(20000)` — and binds a blocker on it to simulate the contended-port case it exercises.

When that port is *already held on the machine*, `blocker.on("listening")` never fires, and the test sits until the 5s deadline. The fixture's own setup is the thing that races.

It is not the deliberate Windows bind race already documented in that file's header (*"the bind, not the logic, was the flake"*) — that one concerns the bridge's own bind, and this is the blocker the test installs before the bridge starts.

## Why it is worth its own issue

- **Pre-existing.** Confirmed by `git diff origin/main...HEAD` on `fix/panel-ask-durable-response-486` containing none of that test's lines. It is unrelated to #486.
- **It will be misattributed.** The same file gained ~360 lines of new grace-clock tests in #811, and those *did* have genuine real-timer races (since made deterministic). A flake in this file will now naturally be blamed on that work. Recording the actual cause here prevents the next person chasing the wrong thing — which already happened once during #811, where an unreproducible 4-failure run turned out to be the clock races rather than this.
- **It fails in the annoying direction**, not the dangerous one: it fails on a *correct* implementation. No false pass, so nothing is silently unverified.

## Fix shape

Bind the blocker on an **OS-assigned** port (`listen(0)`) and read back the assigned port, rather than guessing a random one and hoping it is free. That removes the race rather than widening the deadline — a larger timeout would only make it rarer while leaving the same failure reachable under load.

Whatever the fix, it should assert the blocker is actually listening before the bridge starts, so a failure to set up the fixture reports as a setup failure rather than as a timeout in the behaviour under test.

## Provenance

Diagnosed during #811 review. The agent deliberately did **not** touch it — unrelated to the PR, and that PR had enough review surface already. Correct call; filing it here instead.

Related: #811.

artokun:
[...]

PR #843, opened 2026-08-04
test reliability: a random-port bind that hangs ~1 in 8

artokun:
## Issues in this cluster

- #821

### DIFF
--- a/src/__tests__/services/relay-client.test.ts
+++ b/src/__tests__/services/relay-client.test.ts
@@ -9,9 +9,23 @@ import { RelayClient } from "../../services/relay-client.js";
  * {t,id,d} envelope on the orchestrator leg documented in that repo's README.
  * Exercises RelayClient + UiBridge.attachRelayConnection end to end without
  * needing a live Cloudflare deploy.
+ *
+ * Binds an OS-assigned port (listen(0)) and reads it back — never a guessed
+ * random port, which can already be held on the machine (#821). A bind failure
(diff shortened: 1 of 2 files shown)


artokun:
Verified and merging.

Its CI checks completed at **19:26**, and `main` has moved three times since (#818, #763, #831), so I did not trust them — that staleness took panel `main` red earlier tonight. Instead I built the actual merged tree (`origin/cluster/test-reliability` + `origin/main`, no conflicts) and ran the two touched suites there **nine times consecutively: 159/159 every run.**

Being precise about what that does and does not prove: at the reported ~1-in-8 failure rate, nine clean runs alone would be about 30% likely by chance — not conclusive on its own. **The structural argument is the one that carries it.** `listen(0)` plus reading the assigned port back removes the collision *mechanism*, rather than making it rarer; there is no longer a guessed port that another process could already hold. The repeated runs corroborate that; they are not the evidence.

Two things done right here that are worth naming, because they are the difference between fixing a flake and hiding one:

- **The timeout was not widened.** Widening would have left the same race with a longer fuse, and the next slow machine reopens it.
- **A bind failure now rejects at setup**, so a broken fixture reports as a setup failure instead of surfacing as a timeout inside the behaviour under test. That is the same principle the rest of this codebase applies to runtime errors — do not let an infrastructure failure masquerade as a verdict about the thing you were measuring.

The `ui-bridge` suite gets the same treatment; the run logs show OS-assigned ports throughout.

Issue #848, opened 2026-08-04
panel_restart_comfyui ignores updated Desktop launch arguments


artokun:
### What happened / root cause
After adding `--disable-dynamic-vram` to the active ComfyUI Desktop installation's `installations.json` (with a verified backup), `panel_restart_comfyui` restarted the server successfully but relaunched it with the previous captured argv. The new flag was absent, and startup still reported `DynamicVRAM support detected and enabled`.

### Steps to reproduce
1. Run a Desktop-managed ComfyUI through comfyui-mcp.
2. Update that installation's launchArgs in Comfy Desktop metadata.
3. Call `panel_restart_comfyui`.
4. Inspect `/system_stats` argv or startup logs.

### Exact behavior
Restart returned healthy, but `/system_stats` argv contained the old arguments without `--disable-dynamic-vram`; logs reported DynamicVRAM enabled.

### Fix
Applied locally: no. Upstream-only: yes.
The restart path should reload the current Desktop installation metadata or expose an explicit launch-argument override; reusing stale captured process arguments makes runtime configuration changes appear to succeed while silently not taking effect. Workaround: stop the server and launch the saved Python/core command explicitly with the new flag.

### Environment
Windows 11; ComfyUI 0.30.0; NVIDIA RTX 5090 32 GiB; Python 3.12.9; comfyui-mcp 0.49.5; comfyui-mcp-panel 0.11.38.

Related (same root cause): #368

artokun:
Regression on **comfyui-mcp 0.50.71** / panel 0.11.63 (Windows 11, ComfyUI Desktop 0.31.1, Manager 3.41): `restart_comfyui(action:"restart")` reported the launch arguments **UNCHANGED** while `listener_ownership` was `"unconfirmed"`. The listener on 8188 was actually running a minimal `main.py --port 8188 --enable-manager --enable-cors-header` command, missing the Desktop `--extra-model-paths-config`, `--input-directory`, and `--output-directory` arguments.

Observed impact: model inventory fell from 14 to 8, input resolution fell back to the stock directory and produced false missing-media/model findings, and repeated restarts left 15 `main.py` processes (about 10 GB RSS). This reproduces after the #850 fix shipped in v0.50.0.

artokun:
I claimed the recurrence and spent a pass reading the restart path. **I am not shipping a fix, because the two mechanisms I expected both fail on inspection** — posting what I ruled out so the next attempt starts from evidence rather than repeating this.

[...]

Issue #846, opened 2026-08-04
Session ENVIRONMENT line reports stale comfyui-mcp version


artokun:
### What happened / root cause
The injected session ENVIRONMENT line reports `comfyui-mcp 0.48.18`, while the installed package is 0.49.5 and the MCP process for the resumed session started after that 0.49.5 package was present. Bug triage consequently received the wrong reporter version until the discrepancy was checked manually.

### Steps to reproduce
1. Install/update comfyui-mcp to 0.49.5.
2. Start or resume a panel agent session.
3. Read the injected ENVIRONMENT line: it can still report 0.48.18.
4. Compare with the package loaded by the freshly spawned MCP child.

### Exact error
No thrown error; the environment metadata is stale and misleading.

### Fix
Applied locally: no. Upstream-only: yes. Resolve `mcp_version` from the package actually loaded by the spawned MCP/orchestrator process when composing the session environment, rather than cached install/update metadata. Add a regression covering package update followed by agent respawn/resume.

### Environment
macOS; ComfyUI 0.29.0; Apple MPS/16 GB; actual comfyui-mcp 0.49.5; reported comfyui-mcp 0.48.18; panel 0.11.38.

Related (same root cause): #358

PR #850, opened 2026-08-05
restart + session reporting: stale launch arguments, a premature failure verdict, and a stale version line

artokun:
## Issues in this cluster

- #848 — `panel_restart_comfyui` ignores **updated** Desktop launch arguments
- #367 — `restart_comfyui` reports a **failed startup just before a healthy one**
- #846 — the session ENVIRONMENT line reports a **stale `comfyui-mcp` version**

Grouped because they are one defect seen three ways: **something observed earlier is reported as if it were still true now.** Stale argv, a verdict rendered before the thing it judges finished, and a version string captured at a moment that has passed.

#367 is the sharpest: reporting a failed startup *immediately before* the server comes up healthy is a "could not determine" folded into a definite negative — the deadline expired, which establishes that startup was not confirmed **yet**, not that it failed. The user is told their restart broke, seconds before it works.

### DIFF
--- a/src/__tests__/orchestrator/panel-restart-health-readiness.test.ts
+++ b/src/__tests__/orchestrator/panel-restart-health-readiness.test.ts
@@ -12,9 +12,16 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 const resetClient = vi.fn();
 const resetObjectInfoCache = vi.fn();
+// #848: the post-restart argv read goes through getSystemStats. Default REJECTS, so
+// every pre-existing test keeps the "nothing observed → say nothing" behaviour and
+// only the tests that opt in exercise the drift note.
+const getSystemStats = vi.fn(async () => {
+  throw new Error("ECONNRESET");
+});
 vi.mock("../../comfyui/client.js", () => ({
   getObjectInfo: vi.fn(),
   backfillObjectInfo: vi.fn(),
+  getSystemStats: () => getSystemStats(),
   resetClient: () => resetClient(),
   resetObjectInfoCache: () => resetObjectInfoCache(),
 }));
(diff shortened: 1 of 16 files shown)


artokun:
**Recovered 14 commits that were lost when the agents were killed mid-work by the spend limit.** The branch now carries them, rebased onto current `main`.

They were only ever in a local worktree — never pushed. I preserved them as `wip/agent-ace77ef187ab9feb9` when the agent died, and have now verified the ref is a clean **descendant** of this branch (a fast-forward, so nothing here was overwritten), merged `main` in cleanly with no conflicts, and advanced the branch.

What was recovered, beyond the original `da6c5a1` (*a deadline expiring is not a failure, a restart is not a reconfigure, and a version read at boot is not the version now*): **twelve codex gate rounds**, each closing real findings. A few worth surfacing because they are this cluster's own defect class turning up inside its fix:

- **round 1** — *four findings — my own fix was making the same mistake in three new places*
- **round 5** — the `StartupConfirmation` doc still said *"ComfyUI is down"* after the messages had stopped saying it: the same bucket-narrated-as-a-cause defect **hiding in a comment**
- **round 7** — *an inferred reboot is not an acknowledged one, and the JSON must not say what the prose withheld*
- **round 11** — `restart_comfyui` could **stop instance A and relaunch against B**
- **round 12** — *the round-11 fence covered only half the window, and `stop_comfyui` had none*

Rounds 11 and 12 are the ones I would look at first on review: a restart that stops one instance and brings up a different one is the wrong-target harm class, and the fence for it needed two passes.

## Verification

- **256 files / 5170 tests** pass at full concurrency; `tsc` clean; zero control bytes across every changed file
- Clean merge with `main` (which has since taken 0.49.7 and 0.49.8)

## Still needed

This has **not** had an independent gate in its final state — the twelve rounds are the author's own. It should get one before merge, per the rule that every author-PASS gated this week produced real defects.

Recording the cause plainly since it affected several PRs: when the agents were killed I preserved each worktree's commits to a `wip/` ref. Most were clean continuations like this one; some **diverged** because a replacement agent re-implemented from `main` instead of continuing the branch. Those are flagged individually and must not be force-pushed either way — see #837 for the worst case.

Issue #648, opened 2026-08-01
list_extra_paths ignores saved default workspace when COMFYUI_PATH is unset


artokun:
### What happened / root cause
`get_workspace` correctly returned the saved default ComfyUI workspace, but `list_extra_paths` failed because `extra-paths.js` only checked `config.comfyuiPath` and ignored `workspace-env.getWorkspace().workspace_path`.

### Steps to reproduce
1. Run comfyui-mcp with COMFYUI_PATH unset.
2. Persist a default workspace with set_default_workspace (or existing workspace config).
3. Call get_workspace: it reports workspace_source "default-config" and a valid path.
4. Call list_extra_paths: it throws.

### Exact error (scrubbed)
VALIDATION_ERROR: No local ComfyUI path is known. Set COMFYUI_PATH, set a default workspace, or pass config_path explicitly.

### Fix
applied locally: yes
upstream-only: no

Patch:
```diff
--- a/dist/services/extra-paths.js
+++ b/dist/services/extra-paths.js
@@
 import { resolveServerExtraModelConfig } from "./output-dir.js";
+import { getWorkspace } from "./workspace-env.js";
@@
+async function resolveTargetPathWithWorkspace(opts = {}) {
+  try { return resolveTargetPath(opts); }
+  catch (error) {
+    const target = opts.target ?? "auto";
+    if (!opts.configPath && (target === "auto" || target === "standalone")) {
+      const workspace = await getWorkspace();
+      if (workspace.workspace_path) {
+        return { target: "standalone", path: join(workspace.workspace_path, "extra_model_paths.yaml") };
+      }
+    }
+    throw error;
+  }
+}
@@
- return { ...resolveTargetPath(opts), notes: [] };
+ return { ...(await resolveTargetPathWithWorkspace(opts)), notes: [] };
```

Verified in a fresh Node process: listExtraPaths() resolved ~/Stability Matrix/Data/Packages/ComfyUI/extra_model_paths.yaml and returned the Stability Matrix model groups.

### Environment
Windows 11; ComfyUI 0.29.0; RTX 5060 Ti 16 GB; comfyui-mcp 0.48.32; panel 0.8.0.

Issue #490, opened 2026-07-30
apply_manifest downloads models into stale local COMFYUI_PATH instead of connected ComfyUI


artokun:
While applying the bundled qwen-image-edit manifest against the connected ComfyUI at /home/parn/repositories/wet/wet.graphics.pipeline/ComfyUI, Manager began downloading Qwen_Image_Edit-Q8_0.gguf into /home/parn/ComfyUI/models/diffusion_models/. These are separate directories; the latter is not a symlink and is not the connected instance. apply_manifest also timed out while the Manager task kept running. Expected: manifest model destinations resolve against the connected ComfyUI root. Actual: a stale/local COMFYUI_PATH was used, leaving a multi-GB partial file in the wrong instance. Environment: Linux, ComfyUI 0.29.0 remote, comfyui-mcp 0.48.13.

artokun:
Reopening: this remains reproducible on current comfyui-mcp 0.49.6 (panel 0.11.39). In a real remote `--comfyui-url` session with a stale local `COMFYUI_PATH`, `resolveEffectiveComfyUIBase()` returns that local path before checking remote mode, despite its documented contract. The shared resolver is used by multiple callers, including destructive snapshot and `comfy_cli` node operations, so an operation aimed at the remote server can act on the unrelated local install.

artokun:
Adding a **third instance**, and it is the one that actually executes the destructive commands.

`src/services/comfy-cli.ts:12-13` states:

> Never returns a local path in remote mode (`resolveEffectiveComfyUIBase` already enforces that).

That guarantee fails **twice over**:

```ts
function defaultWorkspace(): string | null {
  return config.comfyuiPath ?? resolveEffectiveComfyUIBase() ?? null;
}
```

1. `resolveEffectiveComfyUIBase` does **not** enforce it — that is this issue.
2. Even if it did, it is **never consulted** in the common case: `config.comfyuiPath ??` short-circuits first, so the delegation the comment relies on does not happen at all whenever `COMFYUI_PATH` is set.

A comment that names another function as the enforcer, for a check that function does not perform, in a path that does not call it — that is how this issue survived being closed once already. Nobody re-derives a guarantee that is stated this confidently.

[...]

PR #854, opened 2026-08-05
panel_show_media: the 20 MB refusal is a dead end for local video (orchestrator half of #648)

artokun:
## Issue

- #648 — `panel_show_media` cannot preview local reference videos over 20 MB

## Why this exists separately

Panel PR **#649** fixed the panel half: video now routes through the existing storyboard pipeline with full disclosure (the video was NOT sent, N *sampled* frames, source size or explicit UNKNOWN, `get_image` as the way to actually see it). It also fixed a worse bug found on the way — **under** the ceiling the panel painted a player for the *user* and returned `{ok:true,count:1}` to an agent that had been shown nothing.

But **#648's literal repro still fails**, because the refusal is raised here, before the panel is ever called:

`src/orchestrator/panel-tools.ts` — `panel_show_media`, absolute-path branch:

```ts
if (stat.size > MAX_BYTES) {
  return fail(
    "file too large (" + (stat.size / 1024 / 1024).toFixed(1) + " MB > 20 MB): " + p,
  );
}
```

## What to do

The refusal itself is **correct** — a 72 MB inline payload is not something to send an agent, and raising the cap is not the fix. The bug is that it is a **dead end**: it names a ceiling with no way past it, so an agent asked to inspect an ordinary local reference video concludes the task is impossible.

Two cases, and they need different answers:

1. **The file is under a ComfyUI root.** Forward it as a `viewRef` instead of refusing — there is no size cap on that path, and it is what a browser panel resolves same-origin anyway. A `viewRef` video arriving at the panel **already** gets the full preview and disclosure today (#649), so this one change completes the feature end to end.
2. **The file is not under a root.** Keep the refusal, but make it name a remedy that works from where the caller is — the size limit, why it exists, and what they can do (move/copy it under a root, or ask the user). **A remedy must be actionable from the caller's current state**; a bare ceiling is not one.

## Constraints

- **Do not raise `MAX_BYTES`.** The cap protects the inline payload path and is doing its job.
- Establish "under a ComfyUI root" from evidence, not assumption. Note **#490 is reopened**: `resolveEffectiveComfyUIBase()` returns `config.comfyuiPath` before checking remote mode, and `comfy-cli.ts` bypasses it entirely. Another agent owns that fix — **do not fix it here**, but do not build on the broken guarantee either. If the root cannot be established, that is *unknown*, not *not-under-a-root*.
- A `stat` that throws is not a "file too large" and not a "file not found" — check what the surrounding branch does with an unreadable file.
- This is the same split-halves pattern as #614 / #809. The panel half is merged-pending; this half closes the issue.

### DIFF
--- a/src/__tests__/orchestrator/panel-show-media-oversized.test.ts
+++ b/src/__tests__/orchestrator/panel-show-media-oversized.test.ts
@@ -24,6 +24,8 @@ const state = vi.hoisted(() => ({
   cloud: false,
   /** Paths statSync should fail on with a non-ENOENT error. */
   statEacces: new Set<string>(),
+  /** absolute path -> errno code readFileSync should fail with. */
+  readFail: new Map<string, string>(),
 }));
 
 vi.mock("../../services/output-dir.js", async (importOriginal) => {
@@ -66,6 +68,17 @@ vi.mock("node:fs", async (importOriginal) => {
       }
       return (actual.statSync as (...a: unknown[]) => unknown)(p, ...rest);
     }) as typeof actual.statSync,
+    // A file whose METADATA reads fine but whose CONTENTS do not — statSync
+    // proves the former, never the latter.
+    readFileSync: ((p: Parameters<typeof actual.readFileSync>[0], ...rest: unknown[]) => {
+      const code = typeof p === "string" ? state.readFail.get(p) : undefined;
+      if (code) {
+        const err = new Error(`${code}: injected read failure, open '${String(p)}'`) as NodeJS.ErrnoException;
+        err.code = code;
+        throw err;
+      }
+      return (actual.readFileSync as (...a: unknown[]) => unknown)(p, ...rest);
+    }) as typeof actual.readFileSync,
   };
 });
 
@@ -144,6 +157,7 @@ beforeEach(() => {
   state.remote = false;
   state.cloud = false;
   state.statEacces.clear();
+  state.readFail.clear();
 });
 
 describe("panel_show_media: an oversized file UNDER a ComfyUI root is forwarded", () => {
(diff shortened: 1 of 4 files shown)


artokun:
**Recovered the real work onto this branch.** It had been carrying only my empty placeholder commit; the agent's actual commits existed solely in a local worktree and were lost when the spend limit killed it.

Safe to replace because the placeholder changed **no files** — I verified that before force-pushing, and used `--force-with-lease` pinned to the placeholder's real SHA so a concurrent push could not have been clobbered. Nothing was overwritten but an empty commit.

What landed:

- **`bf6d67a` — an oversized local video is not a dead end: forward what ComfyUI already serves.** This is the orchestrator half of comfyui-mcp-panel#648. The panel half shipped in 0.11.41, but the 20 MB refusal is raised *here*, before the panel is ever called, so the reporter's literal repro still failed. A `viewRef` video arriving at the panel already gets the full sampled-preview disclosure, so this is what completes it end to end.
- **`7c7ae70` — a failed canonicalisation is a doubt, not a verdict** (codex gate). Exactly the fold I asked the dispatch to watch for: if the ComfyUI root cannot be established, that is *unknown*, and it must not silently become "not under a root" — which would blame the user's file location for something we simply could not determine.

## Verification

- **257 files / 5183 tests** pass at full concurrency after merging current `main`; `tsc` clean; zero control bytes; no merge conflicts.

## Still needed

No independent gate in this final state — the one round it had was the author's own. It needs one before merge.

Also worth confirming on review: the dispatch called out that **#490 is reopened** (`resolveEffectiveComfyUIBase` returning a local path in remote mode) and that this change must not build on that guarantee. `7c7ae70`'s framing suggests it did handle the unknown case correctly, but that is the specific thing to check rather than assume.

When this merges, comfyui-mcp-panel#648 can close — it is currently reopened precisely because this half was outstanding.

PR #891, opened 2026-08-05
generations.db inside the ComfyUI install breaks ComfyUI's git updater (#872)

### LINKED ISSUE #872: generations.db inside the ComfyUI install breaks ComfyUI's git updater on Windows (locked WAL handle)
## Summary

`GenerationTracker` writes its SQLite DB **inside the ComfyUI git working tree** at `<comfyuiPath>/comfyui-mcp/generations.db`. That directory is not ignored by ComfyUI's `.gitignore`, so ComfyUI's own updater sweeps the DB into a commit and then tries to delete it — which Windows refuses while the orchestrator holds the SQLite handle open. **The ComfyUI update aborts, and its rollback fails too.**

Hit twice in a row on the same machine (two failed update attempts ~5 min apart).

## Exact failure

```
[PRE_UPDATE_HEAD] 0764232429b8cfb10b79b6f186c8cb23e0b22897
Creating backup branch: backup_branch_2026-08-05_10_45_03
[BACKUP_BRANCH] backup_branch_2026-08-05_10_45_03
Uncommitted changes saved to backup branch.
Fetching from origin...
Resetting to origin/master
[ERROR] Update checkout failed: "could not find 'C:/Users/.../ComfyUI/comfyui-mcp/generations.db' to remove: The process cannot access the file because it is being used by another process."
[ERROR] Failed to restore pre-update state: '0764232429b8cfb10b79b6f186c8cb23e0b22897'
```

## Mechanism (confirmed on the affected machine)

1. `src/services/generation-tracker.ts:158-166` — `defaultDbPath()` returns `join(config.comfyuiPath, "comfyui-mcp", "generations.db")` whenever `comfyuiPath` is set, i.e. for every normal local install.
2. `journal_mode = WAL` (line 149) means three live files: `generations.db`, `-shm`, `-wal`. The orchestrator (`node dist/index.js`) keeps the handle open for its entire lifetime — the process on this machine had held it since Jul 30.
3. ComfyUI's updater saves uncommitted work with a `git add -A` + commit onto a `backup_branch_<ts>`. Because `comfyui-mcp/` is untracked-but-not-ignored, **all three DB files get staged and committed.**
4. The updater then runs `git reset --hard origin/master`. The DB is now in the index but absent from `origin/master`, so git must delete it from the working tree.
5. Windows denies the unlink (open exclusive handle) → checkout fails → the restore-to-`PRE_UPDATE_HEAD` step fails for the same reason. The repo is left on a detached HEAD with three files staged as `A`.

Verified: the file is genuinely locked (`File.Open(..., FileShare.None)` throws `used by another process`), and the holder is the orchestrator process, not ComfyUI.

## Impact

Any Windows user running comfyui-mcp against a git-based ComfyUI install cannot update ComfyUI while the orchestrator is running — which is most of the time, since it's a long-lived background process. The error names our file but gives no hint that stopping the MCP server is the workaround, so it reads as a broken ComfyUI install. It also silently accretes committed copies of the DB (+ ~115 KB WAL) on every backup branch: 14 backup branches on this machine each carry a copy.

Linux/macOS are unaffected by the *hard failure* (unlink of an open file succeeds), but they still get the DB committed to backup branches, and the running orchestrator keeps writing to a now-unlinked inode — so the generation history silently vanishes on update there.

## Proposed fix

Make the data dir **self-ignoring** at creation. In the `GenerationTracker` constructor (`generation-tracker.ts:144-146`), after `mkdirSync(dir, { recursive: true })`, write a `.gitignore` containing `*` into that directory if one isn't already present:

```ts
mkdirSync(dir, { recursive: true });
// The dir may live inside ComfyUI's git tree; a self-ignoring .gitignore keeps
// ComfyUI's updater from committing (and then failing to delete) the open WAL DB.
const ignore = join(dir, ".gitignore");
if (!existsSync(ignore)) writeFileSync(ignore, "*\n");
```

`*` in a directory's own `.gitignore` ignores everything in it including itself, so `git add -A` skips the DB entirely and `git reset --hard` never tries to delete it. It's the same trick pip/venv use, needs no path migration, and fixes existing installs on next start.

Worth considering separately: writing runtime state into someone else's git checkout is fragile in general — the `COMFYUI_MCP_DATA_DIR` / `~/.comfyui-mcp/instances/<slug>/` path the remote branch already uses (lines 162-165) would avoid the class of problem, but that needs a migration for existing histories, so the `.gitignore` is the low-risk fix.

Caveat: this prevents *future* staging but does not un-track a DB already committed by a previous failed update. A user in that state needs `git rm --cached comfyui-mcp/generations.db*` once (index-only, safe with the file open) — possibly worth doing defensively, or documenting.

## Workaround for anyone hitting this now

From the ComfyUI repo root, with the orchestrator still running:

```bash
git rm --cached comfyui-mcp/generations.db comfyui-mcp/generations.db-shm comfyui-mcp/generations.db-wal
printf '/comfyui-mcp/\n' >> .git/info/exclude
```

`.git/info/exclude` is used rather than `.gitignore` because `.gitignore` is tracked and gets clobbered by the very `reset --hard` we're trying to survive. Update then succeeds with no need to stop the MCP server.

## Environment

- Windows 11 Pro 10.0.26200
- ComfyUI at `v0.30.1` (`0764232`), 12 commits behind `origin/master`
- Orchestrator: `node dist/index.js`, better-sqlite3 in WAL mode

artokun:
Fixes #872.

`GenerationTracker` wrote a live WAL SQLite database — three open files — at `<comfyuiPath>/comfyui-mcp/generations.db`, **inside ComfyUI's own git working tree**, in a directory its `.gitignore` does not cover.

ComfyUI's updater saves uncommitted work with `git add -A`, so it committed our database onto a backup branch; `git reset --hard` then tried to remove it, and Windows refuses to delete a file whose handle we hold for the orchestrator's entire lifetime. **The user's ComfyUI update aborted, and its rollback failed too**, leaving them on the backup branch. Reported twice within five minutes on one machine.

The locked handle is only what made it visible. The mistake was writing our own state into a tree we do not own.

## The fix

The per-user data dir (`~/.comfyui-mcp/instances/<slug>/`) was **already** the path for remote/cloud installs, and its instance slug (host+port) keeps separate installs apart — the only thing the in-tree location was really achieving. That path is now used always.

## Migration

Existing installs have real history at the old location, so it is moved on first construction:

- **All three WAL files travel together.** A `.db` carried across without its `-wal` loses every transaction that had not been checkpointed — silent data loss dressed as a successful migration.
- **An existing database at the new location is never clobbered.** The legacy file is left where it is and its path reported, with the note that its presence can break ComfyUI's updater.
- **A failed move says so**, and says that part of the set may have moved — a `.db` here and its `-wal` there is a corrupt pair.

## On the tests

The existing test asserted the in-tree placement as correct — *"local install: DB stays inside the install's comfyui-mcp dir (unchanged)"*. It now asserts the opposite, plus that nothing of ours lands anywhere under the install.

**Three mutations, each caught by exactly the intended test.** The `-wal` case needed the file recreated by hand first: a clean SQLite close checkpoints it away, so without that the test could not distinguish a migration that carries it from one that drops it.

Full suite: 6006 passing. Control bytes 0, no git-binary files.

### DIFF
--- a/src/__tests__/services/generation-tracker-path.test.ts
+++ b/src/__tests__/services/generation-tracker-path.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { mkdtempSync, rmSync, existsSync } from "node:fs";
+import { mkdirSync, mkdtempSync, rmSync, existsSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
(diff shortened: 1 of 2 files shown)


artokun:
**Confirmed live on the rig**, with a nuance that strengthens the case for moving the file rather than ignoring it.

The database really is inside ComfyUI's git working tree:

```
C:\Users\Artokun\ComfyUI-Installs\ComfyUI\ComfyUI\comfyui-mcp\generations.db   (45 KB)
$ git -C ...\ComfyUI rev-parse --is-inside-work-tree
true
```

But it does **not** currently break the updater on this machine, and the reason matters:

```
$ git -C ...\ComfyUI check-ignore -v comfyui-mcp/generations.db
.git/info/exclude:9:/comfyui-mcp/    comfyui-mcp/generations.db
```

Someone has added `/comfyui-mcp/` to **`.git/info/exclude`**. That file lives inside `.git/`, is **never committed, never cloned, and never shared** — so it is a per-clone workaround that:

- does not exist on a fresh ComfyUI install,
- does not survive a re-clone or a reinstall,
- and was not present on the reporting machine, which is why #872 fired there and not here.

So the current protection is one local file away from disappearing, and its absence is silent — the failure only shows up later, as an aborted ComfyUI update whose rollback also fails.

That is the argument for this PR over the cheaper alternative of shipping a `.gitignore` entry: we cannot make a durable promise about a file we place in someone else's version-controlled tree. Moving it to the per-user data dir removes the question.

The live pass also confirmed the surrounding behaviour is healthy on current `main` — orchestrator up clean, one bridge socket with zero closes across a full ComfyUI restart, a render end-to-end, and `resolveOutputDir` correctly following the live server's `--output-directory` rather than the configured base.

PR #895, opened 2026-08-05
direct tools/call answers a retired name with a bare 404, not the ledger redirect — blocks the 0.50.0 flip

artokun:
## Why now

Consolidation rule 6 is that every RETIRED tool name answers with a self-explaining
redirect — `Unknown tool 'X' — removed in 0.49.0. Call Y (action:"z") instead.` — rather
than a bare 404. That is what makes retiring even a high-traffic name acceptable: the
caller is told, in the error, what to call next.

`retiredToolMessage()` was wired into exactly **two** call paths: the `call_tool` facade
(`src/tools/compact.ts`) and the Ollama backend's router
(`src/orchestrator/ollama-backend.ts`). **Nothing served it on a direct MCP `tools/call`.**
There, `@modelcontextprotocol/sdk`'s own handler answers `Tool <name> not found` and stops.

Measured on `d9e8971` (full mode) against a name the 0.49.0 slices had already retired:

| path | answer |
| --- | --- |
| direct `tools/call` | `MCP error -32602: Tool <name> not found` — **bare 404, rule 6 broken** |
| `call_tool` (compact) | the ledger redirect — correct |
| `call_tool` on full (#616 hatch) | the ledger redirect — correct |

This is invisible today only because compact mode is the DEFAULT, so every call goes
through the facade. **The 0.50.0 flip (#726) makes direct `tools/call` the default path in
the same release that retires 121 names**, so rule 6 would break for the whole surface at
once. Landing this first means every slice's retirements are covered on both paths from
day one. It touches no tool family, so it does not collide with the slice PRs.

## What it does

`src/tools/retired-redirect.ts` captures the `tools/call` handler the SDK installed and
puts a wrapper in front of it. A name that is (a) absent from the SDK's registration table
and (b) present in `DEAD_NAMES` gets the ledger message; **everything else delegates to
the original handler with the request unchanged.**

Installed at the two choke points that serve a direct `tools/call`:

- `createConfiguredServer()` (`src/index.ts`) — after BOTH branches. Full mode is what the
  flip makes default; compact mode matters too, because a client calling a retired name
  directly there is exactly the stale-snapshot client the facade exists for (#616).
- `getCallToolClient()` (`src/orchestrator/index.ts`) — the panel's `call_tool` bridge
  dispatches through `client.callTool()`, i.e. the same direct path. Same gap, never
  covered by the facade, and **not new in 0.50.0**: a retired name has always come back to
  the panel as `error: "Tool <name> not found"`.

### A/B on the real 154-tool surface

With and without the wrapper, **exactly one line differs**:

```
              WITHOUT                                    WITH
RETIRED   Tool get_workspace not found     ->  Unknown tool '…' — removed in 0.49.0. Call workspace (action:"get") instead.
UNKNOWN   Tool totally_made_up_xyz not found            (identical)
REAL-OK   results: [1] 6*7 => 42                        (identical)
BAD-ARGS  Input validation error: …                     (identical)
NOARGS    Input validation error: …                     (identical)
PROTO     Tool __proto__ disabled                       (identical)
CTOR      Tool constructor disabled                     (identical)
EMPTY     Tool not found                                (identical)
```

A genuinely made-up name keeps the SDK's own error. A model told
`totally_made_up_xyz was removed in 0.49.0` would go hunting for a replacement that never
existed, so the ledger is consulted rather than every miss being answered with a redirect.

## The crux: this reads two PRIVATE SDK fields

`server.server._requestHandlers` and `server._registeredTools`. There is no public API for
this — `setRequestHandler` before registration trips the SDK's own
`assertCanSetRequestHandler`; `fallbackRequestHandler` is only consulted when no handler
exists; registering dead names as tools puts them back in `tools/list`; wrapping
`transport.onmessage` means fabricating JSON-RPC outside the Protocol's bookkeeping. All
four are written down in the module header with why they were rejected.

The failure we cannot afford is that a future SDK renames one and the wrapper **quietly**
stops intercepting — CI green, every retired name back to a bare 404. Three things stop it:

1. **Pre-conditions**, each asserted at install time and named in the failure message.
2. **A FUNCTIONAL post-condition.** Identity is not assertable (`setRequestHandler` stores
   its own closure, so the map never holds the function we wrote), and `!== original`
   would pass for any third function an SDK might write into the slot. So after
   installing, the code **drives** the entry in the dispatch map with a synthesized
   `tools/call` for a retired, unregistered name and requires the ledger's exact message
   back. Any rename, re-route or reshape that leaves the wrapper unreachable fails there,
   whatever it did to the plumbing. The probe is side-effect-free by construction: the two
   possible outcomes are our redirect (pass) or the SDK's "not found" (fail); no tool
   handler runs either way.
3. **A canary test** running the real install against a real `McpServer`. Simulating a
   field rename fails 18 of that file's 21 tests, the canary among them, with the moved
   field named.

**Throw vs log.** Production calls a log-and-degrade variant and reports the result on the
startup line. Refusing to BOOT in order to protect the quality of one error message is the
worse outage — degrading restores exactly today's shipped behaviour. That is only
defensible if the SDK a user can resolve is BOUNDED, and it was not: this is normally run
as `npx comfyui-mcp`, which resolves fresh (package-lock protects `npm ci` and nothing
else), so a caret range made "internals nothing ever ran against" the COMMON case. So
`package.json` now caps `@modelcontextprotocol/sdk` at `~1.29.0` (was `^1.12.1`).
`npm install` resolves 1.29.0 unchanged; the lockfile delta is only the range string;
`@anthropic-ai/claude-agent-sdk` declares `^1.29.0` and still dedupes onto the same copy.

## Also: the served surface is `MAX_TOOLS + 3`

The three facade names (`call_tool`, `list_tools`, `describe_tool`) live in no ledger and
no baseline, so full-mode `tools/list` returns 157 against a `MAX_TOOLS` of 154 at the time of measurement. Nothing
asserted that before — every surface test stopped at `registerAllTools` and never saw the
facade. `registry-surface.test.ts` now pins it three ways (the count, WHICH three the
surplus is, and that no ledger tool is dropped or reordered), expressed against
`MAX_TOOLS` so it survives every slice landing.

## Adversarial gate — 3 rounds, both findings fixed

| # | Finding | Disposition |
| --- | --- | --- |
| P1 | Fail-open leaves the guarantee unenforced for any SDK resolved through the caret range | **FIXED** — `~1.29.0` cap; the boolean is reported at both call sites instead of discarded |
| P2 | The post-condition accepted any function that was not the original, so a third function would report "active" while the wrapper never ran | **FIXED** — replaced with the functional self-test above (the reviewer's proposed identity check is not implementable; the SDK stores its own closure) |

Round 3 verdict: **SHIP**, no P0–P3 findings.

## Verification

- Full suite green post-rebase onto slice 8 (286 files / 6059 tests). `origin/main` baseline has one unrelated
  flaky `download-cache` failure this branch does not.
- `npm run lint`, `check:vocabulary`, `vocab:export --check` all pass.
- Every test in the new file mutation-verified: 21 mutations, each producing exactly the
  expected specific failure. Notably, deleting the whole wrapper fails 5 tests — and the
  real-dispatch / validation / disabled / unknown-name tests all still PASS, which is why
  the mechanism tests exist.
- Ratchet artifacts untouched: no change to `TOOL_NAMES`, `MAX_TOOLS`, `BASELINE_SHA256`,
  `docs/design/tool-surface.txt`, or `docs/design/tool-vocabulary.json`. The only
  `vocabulary.ts` change is one per-occurrence `allowedIn` entry for a verbatim quotation
  of measured output.

## Known adjacent gap, deliberately NOT fixed

The panel's own MCP HTTP surface (`registerPanelTools`, `src/orchestrator/panel-mcp-http.ts`)
has the same direct-`tools/call` gap, and the ledger holds two retired `panel_*` names.
It is left alone on purpose: `DEAD_NAMES` is one ledger spanning both surfaces, so
installing there would answer a retired CORE name by naming a core replacement that the
panel surface does not serve — trading a bare 404 for a redirect into a second dead end.
Fixing it properly needs a per-surface ledger view, which is its own change.

Unblocks the #726 flip.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "license": "MIT",
       "dependencies": {
         "@comfyorg/sdk": "^0.1.5",
-        "@modelcontextprotocol/sdk": "^1.12.1",
+        "@modelcontextprotocol/sdk": "~1.29.0",
         "@stable-canvas/comfyui-client": "^1.5.9",
         "better-sqlite3": "^12.6.2",
         "dotenv": "^16.4.7",
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   },
   "dependencies": {
     "@comfyorg/sdk": "^0.1.5",
-    "@modelcontextprotocol/sdk": "^1.12.1",
+    "@modelcontextprotocol/sdk": "~1.29.0",
     "@stable-canvas/comfyui-client": "^1.5.9",
     "better-sqlite3": "^12.6.2",
     "dotenv": "^16.4.7",
--- a/src/__tests__/tools/registry-surface.test.ts
+++ b/src/__tests__/tools/registry-surface.test.ts
@@ -28,15 +28,21 @@ import { MAX_TOOLS, TOOL_NAMES } from "../../tools/vocabulary.js";
  */
 describe("registered tool surface", () => {
   let names: string[];
+  let served: string[];
   const saved: Record<string, string | undefined> = {};
 
   beforeAll(async () => {
-    for (const key of ["COMFYUI_WORKFLOWS_DIR", "COMFYUI_MCP_TOOL_MODE"]) {
+    for (const key of ["COMFYUI_WORKFLOWS_DIR", "COMFYUI_MCP_TOOL_MODE", "COMFYUI_MCP_NO_FACADE"]) {
       saved[key] = process.env[key];
     }
     process.env.COMFYUI_WORKFLOWS_DIR = await mkdtemp(join(tmpdir(), "comfyui-mcp-surface-"));
     delete process.env.COMFYUI_MCP_TOOL_MODE;
+    // The served-surface assertions below are about the DEFAULT full-mode surface,
+    // which layers the facade. `=1` in a dev shell would drop the three meta-tools and
+    // fail those with an off-by-three nobody would guess the cause of.
+    delete process.env.COMFYUI_MCP_NO_FACADE;
     names = await listRegisteredToolNames();
+    served = await listRegisteredToolNames("served");
   });
 
   afterAll(() => {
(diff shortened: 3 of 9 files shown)


artokun:
Rebased onto `8499b0e` (0.50.0 slice 8: runpod 11→2), which moved `MAX_TOOLS` 154 → 145 and `DEAD_NAMES` 41 → 51 entries. Both new test files pass unchanged — that is the design working rather than luck:

- the served-surface assertion is expressed against `MAX_TOOLS`, so `157` became `148` on its own;
- `retired-redirect.test.ts` takes its subject FROM the ledger instead of spelling a name, so slice 8's ten new retirements needed no edit here, and are covered from the moment they land.

End-to-end against the real full surface after the rebase: install reports active, `tools/list` = `MAX_TOOLS + 3`, **all 51** ledger names redirect correctly, and `totally_made_up_xyz` still gets the SDK's own `-32602 ... not found`.

Full suite post-rebase: 286 files / 6059 passed.

PR #913, opened 2026-08-06
bench: tool-reach benchmark for the 0.50.0 consolidation (154 flat vs compact vs 33 action-parameterized)

artokun:
## What this is

A **tool-reach benchmark**: does folding 154 flat MCP tools into 33
action-parameterized ones make models reach for the RIGHT tool more often, or
less often?

A score on the new surface alone cannot answer that, so the same 100
user-phrased requests run against three surfaces:

| arm | surface | how |
|---|---|---|
| `baseline` | 154 flat tools | pinned at **`d9e8971`** — the last commit before slice 8. The old surface is already gone from `main`; pinning the SHA is the only way the "before" number stays reproducible |
| `compact` | 3 meta-tools | today's default registration |
| `consolidated` | the final 33 | `main`, once the slices land — `--ref` makes that one command, not an edit |

```
npm run reach -- --arm baseline
npm run reach -- --arm compact --models claude/sonnet,ollama/qwen3:4b
npm run reach -- --arm consolidated --ref origin/main
npm run reach:report        # re-render from saved runs
npm run reach:selfcheck     # request set + fold map only
```

**Do not merge yet.** This is the measurement instrument plus the `baseline` and
`compact` numbers; the `consolidated` arm needs a re-run once slices 9–16 land.

## Nothing the model picks is executed

Not an optimisation — it is what makes the benchmark possible at all. The
surface under test can delete multi-GB checkpoints, uninstall node packs, rent
paid cloud GPUs, stop and restart the server, drop the GPU cache and file real
GitHub issues, against a **shared** ComfyUI with unrelated unsaved tabs open.
100 requests × 3 surfaces × several models is thousands of calls.

`scripts/tool-reach-dispatch.mjs` is the single choke point and it is an
**allowlist**: the only calls that ever reach real code are compact mode's
`list_tools` / `describe_tool`, because navigating the catalog is the cost being
measured on that arm. The escape hatch is *injected*, which is what makes the
guarantee testable rather than asserted — `tool-reach-dispatch.test.ts` fires
every destructive name on the surface through both the flat and the `call_tool`
path and fails if the hatch is ever touched.

The surface itself is real: `listRegisteredTools()` / `collectToolCatalog()` are
the same entry points `src/index.ts` uses, so the schemas are byte-identical to
what a client is told. That is the thing being measured, so it cannot be
stubbed.

## Scoring compares (tool, action), never the name alone

Name-only matching is a false-acceptance machine on the consolidated arm, and an
**asymmetric** one: `regenerate` folds into `generate_image action:"regenerate"`
and so does a plain render as `generate_image action:"image"`, so a wrong action
scores as a hit — while the 154-name baseline has no action to get wrong and
cannot benefit from the same leniency. The consolidation would look better than
it is, purely as a scoring artefact. (The same bug was live in
`arena-scenarios.mjs`'s `c.tool === "regenerate"` check.)

tool-only is reported as a separate, clearly labelled column, never the
headline. Reported separately: first-reach (tool+action), first-reach
(tool-only), reach-within-3, navigation cost, abstention on distractors, and a
wrong-tool **confusion matrix** — the actionable output, because it names which
tool is stealing which request.

## The request set is stratified, not sampled

57 neutral (the unbiased headline) · 30 targeting the seven predicted reach
failures H1–H7 · 9 ambiguity pairs where two tools legitimately compete · 4
negatives whose correct answer is no tool at all. Without the negatives, a model
that reaches on every turn is indistinguishable from one that reaches correctly.

No request names its own tool. `checkRequestSet()` rejects the snake/space/hyphen
forms, the folded `"<tool> <action>"` phrasing, and bare action tokens; the
irreducible overlaps (a user does say "VRAM") carry a written rationale, and the
tests prove the gate can fail.

## What the harness refuses to do

- **Merge incomparable episodes.** Every episode carries a fingerprint of the
  request as asked and the conditions the model could perceive. Raising the
  result cap or rewording a request invalidates the affected rows, which then
  resurface as missing coverage rather than being averaged in. A row with no
  fingerprint is dropped — "cannot tell" has to mean re-run.
- **Publish a partial arm as complete.** The report labels any row that did not
  cover all 100 requests.
- **Silently skip a model.** Every skip is reported next to the results with its
  reason.
- **Count a provider failure as a model failure.** Errored episodes are excluded
  from the accuracy denominators and reported separately.

## Additive

Nothing in `llm-arena.mjs` or `arena-scenarios.mjs` is touched — those are moving
under several slice PRs. The only edits outside `scripts/tool-reach-*` are three
`npm` scripts, two `.gitignore` lines, and five exact paths appended to
`check-tool-vocabulary.mts`'s `SELF` set for the files that hold retired names as
DATA (the fold map *is* the migration table; per-name `allowedIn` would need an
entry added by whichever slice retires that name, so every slice would collide
on this benchmark's exceptions).


artokun:
Status, so this isn't mistaken for pending work — it needs a rebase, not a review.

**Partly superseded.** #1003 has landed: the tool-reach corpus (`benchmarks/tool-reach.jsonl`), a one-command blind-arm runner, and `tool-reach-corpus.test.ts`, all green against current main.

**What this PR still holds that #1003 does not:** the three-arm comparison — `baseline` pinned at `d9e8971` (the last commit before slice 8), `compact`, and `consolidated` — plus the dispatch/foldmap/merge/report harness. That pinned baseline is the only reproducible "before" number, and once it's gone it cannot be reconstructed. That's why I'm not closing this.

**Why it's red.** It is **108 commits behind main**, and its checks last ran on 2026-08-06 against a main that no longer exists. It touches two files main has moved substantially since the merge base:

| file | commits on main since |
|---|---|
| `package.json` | 11 |
| `scripts/check-tool-vocabulary.mts` | 10 |

The second is the real work: that script was **rewritten** in #970 (it now enumerates via `git ls-files --others --exclude-standard` and reports tracked/untracked counts, because its old scan silently skipped untracked files — a gate with a blind spot). `git merge-tree` confirms conflicts there. So the red is staleness plus a genuine conflict, not a broken benchmark.

**Also worth noting:** the surface settled at **37** tools, not the 33 this PR's arms and title assume. The consolidated arm needs re-pointing before its number means anything.

**Parked behind the stabilization pass**, deliberately. Benchmark tooling isn't a stability fix, and I'd rather leave this honestly labelled than rebase 4,000 lines of it while user-facing defects are still landing. When it's picked up, the order is: rebase onto main, re-point the consolidated arm at 37, re-run all three arms, then read the number.

For anyone judging consolidation in the meantime: #1003's corpus measures the current surface, and it is the one that's green today.

artokun:
Closing — the question this benchmark existed to answer has been answered, and a leaner harness shipped in its place.

**What happened to the two halves of this PR:**

- `scripts/check-tool-vocabulary.mts` **landed on main** separately and is a live gate today.
- The tool-reach harness (`scripts/tool-reach*.mjs`, the stub server, dispatch/foldmap/merge/score) **did not**. `#1003` shipped a different, much smaller implementation of the same idea — `benchmarks/run-arm.mjs` + `benchmarks/tool-reach.jsonl` + a corpus test — targeted at the consolidated 37-tool surface that actually shipped, rather than the 154-flat vs compact vs 33-action comparison this branch was built to measure.

So this is being closed as **superseded in substance, not merged in form**. The consolidation it was meant to inform is done: the surface went 154 → 37 and the default mode flipped, both released in 0.50.x.

**The work is not lost.** Branch `bench/tool-reach` keeps all 5 commits and ~4,000 lines, including the stub server and scoring rig, which are more thorough than what replaced them. If a future consolidation needs to compare arms head-to-head again — rather than score a fixed corpus — reopen this or lift from the branch instead of rebuilding it.

Closing rather than leaving it open because a stale draft that no longer matches the shipped surface reads as pending work and quietly rots.

Issue #633, opened 2026-08-01
Allow download_model destination in configured extra model paths


artokun:
### What happened / root cause
download_model refuses a target_subfolder that is a symlink into a model directory already registered in ComfyUI's live extra model paths. This prevents panel-visible downloads directly to an external model drive.

### Steps to reproduce
1. Configure /Volumes/Render/00_AI/models/unet as an active ComfyUI extra model path.
2. Create ComfyUI/models/external_unet_download -> /Volumes/Render/00_AI/models/unet.
3. Call download_model with target_subfolder=external_unet_download.

### Exact error
MODEL_ERROR: Refusing to write outside the models directory (a symlink in the path escapes it): external_unet_download

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: add an optional destination selector restricted to paths already returned by the connected ComfyUI's configured extra-model paths, or allow target_subfolder to resolve into a configured extra model root after canonical-path validation. Do not weaken the existing escape guard for arbitrary paths.

### Environment
macOS Darwin, Apple MPS 24 GB, ComfyUI 0.29.0, comfyui-mcp 0.48.30, panel 0.11.20.

PR #919, opened 2026-08-06
model destination visibility: nested Diffusers false mismatch + StabilityMatrix junction refusal

### LINKED ISSUE #844: Avoid false workspace mismatch from nested Diffusers files
`download_model` refused a correct local ComfyUI workspace as “a DIFFERENT install.” Its workspace-identity heuristic sampled two ordinary nested files under `models/diffusers` (Hunyuan3D component `.bin` files) that ComfyUI's `/models` endpoint does not enumerate. It treated those expected omissions as proof that the running server did not read the workspace.

### Steps to reproduce
1. Run ComfyUI 0.30.0 from `~/repositories/wet/wet.graphics.pipeline/ComfyUI` on port 6101 with `--enable-manager`.
2. Keep a Diffusers-style nested model directory containing component files such as `hunyuan3d-paint-v2-0-turbo/vae/diffusion_pytorch_model.bin`.
3. Start comfyui-mcp with matching `COMFYUI_URL=http://127.0.0.1:6101` and `COMFYUI_PATH=~/repositories/wet/wet.graphics.pipeline/ComfyUI`.
4. Call `download_model` for any public Hugging Face model.

### Exact error (scrubbed)
```
Refusing to download into "~/repositories/wet/wet.graphics.pipeline/ComfyUI/models": the connected ComfyUI (http://127.0.0.1:6101) does not read from it. Its "diffusers" folder holds 2 model file(s), 2 of which the running server does NOT list (e.g. hunyuan3d-paint-v2-0-turbo/vae/diffusion_pytorch_model.bin, hunyuan3d-paint-v2-0-turbo/text_encoder/pytorch_model.bin) — if it scanned this directory it would see all of them, so this is a DIFFERENT install than the one serving you.
```

The systemd unit's WorkingDirectory and COMFYUI_PATH were already the same correct ComfyUI checkout.

### Fix
Applied locally: yes (configuration workaround)
Upstream-only: no

Added an explicit absolute `--base-directory ~/repositories/wet/wet.graphics.pipeline/ComfyUI` to ComfyUI's existing ExecStart. After restart, the identical `download_model` calls were accepted, resumed partial downloads, wrote to the intended workspace, and ComfyUI confirmed the completed VAE files in its model list.

Suggested upstream fix: do not use unlisted nested Diffusers component files as definitive evidence that the workspace differs. Prefer an authoritative absolute base-directory signal when available; otherwise sample only files belonging to model folders whose REST enumeration contract includes those nested files, or downgrade this mismatch to an unverified warning.

### Environment
Linux; ComfyUI 0.30.0; NVIDIA GB10, ~124GB VRAM; comfyui-mcp 0.49.5; panel 0.11.38.

Related (same root cause): #813

### LINKED ISSUE #870: download_model refuses StabilityMatrix junction model folders (vae/diffusion_models)
On StabilityMatrix installs, ComfyUI `models/vae` and `models/diffusion_models` are **NTFS junctions** pointing at `E:\comfy\StabilityMatrix\models\VAE` / `DiffusionModels`. `download_model` rejects these with:

`Refusing to write outside the models directory (a symlink in the path escapes every registered model root): vae`

`text_encoders` is a real directory under the package models root, so that download succeeds. Same host, same ComfyUI.

Root cause: symlink/junction escape check does not treat the junction *target* as a registered model root even though ComfyUI/StabilityMatrix intentionally uses junctions for shared model storage.

### Steps to reproduce
1. Run ComfyUI via StabilityMatrix (junctioned model folders).
2. call download_model with target_subfolder `vae` (or `diffusion_models`) and any public HF URL.
3. Observe MODEL_ERROR about symlink escape.
4. call download_model with target_subfolder `text_encoders` on same machine — succeeds.

### Exact error (scrubbed)
```
MODEL_ERROR: Refusing to write outside the models directory (a symlink in the path escapes every registered model root): vae
```

### Fix

Precise change needed: when resolving the write path under models/, if a path component is a junction/symlink, **resolve to the real path** and allow the write if the resolved path is still under a known model root OR if the junction itself lives under the ComfyUI models directory (StabilityMatrix pattern). Alternatively register junction targets from extra_model_paths / folder_paths as allowed write roots.

Workaround used: huggingface_hub.hf_hub_download + copy into `E:\comfy\StabilityMatrix\models\VAE` and `DiffusionModels`.

### Environment
Windows 11 / ComfyUI 0.30.0 (StabilityMatrix) / RTX 5060 Ti 16GB / comfyui-mcp 0.49.8 / panel 0.11.38 / Manager 3.x LEGACY

artokun:
Both defects lived in src/services/model-resolver.ts and both are the same fold — a listing/registration the server never makes, read as proof of a negative.

## #844 — nested Diffusers files caused a false "DIFFERENT install" refusal

Verified against ComfyUI master `folder_paths.py`: the `diffusers` category is registered with the extension contract `["folder"]`, which no file matches — so `/models/diffusers` **never lists files** (it loads those models as whole directories). The pre-write workspace check sampled nested component files (`hunyuan3d-paint-v2-0-turbo/vae/diffusion_pytorch_model.bin`, …) and read their contractual absence as proof the server doesn't read the tree.

- Categories whose listing contractually cannot name files (`CATEGORIES_WITHOUT_FILE_ENUMERATION`) now make `liveCategoryListing` return `undefined` (inconclusive = no evidence), which every caller — the pre-write refusal, `verifyLandedModel`, the manifest skip/pending paths — already honors.
- `verifyLandedModel` reports a landed diffusers file as **unknown** with an accurate note, instead of a false "the server does NOT list it".
- A test pins that a real sibling-category mismatch still refuses (the excusal is not a blind spot).

## #870 — StabilityMatrix junction model folders refused

StabilityMatrix junctions `models/vae`, `models/diffusion_models`, … into its shared store and registers nothing — the junction IS the registration, so no live-root lookup could ever authorize the escape. Per the maintainer's ruling in the issue, an in-tree symlink/junction redirect is now authorized by its **location** inside the user's own models tree (ComfyUI's scanner follows it: `recursive_search` walks with `followlinks=True`). This supersedes the #633-era P0d fail-closed authorization.

- **Kept:** the code-root (`custom_nodes`) veto — including a symlink *landing* inside one — the dangling/unresolvable-symlink refusal, and the primary-root code veto.
- **Companion honesty fix:** `verifyLandedModel`'s live-roots membership now runs on the **lexical write path** — the path the server actually scans, following links at any depth — so a file landed through a category-level or nested in-tree junction is not falsely reported "outside every directory the server reads" post-write.

## Accepted residual (called out in code + tests)

An in-tree symlink to an unregistered, non-code location (e.g. `/tmp/evil`) is now allowed: there is no structural property separating it from a legitimate junction target, both are the user's own fixture on their own machine, and a landing the server cannot see is reported honestly by the post-write check rather than pre-emptively refused.

## Verification

- Full suite: 290 files / 6219 tests green; `tsc --noEmit` clean.
- Mutation-verified: removing the contract-skip fails the #844 pre-write test; removing the verifyLandedModel branch fails its note assertion; reinstating the escape refusal fails 8 allowance tests; reverting membership to the realpath'd input fails both junction post-write tests.
- Codex adversarial gate: round 1 FAIL (3 findings — all addressed, incl. a factual error in the first message draft), round 2 PASS.

closes #844
closes #870

### DIFF
--- a/src/__tests__/services/download-dest-roots.test.ts
+++ b/src/__tests__/services/download-dest-roots.test.ts
@@ -7,16 +7,22 @@ import { join, resolve } from "node:path";
 //
 //   (a) #346 — a server with --base-directory != COMFYUI_PATH roots the download
 //       at the SERVER's models dir (where the running ComfyUI actually reads).
-//   (b) #633 — a symlinked target_subfolder resolving into a LIVE, authoritatively
-//       registered extra_model_path is ALLOWED (that's where ComfyUI reads).
-//   (c) path-safety — the symlink allowance never becomes an arbitrary-write /
-//       write-into-code-dir (RCE) hole. Covers the codex + independent-codex P0s:
+//   (b) #633/#870 — a symlink/junction INSIDE the models tree may redirect the
+//       write: into a live-registered extra root (#633), or anywhere else the
+//       code-root veto does not cover (#870 — the StabilityMatrix layout, where
+//       junctioned category folders ARE the registration mechanism and no config
+//       can vouch for their targets; ComfyUI's scanner follows the links).
+//   (c) path-safety — the symlink allowance never becomes a write-into-code-dir
+//       (RCE) hole, and a broken link is never silently followed. Covers:
 //         P0a  dangling / uncanonicalizable symlink -> REFUSED (not skipped).
 //         P0b  primary models root that resolves into custom_nodes -> REFUSED.
 //         P0c  (TOCTOU — inherent to pathname checks in Node; not asserted here).
-//         P0d  authorization is FAIL-CLOSED: only a LIVE, server-authoritative
-//              config may authorize an escape — a stale/static config cannot, and
-//              an unreachable server authorizes nothing.
+//         P0d  SUPERSEDED by #870: authorization used to require a LIVE,
+//              server-authoritative registration of the escape target. That
+//              rule could never cover a junction the server never declares, so
+//              it refused correct StabilityMatrix installs. Authorization now
+//              comes from the symlink's LOCATION (inside the user's own models
+//              tree); the code-root veto below is unaffected by the change.
 //
 // The REAL resolver + guard run; only /system_stats, the filesystem, config, and
 // the static/live registered roots are controlled.
@@ -55,10 +61,11 @@ vi.mock("../../comfyui/client.js", () => ({
   getClient: () => ({ fetchApi: (...a: unknown[]) => fetchApi(...(a as [])) }),
 }));
 
-// Two injected root sources:
+// Two injected root sources, both feeding the code-root VETO only:
 //  - getExtraModelRoots (STATIC-capable): contributes to the code-root VETO only.
-//  - getLiveExtraModelRoots (LIVE, server-authoritative): the ONLY source that may
-//    AUTHORIZE an escaping symlink (fail-closed), plus its custom_nodes veto.
+//  - getLiveExtraModelRoots (LIVE, server-authoritative): its custom_nodes roots
+//    also feed the veto. Since #870 NEITHER source authorizes anything — an
+//    in-tree symlink's redirect is authorized by its location.
 const getExtraModelRootsMock = vi.fn();
 const getLiveExtraModelRootsMock = vi.fn();
 vi.mock("../../services/extra-paths.js", () => ({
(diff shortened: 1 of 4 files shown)


artokun:

PR #920, opened 2026-08-06
download lifecycle: stale download cannot be cancelled after reconnect + ordering-dependent retry test

### LINKED ISSUE #858: Stale download survives reconnect but cannot be cancelled or safely restarted
A local download_model job survived a sidebar/session reconnect as `downloading`, but its heartbeat stayed stale for over 2200 seconds. download_status warns not to reissue, while cancel_download refuses because the job belongs to a previous server session. The expected destination and any matching `.partial` file were absent under the confirmed local workspace, so the user is left unable to recover without manually cancelling in the panel tray.

### Steps to reproduce
1. Start download_model for a large local model.
2. Reconnect/restart the tool session while it is in flight.
3. Query download_status: job remains downloading with stale heartbeat.
4. Call cancel_download with id and tray_id: it refuses due to previous-session ownership.

### Exact error
`Download ... is tracked (status: downloading) but was started by a different/previous server session, so it can't be aborted from here. Stop it from the panel download tray instead.`

### Fix
Applied locally: no. Upstream-only: yes. Allow a new session to adopt/cancel a stale local job, or mark it recoverable after verifying the writer is gone and the `.partial` is not growing. The tool should provide an atomic recovery/restart operation so agents do not risk duplicate writers.

### Environment
Windows 11; ComfyUI 0.28.2; RTX 4080 SUPER 16GB; comfyui-mcp 0.49.6; panel 0.11.38.

Related (same root cause): #761, #822

### LINKED ISSUE #869: download-retry.test.ts staged-file interference test is ordering-dependent — failed once on a CI runner, 5/5 locally
## What happened

During the 0.49.8 release, CI failed on **windows-latest** with:

```
AssertionError: expected 'Download produced a 0-byte file — the…'
  to match /another download is writing the same staged file/i

- Expected: /another download is writing the same staged file/i
+ Received: "Download produced a 0-byte file — the source sent no data. Removed it; retry."

❯ src/__tests__/services/download-retry.test.ts:649:21
```

Established it is **not a regression** before re-running:

- the release branch is `main` + CHANGELOG/version only, and **main's CI is green at the identical commit** (`2ebe8d1`);
- the changed code in that release touched `panel-pin-guard.ts`, not the download staging path;
- the file passes **5/5 locally**;
- a re-run went green on all three platforms.

## Why it is worth naming rather than tolerating

This is a race in a test **about** a race, which is exactly the kind that gets waved through as noise. Both messages are legitimate outcomes of the same setup — which check observes the staged file first decides the verdict — so the test is asserting on a scheduling outcome rather than on a property.

The suite's other known load-sensitivity was diagnosed and fixed in **#852 / #853** (runner starvation, closed by giving the suite a real timeout). This one is different: it is not a timeout, it is genuine non-determinism, so a longer budget will not touch it.

And the precedent this project has set twice is that the fix is to **remove the mechanism, not widen the tolerance** — #821/#843 rebuilt a fixture around `listen(0)` rather than raising a timeout, and comfyui-mcp-panel#652 made an injected delay resolve instantly rather than lengthening a budget. A test that passes or fails on which of two checks wins should be made to control that ordering, not re-run until it agrees.

## Steps to reproduce

Not reliably reproducible on demand — that is the report. It fired once on a GitHub-hosted windows runner under full-suite load and has not reproduced in 5 local runs of the file.

## Fix

Applied locally: no.

Make the interference case **deterministic**: drive the competing writer to a known point before the assertion, so the test observes the interference branch by construction rather than by scheduling. If both outcomes are genuinely valid for that setup, then the assertion is wrong as written and should either accept both explicitly (saying why) or split into two tests that each control which check fires first.

Worth checking the neighbouring cases in the same file at the same time — an ordering assumption that is wrong in one place is rarely alone.

artokun:
## What this fixes

**#858 — a stale download survived reconnect but could not be cancelled or safely restarted.** A download whose owning session died kept a persisted `downloading` record with a stale heartbeat: `download_status` warned "do not re-issue" while `cancel_download` refused as foreign-owned — a wedge with no tool-level exit. Persisted records now carry the writer's **pid** (`src/services/download-progress.ts`), and `cancelDownloadJob` distinguishes **proven dead** from **unknown**:

- **Proven dead** (`process.kill(pid, 0)` → ESRCH) → the record is *reclaimed*: this session's terminal `cancelled` record (marked `reclaimed_dead`) is made durable **first**, the dead owner's record file removed **second** (deletion outcome reported; a leftover is disclosed, not hidden), and the dead writer's tray-row ids are cleared only when no possibly-live job shares them — a *stale-but-unproven* sibling still suppresses the clear, because a stale heartbeat is not death (#761).
- **Alive / unprovable** (live pid, no pid, probe error) → refused exactly as before, with a message that says which case and why, pointing at the panel download tray.
- A foreign **settled** record now reports "already **done/cancelled/error** — nothing to cancel" instead of the old misleading "can't be aborted from here".

`download_status`'s stale-heartbeat note now names the recovery path, and a reclaimed record renders as an administrative close — it never claims a live transfer was aborted or that a partial was deliberately left.

**#869 — the staged-file interference test was ordering-dependent.** It raced a background writer (poll for 4 bytes, sleep 250 ms, overwrite) against the retry loop's attempt unwinding; on a loaded CI runner the unwinding observed the file mid-overwrite (0 bytes) and took the 0-byte refusal branch. The retry loop calls `abortableDelay` exactly once per backoff — strictly **after** recording its baseline snapshot of the staged file and strictly **before** the next attempt re-proves it — so the foreign write now happens inside a one-shot hook in a module-mocked `abortableDelay`: deterministic by construction, no timeouts widened. All three timer-based interference tests converted (HTTP size-change, same-size validator-swap, cloud).

## Verification

- Full suite green (290 files / 6223 tests); `download-retry.test.ts` 20/20 runs; the two download test files 10/10.
- Mutation-verified: size-check and sidecar-check mutants fail the rewritten #869 tests; dead-probe, reclaim-branch, row-sharing (both id shapes + stale sibling), and deletion-failure mutants each fail their #858 tests.
- Codex gate: 3 rounds (cap), 4 findings total, all fixed and re-verified — tray-row overclaim in the reclaim message, single-id row guard, silent deletion failure, fresh-only sibling guard.
- Committed blobs scanned for control bytes: clean.

## Deliberately not done

- No atomic cross-process lease for the start-two-writers-at-the-same-instant race — documented best-effort scope (#529); the reclaim does not change it.
- A cancel that lands while the writer's pid is alive but its heartbeat is stale still refuses — that is the #761-safe direction, now with an explicit message.
- Pre-#858 records (no pid) are never reclaimed — unprovable is not dead.

closes #858
closes #869


artokun:

PR #921, opened 2026-08-06
generation defaults: seed randomization ignores declared max + checkpoint auto-select picks video models

### LINKED ISSUE #865: enqueue_workflow's seed randomization ignores the node's declared max, producing an instant 400 validation failure (e.g. ClaudeNode seed max 2147483647)

`enqueue_workflow` auto-randomizes any input named `seed`, but it draws from a range wider than the target node's **declared** `max`. It does not read the node's `/object_info` INPUT_TYPES config before substituting, so any node whose seed max is smaller than the randomizer's range is **guaranteed** to fail validation.

Hit with core's `ClaudeNode` (`comfy_api_nodes/nodes_anthropic.py`), whose seed is declared `min: 0, max: 2147483647` (int32). I explicitly passed `seed: 1`; the tool replaced it with `2328903345` (> 2^31-1) and ComfyUI rejected the prompt outright.

Note the value I sent was overwritten — this is not "a default was filled in", it is an **explicit user-provided value being replaced with an invalid one**.

### Steps to reproduce

```
call_tool enqueue_workflow {
  "workflow": {
    "1": { "class_type": "ClaudeNode", "inputs": {
             "prompt": "hello", "model": "Sonnet 5", "seed": 1 } },
    "2": { "class_type": "PreviewAny", "inputs": { "source": ["1", 0] } }
  }
}
```

### Exact error (scrubbed)

```
ComfyUI rejected the workflow (400): Prompt outputs failed validation
- ClaudeNode (node 1): Value 2328903345 bigger than max of 2147483647 (seed)

node_errors.1.errors[0]:
  type: value_bigger_than_max
  input_name: seed
  input_config: ["INT", { "default": 0, "min": 0, "max": 2147483647,
                          "control_after_generate": true }]
  received_value: 2328903345
```

### Impact

Any node with a sub-2^32 seed max is unusable through `enqueue_workflow` unless the caller knows to pass `disable_random_seed: true`. Nothing in the error points at the MCP as the source — it reads like the user supplied a bad seed, so the natural (wrong) next move is to "fix" a seed value that was already correct. This is squarely in the path of the documented headless fallback for `panel_run`, which is how I hit it (I was working around panel issue #659).

### Fix

**applied locally: no** · **upstream-only: yes** (the randomizer lives in the MCP server, not in anything I can safely patch from the panel side without a rebuild)

Precise change needed, wherever the seed substitution happens in the enqueue path:

1. Look up the target node's input config from the memoized `/object_info` snapshot (the same snapshot `get_node_info` serves) and clamp the generated value to `[min, max]` for that node's seed input — do not use a single global range.
2. When the config is unavailable (unknown class_type, server unreachable), fall back to the **int32** range `0..2147483647` rather than a wider one. Every stock ComfyUI seed widget is int32 or int64; int32 is valid for both, so it is the safe default and would have avoided this entirely.
3. Consider **not overwriting an explicitly supplied seed at all** — `disable_random_seed` defaults to false, so a caller who deliberately passes `seed: 1` for a reproducible/cacheable run silently loses it. At minimum, only randomize inputs the caller omitted.

(3) is arguably the real bug: the surprising part was not the range, it was that an explicit value was replaced.

### Environment

- OS: Windows 11 Pro for Workstations 10.0.26200
- GPU: NVIDIA GeForce RTX 5090 (32 GB VRAM), 127 GB RAM, CUDA 13.0
- torch 2.10.0, python 3.12
- ComfyUI 0.30.0 (Comfy-Desktop install)
- **comfyui-mcp 0.49.7**
- **comfyui-mcp-panel 0.11.38**
- Workaround that works: `disable_random_seed: true`

### LINKED ISSUE #892: generate_image checkpoint auto-select picks the alphabetically-first checkpoint, which fails outright if it is a video model
## What happens

`generate_image` documents "auto-selecting a local checkpoint when none is given". On this rig that auto-selection is deterministic and always picks `Ltx2.3\ltx2310eros_v14.safetensors` — the alphabetically-first entry in `list_local_models` checkpoints. That is an LTX **video** checkpoint with no CLIP/text encoder, so the generated txt2img graph fails 100% of the time at CLIPTextEncode.

Net effect: a parameterless `generate_image` call — the documented "high-level entry point" and the first thing a model reaches for — never works on a rig whose alphabetically-first checkpoint is a video (or otherwise non-txt2img) model. The failure is also confusing, because the error blames the graph, not the selection.

## Reproduce

```
call_tool generate_image {"prompt":"a single red apple on a plain white table","width":512,"height":512,"steps":8}
```

Returns:
```
{"status":"enqueued","prompt_id":"...","checkpoint":"Ltx2.3\\ltx2310eros_v14.safetensors"}
```

Then `queue action:"status"` on that prompt_id:
```
"status_str": "error",
"node_id": "3", "node_type": "CLIPTextEncode",
"exception_type": "RuntimeError",
"exception_message": "ERROR: clip input is invalid: None\n\nIf the clip is from a checkpoint loader node your checkpoint does not contain a valid clip or text encoder model."
```

Reproduced twice, same checkpoint chosen both times. Passing `checkpoint: "v1-5-pruned-emaonly-fp16.safetensors"` explicitly succeeds in 2.7s, so nothing else in the path is wrong.

## Suggested fix

Auto-select should prefer a checkpoint that can actually drive the graph it is about to build, rather than taking index 0. At minimum, skip checkpoints known to lack a CLIP/text encoder for txt2img, and if none qualifies, fail fast with "no txt2img-capable checkpoint found, pass `checkpoint` explicitly" instead of enqueuing a graph that is guaranteed to error.

## Secondary observation (separate, lower severity)

Unknown argument keys are silently stripped rather than rejected: calling `generate_image` with `model:` (instead of the real parameter name `checkpoint:`) drops the key without complaint and renders with the auto-selected checkpoint. A caller who guesses the wrong parameter name gets a silently wrong render rather than an error. This matters more after the #726 consolidation, where each action-parameterized tool has a much wider flat schema and a field misspelled for your action is silently ignored the same way.

## Environment

ComfyUI 0.30.2, frontend 1.47.12, Windows 11, found on port 8188. Discovered during live validation for the #726 0.50.0 tool consolidation.

artokun:
## What

**#865 — seed randomization vs. declared max + explicit seeds.** `randomizeSeeds` drew `Math.random()*2**32`, overwrote even explicit seeds, and never consulted the node's declared max — a node like core's ClaudeNode (seed max 2147483647) failed validation with a 400 on every enqueue.

- Draws now respect the node's declared `[min, max]` from the memoized `/object_info` snapshot; unknown node/bounds → int32 fallback (valid for every stock seed widget). Declared maxes above 2^53 (KSampler's 2^64-1) clamp the draw ceiling to `MAX_SAFE_INTEGER` so a top-of-range draw can't round past the true max; fractional declared bounds are rounded inward; bounds with no safely-drawable integer range refuse the draw and keep the caller's value.
- **`enqueue_workflow` no longer re-randomizes at all** — every seed in a caller-built workflow is caller-fixed (the issue's core complaint). The `disable_random_seed` param is kept as a documented no-op for compatibility. `rerun_generation`/`regenerate` keep their documented re-randomize-by-default, and a new `preserve_seed_inputs` executor option stops that from clobbering a seed the caller pinned via overrides (regenerate's own description already promised this).
- `generate_image` enqueues with executor randomization off: the composer already draws a random seed when none is given, so the executor pass only ever clobbered the resolved seed (explicit `seed:` was silently broken before).

**#892 — checkpoint auto-select picks video models.** `resolveCheckpoint` returned the alphabetically-first checkpoint; on a rig whose first checkpoint is an LTX/Wan video model (no text encoder) the txt2img graph fails at CLIPTextEncode 100% of the time.

- New `checkpoint-capability` service probes each listed checkpoint's safetensors header (first bytes only, mtime+size-cached): known TE prefixes → `capable`; recognized diffusion/VAE layout with no TE → `incapable`; anything undeterminable (remote server, .ckpt/.pt/.bin, unreadable or unrecognized layout) → `unknown`, deliberately never folded into `incapable`.
- Auto-select prefers known-capable, falls back to unknown (status quo), and fails fast with an actionable error only when EVERY candidate is known incapable.

## Behavior-change note

`enqueue_workflow` no longer randomizes seeds by default — resubmitting the same workflow now reproduces the run (ComfyUI cache included). Fresh-seed re-runs are `rerun_generation`'s documented job. `run_template`/`batches`/`run_workflow_url` keep randomize-by-default (now range-safe).

## Verification

- Full suite green (294 files / 6246 tests), `tsc --noEmit` clean, docs regenerated (`gen-tool-docs`), vocabulary gate OK.
- Mutation-verified: old 2^32 draw (4 tests fail), preserve-guard removal, `min>max → fallback` instead of null (3 fail), capability filter removal (5 fail), tool-layer selection bypass (2 fail), diffusion-layout tier removal (3 fail), ceil/floor removal (1 fail).
- Codex gate: 3 rounds — r1 NO-SHIP (5 findings: unsafe-range fallback, unobserved capability claim in description, non-actionable remedies, 2 weak tests) → fixed; r2 (2 findings: unrecognized-layout-as-incapable, fractional/unsafe declared bounds) → fixed; r3 (1 test-quality finding) → fixed.

## Deliberate non-fixes

- Other generate-* tools (video/audio/conditioned/upscale/remove-background) have the same explicit-seed clobber; untouched here — PR #907 is rewriting those call sites, flagged for coordination.
- Omitted seed inputs are not injected (the issue's optional "at minimum" extension); omitted required inputs fail ComfyUI validation with its own clear error.
- `batches` sweep sets naming `seed` still get re-randomized per that tool's documented behavior.

closes #865
closes #892

### DIFF
--- a/docs/tools/image-generation.mdx
+++ b/docs/tools/image-generation.mdx
@@ -10,7 +10,7 @@ icon: "image"
 
 ## generate_image
 
-Generate an image from a text prompt — the high-level entry point. Builds a txt2img workflow, filling any unspecified parameter from your configured defaults (get_defaults (action:"set") / COMFYUI_DEFAULT_* / config file), auto-selecting a txt2img-capable local checkpoint when none is given (checkpoints without a text encoder, e.g. video models, are skipped). Returns the prompt_id immediately; the resulting asset_id arrives in the completion notification and can be passed to view_image or regenerate. For full control over the node graph, use create_workflow + enqueue_workflow instead.
+Generate an image from a text prompt — the high-level entry point. Builds a txt2img workflow, filling any unspecified parameter from your configured defaults (get_defaults (action:"set") / COMFYUI_DEFAULT_* / config file), auto-selecting a local checkpoint when none is given — checkpoints known to lack a text encoder (e.g. video models) are skipped. Returns the prompt_id immediately; the resulting asset_id arrives in the completion notification and can be passed to view_image or regenerate. For full control over the node graph, use create_workflow + enqueue_workflow instead.
 
 <Tip>**In plain terms:** The one-line way to get a picture. You do not need a workflow open, or any workflow at all — describe the image and this builds and runs a sensible graph for you. Everything except the prompt is optional and falls back to your saved defaults.</Tip>
 
--- a/docs/tools/image-generation.mdx
+++ b/docs/tools/image-generation.mdx
@@ -10,7 +10,7 @@ icon: "image"
 
 ## generate_image
 
-Generate an image from a text prompt — the high-level entry point. Builds a txt2img workflow, filling any unspecified parameter from your configured defaults (get_defaults (action:"set") / COMFYUI_DEFAULT_* / config file), auto-selecting a local checkpoint when none is given. Returns the prompt_id immediately; the resulting asset_id arrives in the completion notification and can be passed to view_image or regenerate. For full control over the node graph, use create_workflow + enqueue_workflow instead.
+Generate an image from a text prompt — the high-level entry point. Builds a txt2img workflow, filling any unspecified parameter from your configured defaults (get_defaults (action:"set") / COMFYUI_DEFAULT_* / config file), auto-selecting a txt2img-capable local checkpoint when none is given (checkpoints without a text encoder, e.g. video models, are skipped). Returns the prompt_id immediately; the resulting asset_id arrives in the completion notification and can be passed to view_image or regenerate. For full control over the node graph, use create_workflow + enqueue_workflow instead.
 
 <Tip>**In plain terms:** The one-line way to get a picture. You do not need a workflow open, or any workflow at all — describe the image and this builds and runs a sensible graph for you. Everything except the prompt is optional and falls back to your saved defaults.</Tip>
 
--- a/docs/tools/workflow-execution.mdx
+++ b/docs/tools/workflow-execution.mdx
@@ -10,7 +10,7 @@ icon: "play"
 
 ## enqueue_workflow
 
-Submit a ComfyUI workflow for execution and return immediately with the prompt_id and queue position. Does not wait for completion. Use queue (action:"status") to check progress later, or get_history to retrieve results and images after completion.
+Submit a ComfyUI workflow for execution and return immediately with the prompt_id and queue position. Does not wait for completion. Seed values in the workflow are used exactly as supplied — they are NOT re-randomized, so a run is reproducible by resubmitting the same workflow (for a fresh-seed re-run of a past job, use rerun_generation). Use queue (action:"status") to check progress later, or get_history to retrieve results and images after completion.
 
 <Tip>**In plain terms:** Runs a workflow that the agent is holding in the conversation — one it just built or loaded from a file. If you want to run the graph you are LOOKING at in ComfyUI, that is the panel's run tool instead, and if you just want a picture, use generate_image.</Tip>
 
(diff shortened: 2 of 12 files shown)


artokun:

PR #922, opened 2026-08-06
test state isolation: full suite can rewrite real ~/.comfyui-mcp state

### LINKED ISSUE #866: Third instance: tests write to real ~/.comfyui-mcp state. Guard at the WRITE, not per-test
## What happened

The suite has now been caught writing to the developer's real `~/.comfyui-mcp` state **three times**, in three different files, each found by accident while chasing something else:

| # | file | found by |
|---|---|---|
| #837 | `.env` — clobbered live API tokens | the secrets agent |
| #859 | the OAuth status mirror (`panel-secrets.json`) | two agents independently, neither able to attribute it |
| this | `panel-pending-ops.json` | chasing a gate finding on #847 |

The third one is live-harmful rather than merely noisy: the orchestrator **reads** that file, so the markers a test run leaves behind make it warn on every pin write that a queued `update_all` or deferred restore may be outstanding. Two such markers were found in the real file and removed.

## Why per-test scoping keeps failing

Each fix so far has been "remember to point the env var at a temp path". That is a convention, and conventions are exactly what a 250-file suite erodes. The specific failures were all *plausible-looking*:

- `node-snapshots.test.ts` set `COMFYUI_MCP_PANEL_PENDING = "\0"` intending an fs-rejected path. **Node truncates a NUL assignment to `""`**, and `panelPendingOpsPath()` treats an empty env var as *unset* — so it silently fell back to the real home. The test asserted nothing it claimed **and** wrote to real state. It read as correct.
- `update-comfyui.test.ts` scoped the path **inside one test** while seven other `updateAllCustomNodes()` calls in the same file did not.
- `panel-pin-bypass.test.ts` and `node-snapshots-tool.test.ts` never scoped it at all.

A rule that can be broken by forgetting one line, or by an env assignment quietly becoming empty, is not a guard.

## Fix

Applied locally: no — the three files above are scoped in the #847 PR, but that is the symptom.

**#837 already built the right mechanism**: it made the credential-store write **refuse at runtime** when a test runner is detected and the redirect env var is not set. It cannot be evaded by how a test spells its import, and it fires at the write rather than after the damage. Extend that same guard to the other real-home writers — at minimum `panel-secrets.json` (including the OAuth mirror) and `panel-pending-ops.json`, and audit for others.

Two constraints learned from #837's version, both worth carrying over:

1. **Detect the runner, not an env var a user could plausibly set.** #837's first cut keyed on `NODE_ENV=test` / a leftover `VITEST` variable, which would refuse a legitimate local user their own write. Use something the runner injects (`process.env.VITEST_WORKER_ID`, a vitest global).
2. **When detection is uncertain, allow the write.** Refusing a real user access to their own state is worse than the leak this prevents.

Also worth fixing while there: `panelPendingOpsPath()` treating an **empty** env var as unset is what made the NUL sentinel silently harmless-looking. An explicitly-set-but-empty redirect is far more likely to be a mistake than an intent to use the default, and failing loudly there would have surfaced this years earlier.

Related (same root cause): #859

### LINKED ISSUE #879: Full test suite intermittently rewrites the developer's real ~/.comfyui-mcp/.env
Observed while verifying the `remote-and-secrets` branch (#837), which is the branch that *adds* the runtime isolation guards. The guards work for every path I could isolate — this is a hole they do not cover.

## What I observed

Running the full suite (`npx vitest run`, 282 files) rewrote `~/.comfyui-mcp/.env`:

```
before: sha256=3179ca4ff50b89ba mtime=1785954687
 Test Files  282 passed (282)
after:  sha256=3179ca4ff50b89ba mtime=1785954830
CONTENT UNCHANGED
!! REWRITTEN by the full suite
```

The **content was byte-identical**, so nothing was lost this time. That is luck, not design: a rewrite that happens to be idempotent today is one refactor away from destroying live tokens. This is the fourth instance of the "suite writes the developer's real state" class (after the `.env` in #837, the OAuth mirror in #859, and `panel-pending-ops.json` in #866).

## It is intermittent

- Full run #1 — **touched**
- Targeted secrets tests (`panel-secrets`, `env-file`, `env-store-atomic-write`, `secret-store-test-isolation`) — clean
- `src/__tests__/orchestrator`, `services`, `tools` each on their own — clean
- Root-level `*.test.ts`, `comfyui/`, `transport/`, `utils/`, `packs/`, `experimental/` each on their own — clean
- Full run #2 — **not touched**

So no single file reproduces it. It needs the full parallel run, which points at an interaction or a worker/main-process split rather than one bad test.

## The likely mechanism

`assertEnvStoreRedirectedInTests` / `assertPanelSecretsRedirectedInTests` key off globals **Vitest injects into worker scope**, and deliberately ALLOW the write when they cannot prove they are under the runner:

> When none is present we do NOT know we are under a runner — and in that uncertainty the write is ALLOWED, deliberately: the failure mode we accept is "a rogue test slips past the runtime guard and is caught by the static sweep instead", never "a real user cannot save a token".

That trade-off is right, and I am not proposing to change it — refusing a real user their own credential is the worse failure. But it means anything writing from **outside a worker** (main process, a global setup/teardown, a module-level side effect evaluated during collection) sails straight through. That matches the evidence: no individual file reproduces it, and the guards never fired.

## Suggested direction

1. Find the writer rather than widening the guard. A temporary `fs.writeFileSync`/`renameSync` trap installed for the whole run, logging a stack whenever the target resolves to the canonical env path, should name it in one run.
2. If it is main-process or collection-time, the fix is to redirect `COMFYUI_MCP_ENV_FILE` for the **entire run** via `globalSetup`, not per-test — which also closes the other three instances of this class permanently.
3. Keep the uncertainty-allows-the-write rule exactly as it is. The static sweep is the right backstop for the runner side.

## Why this matters beyond the tests

Anyone running the suite on a machine with real tokens in `~/.comfyui-mcp/.env` is one non-idempotent rewrite away from losing them, with no backup taken (the only backup present here is from 2026-07-21). Refs #837, #859, #866.

artokun:
vitest has no global HOME redirect; assertPanelSecretsRedirectedInTests deliberately fails open outside vitest workers; other real-home writers (panel-pin-guard pending-ops) are unguarded. Guard at the WRITE, not per-test.

closes #879
closes #866


artokun:

PR #923, opened 2026-08-06
turn lifecycle: Claude backend stall still reported as user rejection + turn registry does not survive reconnect

### LINKED ISSUE #885: Stalled turn is reported to the agent as a USER REJECTION (agent then falsely tells the user they cancelled)

During a long session, an agent turn stalled with no activity during a Bash tool call. The panel correctly told the **user**:

> The agent stopped responding (the turn stalled with no activity). I've cleared it — please try again.

But what the **agent** received as the tool result was the standard user-rejection payload:

> The user doesn't want to proceed with this tool use. The tool use was rejected (eg. if it was a file edit, the new_string was NOT written to the file). STOP what you are doing and wait for the user to tell you how to proceed.

A harness-side stall and a deliberate user rejection are semantically opposite, but they share one message to the agent. The agent has no way to distinguish them.

**Consequence (why this is worth fixing):** the agent obeyed the rejection semantics, stopped, and then told the user *"you stopped the Trellis run"*. That is a false statement about the user's own actions, and the user had to correct it ("I didn't stop anything"). The agent also abandoned a multi-step debugging task that was mid-flight and had just landed two real fixes — work that should simply have been retried.

### Steps to reproduce

1. Start a long-running agent turn in the panel that issues a Bash tool call.
2. Let the turn stall until the panel's watchdog clears it.
3. Observe: the user sees a "turn stalled / I've cleared it" notice, while the agent receives the user-rejection message.

### Exact error (scrubbed)

Agent-visible tool result:
```
The user doesn't want to proceed with this tool use. The tool use was rejected
(eg. if it was a file edit, the new_string was NOT written to the file).
STOP what you are doing and wait for the user to tell you how to proceed.

Note: The user's next message may contain a correction or preference.
```
User-visible panel message (screenshot supplied by user):
```
⚠️ The agent stopped responding (the turn stalled with no activity).
I've cleared it — please try again.
```

### Fix

**applied locally: no**  **upstream-only: yes**

Precise change needed: when the watchdog clears a stalled turn, emit a **distinct** tool result to the agent rather than reusing the user-rejection string. Suggested wording:

```
The turn stalled with no activity and was cleared by the harness.
The user did NOT reject this tool call. It is safe to retry the same call.
```

This lets the agent retry instead of halting, and — critically — stops it from attributing the stop to the user. Anything that preserves the stall-vs-rejection distinction would fix it.

### Related, lower-severity issues observed in the same session (matched current versions)

2. **MCP disconnect/reconnect churn.** Ambient notices repeatedly announced ~264 deferred tools going away and returning (`mcp__blender-mcp__*` 22, `mcp__browsermcp__*` 12, `mcp__chrome-devtools__*` 29, `mcp__civitai__*` 53, `mcp__comfyui__*` 3, `mcp__open-design__*` 18, `mcp__panel__*` 91, `mcp__web-rig__*` 33). Many times in one session, including mid-task, correlated with "connection dropped mid-task" resumes.

3. **Background Workflow runs are silently killed by a panel/agent restart.** A background deep-research Workflow later reported `No completion record was found for background workflow ... it may have been running when the previous Claude Code process exited.` Happened twice. `panel_restart_comfyui` / `panel_reload` take the agent down with them. Suggest either surviving the restart or warning at launch that a restart will kill in-flight background work.

4. **Workflow resume by `resumeFromRunId` does not carry `args`.** `Workflow({scriptPath, resumeFromRunId})` returned instantly with `{"error":"No research question provided. Pass it as args: Workflow({name: 'deep-research', args: '<question>'})"}`. Cached agent results are restored but the `args` global is not, so the script exits immediately. Either persist `args` with the run, or fail loudly at resume explaining they must be re-passed.

5. **`panel_restart_comfyui` confirmation card times out when the tab is backgrounded.** Returned `No confirmation received within 90s, so I did NOT restart ComfyUI. The panel tab may be backgrounded or still reconnecting...` even though the user had already approved the restart in chat. Worked around via the headless `restart_comfyui` MCP tool. Suggest a longer timeout, or an OS-level notification when the tab isn't visible.

### Environment

- OS: Windows 11 (10.0.26200)
- GPU: NVIDIA RTX 3070, 8 GB VRAM
- ComfyUI: 0.30.1 Desktop @8188 (py 3.13.12, torch 2.10.0+cu130)
- Also running: portable ComfyUI 0.3.49 @8190 (py 3.12.10, torch 2.7.1+cu128)
- **comfyui-mcp: 0.49.3** (= npm latest)
- **comfyui-mcp-panel: 0.11.38** (= registry latest, published 2026-08-03T06:30)

Version skew explicitly ruled out via the `panel-version-sync` skill before filing — both halves verified current, and `panel_refresh_nodes` confirmed working end-to-end.

Session shape: several hours, driving two ComfyUI instances, background workflows, and heavy `panel_*` graph editing.

### LINKED ISSUE #886: Turn reported as "unverifiable" — result could not be matched to any submitted turn (claude-opus-5)
### What happened

A completed agent turn was rejected by the panel with:

> ⚠️ The claude-opus-5 turn failed: The result could not be matched to any submitted turn — reporting the turn as unverifiable rather than a success.
>
> Nothing was lost — try again, switch models from the composer picker, or check the terminal running the orchestrator for more detail.

The turn had in fact done real work (tool calls executed, files written). The panel could not correlate the returned result with a submitted turn record, so it discarded the outcome as unverifiable.

### Context — likely related to the same session's instability

This is the same long session as issue artokun/comfyui-mcp#885. Throughout it, the orchestrator repeatedly announced MCP servers disconnecting and reconnecting (~264 deferred tools cycling out and back in, including `mcp__panel__*` itself, 91 tools). A turn submitted before a reconnect and returning after it is a plausible cause of the submitted-turn record being lost — i.e. this may be a **symptom of the reconnect churn rather than an independent defect**. Worth checking whether the submitted-turn map is held in memory that is cleared or re-created on orchestrator reconnect.

### Steps to reproduce

Not deterministically reproduced. Observed conditions:
1. Long-running panel session (several hours) driving two ComfyUI instances plus background workflows.
2. Repeated MCP disconnect/reconnect notices during the session.
3. A turn completes and its result cannot be matched to a submitted turn → reported as unverifiable.

### Exact message (scrubbed)

```
⚠️ The claude-opus-5 turn failed: The result could not be matched to any submitted
turn — reporting the turn as unverifiable rather than a success.

Nothing was lost — try again, switch models from the composer picker, or check the
terminal running the orchestrator for more detail.
```

### Suggested investigation

- Does the submitted-turn registry survive an orchestrator/MCP reconnect? If it lives in memory recreated on reconnect, any in-flight turn spanning a reconnect would become unmatchable.
- Consider persisting submitted-turn ids across reconnects, or reconciling by a stable turn id rather than an in-memory handle.
- If a match genuinely can't be made, consider surfacing the turn's actual tool activity as "completed but unverified" instead of discarding it — in this case real work had been done and the user was told the turn failed.

### Fix

**applied locally: no**  **upstream-only: yes**

### Environment

- OS: Windows 11 (10.0.26200)
- GPU: NVIDIA RTX 3070, 8 GB VRAM
- ComfyUI: 0.30.1 Desktop @8188 (py 3.13.12, torch 2.10.0+cu130); also portable ComfyUI 0.3.49 @8190 (py 3.12.10, torch 2.7.1+cu128)
- **comfyui-mcp: 0.49.3** (npm latest)
- **comfyui-mcp-panel: 0.11.38** (registry latest)
- Model: claude-opus-5

Both components verified current via the `panel-version-sync` skill before filing; version skew ruled out. Likely related to artokun/comfyui-mcp#885 (same session, same reconnect churn).

artokun:
recoverStalledTurn exists only on the Codex backend (codex-backend.ts:1411); a stalled Claude turn still goes straight to interrupt() and the agent tells the user they cancelled. And the claude-backend turn-trace registry is in-memory only, dropped on session (re)start — no persistence across reconnects, no 'completed but unverified' surfacing.

closes #885
closes #886


artokun:

PR #924, opened 2026-08-06
self_update on Windows: EBUSY on own sharp dll, npm error swallowed; updater failure + cancelled restart disconnects bridge

### LINKED ISSUE #912: self_update can never succeed on Windows (EBUSY on its own sharp libvips-42.dll) and swallows the npm error
`self_update {action:'update'}` on a **global Windows install** returns:

```
{ "action": "unavailable", "mode": "global", "from": "0.49.3", "to": "0.49.8",
  "note": "npm update to 0.49.8 failed; staying on 0.49.3." }
```

No cause, nothing actionable. Running the same command by hand shows the real reason:

```
npm error code EBUSY
npm error syscall copyfile
npm error path ~/AppData/Roaming/npm/node_modules/comfyui-mcp/node_modules/@img/sharp-win32-x64/lib/libvips-42.dll
npm error dest ~/AppData/Roaming/npm/node_modules/.comfyui-mcp-I7rTslI1/node_modules/@img/sharp-win32-x64/lib/libvips-42.dll
npm error errno -4082
npm error EBUSY: resource busy or locked, copyfile ...libvips-42.dll -> ...libvips-42.dll
```

Two distinct problems:

1. **Structural (Windows):** the orchestrator process that runs `self_update` has `sharp` loaded, which holds an open handle on `@img/sharp-win32-x64/lib/libvips-42.dll`. Windows refuses to move/overwrite a mapped DLL, so npm's staging copy fails with EBUSY. This means an in-process global `self_update` on Windows fails **every time** — it is not a transient/flaky condition. (Linux/macOS are unaffected because unlink-while-open is legal there, which is likely why this went unnoticed.)
2. **Diagnostics:** `runNpm` is destructured as `const { ok } = await deps.runNpm(args, ...)` in `dist/services/self-update.js:321`, so npm's stdout/stderr is discarded and the user gets a generic "failed" note at `:328` (same pattern again at `:468`). The agent cannot tell EBUSY from a 404, a registry timeout, or an EACCES, so it cannot advise a fix.

### Steps to reproduce
1. Windows, global install (`npm i -g comfyui-mcp`), orchestrator running (sidebar panel connected).
2. Be behind the latest published version.
3. Call `self_update {action:'update'}` → `action:"unavailable"`, generic note.
4. Run `npm install -g comfyui-mcp@latest` in a shell *while the orchestrator is still running* → EBUSY on `libvips-42.dll`.
5. Stop the orchestrator, re-run the same npm command → succeeds.

### Fix

Not patched locally on purpose: the in-flight update replaces `dist/` outright, so a local patch would be clobbered — and the fix belongs in the released package.

Precise changes needed, in `src` behind `dist/services/self-update.js`:

1. **Surface the npm output.** Keep the full result from `runNpm` instead of destructuring only `ok`, and include a trimmed tail of stderr in the failure note at both `:328` and `:468`:

```js
-        const { ok } = await deps.runNpm(args, info.mode === "local" ? info.projectRoot : undefined);
+        const res = await deps.runNpm(args, info.mode === "local" ? info.projectRoot : undefined);
+        const { ok } = res;
         if (!ok) {
+            const err = String(res.stderr ?? res.output ?? "").trim().split(/\r?\n/).slice(-8).join("\n");
+            const busy = /EBUSY|EPERM|resource busy or locked/i.test(err);
             return {
                 action: "unavailable",
                 mode: info.mode,
                 from: info.currentVersion,
                 to: latest,
-                note: `npm update to ${latest} failed; staying on ${info.currentVersion}.`,
+                reason: busy ? "locked-by-running-process" : "npm-failed",
+                note: busy
+                    ? `npm update to ${latest} failed: files are locked by the RUNNING orchestrator ` +
+                      `(Windows holds sharp's libvips DLL open). Staying on ${info.currentVersion}. ` +
+                      `Disconnect the panel / stop the orchestrator, then run: npm i -g ${PACKAGE_NAME}@latest`
+                    : `npm update to ${latest} failed; staying on ${info.currentVersion}.\n${err}`,
             };
         }
```

2. **Make the Windows path actually work.** In-process replacement can't work while the DLL is mapped, so on `process.platform === 'win32'` + `mode === 'global'`, spawn a **detached** updater that outlives this process (retry `npm i -g <pkg>@latest` until the lock clears, then exit), and return `action:"scheduled"` so the agent can honestly report "it will land once the orchestrator restarts" instead of a bare failure. That is exactly the workaround I had to hand-roll here (a detached PowerShell retry loop) — it belongs in the tool.

3. Optionally, only for the check path: `self_update {action:'status'}` should note on Windows that applying an update requires an orchestrator restart, so the agent doesn't promise an in-place update it can't deliver.

### Environment
OS: Windows 11 Pro 10.0.26200 | ComfyUI 0.30.0 (REMOTE) | GPU: RTX 5090 32GB, 127GB RAM, CUDA 13.0, torch 2.9.1, py3.12 | comfyui-mcp 0.49.3 (global npm install) | panel 0.11.36 | node on Windows, npm global prefix under ~/AppData/Roaming/npm

### LINKED ISSUE #916: Official updater failures and cancelled restart disconnects panel bridge
### What happened

While recovering a known stale-workflow-UUID panel bug, the worker recommended upgrading from comfyui-mcp 0.49.5 / panel 0.11.38. Both official update paths failed even though the install was local and writable.

### Steps / exact results

1. call self_update(action:"update")
```
{"action":"unavailable","mode":"local","from":"0.49.5","to":"0.49.8","note":"npm update to 0.49.8 failed; staying on 0.49.5."}
```
No underlying npm error was returned. A manual npm install against the same package root succeeded and self_update(status) then verified 0.49.8.

2. call install_panel(action:"update")
The tool said Manager could not resolve comfyui-agent-panel and refused the source fallback because the running ComfyUI did not corroborate the configured root, despite the connected local instance and the pack being present on disk. Manual clone/replacement to the same verified path succeeded; install_panel(status) then verified 0.11.41 with no shadows.

3. panel_restart_comfyui was cancelled via its confirmation card. Immediately afterward panel_reload and panel_set_workflow_target both reported:
```
Panel not reachable: no panel connected
```
The graph bridge disconnected despite the restart being cancelled.

### Fix

Applied locally: manual package updates only; no source patch.
Upstream-only: yes.

Please expose the underlying self_update npm error, improve local-root corroboration guidance/fallback, and ensure a cancelled restart cannot leave the panel bridge disconnected.

### Environment

Connected ComfyUI 0.29.0; initial mcp 0.49.5 / panel 0.11.38; manually updated on disk to mcp 0.49.8 / panel 0.11.41.

Related (same root cause): #912, #890, #742

artokun:
self-update.ts:414,572 discard npm stderr ({ ok } only) — no EBUSY detection, generic 'staying on' note. #916 adds: panel update corroboration refusal, and a cancelled restart disconnecting the panel bridge with no survival pin.

closes #912
closes #916


artokun:

PR #925, opened 2026-08-06
restart fences the endpoint, not the instance + cannot identify local process without start times

### LINKED ISSUE #871: Restart fences the endpoint, not the instance: a same-URL replacement is invisible
Found by an independent gate on #850 while reviewing the fix for the restart target fences. Not introduced there — it predates the branch — and deliberately left unfixed rather than papered over.

## What is fenced today

`restartViaManagerReboot` guards two different questions with two different tests:

- **Before dispatch** — "is the server I am about to reboot the one whose argv I just read?" Fenced by comparing the **base URL**. (Comparing the target *generation* here was wrong: `setComfyuiTarget` bumps it on every successful call, including a same-URL reaffirmation that moves nothing, so it falsely refused restarts that were never in danger.)
- **The argv comparison** — "do these two readings describe the same server?" Fenced by the **generation**, which correctly catches an A→B→A round trip that leaves the base equal.

Both are right for what they test.

## The gap

Neither detects a server being **replaced at the same URL**. If another agent (or a supervisor, or the user) restarts ComfyUI on the configured target while we are awaiting `readServingArgv()`:

- the base is unchanged — so the pre-dispatch fence passes, and the reboot reaches the **successor**, not the instance whose argv we read;
- the generation is unchanged — it only bumps on *retarget* — so the argv comparison **proceeds**, and can narrate the difference between two distinct instances as one server changing its launch arguments.

The second is the more misleading half: it is a "could not determine X" reported as a determination, which is the fold this cluster exists to remove. Two concurrent agents on one rig is the ordinary way it happens, and that is in scope.

## Why it is not fixed in #850

Closing it needs something that does not exist yet: a **per-instance identity** that survives across our awaits — a boot id, a process start stamp, or a value from `/system_stats` that changes when the server is replaced. Without it there is no observation to fence on, and any check we invent would either miss the case or reintroduce the false refusal the base comparison was written to avoid.

## Suggested direction

Look for a stable per-instance token in `/system_stats` (or add one to what we already read), capture it beside the base and the generation, and re-check it before the dispatch and before the argv comparison. If it cannot be read, that is an **inconclusive** — the comparison should decline rather than proceed, and the dispatch should still go ahead (rebooting whatever serves the configured endpoint is what the caller asked for).

Refs: #850, and the same-instance reasoning in `src/services/process-control.ts` around the pre-dispatch fence.

### LINKED ISSUE #914: panel_restart_comfyui cannot identify local ComfyUI process after verified panel update
`install_panel(action="update")` successfully updated the panel from 0.11.36 to verified 0.11.41 and required a ComfyUI restart. `panel_restart_comfyui` then refused because the server on port 8189 could not be tied to its reported PID: macOS did not expose the process creation time.

### Steps to reproduce
1. Update the local panel through `install_panel(action="update")`.
2. Call `panel_restart_comfyui` while ComfyUI is idle.

### Exact error (scrubbed)
`Refusing to restart ComfyUI: ... PID 92755 could not be read ... host does not expose process start times ...`

### Fix
Applied locally: no. Workaround: restart ComfyUI from the launcher that owns it, then reconnect.

### Environment
macOS; ComfyUI 0.30.0; Apple MPS 48 GB unified memory; Python 3.13; torch 2.12.1; comfyui-mcp 0.49.8; panel updated from 0.11.36 to 0.11.41.

artokun:
restartViaManagerReboot (process-control.ts:3168) compares URLs, so a same-URL replacement is invisible (the code comment says per-instance identity 'the codebase does not have yet; tracked in #871'). Plus observeOwner refuses when process start times are unreadable ('this host does not expose process start times') with no alternative identification path.

closes #871
closes #914


artokun:

PR #926, opened 2026-08-06
panel run/sync truthfulness: duplicate run after reconnect + false 'panel install did NOT land' warning

### LINKED ISSUE #862: Avoid duplicate panel_run after reconnect when the original render is still running
After a panel reconnect, the active render continued on ComfyUI but the resumed agent had no completion event or visible running-state signal. Calling panel_run queued a duplicate first and only then returned a QUEUE WARNING that an identical job was already running. The warning arrives too late to prevent duplicate work.

### Steps to reproduce
1. Start panel_run for a long local render.
2. Let the panel/orchestrator reconnect while ComfyUI keeps executing.
3. Resume the task and call panel_run again.
4. panel_run queues a second job, then says another render is already running.
5. Agent must call get_queue and cancel_queued_job to remove the duplicate.

### Exact behavior
running prompt: b319f9dd-987e-4a50-a386-8773dbeebdd4
duplicate pending prompt: f1e87e12-7616-4a95-831d-f5aab413a83d
The pending duplicate was removed without interrupting the running render.

### Fix
applied locally: no
upstream-only: yes
Before queuing, panel_run should inspect current queue state after reconnect. If a job from the same live workflow/output branch is already running, return that running prompt_id and do not enqueue by default. Alternatively add an explicit allow_duplicate flag defaulting false.

### Environment
Windows 11; ComfyUI 0.29.2; RTX 5090 D 32GB; comfyui-mcp 0.48.23; panel 0.11.20.

Related (same root cause): #450, #468

### LINKED ISSUE #888: Reconnect warning falsely claims panel install missing although panel is installed and compatible
After reconnect, the sidebar injected a warning saying automatic panel sync failed and the pack was not present in custom_nodes, attributing it to stale ComfyUI-Manager 3.x. An immediate authoritative install_panel(action:"status") showed the panel is installed at the expected custom_nodes path, version 0.11.38, with reliable scan and no shadows. install_panel(action:"sync") then correctly returned up-to-date.

The reconnect/install diagnostic appears to retain or emit a stale install-failure assessment instead of re-reading the current on-disk panel state before warning.

### Steps to reproduce
1. Run with legacy ComfyUI-Manager 3.x.
2. Reconnect/resume after a dropped session or restart.
3. Observe injected "Panel install did NOT land" warning.
4. Call install_panel(action:"status") and observe installed=true.
5. Call install_panel(action:"sync") and observe decision=up-to-date.

### Exact error (scrubbed)
"Could not automatically sync the ComfyUI-MCP panel; no update was claimed... Panel install did NOT land: the pack is not present in custom_nodes..."

### Fix
Before emitting the reconnect warning, re-run the same authoritative disk scan used by install_panel(status), and suppress/replace stale installation failure text when installed=true and installedVersion meets requiredPanelVersion.

### Environment
Windows 11; ComfyUI 0.28.2; RTX 4080 SUPER 16 GB; comfyui-mcp 0.49.4; panel 0.11.38; ComfyUI-Manager 3.x legacy.

Related (same root cause): artokun/comfyui-mcp-panel#436

artokun:
panel_run snapshots the queue then queues unconditionally — the already-running warning is appended AFTER the duplicate is enqueued (panel-tools.ts:~6471). And the hello auto-sync catch (orchestrator/index.ts:3036-3058) pushes the raw 'did NOT land' warning with no disk re-scan/suppression when the panel IS installed and compatible.

closes #862
closes #888


artokun:

PR #927, opened 2026-08-06
VRAM handoff during renders skips the llama.cpp backend

### LINKED ISSUE #874: VRAM handoff during renders skips the llama.cpp backend — agent contends with ComfyUI on single-GPU setups
Reported by **jnbackup** in the Discord `#help` "Docker Compose" thread, testing a one-GPU setup via llama-swap's upstream feature:

> i think the mcp probably pinged comfyui in a way that left the original tool request waiting for a response that never arrived.

## The gap

The orchestrator has a VRAM handoff: when a render starts it unloads the local model and **holds** the user's chat until the render finishes, then warms the model back (`COMFYUI_MCP_OLLAMA_PAUSE_ON_GEN`, default on). The user gets "⏸ A render is running, so I've kept the GPU free for it and queued your message."

The gate on that behavior covers **only Ollama and LM Studio**, in both places it's evaluated:

- `src/orchestrator/index.ts:2261-2268` — `anyLocalOllama()` / `anyLocalLmstudio()`
- `src/orchestrator/index.ts:4463-4465` — `tabIsLocalVram` is `(ollamaApi === "ollama" && backendForTab(...) === "ollama") || (backendForTab(...) === "lmstudio" && isLocalLmstudio(...))`

The llama.cpp backend is a first-class provider (`COMFYUI_MCP_LLAMACPP_HOST`, default `http://127.0.0.1:8080/v1`, plus `COMFYUI_MCP_LLAMACPP_MODEL`) and is not in either check. So a local llama-server never yields: the agent keeps issuing completions while ComfyUI is mid-render, both processes hold VRAM on the same card, and the in-flight tool request can stall out — matching the report.

## Expected

`tabIsLocalVram` and the `anyLocal*` pair should include a local llama.cpp host on the same footing, with the same locality test the LM Studio path uses (`isLocalLmstudio`) so a **remote** `COMFYUI_MCP_LLAMACPP_HOST` is correctly treated as someone else's VRAM and left alone.

Two things worth deciding while in there:

1. The env var is named `COMFYUI_MCP_OLLAMA_PAUSE_ON_GEN` but already governs LM Studio too. Extending it to a third backend makes the name actively misleading — worth a neutral alias (`COMFYUI_MCP_PAUSE_LOCAL_ON_GEN`) with the old name still honored.
2. llama-swap adds a wrinkle Ollama doesn't: it swaps models upstream on demand, so "unload to free VRAM" may need to be a no-op-plus-hold rather than an explicit unload call. The hold is the part that matters for contention.

## Environment

Docker Compose, llama-swap upstream, single GPU, Open WebUI frontend.

artokun:
tabIsLocalVram (orchestrator/index.ts:4926-4929) only covers ollama+lmstudio; llamacpp is absent from the per-tab gate and the anyLocal* pair (:2563-2569) — agent contends with ComfyUI on single-GPU setups.

closes #874


artokun:

PR #928, opened 2026-08-06
mobile tunnel link dies on every orchestrator restart

### LINKED ISSUE #875: Mobile tunnel link dies on every orchestrator restart — quick-tunnel hostname and pair token both rotate per run
Reported by **seanmcmagic** in Discord `#help`:

> Is there a way for us to pin a version of the orchestrator while on mobile tunnel via the mobile app? Updating the npm version bricks my communication with the agent

He asked for a version pin, but the version isn't the cause — the *self-restart* is, and it breaks the phone link through two independent rotating values.

## Two rotations, both per-run

**1. Pair token.** `src/orchestrator/index.ts:908-913` — `COMFYUI_MCP_PAIR_TOKEN` if set, else `randomBytes(24)` per session. Unpinned, the phone's saved URL is dead after any restart.

**2. Tunnel hostname.** The default backend is a cloudflared **quick** tunnel → `wss://<random>.trycloudflare.com` (`src/services/secure-bridge.ts:8-9`). New random hostname every run. The code already acknowledges this at `index.ts:966-967`: *"Tunnel pairing stays on-demand: cloudflared quick-tunnel hostnames rotate per run, so a stable token alone can't make a tunnel URL persist."*

Meanwhile the self-restarter is **on by default** and re-checks npm hourly, restarting into new code whenever the agent is idle (`index.ts:5301-5309`). LAN users are fine — pinning `COMFYUI_MCP_PAIR_TOKEN` auto-starts the pairing listener at boot and the phone reconnects silently. **Tunnel users get no equivalent**, so a background auto-update silently invalidates the phone's URL.

## Why the current workarounds aren't good enough

The only ways to keep a tunnel session alive today are:

- Deploy your own `comfyui-mcp-relay` Worker and set `COMFYUI_MCP_TUNNEL_BACKEND=relay` + `COMFYUI_MCP_RELAY_URL` — a stable URL, but standing up a Cloudflare Worker is a steep prerequisite for "my phone should still work tomorrow".
- Disable updates entirely: `COMFYUI_MCP_AUTO_UPDATE_DISABLE=1`, or `COMFYUI_MCP_AUTORESTART=0` to keep notifications without restarts.

Both amount to opting out of a default-on feature to avoid a default-on feature breaking you. A user on mobile tunnel currently has to choose between staying current and staying connected, and nothing surfaces that tradeoff at pair time.

## Suggestions

- At minimum, **warn at pair time** when a tunnel URL is issued without a stable backend: say plainly that this URL dies on restart, and that auto-update will cause restarts.
- Persist the pair token to the config dir by default (generate once, reuse) rather than per-process — removes rotation #1 for everyone with no configuration. The token is already treated as password-equivalent, so persisting it doesn't change the trust model.
- For rotation #2, consider suppressing self-**restart** (not the update check) while a tunnel pairing session is live, deferring it to the next disconnect. The restarter already gates on idle/queue state, so this is another gate rather than new machinery.
- Document the relay-backend path in the mobile pairing flow, since it's currently the only real fix and isn't discoverable from the app.

## Environment

Mobile app over cloudflared tunnel, published npm orchestrator (auto-update on).

artokun:
quick-tunnel hostname and pair token both rotate per run (secure-bridge.ts header documents fresh-random default; pair token randomBytes per session unless pinned, orchestrator/index.ts:1007-1012).

closes #875


artokun:

artokun:
Closing as empty: this branch changes ZERO files against main — it carries only a placeholder commit and the diagnosis in its description. The agent that opened it ran out of credits before writing any fix.

The diagnosis is preserved on #875, which stays OPEN. Reopening this PR would not recover any work; the fix still has to be written.

Issue #950, opened 2026-08-06
install_panel(action:'update') errors as no-op failure even when disk is already at origin/main HEAD (>= required version)


artokun:
### What happened / root cause
graph_add_node (panel_add_node) was refused with: "cannot be safely targeted to the active workflow ... this tab advertised NO panel version ... a graph WRITE needs panel 0.11.35+". The error instructed running install_panel(action:'update').

I checked the on-disk panel checkout at custom_nodes/comfyui-agent-panel: it is a clean git checkout already at origin/main HEAD (766e8f464a4c431bd37c75a020035acac425144a), and PANEL_VERSION in web/js/comfyui-mcp-panel.js reads 0.11.41 — already >= the 0.11.35 required minimum. So the disk install was never stale; the actual cause is that the user's already-open ComfyUI browser tab loaded its panel JS bundle before/without this version, i.e. a stale cached bundle, fixable only by a hard-refresh (Ctrl+Shift+R) of that tab.

I then called install_panel(action:'update') as the suggested self-heal, and it itself errored (did not report a clean 'already up to date'):

"Panel update did NOT apply: nothing changed on disk, and the Manager's queue counts are NOT the proven legacy-3.x empty-queue signature (need total 0 AND done 0 with nothing pending/in-progress; got total=0, done=2, pending=0, in_progress=0) - the git fallback does not fire on an incoherent or partial signature."

Reading dist/services/panel-installer.js (no source checkout present, npx-cached build only), the 'no-op' outcome branch (~line 1677) is INTENTIONALLY fail-closed: the comment explains an unchanged git-HEAD/version cannot distinguish 'already at the upstream tip' from 'the update silently no-op'd', so it always throws rather than ever claim success. This looks like deliberate, carefully-reasoned design (references #639/#641/#724/#766), not an obvious bug — so I did NOT attempt a patch here (didn't want to fight well-reasoned fail-closed logic without being sure of all the invariants it protects). Flagging for maintainer judgment rather than guessing.

### Steps to reproduce
1. Open a ComfyUI browser tab, leave it open/cached across a panel update.
2. From an MCP session, call a graph-write tool (e.g. panel_add_node) against that tab.
3. Get the 'cannot be safely targeted ... panel version' refusal.
4. Call install_panel(action:'update') as instructed -> it throws 'Panel update did NOT apply' even though the on-disk checkout is already current (confirmed via git log / PANEL_VERSION).

### Exact error (scrubbed)
See above — no secrets involved.

[...]

Issue #945, opened 2026-08-06
update_comfyui reports success on detached tag without updating core


artokun:
### What happened / root cause
update_comfyui returned updated:true while ComfyUI stayed on detached tag v0.30.2. The install's remote.origin.fetch only contained a tag refspec, so git pull printed "Already up to date" without moving to current origin/master. The tool treated that as a successful core update and reinstalled requirements for the stale checkout.

### Steps to reproduce
1. Use a Comfy Desktop install checked out at a detached release tag.
2. Configure remote.origin.fetch with only a tag refspec.
3. Run update_comfyui.
4. Observe updated:true and "Already up to date", while git rev remains the detached tag and origin/master is not fetched.

### Exact result
update_comfyui: {"updated":true}; git pull output: "Already up to date."
Before workaround: HEAD detached at v0.30.2 (dec5d945), remote master 2eb60976.

### Fix
Applied locally: yes. Upstream-only: no.
Workaround used:
- preserve the user's server.py change in a stash and backup branch;
- explicitly fetch +refs/heads/master:refs/remotes/origin/master;
- switch/fast-forward master;
- add the master fetch refspec and set master upstream;
- reinstall requirements and restart.
Afterward git pull --ff-only reports cleanly against master, git rev is 2eb60976, frontend is 1.48.6.

Precise tool change needed: before reporting success, detect detached HEAD and inspect remote fetch refspec/upstream. Fetch the intended branch explicitly, verify HEAD/revision changed or is already equal to the remote target, and return a warning/error instead of updated:true when git pull cannot update the checkout.

### Environment
Windows 11; RTX 3060 12 GB; ComfyUI Desktop local; Python 3.13; torch 2.10.0+cu130; comfyui-mcp 0.49.8; panel 0.11.38.

Issue #944, opened 2026-08-06
panel_run reports a FALSE "ComfyUI refused to queue" failure when ComfyUI actually accepted the prompt and only dropped one invalid output node ("Output will be ignored")


artokun:
### What happened / root cause

`panel_run` (no `to_node_id`, full graph) returned a hard failure:

```
Error: ComfyUI refused to queue the workflow
- SaveVideo (node 92): Exception when validating node ('43069')
```

But ComfyUI **accepted the prompt and ran it**. Its own log for the very same submission:

```
2026-08-06T19:17:46.493586 - [INFO] got prompt
2026-08-06T19:17:46.506913 - [ERROR] Failed to validate prompt for output 92:
2026-08-06T19:17:46.507002 - [ERROR] * (prompt):
2026-08-06T19:17:46.507026 - [ERROR]   - Exception when validating node: '43069'
2026-08-06T19:17:46.507051 - [ERROR] * SaveVideo 92:
2026-08-06T19:17:46.507070 - [ERROR]   - Exception when validating node: '43069'
2026-08-06T19:17:46.507088 - [ERROR] Output will be ignored
2026-08-06T19:17:46.558292 - [llama-cpp_vlm] Loading model...
   ... execution proceeds normally ...
```

`Output will be ignored` is ComfyUI's **partial-validation** path: when SOME output nodes fail validation but at least one valid output remains, ComfyUI drops the bad output(s), still returns a `prompt_id`, and executes the rest. `panel_run` treats the presence of `node_errors` / a per-output validation failure as a TOTAL rejection and reports "ComfyUI refused to queue the workflow", discarding the fact that the prompt was queued.

This is the inverse of the failure mode the tool description promises to avoid ("never a false 'queued'") — here it is a **false 'refused'**, which is just as damaging:

1. The agent believes nothing was queued and starts diagnosing/repairing a graph that is *already rendering*.
2. The agent asks the user to make edits (here: "mute node 92") to a graph with a **live render in flight** — edits that were both unnecessary (ComfyUI had already dropped that output itself) and potentially harmful.
3. The agent then issued two further `panel_run(to_node_id=150)` calls attempting to "retry", i.e. it actively tried to double-queue on top of a running job. Those happened to be refused by the #556 graph-changed guard; without that guard this would have queued a duplicate ~20-minute video render.
4. When the user later asked "did you queue this?" the agent had to answer "no" — because the tool said so — and both user and agent were left with a running job neither believed they started.

[...]

Issue #941, opened 2026-08-06
panel_show_media reports success for broken remote viewRef cards


artokun:
### What happened / root cause
`panel_show_media` accepts ComfyUI `viewRef` items and immediately reports them as painted, but it does not wait for or validate the browser image load. On a REMOTE ComfyUI whose headless `/view` endpoint returns HTML, the user sees a broken image card while the tool reports `ok:true, painted:8, unconfirmed:0`.

The output files do exist in the configured local ComfyUI workspace mirror. Supplying the same files as absolute `source.path` works because the orchestrator inlines verified bytes as data URLs. The broken path is specifically unverified `viewRef` forwarding.

### Steps to reproduce
1. Use a remote ComfyUI/reverse-proxy session where headless `/view` returns HTML.
2. Call `panel_show_media` with `{filename, subfolder, type:"output"}`.
3. Observe tool result `ok:true, painted:N, unconfirmed:0`.
4. Observe broken images in the actual panel chat.
5. Call `panel_show_media` with an existing absolute local mirror path; inline media renders instead.

### Exact error / evidence
Headless fetch:
`IMAGE_NOT_FOUND ... got content-type "text/html"`

Original panel result:
`{"ok":true,"count":8,"painted":8,"unconfirmed":0}`

User report:
`채팅창에 뜨는 이미지도 오류야 정상적으로안떠`

### Fix
Applied locally: yes
Upstream-only: no

Patched the packaged runtime's `dist/orchestrator/panel-tools.js`: before forwarding a ComfyUI viewRef, resolve the configured/saved workspace, securely constrain the candidate beneath its output/input/temp root, verify it is an allowed media file under 20 MB, and inline its bytes. Fall back to the existing browser/headless behavior when no verified local mirror exists.

Key change:
```js
const workspace = await getWorkspace();
const root = workspace.workspace_path
  ? resolve(workspace.workspace_path, mediaType)
  : undefined;
const p = root ? resolve(root, src.subfolder ?? "", src.filename) : undefined;
if (root && p && (p === root || p.startsWith(root + sep)) && existsSync(p)) {
  // validate type/size and send data URL
}
```

Verification: `node --check dist/orchestrator/panel-tools.js` passed; orchestrator reload succeeded; a post-reload viewRef dispatch succeeded. The installed release package lacks `tsc`, so `npm run build` could not run (`tsc: not found`). Browser-level rendering requires user confirmation because the tool currently acknowledges before the image load event—the defect being reported.

[...]

PR #956, opened 2026-08-07
cluster: tools that report SUCCESS for work that did not happen

artokun:
The house defect class, pointed at ourselves: a tool returns a success message for an action that never completed. Every issue here is the same shape — the *claim* is stronger than what the code established. Tracking issue: #796.

## Status — 6 of 7 shipped

| Issue | Sev | The false claim | Status |
|---|---|---|---|
| #826 | **P1** | `panel_request_secret` reports the token saved and applied; the comfyui MCP child is never respawned, so downloads keep 401ing | **fixed** (#837, access-time credential getter) |
| #945 | P2 | `update_comfyui` reports success on a detached tag without updating core | **fixed** (#972) |
| #950 | P2 | `install_panel(action:'update')` errors as a no-op failure even when the disk state changed | **already fixed** in 0.50.0 (#824 work) |
| #941 | P2 | `panel_show_media` reports success for broken remote viewRef cards | **fixed** (#982) |
| #889 | P2 | run-errored notification claims "the workflow run you just queued" for a run it cannot attribute | **fixed** (#991) |
| #944 | P2 | `panel_run` reports a FALSE "ComfyUI refused to queue" when ComfyUI accepted it | **fixed** (#971) |
| #933 | P2 | `panel_set_workflow_target({mode:'current'})` reports success but does NOT retarget | **open** — see below |

## The fix shape held up

The premise was that these share a *fix* and not just a symptom: separate **did it happen** from **can we describe it**. Refuse when the action has NOT happened and proceeding is unsafe; disclose when it HAS happened and only its description is uncertain.

That is what each fix turned out to need, and the third state — *could not determine* — carried more weight than expected. #971 in particular ended up with three outcomes rather than two, because a `prompt_id` arriving alongside a rejection is a genuine contradiction that must not be resolved in either direction.

They were fixed individually rather than in one PR. In hindsight that was right: each needed its own mutation-verification, and the shared reasoning transferred through the commit messages without needing shared code.

## What is left

**#933** (and its sibling #934) are the reconnect-wedge pair, and they are only partly ours. #984 fixed the tab-id rendering that made the rebind note read as a no-op. The remaining halves — `mode:'current'` reporting success with no tab connected, `panel_list_workflows` gated by the instance guard, and the guard not consulting connectivity first — are **panel-side**: `workflow instance mismatch` is emitted by `comfyui-mcp-panel.js` (lines 1859 and 13166), not by this repo. Tracked on #932/#933/#934 and in cluster #958.

**#796** remains as the standing tracking issue for the class itself, which is the right home for it — it outlives any one cluster.

Closing this draft: it did its job as a scoping document, and leaving it open would advertise work that is done.


artokun:
Four more members of this class, found by **auditing for it** rather than waiting for reports — all fixed and merged on the panel side tonight.

I swept every mutating panel handler asking your question here: *is the claim stronger than what the code established?*

| fix | the false claim |
|---|---|
| panel#739 | `panel_set_node_mode` returned the **requested** mode, never re-reading the node. Not cosmetic — bypass/mute decide whether a node EXECUTES, so an agent could be told it disabled a node that is still running |
| panel#740 | `panel_remove_group` reported `{removed}` without checking. Its fallback chain was `else if`, and LiteGraph's `remove` is for **nodes** — so on any build lacking `removeGroup` the splice was unreachable and a group still on the canvas read as removed |
| panel#741 | `panel_clear` reported the count it **attempted**, not what left. Worse than the group case: the caller believes the canvas is empty and wires new nodes alongside invisible leftovers |
| panel#742 | both subgraph-conversion paths reported `node_id: res?.node?.id ?? null` — a null sentinel **inside** a success payload, which reads as "created, field missing" rather than "nothing happened" |

**The discriminator turned out to be mechanical**, which is why the sweep went 4-for-4 and is worth reusing on the orchestrator side:

- handlers routing through `graph_edit_node` are HONEST — it builds its reply as `after: summarize(node)`, read from the live node;
- every handler that **hand-built a literal reply object** echoed the request.

So: grep for handlers constructing a reply literal rather than summarizing a live object. That is where this class lives.

Two tells that a report belongs here even when the reporter did not diagnose it: someone saying they verified **"by re-read, not by return value"** (panel#690 said exactly that about `set_node_mode`, which is what pointed me at it), and a workaround that switches tools rather than retrying.

I also audited three adjacent surfaces that came back **clean**, so they need not be re-swept: the orchestrator's 91 `panel_*` wrappers (45 inspect the reply, 36 return `ctx.call` directly which propagates errors correctly), absence-vs-falsiness guards, and stale tool references in panel prose (the vocabulary gate already covers it, and I verified the vendored artefact against mcp source rather than trusting it — 37/91/160, exact match).

Already-honest models worth copying when fixing the remaining members here: `graph_disconnect`'s `verifyDisconnect` (discloses whether the intended link went AND collateral node changes), `graph_move_group`'s `groupBoxIsAt` + `restoreBox`, and `graph_add_subgraph`'s landing check. Where one path has the guard and its sibling does not, that is a live defect — it caught two tonight.

artokun:
Scoping complete — 6 of 7 shipped, the remainder tracked on #933/#934 (mostly panel-side) and #796. Body updated with the outcome of each.

Issue #954, opened 2026-08-06
list_workflow_templates hides target and socket cause behind bare fetch failed


artokun:
### What happened / root cause
`list_workflow_templates` returned only `fetch failed` when its configured headless ComfyUI target (`http://127.0.0.1:8188`) was unreachable, even though the panel bridge itself was connected to a live canvas. The tool catches Node fetch errors through `errorToToolResult`, which keeps only `err.message` and discards `err.cause`, so the target URL and socket cause are lost.

### Steps to reproduce
1. Connect the sidebar panel to a live canvas while the headless MCP `COMFYUI_URL` still points to an unreachable local address.
2. Call `list_workflow_templates`.
3. Observe only `fetch failed`; compare `get_environment`, which reports the unreachable target explicitly.

### Exact error (scrubbed)
`fetch failed`

### Fix
Applied locally: yes. Upstream-only: no.

Moved `const base = getComfyUIBaseUrl()` outside the try and returned a targeted error containing the full `/api/workflow_templates` URL, the primary error, and `err.cause` when available. Syntax verification: `node --check dist/tools/skills-access.js` exited 0. The patch takes effect on the next MCP process respawn.

```diff
+ const base = getComfyUIBaseUrl();
  try {
-   const base = getComfyUIBaseUrl();
    const res = await fetch(`${base}/api/workflow_templates`, ...);
  } catch (err) {
-   return errorToToolResult(err);
+   const primary = err instanceof Error ? err.message : String(err);
+   const cause = err instanceof Error && err.cause ? `; cause: ${err.cause instanceof Error ? err.cause.message : String(err.cause)}` : "";
+   return { isError: true, content: [{ type: "text", text: `Failed to fetch ${base}/api/workflow_templates: ${primary}${cause}` }] };
  }
```

### Environment
Linux; ComfyUI 0.29.0; NVIDIA A100-PCIE-40GB; sidebar orchestrator comfyui-mcp 0.49.8; headless package inspected/patched 0.49.5; panel 0.11.41.

Issue #953, opened 2026-08-06
list_output_images: silent /history fallback on REMOTE omits VHS video outputs — reports populated output dir as empty


artokun:
### What happened / root cause

`list_output_images` returns a confident, unqualified listing on a **REMOTE** ComfyUI even though that listing is derived from `/history` — which systematically does **not** register `VHS_VideoCombine`-style video outputs. The result text carries no indication of this.

`dist/services/image-management.js`:

```js
if (isRemoteMode() || !config.comfyuiPath) {
    return listOutputImagesFromHistory(limit, pattern);
}
```

The code *knows* the caveat — the doc comment right above it literally says *"ComfyUI's /history often does NOT register VHS-style video outputs — so a filesystem scan is the reliable way to confirm a finished video render."* But that warning never reaches the caller. The tool description leads with "USE THIS TO CONFIRM A VIDEO RENDER … the local filesystem scan is the reliable way to verify the .mp4 exists", and only later mentions the remote fallback. On a remote target the tool's headline guarantee silently **inverts**.

Real-world impact: on a remote pod with 30+ ProRes masters under `output/slot/` written by `VHS_VideoCombine`, the tool reported 12 files, none of them under `slot/`. I concluded the directory was empty and **wrote that as a fact into a project handover/contract document**. The user caught it. The files were there the whole time — I confirmed by pulling one through `/view` with `stage_output_as_input`.

The empty-result branch is the most dangerous: `"No output media (images or videos) found."` reads as authoritative absence, when on remote it means only "nothing in /history".

Secondary (minor): the tool silently accepts and ignores params not in its schema. I passed `subfolder: "slot"` (not a real param) and got no error — so I believed I had filtered when I had not, which compounded the wrong conclusion.

### Steps to reproduce

1. Point comfyui-mcp at a REMOTE ComfyUI (`isRemoteMode()` true).
2. Render anything through `VHS_VideoCombine` with `save_output: true` and a `filename_prefix` containing a subfolder (e.g. `slot/clip`).
3. Call `list_output_images`.
4. The file exists in `output/slot/` but is absent from the result, with no warning. `stage_output_as_input` on the same filename succeeds, proving it is on disk.

### Exact error (scrubbed)

No error — silent wrong output. Observed:

```
Found 12 media file(s) (12 video):
1. **final/BASENAME_00018.mov** [video] _(subfolder: final)_
...
```

[...]

Issue #952, opened 2026-08-06
Readonly ComfyUI tools return fetch failed while panel bridge remains connected


artokun:
### What happened / root cause
During a live-panel workflow repair, panel_graph_outline, panel_get_errors, panel_query_graph, and panel_strip_workflow all reached the active ComfyUI canvas successfully. In the same session, get_node_info returned only `fetch failed`; validate_workflow reported `Cannot connect to ComfyUI to validate: fetch failed`; check_workflow_runtime returned `runtime: unknown` with the same fetch failure. The readonly/headless tool channel appears not to share the connected target used by the panel bridge.

### Steps to reproduce
1. Connect the sidebar panel to a live workflow.
2. Confirm panel_graph_outline succeeds.
3. Call get_node_info for an installed node type, then validate_workflow and check_workflow_runtime on a graph stripped from that same live canvas.
4. Observe fetch failures despite successful panel access.

### Exact error (scrubbed)
`get_node_info`: `fetch failed`
`validate_workflow`: `Cannot connect to ComfyUI to validate: fetch failed`
`check_workflow_runtime`: `Could not reach the ComfyUI server to classify nodes (fetch failed)`

### Fix
Applied locally: no. Upstream-only: yes. Bind comfyui_readonly tools to the same resolved ComfyUI target/session used by mcp__panel tools, or expose an explicit target handoff. When unavailable, include the resolved endpoint/channel in the diagnostic so target drift can be identified without guessing.

### Environment
OS: Linux; ComfyUI 0.29.0 LOCAL; GPU NVIDIA A100-PCIE-40GB; comfyui-mcp 0.49.8; panel 0.11.41; backend Codex.

artokun:
Traced this, and the shape is worth stating precisely because the obvious fix carries a decision rather than just code.

**The retarget machinery already exists and is complete.** `applyComfyuiTarget()` validates the URL, calls `setComfyuiTarget()`, and — the part that matters for your symptom — resets the in-process client:

> *"Retarget the orchestrator's OWN in-process ComfyUI client too — the direct call_tool path the mobile app uses (get_workflow, get_image, …) goes through getClient(), which caches against the process-start host. Without this, a retargeted orchestrator keeps that client pinned to the old ComfyUI."*

So once the orchestrator is told where ComfyUI is, the readonly tools follow. `runpod` drives exactly this path when a pod comes up.

[...]

Issue #918, opened 2026-08-06
list_local_models returns "No local models found" when ComfyUI is unreachable/still starting (indistinguishable from an empty install)


artokun:
### What happened / root cause

Against a REMOTE ComfyUI that was still warming up, `list_local_models` returned `No local models found.` — identical to the answer for a genuinely empty install. I reported to the user that their headless ComfyUI URL was misconfigured. It wasn't: a retest a few minutes later returned the full, correct model list. The tool's output caused a confidently wrong diagnosis.

In `dist/services/model-resolver.js` `listLocalModels()`:

1. Each `/models/<dir>` fetch is wrapped in a try/catch that only `logger.debug(...)`s and continues. A non-ok status and a non-array body also `continue` silently. So when the server is down/warming, every category fails silently.
2. `httpReturnedAny` stays `false`, so it falls through to the filesystem fallback.
3. In remote/cloud mode `config.comfyuiPath` is undefined, so it hits `if (!config.comfyuiPath) return results;` and returns `[]`.
4. The tool renders that as `No local models found.`

Nothing anywhere distinguishes "ComfyUI answered and has zero models" from "ComfyUI never answered". The existing comment even calls the empty return "the correct answer for remote/cloud setups" — but it is only correct when the server actually responded.

Related, same class, NOT patched here: `get_node_info` surfaces a bare `Unexpected token '<', "<!doctype "... is not valid JSON` when `/object_info` returns an HTML proxy/login/startup page. That at least fails loudly, but it should say "ComfyUI returned HTML instead of JSON — server may still be starting" rather than leaking a JSON.parse message.

### Steps to reproduce

1. Point comfyui-mcp at a REMOTE ComfyUI (no `COMFYUI_PATH` set, so no filesystem fallback).
2. Stop that ComfyUI, or call during startup before `/models/*` serves JSON.
3. Call `list_local_models`.
4. Observe `No local models found.` — no indication the server was unreachable.
5. Start ComfyUI fully, call again → full correct listing.

### Exact error (scrubbed)

No error surfaced — that is the bug. Output was:

```
No local models found.
```

Sibling tool during the same window:

```
get_node_info -> Unexpected token '<', "<!doctype "... is not valid JSON
```

### Fix

applied locally: yes   upstream-only: no

[...]

PR #957, opened 2026-08-07
cluster: 'could not reach it' reported as 'there is nothing there'

artokun:
> **Empty placeholder** — one commit, zero files changed. Recovered and written up rather than left dangling, because the theme it named turned out to be the most productive line of work in the repo this week.

The defect: a tool that **could not reach** something reports that the thing **is not there**. Absence of evidence rendered as evidence of absence, in a sentence the reader has no way to distinguish from a real finding.

## What shipped under this theme

| Issue | The false absence | Fixed by |
|---|---|---|
| #918 | `list_local_models` printed "No local models found" when ComfyUI never answered — a reporter read it as a bad URL and told the user so | #967 |
| #962 | "No diffusion_models models found" for weights the server was actively loading — the category list was hardcoded, so they were invisible by construction | #995 |
| #953 | `list_output_images` returned a confident list that omits every VHS video by construction; a reporter wrote "the directory is empty" into a handover doc | #975 |
| #954 | `list_workflow_templates` answered `fetch failed` with no target and no cause | #968 |
| #952 | the readonly tools' `fetch failed`, read as "these tools are broken" while the panel worked | #968 (diagnosis half) |
| #449 | `restart_comfyui` reported "no process found" for a process that was answering HTTP | already fixed in 0.48.12 |
| #717 | download tray froze on stale percentages after an orchestrator restart | already fixed in 0.49.1 |

## The shape that generalised

Three states, not two. `answered and empty` is a fact; `could not ask` is not; and the gap between them is where every one of these lived. The fixes that held all did the same thing:

- **say which categories/targets actually answered**, so emptiness is scoped to what was checked (#967's coverage record);
- **discover rather than assume** — `/models` is the server's own answer to "what do I serve?", and a hardcoded list of 15 folder names cannot find weights registered under a sixteenth (#995);
- **name the target and the cause**, because `fetch failed` is four characters that say nothing — not the host, not the port, not DNS-vs-refused-vs-TLS (#968).

And the wording rule that came out of it: *unverified* is not *wrong*. A reader told "failed" fixes something that may not be broken; a reader told "unverified" re-checks.

## What is left

**#952's design half** — whether the readonly tools should *bind* to the panel's connected server rather than `COMFYUI_URL`. Those two addresses are allowed to differ, and merging them has real consequences for remote setups. #968 fixed the diagnosis; the binding question is open and deserves its own issue with the trade-off written down.

Closing this placeholder: everything it scoped is either shipped or tracked on its own issue, and an empty draft PR advertising a live cluster is its own small version of the defect it names.


artokun:
**#828 removed from this cluster — already fixed and shipped.**

It was fixed by #837 (`d9e8971`) in v0.50.0/v0.50.1 via `src/comfyui/json-guard.ts`, but sat open because that PR used an owner-qualified `Closes` reference, which GitHub does not auto-close for same-repo issues. Now closed.

This matters for the cluster's framing: #828 was listed here as the likely **upstream cause** of several bare `fetch failed` reports (#952, #954). Since the HTML-not-JSON path is already guarded, those remaining reports are **not** explained by it — they need their own diagnosis rather than being assumed downstream of #828.

Remaining in this cluster: #962, #918, #449, #947, #952, #953, #954.

artokun:
Empty placeholder (1 commit, 0 files) from the stranded wave. Body written up with what shipped under this theme — six issues fixed, one design question left on #952 — and closed.

Issue #934, opened 2026-08-06
Reconnect recovery loop: panel_set_workflow_target({mode:"current"}) reports success with no tab connected; tab ids truncated to "wf:workf"; panel_list_workflows gated by the instance guard


artokun:
### What happened / root cause

After a ComfyUI restart/reconnect (the agent session resumed with a "the user switched the open workflow on the canvas — it is now 'Untitled 2026-08-06 03-58-41'" notice), **every** `panel_*` call failed, and the documented recovery path reported success while doing nothing. It took 6 tool calls to reach the one error message that actually named the cause (a dropped panel socket).

Four distinct defects:

**(1) `panel_set_workflow_target({mode:"current"})` returns a success payload when no panel tab is connected.**
It returned `{mode:"current", note:"Following the user's current workflow tab. Rebound this session from tab wf:workf onto the active tab wf:workf."}` — and on a second call just `"Following the user's current workflow tab."` — with **no** `deferred:true` and **no** "No panel tab is connected yet" note, even though `panel_reload` immediately afterwards said `Panel not reachable: no panel connected`. `dist/orchestrator/panel-tools.js` (~L5440-5515) *has* a correct `noTabsConnected` → `deferredBind` branch that emits exactly the right guidance, but this session never took it: `awaitReachable` / `rebindToActiveTab` apparently resolved without throwing against a tab that was not actually connected, so `deferredBind` stayed false and the tool reported a clean rebind. Because every other error message instructs the agent to "re-target with `panel_set_workflow_target({mode:'current'})`", a silently-successful no-op sends the agent in circles.

**(2) Tab ids are truncated to 8 chars, which renders every workflow-keyed tab id identically as `wf:workf`.**
`dist/orchestrator/panel-tools.js:5480`:
```js
rebindNote = ` Rebound this session from tab ${before.slice(0, 8)} onto the active tab ${ctx.tabId.slice(0, 8)}.`;
```
[...]

Issue #932, opened 2026-08-06
panel_* tools permanently wedged with "workflow instance mismatch" — documented recovery paths do not clear the guard


artokun:
### What happened / root cause

After a connection drop + ComfyUI reconnect (session resumed automatically) AND a workflow tab switch (panel injected the "The user switched the open workflow on the canvas — it is now 'Untitled 2026-08-06 03-58-41'" notice), **every** `panel_*` call started failing with:

```
workflow instance mismatch: this command targets a different workflow than the active canvas — the workflow was switched or replaced after it was issued. Re-select the intended workflow (panel_open_workflow) or re-target with panel_set_workflow_target({mode:"current"}), then retry.
```

The guard is stuck: both remedies the error message itself recommends **succeed**, yet the very next call still fails with the same mismatch. Notably the guard also blocks **read-only, workflow-agnostic** calls — `panel_list_workflows` (which just lists tabs and shouldn't be instance-scoped at all) fails too, so there is no way to even enumerate tabs to recover.

Likely root cause: the orchestrator's cached workflow-instance token (routing key / workflow_uuid / open_seq) is not refreshed by `panel_set_workflow_target({mode:'current'})` or by a successful `panel_open_workflow`, so the comparison against the live canvas keeps failing. `panel_open_workflow` returned a fresh `workflow_uuid` and `open_seq: 2`, but the following `panel_graph_outline` still mismatched — i.e. the newly-returned instance identity is not adopted as the session's target.

### Steps to reproduce

1. Have a panel session with an open workflow.
2. Drop/restore the ComfyUI connection (e.g. a restart by another agent) so the session resumes.
3. Switch the active workflow tab on the canvas.
4. Call any `panel_*` tool → "workflow instance mismatch".
5. Call `panel_set_workflow_target({mode:"current"})` → succeeds, returns `{mode:'current', note:"Following the user's current workflow tab."}`.
6. Call `panel_graph_outline` → **still** "workflow instance mismatch".
7. Call `panel_open_workflow({path:"Untitled 2026-08-06 03-58-41"})` → succeeds, returns `{path:'workflows/Untitled 2026-08-06 03-58-41.json', workflow_uuid:'<uuid>', modified:true, open_seq:2}`.
8. Call `panel_graph_outline` → **still** "workflow instance mismatch". Same for `panel_list_workflows`.

Wedged permanently; the only remaining escape is `panel_reload`.

### Exact error (scrubbed)

```
[...]

PR #958, opened 2026-08-07
cluster: panel wedged after reconnect — instance mismatch, stale trays, lost tunnels

### LINKED ISSUE #717: Download tray shows stale/frozen % forever after orchestrator restart (lastDownloadSnapshot init bug)
The panel's download tray (DOWNLOADS section) showed stale, frozen progress percentages (e.g. 56%, 89%) for two model downloads (hidream_o1_image_fp8_scaled.safetensors, llama_3.1_8b_instruct_fp8_scaled.safetensors) for 10-13+ minutes with zero underlying file growth. Investigation confirmed both files were actually FULLY DOWNLOADED and byte-identical to their HuggingFace source sizes, with valid, parseable safetensors headers — the downloads had succeeded on disk; only the tray UI was wrong.

Root cause (dist/orchestrator/index.js, pollDownloads):
1. Each orchestrator process computes its own progressDir from `join(tmpdir(), \`comfyui-mcp-progress-\${bridgePort}-\${progressNonce}\`)` — unique per process. A NEW orchestrator process (e.g. after the panel reconnects to a fresh session, or after the process that started the download exits/is replaced) has no visibility into the OLD process's progress temp files, so it can never see, update, or prune those rows itself.
2. `lastDownloadSnapshot` is initialized to the literal string `"[]"`.
3. On its first poll, the new (empty) progressDir yields `downloads = []`, and `JSON.stringify([]) === "[]"`, which is EQUAL to the initial `lastDownloadSnapshot`. The guard `if (snapshot !== lastDownloadSnapshot)` is therefore false, so `bridge.push({ type: "download_progress", downloads: [] })` is NEVER sent.
4. Meanwhile the panel's front-end (comfyui-mcp-panel.js `renderDownloads`) only updates `downloadItems` when it receives a `download_progress` push — it has no independent staleness/reconciliation check. So it keeps rendering whatever the OLD, now-dead orchestrator process last pushed (e.g. 56%/89%), forever, since the new process never sends a corrective (even empty) push.

Net effect: any time the orchestrator process restarts/reconnects while a download row is mid-flight (not yet "done") in the panel's last-known state, that row is stuck showing its last percentage indefinitely, even though the download may have completed successfully in the background under the old process (or via a different, e.g. Ollama-driven, session).

### Steps to reproduce
1. Start a model download via `download_model` (or any path that writes rows to the orchestrator's progress temp dir) and let it reach some %, e.g. 50%, in the panel's DOWNLOADS tray.
2. Let it continue downloading to completion, OR end/restart the orchestrator process (e.g. new Claude Code session, `panel_reload`, or the process that started the download exiting) BEFORE the panel receives the terminal "done" push.
3. Reconnect the panel to the new orchestrator process.
4. Observe: the DOWNLOADS tray keeps showing the stale, frozen percentage from the old process indefinitely, even though the file on disk may already be 100% complete.

### Exact error (scrubbed)
No thrown error — silent stale-state bug (missing corrective push on first poll of a fresh orchestrator process).

### Fix

```diff
--- a/src/orchestrator/index.ts (dist/orchestrator/index.js)
+++ b/src/orchestrator/index.ts (dist/orchestrator/index.js)
@@
-    let lastDownloadSnapshot = "[]";
+    // null (not "[]") so the FIRST poll of a fresh orchestrator process always
+    // pushes its snapshot — even an empty one — correcting any stale download
+    // tray a reconnecting panel is still showing from a PRIOR orchestrator
+    // process (each process gets its own progressDir, so it can't see or prune
+    // that process's rows itself; without this the equality check silently
+    // skipped the corrective push whenever the fresh snapshot also computed to
+    // "[]", leaving completed downloads frozen at their last-seen percentage).
+    let lastDownloadSnapshot = null;
```

Verified: `node --check` passes on the patched dist file; applied via `panel_reload` (orchestrator scope) and confirmed the orchestrator resumed normally.

Suggested follow-up (not applied, for upstream consideration): the panel front-end could also independently clear/mark-stale any download row that hasn't received an update within some bounded time after a bridge reconnect, as defense-in-depth against any other path that could skip a corrective push.

### Environment
OS: Windows 11 Pro 10.0.26200
ComfyUI: 0.29.0 (local)
GPU: NVIDIA GeForce RTX 5090, 32 GB VRAM
comfyui-mcp: 0.49.0
panel: 3.x (legacy)

artokun:
A reconnect or restart leaves the panel in a state it cannot recover from on its own, and the recovery path itself is missing or broken.

## Issues in this cluster

| Issue | Sev | Symptom |
|---|---|---|
| #932 | P2 | `panel_*` tools permanently wedged with "workflow instance mismatch" |
| #934 | P2 | Reconnect recovery loop — `panel_set_workflow_target({mode:"current"})` never breaks the cycle |
| #717 | P2 | Download tray shows a stale/frozen % forever after an orchestrator restart |
| #858 | P2 | Stale download survives reconnect but cannot be cancelled or safely restarted |
| #875 | P2 | Mobile tunnel link dies on every orchestrator restart (hostname AND pair token both rotate) |
| #823 | P2 | Recovery tools missing from Codex panel sessions after a panel version-floor error |

## Why one PR

All six are "state that must survive an orchestrator restart, and doesn't". Two likely duplicate pairs:

- **#932 / #934** — plausibly one bug seen from both ends: the wedge, and the failed escape from it.
- **#717 / #858** — the same orphaned download seen twice: #717 is the frozen percentage the user *sees*, #858 is that the same download can't be cancelled or restarted. A fix that only repairs the tray's display would close #717 and leave the user stuck — the download record itself is the broken thing.

## Known-failed prior attempts

- **#823**: PR #832 already shipped a fix for the *message* (per-session recovery description). The tools are still absent from the surface. A message-level fix has been tried and did not resolve this — do not repeat it.
- **#875**: PR #928 was opened for this and changed **zero files** (closed as empty). Diagnosis preserved on the issue: the quick-tunnel hostname and the pair token rotate **independently**, so pinning only one still breaks the saved link.
- **#932 / #934**: sessions are ORCHESTRATOR-scoped (one per backend), never per-workflow or per-tab. Any fix that re-introduces workflow-scoped session gating regresses #570. `tab_id` is routing only.

**Draft** — issues attached for scoping; no fix yet. Updated: attached #858.


artokun:
**#858 removed from this cluster — already fixed and shipped** (PR #920, v0.50.0/v0.50.1; open only because of an owner-qualified `Closes`).

This changes the cluster's framing: #858 was paired here with #717 as "the same orphaned download seen twice" — the frozen tray percentage and the uncancellable download. The **cancel/restart half is already fixed**, so #717 should be re-checked against 0.50.1 before any work starts. It may be fixed too, or it may now be purely a tray-rendering bug rather than a download-record bug — which is a different fix in a different place.

Remaining: #932, #934, #717, #875, #823.

artokun:
Closing this scoping draft (1 commit, 0 files). The cluster's framing — *"state that must survive an orchestrator restart, and doesn't"* — held up; recording where each item landed.

**Shipped:**
- **#717** (frozen download tray) — already fixed in **0.49.1**. The `lastDownloadSnapshot = "[]"` sentinel the reporter identified was replaced with a `null`-initialised `DownloadProgressSnapshots`, so a fresh process's first poll now sends the corrective empty push. Reported against 0.49.0, one release earlier. Regression test carries the issue number.
- **#858** (stale download uncancellable) — closed earlier.
- **#934's tab-id half** — `.slice(0, 8)` rendered EVERY `wf:workflows/…` tab as the literal `wf:workf`, so the rebind note read *"Rebound this session from tab wf:workf onto the active tab wf:workf"* for a rebind between two different workflows. Fixed in #984.

**Still open, and mostly panel-side:**
- **#932 / #934 / #933** — the guess that these are one bug from several angles was right in substance. `workflow instance mismatch` is emitted by `comfyui-mcp-panel.js` (lines 1859 and 13166), not by this repo, so three of the four defects #934 enumerates need a panel change: `panel_list_workflows` being gated by the instance guard, the guard not consulting connectivity first, and `mode:'current'` reporting success with no tab connected.

  On that last one — the `deferredBind` branch exists and is correct, so the open question is why `awaitReachable`/`rebindToActiveTab` resolved against a tab that was not connected. I could not establish that from the report alone and did not guess.
- **#875** (mobile tunnel dies on every restart) — untouched.
- **#823** (recovery tools missing from Codex sessions after a version-floor error) — untouched.

Each is tracked on its own issue with its own diagnosis. An empty draft PR restating them advertises coverage that does not exist, which is the same defect this repo has spent the week removing from its tools.

Issue #955, opened 2026-08-06
panel_strip_workflow swaps custom-node widgets when UI order differs from object_info


artokun:
### What happened / root cause
On a live canvas, `panel_strip_workflow` converted two `Wildcard Processor [Eclipse]` nodes with the wrong API inputs: the UI showed numeric seeds 101/202, but the stripped Prompt API assigned the preceding `wildcards` combo strings to `seed` and substituted defaults into `wildcards`. `graph_serialize` provides positional `widgets_values` in the custom frontend's visual order (`wildcards` before `seed`), while `convertUiToApi` maps positions using `/object_info` input order (`seed` before `wildcards`). The existing name-keyed `graph_get_state` reconstruction avoids this class of drift but is only used when `graph_serialize` is unsupported.

### Steps to reproduce
1. Add `Wildcard Processor [Eclipse]` on the live canvas.
2. Set `wildcard_text`, `mode=populate`, `seed=101`, and any wildcard combo value.
3. Confirm `panel_query_graph` reports the numeric seed correctly.
4. Call `panel_strip_workflow()` with no explicit graph/path.
5. Observe the wildcard combo string emitted as the Prompt API `seed`.

### Exact error (scrubbed)
Canvas: `seed: 101`, `wildcards: __smart_prompt/01_subjects/10_face__`
Stripped API: `seed: __smart_prompt/01_subjects/10_face__`, `wildcards: Select a Wildcard`
The tool also emitted a conversion note saying the numeric widget value was invalid for `wildcards`.

### Fix
Applied locally: yes. Upstream-only: no.

For live strip, retain the complete `graph_serialize` workflow but additionally request `graph_get_state`; when that snapshot is complete, overlay each node's name-keyed `widgets` object onto its serialized `widgets_values` before `convertUiToApi`. This preserves links/groups/properties/subgraphs while sidestepping positional custom-widget order drift. Ambiguous duplicate names remain governed by the converter's existing fail-closed logic. The overlay is best-effort and falls back to the original serialized graph for older panels. `node --check dist/orchestrator/panel-tools.js` exited 0.

```diff
 const wf = reply?.workflow;
+if (allowStateFallback && Array.isArray(wf.nodes)) {
+  const state = await ctx.bridge.send({ cmd: "graph_get_state" }, ...);
+  if (completeState(state)) {
+    const widgetsById = new Map(state.nodes.map(n => [String(n.id), n.widgets]));
+    for (const node of wf.nodes) {
[...]

Issue #938, opened 2026-08-06
No panel tool to remove a single dynamic widget row (e.g. rgthree Power Lora Loader lora_N rows)


artokun:
### What happened / root cause

User asked me to remove the 3 unused LoRA rows from an rgthree `Power Lora Loader (rgthree)` node on the live canvas. I could not do it with any available panel tool.

The panel exposes:
- `panel_set_widget` — can CHANGE a widget's value (it happily round-tripped the row's object value, see below), but cannot DELETE the widget.
- `panel_remove_node` — deletes an ENTIRE node, not one widget.
- `panel_edit_node` — presentation only (pos/size/title/color/mode).
- `panel_promote_widget` — subgraph boundary promotion only (`demote:true` un-promotes from the parent node; it does not delete a widget on a normal node).

Nodes with DYNAMIC/repeatable widget rows are common in the ecosystem — rgthree Power Lora Loader (`lora_1..lora_N`), Impact/Inspire list nodes, various "add row" style custom nodes. Their add/remove affordance is a DOM-level custom widget (an x button / right-click item) that an agent cannot click, so those rows are effectively un-removable from an agent session. The user has to do it by hand, which breaks the "I can edit your canvas for you" contract.

### Steps to reproduce

1. Have `rgthree-comfy` installed; put a `Power Lora Loader (rgthree)` on the canvas with several lora rows, some toggled off.
2. Ask the agent to remove the unused rows.
3. `panel_query_graph {ids:[<id>], fields:'detail'}` shows the rows as widgets `lora_1`, `lora_2`, ... with object values `{on, lora, strength, strengthTwo}`.
4. There is no tool that removes one of those widgets. `panel_set_widget` can only overwrite the value.

### Exact error (scrubbed)

No error thrown — this is a missing capability, not a crash. `panel_set_widget` succeeded and returned:

```
{"set":{"node_id":"35","widget":"lora_1",
  "previous":{"on":false,"lora":"SDXL\\Realism_Lora_By_Stable_yogi_SDXL8.1.safetensors","strength":0.8,"strengthTwo":null},
  "value":{"on":false,"lora":"None","strength":0,"strengthTwo":null}}}
```

so the value is writable, but the ROW remains. (I reverted this immediately — no user data changed.)

### Fix

applied locally: no   upstream-only: yes

[...]

Issue #910, opened 2026-08-05
panel_add_node refuses any node whose REQUIRED input is a custom-typed SOCKET (misreads sockets as required custom widgets)


artokun:
### What happened / root cause

`panel_add_node` fails with `Required custom widget "<TYPE>" have not registered. They may be custom widgets still loading; retry shortly.` for any node that has a **required input whose type is a custom (non-primitive) type**. The message is misleading in two ways: the input is a **link socket**, not a widget, and retrying never helps — it is not a loading race.

Three confirmed cases in one session:

| Node | Required input | Type |
|---|---|---|
| `SaveVideo` (core) | `video` | `VIDEO` |
| `LoadImage` (core) | `upload` | `IMAGEUPLOAD` |
| `easy multiTrackInfoOutput` (ComfyUI-Easy-Media) | `tracks_info` | `TRACKS_INFO` |
| `easy multiTrackTaskOutput` (ComfyUI-Easy-Media) | `tracks_info` | `TRACKS_INFO` |

Note `SaveVideo` and `LoadImage` are **core ComfyUI nodes**, so this is not limited to custom packs.

Strong signal for the fix: **`panel_copy_nodes` + `panel_paste_nodes` of an existing instance works perfectly every time.** The paste path clones an existing node instead of constructing widgets from the schema, so the defect is specific to the add/construct path's handling of required custom-typed inputs. It appears to classify every required input as a widget needing a registered constructor, rather than checking whether the input is a socket (link-only) vs a widget.

### Steps to reproduce

1. `panel_add_node(class_type: "SaveVideo")` -> error.
2. `panel_add_node(class_type: "LoadImage")` -> error.
3. Install/enable ComfyUI-Easy-Media, restart, then `panel_add_node(class_type: "easy multiTrackInfoOutput")` -> error.
4. Confirm the node IS present server-side: `GET /object_info` lists it with `input.required.tracks_info = ["TRACKS_INFO", {}]`.

Ruled out during this session:
- Not a load race — retried repeatedly over several minutes.
- Not stale node defs — `panel_refresh_nodes` (re-pulls /object_info) did not help.
- Not a stale frontend bundle — a browser hard refresh (Ctrl+Shift+R) did not help.
- Not a missing pack — sibling nodes from the SAME pack that have no custom-typed required input (`easy multiTrackEditor`, `easy minimaxH3ToVideo`) add successfully.

### Exact error (scrubbed)

```
[...]

Issue #845, opened 2026-08-04
panel_view_nodes_in_viewport returns the whole graph (135k chars) for a single node_id; panel_screenshot ignores canvas viewport


artokun:
### What happened / root cause

Three defects hit in one short session while inspecting ONE node (an `LTXDirector`, id 42) in a 175-node workflow.

**1. `panel_view_nodes_in_viewport` floods context.** Called with `node_ids: ["42"]` — a single node — and it returned **135,531 characters across 5,662 lines**, tripping the harness token guard and getting spilled to a file. It appears to serialize every node in the viewport (or the whole graph) rather than the requested node(s). This is the exact failure mode `panel_query_graph` was token-bounded to prevent; `panel_view_nodes_in_viewport` has no such bound. Practical effect: the tool is unusable on any real workflow.

**2. `panel_screenshot` ignores canvas viewport state.** After `panel_canvas {action:"center_on_node", node_id:42}` and then `panel_canvas {action:"zoom", scale:0.55}` (both returned success, with scale 0.55 and a new offset), three consecutive `panel_screenshot` calls returned **pixel-identical fit-all framing** of the whole 175-node graph. The only difference between frames was the FPS counter. So there is no way to visually inspect one node — which is precisely what you need when verifying that a widget write caused a custom node's UI to re-render.

**3. `panel_canvas` arg naming is inconsistent.** `action:"center_on_node"` silently ignores a `zoom` argument (returned scale 0.067). `action:"zoom"` rejects `zoom` and requires `scale`. Two names for the same concept on the same tool; the first fails silently, the second errors.

**4. (minor) `panel_select_nodes` demands numeric node_ids.** Every other panel tool returns node ids as **strings** (`"id": "42"`) and accepts strings. `panel_select_nodes` rejects `["42"]` with a zod `expected number, received string`. Copying an id straight out of `panel_query_graph` output into `panel_select_nodes` therefore always errors on the first try.

### Steps to reproduce

1. Load a large workflow (175 nodes; this was `LTX 2.3 Director - Clean.json`).
2. `panel_view_nodes_in_viewport {node_ids: ["42"]}` → 135k-char response.
3. `panel_select_nodes {node_ids: ["42"]}` → validation error; `[42]` succeeds.
4. `panel_canvas {action:"center_on_node", node_id:42, zoom:0.55}` → returns `scale: 0.067` (zoom ignored).
[...]

PR #959, opened 2026-08-07
cluster: graph read/write fidelity — strip swaps widgets, add_node refuses, viewport returns everything

artokun:
Tools that read or rewrite the graph return something that does not faithfully represent it. These corrupt or hide the user's actual workflow, so they rank above cosmetic issues.

## Issues in this cluster

| Issue | Sev | Fidelity broken | Status |
|---|---|---|---|
| #961 | P2 | `panel_strip_workflow` shifts Eclipse wildcard widgets when frontend order differs from the definition | **fix in #966** |
| #955 | P2 | `panel_strip_workflow` swaps custom-node widgets when UI order differs from the definition | **fix in #966** |
| #361 | P2 | `panel_strip_workflow` returns wrong dynamic widget values and drops Set/Get nodes | **widget half fixed in #966**; Set/Get link loss not re-verified |
| #910 | P2 | `panel_add_node` refuses any node whose REQUIRED input is a custom type | open |
| #845 | P2 | `panel_view_nodes_in_viewport` returns the whole graph (135k chars) instead of the viewport | open |
| #938 | P3 | No panel tool to remove a single dynamic widget row (e.g. rgthree Power Lora Loader) | open |

## #961 / #955 / #361 were ONE bug reported THREE times — fixed in #966

Three independent reports, three different node packs (Eclipse wildcards, generic custom nodes, Set/Get), one mechanism: **widgets are mapped by INDEX, and the frontend's widget order does not match the node definition's order.** Whenever those two orders disagree, values land on the wrong widget.

That is data corruption, not a display bug — a shifted widget silently changes what the user renders, and the workflow they save is wrong. #361 is the severity case: five model/LoRA/VAE filenames were each replaced by a *different file on disk*.

**#966** fixes all three together, as this cluster required: it takes a second, name-keyed read of the live canvas (`graph_get_state` reports widgets BY NAME) and maps from that instead of from positions. It refuses — and discloses — when it cannot prove the two captures describe the same nodes.

The prediction in the "known-failed prior attempts" note below held: the fix had to leave positional mapping behind entirely, and #966 does.

## Still open in this cluster

**#910**, **#845**, **#938** are untouched. #845 is the cheapest and most annoying (a viewport read returning 135k chars defeats the point of scoping to the viewport). #361 also reports **Set/Get links going missing** on the same canvas — #966 does not address that half, and it has not been re-verified since; it may have been a downstream effect of the widget corruption, or a separate bug.

## Known-failed prior attempts

#361 is long-standing and survived earlier passes. That longevity was the signal: any fix that kept mapping by position would keep failing, and would keep producing a report per node pack. The arrival of #961 as a *third* instance was the evidence that the earlier narrowing was wrong.

Mutation-verification was treated as mandatory and #966 carries it: breaking each guard kills exactly the tests that claim to cover it.


artokun:
**Investigation notes — where the index mapping actually lives.** Recording this so the fix does not start from scratch.

The slot mapping is driven by `isPositionalWidgetSpec()` in `src/services/workflow-converter.ts` (~line 95). It decides which inputs occupy a slot in the UI's `widgets_values` array, walking the **node definition's** input order and counting positional widgets: inline combos (`["a","b"]`), v3 dynamic combos (a string type carrying an `options` list), and scalar widget types qualify; link/list types (IMAGE, COMFY_AUTOGROW_V3, …) do not.

**That is the defect, stated precisely:** the mapping assumes the definition's positional order equals the frontend's actual widget order. All three reports say exactly the opposite — #961 (Eclipse wildcards), #955 (generic custom nodes), #360/#361 (Set/Get). When a custom node's frontend widget order differs from its definition order, every value after the divergence lands on the wrong widget.

So the fix is not a new special case per node pack. It needs the **frontend's real widget order** as the source of truth rather than the definition's, and the two must be reconciled explicitly where they disagree — with the disagreement surfaced rather than silently resolved, since picking the wrong one silently changes what the user renders.

Two things to be careful of when implementing:

1. `ASSET_FILE_RE` / `looksLikeAssetFilename` nearby exists to distinguish an **asset-selection** combo (where substituting a different installed file silently swaps the user's model) from a plain enum combo where falling back to the first option is harmless. A shifted widget can turn one into the other — the corruption path is worse for asset combos.
2. This is data-corruption shaped, so mutation-verification is mandatory: break the mapping deliberately, prove the specific test fails, restore. A test that passes both ways proves nothing here.

Not started beyond this analysis.

artokun:
Closing this scoping draft (1 commit, 0 files). The body above is current; summarising the outcome here.

**Shipped:** #961, #955 and #361 were one bug reported three times — widgets mapped by INDEX against `object_info`'s order while `graph_serialize` stores them in the FRONTEND's. Fixed together in #966, which takes a second name-keyed read of the canvas and refuses (with disclosure) when it cannot prove the two captures describe the same nodes.

The cluster's own prediction held: *"any fix that keeps mapping by position will keep failing, and will keep producing a report per node pack."* The fix had to leave positional mapping behind entirely, and it did.

**Still open, and all panel-side:**
- **#910** — `panel_add_node` refuses any node whose REQUIRED input is a custom-typed socket. The refusal text (`Required custom widget "TRACKS_INFO" have not registered`) is emitted by `comfyui-mcp-panel.js:6489`; the add/construct path misreads a link socket as a widget needing a registered constructor. `panel_copy_nodes`/`panel_paste_nodes` of an existing instance works every time, which is the strong signal.
- **#845** — `panel_view_nodes_in_viewport` returns the whole graph.
- **#938** — no tool to remove a single dynamic widget row.

Each is tracked on its own issue with its own diagnosis, so an empty draft PR restating them adds noise rather than coverage.

One thread from #361 is worth not losing: it also reported **Set/Get links going missing** on the same canvas. #966 did not address that, and it has not been re-verified since — it may have been a downstream effect of the widget corruption, or a separate bug. Noted on the issue.

Issue #949, opened 2026-08-06
Fix batch panel_run completion correlation for prompt_ids


artokun:
### What happened / root cause
panel_run with batch_count > 1 returned both prompt_id and prompt_ids, but dist/orchestrator/panel-tools.js opened a RunCompletions ticket only for parsed.prompt_id. Completions for the remaining prompt_ids were therefore classified as foreign/UNDETERMINED even though the agent queued them.

### Steps to reproduce
1. Call panel_run with batch_count: 3.
2. Observe the tool result includes prompt_ids for all three runs.
3. Let the renders complete.
4. Secondary completion events say they do not match any run queued by panel_run.

### Exact error (scrubbed)
This run (prompt <uuid>) does NOT match any run you queued with panel_run - its origin is UNDETERMINED.

### Fix
applied locally: yes  upstream-only: no
Parse every valid id in prompt_ids, include prompt_id if absent, call QueueMonitor.markSelfQueued for each id, and open a RunCompletions ticket for each id. Keep the no-id fallback unchanged. Syntax verified with node --check; orchestrator soft-reloaded.

Patch summary:
- const queuedId -> const queuedIds
- collect parsed.prompt_ids plus parsed.prompt_id
- markSelfQueued for every queued id
- RunCompletions.openRun for every queued id

### Environment
Windows 11 / ComfyUI 0.30.0 / RTX 4070 SUPER 12 GB / comfyui-mcp 0.49.8 / panel 0.11.38

Issue #948, opened 2026-08-06
Provider CLI auth errors are surfaced verbatim into panel chat — "please run /login" names a command that does not exist where the user is standing


artokun:
Reported by **seanmcmagic** in Discord `#help`:

> for some reason it logged me out of the panel and is saying "Not logged in, please run /login" but that doesn't exist as a command

He's right. `/login` is a slash command typed **inside the `pi` CLI**. It is not a panel command, not a ComfyUI command, and not a shell command — and the message names none of that. In a chat panel a leading `/` reads as *"type this here"*, so the instruction actively points the user at the wrong place.

## Why this is ours and not upstream's

Our own wrapped messages get it right. `index.ts` and `pi-backend.ts` both say some form of:

> run `pi` once and `/login` (stored in `~/.pi/agent/auth.json`), then Disconnect → Connect

That names the tool, the file, and the follow-up action. The defect is that a **raw provider-CLI error can reach panel chat without passing through that wrapping** — so the user sees the CLI's own phrasing, which assumes you are already standing in the CLI.

It also misleads across providers: a user on the Claude or Codex chip has no `/login` at all (`claude setup-token` / `codex login`), so if this string can surface for them it is not merely unhelpful but wrong.

## Suggested fix

Treat provider auth failures as a class with one wrapper, rather than passing the child's stderr through:

1. Detect the not-authenticated condition per backend.
2. Emit **our** message, which must name **where** to run the remedy, not just what to type — the failure mode here is a correct instruction delivered to the wrong context.
3. Never surface a bare `/`-prefixed instruction into chat; if a remedy is a CLI slash command, render it as ``run `pi`, then type `/login` at its prompt``.

Related class: `#809` (truncated results must name their own remedy) — same principle, that a message which tells the user to do something must be actionable from where they are reading it.

## Also worth splitting out

The reporter said it "logged me out". Worth establishing whether the session was genuinely invalidated or whether the panel merely lost the ability to *read* the existing credential — a stale or unreadable `~/.pi/agent/auth.json` presents identically from the panel's side, and the two have different remedies. If they are indistinguishable today, that is a second defect: the panel is reporting a cause it cannot actually determine.

artokun:
## The reporter demonstrated the defect while we were debugging it

Told to "run /login", he opened Command Prompt and typed it there:

```
C:\Users\swilb>pi
'pi' is not recognized as an internal or external command,
operable program or batch file.

C:\Users\swilb>/login
[...]

Issue #946, opened 2026-08-06
call_tool: upload_image/upload_video 'filename' override containing '/' throws a raw JSON.parse SyntaxError instead of uploading


artokun:
### What happened / root cause
Calling the compact `call_tool` facade to invoke `upload_image` (and separately `upload_video`) with a `filename` override that contains a forward slash (intended to place the upload in a subfolder of ComfyUI's input/ dir, e.g. `assets/buildings.mp4`) fails immediately with a raw JSON parse error instead of either uploading to that subfolder or returning a normal validation error:

```
Unexpected non-whitespace character after JSON at position 4 (line 1 column 5)
```

The identical call with the `filename` field omitted (same `source_path`, no override) succeeds normally and uploads to ComfyUI's input/ root. Also, `source_path` itself routinely contains many forward slashes (e.g. `C:/Users/.../ltxv_assets/buildings.mp4`) and does NOT trigger the error — only a slash inside `filename` does. So this isn't 'any slash in any string arg breaks JSON.parse'; it's specific to how the `filename` argument is handled somewhere between the client request and the tool handler.

I read `dist/tools/compact.js`'s `call_tool` handler (the code backing this facade): when `args` arrives as an object (the normal case for this client), it's used directly at `rawArgs = args` with no `JSON.parse` at all — the `JSON.parse(args)` branch only runs when `args` arrives as a raw string. That means either (a) something upstream of `compact.js` (transport layer, or the calling client's own serialization of the nested `mcp__comfyui__call_tool` invocation) is re-stringifying `args` and re-parsing it, and a `/` inside one particular field corrupts that round-trip, or (b) there's a second code path I didn't find that treats `filename` specially (e.g. splitting on `/` for a path/pointer). I could not pin down the exact location from the installed `dist/` alone without request-level server logs, so I'm filing with a solid repro rather than guessing further and risking a wrong patch.

### Steps to reproduce
1. Have a local file to upload, e.g. `C:/Users/<user>/AppData/Local/Temp/somefile.mp4`.
2. Call `call_tool` with `{"name": "upload_video", "args": {"source_path": "C:/Users/<user>/AppData/Local/Temp/somefile.mp4", "filename": "assets/somefile.mp4"}}`.
3. Observe: `Unexpected non-whitespace character after JSON at position 4 (line 1 column 5)` — no upload happens.
[...]

Issue #937, opened 2026-08-06
report-bug intake: Cloudflare 403 / error 1010 blocks `Python-urllib` by UA signature — single-variable repro (not rate limiting)


artokun:
### What happened / root cause

The intake Worker returned **HTTP 403 with Cloudflare `error code: 1010`** for a programmatic POST. The identical request succeeded **13.8 seconds later** with only a browser `User-Agent` header added.

> Filed here rather than in `comfyui-mcp-issue-worker` because `report_issue` only auto-files into `comfyui-mcp` / `comfyui-mcp-panel` — the worker repo returns a prefilled link only, so worker defects can't be reported autonomously. Suggested fixes 3 and 4 below are `comfyui-mcp` changes regardless (the `report-bug` skill snippet). Please move if the WAF config is the preferred home.

Follow-up to the discussion on comfyui-mcp-panel#690. My first diagnosis there — *"non-browser clients silently can't file"* — was **wrong**, and the maintainer correctly refuted it: plain `curl`, Chrome-UA `curl`, and Node `fetch` all return 401 identically when unauthenticated (auth is evaluated before the bot check), and #892/#902/#912/#914/#916/#918 all filed successfully via Node `fetch` with no browser UA. A blanket non-browser block is definitively not in force.

**The narrower explanation that fits all the evidence: Cloudflare is banning the `Python-urllib` UA signature specifically.** CF error 1010 is the browser-signature ban — not rate limiting, not a challenge. The failing client was Python's `urllib.request`, whose default header is `User-Agent: Python-urllib/3.13`. Neither probe client sends that — `curl` sends `curl/8.x`, Node `fetch` sends an undici/node UA — which is exactly why the probe couldn't reproduce it, and why the Node-fetch `report_issue` path is unaffected.

### Steps to reproduce

Three POSTs, same machine, same IP, same `X-Client-Key` (the default soft anti-spam key shipped with the panel — `[REDACTED]`), **byte-identical JSON body read from the same file**, all within 14 seconds:

| # | timestamp (UTC) | client | User-Agent | result |
|---|---|---|---|---|
| 1 | `2026-08-06T02:19:10.063Z` | `urllib.request` | default (`Python-urllib/3.x`) | **HTTP 403** |
| 2 | `2026-08-06T02:19:16.587Z` | `urllib.request` | default (`Python-urllib/3.x`) | **HTTP 403**, body: `error code: 1010` |
[...]

PR #960, opened 2026-08-07
cluster: ignored search filters, unsafe upload names, bad defaults, credential races

artokun:
Independent defects that are each small, share no root cause, and are cheapest to land as one reviewed batch.

## Status — 6 of 9 shipped

| Issue | Sev | Defect | Status |
|---|---|---|---|
| #946 | P2 | `call_tool`: `upload_image` `filename` override containing `/` | **fixed** (#985) |
| #949 | P2 | Batch `panel_run` completion correlation for `prompt_ids` | **fixed** (#994) |
| #948 | P2 | Provider CLI auth errors surfaced verbatim into panel chat | **fixed** (#989) |
| #937 | P2 | report-bug intake: Cloudflare 403 / error 1010 blocks `Python-urllib` by UA signature | **fixed** (#1000) |
| #890 | P2 | Legacy Manager update no-op leaves panel graph writes blocked | **fixed** in 0.50.1 (remedy now names the exact `git -C … pull --ff-only`) |
| #892 | P2 | `generate_image` checkpoint auto-select picks the alphabetically-first checkpoint | **fixed** (closed earlier) |
| #935 | P2 | `panel_civitai_search`: documented filters appear to be ignored | **half fixed** (#993) — see below |
| #881 | — | Credential store: a concurrent save is silently overwritten and still reports success | **open, and still live** — see below |
| #908 | P2 | RunPod Deploy/Start buttons dead since #278 | **open** — #1008 is in flight |

## #908 — the diagnosis moved

The original reading was that the panel calls the two retired names over the admission channel. The 0.50.0 fold changed the shape: `runpod` is now a single admitted *name* with an action allowlist, so those buttons route through the **action** branch — which returned a bare `tool "runpod" is not permitted for action "start"`, the same dead end the NAME branch had just been taught not to return (#965).

So the consolidation quietly undid #908's fix for the two buttons it was reported about. #1008 addresses that. **The exclusion itself is unchanged and stays correct** — deploy and start both begin billing, and a confirmation-less channel must not spend money.

## #935 — half is panel-side

#993 extended the unapplied-filter guard past `creator`, so a supplied-but-ignored `modelSort`/`baseModels` is now reported instead of answered with a plausible list. Whether the filters actually *apply* is panel-side (`web/js/cmcp-civitai.js`, `h.civitai.search`) and remains open — which is why #993 was `Refs`, not `Closes`.

## #881 — closed as fixed, and reopened with evidence

It was closed today as completed against #837. It is not fixed: the reproduction is still `it.fails` on main, and flipping it to a plain `it()` on unmodified `main` fails with `expected … to contain 'HF_TOKEN=theirs'`.

#837 genuinely closed the **read→swap** window. This issue is the **link→rename** window, which the harness targets deliberately so the compare-and-swap cannot catch it. Reopened with that account.

I also attempted it and did not land it — the inode premise is sound and the call path is right, but I could not establish whether my guard executed, and I reverted rather than ship an unverified change into the credential store.

## Keeping this open

Unlike #956, this one still has live work: #908 (in flight), #935's panel half, and #881. Body updated so it stops advertising the six that are done.


artokun:
**#892 removed from this cluster — already fixed and shipped** (PR #921, v0.50.0/v0.50.1; owner-qualified `Closes`, so it never auto-closed).

The cluster body already warned to read what PR #921 landed before assuming #892 was untouched. It was not untouched — it was finished: `fix(#865,#892)` plus `test(#892)` across several gate rounds.

Remaining: #935, #946, #949, #948, #881, #890, #937, #908.

artokun:
Retiring this — it holds **no commits** (`git diff origin/main...` is empty). It was a coordination placeholder, and every item in it has since landed as its own reviewed PR, so leaving an empty branch open only invites someone to mistake it for pending work.

Final accounting — **8 of 9 shipped**:

| Issue | Landed as |
|---|---|
| #946 upload filename containing `/` | #985 |
| #949 batch `panel_run` completion correlation | #994 |
| #948 provider CLI auth errors verbatim into chat | #989 |
| #937 report-bug intake blocked by UA signature | #1000 |
| #890 legacy Manager update no-op | 0.50.1 |
| #892 `generate_image` alphabetically-first checkpoint | closed earlier |
| #935 civitai filters supplied-but-ignored | #993, issue now closed |
| #908 RunPod Deploy/Start dead | #1008 — the 0.50.0 fold had quietly moved the refusal from the NAME branch to the ACTION branch, re-opening the dead end #965 had just closed |

**Still open, and tracked on its own:** #881 — the credential store's concurrent-save window. Not folded into any PR, and deliberately so: I implemented the inode drift check, could not get the reproduction past  after three instrumentation attempts, and reverted rather than ship an unverified change into the credential store. #837 closed the read→swap window; this is the link→rename one, and a second writer (`writeFileAtomic`) has no sentinel at all. The blocker is instrumentation, not the guard design.

Nothing is lost by closing this — #881 is the only live work and it lives at #881.

artokun:
Correction to one sentence above — the shell ate a backticked token. It should read: "could not get the reproduction past `it.fails` after three instrumentation attempts."

PR #966, opened 2026-08-07
fix(#959): map live-canvas widgets by NAME, and disclose when we can't

### LINKED ISSUE #361: panel_strip_workflow returns wrong dynamic widget values and drops Set/Get links
`panel_strip_workflow()` on the live SteadyDancer canvas returned a stripped API graph whose widget values and resolved Set/Get wiring do not match `panel_graph_outline()` for the same live canvas. This appears to be an upstream serialization/conversion defect, likely around dynamic COMBO widgets and rgthree SetNode/GetNode wildcard links.

### Steps to reproduce
1. Open the workflow `????SteadyDancer????` containing dynamic model combo widgets and SetNode/GetNode buses.
2. Call `panel_graph_outline()` and record live values/wiring.
3. Call `panel_strip_workflow()` with no arguments.
4. Compare the returned API graph.

### Exact error (scrubbed)
No exception is thrown; the tool succeeds with misleading output. Examples:
- node 3: live `clip_vision_h.safetensors`; stripped `CLIP-ViT-H-14-laion2B-s32B-b79K.safetensors`
- node 11: live `umt5-xxl-enc-bf16.safetensors`; stripped `ViT-L-14-TEXT-detail-improved-hiT-GmP-HF.safetensors`
- node 13: live LightX2V WAN LoRA; stripped `Florence-2-Flux-Large.safetensors`
- node 32: live `Wan2.1_VAE.pth`; stripped `LTX23_audio_vae_bf16.safetensors`
- node 53: live SteadyDancer WAN model; stripped LTX-2.3 GGUF
- node 57: live pasted input image; stripped `Architecture.png`
- Set/Get-resolved links are also missing from nodes 22, 42, 59, 62, and 63 despite the outline showing those virtual-bus connections.

### Fix
Suggested fix: preserve live UI widget values by name during UI-to-API conversion instead of re-resolving dynamic COMBO indices against the current server list; resolve rgthree SetNode/GetNode wildcard outputs before removing bus nodes, and validate that every downstream input formerly fed by a GetNode still has a concrete connection.

### Environment
Windows 10; ComfyUI 0.27.0 local; RTX 3090 24 GB; comfyui-mcp 0.48.5; panel version unavailable (`install_panel status` reports no separately installed panel pack).

### LINKED ISSUE #955: panel_strip_workflow swaps custom-node widgets when UI order differs from object_info

### LINKED ISSUE #961: panel_strip_workflow shifts Eclipse wildcard widgets when frontend order differs from object_info
On a live canvas containing `Wildcard Processor [Eclipse]`, `panel_strip_workflow` reconstructed the node's widget values positionally using `object_info.input_order`, but the frontend serializes `wildcards` before `seed`. The resulting prompt graph shifts values: an unlinked node can emit `seed: "Select a Wildcard"`, while linked-seed nodes report `wildcards` as an empty invalid option and substitute it. The live canvas itself reports the correct named values.

### Steps to reproduce
1. Add `Wildcard Processor [Eclipse]` to a live canvas.
2. Set `wildcard_text`, `mode=populate`, `wildcards=Select a Wildcard`, and `seed=0`; optionally link an INTConstant to `seed`.
3. Confirm `panel_query_graph` or `panel_graph_outline` reports the correct named values.
4. Call `panel_strip_workflow`.
5. Observe the conversion warning for `wildcards`; without a linked seed, observe `seed: "Select a Wildcard"` in the stripped API graph.

### Exact error (scrubbed)
`Node 296 (Wildcard Processor [Eclipse])` was stripped with `seed: "Select a Wildcard"` despite the live node reporting `seed=0`, and linked-seed nodes warned that widget `wildcards` had value `""` even though the live node reported `Select a Wildcard`.

### Fix
Applied locally: no. Upstream-only: yes. The active orchestrator is 0.49.8, while the only inspectable package copy on the host is 0.49.5, so patching it would not affect the running tool and would be unsafe.

Precise change needed: when `panel_strip_workflow` receives a live-canvas capture, prefer the panel's authoritative name-to-value widget map over positional `widgets_values` for custom nodes whose frontend order differs from `object_info.input_order`. At minimum add a regression for `Wildcard Processor [Eclipse]` where the serialized order is `[wildcard_text, populated_text, mode, wildcards, seed, ...UI buttons]` while object_info orders seed before wildcards. Ensure output keeps `seed=0`, `wildcards=Select a Wildcard`, and a linked top-level seed does not consume or shift an uncertain placeholder slot.

### Environment
Connected runtime: Linux; ComfyUI 0.29.0; NVIDIA A100-PCIE-40GB (39 GB VRAM); comfyui-mcp 0.49.8; panel 0.11.41.

artokun:
Closes #961
Closes #955
Closes #361

## The bug

`panel_strip_workflow` mapped widget values by **position**. The positions come from `graph_serialize`, whose order is the **frontend's**; the names come from `object_info`, whose order is the **server's**. For custom nodes that add, hide, or reorder widgets in JS the two disagree, and every value from the first disagreement onward lands on the wrong widget — with no error anywhere.

Reported three times against three node packs before it read as one bug:

| Issue | Symptom |
|---|---|
| #961 | Eclipse `wildcards` serialized before `seed` → a wildcard **string** emitted as the seed |
| #955 | Same node, seeds 101/202 swapped out for combo strings |
| #361 | SteadyDancer canvas: five model/LoRA/VAE filenames each replaced by a **different file on disk** |

#361 is the severity argument — the user renders something they never asked for, and nothing reports a problem.

The shift also **compounds**: `seed` carries a `control_after_generate` phantom slot, so once the string is taken as the seed the real seed is swallowed as the phantom and the trailing widget is never assigned at all. One inverted pair costs both widgets.

## The fix

Take a second, name-keyed read of the same canvas — `graph_get_state` already reports each node's widgets **by name**, straight off the live litegraph objects — and hand the converter the named map. A name-keyed map has no order to disagree about, and it routes into the converter's existing object branch, which already fails closed on an ambiguous name.

The capture lands in its own `capturedWidgetValues` field rather than being written into `widgets_values`. Several call sites read that array **by index** — a subgraph node's `proxyWidgets`, a `PrimitiveNode`'s literal at `[0]`, a Set/Get node's bus name — and handing them an object would silently drop all three. A test pins the PrimitiveNode case.

## What it refuses

It refuses rather than guesses, and every refusal is disclosed:

- **Subgraph scope.** `graph_serialize` always serializes the ROOT graph, but `graph_get_state` reports whichever graph the user is *inside*. With a subgraph open the two id spaces belong to different graphs — overlaying node 7 of a subgraph onto node 7 of the root would be a worse bug than the one being fixed. Scope must be proven `root`; an absent scope refuses too.
- **Duplicate node id** — can't be attributed to either claimant, so neither wins.
- **Partial capture** (the panel reports at most 100 nodes per read) — fixes what it covers, names what it didn't.

Every refusal says the mapping stayed positional and **MAY** be wrong. Unverified is not the same as known wrong, and asserting the stronger claim is the defect class this repo keeps hitting. The notes ride with strip's existing conversion notes.

## Why this shape and not the narrower one

#361 has survived earlier passes. That longevity is the signal: any fix that keeps mapping by position keeps failing, and keeps producing one report per node pack. #961 arriving as a *third* instance is the evidence the earlier narrowing was wrong.

## Verification

- 13 new tests; full suite **318 files / 6853 passed**; `tsc --noEmit` and `check:vocabulary` clean.
- **Mutation-verified**: reverting the converter to the positional source kills 5 tests including all three corruption cases; removing the root-scope guard kills 2; accepting duplicate ids kills 1.

### DIFF
--- a/src/__tests__/services/live-widget-overlay.test.ts
+++ b/src/__tests__/services/live-widget-overlay.test.ts
@@ -1,0 +1,277 @@
+// #959 — one bug reported three times: #961 (Eclipse wildcards), #955 (the same
+// node with numeric seeds), #361 (five model/LoRA/VAE filenames each replaced by a
+// different file on disk). The mechanism is shared: `graph_serialize` stores widget
+// values POSITIONALLY in the FRONTEND's order, and the converter names those
+// positions from object_info's input order. Custom nodes that add or reorder
+// widgets in JS make the two disagree, and every value from the first disagreement
+// onward lands on the wrong widget — silently, with a valid-looking graph.
+//
+// These tests cover both halves of the fix: the overlay decides whether the
+// name-keyed capture describes the SAME nodes as the serialized graph (refusing
+// when it can't tell), and the converter then maps by name instead of by index.
+
+import { describe, expect, it } from "vitest";
+import { applyCapturedWidgetValues } from "../../services/live-widget-overlay.js";
+import { convertUiToApi } from "../../services/workflow-converter.js";
+
+// ── the reported node ───────────────────────────────────────────────────────
+// object_info orders `seed` BEFORE `wildcards`; the Eclipse frontend serializes
+// `wildcards` BEFORE `seed`. That single inversion is the whole bug.
+const ECLIPSE = "Wildcard Processor [Eclipse]";
+const OBJECT_INFO = {
+  [ECLIPSE]: {
+    input: {
+      required: {
+        wildcard_text: ["STRING"],
+        populated_text: ["STRING"],
+        mode: [["populate", "fixed"]],
+        seed: ["INT"],
+        wildcards: [["Select a Wildcard", "__smart_prompt/01_subjects/10_face__"]],
+      },
+    },
+  },
+} as never;
+
+const WILDCARD = "__smart_prompt/01_subjects/10_face__";
+
+/** The live canvas as `graph_serialize` returns it: frontend widget order. */
+const serializedCanvas = () => ({
+  nodes: [
+    {
+      id: 296,
+      type: ECLIPSE,
+      mode: 0,
+      inputs: [],
+      outputs: [],
+      // frontend order — wildcards BEFORE seed
+      widgets_values: ["a cat", "a cat", "populate", WILDCARD, 101],
+    },
+  ],
+  links: [],
+});
+
+/** The same canvas as `graph_get_state` returns it: name-keyed, order-free. */
+const nameKeyedState = () => ({
+  viewing: { scope: "root" },
+  node_count: 1,
+  truncated: false,
+  nodes: [
+    {
+      id: 296,
+      type: ECLIPSE,
+      widgets: {
+        wildcard_text: "a cat",
+        populated_text: "a cat",
+        mode: "populate",
+        seed: 101,
+        wildcards: WILDCARD,
+      },
+    },
+  ],
+});
+
+describe("#959: the bug, reproduced", () => {
+  // This is the failure the three reports describe. It is asserted deliberately:
+  // if this test ever stops failing on its own, the positional path changed and
+  // the overlay below is no longer testing what it claims to.
+  it("without the capture, the wildcard STRING is emitted as the seed", () => {
+    const { workflow } = convertUiToApi(serializedCanvas() as never, OBJECT_INFO);
+    expect(workflow["296"].inputs.seed).toBe(WILDCARD);
+    expect(workflow["296"].inputs.seed).not.toBe(101);
+  });
+});
+
+describe("#959: the capture fixes it", () => {
+  it("with the name-keyed capture, every widget gets its own value", () => {
+    const ui = serializedCanvas();
+    const out = applyCapturedWidgetValues(ui, nameKeyedState());
+    expect(out.applied).toBe(1);
+    expect(out.refused).toBeUndefined();
+    // A fully covered graph has nothing left to caveat, so it says nothing.
+    expect(out.notes).toEqual([]);
+
+    const { workflow } = convertUiToApi(ui as never, OBJECT_INFO);
+    expect(workflow["296"].inputs.seed).toBe(101);
+    expect(workflow["296"].inputs.wildcards).toBe(WILDCARD);
+    expect(workflow["296"].inputs.mode).toBe("populate");
+    expect(workflow["296"].inputs.wildcard_text).toBe("a cat");
+  });
+
+  // The shift compounds: `seed` is an INT with control_after_generate, so the
+  // positional pass consumes a PHANTOM slot after it. Having already taken the
+  // wildcard string as the seed, it then swallows the real seed as that phantom
+  // and runs out of values — so `wildcards` is never assigned at all and falls
+  // back to the node default. One inverted pair costs BOTH widgets, and nothing
+  // in the output says so.
+  it("the trailing widget is lost entirely, not merely misassigned", () => {
+    const before = convertUiToApi(serializedCanvas() as never, OBJECT_INFO);
+    expect(before.workflow["296"].inputs.wildcards).not.toBe(WILDCARD);
+
+    const ui = serializedCanvas();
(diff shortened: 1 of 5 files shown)


artokun:
Two trade-offs worth stating plainly rather than leaving for someone to find.

**1. Fail-closed can now default a value that positional mapping happened to get right.**

Routing live strips through the name-keyed branch also subjects them to that branch's flat-slot ambiguity guard. When a V3 dynamic combo's nested leaf shares a name with another widget on the same node, the guard refuses *both* and leaves them at the node default with a loud warning — where positional mapping would have assigned something, possibly correctly.

That is a deliberate trade: a warned default beats an unwarned wrong value, and it's the behaviour the `graph_get_state` fallback path has had all along. It needs a same-named nested leaf and a top-level widget on one node, so it should be rare — but it is a behaviour change, not a pure improvement.

**2. Panels older than 2026-06-19 get a note on every live strip.**

The root-scope guard keys off `viewing.scope`, added to the panel in `0094ed0` (2026-06-19). A panel older than that answers `graph_get_state` without it, so the overlay refuses and discloses each time. Fail-closed is right here — a subgraph/root id mix-up would be worse than the bug being fixed — and the note names the remedy, but those users trade silence for a repeated caveat until they update.

**Not addressed by this PR** (still open under #959): #910 `panel_add_node` refusing custom required types, #845 viewport returning the whole graph, #938 no way to remove one dynamic widget row.

Issue #899, opened 2026-08-05
list_output_images returns filenames without the directory, so callers guess — and the obvious guess is wrong


artokun:
Found during a live browser pass on `main` (not a unit-test finding — no test would surface this).

## What happened

Asked the panel agent: *"List my 5 most recent output images and state the exact directory path they were read from."*

It answered correctly, but only after doing forensic work, and flagged the problem itself:

> `list_output_images` returns filenames but **not** the directory, so I had to derive it. The obvious guess — the workspace path from `get_environment` (`C:\Users\Artokun\ComfyUI-Installs\ComfyUI\ComfyUI\output`) — was **wrong**; those files don't exist there. Your ComfyUI is launched with `--output-directory C:\Users\Artokun\ComfyUI-Shared\output` (and `--input-directory ...`), which I read off the process command line and then confirmed by stat'ing all five files. Sizes and mtimes match the tool's report exactly, so the path above is verified, not inferred.

## Why it matters

The tool **knows** the directory — `resolveOutputDir()` just resolved it, from the live server's argv, which is exactly the machinery that makes it correct on this install. It then throws that away and returns bare filenames.

So a caller that needs the path has two options: guess, or go and re-derive it from the process command line. On this install (ComfyUI Desktop with an external `--output-directory`) the natural guess — the workspace path from `get_environment` — is **wrong**, and wrong in the silent way: the filenames look plausible against it, so a caller that does not stat them will confidently report a path that holds nothing.

This is the same family as #877 without being a false answer: the tool holds the fact and does not pass it on, so the caller reconstructs it and can reconstruct it wrong.

## Suggested fix

Return the resolved directory alongside the entries — the `OutputImage` entries already carry an absolute `path`, so the JSON shape mostly has it; the markdown/prose form is where it goes missing. Naming it once at the top of the response is enough, and it should say which source established it (live argv vs configured base), because those differ on exactly the installs where guessing fails.

Related: #877 / #883 (the resolver that makes the path correct in the first place).

PR #975, opened 2026-08-07
fix(#953): a remote listing is incomplete BY CONSTRUCTION — say so, on every result

### LINKED ISSUE #953: list_output_images: silent /history fallback on REMOTE omits VHS video outputs — reports populated output dir as empty

artokun:
Closes #953

## The bug

`list_outputs` already named its source (#899): a local scan says which directory it read; a remote answer says it came from ComfyUI's `/history`. But what that wording described was **staleness** — *"what the server remembers this session, not what the output folder holds."*

The reporter's files were never in `/history` **at all**. `VHS_VideoCombine` and similar video nodes write their file *without registering an output entry*, so those videos are absent from a history-derived listing **by construction**, not by age.

On a remote pod with 30+ ProRes masters under `output/slot/`, the tool returned a confident:

```
Found 12 media file(s) (12 video):
1. **final/BASENAME_00018.mov** [video] _(subfolder: final)_
```

…with none of them under `slot/`. They concluded the directory was empty and **wrote that into a project handover document**. The user caught it. The files were fetchable by name the whole time.

The service layer has known this for as long as the function has existed — its doc comment says *"/history often does NOT register VHS-style video outputs"*. That warning simply never reached the caller.

## The fix

Every history-derived result now states the omission as a property of the source, says absence is not evidence, and names the check that *does* work remotely:

> Read from ComfyUI's generation history over HTTP — NOT from disk. This listing is **INCOMPLETE BY CONSTRUCTION**: `VHS_VideoCombine` and similar video nodes write their file without registering an output entry, so those videos never appear here even though they are on disk… Absence from this list is **NOT evidence** the file is missing. To check a specific file, fetch it by name with `action:"get"` or `upload_image (action:"stage")` — both read the server's `/view` endpoint, which serves straight from the output directory.

Two deliberate choices:

- **The caveat rides on the populated result too**, not just the empty one. The reporter's listing was populated — that is precisely what made it convincing.
- **The local scan inherits none of it.** Hedging an answer that really did read the directory would make a trustworthy result look doubtful.

The tool description carried the same inversion — it promised *"the local filesystem scan is the reliable way to verify the .mp4 exists"* and mentioned the remote fallback separately, so its headline guarantee silently reversed on remote. It now says the guarantee is local-only.

## Also fixed

A test fixture that was lying quietly: the output directory was written `"C:\Comfy\output"` in a JS string, where `\C` and `\o` are not escapes — so it evaluated to `C:Comfyoutput`, a Windows path fixture with no separators in it. Two assertions were pinned to the mangled value.

## Not addressed

The report's secondary point — the tool accepts and ignores params outside its schema, so passing `subfolder: "slot"` produced no error and the reporter believed they had filtered when they had not. That is a schema-strictness question across the whole tool surface, not specific to this tool, and it deserves its own change.

## Verification

- 4 new tests; full suite **321 files / 6922 passed**; `tsc --noEmit` and `check:vocabulary` clean.
- **Mutation-verified**: reverting the history wording to a bare source label kills 2 tests.
- `check:vocabulary` caught `stage_output_as_input` (retired in 0.50.0) in the new user-facing text — fourth catch this session.


artokun:
Make a new pr for that tool

PR #976, opened 2026-08-07
fix(panel-fence): an untrusted hello must not ERASE the tab command stamp (panel #689/#688/#607/#702)

artokun:
Four independent live reports — panel #689, #688, #607, #702 — are one defect. The panel refuses **every** command with `workflow instance mismatch` after a reconnect or workflow switch, including `panel_list_workflows`, and neither `panel_open_workflow` nor `panel_set_workflow_target({mode:"current"})` clears it.

## The wedge is an ABSENT stamp, not a stale one

The hello handler deleted the tab's command stamp whenever it could not derive a trusted identity:

```ts
if (newIdentity) { tabCommandWorkflowUuid.set(panelTab, newIdentity.uuid); }
else { tabCommandWorkflowUuid.delete(panelTab); }   // ← removed
```

`workflowIdentityParts` returns `undefined` when the uuid is missing/malformed **or** the server-observed origin is empty — and a reconnect hello that lands before the canvas identity is readable carries no uuid. So the reconnect itself erases the stamp.

With no stamp, `UiBridge` sends frames with no `workflow_uuid` at all (`else delete frame.workflow_uuid`), and the panel counts an unstamped command as a mismatch too (#718, deliberately fail-closed). **Erasing did not relax the fence — it refused harder**, across everything the fence covers. That is why even `panel_list_workflows` was gated: it is fenced by design ("workflow_list reports the ACTIVE workflow — a stale-stamped call must not answer for the wrong tab").

## Why it was permanent rather than self-healing

The two outcomes differ in *recoverability*. A stale stamp is replaced by the next hello that does resolve an identity. An absent one is not: the panel's re-advertise repair (#607) is capped at `MISMATCH_REHELLO_MAX_PER_IDENTITY = 3` per identity. Once those attempts are spent against an orchestrator that still cannot derive an identity, nothing retries and the tab is refused for the rest of the session.

## Why retaining is safe

The load-bearing claim: **the panel authorizes a fenced command IFF `stamp === the live active workflow uuid`.** So a retained stamp can only ever authorize a command naming the canvas actually mounted right now. It permits no write that a correct stamp would not permit — if the workflow changed, the retained stamp mismatches and is refused exactly as before. An untrusted hello means "no NEW evidence", not "the previous evidence is now false".

This also makes the hello path consistent with the **refresh** path, which already retains on untrusted input — `#716`'s test asserts *"a malformed/late response cannot erase or replace the established stamp."* Same situation, opposite behaviour, now aligned.

`delete(migratedFrom)` on the migration path is a different decision (a tab id going away) and is deliberately untouched — pinned by a test so it cannot be removed as collateral.

## Verification

- New `hello-stamp-retention.test.ts` — 3 tests: the source pin, proof that the untrusted branch is reachable from real reconnect inputs (not a hypothetical), and the migration-delete pin.
- **Mutation-verified**: restoring `delete(panelTab)` fails the pin (`an untrusted hello must NOT erase the stamp`); removing the mutation restores green. My first mutation attempt applied 0 replacements (CRLF), so that pass was discarded and redone.
- `tsc --noEmit` clean; **full suite 322 files / 6921 passed**, no regressions.
- Control-byte scan clean on both changed files; no binary files in `--numstat`.
- No tool-surface change, so the vocabulary ratchet is untouched.

## Not claimed

This removes the mechanism that makes the wedge permanent. Whether it closes every reporter's case needs confirmation on a released build — per the standing rule the four issues stay **open** until someone confirms there.

Refs artokun/comfyui-mcp-panel#689, artokun/comfyui-mcp-panel#688, artokun/comfyui-mcp-panel#607, artokun/comfyui-mcp-panel#702

### DIFF
--- a/src/__tests__/orchestrator/hello-stamp-retention.test.ts
+++ b/src/__tests__/orchestrator/hello-stamp-retention.test.ts
@@ -1,0 +1,84 @@
+// #689/#688/#607/#702 — the panel wedges "workflow instance mismatch" on EVERY
+// command (including panel_list_workflows) after a reconnect or workflow switch,
+// and neither panel_open_workflow nor panel_set_workflow_target({mode:"current"})
+// clears it.
+//
+// The wedge is not a STALE stamp, it is an ABSENT one. The hello handler used to
+// DELETE the tab's command stamp whenever it could not derive a trusted identity
+// from the hello (`workflowIdentityParts` returns undefined when the uuid is
+// missing/malformed OR the server-observed origin is empty — a reconnect hello
+// that lands before the canvas identity is readable carries no uuid). With no
+// stamp, UiBridge sends frames with no `workflow_uuid` at all, and the panel
+// counts an unstamped command as a mismatch too (#718, deliberately fail-closed).
+// So erasing the stamp does not relax the fence — it refuses HARDER, across
+// everything the fence covers.
+//
+// The two outcomes are asymmetric in RECOVERABILITY, which is what made this
+// permanent: a stale stamp is replaced by the next hello that does resolve an
+// identity, while an absent one is not, because the panel's re-advertise repair
+// is capped (MISMATCH_REHELLO_MAX_PER_IDENTITY = 3) per identity. Once those are
+// spent against an orchestrator that still cannot derive an identity, nothing
+// retries and the tab is refused for the rest of the session.
+//
+// Retaining is SAFE, and that is the load-bearing claim these tests defend: the
+// panel authorizes a fenced command IFF stamp === the live active workflow uuid,
+// so a retained stamp can only ever authorize a command naming the canvas that
+// is actually mounted. It permits no write a correct stamp would not permit.
+
+import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
+import { workflowIdentityParts } from "../../orchestrator/session-store.js";
+
+const indexSrc = (): string =>
+  readFileSync(new URL("../../orchestrator/index.ts", import.meta.url), "utf8");
+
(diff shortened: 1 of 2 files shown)


artokun:
**Consumer audit** — `tabCommandWorkflowUuid` is not only the frame stamp, so retaining rather than deleting has to be checked at every reader, not just at the definition. Three consumers:

1. **The frame stamp** (`resolveTabWorkflowUuid` → `UiBridge.send`) — the case the fix is about. Retained ⇒ the panel judges it on equality and refuses on mismatch, exactly as before. Deleted ⇒ unstamped ⇒ refused unrecoverably.

2. **`turnOrigins.recordForMid`** — stamps scope-addressed mutations with the value the turn STARTED with, "so a mid-turn switch to another workflow makes late edits fail the panel's fence loudly instead of silently re-aiming." A retained-but-stale value is stamped, mismatches the live canvas, and is refused **loudly** — which is the documented intent. An absent value is refused too, but as the unstamped case. Equal-or-better, and fail-closed either way; there is no path where a retained stamp re-aims a mutation, because authorization still requires equality with the live active uuid.

3. **`messageOrigin` / `workflowOriginNote`** — the informational "which canvas you are operating on" note, which fires only on a *change*. This is the one place the change is a downgrade: if the workflow genuinely switched while the hello was untrusted, the retained value keeps the origin looking unchanged and the note is suppressed, where deletion would have fired one. Informational only — no routing effect (routing is by tab, not uuid), and any mutation issued under a stale stamp is still refused at the fence.

So the trade is: a possibly-missed origin note in the rare untrusted-hello case, against a permanent total denial of every panel command that four separate live reports hit. Recording it explicitly rather than leaving it to be found later.

PR #1047, opened 2026-08-07
fix(turn-origins): a save's tmp:→wf: rename is one origin, not a mixed batch

artokun:
## What happens

A save renames the tab (`tmp:<uuid>` → `wf:<path>`). Two messages issued from **one**
tab either side of that rename therefore arrive carrying **different ids**. The batch
aggregator counted them as minted, got `tabs.size === 2`, and declared a single-tab turn
mixed-origin — so routing and graph mutations failed closed with:

```
no connected tab can be chosen for "orchestrator::claude": the current turn was
issued from multiple workflows at once, so its target is ambiguous.
```

…naming two entries that are the same workflow.

## Observed live

On released builds — `mcp=0.50.14`, `panel=0.11.44` — a turn opened with two origin
banners:

```
1) "Untitled 2026-08-07 14-21-08"   tmp:95b5c183-…
2) "Untitled 2026-08-07 14-21-08"   wf:1927d99c-…:workflows/Untitled 2026-08-07 14-21-08.json
```

Same title, same workflow, two ids. Four consecutive `panel_*` calls were refused, each
self-reporting `was not dispatched — nothing was applied`.

## The same bug, a few lines up, already fixed

The backend-ownership check immediately above judges on `liveOrigin` — `liveTabIdFor`,
which path-compresses every historical alias onto the newest live id — and its comment
says exactly why:

> A same-socket migration (A→B: a workflow switch, a save, tmp:→wf:) moves the backend
> mapping to B and deletes A, so asking `backendForTab(A)` returns the DEFAULT backend and
> a perfectly healthy Codex turn requeued after such a migration was judged foreign and
> wedged until explicit recovery — **a false refusal, not a leak**.

The tab **count** was left judging on the minted id. Same migration, same false refusal,
one consumer over.

## The fix

Ambiguity is judged on a new `routes` set built from that same `liveOrigin`. The **pin**
still carries the id as recorded, so gate 4's contract is untouched — the bridge resolves
it onward through the alias.

Two sets rather than one, because the two consumers genuinely want different things.
Conflating them broke the existing gate-4 test, which is what surfaced the distinction —
that test is left exactly as it was.

## What stays refused

- **Genuinely distinct tabs** still resolve distinctly, so a real mixed batch still fails
  closed. This is the P0 the gate exists for and it has its own test.
- **Halves that disagree about the workflow uuid**: the tab pins (routing recovers) but the
  stamp is still withheld (mutations stay fenced). Pinning a tab is not a vouch for a
  workflow.

## Tests

35 in the file, **7193 in the full suite**, typecheck and `check:vocabulary` green — the
vocabulary gate is not part of `vitest`, so it is run separately.

Mutation-verified: reverting either size test to the minted-id set fails 3 tests, with
marker counts asserted before and after.

## Relationship to the panel-side fix

Independent, and both are needed. This one stops a false *ambiguity* refusal before
dispatch. artokun/comfyui-mcp-panel#759 stops the *fence* refusing the recovery probe once
a command does reach the panel. The live session hit both in sequence.

Does not close the reports on its own: they close when someone confirms on a released build.

### DIFF
--- a/src/__tests__/orchestrator/turn-origins.test.ts
+++ b/src/__tests__/orchestrator/turn-origins.test.ts
@@ -70,6 +70,75 @@ describe("TurnOriginTracker — pins and stamps land at dequeue (#884)", () => {
     expect(warnings.join("\n")).toMatch(/mixed\/unknown-origin/);
   });
 
+  // A SAVE renames the tab (tmp:<uuid> → wf:<path>), so two messages issued from
+  // ONE tab either side of it carry different ids. Counting the minted ids made
+  // tabs.size === 2 and refused a single-tab turn as "issued from multiple
+  // workflows at once" — naming two entries that are the same workflow.
+  // Observed live on mcp 0.50.14 / panel 0.11.44: two origin banners, both
+  // titled "Untitled 2026-08-07 14-21-08", one tmp: and one wf:.
+  it("a tmp:→wf: SAVE migration is ONE origin, not a mixed batch", async () => {
+    const { tracker, tabAliases } = makeTracker();
+    // The save rewrote tmp:a onto wf:a; the instance uuid is carried across the
+    // swap (the tab is the same instance, only its id changed).
+    tabAliases.set("tmp:a", "wf:a");
+    tracker.recordForMid("m1", "uuid-a", "tmp:a"); // issued before the save
+    tracker.recordForMid("m2", "uuid-a", "wf:a"); //  issued after it
+    tracker.onSeen(KEY, "m1");
+    tracker.onSeen(KEY, "m2");
+    await flushMicrotasks();
+    // Not null — that is the whole point: routing resolves instead of refusing.
+    // The pin keeps the RECORDED identity (the bridge resolves it onward through
+    // the alias), so this asserts the contract gate 4 established, not the live id.
+    expect(tracker.pinOf(KEY)).toBe("tmp:a");
+    // …and the stamp survives, so graph mutations are not fenced out either.
+    expect(tracker.stampOf(KEY)).toBe("uuid-a");
+  });
+
+  it("the migration collapse holds whichever ORDER the two halves dequeue in", async () => {
+    // The post-save message can dequeue first (a re-queue, or simply the order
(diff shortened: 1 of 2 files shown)


artokun:
## Verified end to end on a live rig

All three fixes applied together, then the exact reported sequence run over the loopback panel MCP — **no agent involved**, so nothing could paper over a refusal.

Setup: orchestrator built from this branch (`mcp=0.50.15` + this fix), panel carrying #759 and #761, ComfyUI 0.30.2, a **fresh browser document** (verified by `performance.timeOrigin`, after every patch — a stale module is the standing false-negative on this rig). All three fixes confirmed present in the *running* module, not merely on disk:

```
fix759_predicate        true      fix759_listExempt      true
fix759_pinGuard         true      fix761_unsavedBranch   true
fix759_graphStillFenced true   ← the fence still refuses graph work
```

The tab under test was an **unsaved** workflow — the #761 scenario exactly.

| step | before | after |
|---|---|---|
| `panel_new_workflow` | `created:true`, fence left stale | `created:true`, key `tmp:c2512bcc…` |
| `panel_query_graph` | `workflow instance mismatch` | **success** — 0 nodes |
| `panel_add_node` | `workflow instance mismatch` | **success** — node id `1` |

And confirmed against the live canvas rather than the tool's own reply (the #710 lesson — a handler must be checked against observed effect, not its own report):

```js
activeTitle: "Unsaved Workflow (2)"   // the workflow new_workflow created
nodeCount:   1                        // the CheckpointLoaderSimple genuinely landed
```

Before these fixes the same sequence produced one success and three hard refusals, with `node_count: 0` proving nothing landed.

### One observation, not a claim

The 4th call failed with `no connected tab with id "tmp:10cc"` — my harness pinned a tab id by prefix, and that id migrated to `tmp:c251` when the new workflow was created. Calls 2 and 3 routed fine, so this is my test harness holding a stale pin across a workflow switch, not a reproduction of the reported bug. Whether the loopback MCP's path resolution should consult the same migration aliases `resolveTarget` does is worth a look on its own, and I have not investigated it.

This does not close the reports — they close when someone confirms on a released build.

PR #1102, opened 2026-08-08
fix(manager): a 403 says it is a security gate, not a missing credential

### LINKED ISSUE #1089: Remote update_all returns unhandled ComfyUI-Manager 403
Against remote ComfyUI 0.30.1 at a reachable server, install_comfyui(action:'update_all') called ComfyUI-Manager /v2/manager/queue/update_all?mode=remote and received HTTP 403 Forbidden. The tool returned NODE_MANAGEMENT_ERROR, so all-custom-node update could not proceed. Environment: Linux, Python 3.14.4, PyTorch 2.12.1+cu130, RTX 3090, comfyui-mcp 0.50.25, panel 0.11.38. Please clarify required Manager permissions or provide a supported remote update path.

artokun:
Fixes #1089.

A reporter's remote `install_comfyui(action:"update_all")` got HTTP 403 from `/v2/manager/queue/update_all` and a bare `NODE_MANAGEMENT_ERROR`, then asked us to *"clarify required Manager permissions"*.

**They were missing none.** That 403 is ComfyUI-Manager's *own* security gate: Manager refuses installs/updates/uninstalls when it considers the instance exposed, and a server bound to a non-loopback address (`--listen`, a container, a tunnel) is treated that way by default. No credential this MCP could send would change it — which is precisely why it needs saying, because the obvious reading of 403 is "authenticate" and there is nothing here to authenticate against.

This file already **named** the cause internally: two comments call 403 *"security_level gating"* while deciding it isn't a dialect signal. The knowledge was here; it was just never shown to anyone.

## Where it went

At `managerFetch`'s single non-2xx throw, so **every** Manager operation benefits rather than only `update_all`.

The message names `security_level` as Manager's own `config.ini` setting (explicitly **not** something this MCP controls), explains why it fires on a server that works fine locally, and orders the options: run it over loopback from the host, or update on the host directly, and only then — stated as an exposure tradeoff rather than a recommendation — lower the level, since loosening it lets anything that can reach the port install arbitrary packages.

**Every other status is untouched.** 405 in particular must keep its meaning: this codebase reads it as *wrong dialect* (ComfyUI's frontend catchall answers every unregistered POST with 405), and attaching a security explanation there would send the reader down exactly the wrong path. There's a test asserting the hint is empty for 400/404/405/409/500/502.

## Verification

| Mutation | Result |
|---|---|
| change the 403 guard | 4 of the 5 helper tests fail |
| **drop the call site in `managerFetch`** | killed **zero** helper tests |

That second one is the point. A pure-function test cannot see whether anyone calls the function — so I added a **wiring** test that drives a real 403 through `managerFetch` and asserts the explanation reaches the caller's error. The same mutation now fails it.

Full suite green — 361 files, 7361 passed.

### DIFF
--- a/src/__tests__/services/manager-403-hint.test.ts
+++ b/src/__tests__/services/manager-403-hint.test.ts
@@ -1,0 +1,74 @@
+// #1089 — a bare 403 from ComfyUI-Manager, and a reporter asking what permission
+// they were missing.
+//
+// `install_comfyui(action:"update_all")` hit /v2/manager/queue/update_all and got
+// HTTP 403. The tool returned NODE_MANAGEMENT_ERROR with nothing else, so the
+// reporter asked us to "clarify required Manager permissions".
+//
+// The answer is that they were missing NONE. That 403 is ComfyUI-Manager's own
+// security gate: Manager refuses installs/updates/uninstalls when it considers
+// the instance exposed (a server bound to a non-loopback address — --listen, a
+// container, a tunnel — is treated that way by default), and no credential this
+// MCP could send would change it. That distinction is the entire value of the
+// message, because the obvious reading of 403 is "authenticate" and there is
+// nothing here to authenticate against.
+//
+// This file already NAMED the cause internally — two comments call 403
+// "security_level gating" while deciding it is not a dialect signal — so the
+// knowledge was present and simply never shown to anyone.
+
+import { describe, expect, it } from "vitest";
+import { managerStatusHint } from "../../services/node-management.js";
+
+describe("a Manager 403 explains itself", () => {
+  it("says it is a security gate and NOT an auth failure", () => {
+    const hint = managerStatusHint(403, "/v2/manager/queue/update_all");
+
+    expect(hint).toMatch(/SECURITY GATE/);
+    expect(hint).toMatch(/not an authentication failure/i);
+    // The actionable half of that: stop hunting for a token.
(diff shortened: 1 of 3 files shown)


artokun:
Closing this — it is a duplicate of work already on `main`, and the version there is better than mine.

`explainManagerForbidden()` (shipped for the same #1089) reads ComfyUI-Manager's **machine-readable 403 body** and distinguishes three real causes:

- `security_level` — Manager's own gate, with the levels that permit the route;
- `comfyui_outdated` — ComfyUI predates the system-user API Manager checks *before* it ever consults `security_level`;
- a named install flag — "that flag, not the general security level, is what governs this route".

Mine inferred the cause from the request **path** instead. That is not just weaker, it is wrong in one of those cases: for `comfyui_outdated` it would have told the user to lower `security_level` — a setting change that, in the other implementation's own words, "will not clear this". It also correctly says nothing when the body is not Manager's JSON, rather than attributing a Manager reason to whatever else answered.

I found this the hard way: a `git stash pop` during a branch switch conflicted with the version already on main, I committed the conflict markers without re-running the build, and CI caught it — `TS1185: Merge conflict marker encountered`. My post-commit check grepped that my function was *present* and never that markers were *absent*, which is the check that would have caught it. Worth stating plainly rather than quietly force-pushing over.

No changes needed here. #1089 is handled on main.

Issue #1086, opened 2026-08-08
Remote download_model ignores extra model paths and writes into 20GB overlay


artokun:
### What happened / root cause
The connected remote ComfyUI is configured to use /workspace/models as the persistent model root, while /workspace is the 100GB volume and / is only a 20GB overlay. download_model action:"download" with target_subfolder dispatches to ComfyUI-Manager, which writes into the default /opt/ComfyUI/models instead of the configured /workspace/models path.

This caused the large Wan 2.2 high-noise file to fail/corrupt and a required clip vision file to be missing after pod restart.

### Steps to reproduce
1. Connect comfyui-mcp to the remote RunPod ComfyUI.
2. Configure extra model paths so ComfyUI lists models from /workspace/models.
3. Call download_model action:"download" with target_subfolder:"clip_vision" or "diffusion_models".
4. Inspect the remote ComfyUI logs and model roots.

### Exact evidence
get_system_stats action:"logs" returned:
"Install model 'clip_vision_h.safetensors' ... into '/opt/ComfyUI/models/clip_vision/clip_vision_h.safetensors'"
The user filesystem reports:
- overlay /: 20G total
- /workspace/models: 81G used on the persistent volume
The high-noise file was absent from /workspace/models after the transfer was reported done.
health reported clip_vision EMPTY while diffusion_models listed names.

### Fix
Applied locally: no. Upstream-only report. Remote Manager downloads need to honor the live ComfyUI extra model roots, or download_model needs an explicit verified destination root for remote targets and must not report accepted Manager tasks as landed.

### Environment
OS: Linux remote RunPod
GPU: NVIDIA H100 PCIe 79GB
ComfyUI: 0.28.0
comfyui-mcp: 0.50.23
panel: 0.11.44

Related (same root cause): #143, #633, #369

artokun:
Confirmed as a real gap, and **not** covered by the #369/#794 destination fix — you are on 0.50.23 and that shipped in 0.49.4, so you have it and this still happens. Different path, different cause.

## What #794 fixed, and why it does not reach this

#794 made **our own** download path resolve its destination from the live server rather than a stale `COMFYUI_PATH`. Your failure is on the **Manager dispatch** path: `download_model` hands the job to ComfyUI-Manager, which runs on the remote box and picks its own root. Your evidence shows Manager making that choice, not us:

```
Install model 'clip_vision_h.safetensors' ... into '/opt/ComfyUI/models/clip_vision/...'
```

[...]

PR #1110, opened 2026-08-08
feat(ci): gate the "could not determine" → "determined not" collapse

artokun:
Refs #796

The tracked defect class, in one line: **a state meaning "I could not OBSERVE this" is folded into a state meaning "I observed that it is NOT SO."**

Twenty-seven instances have been fixed by hand over two months — a pack reported "not installed" by code that never reached the disk, a port reported free by a lookup that failed, a download reported verified after an unverified server swap. Every one had tests. Every one tested a real negative rather than a *failed observation*, which is exactly why the tests passed.

#796 asks for two things and this lands both: a review-checklist line in `CONTRIBUTING.md`, and a lint for the shapes that are actually greppable.

## What it does *not* flag matters more than what it does

```ts
await client.close().catch(() => {});   // NOT this defect — nothing reads it
```

That catch exists so a cleanup failure cannot mask the real error on the way out. On this repo that shape is **40 of 99 raw matches**, so a gate that cried about them would be switched off within a week. A match counts only when the value is **consumed**: bound, returned, chained onward, or passed as an argument.

Most of the test file is that negative half — proof the noisy shapes stay quiet, rather than a claim that they do.

### The discriminator is trailing punctuation, not leading text

```ts
await load().catch(() => null),        // an argument  — consumed
await rm(tmp).catch(() => undefined);  // a statement  — discarded
```

Both open with `await`, so what comes *before* cannot separate them. A terminating `;` on a statement that binds nothing is the one shape where the value provably goes nowhere.

## The ratchet runs downward

The baseline is **debt, not history** — 38 pre-existing sites, and the file may only **shrink**. Fixing a site means deleting its line, and the gate fails if a listed site no longer exists, so the list cannot rot into standing pre-approval for a re-introduced defect.

This is deliberately the opposite direction from `docs/design/tool-surface.txt`, which is append-only history.

## The escape hatch is the point

```ts
// unknown-ok: <why a failed observation and a genuine negative should lead to
// the same behaviour at THIS call site>
```

A **bare** `unknown-ok` is rejected — the reason is the deliverable, and writing it is where an author notices the collapse is not safe. A wrapped reason counts in full, so nobody shortens an explanation to fit a linter. A marker above the enclosing *function* does not reach the site.

Two waivers land with it, both verified by reading the code:

- `training-jobs` — `null` is the UNKNOWN state, deliberately distinct from a probed `false`, and the branch below acts on that distinction
- `queue-manager` — a failed poll is recorded as a failed poll and carried into the returned `continuous` flag

Those two were **spot-checks** of the baseline, and both turned out already correct. That suggests the remaining 36 are debt to *review* rather than debt to *fix* — they have not been read, and this PR does not bless them.

## Verification

| mutation | result |
|---|---|
| a real new collapse inserted into `queue-manager.ts` | exit 1, named the line |
| the consumed-value test disabled | 38 sites becomes **97** |
| a waiver reason stripped to a bare marker | exit 1, named the line |

Cross-checked both directions against a naive grep: **zero** false negatives versus every single-line binding/return form, and the scanner finds 4 more (multi-line chains and argument positions) that grep misses.

20 gate tests · full suite **7450 passed | 2 skipped** · `tsc`, `lint`, `check:vocabulary`, `check:unknown-collapse` all exit 0 · control-byte scan clean on all 8 changed files.

Left open deliberately: this catches the common *written form*, not the *shape*. `existsSync` guarding a meaningful read is in the issue's list and is not greppable without type information — the CONTRIBUTING line carries that half.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,15 @@ jobs:
       # becomes real once the matrix actually varies the OS — comfyui-mcp-bi9.)
       - run: npm run check:vocabulary
 
+      # #796 — the "could not determine" -> "determined not" collapse. A caught
+      # failure answered with [] / null / false tells the caller a negative was
+      # OBSERVED, and the caller then reports it to a user as fact. Twenty-seven
+      # instances were fixed by hand before this gate existed; each had tests, and
+      # each tested a real negative rather than a failed observation. Ratchets down
+      # only — a site is cleared by fixing it or by a reasoned `unknown-ok:` at the
+      # line, never by appending to the baseline.
+      - run: npm run check:unknown-collapse
+
       # docs/design/tool-vocabulary.json is what comfyui-mcp-panel VENDORS to
       # validate its callTool("…") string literals. If the vocabulary changes here
       # and the artefact does not, the panel keeps checking against the old surface
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,6 +115,17 @@ install stays lean.
 - Keep PRs focused; include tests for behavior changes.
 - **PR checklist:** `npm run build` ✓ · `npm test` ✓ · docs regenerated if tools changed ✓ ·
   a clear description of what and why.
+- **When you add or touch a predicate, ask what it returns when the observation
+  FAILS.** If that is the same value it returns for a genuine negative, you have
+  written the most common defect in this codebase: "I could not determine this"
+  collapsing into "I determined it is not so." The user is then told a confident
+  wrong answer — a pack reported "not installed" by code that never reached the
+  disk, a port reported free by a lookup that failed, a download reported
+  verified after an unverified server swap. Give unknown its own representation
+  (`yes | no | unknown`, or a tagged result), and **test the failed observation,
+  not just the negative answer** — every instance found so far had tests, and
+  none of them tested the probe failing. `npm run check:unknown-collapse` catches
+  the common written form; it cannot catch the shape, only you can.
 
 Open a GitHub issue first for large or potentially-breaking changes so we can align on the approach.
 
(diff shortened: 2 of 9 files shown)


artokun:
Second commit — two detector defects, both found by pointing the gate at real code and asking why a site I *expected* was absent.

I was reading `src/tools/model-management.ts` for #1086 and grepped the baseline for it. It wasn't there.

**Missed: a catch appended to a wrapped call's closing paren.**

```ts
const managerSeen = job.viaManager
  ? await verifyManagerVisibility(
      job.target_subfolder,
      job.filename,
      { attempts: 1 },
    ).catch(() => undefined)
  : undefined;
```

Bare `)` before the match, nothing after it — both heuristics read it as a discarded statement. That site is the #1086 download verification, the most safety-critical instance in the repo and precisely the shape this gate exists for. The walk back to the statement start is now bracket-balance aware.

**False positive introduced by that fix: a `return` inside a callback.**

```ts
fetchIt(url)
  .then((r) => {
    if (!r.ok) return;      // belongs to the CALLBACK
    push({ ready: true });
  })
  .catch(() => {});
```

Walking to the statement start now sweeps up the callback body, and that `return` turned a fire-and-forget readiness probe into a reported defect. Consumption is now judged from the line the statement *begins* on rather than everything swept up reaching it — simpler than what it replaces. The waiver lookup had the same locality bug and is fixed the same way: a reader writes the explanation above the **statement**, not above whichever fragment the match landed on.

`model-management` is **waived, not baselined** — `undefined` there is the could-not-verify state, kept distinct from both `visible` and `not-listed`, and the render never upgrades a missing verification into a confirmation. Verified the waiver is what excludes it rather than a detector regression: removing the marker brings the site back (1 → 0 → 1).

**Net effect on the repo: the same 38 sites. None dropped, none added** — `model-management` is now found *and* waived, and the orchestrator probe is correctly silent. Diffed the site list, not just the count, at every step; twice a stable count was hiding a swap.

22 gate tests · full suite **7452 passed | 2 skipped** · all four gates exit 0.

PR #1124, opened 2026-08-08
feat: attach UI workflow metadata to API-enqueued prompts

kamekazu:
## Why

Prompts submitted through the ComfyUI API currently retain only the API graph.
As a result, ComfyUI History and Media Assets cannot use “Open as workflow” for
generations created by comfyui-mcp.

## What changed

- Generate a UI graph from the final post-seed-randomization API graph in the
  shared enqueue service.
- Nested-merge it into `extra_data.extra_pnginfo.workflow`.
- Preserve a valid caller-provided workflow and all existing `extra_data`.
- Reuse the existing `/object_info` TTL cache and single-flight request.
- Log conversion warnings and failures, then fail open so metadata generation
  can never block or duplicate the original enqueue.

All generation paths continue to use the same shared enqueue service. No
tool-specific changes were added.

## Validation

- 7,483 tests passed (2 skipped)
- TypeScript typecheck and build passed
- Representative rerun, template, URL, image, and video paths passed
- Live ComfyUI smoke retained existing `extra_data` and stored a valid
  2-node / 1-link UI graph in History
- H3 8-node conversion timing:
  - cold: 839.30 ms
  - warm median: 0.08 ms
  - warm max: 0.19 ms

### DIFF
--- a/src/__tests__/services/workflow-executor.test.ts
+++ b/src/__tests__/services/workflow-executor.test.ts
@@ -357,3 +357,31 @@ describe("enqueueWorkflow UI metadata", () => {
     );
   });
 });
+
+// A non-object extra_pnginfo cannot be merged into, and spreading it would
+// DISCARD whatever the caller put there. Attaching UI metadata is an
+// enhancement; dropping a field the caller set would be a regression.
+describe("enqueueWorkflow UI metadata — a field we cannot merge into is left alone", () => {
+  it("does not drop an ARRAY extra_pnginfo to make room for the workflow", async () => {
+    const extraData = { extra_pnginfo: ["caller", "data"], client_id: "c1" };
+
+    await enqueueWorkflow(
+      { "1": { class_type: "ClaudeNode", inputs: { prompt: "hi", seed: 7 } } },
+      { disable_random_seed: true, extra_data: extraData },
+    );
+
+    expect(enqueuedExtraData()).toEqual(extraData);
+    expect(enqueuedExtraData()!.extra_pnginfo).toEqual(["caller", "data"]);
+  });
+
+  it("does not drop a STRING extra_pnginfo either", async () => {
+    const extraData = { extra_pnginfo: "opaque" };
+
+    await enqueueWorkflow(
+      { "1": { class_type: "ClaudeNode", inputs: { prompt: "hi", seed: 7 } } },
+      { disable_random_seed: true, extra_data: extraData },
+    );
+
+    expect(enqueuedExtraData()).toEqual(extraData);
+  });
+});
--- a/src/__tests__/services/workflow-executor.test.ts
+++ b/src/__tests__/services/workflow-executor.test.ts
@@ -23,6 +23,7 @@ import {
   seedInputRange,
 } from "../../services/workflow-executor.js";
 import type { ObjectInfo, WorkflowJSON } from "../../comfyui/types.js";
+import { logger } from "../../utils/logger.js";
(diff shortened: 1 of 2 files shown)


artokun:
Reviewed — this is good work, and it's ready. Sorry it sat; I've pushed two commits to your branch so it can land in one piece rather than costing you another round-trip.

**1. Merged current `main`.** Your checks were green, but they ran against a `main` from ~8 hours and four releases earlier (0.50.36 → 0.50.40 landed in between). A green check is only green for the `main` it ran against, so I re-verified locally after the merge: clean merge, no conflicts, **372 files / 7,549 tests passing**, typecheck and the vocabulary gate clean.

**2. One edge case: a non-object `extra_pnginfo` was being dropped.**

`isRecord(extraData?.extra_pnginfo) ? … : {}` falls to `{}` for an array or a string, and the spread then replaces the caller's value. I probed it against your code:

```
extra_data: { extra_pnginfo: ["a","b"] }
  → sent: { extra_pnginfo: { workflow: {…generated…} } }     // "a","b" gone
```

Not a shape anyone should be sending, but attaching UI metadata is an *enhancement* and it shouldn't buy itself room by discarding a field the caller set. Such a request now goes out exactly as it came, with a debug line saying why no metadata was attached. Two tests, mutation-checked in both directions (removing the guard, and widening it so the feature never attaches).

## What I checked, and what held up

- **Single choke point.** Your claim that all generation paths share the enqueue service is correct — `enqueueWorkflow` is the only caller of `enqueuePrompt` that builds `extra_data`, and `requeuePendingJob` re-sends the extra_data a job already carries, so re-queues keep the metadata rather than losing or double-generating it.
- **Fails open, verifiably.** With `getObjectInfo` rejecting, the prompt still enqueues **once**, with the caller's `extra_data` untouched, and returns its `prompt_id`. Your test pins it and my independent probe agreed.
- **Cannot hang.** My one real worry was that `disable_random_seed: true` now adds an `/object_info` fetch to a path documented as returning immediately. It's bounded by `comfyHttpTimeoutSeconds()`, so worst case is one timeout and then the fail-open — fine.
- **No second fetch in practice.** The default path calls `getObjectInfo` twice, but the second is a TTL cache hit; only a mock without the cache shows two round-trips.
- **Converter is proven ground.** `convertApiToUi` already backs a shipped `workflow-library` path and has ~100 dedicated tests including a full `convertUiToApi(convertApiToUi(x)) === x` round-trip. Reusing it was the right call over anything bespoke.
- **A valid caller canvas is authoritative**, and you skip the `/object_info` fetch entirely in that case. That ordering is exactly right — it keeps the caller's real layout and provenance instead of a synthesized one.

The thing I appreciated most: generating the UI graph from the **post-seed-randomization** graph, and testing that the sampler's `widgets_values` carries the randomized seed. "Open as workflow" reproducing a *different* seed than the one that ran would have been a subtle and very annoying bug, and you closed it before anyone hit it.

Merging once CI comes back green. Thanks for this — History "Open as workflow" on MCP-generated outputs is a real gap closed.

artokun:
Merged, and shipping in **0.50.41** — thanks @kamekazu. History "Open as workflow" now works for anything comfyui-mcp generates.

For anyone finding this later, the design decision worth copying: the UI graph is derived from the **post-seed-randomization** API graph, so the reopened workflow carries the seed that actually ran rather than the one that was requested. That one is easy to get wrong and would have been very hard to notice.

Issue #884, opened 2026-08-05
P0: agent sessions are workflow-scoped — context is lost on Workflow → New, on switching workflows, and across tabs


artokun:
Reported by **seanmcmagic** in Discord `#help`:

> I'm noticing that every time the agent does "Workflow_New" it loses all of its context

## Not a bug — but a real gap

This is the per-workflow session design working correctly, so filing as an enhancement rather than a defect.

The panel sends a `tab_id` that *is* the workflow identity (`comfyui-mcp-panel.js:1871` `workflowTabId()`): `wf:<path>` for a saved workflow, `tmp:<uuid>` for an unsaved one. The orchestrator keys the agent on that plus the backend (`orchestrator/index.ts:1430` `agentKeyFor`). So each workflow owns its conversation, which is deliberate and prevents cross-workflow contamination.

`Workflow → New` creates a fresh workflow object → fresh `tmp:<uuid>` → different session key → new agent with no history.

Notably the identity layer **already distinguishes continuity from forking** (`web/js/lib/workflow-chat-identity.js:98-113`): Save-As / rename mutate the same live object so "continuity wins" and context survives, whereas copied JSON resolving to a different path forks to a fresh identity. So the concept of *"this new file should keep the old conversation"* is already implemented and load-bearing — it's just not reachable from `New`.

## The gap

There's no explicit **"new workflow, same conversation"** affordance. A user mid-conversation ("ok, now build me a variant of this in a blank graph") has to either stay in the current workflow or re-explain everything. The workaround — use Save-As instead of New — works today but is entirely non-obvious; nothing in the UI hints that one preserves context and the other doesn't.

## Suggestions

Roughly in order of cost:

1. **Signpost it.** When a new session starts because the workflow changed, have the agent's first message say so — one line, e.g. "New workflow, so this is a fresh conversation. Save-As from the previous workflow instead if you wanted to keep that context." Cheap, and converts a confusing silence into a documented behavior.
2. **A "carry context" option on New** — e.g. `New (keep chat)`, which mints the new workflow identity but seeds it from the previous session, reusing the same continuity path Save-As already takes.
[...]

Issue #1133, opened 2026-08-08
ComfyUI Desktop: root corroboration refuses on a Desktop argv, and restart_comfyui declines — no in-tool path to complete a panel update


artokun:
### What happened / root cause

On a **ComfyUI Desktop** install with panel 0.9.8 (below the 0.11.35 that mcp 0.50.36 requires for graph writes), every documented route to update the panel dead-ends. The agent can read the graph but cannot write to it, and cannot self-heal.

The three failures compound:

**1. `install_comfyui(action:'panel', panel_action:'update')` misreports a *staged* update as a *failed* one.**

The tool returned:

> Panel update did NOT apply (ComfyUI-Manager reported the queue drained without moving the pack on disk), and the reinstall-from-source fallback is REFUSED …

But the ComfyUI log shows Manager did **not** no-op — it deferred:

```
[2K Updating: comfyui-agent-panel
Installation reserved: comfyui-agent-panel
[INFO] After restarting ComfyUI, please refresh the browser.
```

`Installation reserved` is ComfyUI-Manager staging the update to apply at next **startup**, which is expected Manager behaviour for a pack whose files are in use. The pack legitimately has not moved on disk yet, so the "queue drained without moving the pack" heuristic fires — but the correct report is **"staged — restart to apply"**, not "did NOT apply". As written it sends the agent down the reinstall-from-source path for a problem that a plain restart would have resolved.

This is not a one-off: the same reserve appears **four times** in the log (18:20:11, 18:20:28, 18:20:50, 18:27:57), i.e. repeated update attempts each re-reserving and each reported as a failure, with nothing ever applying.

**2. The reinstall-from-source fallback refuses on a resolvable root-corroboration check.**

> the running ComfyUI answered but its reported launch arguments did not identify an install root holding custom_nodes (a relative main.py and no reported working directory cannot be resolved) … this is the CONFIGURED path rather than a corroborated one. On a split install (Comfy Desktop's --base-directory, #766) those differ…

The caution is sound in principle, but here it was decidable from evidence the tool already had:

- `server_argv` contains **no `--base-directory`** — the only flag that creates the split this guard defends against. It does contain `--input-directory` / `--output-directory` (which relocate media, not `custom_nodes`).
- The machine has exactly **one** `custom_nodes` tree and exactly **one** panel pack inside it.
- The running server's own log names `comfyui-agent-panel` as the pack Manager is updating, tying the running process to that tree.

[...]

PR #1151, opened 2026-08-08
spec: mobile reattach — decouple the tunnel's lifecycle from the orchestrator's restart

artokun:
Draft ask for handoff — not scoped for build. Full spec: `docs/design/mobile-tunnel-reattach.md`.

> **Re-verified against `origin/main` on 2026-08-14.** The spec is long-lived and the repo is under a feature freeze, so it was re-read against today's source rather than implemented. One claim was wrong and is corrected in place; one open question is now answered. Details at the bottom.

## Origin

artokun, Discord `#help`: *"Can we hold the connection separate from the orchestrator restart group that way we can reattach between updates?"* / *"It really is annoying to lose your connection on mobile."*

## Why `#875` (shipped) isn't the final answer

`#875` shipped in **three** parts, not one:

- `65058e2` (#1113) — the pairing **token** is now persisted to `~/.comfyui-mcp/pair-token`, so a **LAN** pairing URL already survives a restart.
- `1764e2a` (#1020) — `pair-durability.ts` discloses at pair time which parts of the URL rotate.
- `30a838b` (#1115) — the **deferral**: the self-restart is postponed while a phone has a live tunnel connection.

The deferral is a real fix for the reported symptom, but updates never land while the phone stays connected. Fine for a short session; worse for someone leaving a tunnel open at work all day.

Critically, with the token now stable, the **quick-tunnel hostname is the last remaining rotating part** of a tunnel pairing URL. `pair-token-store.ts` says so itself — *"This is HALF the fix … a stable token makes the LAN URL durable and a tunnel URL still rotates."* This spec is the other half, which makes it **narrower** than when it was written, not obsolete.

## The alternative

Let updates apply promptly. Give the phone a **stable hostname** to reconnect to instead of a dead one — the reconnect becomes a brief blip rather than either a silent drop or an indefinitely-deferred update. `#884` (orchestrator-scoped sessions, shipped) is what makes this worth doing: a phone reconnecting to the same hostname resumes the same conversation, no re-pairing.

## What's coupled today (re-verified 2026-08-14)

There are **two independent quick-tunnel call sites**, and this ask is about the first:

1. **Phone pairing** — `src/orchestrator/index.ts`, in the `pair` handler: `startQuickTunnel(pairPort, "127.0.0.1")` cached in `pairTunnel`, torn down by its own `process.once("exit", …)` right there.
2. **The `connect` path's remote-pod bridge** — `src/services/secure-bridge.ts` → `setupCloudflaredBridge()`. Same helper, different consumer, its own teardown.

Both go through `src/services/tunnel.ts` → `startQuickTunnel()`, which spawns via the `cloudflared` package's `Tunnel.quick()` — and that exposes **no detach option**, so "spawn detached" is a rewrite of the spawn, not a flag.

The deferral itself is **not** in `self-restart.ts` (which knows nothing about tunnels). It is the single `!tunnelPairingLive()` clause in the orchestrator's `allIdle` gate, where `tunnelPairingLive()` requires *both* a tunnel and a client connected right now.

## The one thing that can't be avoided

cloudflared tunnels TCP, not the WebSocket session — the live connection still dies the instant the old process exits. What survives is the hostname. Stated plainly in the spec so nobody scopes a build expecting a connection that survives untouched.

## Two open decisions (owner's call)

- Does this replace `#875`'s deferral, or layer on top of it? (Mechanically a one-clause edit either way.)
- How does the new process validate a detached tunnel it's inheriting is still healthy, vs. a stale one surviving a genuine crash?

## Explicitly out of scope

The relay-backend path is unaffected — and per `pair-durability.ts` it is not even *reachable* from the pairing path, so it can't be offered as a workaround. A real named-cloudflared-tunnel integration was raised as a separate, larger alternative and is not scoped here.

---

## What the 2026-08-14 re-verification changed

- **CORRECTED — wrong file.** The draft named `src/services/secure-bridge.ts` as "what tears the tunnel down today", under a heading claiming it was verified against source. That's the `connect`-path remote-pod bridge, **not** phone pairing. The pairing tunnel has lived in `src/orchestrator/index.ts` since `5129663` (#180) — so the claim was wrong *when written*, not drifted. A build scoped from that sentence would have edited a file the phone never touches.
- **UNDER-CREDITED — `#875` scope.** The draft credited it with the deferral only; the persisted token and the pair-time disclosure also shipped under it.
- **ANSWERED — open item 4.** `comfyui-mcp-mobile`'s `bridge_client.dart` already re-dials the same URL after an unintentional drop (its `#37`): 1/2/4/8/16/30 s backoff with a probation-gated reset. No new retry mechanism is needed — **but** it is capped at `_maxRetries = 6` (~1 min), and that cap is justified *by* the rotating hostname. Make the hostname stable and the cap becomes the new binding constraint, so revisiting it moves into scope.
- **Still accurate:** the TCP/WebSocket limitation, the `#884` session-resume premise (`~/.comfyui-mcp/sessions`), and the relay exclusion.
- **Nothing in "shape of the change" is implemented, and nothing is rejected.** No detached cloudflared spawn exists anywhere in `src/`, and there is no durable tunnel record.

Refs #875

### DIFF
--- a/docs/design/mobile-tunnel-reattach.md
+++ b/docs/design/mobile-tunnel-reattach.md
@@ -2,42 +2,73 @@
 
 **Status:** draft ask, not scoped for build. Two open decisions below, both owner's call.
 **Origin:** artokun, Discord `#help`, 2026-08-08 — "It really is annoying to lose your connection on mobile" / "Can we hold the connection separate from the orchestrator restart group that way we can reattach between updates?"
-**Related:** #875 (shipped, `30a838b` — defers the self-restart while a phone is connected), #884 (orchestrator-scoped sessions, shipped — the mechanism this ask leans on)
+**Related:** #875 (shipped, in three parts — see below), #884 (orchestrator-scoped sessions, shipped — the mechanism this ask leans on)
+**Re-verified against source:** 2026-08-14, against `origin/main`. Every "verified" claim below was re-read on that date. **One claim in the original draft was wrong and is corrected in place** — see "What's coupled today".
 
 ## The problem, as it stands today
 
-The default mobile transport is a cloudflared **quick tunnel**: no Cloudflare account required, URL is `wss://<random>.trycloudflare.com`, a fresh random hostname **every process run**. `#875` shipped a fix for the resulting disconnect, but it is a *deferral*: the hourly self-restart update check still runs, and while a phone has a live tunnel connection the restart is postponed until the phone disconnects. Updates queue up and land in one shot next time you disconnect.
+Phone pairing offers two transports: `lan` (`ws://<lan-ip>:<pairPort>/?token=…`, same wifi) and `tunnel` (anywhere). The tunnel option is **always** a cloudflared **quick tunnel** — no Cloudflare account required, URL is `wss://<random>.trycloudflare.com`, and the hostname is minted fresh **every process run** with no way to pin it.
 
-That is a real fix for the reported symptom (a background update silently killing an in-use session), but it trades "you get disconnected without warning" for "you never get updated while connected" — fine for a short session, worse for someone leaving a tunnel open at work all day, since it means updates never land until they explicitly step away.
+`#875` has shipped, and it shipped more than the "deferral" this draft originally credited it with. Three separate things landed:
+
+- **`65058e2` (#1113) — the pairing token stopped rotating.** It is now persisted to `~/.comfyui-mcp/pair-token` (mode 0600, write-then-rename) instead of being minted per run — `src/orchestrator/pair-token-store.ts`. A **LAN** pairing URL therefore already survives a restart with no configuration.
+- **`1764e2a` (#1020) — pair-time disclosure.** `src/orchestrator/pair-durability.ts` tells the user, at the moment the URL is minted, which parts of it rotate on a restart and what to do about it.
+- **`30a838b` (#1115) — the deferral.** While a phone is connected over a tunnel, the automatic update-restart is postponed until it disconnects.
+
+**This is what makes the ask below narrower than it looks.** With the token stable, the quick-tunnel **hostname is the last remaining rotating part** of a tunnel pairing URL. `pair-token-store.ts` says so in its own header: *"This is HALF the fix … a stable token makes the LAN URL durable and a tunnel URL still rotates."* This document is the other half.
+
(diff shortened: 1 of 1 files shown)


artokun:
### Item 4 answered from the mobile source: **yes, this needs mobile-side work.**

The spec said *"check what `comfyui-mcp-mobile`'s `bridge_client.dart` does today on an unexpected close before assuming this needs new mobile-side work."* Checked — and the assumption holds, so that open item can be closed as a known cost rather than a maybe.

**There is no reconnect anywhere on the mobile side.** `lib/core/bridge/bridge_client.dart:102-109`:

```dart
onError: (Object _) { _setStatus(BridgeStatus.error);        _failPending(const BridgeConnectionLost()); },
onDone:  ()          { _setStatus(BridgeStatus.disconnected); _failPending(const BridgeConnectionLost()); },
```

Status change and pending-call failure. No timer, no backoff, no retry. And `chat_controller.dart:469-486` is a *teardown* handler, not a recovery one — on `error`/`disconnected` it clears in-flight option switches, drops live queue state, cancels the render probe. Its comments consistently say *"a reconnect re-seeds it on hello"*, which describes what happens **when** a reconnect occurs, not anything that causes one. No lifecycle-driven reconnect either: nothing in `app_lifecycle.dart` reconnects, and there is no foreground/resume retry path anywhere in `lib/`.

So today a drop is terminal until the user acts. That has two consequences for this spec:

1. **The mobile work is not optional polish — it is half the feature.** A stable hostname the app never retries is worth nothing to the user; they still see a dead session. Points 1–3 (detached spawn, skip `stop()` on the self-restart path, discover-and-adopt on startup) deliver *nothing observable* without point 4.
2. **It sequences the other way round from how the spec reads.** The mobile retry is independently useful *before* any orchestrator change lands: with `#875`'s deferral shipped, a phone that reconnects on drop already recovers from ordinary network blips, backgrounding and Wi-Fi/LTE handoff — none of which involve a restart. That is a smaller, lower-risk change with its own payoff, and it de-risks the tunnel work by making the client side already proven when the hostname becomes stable.

**Suggested revision to the spec:** move point 4 out of "shape of the change" and into its own first step, noting it stands alone. That also softens open question 1 — if the client retries, deferral (`#875`) and prompt-update-plus-reattach stop being alternatives, because the reconnect blip is handled either way.

The two open questions remain owner's calls and I have not filled them in. Question 2 in particular (a detached tunnel surviving a **crash**, and the next process needing to validate it before trusting it) is the part I would treat as the real risk — a stale tunnel silently adopted is worse than today's clean slate, and it is the same shape as the corroboration problem in #1133: absence of evidence must not read as evidence of health.

PR #1185, opened 2026-08-09
fix(restart): a BACKGROUNDED phone is not a departed phone

### LINKED ISSUE #1176: #875's restart deferral doesn't hold for a BACKGROUNDED phone — 'live socket' is the wrong liveness signal for a mobile client
Reported by **seanmcmagic** in Discord `#help`. This is a confirmed gap in `#875`'s deferral fix, not a recurrence of the original bug.

## The report

Phone disconnected from the mobile tunnel at work, ~12:30–1:00pm. Confirmed version movement: **0.50.39 → 0.50.41**. Both are well past **0.50.35**, where the deferral fix (`30a838b`) shipped — so the gate was present and still let the restart through.

Screenshot from the phone after the drop:

```
WebSocketChannelException: SocketException: Failed host lookup:
'cameron-timing-face-spies.trycloudflare.com'
(OS Error: No address associated with hostname, errno = 7)
```

DNS failure, not a timeout or refusal — the hostname no longer exists, which is what a rotated quick-tunnel looks like from the client side.

## Root cause: "live" is the wrong liveness signal for a phone in a pocket

The gate (`src/orchestrator/index.ts:1075`):

```ts
const tunnelPairingLive = (): boolean => pairTunnel !== null && bridge.hasLiveHeadlessClient();
```

and (`src/services/ui-bridge.ts:2757`):

```ts
hasLiveHeadlessClient(): boolean {
  for (const c of this.conns.values()) if (c.headless === true) return true;
  return false;
}
```

`this.conns` holds **currently-open sockets only**. A phone at work — screen off, app backgrounded — has its WebSocket suspended or closed by the mobile OS. So `hasLiveHeadlessClient()` correctly returns `false`, `tunnelPairingLive()` goes false, the restart gate legitimately opens, the process restarts, and cloudflared mints a new random hostname. The user picks up their phone and finds a dead URL.

**Nothing here is behaving incorrectly by its own contract.** The accessor's own doc comment shows the tradeoff was considered deliberately:

> Deliberately NOT `isHeadless()`, which is STICKY by design… using it would report a phone that paired once and left as still connected, and the self-restarter would defer updates forever.

That reasoning is sound. But it leaves a hole between the two options:

| signal | behavior | problem |
|---|---|---|
| `isHeadless()` (sticky) | true once a phone ever paired | defers updates forever after a single pairing — correctly rejected |
| `hasLiveHeadlessClient()` (live) | true only while a socket is open | a backgrounded phone reads as "gone" — **this bug** |

The missing concept is a phone that is **paired and intended to be in use, but not holding a socket at this instant**. That is the normal state of a mobile client, not an edge case — a phone in a pocket is the *expected* condition for someone who paired it to use at work.

## Possible directions (not prescribing one)

- **Bounded recency window** — defer if a headless client was connected within the last N minutes. Keeps the "don't defer forever" property that killed the sticky option, while covering a backgrounded phone. Needs N chosen deliberately.
- **Explicit pairing lifetime** — treat a pairing as live until explicitly unpaired or a long idle timeout, rather than inferring from socket state.
- **Make it moot structurally** — `#1151` (spec, filed) decouples the tunnel's lifecycle from the restart so the hostname survives. A backgrounded phone would then reconnect to a hostname that still exists, regardless of whether a restart happened. **This report is direct evidence for that spec**: the deferral approach has to correctly guess "is the user still using this?", and the reattach approach never has to ask.

## Environment

Mobile app over cloudflared quick tunnel, published npm orchestrator, auto-update on. Reporter's orchestrator self-updated 0.50.39 → 0.50.41 during the disconnect window.

artokun:
Fixes #1176. **Draft — claiming the issue, work in progress.**

## The gap

`#875`'s deferral gate asks `hasLiveHeadlessClient()`, which reads `conns` — currently-open sockets only. A phone at work with its screen off has had its WebSocket suspended or closed by the mobile OS, so the gate correctly reads "no headless client", opens, and the restart mints a new cloudflared hostname. The user picks up their phone and finds a dead URL:

```
WebSocketChannelException: SocketException: Failed host lookup:
'cameron-timing-face-spies.trycloudflare.com' (errno = 7)
```

Nothing misbehaves by its own contract — and the accessor's doc comment shows the sticky alternative was rejected for a good reason (it would defer updates forever after a single pairing). The hole is between the two:

| signal | behavior | problem |
|---|---|---|
| `isHeadless()` (sticky) | true once a phone ever paired | defers forever — correctly rejected |
| `hasLiveHeadlessClient()` (live) | true only while a socket is open | a backgrounded phone reads as gone — **this bug** |

The missing concept is *paired and in use, but not holding a socket right now* — which is the **normal** state of a mobile client, not an edge case.

## Plan

Bounded recency window: defer while a headless client was connected within the last N minutes. Keeps the "don't defer forever" property that killed the sticky option, while covering a pocketed phone. N chosen deliberately and stated in the code.

Not attempting #1151's structural fix here (decoupling the tunnel lifecycle so the hostname survives a restart) — that is the better long-term answer and this report is evidence for it, but it is a different change.

### DIFF
--- a/src/__tests__/services/headless-recency-window.test.ts
+++ b/src/__tests__/services/headless-recency-window.test.ts
@@ -24,7 +24,9 @@ function bridgeWithHeadlessGoneFor(ms: number | null): UiBridge {
   const b = new UiBridge() as UiBridge & {
     markHeadlessDisconnectForTests(at: number): void;
   };
-  if (ms !== null) b.markHeadlessDisconnectForTests(Date.now() - ms);
+  // Monotonic, matching the field — the recency window measures ELAPSED time,
+  // which the wall clock does not (codex review: an NTP step could skip it).
+  if (ms !== null) b.markHeadlessDisconnectForTests(performance.now() - ms);
   return b;
 }
 
--- a/src/__tests__/services/headless-recency-window.test.ts
+++ b/src/__tests__/services/headless-recency-window.test.ts
@@ -1,0 +1,61 @@
+// #1176 — a BACKGROUNDED phone is not a departed phone.
+//
+// #875 defers a self-restart while a paired phone is present, because a restart
+// mints a NEW cloudflared quick-tunnel hostname and the old one stops resolving.
+// The gate asked `hasLiveHeadlessClient()`, which reads currently-OPEN sockets —
+// and a phone at work with its screen off has had its socket suspended or closed
+// by the mobile OS. So the gate opened, the orchestrator self-updated
+// 0.50.39 → 0.50.41, and the reporter picked up their phone to:
+//
+//   Failed host lookup: 'cameron-timing-face-spies.trycloudflare.com'
+//   (No address associated with hostname, errno = 7)
+//
+// A DNS failure, not a timeout or a refusal: the hostname no longer existed.
+//
+// The sticky isHeadless() was rejected for this gate and that was right — a phone
+// that paired once and left would defer updates forever. But those are not the
+// only two options, and the state between them is the NORMAL one for a phone.
+
+import { describe, expect, it } from "vitest";
+import { UiBridge, HEADLESS_RECENCY_MS } from "../../services/ui-bridge.js";
+
+/** A bridge with its headless-disconnect clock placed at a chosen age. */
+function bridgeWithHeadlessGoneFor(ms: number | null): UiBridge {
+  const b = new UiBridge() as UiBridge & {
+    markHeadlessDisconnectForTests(at: number): void;
+  };
+  if (ms !== null) b.markHeadlessDisconnectForTests(Date.now() - ms);
+  return b;
+}
+
+describe("the restart-deferral gate covers a pocketed phone (#1176)", () => {
+  it("still defers seconds after the socket closed — the reported case", () => {
+    // The mobile OS suspends a backgrounded socket within seconds.
+    expect(bridgeWithHeadlessGoneFor(5_000).hasLiveHeadlessClient()).toBe(true);
+  });
+
+  it("still defers across a meeting", () => {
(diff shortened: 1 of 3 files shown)


artokun:
**Codex adversarial review — one real defect found and fixed, four lines of attack came back clean.**

**DEFECT (fixed in the follow-up commit):** the window measures *elapsed time*, and `Date.now()` is not a measure of elapsed time. An NTP correction stepping the wall clock forward by more than the window would make a phone that disconnected seconds ago look long gone — the restart would rotate the tunnel hostname, which is **the exact bug this window exists to prevent**, reintroduced by the clock rather than by the logic. A backward step would defer far longer than intended.

Now on `performance.now()`, monotonic from process start. That is the right lifetime: the value is only ever compared against another reading from the same process, and a restart clears it — after which there is no phone this process has seen, which is the correct answer anyway. Reader, writer and test seam all moved together, so the tests now catch a clock-domain **mismatch** as well as the wrong clock (reverting either side alone fails them).

Clean, each verified with file:line rather than asserted:

1. **Tab migration cannot silently lose a headless client.** The second `conns.delete` is a same-socket re-hello; the handler synchronously reinstalls that same socket under the new id with no interleaving point, so a later close still sees it and arms the clock.
2. **A non-headless socket cannot arm the clock.** The write is guarded by `headless === true`, socket kind is pinned from the first hello, and cross-kind takeovers are refused — so a desktop-only install can never defer.
3. **The restart gate is still the only production caller**, it requires an extant `pairTunnel`, and LAN-only installs are unaffected.
4. **`markHeadlessDisconnectForTests` has no protocol route** — it cannot be reached from a panel or mobile WebSocket message.

Full suite: 382 files, 7830 passed. Lint and the vocabulary gate clean.

PR #1190, opened 2026-08-09
fix(download): report the destination ComfyUI-Manager actually chose

### LINKED ISSUE #1086: Remote download_model ignores extra model paths and writes into 20GB overlay

artokun:
Fixes #1086. **Draft — claiming the issue, work in progress.**

## The report

A remote RunPod ComfyUI read models from `/workspace/models` (the 100GB volume) via `extra_model_paths`, while `/` was a 20GB overlay. `download_model` dispatched to ComfyUI-Manager, which wrote into `/opt/ComfyUI/models` instead. A multi-GB Wan 2.2 file and a required clip-vision file were gone after a pod restart.

## What we say today, and why it is not enough

`managerDestinationCaveat()` already tells the caller we have **not** established where the file lands, and `verifyManagerVisibility` asks the server afterwards whether it lists the file. Both were the right calls at the time — the reasoning being that `extra_model_paths.yaml` lives on the target filesystem, which this process cannot read.

But the reporter's own evidence shows the destination **is** observable:

```
get_system_stats action:"logs" →
  Install model 'clip_vision_h.safetensors' ... into '/opt/ComfyUI/models/clip_vision/clip_vision_h.safetensors'
```

Verified against the installed Manager source — both generations log it:

- `comfyui_manager/glob/manager_server.py:1063`
- `comfyui_manager/legacy/manager_server.py:634`

```py
f"Install model '{json_data['name']}' from '{model_url}' into '{model_path}'"
```

So "we cannot know where it lands" is true only before we look. `/internal/logs` is already read by the health check.

## Plan

After a Manager dispatch, read back the destination Manager logged and report it — and where the live models root is known, say whether that destination is inside it. Turns a permanent caveat into a specific, checkable answer, without pretending we controlled the destination.

Three-state, as ever: a destination we read, a destination the log did not carry, and a log we could not read are different facts and must not render alike.

### DIFF
--- a/src/__tests__/services/manager-destination.test.ts
+++ b/src/__tests__/services/manager-destination.test.ts
@@ -63,9 +63,16 @@ describe("describeManagerDestination", () => {
 
   it("WARNS when the destination is outside the root the server reads", async () => {
     // The reported case, and the one that costs a multi-GB file.
+    //
+    // extraRoots is INJECTED, and must be: without it this falls through to the
+    // live read, which on a developer machine reaches the real ComfyUI and on CI
+    // reaches nothing. That difference is exactly what made this pass locally and
+    // fail on CI — the same "a test silently queried the real server" trap this
+    // suite has hit before. A hermetic test states its own world.
     const note = await describeManagerDestination("clip_vision_h.safetensors", {
       readLog: readLog(LOG("/opt/ComfyUI/models/clip_vision/clip_vision_h.safetensors")),
       liveModelsDir: "/workspace/models",
+      extraRoots: async () => [],
     });
     expect(note).toMatch(/OUTSIDE the models directory/);
     expect(note).toContain("/opt/ComfyUI/models");
@@ -77,6 +84,7 @@ describe("describeManagerDestination", () => {
   it("confirms a destination INSIDE the live root", async () => {
     const note = await describeManagerDestination("clip_vision_h.safetensors", {
       readLog: readLog(LOG("/workspace/models/clip_vision/clip_vision_h.safetensors")),
+      extraRoots: async () => [],
       liveModelsDir: "/workspace/models",
     });
     expect(note).toMatch(/INSIDE the models directory/);
@@ -88,6 +96,7 @@ describe("describeManagerDestination", () => {
     // half is what decides usability. Say the first without implying the second.
     const note = await describeManagerDestination("clip_vision_h.safetensors", {
       readLog: readLog(LOG("/opt/ComfyUI/models/clip_vision/clip_vision_h.safetensors")),
+      extraRoots: async () => [],
       liveModelsDir: undefined,
     });
     expect(note).toContain("/opt/ComfyUI/models");
@@ -101,6 +110,7 @@ describe("describeManagerDestination", () => {
     expect(
       await describeManagerDestination("x.safetensors", {
         readLog: readLog("[INFO] nothing relevant"),
+        extraRoots: async () => [],
         liveModelsDir: "/workspace/models",
       }),
     ).toBeUndefined();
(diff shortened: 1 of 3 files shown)


artokun:
**Codex adversarial review — three real defects found, all fixed.** Two were wrong in the dangerous direction: a false `OUTSIDE` cries wolf about a file that is fine **and** contradicts the `LISTED` verdict rendered beside it in the same message.

**1. Path domain.** The destination comes from the *remote* server's log, in that host's syntax. `currentLiveModelsRoot()` puts its value through node's `resolve()`, which on a **Windows orchestrator** turns `/workspace/models` into `C:\workspace\models`. Verified locally — `resolve("/workspace/models")` really does yield `C:\workspace\models` here. Comparing those always answers `OUTSIDE`, for a destination that is exactly right. The verdict is now withheld unless both paths are written in the same syntax.

**2. Extra roots.** The check compared **one** primary root while a server reads several, so a destination under a registered `extra_model_paths` root was reported unusable. Extra roots are now consulted — and an **unreadable** extra-root list withholds the verdict rather than reading as "there are none". That fold is the #796 defect, and here it manufactures a false unusable.

**3. The filename — a PRE-EXISTING bug the review exposed.** `job.filename` is optional, and for a Manager dispatch `job.path` is a *descriptor*, not a path:

```
"checkpoints/foo.safetensors (dispatched to the remote ComfyUI via …)"
```

so the fallback `job.path.split(/[\/]/).pop()` returned the whole trailing note. **This is not new to this PR** — `verifyManagerVisibility` has used the identical fallback since #1086's first half, so every URL-only Manager download has been probing a name the server could never list and quietly reporting itself unverifiable. `managerJobFilename` fixes both call sites.

Clean, and checked rather than asserted: the regex escaping is complete with no catastrophic-backtracking shape, and `/internal/logs` is a bounded deque (default 300 lines) so reading it back is not a resource regression.

**The fix is not blunted** — a destination outside the primary root, with extra roots known and clear, still gets the full warning and the overlay risk. There is a test pinning exactly that.

Mutation-verified: removing the path-domain guard, treating unknown extra roots as empty, not consulting extra roots at all, and re-breaking the filename fallback each fail the suite.

Full suite: 383 files, 7846 passed. Lint, the #796 gate, and the vocabulary gate all clean.

Issue #420, opened 2026-07-30
Reconnect loses COMFYUI_PATH and blocks model download resume


artokun:
### What happened / root cause
After the sidebar/orchestrator session reconnected, download IDs from the previous session correctly expired, but a new download_civitai_model request against the same connected local ComfyUI failed immediately because COMFYUI_PATH was no longer configured. The previous session had successfully started the same download and list_local_models still reaches the connected ComfyUI. The reconnect appears to lose local-path/download routing state instead of falling back to the connected Manager route.

### Steps to reproduce
1. Start download_civitai_model on a local ComfyUI; observe streaming begins.
2. Restart/reconnect the panel/orchestrator session.
3. Verify list_local_models still queries the connected ComfyUI.
4. Retry download_civitai_model.
5. It fails immediately with COMFYUI_PATH is not configured.

### Exact error (scrubbed)
COMFYUI_PATH is not configured. Set the COMFYUI_PATH environment variable.

### Fix
Applied locally: no. Upstream-only: yes.
Preserve/reconstruct COMFYUI_PATH after panel reconnect, or dispatch through the connected ComfyUI Manager when a server connection exists and the local path is absent.

### Environment
Windows 11 / ComfyUI 0.28.3 / NVIDIA RTX 5070 Ti Laptop 12 GB / comfyui-mcp 0.48.5 / panel version unavailable.

artokun:
Fixed in **mcp 0.48.10** (#440): after a reconnect that loses the local ComfyUI base, downloads route through the still-connected ComfyUI-Manager instead of failing with 'COMFYUI_PATH is not configured'; the in-flight job registry is dual-keyed so a Manager↔local route flip can't spawn a duplicate same-file writer. Update with `npm i -g comfyui-mcp@latest` and reopen if it persists.

Issue #357, opened 2026-07-29
get_node_info misses installed V3 ComfyUI nodes that panel recognizes


artokun:
### What happened / root cause
After installing and enabling `comfyui-qwenmultiangle` 0.7.0, `panel_get_errors` no longer reported `QwenMultiangleCameraNode` as missing and the live canvas recognized it. However, `get_node_info` with the exact node type returned no matches. The pack uses the ComfyUI V3 node API, suggesting the query path omits V3 definitions or reads a different object-info surface than the panel.

### Steps to reproduce
1. Install and enable `comfyui-qwenmultiangle` 0.7.0.
2. Load a workflow containing `QwenMultiangleCameraNode` and confirm the panel does not report the type missing.
3. Call `get_node_info` with `node_type=QwenMultiangleCameraNode`.

### Exact error (scrubbed)
No nodes found matching "QwenMultiangleCameraNode"

### Fix
Applied locally: no. Upstream-only: yes.
Include ComfyUI V3 node definitions in `get_node_info`, or query the same node-definition source used by panel missing-node detection.

### Environment
Windows 11; ComfyUI 0.29.0 LOCAL; RTX 5070 Ti 16 GB; comfyui-mcp 0.48.5; panel 0.11.3.

PR #1193, opened 2026-08-09
fix(#1133): resolve the panel base through the OS-observed process on Desktop

### LINKED ISSUE #1133: ComfyUI Desktop: root corroboration refuses on a Desktop argv, and restart_comfyui declines — no in-tool path to complete a panel update

artokun:
Working the remaining half of #1133 — failure (2), which the owner left open deliberately.

## What the investigation found

The theory in the report — and in my own first comment — was that `--input-directory`/`--output-directory` were being read as evidence of a split install, and the fix was to narrow the guard to `--base-directory`.

**That fix would have been a no-op.** Neither flag appears anywhere in `panel-installer.ts`, `panel-workspace.ts` or `workspace-env.ts`. `parseBaseDirFromArgv` reads only `--base-directory`. "Only `--base-directory` creates the split this guard defends against" is *already* how the code behaves — so narrowing it would have changed nothing while looking, in a changelog, like the report had been addressed.

## The actual mechanism

Desktop reports its launch argv as a **relative** `ComfyUI\main.py` with **no** cwd. Given the reporter's argv (in the issue's Environment block — no `--base-directory`):

- `parseBaseDirFromArgv` → nothing (the flag is absent)
- `liveRootFromArgv` → nothing (a relative dir with no absolute cwd is unresolvable, and it correctly refuses to return a bogus relative root)

So the configured path was the only candidate left — and by construction the running server never vouched for it. `isLiveDerivedBase` was therefore false and the corroboration gate refused. **On every Desktop install, universally.** Not a heuristic misfiring on media flags; a live-derived candidate that never existed.

## The fix

`resolveLiveServerRoot` already has a tier built for exactly this shape (#369): it asks the OS which process is listening on our port, correlates it against that server's own argv, and re-anchors the relative `main.py` on that interpreter's install tree.

`resolvePanelBase` simply was not using it — it re-parsed argv itself. That is precisely what `LiveServerSnapshot.liveRoot`'s own doc warns consumers against: *"must use this rather than re-parsing argv, or they miss the relative-`main.py` shape entirely."* The panel base resolver was that consumer.

So this is **not** a relaxation of a guard that gates a destructive directory swap — it supplies live-derived evidence the system already had. The asymmetry the owner asked for is held explicitly, each pinned by a test:

- `--base-directory` still wins **ahead of** the observed root → a genuinely split install resolves exactly as before
- the observed root is still gated on actually holding `custom_nodes`
- a **failed** observation still leaves the resolution uncorroborated and the gate still refuses — the same tri-state discipline (1)/#1141 used

The anchor is not a guess either: `anchorRelDirOnInterpreter` accepts a candidate only when it really holds `main.py` **and** the observed interpreter belongs to that install, and `resolveLiveInterpreter` correlates against the server's own argv so a proxy cannot impersonate it.

## Verification

- New suite `panel-desktop-observed-root.test.ts` (6 tests), driving `runPanelActionInner` with the **production** dep set over a real filesystem.
- **Both mutation classes checked**, since an install line is invisible to tests: dropping `live-observed-root` from `isLiveDerivedBase` turns 2 red; disabling the call site turns 2 red. Neither passes silently.
- Full suite: **7848 passed**, 1 pre-existing environment-dependent failure (see below).
- `tsc --noEmit` clean; `check:unknown-collapse` reports no new collapses.

Three existing test harnesses needed `resolveLiveServerRoot` added to their `workspace-env` mock (it is newly imported by `panel-workspace`). The observed tier is **off by default** in each, so every pre-existing assertion still describes exactly the server it described before — 331 tests across those files pass unchanged.

## Out of scope

Failure (3) — `restart_comfyui` declining on an unverifiable Desktop supervisor — remains ruled correct behaviour. (1) shipped in 0.50.39.

## Noted, not fixed here

`download-manager-routing.test.ts` › *"#420: no local base + reachable server WITHOUT `--base-directory` → Manager route"* fails on this machine **on a clean checkout of `main` as well**. It is host-dependent: the test's argv is relative with no cwd, so the observed-process tier shells out and finds the *real* ComfyUI listening on :8188 on the dev rig, resolving a genuine root where the test expects none. Pre-existing and unrelated to this diff; filing separately.

Closes #1133


artokun:
### Traced end to end. The draft was an empty stub — here is the actual change, specified.

`98f80da` contains **zero files** (`git show --stat` is empty, and it is not an ancestor of main). So there was nothing here but a claim. Recording the full path so the next pass starts from code, not from the issue.

**Where the refusal is produced:** `panel-installer.ts:2550-2577`, the `liveRootUnderivable` branch — the message the reporter quoted verbatim.

**Where the decision is made:** `panel-workspace.ts:161-168`.

```ts
const liveRootUnderivable = snapshot.reachable;
if (configured) {
  return { base: configured, source: "configured", liveProbeFailed, liveRootUnderivable };
}
```

So `liveRootUnderivable` is set purely from *"the server answered and no candidate root was derivable"*. It never inspects **which** flags the argv carried — which is exactly the reporter's point: `--input-directory`/`--output-directory` relocate media, not `custom_nodes`, and only `--base-directory` creates the split this guard defends against.

**The change:** when the server answered AND reported an argv AND that argv carries no `--base-directory`, the split cannot exist, so `configured` is corroborated rather than merely configured. Three conditions, and the middle one matters — an argv we never received is not the same as an argv without the flag, and collapsing those would be the #796 fold inside the fix for it.

## Why I am not writing it in this pass

This gate protects a **destructive directory swap**, and the owner's own note said relaxing it deserves its own change and its own review. Three things must be established first, none of which I can do well right now:

1. **What the snapshot actually carries.** I have not confirmed the argv is on `snapshot` at this point, or how "no argv reported" is represented. If it is absent rather than empty, the middle condition above has no signal to read and the change is unsafe as written.
2. **The other consumer.** `liveRootUnderivable` is also read at `panel-installer.ts:2008`. Loosening it here changes that message too, and I have not read what it gates.
3. **The negative test is the deliverable.** A Desktop install *with* `--base-directory` must still refuse. A relaxation that passes the reporter's case and also passes the split-install case is worse than the bug — it would move a copy nobody serves and report success, which is the precise harm the guard exists to prevent.

I have shipped four inert guards today by testing the case I had in mind rather than the case that distinguishes. On a destructive path that is not a risk worth taking to close a draft faster.

Draft stays open, now with a specification instead of an empty commit.

artokun:
### Unknowns 1 and 2 resolved. The signal already exists; the remaining question is where to apply it.

**1. The argv IS available, and already parsed for this exact flag.** `panel-workspace.ts:139-145`:

```ts
const snapshot = await getLiveServerSnapshot();
if (snapshot.reachable) {
  [parseBaseDirFromArgv(snapshot.argv, snapshot.cwd), "live-base-directory"],
  [liveRootFromArgv(snapshot.argv, snapshot.cwd),    "live-argv-root"],
```

So `parseBaseDirFromArgv` runs on every reachable snapshot **twelve lines above** the decision that ignores its result. The reporter's discriminator is not missing — it is computed and discarded. No new parsing, and no new notion of "argv absent vs argv without the flag": that distinction already lives inside the existing helper.

**2. The second consumer agrees with the change rather than conflicting.** `panel-installer.ts:2002-2016` is the *no panel found* message, and its outer clause is *"this is the CONFIGURED path … so 'not installed' is UNCORROBORATED"*. If a missing `--base-directory` corroborates the configured path, that message should stop firing too — same reasoning, same direction. Both sites improve together.

### The one real design question left

`liveRootUnderivable` cannot simply be flipped to `false`. It selects the *remedy*, and the `false` branch says **"Start/reach ComfyUI so its own install root can be read"** — which is wrong here, because ComfyUI **is** running. Flipping it would swap a wrong refusal for a wrong instruction, which is the trade this issue is about.

So the change needs a distinct signal — corroborated-because-no-split — rather than reusing the underivable flag. That is a small addition to `PanelBaseResolution` plus the refusal condition at `panel-installer.ts:2550`, which is the last thing to read.

### Still the deliverable, unchanged

A Desktop install **with** `--base-directory` must still refuse. That is the test that distinguishes a fix from a hole, and it is the one I would write first.

Stub → specification → two unknowns closed with code references. Remaining: read the refusal condition, add the field, write the negative test first.

artokun:
### Implemented it, and an existing test caught a hole. Reverted — the finding is worth more than the patch.

Wrote the change: a `liveSplitImpossible` signal set when a reported argv carries no `--base-directory`, plus a narrow early-return in `assertSwapTreeCorroborated`. Five tests, all three conditions mutation-verified.

**`panel-update-corroboration-message.test.ts` then failed** — *"reachable-but-underivable: no 'Start ComfyUI'"* — and it is right to.

Its scenario is `liveArgv = ["main.py"]`: reachable, relative launch path, no cwd, **and no `--base-directory`**. My condition says "no split possible, proceed." But a relative argv means we cannot tell **which install is running at all**. Absence of `--base-directory` rules out the *Desktop-style* split; it does not establish that the configured path is the served tree. The orchestrator can be pointed at install A while the server runs install B — updating there moves a copy nobody serves and reports success, which is the exact harm.

**The reporter anticipated this and I missed it.** Their suggestion was:

> when `server_argv` lacks `--base-directory`, treat the configured path as corroborated **(or fall back to the single-`custom_nodes`-tree check)**

I implemented the first clause and skipped the parenthetical. That clause is not a nicety — it is what supplies the missing half of the proof. Their own case had it: *"exactly one `custom_nodes` tree and exactly one panel pack inside it"*, plus the running server's log naming that pack.

### So the correct condition is a conjunction, not a single flag

1. a reported argv with no `--base-directory` (no Desktop split), **AND**
2. positive evidence that the configured tree is the served one — a derivable root, or the single-`custom_nodes`-tree check.

`["main.py"]` satisfies (1) and fails (2), which is why that test must keep refusing. Any implementation that turns it green without adding (2) is a hole, and it would be invisible except for this test.

Reverted rather than left applied. What this pass produced: the change is fully specified, its obvious form is now known to be **wrong**, and the test that proves it is named. That is worth more than a patch I would have had to withdraw after review — I have already done that once today.

Remaining: implement the conjunction, and the FIRST test to write is `["main.py"]`-still-refuses, since it is what separates the fix from the hole.

artokun:
### The conjunction has a hard prerequisite that does not exist yet — and that makes the no-flag signal worthless on its own

There is **no single-`custom_nodes`-tree helper** in the repo. `hasCustomNodes` (`panel-workspace.ts:120`) answers *"does THIS base have one"*; nothing enumerates trees or asserts there is exactly one. So condition 2 needs a new filesystem capability, not a call to something existing.

That matters more than it sounds, because of what happens when you look for a cheaper condition 2:

- **If the root IS derivable** (absolute `main.py`, or a cwd to resolve against), corroboration already succeeds through `liveResolutionNamesTree`, and `assertSwapTreeCorroborated` returns before my early-return is ever reached.
- **If the root is NOT derivable**, that is precisely the case with no evidence tying the configured tree to the served one — the `["main.py"]` case the existing test protects.

So the set of cases the no-`--base-directory` signal would newly admit is **exactly the unsafe set**. Every safe case is already corroborated by the existing path. There is no partial version of this fix that is both useful and sound: without the single-tree check it adds only the hole, and with it, the single-tree check is doing all the work.

That reframes the issue. It is not "relax an over-strict guard" — the guard is strict in exactly the right places. It is **"add a second corroboration source for installs whose argv cannot be resolved"**, which is a new capability with its own failure modes to think through (symlinked trees, a portable install beside a Desktop one, `custom_nodes` under a base the orchestrator was never told about).

I am not building that on this pass, and I would push back gently on doing it quickly: it is a filesystem scan feeding a gate that authorises a destructive directory swap, and getting it wrong recreates #1133's harm rather than fixing it.

**Draft state:** empty stub → refusal site, decision site, already-computed discriminator, both consumers checked, obvious implementation written and proven wrong by an existing test, and now the prerequisite identified with the reason no cheaper version exists. Whoever picks this up should start from "how do we corroborate an unresolvable-argv install", not from the argv flag.

artokun:
## Live verification against a real install (not just mocks)

This rig's ComfyUI reports the same *class* of argv the report describes — relative script, no cwd:

```
argv: ["main.py","--port","8188","--listen","127.0.0.1"]
cwd : undefined
```

Running the **built** `resolvePanelBase` against the live server, before/after:

| argv shape | `resolveLiveServerRoot` | panel base |
|---|---|---|
| `ComfyUI\main.py` — **the #1133 Desktop shape** | `observed-process` → `…\ComfyUI-Installs\ComfyUI\ComfyUI` | **corroborated** |
| `main.py` — bare, this rig | `unresolved` (relDir `.`) | still refuses |

The anchored root is real: `…\ComfyUI-Installs\ComfyUI\ComfyUI\main.py` and `…\custom_nodes` both exist, and the interpreter the OS reported for the process on :8188 (`…\ComfyUI\standalone-env\python.exe`) genuinely belongs to that install. So the Desktop path is verified end-to-end against a real on-disk layout with a real process observation — not only against the mocked suite.

## Residual gap, stated plainly

The **bare `main.py`** shape (relDir `.`) still does not resolve, and this PR does not change that. `anchorRelDirOnInterpreter` ascends from the interpreter's directory joining `relDir`, so with `relDir = "."` it can only ever land on ancestors of the interpreter — and the portable/standalone layout puts the launcher's `standalone-env\python.exe` as a **sibling** of the nested `ComfyUI\main.py`, never an ancestor of it.

That is the `python main.py` shape #916 addressed by making the refusal *honest* rather than by corroborating it, and I am deliberately not widening the anchor here: reaching a nested sibling would mean searching `serverRootsUnder()` from an interpreter anchor, which is a different inference with a different risk profile — on a gate that authorises a destructive directory swap. It deserves its own change and its own review, exactly as the owner asked for this one.

So: #1133's reported shape is fixed and verified. The bare-`main.py` shape remains underivable by design, with the honest refusal #916 gave it.

artokun:
## Codex review — round 1: NO-SHIP, 3 real bugs

Disclosed: adversarial review by Codex (source-traced; its sandbox denied process creation, so it did not run the suite). It was right, and the findings were inside this diff's own blast radius.

**1+2 — the new tier could license a swap of the WRONG tree.** `parseBaseDirFromArgv` returns `undefined` for a `--base-directory` that is *present but relative* with no reported cwd — it genuinely cannot be resolved. The candidate list read that as "the flag isn't there" and let the next candidate win. But ComfyUI derives custom_nodes from `<cwd>/<flag>`, so the observed root names a tree that flag **has already overridden** — a confidently wrong directory handed to a destructive swap.

Fixed with `hasUnresolvableRelativeBaseDirFlag`, which poisons *every* candidate for that shape. Deliberately narrower than the existing `hasUnresolvableRelativeModelDirFlag`: that one also fires on `--models-directory`, which relocates models and says nothing about custom_nodes, so folding it in would re-create the exact over-refusal this issue is about.

Worth noting this hole was **not new** — an absolute `main.py` alongside an unresolvable relative `--base-directory` already resolved to the install root on `live-argv-root`. My diff extended it to the relative-argv case; the fix closes both.

And the refusal it produced would have advised *"relaunch with an ABSOLUTE path to main.py"* — which fixes nothing here, because the script path is not what failed to resolve. That is the #916 defect pointed at a new branch, so it gets its own `baseDirUnresolvable` flag and its own message naming the flag.

**4 — the probe ran when it could not matter, and it blocks.** Worse than codex knew: `resolveLiveServerRoot` is **synchronous**, so its netstat + WMI walk blocks the event loop. Measured on this machine: **~1.27s**, and far worse on the 5s/8s timeout paths. It was running *before* candidate precedence, so a server whose `--base-directory` was going to win paid for a process-table walk that could never be used. Now reached only after the flag candidate fails.

**5 / P2 — mock faithfulness.** Fair. The new suite no longer stubs `output-dir` at all: `--base-directory` is appended to the reported argv so production's own parser decides what "unresolvable" means, and the observed-tier stub now requires a real `main.py` on disk, matching `anchorRelDirOnInterpreter`'s precondition. A stub that disagrees with the code under test about its own preconditions can only manufacture agreement.

### Not taking finding 3 (process-replacement / spoof race)

Two reasons, stated rather than waved off:

1. This repo's standing threat model is a single local trust domain — accidental bugs, not an adversary who can bind our port and forge a matching argv.
2. More decisively, it is **not a new trust boundary**. This exact tier already decides where multi-GB model downloads are written (#369). If the observed-process anchor were unsound, the damage would land there first and harder. Promoting it to the panel gate does not widen it.

The *accidental* form — ComfyUI restarting between the snapshot and the probe — is bounded by the target/generation guards, which drop any resolution whose target moved mid-probe.

## Round 2 — clean

Re-reviewed the fix commit itself (never merge on the author's PASS). It confirmed the equals-form (`--base-directory=x`), duplicate-flag last-wins, empty-string value, and relative-cwd edges all poison correctly; that a resolvable flag without `custom_nodes` still falls through to the argv root exactly as before; and that the new refusal branch is reachable and correctly ordered after the mismatch branch. **No new defects.**

## State

- 7851 pass, 1 pre-existing environment-dependent failure (#1263, reopened — fails on clean `main` here too)
- `tsc --noEmit` clean, no new unknown-collapses
- Mutation-pinned: dropping the poison guard turns 2 red; dropping `live-observed-root` from `isLiveDerivedBase` turns 2 red; disabling the call site turns 2 red

artokun:
## Merge with `main` (0.50.88) — two guards combined, not chosen between

`main` landed the #796 sweep in the same function: `hasCustomNodes` became a **tri-state** `customNodesState`, so "provably absent" is told apart from "could not be read", and the latter is carried out as `liveRootUnreadable`. This branch had meanwhile restructured the same candidate list.

Kept both. Taking either side wholesale would have dropped a guard — main's alone loses the poisoned-flag refusal; this branch's alone re-folds "unreadable" back into "absent", which is precisely the #796 defect.

### The merge made someone else's test host-dependent

Routing through `resolveLiveServerRoot` pulled three test files into the set that reaches `resolveLiveInterpreter` — including **main's own** `custom-nodes-unreadable`. Main's new `live-probe-gate` (#1290) computes that set rather than listing it, and failed. Stubbed at the module boundary in all three, as it prescribes.

That gate is the fix for #1263 — which I filed from the other direction earlier in this same session, before pulling main. Filed as a new report, triage matched it to a closed #1263 and reopened it; main had already built the general form.

And `custom-nodes-unreadable` stubbed `liveRootFromArgv`, which `resolvePanelBase` no longer calls. Left alone it would have produced no live candidate at all and passed **vacuously** — asserting about a disqualification that never happened. Now stubs the resolver actually used.

## Codex round 3 — one finding, rejected on the evidence

It reported that an unreadable *flag* candidate followed by an accepted *live* candidate returns without `liveRootUnreadable`, where "origin/main would have returned it."

Checked against `git show origin/main:src/services/panel-workspace.ts`: main's accept path is

```ts
return { base: candidate, source, overriddenConfiguredBase: … };
```

— which also omits the field. Main only ever emitted `liveRootUnreadable` on the **fallback** path after the loop, which is exactly what the merged version does. No behaviour change, so no defect. (Codex noted its sandbox denied process creation, so it reviewed GitHub-fetched content rather than local git.)

Its other three checks — the poisoned short-circuit correctly leaving `liveRootUnreadable` undefined, flag-before-live ordering under the tri-state, and no accept/reject flips — confirmed clean.

## State at merge

- `tsc --noEmit` clean
- 8372 pass. One red under full-suite load — `ui-bridge` #357 tolerant-read timing — **passes in isolation**; it is load-sensitive and unrelated to this change.
- Desktop path verified live against a real install earlier in this PR.

Issue #1148, opened 2026-08-08
download_model: tracked background download vanishes silently after a session reconnect (no error, no file, no tray record)


artokun:
### What happened / root cause

A `download_model action:"download"` job that was reported as STARTED and progressing was **silently lost** when the agent session reconnected (the panel emitted the usual "Your connection dropped mid-task ... session resumed with full context" notice).

After the reconnect:

- `download_model action:"status"` with the returned id replied **`No download matching id 53012d3181fd46b6`**
- `download_model action:"status"` with no arguments replied **`No downloads are being tracked.`**
- The destination directory contained **no file and no partial/.part file** — the transfer had not merely lost its tracking record, the bytes were gone too.
- **No error event** was emitted. The panel had earlier emitted a `Model download FAILED` event for unrelated 401s, so the event channel was alive; this job just disappeared.

The tool's own status text explicitly promises the opposite behaviour, which is why this is a trap:

> "Survives a sidebar/tool-session reconnect: an in-flight download started in a previous session is still resolvable by its `id` (or by `url`), so you can confirm it's still running instead of starting a duplicate."

Re-issuing the identical `action:"download"` call started a **fresh** transfer from 0% (it did not adopt anything), which confirms nothing survived.

Impact: an agent that follows the documented contract ("do NOT call action:download again for this URL") and simply waits will wait forever for a 12GB model that is not downloading. The failure is invisible unless the agent independently stats the filesystem. In this session it cost ~40 minutes of a 12GB transfer.

### Steps to reproduce

1. Start a large background download so it stays in flight for a while:
   `download_model action:"download", url:"https://huggingface.co/Aitrepreneur/FLX/resolve/main/krea2_turbo_fp8.safetensors", target_subfolder:"diffusion_models"`
   → returns `Download STARTED and is still running (0%) - id 53012d3181fd46b6`
2. Confirm it is progressing: `download_model action:"status", id:"53012d3181fd46b6"` → `downloading 0.67/12.01 GB (5%)`
3. Let the agent session drop/reconnect (in this session it happened on its own; the panel showed the standard "connection dropped mid-task / session resumed" banner).
4. `download_model action:"status", id:"53012d3181fd46b6"` → **`No download matching id`**
5. `download_model action:"status"` (no args) → **`No downloads are being tracked.`**
6. `ls <ComfyUI>/models/diffusion_models` → no `krea2_turbo_fp8.safetensors`, no partial file.

[...]

PR #1194, opened 2026-08-09
fix(#1148): download status must not promise survival across an orchestrator restart

artokun:
Fixes item 2 of #1148 — the description over-promise. **The issue stays open for item 1** (making downloads outlive the orchestrator), per the owner's framing that the honest wording depends on it.

## The fold

`download_model action:"status"` said:

> Survives a sidebar/tool-session reconnect: an in-flight download started in a previous session is still resolvable by its `id` … so you can confirm it's still running instead of starting a duplicate.

Two events with **opposite answers**, folded into one word. Established from `src/services/download-progress.ts`, not from the changelog:

| event | record findable? | transfer running? |
|---|---|---|
| agent / sidebar reconnect | yes, `downloading` | **yes** — still streaming |
| orchestrator restart | yes, terminal `error` + `interrupted_by_restart` | **no** — it streamed inside the exited process |

So in the second case the sentence told callers to do the **opposite** of the right thing: "confirm it's still running instead of starting a duplicate", when re-issuing is correct. That is what cost the reporter 40 minutes, twice. #1170 made that case legible; the description still described the other one as universal.

The new wording names both cases, says the state is what distinguishes them, and keeps the true half — #529's adoption contract is real, and a caller who stops trusting it starts duplicating live transfers.

## A second fold, found while checking the siblings

`No download matching …` asserted the record was absent because it *"finished long ago … or never started"* — two possibilities, presented as exhaustive. There is a third, and it is load-bearing:

- `migrateInFlightJobs` returns `0` on a readdir failure (`download-progress.ts:106`)
- its caller wraps it in `catch { /* best-effort */ }` (`orchestrator/index.ts:1345`)
- and the old dir is `rmSync`'d **either way**

So a missing record is **not** evidence the download is alive. That is #796's class one layer over, now stated as three causes this lookup cannot distinguish.

## Deliberately unchanged

The stale-heartbeat NOTE — *"may have reconnected, and the transfer may still be running"* — is already honest about what it does not know.

## Verification

- The test asserts the **registered** description via the tool registration, not a copy of the string, so a reworded-but-still-over-promising edit fails it.
- **Mutation-checked.** Worth recording that my first mutation attempt was *invalid*: an unescaped apostrophe broke the file, so the suite failed to parse and reported "no tests" — a failure for the wrong reason, which proves nothing. Redone as a semantic mutation (`is NOT running` → `may still be running`), which fails exactly one test.
- Full suite: **384 files, 7847 passed, 2 skipped**. `tsc --noEmit` clean.
- `docs/tools/models.mdx` regenerated (1 line); the other `docs/` churn was line-endings only and is not in the diff.


artokun:
**Marking this draft. It should land after #1197, or as part of it — not before.**

Five review rounds, and the pattern stopped being convergence. Each round found a real defect in the previous round's *fix*, and the last two escalated:

- **Round 4** found a P1 in the product, not the prose: a Manager dispatch stays `downloading` for the whole server-side transfer (without an aria2 sidecar), so `migrateInFlightJobs` marks a **live** host-side download INTERRUPTED and strips `via_manager`. Filed as #1197.
- **Round 5** found that this PR's protection was **inert** — the record's own text said *"Re-issue the download."* and the description said don't; the record wins. Then my fix for *that* shipped a worse instruction: a two-minute timer that can only terminate the dangerous way, because an in-progress file is never listed. It fired on healthy transfers and caused the corruption it warned about.

That is now backed out, along with an inert assertion and two comments still asserting retracted claims. What remains here is honest and narrower than main.

**But the right fix is one field, and it belongs to #1197.** Copying `via_manager` in the migration lets the record split:

- **local** → "nothing is writing it; re-issue, it resumes from the .partial" — short, correct, terminating
- **Manager** → "the host may still be fetching; the file appears only when it COMPLETES, so do not re-issue on a timer"

That removes the hedging this PR has spent five rounds trying to word around, and removes the need for a two-minute heuristic at all. Doing it here would mean a description PR quietly becoming the migration fix, with `trayId`/`pid`/`started_at` and the missing Manager test coverage trailing behind it.

So: this branch is green (CI 4/4, suite 385 files) and net-better than main, but I would rather it wait than ship a fifth iteration of prose compensating for a code defect one field away from being fixed properly.

Recommend: land #1197 first, then rebase this and delete most of the hedging.

PR #1196, opened 2026-08-09
fix(models): a 404 category is an ANSWER, not an unreadable one

### LINKED ISSUE #1015: list_local_models returns partial inventory when clip and unet category payloads fail to parse
While researching locally deployable models, list_local_models(action:"list") returned a partial inventory instead of a complete authoritative result. It successfully listed several categories but ended with: "Partial listing — 2 categories could not be read (clip: Unexpected end of JSON input; unet: Unexpected end of JSON input). Models in those categories are NOT absent, they are unlisted — retry if something you expect is missing."

Two plausible causes:
1. The connected ComfyUI model endpoint returned truncated or malformed JSON for the clip and unet categories; evidence: the exact category-specific parser error "Unexpected end of JSON input".
2. The MCP aggregation/parsing path mishandled those category responses; evidence: other categories in the same call were returned normally and the tool explicitly downgraded the result to partial rather than confirming absence.

### Steps to reproduce
1. Call list_local_models(action:"list") on this local ComfyUI.
2. Observe that checkpoints, loras, vae, diffusion_models, text_encoders and other categories may return, while clip and unet report JSON parse errors and are omitted.

### Exact error (scrubbed)
"Partial listing — 2 categories could not be read (clip: Unexpected end of JSON input; unet: Unexpected end of JSON input). Models in those categories are NOT absent, they are unlisted — retry if something you expect is missing."

### Fix / workaround
Applied locally: no source patch; this is report-only/upstream-only. Workaround used: query model categories separately and treat omitted categories as unknown, never as absent. Requested fix: make category parsing robust to empty/truncated responses and return the raw endpoint/status so callers can distinguish an empty category from a transport/parse failure.

### Environment
OS: Windows 11
ComfyUI: 0.30.0 local
GPU: NVIDIA GeForce RTX 5080 16 GB
CUDA: 13.0
Torch: 2.13.0
Python: 3.13
comfyui-mcp: 0.50.2
panel: 0.11.43

Related (same root cause): #918

artokun:
Fixes #1015. **Draft — claiming the issue, work in progress.**

## What the reporter sees

```
Partial listing — 2 categories could not be read
(clip: Unexpected end of JSON input; unet: Unexpected end of JSON input).
```

and on their follow-up, the precise shape: `/models/clip` and `/models/unet` each return **HTTP 404 with an empty body**, while `/system_stats`, `checkpoints`, `vae`, `diffusion_models` and `text_encoders` all answer normally.

## Reproduced on the rig

```
clip               HTTP 404 bytes=0
unet               HTTP 404 bytes=0
text_encoders      HTTP 200 bytes=297
diffusion_models   HTTP 200 bytes=360
```

## The cause

`MODEL_SUBDIRS` still carries `clip` and `unet`. Modern ComfyUI renamed those folders — `clip` → `text_encoders`, `unet` → `diffusion_models` — so a current server does not register them at all and answers 404.

The scan treats **any** non-OK status as "could not be read", which is right for a 502 or a proxy page and wrong for a 404: a 404 is the server answering **definitively** that it does not serve that category. Folding the two produces a permanent "partial listing" warning on every healthy modern install, telling the user their inventory is incomplete when it is complete — and the categories named are aliases whose real contents are already listed under their modern names.

That is the house defect pointing the other way: "determined not present" reported as "could not determine".

## Plan

Distinguish a 404 from an unreadable answer, and stop counting a legacy alias whose modern equivalent answered as a coverage gap. The listing must stay honest for a genuine read failure — a 502, HTML, or a transport error still degrades to partial.

### DIFF
--- a/src/__tests__/services/model-listing-coverage.test.ts
+++ b/src/__tests__/services/model-listing-coverage.test.ts
@@ -477,3 +477,46 @@ describe("#1015: all-404 is NOT a verified empty install", () => {
     expect(text).toMatch(/vae/);
   });
 });
+
+// Adversarial review of PR #1196 (my own, before merge) — the all-404 branch
+// answered a FILTERED call with the UNFILTERED diagnosis.
+//
+// list_local_models({model_type:"clip"}) on a healthy modern server returned
+// "usually an older ComfyUI, or a proxy answering in front of it. Check the
+// ComfyUI URL…" — sending someone to debug a URL that works perfectly, when the
+// real answer is that the folder was renamed. A wrong remedy is worse than a
+// vague one: it costs time and teaches distrust of a working setup.
+describe("#1015: a filtered 404 names the RENAME, not a broken server", () => {
+  const base = { answered: [], unanswered: [], usedFilesystem: false };
+
+  it("points clip → text_encoders", () => {
+    const text = describeEmptyModelListing("clip", { ...base, absent: ["clip"] });
+    expect(text).toMatch(/LEGACY name/);
+    expect(text).toContain("text_encoders");
+    // The wrong remedy must be gone.
+    expect(text).not.toMatch(/older\s+ComfyUI, or a proxy/);
+    expect(text).not.toMatch(/Check the ComfyUI URL/);
+  });
+
+  it("points unet → diffusion_models", () => {
+    const text = describeEmptyModelListing("unet", { ...base, absent: ["unet"] });
+    expect(text).toContain("diffusion_models");
+    expect(text).toMatch(/LEGACY name/);
+  });
+
+  it("a NON-legacy 404 category says what to do without inventing a rename", () => {
+    const text = describeEmptyModelListing("gligen", { ...base, absent: ["gligen"] });
+    expect(text).toMatch(/does not serve a "gligen" model category/);
+    expect(text).toMatch(/NO model_type/);
+    expect(text).not.toMatch(/LEGACY name/);
+  });
+
+  it("the UNFILTERED all-404 case keeps the old-server\/proxy diagnosis", () => {
+    // That message is right there and must not be lost to the filtered branch.
+    const text = describeEmptyModelListing(undefined, {
+      ...base,
+      absent: ["checkpoints", "loras"],
+    });
+    expect(text).toMatch(/older\s+ComfyUI, or a proxy/);
+  });
+});
--- a/src/__tests__/services/model-listing-coverage.test.ts
+++ b/src/__tests__/services/model-listing-coverage.test.ts
@@ -386,3 +386,94 @@ describe("#962: the message names the other categories instead of claiming none"
     expect(text).not.toMatch(/fact about ONE folder/);
   });
 });
+
+// #1015 — a 404 category is an ANSWER, not an unreadable one.
+//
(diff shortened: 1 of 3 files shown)


artokun:
## Review — disclosure first

**The Codex review did not complete.** It was dispatched twice: the first attempt was blocked by a read-only sandbox and honestly declined rather than inventing findings; the second fetched the real diff and sources but did not finish inside its execution window. So **no Codex findings informed this PR** — what follows is my own adversarial pass, and I am labelling it that way rather than letting the earlier "run a review" step imply a second opinion that does not exist.

## What my own review found — one real defect, fixed

The all-404 branch answered a **filtered** call with the **unfiltered** diagnosis:

```
list_local_models({model_type:"clip"})
  → "…usually an older ComfyUI, or a proxy answering in front of it.
     Check the ComfyUI URL with get_system_stats (action:'health')…"
```

on a completely healthy modern server. That sends someone to debug a URL that works perfectly, when the real answer is that the folder was **renamed** — the whole origin of this issue. A wrong remedy is worse than a vague one.

Now:

> The connected ComfyUI does not serve a `"clip"` model category — it answered 404 for that route, which is a definite answer, not a failed read.
>
> `"clip"` is the LEGACY name for this folder; current ComfyUI calls it `"text_encoders"`. Ask for `"text_encoders"` instead — the models you are looking for are almost certainly listed there.

A non-legacy 404 says what to do without inventing a rename, and the unfiltered all-404 case keeps the old-server/proxy diagnosis.

## The other questions, answered honestly

- **Is a 404 always safe to read as "does not exist"?** Not *always*, and I will not claim it is. A gateway that 404s unknown paths, or an auth gate hiding routes, would also 404 — but those 404 the whole `/models/*` prefix uniformly, which lands in the all-404 branch that refuses to conclude anything. The residual risk is a proxy that 404s **some** categories and not others; I could not construct a realistic configuration that does that, since proxies route by prefix. Weighed against what it replaces — a **constant** false "your inventory is incomplete" on every healthy modern install — this is the right trade, but it is a trade.
- **Consumers of the optional `absent`.** There are exactly three reads, all guarded (`??=`, `?? []`, `?? []`). The type is `ModelListingCoverage`, consumed in only two places besides its constructor; both handled.
- **Filesystem fallback.** `dirsToScan` is unchanged, so a category that 404s over HTTP but exists on local disk is still scanned and still lists. `absent` only feeds the empty-listing message, which is not reached when models were found.
- **Counting.** The "N categories could not be read" phrasing is built from `coverage.unanswered` alone, so excluding 404s makes that count *more* accurate, not less.

## Testing

Mutation-verified, both the fix and the review fix:

| mutation | result |
|---|---|
| 404 back to `unanswered` (the reported partial-listing returns) | ✗ killed (3 tests) |
| every non-OK treated as absent (a 502 stops degrading) | ✗ killed |
| all-404 rendered as a verified empty install | ✗ killed |
| all-404 branch swallows the unanswered case | ✗ killed |
| filtered branch removed (wrong remedy returns) | ✗ killed (3 tests) |
| rename map emptied | ✗ killed |
| filtered branch swallows the unfiltered case | ✗ killed |

Full suite: 384 files, 7870 passed. Lint, the #796 gate and the vocabulary gate clean. CI green on all three platforms.

Issue #1174, opened 2026-08-09
panel_save_workflow(name:) Save-As fences the session, and panel_set_workflow_target({mode:'current'}) cannot recover it


artokun:
### What happened / root cause

Calling `panel_save_workflow({name: "..."})` on a never-before-saved (Untitled) workflow performs a Save-As that changes the active workflow's identity/uuid, but the session's workflow-instance fence is NOT re-derived by the save itself. Every subsequent `panel_*` call is then refused with `workflow instance mismatch`, even though the save succeeded and the user never switched tabs.

The save-as is the very operation that invalidated the fence, so it should re-derive the fence from its own reply (exactly as `workflow_open` already does). Instead the caller is left in a fenced state immediately after a successful, agent-initiated write.

Worse, the documented recovery path does not work.

### Steps to reproduce

1. Start on a blank/Untitled workflow.
2. Build a graph with `panel_add_node` / `panel_connect` / `panel_set_widget` (all succeed).
3. `panel_save_workflow({name: "Video Upscale 2x (ESRGAN)"})`
   → returns `{saved: true, saved_as: true, copied_from: "Untitled ...", original_on_disk: true}` **plus** a fence error.
4. Any `panel_*` call now fails with `workflow instance mismatch`.

### Exact error (scrubbed)

Step 3 — the save succeeded but reported:

```
The workflow WAS saved. Could NOT read the live canvas identity — the panel did not
answer, so NOTHING about this session's graph binding was observed. It is still fenced
to the previous workflow instance (921827c8-...), which the panel will keep refusing.
This is not a rebind — it is an unknown.
```

Step 4 — `panel_graph_outline`:

```
Error: workflow instance mismatch: this command targets a different workflow than the
active canvas — the workflow was switched or replaced after it was issued.
```

Recovery attempt A — `panel_set_workflow_target({mode:"current"})` **failed**, and its own error text concedes the recovery is structurally impossible from this state:

```
panel_set_workflow_target({mode:"current"}) did NOT restore this session's graph binding.
APPLIED: the workflow target is now mode:"current".
NOT APPLIED: Could NOT read the live canvas identity — the panel did not answer ...
The panel refused the read because of the FENCE itself, so retrying this call cannot work
— it needs the very read the fence blocks.
```

This is the bug within the bug: `mode:'current'` is documented as *the* recovery signal for an instance mismatch, but it needs a fenced read to do its job, so it can never clear the fence it is meant to clear.

[...]

Issue #1077, opened 2026-08-08
Workflow-instance command fence permanently REFUSED even after panel_list_workflows/panel_open_workflow confirm the exact same uuid (RunPod/remote session)


artokun:
### What happened / root cause

During a remote (RunPod) ComfyUI session, `panel_graph_outline` and `panel_set_workflow_target({mode:"current"})` got permanently stuck refusing to adopt the workflow-instance command fence, even though two OTHER read tools (`panel_list_workflows`, `panel_open_workflow`) both independently confirmed the exact same live `workflow_uuid`. Multiple hard browser refreshes (Ctrl+Shift+R) and even closing/reopening the tab did not clear it — the failure was structural for the session, not a transient reconnect glitch.

I traced the fence-adoption path in the installed build (`dist/orchestrator/index.js` `setTabWorkflowUuidResolver`'s refresh callback, `dist/orchestrator/panel-tools.js` `rebindWorkflowFence`/`corroborateActiveForFence`, `dist/orchestrator/session-store.js` `workflowIdentityParts`, `dist/services/ui-bridge.js` `tabServerOrigin`/`handleConnection`/`attachRelayConnection`). Two findings:

[...]

PR #1199, opened 2026-08-09
fix(fence): a rebind that fails on an OLD panel says so, and names the version

### LINKED ISSUE #1043: panel_new_workflow leaves session fenced to previous workflow instance; set_workflow_target({mode:'current'}) cannot recover it

`panel_new_workflow` created the blank workflow tab successfully, but the follow-up read of the live canvas identity did not get an answer from the panel. The tool reported this honestly ("Could NOT read the live canvas identity"), but the session's workflow-instance fence was left pointing at the PREVIOUS workflow instance (`8a1de94f-...`). From that point every `panel_*` graph call — including read-only ones — is refused with:

`workflow instance mismatch: this command targets a different workflow than the active canvas`

The documented recovery, `panel_set_workflow_target({mode:"current"})`, does NOT recover it. It reports `APPLIED: the workflow target is now mode:"current"` but then `NOT APPLIED: Could NOT read the live canvas identity — the panel did not answer`, and the fence stays on the stale instance. So the recovery path depends on the exact same canvas-identity read that is already failing, which makes it unable to ever break the deadlock.

Net effect: the agent session is hard-wedged out of ALL graph tools (read AND write) and the only escape is asking the human to manually F5 the browser tab. `panel_reload` is explicitly contraindicated here (#803).

### Steps to reproduce

1. Build and run a graph in workflow A (several `panel_add_node` / `panel_connect` / `panel_set_widget` / `panel_run` cycles).
2. `panel_save_workflow({name:"..."})` — succeeds (Save-As).
3. `panel_new_workflow()` — the tab IS created, but the canvas-identity read fails.
4. Any `panel_*` call → `workflow instance mismatch`.
5. `panel_set_workflow_target({mode:"current"})` → fails to rebind, same message.
6. Repeat 4 and 5 — deadlock persists indefinitely.

Observed sequence in one session: `panel_new_workflow` → `panel_graph_outline` (fail) → `panel_set_workflow_target` (fail) → `panel_list_workflows` (fail) → `panel_set_workflow_target` (fail).

### Exact error (scrubbed)

```
The blank workflow WAS created. Could NOT read the live canvas identity — the panel
did not answer, so NOTHING about this session's graph binding was observed. It is
still fenced to the previous workflow instance (8a1de94f-...), which the panel will
keep refusing. This is not a rebind - it is an unknown.

UNDERLYING CAUSE: Error: workflow instance mismatch: this command targets a different
workflow than the active canvas - the workflow was switched or replaced after it was
issued.
```

### Fix

applied locally: no
upstream-only: yes

Two suggested changes, either of which breaks the deadlock:

1. **Make `panel_new_workflow` adopt the instance it just created.** It already knows the new tab's key (it returns `key`/`routing_key`, e.g. `tmp:341bfac3-...`). When the confirmation read fails, the fence should be optimistically set to that just-created key rather than left on the old instance — the tool created the tab, so it is the best available authority on what the canvas now is. Leaving the fence on a workflow that is provably no longer active is strictly worse than adopting the new key.

2. **Give `panel_set_workflow_target({mode:"current"})` a fence-clearing escape hatch.** Right now it can only rebind if the canvas-identity read succeeds, so it cannot recover the one situation it is documented to recover. It should support clearing/invalidating the stale fence unconditionally (e.g. `mode:"current"` with a `force`/`reset` flag, or simply dropping the fence when the read fails), so the NEXT graph call re-derives the binding instead of being pre-refused against a known-dead instance.

A retry/backoff on the canvas-identity read inside `panel_new_workflow` would also reduce how often this is hit, but does not fix the unrecoverable state on its own.

### Environment

- ComfyUI: 0.30.2 (local, Comfy-Desktop standalone)
- GPU: NVIDIA GeForce RTX 5080, 16 GB VRAM; 95 GB RAM
- torch 2.10.0+cu130 / CUDA 13.0 / python 3.13.12
- comfyui-mcp: 0.50.13
- panel: 0.11.43

Note: no generation was running at the time; the preceding `panel_run` had completed and its result had been delivered.

artokun:
Fixes #1043. **Draft — claiming the issue, work in progress.**

## The pattern across three reports

#1043, #1077 and #1174 all end in the same deadlock: a command that re-points the active workflow (`panel_new_workflow`, a Save-As) succeeds, the session's fence is not re-derived, and every later `panel_*` call is refused with `workflow instance mismatch`. The documented recovery, `panel_set_workflow_target({mode:"current"})`, then fails because it needs a read the fence itself blocks.

The orchestrator already has the fix for this: `refreshFenceFromOwnReply` (#1161, shipped in mcp 0.50.43) trusts the `workflow_uuid` the command's **own reply** carries, before ever attempting the round trip that gets refused.

**It depends on the panel publishing that uuid — which is panel #800, first shipped in panel 0.11.45.** And every one of these reporters is below it:

| issue | panel |
|---|---|
| #1043 | 0.11.43 |
| #1174 | 0.11.44 |
| #1077 | (checking) |

So the fix is present and inert, and the user is told only that the identity "could not be read" — an unexplained unknown, when the actual answer is "your panel is too old to answer this, update it".

## Plan

The orchestrator knows the panel's version from its `hello` (`panelVersion`), and there is already a semver gate for exactly this class of judgement. When the reply-trust path finds no uuid **and** the connected panel is below the version that publishes one, say that — with the version they have, the version they need, and how to update — instead of reporting a bare unknown.

Strictly additive: a panel new enough to publish the uuid is unaffected, and a panel of unknown version keeps the existing wording rather than guessing.

### DIFF
--- a/src/__tests__/services/panel-version-gap.test.ts
+++ b/src/__tests__/services/panel-version-gap.test.ts
@@ -18,12 +18,14 @@
 import { describe, expect, it } from "vitest";
 import { UiBridge, PANEL_MIN_VERSION_REPLY_UUID } from "../../services/ui-bridge.js";
 
-/** A bridge whose resolveTarget reports the given advertised panel version. */
-function bridgeWithPanel(version: string | undefined): UiBridge {
+/** A bridge whose resolveTarget reports the given panel version. `advertised`
+ *  models whether THIS connection's hello carried it (vs inheriting it from a
+ *  previous one) — the distinction codex review showed is load-bearing. */
+function bridgeWithPanel(version: string | undefined, advertised = true): UiBridge {
   const b = new UiBridge(0) as UiBridge & {
-    resolveTarget: (id: string) => { panelVersion?: string };
+    resolveTarget: (id: string) => { panelVersion?: string; panelVersionAdvertised?: boolean };
   };
-  b.resolveTarget = () => ({ panelVersion: version });
+  b.resolveTarget = () => ({ panelVersion: version, panelVersionAdvertised: advertised });
   return b;
 }
 
@@ -49,6 +51,15 @@ describe("panelTooOldForReplyUuid (#1043)", () => {
     for (const v of [undefined, "", "dev", "nightly", "not-a-version"]) {
       expect(bridgeWithPanel(v).panelTooOldForReplyUuid("t").tooOld, String(v)).toBe(false);
     }
+  });
+
+  it("an INHERITED version is not narrated as too old (codex review)", () => {
+    // A re-hello that omits panel_version inherits the previous value, so a panel
+    // that was updated and reloaded without re-advertising would still read as
+    // old — and we would tell someone to update a panel they just updated.
+    expect(bridgeWithPanel("0.11.43", false).panelTooOldForReplyUuid("t").tooOld).toBe(false);
+    // …while a freshly advertised old version still is.
+    expect(bridgeWithPanel("0.11.43", true).panelTooOldForReplyUuid("t").tooOld).toBe(true);
   });
 
   it("never throws when the tab cannot be resolved", () => {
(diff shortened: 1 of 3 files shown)


artokun:
**Codex adversarial review — two real defects found, both fixed.** Both produced the same user experience this note exists to prevent: a confident **wrong remedy**.

Codex could not start a shell in this worktree, so it fell back to the GitHub API and grounded every claim against the real PR head `89e7d98` — and flagged that it therefore could not run tests. Worth stating, since that limits what its "no defect" answers cover.

**1. The claim was false on one of its three call sites.** The note says that on 0.11.45+ *"the command that re-pointed the canvas repairs the fence from its own reply and never makes this read"*. That is true of `panel_save_workflow` and `panel_new_workflow`, whose replies carry a `workflow_uuid`. It is **not** true of `panel_set_workflow_target`, which has no own-reply uuid to gain — an updated panel would still make that read, so "updating removes this failure" is simply wrong there.

A user may well have arrived at that recovery *from* a save/new that hit the gap. But arriving there does not establish it, and guessing which it was is exactly the unearned claim this area exists to avoid. The two paths that can prove it now say it at the moment it is provable.

**2. The version could be stale.** A re-hello that omits `panel_version` **inherits** the previous value, so a panel that was updated and reloaded without re-advertising would still read as old — and we would tell someone to update a panel they had just updated. `panelVersionAdvertised` exists for precisely this (the proactive command gate already refuses to act on an inherited value), and the accessor now requires it.

Clean, each grounded in real code:

3. The note is correctly scoped to the `unreadable` branch — the other six rebind outcomes have a different demonstrated cause.
4. `install_comfyui(action:"panel", panel_action:"sync")` is a real route that works on-disk without needing a bridge reply, and correctly requires a restart.
5. `SEMVER_RE` gates malformed/2-part/empty/leading-`V` strings before `compareSemver`, so the tri-state cannot misfire in either direction.

Mutation-verified for both fixes: trusting an inherited version again, and restoring the note on `set_workflow_target`, each fail the suite. The wiring test now pins the count at **exactly two** call sites rather than "all of them" — which is what would have let defect 1 back in.

Full suite: 385 files, 7878 passed. Lint, #796 gate and vocabulary gate clean.

PR #1200, opened 2026-08-09
fix(#1197): a live ComfyUI-Manager download must not be reported INTERRUPTED

### LINKED ISSUE #1197: An orchestrator restart tells you to re-issue a LIVE ComfyUI-Manager download — which corrupts the model
Found during the adversarial review of #1194 (issue #1148). This is a code defect, not a wording one, and it ends in the corrupt-model path this repo documents elsewhere.

## The path

1. Remote/RunPod target, so `shouldDispatchDownloadToManager()` is true. `download_model` dispatches a 12 GB fetch to ComfyUI-Manager, which runs **on the ComfyUI host**.
2. `queueManagerTask` → `runManagerQueue` **polls until the Manager queue drains**, under the model budget (`modelTimeoutMs: 4 * 60 * 60_000`). `commitDone` for the Manager route only runs when that returns (`src/services/download-jobs.ts:612`). So for the entire server-side transfer the persisted record sits at `status:"downloading"` with `via_manager:true`.
3. The orchestrator restarts 20 minutes in.
4. `migrateInFlightJobs` migrates **every** `downloading` record (`src/services/download-progress.ts:118`) with **no `via_manager` check**, writing a terminal record whose text is:

   > It is NOT running and will not resume on its own — nothing is waiting to finish… Re-issue the download.

   That is false. The host never stopped; a restart here does not touch it.
5. The migrated record copies `id/name/url/dest/target/dest_key/req_key/total/received` and **not `via_manager`** (`download-progress.ts:123-141`), so it renders through the plain LOCAL `INTERRUPTED` branch. Neither the agent nor the user can tell it was a Manager job.
6. The caller does as instructed and re-issues → a second server-side fetch to the same destination. This repo already states the consequence (`node-management.ts:971-979`): two concurrent writes produce **a corrupt model**, surfacing later as `shape is invalid for input of size N`.

## Why it went unnoticed

The word "manager" does not appear anywhere in `src/__tests__/services/download-progress.test.ts`. The migration has no Manager case at all.

It is also easy to misread the code: `download-jobs.ts:611` has the comment *"commit done here when it returns (dispatch accepted…)"*, which reads as commit-at-acceptance. Two separate reviewers and I all read it that way. `queueManagerTask` actually drains first — the parenthetical is describing what `done` MEANS for a Manager job, not when it fires.

## Related, same migration

`migrateInFlightJobs` also drops `trayId`, `pid` and `started_at`. Consequences: a caller who passes the `tray_id` they were handed (the tool's own #822 guidance) gets "not found" on a record that exists; the row renders `(tray \`undefined\`)`; and the candidate listing prints `NaN s ago`.

And the INTERRUPTED verdict itself is asserted without evidence. Every persisted record carries the writer's pid, and `writerProcessGone()` exists precisely to answer this — the cancel path **refuses** to close a stale record until that probe returns ESRCH (#761/#858: *"Absence of evidence is NOT evidence of absence"*). The migration reads neither.

## Suggested direction

Skip `via_manager` records in the migration, or migrate them with a Manager-specific note that does not say the transfer stopped; copy `via_manager`/`trayId`/`pid`/`started_at` for the ones it does migrate; and consider probing the pid before writing "It is NOT running", to match the standard the cancel path already holds.

Environment: present on main at 0.50.52; the reproduction is configuration-dependent (remote target + Manager dispatch + a restart mid-transfer), not platform-dependent.

Related (same root cause): #1148, #761, #858

artokun:
Fixes #1197. **Draft — claiming the issue.**

## The path

A Manager dispatch's record stays `downloading` for the whole server-side transfer (without an aria2 sidecar, where `runManagerQueue` drains only on completion). An orchestrator restart then runs `migrateInFlightJobs`, which:

1. migrates **every** `downloading` record with **no `via_manager` check**, writing a record that says the transfer stopped — for a fetch running on the ComfyUI **host**, which a restart here does not touch;
2. **drops `via_manager`**, so the record renders as a plain LOCAL interruption and nobody can tell the difference.

Re-issuing then writes a second copy to the same destination. `node-management.ts:971-979` says what that produces: **a corrupt model**.

## Why the fix is a field, not wording

#1194 spent five review rounds trying to word around this, and each attempt produced a new false claim — because the record cannot say the right thing for a route it has thrown away. Carry `via_manager` and the text splits cleanly:

- **local** → nothing is writing it; re-issue, it resumes from any surviving `.partial`. Short, correct, terminating.
- **Manager** → the host may still be fetching; the file is not listed until it COMPLETES, so absence proves nothing and a timer is not a test.

## Plan

- [ ] Carry `via_manager` (plus `trayId`, `pid`, `started_at` — dropped today, and load-bearing: the row renders `(tray undefined)` and the listing prints `NaN s ago`)
- [ ] Split the migrated record's text on that flag
- [ ] Do NOT claim death for either route — the migration reads neither `pid` nor `owner`, while `writerProcessGone()` exists to answer exactly that and the cancel path refuses to act without it (#761/#858)
- [ ] Tests: the word "manager" currently appears **nowhere** in `download-progress.test.ts`, which is why this shipped

## Config note

Reproduces **without** an aria2 sidecar. With one — and `docker/runpod/post_start.sh` exports `COMFYUI_MANAGER_ARIA2_SERVER`, so this project's own RunPod image has it — the queue drains at handoff, the record goes terminal `done`, and nothing migrates. A repro needs that variable unset, or the bug looks unreal.

Unblocks #1194, which is parked in draft pending this.

### DIFF
--- a/src/__tests__/services/download-progress.test.ts
+++ b/src/__tests__/services/download-progress.test.ts
@@ -376,7 +376,7 @@ describe("migrateInFlightJobs (#1148)", () => {
       status: "downloading",
       via_manager: true,
       name: "wan22.safetensors",
-      tray_id: "tray-abc",
+      trayId: "tray-abc",
       filename: "wan22.safetensors",
       target_subfolder: "diffusion_models",
       started_at: 1_700_000_000_000,
@@ -424,7 +424,10 @@ describe("migrateInFlightJobs (#1148)", () => {
       writeJob(oldDir, managerJob);
       mod.migrateInFlightJobs(oldDir, dir);
       const rec = mod.readPersistedDownloadJob("mgr-12gb")!;
-      expect(rec.tray_id).toBe("tray-abc");
+      // Asserted against the REAL persisted key. The first version of this
+      // test used `tray_id` in both the fixture and the assertion, so it
+      // passed while the migration copied nothing.
+      expect(rec.trayId).toBe("tray-abc");
       expect(rec.filename).toBe("wan22.safetensors");
       expect(rec.target_subfolder).toBe("diffusion_models");
       expect(rec.started_at).toBe(1_700_000_000_000);
--- a/src/__tests__/services/download-progress.test.ts
+++ b/src/__tests__/services/download-progress.test.ts
@@ -366,6 +366,50 @@ describe("migrateInFlightJobs (#1148)", () => {
       JSON.stringify(rec),
     );
 
+  // The structural hole that produced BOTH key bugs in this function: every
+  // fixture here is hand-built, so a key the WRITER never emits still reads as
+  // "carried" and the test agrees with the mistake. `tray_id` vs `trayId` was
+  // caught by hand; `name`/`dest`/`target`/`total`/`received` — five keys that
+  // belong to the tray-row interface, not this record — survived for months and
+  // even had an assertion claiming bytes that are always undefined.
+  //
+  // This closes it by ROUND-TRIPPING a record the real writer produced.
+  it("round-trips a REAL persisted record, not a hand-built fixture", async () => {
+    const src = mkdtempSync(join(tmpdir(), "cm-rt-"));
+    const writer = await (async () => {
+      vi.resetModules();
+      process.env.COMFYUI_MCP_PROGRESS_DIR = src;
+      return import("../../services/download-progress.js");
+    })();
+
+    writer.persistDownloadJob({
+      id: "rt-1",
+      trayId: "tray-rt",
+      url: "https://example.invalid/m.safetensors",
+      target_subfolder: "checkpoints",
+      filename: "m.safetensors",
+      status: "downloading",
+      started_at: 1_700_000_000_000,
+      via_manager: true,
+    } as never);
+
+    // Migrate with the module under test (its own dir), reading what the WRITER
+    // actually wrote.
+    mod.migrateInFlightJobs(src, dir);
+    const rec = mod.readPersistedDownloadJob("rt-1");
+    rmSync(src, { recursive: true, force: true });
+
+    expect(rec, "a genuinely-written record must migrate").not.toBeNull();
+    // Every field the migration claims to carry, verified against the writer's
+    // own output rather than a fixture that could share my typo.
+    expect(rec!.trayId).toBe("tray-rt");
+    expect(rec!.filename).toBe("m.safetensors");
+    expect(rec!.target_subfolder).toBe("checkpoints");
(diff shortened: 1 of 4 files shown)


artokun:
**Review returned NO-SHIP. Two fixed, F1 deliberately left — and F1 is the one that matters.**

## Fixed here

- **F2 (HIGH)** — `name`, `dest`, `target`, `total`, `received` are not keys on `PersistedDownloadJob` at all; they belong to the tray-row interface. The original #1148 author copied the wrong one and I propagated it, *while claiming I had checked against the interface* — the check only covered the keys I added. It had already produced a lying test asserting bytes on a field that is always `undefined`.
- **The structural hole behind it.** Every fixture here is hand-built, so a key the writer never emits still reads as "carried". There is now a test that round-trips through the real `persistDownloadJob`; dropping `trayId` fails two tests including that one.
- **F5** — "resumes from any surviving `.partial`" overclaimed; without a validator sidecar the partial is discarded and the download restarts from zero.

## Not fixed — required before merge

**F1 (HIGH).** The renderer was never touched. It still prints `**error**` and `INTERRUPTED —` beside the new text, and `action:"cancel"` says *"already error — nothing to cancel"*. So the caller sees:

> `- id (tray x)` **error** … `INTERRUPTED — This download … is very likely STILL RUNNING`

The token asserting failure wins by position, and cancel reinforces it. **The string fix is inert at the point of delivery** — the third time in this cluster a fix of mine failed to reach the user. Needs three sites in `model-management.ts` branched on `via_manager` for the interrupted case.

**F3 (MEDIUM).** "very likely STILL RUNNING" overclaims in the other direction for three states: never-dispatched (the record is persisted before the dispatch is enqueued), a failed server-side fetch, and — the important one — **a RunPod pod restart takes ComfyUI and the orchestrator together**, since `post_start.sh` boots both in one container. There, *"the ComfyUI host, which the restart here did not touch"* is exactly inverted. The message also deleted the actionable half that the sibling text at `node-management.ts:975` keeps, so it reintroduces #1148's "agent will not act" harm for the Manager route.

**F4/F6** and two surviving copies of the retracted "the transfer IS dead" rationale (`orchestrator/index.ts:1340`, and this test file's own header).

## Why I stopped

Fixing F1 means branching three renderers in a different file and layer, untested by this PR, and F3 needs a positive test restored to the message. Both are real work, not touch-ups. Given this cluster has produced a new false claim in every single round where I pushed on, the honest move is to leave the remaining findings written down rather than take a fourth swing at delivering a message that keeps not arriving.

Suite green (42 in this file, full suite green as of the previous commit); CI was 4/4 before these commits.

Issue #485, opened 2026-07-30
enqueue_workflow discards ComfyUI prompt validation details on HTTP 400


artokun:
### What happened / root cause
When ComfyUI rejected a workflow during prompt validation, `enqueue_workflow` returned only `Endpoint Bad Request (400 Bad Request): http://127.0.0.1:8188/prompt?...` and discarded the response/log details identifying the invalid node. The actual cause was visible only after a separate get_logs call: LoadImage node 5 referenced an invalid input file because ComfyUI was launched with a custom shared input directory.

### Steps to reproduce
1. Submit an API workflow with a LoadImage node referencing a file absent from the server's configured input directory.
2. Call enqueue_workflow.
3. Observe generic HTTP 400 with no node id, input name, or validation message.
4. get_logs reveals `LoadImage 5 ... Invalid image file`.

### Exact error (scrubbed)
Tool: `Endpoint Bad Request (400 Bad Request): http://127.0.0.1:8188/prompt?clientId=comfyui-mcp`
Server log: `Failed to validate prompt for output 11: * LoadImage 5: Custom validation failed for node: image - Invalid image file: RogueTD_Ara_Locomotion_Source_20260730.png`

### Fix
applied locally: no  upstream-only: yes
No runnable comfyui-mcp source checkout was available in the visible install. Suggested change: parse and return ComfyUI's 400 response body (`node_errors` / prompt validation details) from enqueue_workflow, including output id, node id/type, input, and message. Fall back to the generic HTTP status only when the response body is empty or malformed.

### Environment
Windows 11 / ComfyUI 0.29.0 / RTX 3080 10 GB / comfyui-mcp 0.48.15 / panel 0.11.4.

PR #1202, opened 2026-08-09
fix(enqueue): redact the response body in the enqueue error builders

### LINKED ISSUE #1191: buildEnqueueError and the cloud enqueue path print up to 500 RAW body bytes into a tool result
Found during the adversarial review of #1187 (issue #385). Filing separately because it is genuinely pre-existing and out of that PR's scope — both sites are on the `comfyuiFetch` path and have always been reachable, so #1187 neither causes nor fixes them.

## What

Two enqueue error builders interpolate the response body verbatim:

- `src/comfyui/client.ts:674` — `buildEnqueueError`: `` `${generic}: ${bodyText.slice(0, 500)}` ``
- `src/comfyui/cloud-client.ts:66` — identical raw-500-byte shape on the cloud path

Both also print `res.statusText` unscrubbed (`client.ts:624`).

## Why it matters

This is the same class #828 closed for the diagnosis path and #1187 closed for the branches it made reachable: **a gateway that reflects the request can echo our own credential in the body it answers with**, and that body goes straight into a tool result the agent reads and users paste into bug reports. The repo already has the tool for this — `bodyPrefixOf` in `src/comfyui/json-guard.ts`, which redacts by shape as well as by known value and fails closed when it cannot substitute safely. `describeStatus` is the matching one for the reason phrase, which is attacker-influenceable on a hostile proxy.

`/prompt` is the worst endpoint to have this on: it is the one call that enqueues work, so it is exercised on every render, and its error path is the one users are most likely to hit and share.

## Note on scope

An earlier note of mine described this as "prints a raw statusText". That undersold it — the review corrected me, and the **body** is the larger exposure of the two. Naming line 674 specifically so a fix does not stop at the status line.

The fix looks like a two-line change at each site (`bodyPrefixOf(bodyText)` and `describeStatus(res.status, res.statusText)`), but it wants a test with a reflecting-gateway body, and ideally one that configures a real `COMFYUI_AUTH_*` value so the known-value redaction pass is exercised rather than only the shape-based passes.

Environment: present on main at 0.50.50; platform-independent (pure string construction).

Related (same root cause): #828

artokun:
Fixes #1191. **Draft — claiming the issue, work in progress.**

## What

Two enqueue error builders interpolate the response body verbatim, and print `statusText` unscrubbed:

- `src/comfyui/client.ts` — `buildEnqueueError`: `` `${generic}: ${bodyText.slice(0, 500)}` ``
- `src/comfyui/cloud-client.ts` — the identical raw-500-byte shape on the cloud path

## Why it matters

A gateway that reflects the request can echo **our own credential** in the body it answers with, and that body goes straight into a tool result the agent reads and users paste into bug reports. `/prompt` is the worst endpoint to have this on: it is the one call that enqueues work, so it runs on every render, and its error path is the one users are most likely to hit and share.

This is the same class #828 closed for the diagnosis path and #1187 closed for the branches it made reachable — genuinely pre-existing on both sites, which is why it was filed separately rather than folded into that PR.

## Plan

The repo already has the right tools:

- `bodyPrefixOf` (`src/comfyui/json-guard.ts`) — redacts by shape as well as by known value, and fails closed when it cannot substitute safely;
- `describeStatus` — the matching one for the reason phrase, which is attacker-influenceable on a hostile proxy.

Use both at each site. The change is small; the care is in verifying the redaction genuinely fires for a credential-bearing body, and that a legitimate ComfyUI validation error (the `node_errors` diagnosis #485 exists to surface) is still readable afterwards — a redaction that eats the useful half would trade one defect for another.

### DIFF
--- a/src/__tests__/comfyui/cloud-client-hardening.test.ts
+++ b/src/__tests__/comfyui/cloud-client-hardening.test.ts
@@ -20,6 +20,10 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 
 vi.mock("../../config.js", () => ({
+  // #1191 — the scrubber now redacts the CLOUD API KEY by known value, so it
+  // reads `config` directly. Stubbed empty: this test is about the error's
+  // delivery-doubt wording, not about redaction.
+  config: {},
   getApiKey: () => "test-key",
   getCloudUrl: () => "https://cloud.comfy.org",
   getComfyUIAuthHeaders: () => ({}),
--- a/src/__tests__/comfyui/cloud-key-redaction.test.ts
+++ b/src/__tests__/comfyui/cloud-key-redaction.test.ts
@@ -1,0 +1,49 @@
+// #1191 (codex review) — the CLOUD API KEY was not among the known values the
+// scrubber redacts.
+//
+// `getComfyUIAuthHeaders()` carries the ComfyUI auth token and the Cloudflare
+// Access pair. It does NOT carry COMFYUI_API_KEY, which cloud-client sends as
+// `X-API-Key` — so the cloud enqueue error path, whose body is the one most
+// likely to reflect that key back, had no known-value pass for it at all.
+//
+// The by-name pass catches a LABELLED reflection ("api-key: …"). An UNLABELLED
+// echo of a short key has neither a recognized label nor the 24-character
+// opaque-run shape, and went straight out. That is the case pinned here.
+
+import { describe, expect, it, vi } from "vitest";
+
+const CLOUD_KEY = "cmfy-7Q2w9Z";      // short + unlabelled: no other pass catches it
+const HF_TOKEN = "hf_aBcD3fGh1jK";
+
+vi.mock("../../config.js", () => ({
+  getComfyUIAuthHeaders: () => ({}),   // deliberately EMPTY: the old known-value source
+  getComfyUIBaseUrl: () => "http://127.0.0.1:8188",
+  config: { comfyuiApiKey: CLOUD_KEY, huggingfaceToken: HF_TOKEN, civitaiApiToken: undefined },
+}));
+
+const { scrubSecretShapedText, bodyPrefixOf } = await import("../../comfyui/json-guard.js");
+
+describe("the scrubber knows every configured credential (#1191)", () => {
+  it("redacts an UNLABELLED reflected cloud API key", () => {
+    const out = scrubSecretShapedText(`upstream rejected ${CLOUD_KEY} at edge`);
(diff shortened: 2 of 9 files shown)


artokun:
**Codex adversarial review — two real leaks found that this PR's first pass missed.** One of them was on the very path the fix was written for, so the fix was incomplete exactly where it mattered most.

**1. The CLOUD API KEY was never a known value.** `scrubSecretShapedText`'s known-value pass iterates `getComfyUIAuthHeaders()`, which carries the ComfyUI auth token and the Cloudflare Access pair — and **not** `COMFYUI_API_KEY`, which `cloud-client` sends as `X-API-Key`. So the cloud enqueue error path, whose body is the one most likely to reflect that key back, had no known-value pass for it at all.

The by-name pass catches a *labelled* reflection (`api-key: …`). An **unlabelled** echo of a short key has neither a recognized label nor the 24-character opaque-run shape, and went straight out. The HF/CivitAI tokens are added for the same reason — which credential leaked is not a per-call-site decision.

**2. The structured branch bypassed redaction entirely.** `buildEnqueueError` parses the body **first** and copied `error.message`, each node's `class_type`, and every `errors[].message` / `.details` verbatim. A gateway answering with JSON shaped like a ComfyUI validation failure therefore leaked while the "redacted" fallback never ran. `safeField` now scrubs each field and withholds it when the scrub cannot substitute safely.

**Fail-closed must not mean fail-noisy.** A genuine ComfyUI validation error carries no credential and is returned unchanged, so the #485 diagnosis reads exactly as before — pinned by its own test.

**Also flagged, and filed rather than folded in:** `getLogs()` returns raw `/internal/logs` lines into a tool result with no scrub → **#1206**. Real, genuinely pre-existing, and not an enqueue error path.

Confirmed clean by the review: the `#485` node_errors branch does return before the raw-body fallback (so the 160-char clip cannot truncate it), and neither helper can throw for a real `Response`.

Mutation-verified, including the check that matters most — restoring the raw 500-byte body fails **both** the credential test and the length bound, so these tests genuinely catch the reported defect rather than passing either way:

| mutation | result |
|---|---|
| raw 500-byte body (the original defect) | ✗ killed (2 tests) |
| raw `statusText` | ✗ killed |
| known values back to auth-headers-only (cloud key leaks) | ✗ killed (3 tests) |
| structured fields unscrubbed (JSON-shaped leak) | ✗ killed (2 tests) |
| `safeField` withholds everything (fail-noisy) | ✗ killed (3 tests) |

Full suite: 387 files, 7891 passed. Lint, #796 gate, vocabulary gate clean.

PR #1209, opened 2026-08-09
fix(logs): scrub /internal/logs before it reaches a tool result

### LINKED ISSUE #1206: getLogs() returns RAW /internal/logs lines into a tool result — no credential scrub
Found during the adversarial review of #1191 (PR #1202). **Pre-existing and out of that PR's scope** — it is not an enqueue error path — so filing separately rather than widening the fix.

## What

`getLogs()` reads `/internal/logs` with `r.text()` and returns the split raw lines (`src/comfyui/client.ts`, the `getLogs` body). Its tool caller turns those lines directly into a text tool result (`src/tools/diagnostics.ts`, the `logs` action).

Nothing on that path goes through `bodyPrefixOf` or `scrubSecretShapedText`.

## Why it matters

ComfyUI's log is a plausible place for a credential to appear:

- a custom node that logs the URL it fetched, query string included — CivitAI and HuggingFace download URLs routinely carry a token;
- ComfyUI-Manager logging an install URL;
- any node that logs a request header on failure.

`get_system_stats action:"logs"` is a **diagnostic** tool, which means it is called precisely when something is wrong — and its output is exactly what a user pastes into a bug report. That combination is what makes this worth fixing rather than tolerating.

## Not a rewrite

The repo already has the tool. `scrubSecretShapedText` redacts by shape as well as by configured value and fails closed when it cannot substitute safely; #1191 additionally taught it the cloud API key and the download tokens, so a token echoed from a CivitAI/HF URL is now a known value.

The care is in not destroying the diagnostic: a log line is the thing being asked for, so per-line scrubbing (keeping the line, redacting the match) is right, and withholding the whole log is not.

## Related

- #1191 — the same class on the enqueue error builders (fixed).
- #828 — the same class on the diagnosis path.
- #1187 — the same class on the branches it made reachable.

artokun:
Fixes #1206. **Draft — claiming the issue, work in progress.**

## What

`getLogs()` reads `/internal/logs` with `r.text()` and returns the split raw lines; its tool caller turns those directly into a text tool result. Nothing on that path goes through `bodyPrefixOf` or `scrubSecretShapedText`.

## Why it matters

A ComfyUI log is a plausible place for a credential to appear:

- a custom node logging the URL it fetched, query string included — CivitAI and HuggingFace download URLs routinely carry a token;
- ComfyUI-Manager logging an install URL;
- any node that logs a request header on failure.

`get_system_stats action:"logs"` is a **diagnostic** tool, so it is called precisely when something is wrong — and its output is exactly what a user pastes into a bug report.

Found during the adversarial review of #1191; genuinely pre-existing, which is why it was filed separately rather than widened into that PR.

## Plan

Per-line scrubbing. The log line **is** the thing being asked for, so the fix must keep the line and redact the match — withholding the whole log would trade one defect for another, and a diagnostic that refuses to diagnose is not an improvement.

`scrubSecretShapedText` already redacts by shape as well as by configured value and fails closed when it cannot substitute safely; #1191 additionally taught it the cloud API key and the download tokens, so a token echoed from a CivitAI/HF URL is now a known value.

The care is in the fail-closed case: a line it cannot scrub safely must be replaced rather than dropped silently, so the log stays readable and the reader knows something was withheld.

### DIFF
--- a/src/__tests__/comfyui/log-scrub-withhold.test.ts
+++ b/src/__tests__/comfyui/log-scrub-withhold.test.ts
@@ -1,0 +1,40 @@
+// #1206 — the FAIL-CLOSED half of the log scrub, in its own file because it
+// needs a different configured credential than the redaction tests.
+//
+// `scrubSecretShapedText` returns null when it cannot substitute a configured
+// credential without mangling unrelated text — documented trigger: a credential
+// shorter than 8 characters. That must REPLACE the line with a marker, never
+// delete it: a shorter log with no explanation is a worse diagnostic than one
+// that says what it withheld, and a caller counting lines would be misled.
+
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("../../config.js", () => ({
+  // Deliberately SHORT — this is the documented "too short to replace safely"
+  // case, and it is the only reliable way to reach the null branch.
+  getComfyUIAuthHeaders: () => ({ Authorization: "abc" }),
+  getComfyUIBaseUrl: () => "http://127.0.0.1:8188",
+  config: {},
+}));
+
+const { scrubLogLines } = await import("../../comfyui/json-guard.js");
+
+describe("a line that cannot be scrubbed is withheld VISIBLY (#1206)", () => {
+  it("replaces the line with a marker rather than emptying it", () => {
+    const out = scrubLogLines(["[INFO] header was abc"]);
+    expect(out).toHaveLength(1);
+    expect(out[0]).toMatch(/withheld/);
+    // An empty string would satisfy "no credential" while silently deleting the
+    // entry — the failure mode this wording exists to prevent.
+    expect(out[0].length).toBeGreaterThan(0);
+    expect(out[0]).not.toContain("header was abc");
+  });
+
+  it("withholds only the affected line, never the whole log", () => {
+    const out = scrubLogLines(["fine one", "[INFO] header was abc", "fine two"]);
+    expect(out).toHaveLength(3);
+    expect(out[0]).toBe("fine one");
+    expect(out[2]).toBe("fine two");
+    expect(out[1]).toMatch(/withheld/);
+  });
+});
--- a/src/__tests__/comfyui/log-scrub.test.ts
+++ b/src/__tests__/comfyui/log-scrub.test.ts
@@ -1,0 +1,93 @@
+// #1206 — /internal/logs reached a tool result with no credential scrub.
+//
+// A ComfyUI log is a plausible place for one to appear: a custom node logging
+// the URL it fetched (CivitAI and HuggingFace download URLs routinely carry a
+// token in the query string), ComfyUI-Manager logging an install URL, or any
+// node logging a request header on failure.
+//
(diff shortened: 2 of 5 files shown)


artokun:
## Review — disclosure first

**No Codex findings informed this merge.** It was dispatched and ran, but did not return findings within its execution window; the forwarding agent correctly declined to guess rather than manufacture an answer. What follows is my own adversarial pass, labelled as such.

That pass did earn its keep — it caught a real defect *before* the first commit, described below.

## The defect my own testing caught

Running `scrubSecretShapedText` over a whole log line **destroys it**. Its opaque-run pass matches 24+ characters of an alphabet that includes `/`, `.` and `-`, so an ordinary log URL collapses to `https:«redacted»` — and a log line whose URL is gone has lost the fact it exists to report ("which model was it fetching?"). Nearly every ComfyUI line carrying a URL would have been gutted.

The repo had already solved this one layer over: `redactUrlForDiagnosis` redacts by URL **structure**, keeping origin, route and parameter names while failing closed on userinfo, opaque path segments and unknown query values. So a line is now **split** — URLs to the URL redactor, the gaps between them to the shape scrubber — and neither pass sees text the other owns:

```
[INFO] fetching https://huggingface.co/org/repo/resolve/main/model.safetensors?token=«redacted»
RuntimeError: CUDA out of memory
```

## The questions I set for the review, answered

**Awkward URL shapes** — eight probed, all safe, and now pinned as a permanent test that asserts *no shape leaks* rather than which pass caught it: parenthesised, trailing period, backticked, `ws://` (URL redactor refuses it → falls to the shape pass), scheme-relative `//host/…` (never matches the regex → shape pass), userinfo, fragment, and a bare token with no URL.

**A non-URL secret glued into a URL** (`?x=1&hdr=…`) — safe: it is inside the URL match, and unknown query values are redacted.

**Performance** — `/internal/logs` is a bounded deque (~300 lines), so per-line work is trivial.

## Residual, stated rather than papered over

A credential **split across two log lines** is only partially handled. The line-wrap detection inside `scrubSecretShapedText` operates per call, so with per-line scrubbing each half is judged alone. In practice the labelled half still redacts (`token=«redacted»`) and only an unlabelled tail fragment survives — not a usable credential — but it is not nothing, and I would rather record it than claim complete coverage.

Per-line is still right: scrubbing the whole blob as one string would let a single unscrubbable line withhold the **entire** log, and a diagnostic that refuses to diagnose is not an improvement.

## Testing

| mutation | result |
|---|---|
| no scrub at all (the reported leak) | ✗ killed (2 tests) |
| shape pass over URLs too (URL collapses) | ✗ killed |
| unscrubbable line emptied instead of marked | ✗ killed (2 tests) |
| unscrubbable line dropped from the list | ✗ killed (2 tests) |
| `getLogs` stops scrubbing | ✗ killed |
| health report stops scrubbing | ✗ killed |

**A second test bug worth naming:** my first fail-closed fixture used a zero-width space, which `\s` does not match — so it never reached the null branch and passed for the wrong reason. That case now lives in its own file with a short configured credential, the documented trigger.

Full suite: 390 files, 7911 passed. Lint, #796 gate and vocabulary gate clean; CI green on all three platforms.

PR #1211, opened 2026-08-09
fix(#1204): typecheck the tests, so a fixture cannot claim a shape the writer never produces

### LINKED ISSUE #1204: tsconfig excludes src/__tests__, so no test fixture is ever typechecked — the root enabler of the wrong-key bug class
Found during the adversarial review of #1200 (issue #1197). Filing separately because it is repo-wide and it is the *reason* two separate wrong-key bugs survived in one function.

## The gap

`tsconfig.json` has `"exclude": ["node_modules", "dist", "src/__tests__"]`.

```
$ npx tsc --listFiles --noEmit | grep -c "__tests__"
0
```

**Zero test files are typechecked.** A fixture can claim any shape and nothing verifies it against the interface it purports to model.

## What it let through

In `migrateInFlightJobs` alone, two instances in the same object literal:

1. **`tray_id` vs `trayId`.** The persisted record's one camelCase key, and the only key url lookup matches on. The migration read `str("tray_id")` → always `undefined` in production. The test passed because its fixture used the same wrong key.
2. **Five keys from the WRONG INTERFACE** — `name`, `dest`, `target`, `total`, `received` belong to `DownloadProgress` (the tray rows), not `PersistedDownloadJob`. They were copied for months. One test even asserted `received` was 700000000, commented *"the bytes it had, so a reader can see what was lost"*, on a field that is always undefined.

In both cases the fixture agreed with the mistake, so a green suite meant nothing.

## Why runtime tests cannot close this

Worth stating, because it is not obvious and I got it wrong first: a runtime assertion **cannot** catch these. The dead keys are always `undefined`, and `JSON.stringify` drops undefined keys — so they never reach disk, and no assertion on the round-tripped record can see them. I wrote a key-set assertion comparing the migrated record against the writer's own output; restoring all five dead keys still passed it.

Only the type layer catches it. Annotating the literal (`const rec: PersistedDownloadJob = {...}`) turns a dead key into `TS2353: Object literal may only specify known properties`. That is now done for this one function — and it immediately caught a latent bug on contact: `url`, `target_subfolder` and `started_at` are REQUIRED on that interface and the migration could emit them as `undefined`, so it had been producing silently-invalid records.

`persistDownloadJob` cannot help either: it does `const safe: PersistedDownloadJob = { ...job, ... }`, and spreading a *variable* defeats excess-property checking, so the writer imposes no schema on its callers.

## Suggested direction

Typecheck the tests. Either drop `src/__tests__` from `exclude`, or add a second `tsconfig.test.json` that includes them and wire it into `npm run lint` / CI. Expect a first-run backlog — that backlog *is* the finding, since every fixture in it is currently unverified against the type it models.

A narrower interim step, if the full sweep is too disruptive: annotate the object literals that construct persisted records, which is where a wrong key silently becomes data loss rather than a test failure.

Environment: main at 0.50.54; platform-independent (build configuration).

Related (same root cause): #1197

artokun:
Fixes #1204. **Draft — claiming the issue.**

## The gap

`tsconfig.json` excludes `src/__tests__`, so `npx tsc --listFiles --noEmit | grep -c __tests__` returns **0**. No fixture is ever checked against the interface it claims to model.

## What it let through, today alone

Two wrong-key bugs in a single object literal in `migrateInFlightJobs`:

1. **`tray_id` vs `trayId`** — the record's one camelCase key, and the only key url lookup matches on. The migration copied nothing in production; the test passed because its fixture used the same wrong key.
2. **Five keys from the wrong interface** — `name`, `dest`, `target`, `total`, `received` belong to `DownloadProgress` (tray rows), not `PersistedDownloadJob`. They survived for months, and a test even asserted `received` was 700000000 on a field that is always `undefined`.

In both cases the fixture agreed with the mistake, so green meant nothing.

## Why only the type layer closes it

Worth stating because I got it wrong first: a runtime assertion **cannot** catch a dead key. It is always `undefined`, and `JSON.stringify` drops undefined before it reaches disk — I wrote a key-set assertion against the round-tripped record and restoring all five dead keys still passed. Annotating the literal turns it into `TS2353` immediately, and that annotation caught a latent bug on contact (three required fields the migration could emit as `undefined`).

## Plan

- [ ] Include `src/__tests__` in typechecking — either drop the exclude or add a `tsconfig.test.json` wired into `npm run lint` and CI
- [ ] Measure the backlog first: it IS the finding, since every error is a fixture currently unverified against its type
- [ ] Triage the backlog — real shape mismatches get fixed; deliberate partial doubles get an explicit, commented cast rather than a silent one
- [ ] Do NOT make it pass by loosening types or blanket-casting; that reproduces the gap with extra steps

Draft until the backlog is measured, since its size decides whether this lands in one pass or needs staging.

### DIFF
--- a/src/__tests__/services/panel-sync.test.ts
+++ b/src/__tests__/services/panel-sync.test.ts
@@ -398,9 +398,26 @@ function makeDeps(opts: {
     },
     // #1204 — this guard used to be on `gitPullFfOnly`, which is NOT a member of
     // PanelInstallerDeps: nothing ever called it, so the assertion could not
-    // fire while the REAL pull path (gitFetch → gitMergeFfOnly) went unguarded.
-    // A safety net that cannot catch anything is worse than none, because it
-    // reads as coverage.
+    // fire at all.
+    //
+    // It is replaced by the deps the fallback ACTUALLY touches, in the order it
+    // touches them. To be accurate about what that buys: the entry point was
+    // already covered — `gitStatusPorcelain` (guarded above) runs at
+    // panel-installer.ts:2675, before `gitFetch` at :2704, and there is no route
+    // to the pull that skips it. So these are belt-and-braces on the only
+    // reachable path, not a hole being closed. `gitWorktreeRoot` is the genuinely
+    // new one: it is the FIRST dep the fallback calls (:2652), and its absence
+    // was also leaving this literal short of PanelInstallerDeps's required
+    // members — a TS2739 the excess `gitPullFfOnly` had been masking.
+    gitWorktreeRoot: () => {
+      throw new Error("git fallback must not run in this persona (not a git checkout)");
+    },
+    gitUpstreamRev: () => {
+      throw new Error("git fallback must not run in this persona (not a git checkout)");
+    },
+    gitIgnoredPullConflicts: () => {
+      throw new Error("git fallback must not run in this persona (not a git checkout)");
+    },
     gitFetch: () => {
       throw new Error("git fallback must not run in this persona (not a git checkout)");
     },
--- a/src/__tests__/services/panel-sync.test.ts
+++ b/src/__tests__/services/panel-sync.test.ts
@@ -396,7 +396,15 @@ function makeDeps(opts: {
     gitStatusPorcelain: () => {
       throw new Error("git fallback must not run in this persona (not a git checkout)");
     },
-    gitPullFfOnly: () => {
+    // #1204 — this guard used to be on `gitPullFfOnly`, which is NOT a member of
+    // PanelInstallerDeps: nothing ever called it, so the assertion could not
+    // fire while the REAL pull path (gitFetch → gitMergeFfOnly) went unguarded.
+    // A safety net that cannot catch anything is worse than none, because it
+    // reads as coverage.
+    gitFetch: () => {
+      throw new Error("git fallback must not run in this persona (not a git checkout)");
+    },
+    gitMergeFfOnly: () => {
       throw new Error("git fallback must not run in this persona (not a git checkout)");
     },
     readPin: () => opts.pin ?? UNPINNED,
--- a/src/__tests__/services/training-jobs.test.ts
+++ b/src/__tests__/services/training-jobs.test.ts
@@ -1,4 +1,8 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+// #1204 — import the REAL progress type instead of restating a narrower one.
+// The local shape omitted `loss`, which `training-jobs.ts:1654` actually reads,
+// so six tests were passing a field their own annotation said could not exist.
+import type { TrainingProgress } from "../../services/ai-toolkit.js";
 import { existsSync, mkdirSync, mkdtempSync, readdirSync, readFileSync, rmSync, symlinkSync, utimesSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
@@ -377,7 +381,7 @@ describe("startTrainingJob", () => {
     vi.useFakeTimers();
     try {
       const d = deferred<{ code: number; tail: string }>();
-      let tick: ((p: { step?: number; totalSteps?: number; raw: string }) => void) | undefined;
+      let tick: ((p: TrainingProgress) => void) | undefined;
       const dataset = join(mod.datasetsRoot(), "dataset");
       mkdirSync(dataset, { recursive: true });
       makeImage(dataset, "img_00001.png");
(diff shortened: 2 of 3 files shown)


artokun:
**Stage 1 of #1204.** Backlog measured at **370** errors when `src/__tests__` is typechecked; this takes the TS2353 class (a fixture claiming a shape the type does not have) from **11 → 3**. Typechecking is deliberately NOT wired into lint/CI yet — it cannot be green until the backlog is worked, and a config that always fails gets switched off.

Three fixes, three distinct shapes, all invisible for the same reason:

| Site | Shape |
|---|---|
| `training-jobs.test.ts` ×6 | mock **narrowed** a real type — `onProgress` omitted `loss`, which `training-jobs.ts:1654` reads |
| `model-management.test.ts` | mock return **inferred from its default literal**, narrower than the real verdict; `verifiedAgainstRoot` is real and consumed at `download-jobs.ts:623` |
| `panel-sync.test.ts` | mock **invented a member** — `gitPullFfOnly` exists nowhere in `src` |

I filed #1204 expecting fixtures that *claim* shapes the code lacks. Two of the four instances were the reverse — fixtures *denying* fields the code has. Both directions are equally invisible.

**One nuance on the third fix, since a green suite does not establish it.** The old guard set `gitPullFfOnly` to throw "git fallback must not run in this persona"; nothing ever called it. I moved it onto the real path (`gitFetch` → `gitMergeFfOnly`). I verified `deps.gitFetch` is reached **only** inside the git-fallback branch (`panel-installer.ts:2704`), so it is correctly placed — but it sits inside a `try/catch` that converts throws into a `PanelInstallError`, so the guard surfaces as that error rather than as a raw loud failure. Better than a guard that cannot fire; not literally "fail the test loudly" as the original comment claimed.

No production code is touched. Suite: **389 files, 7911 passed, 2 skipped**; `tsc --noEmit` clean on src.

The three remaining TS2353s are each documented on #1204 as their own question. One is not stage-1 work: `panel-restart-legacy-fallback.test.ts` mocks a `RestartResult` missing `startup`, which the real type marks *"REQUIRED on every path (#367)"* — those tests exercise a restart outcome that cannot occur, and typing the mock honestly will cascade into the assertions.

artokun:
## Review: SHIP, with two corrections applied

Full adversarial review came back **SHIP** — three real defects, correctly diagnosed and fixed, provably runtime-inert (0 production files), 139 tests green, CI untouched by design. It also caught four overstatements in my own commit messages, and one of them was substantive. Recording all of them, because half the value of #1204 is not repeating this class.

### Substantive — fixed in `d1c1c9a`

My comment claimed the *"REAL pull path (gitFetch → gitMergeFfOnly) went unguarded"*. **It was not unguarded.** `gitStatusPorcelain` — already guarded, unchanged — runs at `panel-installer.ts:2675`, before `gitFetch` at `:2704`, and there is no route to the pull that skips it. My new guards are belt-and-braces on the only reachable route.

That is a comment asserting coverage that does not exist, in the branch whose entire subject is comments asserting things that are not true. Corrected to say what the guards actually buy.

The reviewer also spotted that removing the phantom `gitPullFfOnly` **unmasked a pre-existing TS2739** — `gitWorktreeRoot`, `gitUpstreamRev`, `gitIgnoredPullConflicts` are required members and were absent, so TS had been reporting the excess property instead. Net movement on that commit was therefore **zero**, not −1. Adding all three as throwing guards fixes the error *and* covers the fallback's real entry point (`gitWorktreeRoot` is the first dep it calls, `:2652`).

Re-measured with tests included: **370 → 361**, TS2353 **11 → 3**, TS2739 **1 → 0**, panel-sync now contributes zero errors. `npx tsc --noEmit` clean; 51/51 green.

### Commit-message overstatements — no code impact, corrected here

- `1645197` says the six tests pass `loss` *"while asserting it propagates to `job.progress.loss`"*. **No test in the file reads `loss`.** The `loss:` in the fixtures is realism — `parseTrainingProgress` only assigns step/totalSteps when a loss is present — not an assertion. The type fix is still right (`training-jobs.ts:1654` reads `p.loss`, the local restatement omitted it); the coverage claim was not.
- "six tests" is **five** — the 6 violations sit at six call sites across five tests, two of them in the rollback test. The sixth annotation I widened had no violating call; widening it for consistency is fine, counting it as a defect isn't.
- `a2957f4` is **empty** (0 files). Collapsed by the squash merge.

### Confirmed sound

`Awaited<ReturnType<typeof import(...)>>` verified by transform, not by argument: the reviewer ran the file through esbuild and grepped the emitted JS — the alias emits nothing, no runtime import, no cycle, nothing `vi.mock` hoisting can observe. It also killed a second latent error (`model-management.test.ts(208,15)` TS2345) that I hadn't noticed. Fair nit: `LandedModelVerification` is already exported, so a plain `import type` would be equivalent and plainer — I'm keeping the indirect form because it tracks the function rather than the name, which is the property that was missing in the first place.

Issue #467, opened 2026-07-30
download_model resume silently discards a 96%-complete .partial with no warning when Xet/CAS CDN omits ETag/Last-Modified


artokun:
### What happened / root cause
Downloading a large model (qwen_image_edit_2511_bf16.safetensors, 38.05GB) from Hugging Face via `download_model`. The MCP server process self-restarted mid-download (unrelated self-update/restart event). At the moment of restart, `.comfyui-mcp/cache/<hash>.safetensors.partial` had reached 39,157,137,284 bytes (~96% complete, ~36.5GB).

After the restart, re-issuing the exact same `download_model` call (same URL/target) reused the same job id but started downloading from 0% instead of resuming — the entire 36.5GB partial was discarded silently, with zero log/warning explaining why, and no signal in the tool's response that a resume was attempted-and-declined vs never-attempted.

Traced the root cause in `dist/services/download-cache.js` (`streamUrlToFile`): resume is only attempted when a validator sidecar (`${partial}.etag`, holding an ETag or Last-Modified value) exists next to the `.partial` file. That sidecar is only written when the *original* download response included an `ETag` or `Last-Modified` header (see the `resumable && !appendMode` branch). In this case no `.etag` sidecar was ever created — Hugging Face's Xet/CAS CDN (the backend that serves large HF model files, e.g. via a `cas-bridge.xethub.hf.co` redirect) apparently did not return either header on the original response. So on the retry, `readValidatorSidecar` returns null, `effectiveResume` is silently reset to `0`, and the code falls back to a full restart — which is the deliberate/documented safe behavior (see the code's own comment referencing issue #343: an unverified resume risks corrupting the file if the upstream changed). The *logic* is intentionally safety-first and correct. The *gap* is that this fallback is completely silent: no log line, and nothing surfaces to the calling agent/user that ~36GB of already-downloaded data was just discarded and why. From the outside it looks identical to "never got a validator, resume was never possible" vs "resume was possible but something else went wrong" — indistinguishable, and costs a full multi-GB, multi-hour re-download with no diagnostic trail.

### Steps to reproduce
1. `download_model` a large file from a Hugging Face repo whose resolve URL redirects to the Xet/CAS CDN (e.g. any `Comfy-Org/*_ComfyUI` split_files repo).
2. While the transfer is in flight (well before completion), restart/kill the comfyui-mcp server process (e.g. via its self-restart mechanism, or a manual kill).
[...]

Issue #1208, opened 2026-08-09
Flaky test: "lists newest first" relies on a 2 ms gap between two download jobs


artokun:
Observed blocking the `v0.50.56` release build — **all three platforms failed the same assertion**, and an unchanged re-run went green.

## The test

`src/__tests__/services/download-jobs.test.ts`

```ts
it("lists newest first", async () => {
  await startDownloadJob(URL_A, "checkpoints");
  await new Promise((r) => setTimeout(r, 2));
  await startDownloadJob(URL_B, "loras");
  expect(listDownloadJobs()[0].url).toBe(URL_B);
});
```

`listDownloadJobs()` sorts with `b.started_at - a.started_at`, a **millisecond** timestamp. A 2 ms sleep is a thin margin, and the comparator has no tiebreak — so when both jobs land in the same millisecond the order is whatever `Map` iteration yields, which is stable locally and evidently not stable on a loaded CI runner.

## Why it deserves a fix rather than a re-run

A flake that fails **all three platforms at once** does not read like a flake — it reads like a real regression, and that is what it cost here: the release build failed, and ruling my change out meant reading the diff, confirming `download-jobs.ts` was untouched, and checking the failure history. A test that cries wolf on a release branch is expensive even when it is wrong.

There is also a second-order risk: `listDownloadJobs()` folds in `listPersistedDownloadJobs()`, so a record left by a neighbouring test can join this list. That makes the ordering sensitive to test **scheduling**, not just to the clock — which would explain a simultaneous multi-platform failure better than clock granularity alone.

## Suggested fix

Do not race the clock:

- give the comparator a deterministic tiebreak (e.g. `id`/`trayId`) so equal timestamps still order stably; **and**
- have the test assert on something it controls — e.g. stub the clock, or assert set membership plus an explicit `started_at` relationship — rather than on a 2 ms wall-clock gap.

Not urgent, but it will keep costing a re-run and a diff read every time it fires.

Found while releasing #1191 (PR #1202 / #1207).

artokun:
**A second flake, in a different test.** Recording it here rather than opening a near-duplicate, since the ask is the same: these cost a re-run and a diff read every time.

> `an UNREADABLE disk version leads with the two free checks, not with an update`

Failed once during a local full-suite run and passed on an immediate unchanged re-run (392 files, 7926 tests). I have not traced it; noting it so the next person to touch flakes has both.

[...]

PR #1213, opened 2026-08-09
fix(download): status must not report a LIVE download as missing

### LINKED ISSUE #1183: download_model action:"status" falsely reports "No download matching id" mid-transfer, hours into a live 26GB local download

Started a local model download via `download_model action:"download"` (target_subfolder: checkpoints, url: a HuggingFace safetensors, ~26.5GB). Polled `action:"status"` with the returned `id` repeatedly over ~30 minutes as it progressed normally: 20% -> 67% -> 79% -> 80% -> 85% -> 88% -> 89% -> 95%. On the NEXT poll (same id, no other args), status returned:

> No download matching id `<id>`. It has either finished long ago (settled records are pruned after a while) or never started — check the panel download tray before re-downloading.

The download had NOT finished (no file existed anywhere in any ComfyUI models directory — I searched the whole filesystem) and had NOT been cancelled. Re-issuing `action:"download"` with the exact same URL immediately re-adopted the SAME id and reported it "still running (97%)" — so the job was alive in-process the entire time; only the status-by-id lookup produced a false negative for one poll.

Traced this in the installed package (`dist/services/download-jobs.js`). `getDownloadJob(id, trayId)` (no `trayId` passed, matching how `action:"status"` calls it with only `id`) does, in order:
1. `jobs.get(id)?.job` (in-memory registry) — if live+`downloading`, checks `hasAmbiguousForeignSibling(id, live.trayId)` (line ~685) and returns `undefined` if a persisted record shares the id with a different `trayId`/`owner` even when there is really only one physical transfer.
2. Falls through to the persisted-store path (lines ~889-908) if the in-memory hit is absent — checks `matches.filter(status==="downloading" && fresh heartbeat)` and again bails to `undefined` if it sees >1 distinct `trayId` ("ambiguous live").

The in-memory index itself is also non-trivial: `registerKey`/`retireEntry` (lines ~120-143) key the `jobs` Map by request/dest key and only delete an index row if the map still points at that exact entry object — so a key can plausibly get re-pointed or transiently miss depending on timing between the two indices (in-memory `jobs` Map vs. the persisted store) whenever both are consulted in the same lookup. I could not fully pin the exact race in one bounded pass (see below), but the effect reproduced cleanly once: a solo, genuinely-still-downloading job momentarily failed a by-id lookup and was reported as gone.

### Steps to reproduce
1. `download_model action:"download"` a large (20GB+) file to a LOCAL ComfyUI so the transfer takes many minutes.
2. Poll `action:"status", id:"<id>"` repeatedly (every few minutes) while it's genuinely still downloading.
3. At some point (observed once, ~89-95% in, ~28 minutes into the transfer) a poll returns "No download matching id" instead of the in-progress percentage.
4. Re-issue `action:"download"` with the same URL — it immediately re-adopts the same id and reports it still running at the correct (advanced) percentage, proving the job was never actually gone.

### Exact error (scrubbed)
```
No download matching id `cb43b8c206bec9b1`. It has either finished long ago (settled records are pruned after a while) or never started — check the panel download tray before re-downloading. Within the SAME session, re-issuing an identical in-flight download adopts it rather than duplicating; across a reconnect, confirm via the tray first.
```
(id shown is this download's own id, not a secret — included for reproducibility.)

### Impact
An agent (or the panel tray) trusting this response mid-download will wrongly conclude the transfer failed/vanished and may: report a false failure to the user, or worse, start a REDUNDANT multi-GB download (the docs say re-issuing "adopts" an in-flight job rather than duplicating, which is what saved this case — but that safety net depends on the caller re-issuing with the exact same URL/args, which an agent reacting to "not found" has no reason to do).

### Fix
<applied locally: no> <upstream-only: yes>

I reviewed `dist/services/download-jobs.js` (`getDownloadJob`, `hasAmbiguousForeignSibling`, `registerKey`/`retireEntry`) but this is a job-registry consistency issue between an in-memory `Map` and a persisted on-disk store, with disambiguation-by-owner/trayId logic that I could not safely change in one bounded, minimal, reversible edit without risking the existing anti-duplication/anti-clobber guarantees (issues #420, #467, #822, #529 are all referenced in comments around this code, so this area has already been hardened multiple times against related races — a shallow patch felt more likely to reintroduce one of those than to cleanly fix this one). Precise area to fix: `getDownloadJob()` in `download-jobs.js` should not return `undefined` for a by-id (no trayId) lookup when there is in fact only ONE physically-live job for that id — the "ambiguous foreign sibling" / "ambiguous live" bail-out paths (`hasAmbiguousForeignSibling`, and the `foreignLive` trayId-Set-size check around line 901) appear to be triggering in a single-job case, possibly due to an owner/trayId mismatch between this session's in-memory record and a persisted record for the SAME physical transfer (rather than a genuine second concurrent transfer), or a transient window where the in-memory registry key was momentarily unindexed. Suggest: (a) verify with a repro harness whether `hasAmbiguousForeignSibling`/the `foreignLive` check can be tripped by a single real transfer whose persisted `owner`/`trayId` diverges from the querying session's, and if so, treat "single physical destination file, still within the fresh-heartbeat window" as authoritative over an owner/trayId mismatch; (b) as a cheaper harm-reduction fix regardless of root cause, have `action:"status"` fall back to checking `findDownloadJob({url})`-style resolution (which DID find it) before returning the terminal "No download matching" text, so a caller who has the original URL handy (common right after starting a download) gets a correct answer even if the id-only path has a blind spot.

### Environment
OS: Windows 11 Pro (10.0.26200) | ComfyUI: 0.31.1 | ComfyUI frontend: 1.48.7 | GPU: NVIDIA GeForce RTX 5090, 32GB VRAM | comfyui-mcp: 0.50.47 | panel: 0.11.38

artokun:
Fixes #1183. **Draft — claiming the issue, work in progress.**

## The report

A local 26.5GB download polled normally for ~30 minutes — 20% → 67% → 79% → 85% → 89% → 95% — and then one poll of `action:"status"` with the same id returned:

> No download matching id `cb43b8c206bec9b1`. It has either finished long ago … or never started

The file did not exist anywhere on the filesystem, and nothing had been cancelled. Re-issuing `action:"download"` with the same URL **immediately re-adopted the same id and reported 97%** — so the job was alive in-process the whole time. Only the by-id lookup produced a false negative, for one poll.

## Why it matters

An agent that trusts it mid-download will report a false failure, or start a **redundant multi-GB download**. The anti-duplication safety net only holds if the caller re-issues with the exact same URL — which an agent reacting to "not found" has no reason to do.

## What the reporter already established

They traced it to `getDownloadJob(id)` in `download-jobs.js` and named the suspects — the `hasAmbiguousForeignSibling` / `foreignLive` trayId-set bail-outs firing for what is really a **single** physical transfer, most likely an owner/trayId divergence between the in-memory record and the persisted one for the same download.

They also explained why they did not patch it, and the reasoning is right: this area has been hardened repeatedly (#420, #467, #822, #529) and a shallow fix risks reintroducing one of those races.

## Plan

Reproduce first with a harness that drives the real registry, rather than reasoning from the code — the whole point of the ambiguity guards is that they are subtle, and I want to see which branch actually fires before touching any of them.

The fix must not weaken the guarantee those guards exist for: a genuinely ambiguous id (two concurrent physical transfers sharing an id) must still refuse rather than guess. The target is the case where ambiguity is *claimed* but only one live job exists.

### DIFF
--- a/src/__tests__/services/download-status-decline.test.ts
+++ b/src/__tests__/services/download-status-decline.test.ts
@@ -88,9 +88,14 @@ describe("describeUnresolvedDownload (#1183)", () => {
     expect(mod.describeUnresolvedDownload("done1")).toBeUndefined();
   });
 
-  it("says nothing for a STALE in-flight record — the writer is gone", () => {
-    // Beyond the freshness window this is not evidence of a live transfer, and
-    // claiming it is would be the same defect in the other direction.
+  it("REPORTS a stale in-flight record, hedged — a missed heartbeat is not proof", () => {
+    // This assertion used to demand silence here, and that was the bug in my own
+    // first version: it used the same 60s window the lookup uses, so it went
+    // quiet in exactly the situation most likely to have produced the reporter's
+    // one-poll miss — a heartbeat delayed while the writer was busy at 90% of a
+    // 26GB file. The codebase already states the rule (#761): a missed heartbeat
+    // is a liveness HINT, not proof the transfer stopped, which is why in-flight
+    // records are retained for 6h rather than reaped at 60s.
     writeFileSync(
       join(dir, `${progress.CONTROL_PREFIX}job-stale1-ownerA.json`),
       JSON.stringify({
@@ -101,7 +106,14 @@ describe("describeUnresolvedDownload (#1183)", () => {
         updated: Date.now() - progress.PERSISTED_INFLIGHT_STALE_MS - 1000,
       }),
     );
-    expect(mod.describeUnresolvedDownload("stale1")).toBeUndefined();
+    const note = mod.describeUnresolvedDownload("stale1");
+    expect(note).toBeDefined();
+    expect(note).toMatch(/stopped reporting/);
+    expect(note).toMatch(/liveness HINT, not proof/);
+    expect(note).toMatch(/NOT the same as it being gone/);
+    // Hedged, not asserted: it must not claim the transfer IS running.
+    expect(note).not.toMatch(/IS still running/);
+    expect(note).toMatch(/BEFORE re-downloading/);
   });
 });
 
--- a/src/__tests__/services/download-status-decline.test.ts
+++ b/src/__tests__/services/download-status-decline.test.ts
@@ -1,0 +1,130 @@
+// #1183 — a DECLINE is not an ABSENCE.
+//
+// A reporter polled a live 26.5GB local download for ~30 minutes (20% → 67% →
+// 79% → 85% → 89% → 95%) and then, for ONE poll, got:
+//
+//   No download matching id `cb43b8c206bec9b1`. It has either finished long ago
+//   … or never started
+//
+// The file existed nowhere on disk and nothing had been cancelled. Re-issuing the
+// same URL immediately re-adopted the SAME id at 97% — the job had been alive in
+// process the whole time.
+//
+// getDownloadJob DECLINES rather than guessing when two live transfers share an
+// id, and that refusal is right: picking one arbitrarily is how you report the
+// wrong transfer's progress. What is wrong is rendering the refusal as an
+// absence. "I will not choose between two" and "there is nothing here" are
+// different facts, and only the second justifies telling someone their download
+// vanished — which is what invites a redundant multi-gigabyte re-download.
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+let dir: string;
+let mod: typeof import("../../services/download-jobs.js");
+let progress: typeof import("../../services/download-progress.js");
+
+beforeEach(async () => {
+  dir = mkdtempSync(join(tmpdir(), "cm-dl-decline-"));
+  vi.resetModules();
+  vi.stubEnv("COMFYUI_MCP_PROGRESS_DIR", dir);
+  progress = await import("../../services/download-progress.js");
+  mod = await import("../../services/download-jobs.js");
+});
+afterEach(() => {
+  vi.unstubAllEnvs();
+  rmSync(dir, { recursive: true, force: true });
+});
+
+/** A persisted in-flight record, as a foreign session's heartbeat would write it. */
+function writeLive(id: string, trayId: string, owner: string) {
+  writeFileSync(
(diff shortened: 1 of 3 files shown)


artokun:
## Review — disclosure first

**No Codex findings informed this merge.** The task ran but the relay returned no content, so there is nothing to report from it. What follows is my own adversarial pass, labelled as such.

It found something important, so it was worth doing properly.

## The decisive question, and it nearly sank the PR

I set the review the question *"does this actually fix the REPORTED case, or is it a message improvement that misses it?"* — and the honest answer was **probably the latter**.

The explainer used the **same 60-second freshness window the lookup uses**. So it went silent in exactly the situation most likely to have produced a one-poll miss during a 26GB transfer: a heartbeat delayed past 60s while the writer was busy at ~90%. In that case the reporter would *still* have been told "No download matching id", and this PR would have improved the message for a case they did not hit.

The codebase already states the right rule one file over (#761):

> A missed heartbeat is only a liveness hint, not proof the transfer stopped

— which is why an in-flight record is retained for **6 hours** rather than reaped at 60 seconds. Using the 60s window here repeated the very fold this fix exists to remove, one level down.

A stale in-flight record is now reported, **hedged and distinctly**: it says the record stopped reporting N seconds ago, that a missed heartbeat is a hint rather than proof, and to check the tray or list before re-downloading. It deliberately does **not** say "IS still running" — that is the opposite over-claim, and there is a mutation pinning it.

## Scope, deliberately narrow

The ambiguity guards are **untouched**. Nothing about *which* job the lookup returns has changed — only what is said when it returns nothing. The reporter's own reasoning for not patching those guards was right: the area has been hardened repeatedly (#420, #467, #822, #529) and a shallow change risks reintroducing one of those races.

## Testing

| mutation | result |
|---|---|
| explainer always silent (false "vanished" returns) | ✗ killed (2 tests) |
| a settled record counted as live | ✗ killed |
| a stale record counted as *definitely* live | ✗ killed |
| **stale records ignored again** (the reported case uncovered) | ✗ killed |
| stale hedge upgraded to "IS still running" | ✗ killed |
| **renderer drops the explainer** | ✗ killed |

The last row survived until a wiring test existed — the call-site blindness this repo keeps getting caught by, and worth stating rather than quietly adding.

Full suite: 392 files, 7926 passed. Lint, #796 gate and vocabulary gate clean.

**One process note:** I committed once while a test was red, then found it was an unrelated flake (recorded on #1208). That was sloppy — the suite is green now and I re-verified before pushing.

PR #1214, opened 2026-08-09
fix(bridge): surface a late reply to a timed-out mutation instead of dropping it (#694)

artokun:
Closes the recoverability half of #694.

Most of #694 shipped in #704/#707 (retry_of, the #521 ledger, session epoch). The reopen was a recurrence #1154 explained. The owner's note scoped what remained, and it is this PR's whole boundary:

> this reduces how often the ambiguity happens, it does not make it recoverable

## The gap

A mutation that times out and then applies is invisible. The panel *does* reply — just after the deadline — and the message loop drops it. `ui-bridge.ts:2230` says so outright: **"Everything else drops."** The reporter's case: `graph_set_node_mode` timed out at 6000 ms, `panel_query_graph` immediately after showed the node already switched. The write landed. Nothing ever said so.

## Two commits

**1 — retention.** Two bounded maps: `timedOutMutations` (written only when the reply timer fires, only for `ctx.mutating`) and `lateMutations`, drained once. 10-min TTL + 256 cap.

The negatives are the design:
- **Reads excluded.** #1154's asymmetry — an abandoned read costs nothing, you retry it. Retaining reads is unbounded noise.
- **Late `ok:false` not retained.** The caller was already told the write didn't complete, and that was true. Promoting a late refusal to "it failed" converts *we never found out* into a verdict we didn't earn — the #796 fold, reversed.
- **On-time replies record nothing.**

Not modelled on #486, deliberately: that buffer serves a caller still alive polling in a grace window. A mutation caller has been handed an outcome-unknown rejection and moved on. There is no poller.

**2 — wiring.** `withRetryToken` drains on `retry_of` and prepends the outcome. **Not a short-circuit** — skipping the dispatch because a rid landed is #683's inference with the arrow reversed: the bridge stores the rid, not a fingerprint, so "the write for this rid landed" does not prove "the write asked for now is that write". A stale token would swallow a real mutation. #687 reverted that class for cause; dedupe stays the #521 ledger's job.

## Verification

9 tests. The 5 bridge tests were **red before the change**; every guard is mutation-verified to bite:

| mutation | test killed |
|---|---|
| drop the `ctx.mutating` guard | the read test |
| remove the `ok !== true` return | the late-failure test |
| record on the resolve path | the on-time test |
| remove the notice, keep the drain | the report test |
| probe unconditionally | the no-retry-token test |

I checked these because the first three would all have passed against a stub returning `undefined` — same-reason-red is weak red.

Full suite: **391 files, 7929 passed, 2 skipped.** `tsc --noEmit` clean.

## Not covered, stated plainly

A caller told outcome-unknown who **never retries** still learns nothing — the notice sits until TTL. That needs a tab-scoped drain on the next panel call, a wider blast radius than this. #694 should stay open for it, or get a follow-up issue.

Refs #517, #521, #683, #687, #704, #707, #1154.

### DIFF
--- a/src/__tests__/orchestrator/late-mutation-e2e.test.ts
+++ b/src/__tests__/orchestrator/late-mutation-e2e.test.ts
@@ -1,0 +1,314 @@
+// #694 — the late-mutation notice, END TO END over a REAL bridge.
+//
+// WHY THIS FILE EXISTS, when both halves already have coverage.
+//
+// ui-bridge.test.ts drives the real bridge but reads the outcome by calling
+// `bridge.takeLateMutation(rid)` with a rid harvested off the wire.
+// panel-retry-identity.test.ts drives the real tool defs and the real
+// makePanelToolCtx, but against a bridge whose `takeLateMutation` is a stub fed
+// from a scripted map. Both pass if the two halves disagree about WHICH rid.
+//
+// And they could disagree, silently. The caller is handed a token by
+// panel-tools (`retry_of:"<dispatchedRid>"`, from the onDispatchedRid observer);
+// retention is keyed by a rid minted inside UiBridge.dispatch. Nothing in the
+// suite asserts those are the same string. If they ever diverge — a re-dispatch
+// that mints a fresh rid, an observer moved before the mint, a rebind — the
+// feature is inert in production while every existing test still passes,
+// because every existing test supplies the rid itself.
+//
+// So this file hands NOTHING to the drain. It reads the rid out of the retry
+// hint the way an agent would (regex over the error text), and asserts that
+// string is the one the bridge saw on the wire and keyed retention by. That
+// identity is the feature's single point of failure.
+//
+// DELIBERATELY SHORT TIMEOUTS. The production default for a write is 20 s
+// (BRIDGE_DEFAULT_TIMEOUT_MS), and an earlier throwaway version of this probe
+// used it: first attempt times out at 20 s, panel replies at 21.5 s, retry
+// succeeds. It passes, and it costs 25 s of CI for no extra proof — the rid is
+// minted, observed, retained and drained by the same code paths regardless of
+// the number. The deadline is not what is under test, so `fastCtx` below shrinks
+// it and everything else stays real.
+
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { createServer } from "node:net";
+import { readFileSync } from "node:fs";
+import WebSocket from "ws";
+import { UiBridge } from "../../services/ui-bridge.js";
+import {
+  buildPanelToolDefs,
+  makePanelToolCtx,
+  RETRY_TOKEN_CMDS,
+  type PanelToolCtx,
+  type ToolResult,
+} from "../../orchestrator/panel-tools.js";
+
+const textOf = (res: ToolResult): string => (res.content[0] as { text: string }).text;
+
+function freePort(): Promise<number> {
+  return new Promise((resolve, reject) => {
+    const srv = createServer();
+    srv.once("error", reject);
+    srv.listen(0, "127.0.0.1", () => {
+      const addr = srv.address();
+      const p = addr && typeof addr === "object" ? addr.port : 0;
+      srv.close((err) => (err ? reject(err) : resolve(p)));
+    });
+  });
+}
+
+/** Same shape as ui-bridge.test.ts's harness: retry past the close→rebind race
+ *  so a bind flake is never reported as a feature failure. */
+async function startBridgeOnFreePort(): Promise<{ bridge: UiBridge; port: number }> {
+  let lastErr = "no attempt made";
+  for (let attempt = 0; attempt < 6; attempt++) {
+    const p = await freePort();
+    const b = new UiBridge(p);
+    b.start();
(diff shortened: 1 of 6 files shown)


artokun:
### First finding: the mechanism exists, it is just wired to one command

`ui-bridge.ts:1474` documents the `ask_user` exemption as a rid-keyed map (`rid → {ask_id, ts}`) whose whole purpose is to let *a reply that validates after the deadline reach the caller, instead of being discarded as a "late reply for a timed-out command"* (#486). Timestamped so an abandoned mapping can be reaped.

That is the shape #694 wants, already built and shipped. So the remaining work is likelier **generalising an existing retention path to mutating commands** than inventing one — which is a much smaller change than the issue body implies, and changes how I would review it.

Two things to establish before writing any code, and I have not established either yet:

1. **Why the exemption is command-scoped.** `ask_user` has a property mutations may not share: its late reply is a *user answer*, inherently still meaningful, and #486 was about not losing it. A late mutation reply is a *completion notice* for an operation the caller has already been told is outcome-unknown. Whether the retention path can carry that without lying — e.g. if the panel's late reply reports success for a write that a subsequent retry has since re-applied — is the actual design question. The `retry_of` ledger (#521) may already resolve it, in which case the two mechanisms compose; if not, surfacing the late reply could contradict the ledger.
2. **A red test first.** I want a test proving a late mutation reply is discarded *today*, before touching anything. Per this repo's pattern, a fix whose test was written after the fact tends to assert the fix rather than the defect.

Also noting for the record, since it changes what "fixing #694" means: the disclosure text at `ui-bridge.ts:1234-1253` already reasons carefully that a reply-timeout is the *likelier* of the two failure shapes to have applied, and that the bridge deliberately does not auto-retry (#334). Whatever surfaces here has to stay consistent with that text, or one of the two becomes false — which is the failure mode #1204 and #796 are both about.

No implementation this pass. Next: (1) find the drop site and the exact predicate that excludes non-`ask_user` rids, (2) red test, (3) decide question 1 above with the ledger's behaviour measured rather than assumed.

artokun:
## Review verdict: works, one blocking defect — fixed in `bdd90ba`

The reachability question came back **negative, and measured**: the rid the caller is handed at `panel-tools.ts:5071` is the same closure variable the timer keys retention by (`ui-bridge.ts:3841` → `:3890` → `:3947`), mutations never mint a second rid, and an e2e probe on a real socket with a real 20 s timeout round-tripped the hint back into the notice. Not hand-fed. That was the thing that could have sunk it.

### F2 — blocking, and mine twice over

**"READS are excluded" was false as shipped.** `ctx.mutating` is `!BRIDGE_READONLY_CMDS.has(cmd)` — the discriminator `ui-bridge.ts` *itself documents* as misclassifying `graph_find_nodes`, `graph_list_subgraphs`, `graph_screenshot`, `graph_canvas`, `graph_select_nodes`, `graph_enter/exit_subgraph`, `graph_copy_nodes`. That's #778, re-asked inside the file that warns against it. Seven reads retained, under a comment saying reads were excluded.

Worse than the bug: **my test certified almost nothing.** It used `graph_outline`, one of the twelve where the discriminator happens to be right. The mutation-check passed honestly and proved a far narrower property than the prose claimed. This is the fourth inert guard I've shipped this week and the pattern is identical every time — I tested the case I had in mind rather than the case that distinguishes.

Fixed by inverting who answers: the retry-token layer installs `setLateMutationFilter` from `RETRY_TOKEN_CMDS`. **Fail closed** — no filter, no retention. The bridge shouldn't be guessing; the question was never "does this mutate" but "can this rid come back as a retry token", which only that layer knows.

### F3/F4/F5 — fixed together

Drain now runs **after** the handler and is skipped when the retry errored. Draining first meant a failed retry consumed the notice permanently — and gating on a thrown exception wasn't enough, because `ctx.call` converts nearly every dispatch failure into an `isError` result. That was the same defect one layer over, and I only found it because the test I wrote for F3 kept failing.

Draining later also **widens the window** F1a identified: a late reply landing while the retry is in flight is now caught instead of stranded until TTL.

The notice is matched on **command**, not rid alone. My own test had baked in the false version — a `graph_set_node_mode` notice printed for a `panel_add_node` retry, under a sentence claiming "the attempt *you are retrying*" completed.

### F6 — dead code removed. F7/F8 — confirmed clean, no change.

### Claims corrected

Struck from the PR body: "Reads excluded" (false), "every guard is mutation-verified to bite" (each mutation bit, but the read guard certified 1 of 12). Added to **Not covered**: F1a (an agent retrying faster than the panel's late reply still learns nothing — now narrowed by the later drain, not eliminated) and F1b (**the mid-command-disconnect half of outcome-unknown can never produce a notice**, though `panel-tools.ts:5044` mints a token for it — so half the token population is inert here).

4 tests added, each fix mutation-verified: restoring the `ctx.mutating` gate kills **both** new bridge tests; dropping the cmd match kills the wrong-command test; draining before the handler kills the failed-retry test.

Full suite: **391 files, 7933 passed, 2 skipped.**

Still open for a follow-up, not this PR: F1b, and the e2e test the reviewer offered — currently all four wiring tests stub `takeLateMutation`, so **nothing in the suite connects the bridge's retention to the tool's drain**. The rid semantics hold today but are unpinned. That's a real gap and it should be closed by the test, not by my having traced it once.

PR #1216, opened 2026-08-09
fix(test): de-flake two timing-dependent tests

### LINKED ISSUE #1208: Flaky test: "lists newest first" relies on a 2 ms gap between two download jobs

artokun:
Fixes #1208. **Draft — claiming the issue, work in progress.**

## Two flakes, both observed in one session

**1. `lists newest first`** (`download-jobs.test.ts`) — failed on **all three platforms** and blocked the `v0.50.56` release build; an unchanged re-run went green. It starts two jobs 2 ms apart and asserts ordering from `b.started_at - a.started_at`, a **millisecond** timestamp with no tiebreak.

**2. `an UNREADABLE disk version leads with the two free checks, not with an update`** (`ui-bridge.test.ts`) — failed once during a local full-suite run, passed on an immediate unchanged re-run.

## Why these are worth fixing rather than re-running

A flake that fails **all three platforms at once** does not read like a flake — it reads like a real regression, and that is what it costs. Ruling the first one out meant reading the diff to confirm `download-jobs.ts` was untouched and checking the failure history. The second one caused me to commit while the suite was red, then re-verify.

The expense is not the re-run. It is the minutes spent proving a change was innocent, every time, on the branch where you can least afford the doubt.

## Plan

Do not race the clock:

- give `listDownloadJobs()`'s comparator a deterministic tiebreak so equal timestamps still order stably;
- have the test assert on something it controls rather than on a 2 ms wall-clock gap.

Then diagnose the second one properly rather than assuming it is the same shape — I have not traced it yet, and guessing would be the same mistake this issue is about.

### DIFF
--- a/src/__tests__/services/download-jobs.test.ts
+++ b/src/__tests__/services/download-jobs.test.ts
@@ -108,6 +108,7 @@ import {
   startDownloadJob,
   getDownloadJob,
   findDownloadJob,
+  compareTrayIds,
   listDownloadJobs,
   listDownloadJobCandidates,
   cancelDownloadJob,
@@ -2107,5 +2108,41 @@ describe("download job registry", () => {
       expect(r.confirmed).toBe(false);
       expect(r.warning).toMatch(/NOT verified as landed/);
     });
+  });
+});
+
+// #1208 (codex review) — the tiebreak must be DETERMINISTIC ACROSS MACHINES.
+//
+// The first version used `String(a.trayId).localeCompare(String(b.trayId))`.
+// localeCompare is locale-aware by definition and its collation depends on the
+// runtime's ICU build, so `tray-B` vs `tray-a` orders one way here and can order
+// the other way elsewhere:
+//
+//     "tray-B".localeCompare("tray-a")  →  1
+//     "tray-B" < "tray-a"               →  false  (raw: -1 the other direction)
+//
+// That would have traded a timing flake for a portability flake, in the one
+// function whose job is to be identical on every machine.
+describe("compareTrayIds (#1208)", () => {
+  it("orders by RAW string comparison, not locale collation", () => {
+    // The exact pair where the two disagree.
+    expect(compareTrayIds("tray-B", "tray-a")).toBeLessThan(0);
+    expect("tray-B".localeCompare("tray-a")).toBeGreaterThan(0);
+  });
+
+  it("is antisymmetric and reflexive — a sort comparator must be", () => {
+    expect(compareTrayIds("a", "b")).toBeLessThan(0);
+    expect(compareTrayIds("b", "a")).toBeGreaterThan(0);
+    expect(compareTrayIds("a", "a")).toBe(0);
+  });
+
+  it("sorts a MISSING trayId last without colliding on 'undefined'", () => {
+    // String(undefined) === "undefined" would have made every id-less job equal
+    // to every other AND sortable against real ids by that literal.
+    expect(compareTrayIds(undefined, "a")).toBeGreaterThan(0);
+    expect(compareTrayIds("a", undefined)).toBeLessThan(0);
+    expect(compareTrayIds(undefined, undefined)).toBe(0);
+    // …and it must not sort as the literal string.
+    expect(compareTrayIds(undefined, "zzzz")).toBeGreaterThan(0);
   });
 });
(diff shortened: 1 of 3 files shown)


artokun:
**Codex adversarial review — one real defect, and it was mine.** I fixed a timing flake by introducing a portability one.

**`localeCompare` was the wrong tool.** It is locale-aware by definition and its collation depends on the runtime's ICU build, so the same two tray ids can order differently on different machines. Demonstrated concretely:

```
"tray-B".localeCompare("tray-a")  →   1
"tray-B" < "tray-a"               →  -1   (the other direction)
```

In the one function whose job is to be identical everywhere, that swaps a flake that fails on a busy runner for one that fails on a different ICU build — strictly harder to diagnose than what it replaced. Now raw comparison.

**Also from the same review:** `String(undefined)` is `"undefined"`, so a missing `trayId` did not merely fail to break a tie — it sorted *as that literal*, placing id-less jobs among real ids by wherever `"undefined"` happens to fall. A missing id now sorts last and never displaces a present one.

The helper is exported and unit-tested, which the timestamp-collision case could not be: it is pure, so antisymmetry, the locale-divergent pair, and every `undefined` combination are directly checkable. Mutation-verified — restoring `localeCompare`, collapsing `undefined` to the literal, and breaking antisymmetry each fail the suite.

## On the review's other points

**Q1 (mutating `a.job.started_at` may not reach `listDownloadJobs`)** — a fair flag, and I checked rather than assumed: inverting the comparator *does* fail the test, so it is exercising the real ordering path for these in-memory jobs. It is fragile in the sense the review means — it depends on object identity — so that is worth knowing if the persisted path ever feeds this test.

**Q3 (jobs sharing both `started_at` and `trayId`)** — real, and left alone deliberately. Two entries sharing both are the same physical download by this file's own identity rule `(id, trayId)`, so they are deduped upstream; adding a third key would be defending against a state the registry does not produce.

**Q4** — matches what I traced independently before the review: the wording varies with `installPanelUsable`, not with shared module state. Reproduced at ~1 run in 6, now 0 in 22.

## What is deliberately NOT here

A test asserting the tiebreak through `listDownloadJobs()`. I wrote one, and mutation testing showed it vacuous — it mutated `started_at` on the array that function *returns*, which is fresh objects, so it never touched the registry and failed with **and** without the tiebreak. Forcing two jobs into one millisecond is not reachable through the public API, so the integration-level property stays uncovered and says so in a comment, rather than shipping a test that passes for the wrong reason.

Full suite: 392 files, 7929 passed. Lint, #796 gate and vocabulary gate clean.

Issue #1136, opened 2026-08-08
An unreachable host reads as an empty result — blocked/filtered users are told 'nothing found' instead of 'could not reach X'


artokun:
An unreachable external host currently presents to the user as an **empty result**, not as an error. Empty reads as "nothing matched", so the user concludes their query was wrong, or that the thing they are looking for does not exist — and goes on trying variations of an action that can never succeed.

## Where this came from

A Chinese-speaking user (`yuanbrand`, Discord `#adult-mode-ama`) was told to update the panel through ComfyUI Manager. Their reply:

> 我这边搜不到任何内容 — "I can't find anything when I search"

I read that as a stale Manager cache and sent them two more routes. In hindsight a blocked registry fits the evidence better, and CivitAI is very likely blocked for them too. **Three consecutive pieces of my advice were useless because I was sending them at a door that could not open, and neither of us could see that from the symptom.**

## Hosts we depend on

- **CivitAI** (`civitai.com`, `search-new.civitai.com`) — model browser, search, discovery
- **ComfyUI Manager's catalogue** — the path we tell everyone to use to install and update the panel
- **npm registry** — `npx comfyui-mcp@latest` is how the orchestrator starts at all
- **GitHub releases** — the `--front-end-version` path

## What good looks like — we already do it in one place

`panel_civitai_results` already distinguishes an upstream failure from a genuine empty set, and the tool description says so explicitly: a `total:0` result is **not** automatically "no matches", because an `error` field (e.g. a 503) may be present, and callers are told to check it before concluding anything from an empty grid.

That is exactly the right pattern. It has simply never been applied to the install/update paths or to the panel's own network-dependent surfaces.

## Ask

Not CDN mirrors, not a China-specific mode — just make unreachable say *unreachable*:

1. When a fetch to an external host fails (DNS, timeout, connection refused, TLS), surface **that**, distinctly from an empty result set.
2. Name the host that could not be reached, so a user behind a filter can recognise their own situation immediately.
3. Anywhere we tell a user to "search for X in Manager", the failure to reach the catalogue should be legible rather than presenting as "X does not exist".

## Family

Same class as the rest of this week: [`artokun/comfyui-mcp#948`](https://github.com/artokun/comfyui-mcp/issues/948) (a remedy naming a command that is not there), `artokun/comfyui-mcp-panel#779` (a panel that renders nothing and says nothing). In each case the software knew more than it told the user, and the user paid for the gap in wasted hours.

[...]

PR #1217, opened 2026-08-09
fix: stop mislabelling reachability — npm and workflow-url (#1136)

artokun:
Addresses #1136 for **every site we own and can fix**. That turned out to be two hosts, not four — see the inventory.

## What #1136 asked for

An unreachable host presents as an **empty result**. Empty reads as "nothing matched", so a filtered user concludes the thing does not exist and keeps trying variations of an action that cannot succeed. The reported harm: a Chinese-speaking user was told to search ComfyUI Manager, the registry was unreachable, the grid was empty, and three rounds of advice went to a door that could not open.

## Inventory — six paths read individually

| host | verdict |
|---|---|
| ComfyUI host (`comfyui/fetch.ts`) | **already correct** (#954) — the reference the rest were measured against |
| CivitAI (`civitai-resolver.ts`) | **already correct** — `civitaiFetch`/`transportError` name the host and separate timeout / network / status / genuine-empty |
| `civitai-lookup.ts` | **dead code** — no importers anywhere in `src/`. Separate issue: delete or revive |
| npm (`self-update.ts`) | **fixed here** |
| `workflow-url.ts` | **fixed here** |
| ComfyUI-Manager | **upstream** — `manager_server.py:947` returns `dict(channel, node_packs)`; an unreachable registry and an empty one are byte-identical to us. No signal exists to read |

## The two real defects, which point in opposite directions

**npm was over-claiming.** `defaultGetLatestVersion` returned `string | undefined` with four conditions collapsing into that `undefined` — DNS/TLS failure, non-2xx, missing `version` field, unparseable body — and the consumers then asserted the most specific: *"npm registry unreachable (offline or timed out)."* A 429 sent the user off to debug their network. Not #796's could-not-determine folded into determined-not; **determined-not asserted from could-not-determine**.

Now three-state: `unreachable` only when the transport proves it (naming the host), `undetermined` with the specific reason when the registry answered, `version` otherwise. Both consumers fixed — the second was still calling `getLatestPublishedVersion`, whose job is to flatten the three states back into `string | undefined`, so the discrimination existed and was discarded one layer down.

**workflow-url was under-claiming.** Its timeout branch already said `Timed out after Nms fetching workflow from "<host>"`. The transport fallthrough said `Failed to fetch workflow URL: fetch failed` — no host, no sign the problem is reachability. A filtered user reads that as "my link is wrong". The two branches now agree.

## The classifier

`describeUnreachableHost` in `comfyui/fetch.ts`, a deliberate **subset** of the existing `NEVER_DELIVERED_CODES`. That set answers *"could this have been delivered?"*, for which a malformed URL and a filtered DNS lookup are equivalent — neither arrived. This one answers *"is this host unreachable?"*, for which they are opposites. `ERR_INVALID_URL` and `ERR_UNSUPPORTED_PROTOCOL` are excluded; reusing the set wholesale would tell a user with a typo to inspect their firewall.

## Verification

12 tests. The **negatives** are the discriminating ones throughout — a suite checking only positives would pass just as happily against the wrong set:

| mutation | test killed |
|---|---|
| add `ERR_INVALID_URL` to the unreachable set | malformed-URL test |
| remove the `workflow-url` throw | DNS test (ECONNRESET test unaffected — correctly) |
| — | a registry that ANSWERED must never produce "unreachable" |

Suite: **395 files, 7959 passed, 2 skipped.** `tsc --noEmit` clean.

## Three corrections I made to my own analysis, recorded in the comments

I asserted things three times about code I had read partially: a producer without its consumer (npm's real defect is the reverse of what I first described), a call without its wrapper (CivitAI was already correct), and a module without its importers (`civitai-lookup` is dead). All three were caught by re-reading, none by a test — there is no test that fails when an inventory is wrong.

Follow-ups, deliberately not folded in: the dead `civitai-lookup.ts`, and an upstream issue asking ComfyUI-Manager's `getlist` to report refresh failure or cache age.

Refs #796, #948, #954.


artokun:
### First inventory pass — the shape is not what I assumed

I went looking for the obvious form of this bug: `catch { return [] }`. Across `src/services` and `src/comfyui` there is essentially **one** (`comfyui/fetch.ts:72`). So the defect is almost certainly *not* "a catch swallows the error."

That matters, because it means the fix is not a sweep of catch blocks. The likelier shape — and the one this codebase has produced repeatedly — is that the error **is** captured and then **discarded downstream**: carried on a result object that a renderer or a caller never reads, exactly like the record-vs-renderer split in #1197 and the description-vs-record split in #1148. `panel_civitai_results` supports that reading: it already *has* the `error` field, which means the vocabulary exists and the question is who fails to consult it.

So the next pass is a consumer audit, not a producer audit: for each of the four hosts, find where a populated `error` stops being propagated. That is a different search and I would rather redo it than pattern-match the wrong layer — the four inert guards I shipped earlier this week were all "fixed the layer I happened to be looking at."

Not claiming an inventory yet. What I have is one negative result that redirects the search, which is worth recording precisely because it would otherwise be redone.

artokun:
### Second pass — found it, and the fix is narrower than the issue implies

The transport-failure vocabulary this issue asks for **already exists and already names its host**: `comfyui/fetch.ts` classifies `ECONNREFUSED` / `ENOTFOUND` / `EAI_AGAIN` / `ETIMEDOUT` (`:128-130`) and was built for #954, when a bare `TypeError: fetch failed` was reaching tool results verbatim (`:97`). That is precisely the "say unreachable, name the host" behaviour #1136 wants.

It is bound to **one** host. `comfyuiFetch` (`fetch.ts:264`) is the ComfyUI server path. The external hosts in scope do not use it:

- `civitai-lookup.ts:44` — bare `fetch(url, {headers, signal: AbortSignal.timeout(10000)})`
- `civitai-resolver.ts:170` — bare `fetch(...)`

So a DNS failure or a filtered connection to `civitai.com` produces a raw rejection with no host attribution, and whatever sits above it renders as empty. Same expected story for the Manager catalogue, npm and GitHub-releases paths — still to confirm individually rather than assumed, since assuming is how this class survives.

**This reframes the work.** It is not "add error handling to N call sites" — it is *generalise the existing classifier off the ComfyUI host and route external fetches through it*, then make the consumers distinguish its output from an empty set. That reuses the vocabulary rather than inventing a second one, which was the plan's stated constraint, and it means the behaviour is already proven in production for one host.

One caution I want on the record before writing anything: `comfyuiFetch` carries ComfyUI-specific behaviour (`describeTargetDrift`, connected-panel origins, `comfyHttpTimeoutSeconds`). Pointing CivitAI at it wholesale would drag that along and produce confidently wrong diagnostics about panel origins for a host that has none. The extraction has to take the classifier, not the wrapper.

Next: confirm the Manager/npm/GitHub paths individually, then a red test per host that an unreachable host yields a message naming it.

artokun:
### Third host confirmed — and it is the cleanest example in the repo

`self-update.ts:133-144`, `defaultGetLatestVersion`:

```ts
if (!res.ok) return undefined;
const json = (await res.json()) as { version?: unknown };
return typeof json.version === "string" ? json.version : undefined;
} catch {
  return undefined;
}
```

**Four distinct conditions collapse into one `undefined`:**

1. the npm registry could not be reached (DNS / filtered / timeout),
2. the registry answered with a non-2xx,
3. the response had no `version` field,
4. …and whatever the caller reads `undefined` as, which for an update check is almost certainly "nothing newer".

That is #796 in its textbook form, and it is the *same* failure the reporter's user hit — one door down. Someone behind a filter is silently told they are up to date, by a function that never found out. The `catch {}` with no binding makes it literally impossible to distinguish, so this one is not a propagation problem like CivitAI; the information is destroyed at the source.

Confirmed inventory so far — **3 of 4** hosts, each verified by reading the site, not inferred from the others:

| host | site | shape |
|---|---|---|
| CivitAI | `civitai-lookup.ts:44`, `civitai-resolver.ts:170` | bare `fetch`, no host attribution |
| npm registry | `self-update.ts:135` | 4 conditions → one `undefined` |
| GitHub / raw | `workflow-url.ts:251`, `skill-generator.ts`, `code-provider-auth.ts` | still to read |
| Manager catalogue | not yet located | — |

Before fixing #2 I need the caller: `undefined` may already be handled as "could not check" somewhere upstream, in which case the defect is narrower than it looks. Checking that rather than assuming it — the last two issues I worked both turned on a claim that was true at one layer and false at the next.

artokun:
### Fourth host: `workflow-url.ts` mostly already does the right thing

`workflow-url.ts:249-266` is the closest thing to a counter-example in the inventory, and it changes the shape of the fix:

```ts
if (e?.name === "AbortError") {
  throw new ValidationError(`Timed out after ${timeoutMs}ms fetching workflow from "${u.hostname}".`);
}
throw new ValidationError(`Failed to fetch workflow URL: ${e?.message ?? err}`);
```

It **throws** rather than returning empty, and the timeout branch **names the host**. That is #1136's ask, satisfied, for one of two branches.

The gap is the fallthrough: a DNS failure or a refused connection produces `Failed to fetch workflow URL: fetch failed` — no host, no indication it is a reachability problem rather than a bad URL. For a filtered user that reads as "my link is wrong", which is the same wrong conclusion as "it does not exist", just phrased differently.

**Inventory now 4 of 4 located**, each read rather than inferred:

| host | site | verdict |
|---|---|---|
| npm registry | `self-update.ts:135` | **worst** — 4 conditions → one `undefined` |
| CivitAI | `civitai-lookup.ts:44`, `civitai-resolver.ts:170` | bare `fetch`, no attribution |
| GitHub / URL fetch | `workflow-url.ts:251` | throws + names host on timeout; **transport branch loses both** |
| ComfyUI host | `comfyui/fetch.ts:128-130` | **already correct** — the reference impl (#954) |

Manager catalogue is the remaining unknown: it is reached through ComfyUI's own Manager endpoints rather than a direct outbound fetch from us, so the failure may surface via the ComfyUI client path instead. That needs checking before I claim anything about it — it is the host the issue's user actually hit, so getting it wrong would miss the point of the report entirely.

This also settles the design question the plan left open. There is no single choke point to fix: one path is already correct, one is half correct, two are wrong in different ways. The work is to extract the `fetch.ts` classifier so it can name **any** host, then apply it at four sites — not to add a wrapper and route everything through it.

artokun:
### Manager catalogue: the classifier CANNOT fix this host, and it is the one the user hit

`workflow-deps.ts:189` reaches the catalogue via `managerFetch("/customnode/getlist?mode=cache&skip_update=true")`. ComfyUI-Manager is a **ComfyUI extension**: that request goes to the local ComfyUI server, and *ComfyUI* then talks to the registry.

So when a user is behind a filter, the DNS/TLS failure happens **on the ComfyUI host, in a process we do not own**. We receive a perfectly healthy HTTP 200 carrying an empty or stale list. There is no `ENOTFOUND` for us to classify, no host for us to name, and no transport error anywhere in our stack.

**This is why the reporter's user could not be helped, and why I could not see it either.** The advice was "search in Manager"; Manager returned an empty grid; every layer including ours was working correctly and reporting truthfully on what it could see. The information was destroyed one process away.

Consequences for this PR, stated before I write code rather than after:

1. **Extracting the `fetch.ts` classifier does not fix the Manager path.** It fixes npm, CivitAI, and the `workflow-url` fallthrough — three of four. Claiming #1136 closed on the classifier alone would leave the reporter's own scenario untouched while the PR title says otherwise.
2. That host needs a **different** signal: Manager's own response carries `channel`, and `mode=cache` means we may be reading a cache that was never refreshed. Whether an empty `node_packs` can be distinguished from a failed upstream refresh is an open question I have not answered — and answering it may require a Manager-side change we do not control.
3. Separately, `:188` returns `{packs: [], directInstall: true}` for a **v4 API**, a legitimate non-failure empty. Any caller reading `packs: []` as "no packs exist" is already wrong today, independent of reachability.

Revised plan: ship the classifier for the three hosts where we own the fetch, and treat the Manager catalogue as its own investigation with its own issue, rather than quietly letting it ride on a PR that cannot deliver it. I would rather close 3 of 4 honestly than 4 of 4 on paper.

I will confirm point 2 before proposing anything — the whole issue is about a system that knew more than it told the user, and guessing here would reproduce it.

artokun:
### Point 2, answered — and the answer needs a caveat I am not going to skip

`ManagerListResponse` (`workflow-deps.ts:55-58`) is:

```ts
type ManagerListResponse = {
  channel?: string;
  node_packs?: Record<string, ManagerNodePack>;
};
```

No error field, no staleness field, no refresh timestamp. So **on the shape we model, an empty `node_packs` is indistinguishable from a failed upstream refresh.**

The caveat: that type is *our declaration of their response*, not their response. Manager may well send fields we discard at the cast on `:190`. What I have verified is "our model carries no signal" — which is **not** the same claim as "Manager sends no signal", and the difference decides whether this is fixable on our side alone.

Saying it the loose way would be the exact move this issue is about: reporting a confident conclusion about something one layer out that I did not actually check. Three of my last four defects were that pattern.

Resolving it properly means reading ComfyUI-Manager's `/customnode/getlist` handler — their source, not ours — to see what it emits when its registry refresh fails. That is a real answer available for the cost of reading one upstream file, and it decides the shape of the Manager half entirely:

- **if Manager does signal it** → we are discarding it at `:190` and the fix is ours, small, and belongs in this family
- **if Manager does not** → the fix is upstream, and the honest thing here is a separate issue plus possibly a report to them

Until that is read, the Manager half stays out of this PR's scope, and the PR will close 3 of 4 hosts with the fourth named explicitly as unfixed and why. Overstating coverage on an issue about a system that overstated what it knew would be a poor way to close it.

artokun:
### Point 2 settled — read from ComfyUI-Manager's source, not from our types

`manager_server.py:947` `fetch_customnode_list`, last line:

```python
result = dict(channel=channel, node_packs=node_packs.to_dict())
```

**Manager genuinely sends no failure signal.** Two fields, exactly the two we model — so my earlier caveat resolves the unlucky way: `ManagerListResponse` is not lossy, there is simply nothing to lose. A registry refresh that failed and a registry that returned nothing produce byte-identical responses.

Note `core.get_unified_total_nodes(channel, mode, 'cache')` — the call is hardcoded to `'cache'`, and our request already passes `skip_update=true`, so no upstream fetch is even attempted on this path. A user behind a filter gets whatever the on-disk cache holds, which for a fresh install is nothing, reported as a successful empty list.

**This closes the Manager question for us: it is upstream, not ours.** Nothing we can add on our side distinguishes the two cases, because the distinction never crosses the process boundary. Options, none of them this PR:

1. an issue against ComfyUI-Manager asking `getlist` to report refresh failure or cache age,
2. a heuristic on our side — an empty `node_packs` with a non-`local` channel is *suspicious* — which is a guess dressed as a diagnosis and would tell some users "you may be blocked" when they simply have no packs. Given the issue is about a system asserting more than it knew, shipping that would be self-defeating.

So this PR closes **3 of 4** hosts (npm, CivitAI, the `workflow-url` transport fallthrough) by extracting the `comfyui/fetch.ts` classifier so it can name any host. The Manager path gets a filed upstream issue and a note in ours.

I am glad I read their file. From our side the evidence pointed at "we are discarding a signal", which would have produced a confident fix that changed nothing — the same shape as the bug, and the third time this week that a claim which was true at one layer was false at the next.

artokun:
### The extraction target, and a trap in it

`comfyui/fetch.ts:127-139` already holds the right set — `NEVER_DELIVERED_CODES`, the codes that *positively prove* a request never reached the host. That is exactly #1136's "unreachable", so the classifier does not need writing, only exposing.

**But it must not be reused wholesale.** Four of its ten members are not reachability failures:

```
ECONNREFUSED  ENOTFOUND  EAI_AGAIN                      <- unreachable (host's fault / network)
UNABLE_TO_VERIFY_LEAF_SIGNATURE  DEPTH_ZERO_SELF_SIGNED_CERT
SELF_SIGNED_CERT_IN_CHAIN  CERT_HAS_EXPIRED
ERR_TLS_CERT_ALTNAME_INVALID                            <- reached it, TLS refused (arguably unreachable)
ERR_INVALID_URL  ERR_UNSUPPORTED_PROTOCOL               <- CALLER error, nothing to do with the host
```

The set exists to answer a *different question* — "could this request have been delivered and acted on?" — for which a malformed URL and a filtered DNS lookup are equivalent (neither was delivered). For #1136's question — "is this host unreachable from here?" — they are opposite: telling a user with a typo'd URL "could not reach civitai.com" sends them to check their firewall over a missing character.

That is the same shape as everything else in this issue: a predicate that is correct for the question it was written for, reused for a question it does not answer. Given #1136 exists *because* a system reported confidently on something it had not checked, importing this set unexamined would be a poor way to close it.

**Concrete plan, ready to implement:**

1. In `comfyui/fetch.ts`, export `describeUnreachableHost(err: unknown, url: string): string | null` — returns a message naming the host for the DNS/refused/TLS codes only, `null` otherwise. The URL-shape codes stay out, and stay in `NEVER_DELIVERED_CODES` where they are correct.
2. Apply at the three sites we own: `self-update.ts:135` (return a discriminated result rather than `undefined` — 4 conditions currently collapse into it), `civitai-lookup.ts:44` / `civitai-resolver.ts:170`, and `workflow-url.ts:265`'s non-timeout branch (its timeout branch already names the host and is the in-repo model for the message shape).
3. Red test per site first: an unreachable host yields a message naming it; a genuinely empty result still reads as empty.

Manager stays out, per the upstream finding above.

artokun:
### Correction: I mis-described the npm path, and the real defect points the other way

Earlier I called `self-update.ts` the **worst** site — "4 conditions collapse into one `undefined`", which then "reads as *nothing newer*". The first half is true. The second is **wrong**, and I should have read the consumer before saying it.

`self-update.ts:733-739`:

```ts
const latest = await deps.getLatestVersion();
if (!latest) {
  return { action: "unavailable", …, note: "npm registry unreachable (offline or timed out)." };
}
```

The caller already maps a missing version to **unavailable**, and already says *unreachable*. So npm does not silently report "up to date" — that part of my description was invented from the producer's shape without checking who consumed it.

The actual defect is the **same fold pointing the opposite way**. Four conditions still collapse into one `undefined`, but the consumer then asserts the *most specific* of them:

| what happened | what the user is told |
|---|---|
| DNS/TLS failure, host truly unreachable | "npm registry unreachable" ✓ |
| registry answered non-2xx (500, 429) | "npm registry unreachable" ✗ |
| response had no `version` field | "npm registry unreachable" ✗ |
| JSON parse failure | "npm registry unreachable" ✗ |

So this is not #796's *could-not-determine folded into determined-not*; it is **determined-not asserted from could-not-determine**. Same collapse, opposite overclaim — and it sends a user with a rate-limited registry to debug their network. It also never names the host, which is #1136's second ask.

That changes the fix here: not "start reporting unreachable" (it already does), but *stop reporting unreachable when we do not know that*, and name the host when we do. `describeUnreachableHost` gives exactly that discrimination — it returns `null` for the three non-reachability cases.

Recording the error rather than quietly editing the earlier comment: I asserted a downstream behaviour from an upstream shape, on an issue about software asserting more than it knows. Third time this week — and the reason my own rule is *read the consumer, not the producer*.

artokun:
### CivitAI site analysed — needs a return-type change, so it is NOT landing in this pass

`civitai-lookup.ts:42-76` returns `null` for **three** distinct outcomes:

| case | line | what `null` means to the caller |
|---|---|---|
| HTTP 404 — genuinely not on CivitAI | `:46` | correct, and correctly cached via `updateCivitaiInfo(…, null, null, null)` |
| non-2xx (500, 429, auth) | `:53` | wrong — reads as "not on CivitAI" |
| transport failure (DNS/TLS/refused) | `:71` | wrong — the #1136 harm exactly |

The 404 branch is right and is the one that should keep meaning "absent". The other two are the fold. Worth noting the code is already careful in one respect: only the 404 path writes to the hasher cache, so a filtered user does not get "not on CivitAI" *persisted* — the wrong answer is transient rather than sticky. That is a smaller blast radius than I expected.

**Why I am stopping short of fixing it here.** `null` is load-bearing in the signature (`Promise<CivitaiModelVersion | null>`), and `batchLookup` at `:82` propagates it into a `Map<string, CivitaiModelVersion | null>` consumed further out. Making the distinction visible means changing that type and following it through every consumer — a bigger change than the two already on this branch, and not one to start with the analysis this thin.

Landing what is proven and leaving this explicitly undone is the right split. The commits so far are self-contained: the classifier plus the npm site, both fully tested.

**Branch state:**

- `describeUnreachableHost` extracted, 7 tests, mutation-verified on the subset boundary
- `self-update.ts` fixed — both consumers, 3 tests
- **still to do:** CivitAI ×2 (needs the type change above), `workflow-url.ts`'s non-timeout branch (small — its timeout branch is already the model)
- **out of scope, upstream:** ComfyUI-Manager, per `manager_server.py:947`

Suite on the branch: 394 files, 7957 passed, 2 skipped.

artokun:
### Correction: `civitai-lookup.ts` is DEAD CODE — I nearly spent a type change on it

I listed `civitai-lookup.ts:44` as an in-scope site three times in this PR and planned a cascading return-type change through `batchLookup`'s consumers. It has **no consumers**:

```
$ grep -rn "civitai-lookup" src/
(nothing)
$ grep -rn "lookupByHash" src/
src/services/civitai-lookup.ts:20:  export async function lookupByHash(
src/services/civitai-lookup.ts:105:   const result = await lookupByHash(...)   # itself
```

Nothing in `src/` imports the module — not production, not tests. The only reference is `batchLookup` calling `lookupByHash` inside the same file. Fixing its `null`-means-three-things problem would have been fixing code that never runs, and the "blast radius through every consumer" I used to justify deferring it was a blast radius of zero.

**The live CivitAI path is `civitai-resolver.ts`**, imported by `tools/missing-models.ts` and `tools/model-extras.ts` (and covered by three test files). That is where the remaining work actually is, and it is a different file with a different shape than the one I analysed.

How I got here is worth naming, because it is this issue's own defect class once more: I read a file, found a real bug in it, and never asked whether anything calls it. Same omission as reading a producer without its consumer — one question earlier in the chain.

**Revised remaining work:**
- `civitai-resolver.ts:170` — the live site, still to analyse (read its consumers in `missing-models.ts` / `model-extras.ts` FIRST this time)
- `civitai-lookup.ts` — out of scope here; it is dead and should be deleted or revived under its own issue, not quietly patched to look maintained

Landed and tested so far: the classifier (7 tests), `self-update.ts` (3 tests, both consumers), `workflow-url.ts` (2 tests). Suite 395 files / 7959 passed.

artokun:
### Third correction, and it closes the issue's scope: **CivitAI was already fixed**

I have said four times in this PR that CivitAI is a bare `fetch` with no host attribution. That came from reading `civitai-resolver.ts:170` — the `fetch(` call — **without reading the wrapper it sits inside**.

`civitaiFetch` (`:165`) routes every failure through `transportError` (`:139`), which returns a distinct `ModelError` naming CivitAI, tagged `{timeout:true}` or `{network:true}`. Its own docblock says why it exists:

> …into a DISTINCT, actionable ModelError. A bare `fetch` rejects with an opaque `TypeError: fetch failed` … indistinguishable from any other crash — exactly the "silently swallowed" failure WS-6 set out to kill.

And the next helper turns a non-OK status into its own error so it is never conflated *"with a genuine empty result."*

That is #1136's ask, implemented, in the live CivitAI path, before this issue was filed.

**So the real inventory — after three corrections, all mine:**

| host | verdict |
|---|---|
| CivitAI (`civitai-resolver.ts`) | **already correct** — names the host, distinguishes timeout / network / status / empty |
| `civitai-lookup.ts` | **dead code**, no importers; separate issue |
| npm (`self-update.ts`) | was **over**claiming unreachable from 4 conditions — **fixed here** |
| workflow-url | timeout branch correct, transport branch silent — **fixed here** |
| ComfyUI-Manager | **upstream** — `getlist` returns `dict(channel, node_packs)`, no failure signal exists to read |
| ComfyUI host (`comfyui/fetch.ts`) | already correct (#954) — the pattern everything else was measured against |

**This PR now covers every site we own and can fix.** Two real defects, both fixed and tested; two already-correct implementations I mistook for gaps; one dead file; one upstream.

The honest summary of my own work here: the issue described a four-host problem, and the *actual* problem was two hosts. I found that by repeatedly asserting things about code I had read partially — the producer without the consumer, the call without its wrapper, the module without its importers. Each was caught, but by me re-reading rather than by any test, because there is no test that fails when your inventory is wrong.

Marking ready for review.

artokun:
## Review verdict: the classifier should not exist. Converting this PR.

Nine false claims, three of them structural. Taking the findings in order of what they change.

### The one that invalidates the approach

**`unreachableHostMessage` already exists** — `src/utils/errors.ts:178`, shipped in `f4e9838` *"say 'could not reach <host>', not 'nothing found' (#1139)"*. It answers the same question, names the host, says "This is NOT an empty result", and handles `UND_ERR_CONNECT_TIMEOUT`, `UND_ERR_HEADERS_TIMEOUT` and `ETIMEDOUT` — **the three codes mine misses**. It is already wired at `model-resolver.ts:234` (HuggingFace) and `registry-client.ts:63`, and `registry-client.ts:57` carries an explicit `#1136:` comment.

So my "six paths read individually" inventory missed two sites already fixed *for this issue* plus the helper they use, and `f37856d`'s "exposes rather than invents" is exactly backwards: I invented a second, weaker classifier for a question already answered on main. My import also drags `config.js` and its top-level network scan into two leaf modules; `utils/errors.ts` is a leaf and costs nothing.

**`describeUnreachableHost` comes out. Both sites call `unreachableHostMessage`.**

### Blocking defects in what remains

- **P0-1 — "both consumers fixed" is false; there are three.** `selfUpdateStatus:872` still calls `getLatestPublishedVersion`, which flattens the three states back to `string | undefined` at `:576`. Measured: a 429 still yields *"npm registry unreachable"*, and a genuinely unreachable host has my composed message **discarded, losing the host name**. Both are the same tool (`tools/self-update.ts:20`), so `action:"status"` and the update path now contradict each other — and `status` is the read-only one an agent reaches for while diagnosing exactly the #1136 scenario. I diagnosed this exact fold in the commit message and then failed to apply it to the third call.
- **P2-6 — I reintroduced an "up to date" overclaim.** `{"version":""}` now returns `action:"up-to-date", to:""`. On main `if (!latest)` caught the empty string; `"version" in probe` does not. A non-answer read as a reassuring answer — the thing my own test asserts against — reintroduced through my own discriminator. Needs `json.version.trim() !== ""`.
- **P0-2 — at workflow-url the change fires for TLS only.** `assertHostResolvesSafe:203` pre-resolves DNS, so `ENOTFOUND` can never reach my catch, and the pre-existing message already names the host. My commit's stated motivation is false for DNS. Worse, a dropped connection (how filters usually present) hits undici's 10s connect timeout before `DEFAULT_TIMEOUT_MS = 15_000`, yielding `UND_ERR_CONNECT_TIMEOUT` — not in my set. My test stubs `fetch` to throw `ENOTFOUND` while letting the real DNS pre-check pass: **a state production cannot produce.** The mutation kill was genuine against a dead path.
- **P2-5 — the TLS inclusion is the same overclaim I removed from npm.** A cert error proves TCP connected and a certificate arrived. I exclude `ECONNRESET` reasoning "the host WAS reached" — that applies verbatim to TLS. `unreachableHostMessage` handles this better.

### And the scope verdict was wrong in the user's favour

**P1-4:** I filed the Manager half as upstream-only. But `workflow-deps.ts:467` has no `packs.length === 0` guard, so an empty legacy catalogue sends every pack to `unresolved`, rendered as *"neither installed nor known to ComfyUI-Manager"* and *"Not found in ComfyUI-Manager — install manually"* (`skills-access.ts:737`, `:788`). An empty legacy catalogue is a strong local signal — a healthy one carries thousands of entries. **That is #1136's harm verbatim, in the surface the reported user was sent to three times**, and it is fixable locally. It is the highest-value fix in the repo for this issue and I classified it as someone else's problem.

### What I got right, per the reviewer

The fixture shape is real (verified against Node 24 with live unreachable hosts), the npm three-state discrimination is correct at the two sites it reaches, the `ECONNRESET` negative genuinely discriminates, `civitai-lookup` is dead (no dynamic import, no barrel, no re-export), and the suite counts are exact.

### Plan

Rewrite rather than patch: drop the duplicate classifier, route both sites through `unreachableHostMessage`, fix the third consumer, add the trim guard, correct every claim above, and pull P1-4 into scope since it is the reported harm. Back to draft.

artokun:
### Read `unreachableHostMessage`'s signature — it resolves the TLS debate by not having it

```ts
export function unreachableHostMessage(
  err: unknown, url: string, what: string, opts?: { remedy?: string },
): { message: string; code?: string }
```

It **always returns a message** — there is no `null` branch. That is not an oversight; it is a different and better model of the problem than mine.

Mine asks *"do I have positive proof the host was unreachable?"* and returns `null` otherwise, which forced me into adjudicating `ECONNRESET` vs `CERT_HAS_EXPIRED` vs `UND_ERR_CONNECT_TIMEOUT` one code at a time — and the review was right that I got the line in the wrong place. The existing helper asks the question that actually matters: **`fetch()` rejected, so no response was received, so nothing follows about whether the thing exists.** Whether the TCP handshake completed is invisible to the user and irrelevant to the harm.

It also phrases it better than I did — it splits *"did not respond in time"* from *"could not reach"* using the timeout codes, and always says *"This is NOT an empty result — nothing was queried, so nothing follows about whether what you asked for exists."*

**So the swap is not a drop-in, and it is better than a drop-in.** It is called unconditionally inside the `catch`, with no code-set guard at all:

- `UNREACHABLE_HOST_CODES` and `describeUnreachableHost` are deleted outright, not ported.
- `unreachable-host.test.ts` goes with them — five of its seven tests assert the `null` branch, which is the design being abandoned. The remaining behaviour is already covered where the helper lives.
- The `VersionProbe` three-state stays: the `unreachable` vs `undetermined` split is about *whether we got a response at all*, which is exactly the boundary the helper sits on. `catch` → `unreachable`, non-2xx/unparseable → `undetermined`.

That also disposes of P2-5 without my having to win the TLS argument, and picks up the three timeout codes for free.

Recording this before implementing, because it is the piece that makes the rewrite small: I had assumed the swap meant porting my guard onto their helper, and it means deleting the guard.

Issue #851, opened 2026-08-05
download_model refuses when ComfyUI is Docker/data-dir layout (COMFYUI_PATH has models/ but no main.py)


artokun:
## What happened
download_model fails with VALIDATION_ERROR: models directory could not be determined, even though COMFYUI_PATH points at a working data directory that ComfyUI actually uses for models.

## Environment
- OS: linux
- ComfyUI 0.30.0 running in Docker (python3 main.py --listen 0.0.0.0 --port 8188), argv shows relative main.py
- Host data dir: /home/gradispo/comfyui-data (models/, input/, output/, custom_nodes/ bind-mounted into container)
- COMFYUI_PATH=/home/gradispo/comfyui-data (env set)
- COMFYUI_URL=http://127.0.0.1:8188
- comfyui-mcp 0.49.6 · panel 0.11.38
- GPU: NVIDIA GB10 122GB

## Error (exact)
```
VALIDATION_ERROR: The models directory of the CONNECTED ComfyUI could not be determined...
- running server reported launch script as RELATIVE path "main.py" and did NOT report a working directory
- configured local base "/home/gradispo/comfyui-data" does not contain main.py
- refusing to write to a guessed directory
```

## Evidence path is correct
object_info/UNETLoader lists the exact same files as /home/gradispo/comfyui-data/models/diffusion_models/. README documents host models/ → container /workspace/ComfyUI/models.

## Expected
download_model should accept COMFYUI_PATH when it contains models/ (data-dir / base-directory layout), or honor models path when it matches live object_info listings, without requiring main.py next to COMFYUI_PATH.

## Workaround used
Shell wget into /home/gradispo/comfyui-data/models/{diffusion_models,text_encoders,vae}/ verified against object_info.

## Repro
1. Run ComfyUI in Docker with host data dir bind-mount (no main.py on host path)
2. Set COMFYUI_PATH to the host data dir, COMFYUI_URL to the container
3. call download_model with any HF URL + target_subfolder diffusion_models
4. Observe VALIDATION_ERROR

Related (same root cause): #813, #844

artokun:
Routing to the **#835** cluster (path + environment probing), which is in flight — and flagging that this is the **third instance of one family tonight**.

[...]

PR #1230, opened 2026-08-09
fix(restart): the confirmation timeout must not recommend a server it never checked (#851)

artokun:
Orchestrator half of artokun/comfyui-mcp-panel#851. The panel half shipped in panel **0.11.54** (#858); this is the part that lives here, which the panel-side work explicitly could not reach.

## The harm

`panel_restart_comfyui` timed out waiting for the confirmation card and returned:

> …so I did NOT restart ComfyUI. … or use `restart_comfyui` to restart the server directly without a panel card.

The reporter followed that advice. `restart_comfyui` targets `COMFYUI_URL` (`127.0.0.1:8188`), **not** the browser tab's connected ComfyUI, and failed with `No ComfyUI process found on port 8188 to restart` — while panel tools had been operating happily on the live canvas the whole time.

So the fallback was recommended without ever checking that it points at the same server. When the panel drives a ComfyUI that is not the headless target — a remote, a tunnel, a second instance — the advice is not merely unhelpful, it aims a restart at the wrong machine.

## Why it is fixable now

The information was already here and simply not consulted on this branch. `captureRebootHealthBase(ctx)` returns the ComfyUI the **panel** is actually running inside, and `sameHttpBase()` compares it to `getComfyUIBaseUrl()`. The adjacent decline branch already calls it (`panel-tools.ts:9453`); the timeout branch does not.

And since panel 0.11.54 the panel carries `target` on every `comfy_reboot` reply, sourced from the same `comfyuiUrlForAgent()` value it hands over in `hello` — so the two sides cannot drift.

## The change

Make the recommendation conditional on the targets matching:

- **same server** → unchanged advice; `restart_comfyui` is a correct fallback.
- **different, or unknowable** → say so, name both origins, and do not recommend a restart of a server this call never checked.

The timeout branch is deliberately the only one touched — the decline and success paths already reason about the bound instance.

## Not claimed

The reporter also saw `backend-reconnecting`, then a successful `panel_refresh_nodes`, then `Connected: none` until a browser refresh. That is a reconnect-lifecycle question and is not addressed here; conflating it with the fallback advice would let a wrong recommendation hide behind a genuinely separate symptom.

### DIFF
--- a/src/__tests__/orchestrator/restart-timeout-target.test.ts
+++ b/src/__tests__/orchestrator/restart-timeout-target.test.ts
@@ -1,0 +1,17 @@
+// panel#851 — placeholder pinning the CURRENT wording; the fix follows.
+import { describe, it, expect } from "vitest";
+import { readFileSync } from "node:fs";
+
+const src = readFileSync("src/orchestrator/panel-tools.ts", "utf8");
+
+describe("#851 restart confirmation timeout", () => {
+  it("today: recommends restart_comfyui unconditionally, with no target check", () => {
+    const branch = src.slice(
+      src.indexOf('No confirmation received within'),
+      src.indexOf('No confirmation received within') + 600,
+    );
+    expect(branch).toContain("use restart_comfyui to restart the server directly");
+    // Nothing compares the panel's ComfyUI against the headless target here.
+    expect(branch).not.toContain("captureRebootHealthBase");
+  });
+});


artokun:
## Verification

- **1885 orchestrator tests pass** (100 files).
- CI green on ubuntu / windows / macos, plus pack & install.
- Three mutations, each confirmed applied before its result was trusted, each **killed**:

| mutation | killed by |
|---|---|
| always give the plain recommendation (revert the fix) | proven-different case |
| fold UNPROVEN into DIFFERENT (cry wolf on remote installs) | unproven case |
| degrade `sameHttpBase` to a host-only compare | basePath-mount case |

## A test weakness this fix exposed, and fixed

The first version of this change was covered by **source-text assertions** on the branch. A mutation that inverted the decision — `sameTarget && false` — while leaving both branches present in the source **survived** them. The assertions were checking that the code *contained* the right shapes, not that it *decided* the right way.

That is the same weakness flagged on panel#832, which I acknowledged there rather than fixed. Here it actually bit, so the decision is now a pure helper (`restartTimeoutFallbackAdvice`) driven directly by the tests. The same mutation is killed, and so are two others that source-text assertions could never have caught.

Worth stating as a general point about this file: a branch whose only observable output is *message text* is exactly where source-text tests feel adequate and are not. If the decision has more than two outcomes, it wants to be a function.

## Reviewer notes

The one judgement call I would most like challenged is using `captureRebootHealthBase()` as the same-target oracle. It is documented as a security-grade **proof** for reboot health — null in remote/cloud mode, requires `tabIsLocal` and a server-observed handshake Origin matching the boot base. I am reading a null as "unknown" rather than "different", which is why the third branch exists. If it can ever return non-null while the panel actually fronts a *different* server, that would be a false "proven same" and would restore the original harm.

artokun:
## Answering my own reviewer note, since I could verify it directly

I flagged the risk that `captureRebootHealthBase()` might return non-null while the panel actually fronts a **different** server — which would make the "proven same" branch a false certification and restore the original harm. Read through, it cannot, and the guard is stronger than I assumed:

```ts
const origin = ctx.bridge?.tabServerOrigin?.(ctx.tabId);
if (!sameHttpBase(origin, base)) return null;
```

`tabServerOrigin` is the **server-observed handshake Origin** — set by the browser on the WS upgrade and not forgeable by page JS — compared path-aware against the boot base. It fails closed (null) for a different instance, a different loopback family, a different path, an absent Origin, a DNS-ambiguous `localhost` on either side, and a basePath mount whose pathless Origin cannot prove which mount it fronts. Two of those are recorded as a codex P1 and a coordinator P0, so they are deliberate rather than incidental.

So non-null means *proven same origin as the boot base*, and the "proven same" branch cannot be reached by a foreign server.

**One thing worth naming that I had not considered.** `captureRebootHealthBase` proves the tab matches the **boot** base (`getBootLocalComfyUIBaseUrl()`), while the advice compares against the **current** target (`getComfyUIBaseUrl()`). Those differ after a runtime retarget — and that asymmetry is *correct here*, because the question this advice answers is "would `restart_comfyui` hit the server the panel is on?", and `restart_comfyui` targets the current one. Comparing current-target against proven-panel-origin is exactly right. It would have been wrong to compare boot base against boot base.

Still want the review's independent read on priorities 2, 4 and 6 — particularly whether the proven-different message *understates* the risk. It says the other target "may find nothing there at all", which was this reporter's outcome, but a restart aimed at a *live* wrong server would succeed and take down something the user cares about. That is arguably the worse case and the text does not say so.

artokun:
Closing this **without merging** — it would have regressed `main`.

## What this branch actually contains

One commit, one file, 17 lines: the *placeholder* test, whose own header says `// panel#851 — placeholder pinning the CURRENT wording; the fix follows.` It asserts the **unfixed** behaviour:

```ts
expect(branch).toContain("use restart_comfyui to restart the server directly");
expect(branch).not.toContain("captureRebootHealthBase");
```

Merging it would have overwritten the 75-line behavioural suite now on `main` with a 17-line test that pins the bug — and that test **fails** against the fixed code. Adversarial review caught this; I had believed the branch carried the implementation and it did not.

## Where the fix actually went

`restartTimeoutFallbackAdvice()` and the behavioural tests are on `main` and shipped in **v0.50.64** (`422fb56`, published to npm). So the fix is live — it simply did not travel through this PR.

The cause: my implementation commit landed while the shared working tree was on `release/0.50.64` rather than this feature branch, so it was swept into that release. The push to this branch carried only the earlier placeholder commit. That is a working-tree collision, not a merge conflict, which is why nothing flagged it — `git push` succeeded, CI passed (a placeholder test is green against unfixed code by construction), and the PR looked healthy.

## Two substantive findings from the review, which remain open

Both sit inside the blast radius the original report described, so they are worth their own issue rather than being lost with this PR:

1. **The "second local instance" case still gets the weaker advice.** `captureRebootHealthBase()` is a proof-of-*sameness* oracle: `null` conflates *provably different* with *cannot tell*. When the tab fronts a second local ComfyUI on another port, `tabServerOrigin` correctly reports the mismatch — and that positive evidence is then discarded, so the advice lands in the "unproven" branch rather than "proven different". Only a runtime retarget reaches the strong warning.

2. **"may find nothing there at all" understates the risk.** In the proven-different branch the other target is a *live* server by construction, so restarting it can just as easily **succeed** — taking down a queue the user did not intend to touch. The message names only the no-op outcome.

Confirmed sound by the same review: the oracle cannot produce a false "proven same" (the WS-handshake Origin is server-observed and unspoofable), `sameHttpBase` is path-aware and matches the idiom used at five other call sites, and all three claimed mutation kills reproduce.

Issue #1001, opened 2026-08-07
Panel graph tools need manual rebinding after multi-workflow turn


artokun:
### What happened / root cause
The live panel graph tools initially failed after the session resumed and the user message carried multiple workflow-origin blocks. The error explicitly said the target was ambiguous because the current turn was issued from multiple workflows, and also reported that the panel was still reconnecting after a restart/reload. The graph tools became usable only after explicitly pinning to workflows/video_minimax_h3_t2v.json and then rebinding with panel_set_workflow_target({mode:"current"}).

Two plausible causes:
1. Workflow-origin resolution cannot select the active tab when one turn contains multiple workflow origins; evidence: the returned error says "no connected tab can be chosen ... current turn was issued from multiple workflows at once".
2. Reconnect readiness is reported before the graph tool route is usable; evidence: the same error says "still reconnecting after a restart/reload", while the explicit rebind returned graph_binding:"bound" and subsequent reads succeeded.

### Steps to reproduce
1. Resume a panel session after a reconnect/reload.
2. Send a turn containing more than one workflow-origin context block while a specific workflow is active.
3. Call panel_graph_outline or panel_list_workflows.
4. Observe the target ambiguity/reconnecting error.
5. Call panel_set_workflow_target({mode:"pinned", path:"workflows/video_minimax_h3_t2v.json"}), then panel_set_workflow_target({mode:"current"}); graph reads recover.

### Exact error (scrubbed)
"graph_outline could not reach the ComfyUI panel ... (no connected tab can be chosen for orchestrator::codex: the current turn was issued from multiple workflows at once, so its target is ambiguous. Target a workflow explicitly ... or wait for the next single-origin message.)"

### Fix / workaround
Applied locally: no source patch was available in the workspace. Rebound explicitly to the active workflow and then restored current-tab mode. This recovered the graph tools without changing user data. Requested upstream improvement: allow explicit workflow-origin context to disambiguate automatically, or make panel_set_workflow_target({mode:"current"}) the documented first recovery path when the active tab is already known.

### Environment
OS: Windows 11
ComfyUI: 0.30.0 local
GPU: NVIDIA GeForce RTX 5080 16 GB
CUDA: 13.0
Torch: 2.13.0
Python: 3.13
comfyui-mcp: 0.50.2
panel: 0.11.43

artokun:
[...]

PR #1234, opened 2026-08-09
fix(panel): graph tools should not need manual rebinding after a multi-workflow turn (#1001)

artokun:
Draft — **claim and scoping only. No implementation yet**, and the branch is an empty commit off main so nothing here is reviewable work. Claiming it so it is not double-picked.

## The report

Panel graph tools failed after a session resumed with a user message carrying **multiple workflow-origin blocks**. They became usable only after pinning to a specific workflow and then `panel_set_workflow_target({mode:"current"})`.

The reporter names two plausible causes and, usefully, quotes the error for each:

1. **Origin resolution cannot pick the active tab** when one turn carries multiple workflow origins — *"no connected tab can be chosen … current turn was issued from multiple workflows at once"*.
2. **Reconnect readiness is reported before the graph route is usable** — the same error also says the panel was *"still reconnecting"*.

## Why the two-causes framing matters here

They are not alternatives to choose between — the error text contains **both** clauses, so the message is already reporting two conditions and the user cannot tell which one blocked them. That is the shape this repo keeps producing (#796): distinct states folded into one outcome. If both are real, fixing either alone leaves the report open.

## First steps, in this order

1. **Find the error site and determine whether the two clauses can fire independently.** If they always print together, one of them is decoration and the diagnosis is wrong for half of readers.
2. **Read the consumer of the origin-resolution result before touching the resolver** — the last three issues I worked all turned on a claim that was true at the producer and false one layer down.
3. Check `git tag --contains` on anything already shipped in this area: the report is from 2026-08-07 and the panel has moved several releases since. This may be partly or wholly fixed already, in which case the deliverable is evidence and a close, not a patch.

Explicitly deferred until step 3 answers: any change to reconnect-readiness reporting, which overlaps the tab-readiness work and should not be touched speculatively.

Refs #796.


artokun:
Not needed — #1001 was already fixed by #1047 (shipped v0.50.17); the reporter was on 0.50.2. Evidence and closure posted on the issue instead of a patch. This branch carried no code.

PR #1235, opened 2026-08-09
fix(restart): a CONFIRMED origin mismatch must not be lost inside the proof (#1233)

### LINKED ISSUE #1233: restart-timeout advice: 'unproven' discards positive evidence of a DIFFERENT target, and the warning understates the risk
Follow-up to panel#851, whose fix shipped in **v0.50.64**. Two findings from the adversarial review of that change. Both sit inside the exact blast radius the original report described, so they are filed rather than left in a closed PR.

## 1. `null` conflates "provably different" with "cannot tell"

`restartTimeoutFallbackAdvice()` takes `panelBase` from `captureRebootHealthBase()`, which is a **proof-of-sameness** oracle. It returns non-null only after confirming the tab's server-observed WS-handshake Origin matches the boot base; every other outcome — including a *confirmed mismatch* — collapses to `null`.

So the advice has three branches but the oracle can only distinguish two:

| actual situation | oracle | branch taken |
|---|---|---|
| tab fronts the boot instance | base | proven same ✅ |
| runtime retarget moved the headless target | base | proven different ✅ |
| **tab fronts a SECOND local ComfyUI on another port** | **null** | **unproven** ❌ |

The third row is the scenario panel#851's own motivation cites. `tabServerOrigin` correctly reports `:8189` against a boot base of `:8188`, `sameHttpBase` returns false — and that positive evidence of a mismatch is discarded at `captureRebootHealthBase`'s return, so the caller gets the mild "check that this is the same ComfyUI" rather than the strong "do NOT reach for restart_comfyui".

Better than the pre-fix unconditional advice, but it does not close the case it was written for. The fix is to surface *why* the proof failed — a confirmed origin mismatch is different information from an absent or unprovable origin — rather than to loosen `captureRebootHealthBase`, which is correctly strict for its own purpose.

## 2. The proven-different warning understates the risk

The message says the other target "may find nothing there at all". That was the reporter's outcome, but in that branch `headlessBase` is a **live, different server** by construction. Restarting it can just as easily **succeed** — taking down a process or queue the user did not intend to touch, which on a shared instance is the worse outcome by some margin.

The text should name that case too, not only the no-op one.

## Not in scope

Loosening `captureRebootHealthBase()` itself. It is a security-grade proof used for reboot-health certification, its strictness is deliberate (a codex P1 and a coordinator P0 are recorded in its comments), and this advice path should not be the reason it is weakened.

artokun:
Fixes #1233. Follow-up to panel#851 (shipped in v0.50.64), from the adversarial review of that change.

`captureRebootHealthBase()` is a proof-of-**sameness** oracle — it returns non-null only when the tab's server-observed handshake Origin provably matches the boot base, and collapses everything else, *including a confirmed mismatch*, into `null`.

So the advice had three branches while the oracle could distinguish two:

| actual situation | oracle | branch |
|---|---|---|
| tab fronts the boot instance | base | proven same |
| runtime retarget moved the headless target | base | proven different |
| **tab fronts a SECOND local ComfyUI** | **null** | **unproven** ← wrong |

The third row is the case panel#851 was filed about. `tabServerOrigin` correctly reported the mismatch and it was discarded at the proof's return.

The observed Origin is now passed alongside the proof: a confirmed mismatch reaches the strong warning, a matching-but-unproven origin does **not** manufacture one, and an absent origin stays unproven — absence is not evidence. **The proof itself is untouched**; it is correctly strict for reboot-health certification and this advice path is not a reason to weaken it.

Also widens the warning: it said the other target "may find nothing there at all", but in that branch the other server is **live** by construction, so a restart can SUCCEED and take down a ComfyUI the user did not mean to touch.

## Gates
- 1903 orchestrator tests pass (101 files)
- three mutations killed: discarding the observed origin; treating any observed origin as a mismatch; dropping the succeed-on-wrong-server risk

### DIFF
--- a/src/__tests__/orchestrator/restart-timeout-target.test.ts
+++ b/src/__tests__/orchestrator/restart-timeout-target.test.ts
@@ -117,4 +117,22 @@ describe("#1233 confirmed mismatch reaches the strong warning", () => {
     expect(a).toContain("may find nothing there");
     expect(a).toContain("SUCCEED and take down a ComfyUI you did not mean to touch");
   });
+
+  it("#1233 a basePath mount is NOT a mismatch — the Origin is pathless and cannot say", () => {
+    // Found by probing this fix, not by review: tabServerOrigin carries scheme+host+port
+    // only, so a path-aware compare against a :8188/comfy base read a correctly-mounted
+    // install as a different server and fired the strong warning at the right one.
+    const a = advise("http://127.0.0.1:8188/comfy", null, "http://127.0.0.1:8188");
+    expect(a).not.toContain("Do NOT");
+    expect(a).toContain("could not confirm which one this panel is running inside");
+  });
+
+  it("#1233 a genuinely different PORT is still caught despite the origin-only compare", () => {
+    expect(advise("http://127.0.0.1:8188/comfy", null, "http://127.0.0.1:8189")).toContain("Do NOT");
+  });
+
+  it("#1233 loopback family folding does not manufacture a mismatch", () => {
+    // 127.0.0.1 vs a 0.0.0.0 bind is the same instance; sameHttpOrigin folds the family.
+    expect(advise("http://0.0.0.0:8188", null, "http://127.0.0.1:8188")).not.toContain("Do NOT");
+  });
 });
--- a/src/__tests__/orchestrator/restart-timeout-target.test.ts
+++ b/src/__tests__/orchestrator/restart-timeout-target.test.ts
@@ -135,4 +135,21 @@ describe("#1233 confirmed mismatch reaches the strong warning", () => {
     // 127.0.0.1 vs a 0.0.0.0 bind is the same instance; sameHttpOrigin folds the family.
     expect(advise("http://0.0.0.0:8188", null, "http://127.0.0.1:8188")).not.toContain("Do NOT");
   });
+
+  it("#1233 (codex) `localhost` vs a loopback literal is UNPROVEN, not a mismatch", () => {
+    // loopbackFamily() deliberately excludes localhost (coordinator P0: it can resolve to
+    // either family, or elsewhere), so it canonicalizes unequal to 127.0.0.1 — and firing
+    // the strong warning on that would accuse what is very likely the same instance.
+    expect(advise("http://localhost:8188", null, "http://127.0.0.1:8188")).not.toContain("Do NOT");
+    expect(advise("http://127.0.0.1:8188", null, "http://localhost:8188")).not.toContain("Do NOT");
+    expect(advise("http://localhost:8188", null, "http://[::1]:8188")).not.toContain("Do NOT");
+  });
+
+  it("#1233 (codex) an ambiguous host does not mask a genuinely different PORT", () => {
+    // Degrading to unproven must not become a blanket amnesty: the port still differs.
+    const a = advise("http://localhost:8188", null, "http://localhost:8189");
+    expect(a).not.toContain("Do NOT"); // ambiguous on both sides -> unknown, by design
+    // …but a concrete literal on both sides still proves it.
+    expect(advise("http://127.0.0.1:8188", null, "http://127.0.0.1:8189")).toContain("Do NOT");
+  });
 });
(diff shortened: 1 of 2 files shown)


artokun:
## Self-caught defect, pushed before review returned

I listed "does comparing a pathless Origin against a basePath-carrying base produce a FALSE mismatch" as the priority I was most suspicious of when briefing the review — then probed it instead of waiting. It was real.

`tabServerOrigin` is **pathless**: the browser sends scheme+host+port on a WS upgrade and no path. `headlessBase` may carry a basePath mount (`:8188/comfy`). I compared them with the path-**aware** `sameHttpBase`, so:

```
headlessBase   http://127.0.0.1:8188/comfy
observedOrigin http://127.0.0.1:8188
               -> "Do NOT reach for restart_comfyui…"
```

A correctly-mounted install read as a different server, and the strong warning fired at the server that **is** the right one. That is the cry-wolf failure this three-way design exists to avoid, reintroduced one line over by the fix for it.

**Fixed:** the raw Origin is now compared **origin-only** (`sameHttpOrigin`), so it is used only for the claim it can support — it can prove a different `host:port`, it cannot prove a different path. `captureRebootHealthBase` fails closed on exactly this ambiguity for exactly this reason, so the file already reasons this way.

Three new cases pin it:

| case | expected |
|---|---|
| basePath mount, pathless Origin, same host:port | no warning |
| basePath mount, Origin on a **different port** | warning still fires |
| `0.0.0.0` bind vs `127.0.0.1` Origin (family folding) | no warning |

Mutating the compare back to path-aware is killed.

**Still 1906 orchestrator tests passing** (101 files); 14 on this helper.

Review priorities 1, 2, 5 and 6 remain open and I would still like them checked independently — particularly (1), whether trusting the raw Origin is sound at all given `panelBase` came from a proof, and (2), the full 8-combination enumeration. Priority 3 is answered above; priority 4 is pinned by the family-folding case.

artokun:
## Codex review — 2 blockers. One fixed; one declined on a standing project rule.

### FIXED — `localhost` produced a false "do NOT" warning

`loopbackFamily()` deliberately excludes `localhost`: it can resolve to either family or elsewhere entirely, and a recorded **coordinator P0** keeps it out so it never matches a concrete literal during reboot-health certification. Correct there — and it meant:

```
headlessBase   http://localhost:8188
observedOrigin http://127.0.0.1:8188   -> "Do NOT reach for restart_comfyui…"
```

…at what is almost certainly the same instance.

Same shape as the basePath defect fixed one commit earlier: **an inference drawn from evidence that cannot support it.** An ambiguous host now degrades to UNPROVEN — the direction `captureRebootHealthBase` already takes for the same input — rather than to "different".

`loopbackFamily` is **not** changed. Weakening it would reach into the certification path this advice has no business affecting, and the P0 that keeps `localhost` out is right.

Codex also caught a stale comment on `sameHttpOrigin` claiming `localhost/127.0.0.1/0.0.0.0 compare equal`. They do not — and that sentence is what made the behaviour look deliberate. Corrected.

### DECLINED — raw `Origin` forgeable by a malicious LOCAL process

Accurate as a mechanism: the tokenless loopback listener stores whatever `req.headers.origin` parses to, and a native local WebSocket client can set it, then take over a tab id.

Not acted on, for a stated reason rather than by omission:

- This project's standing constraint is **one machine, one trust domain — accidental bugs only.** A hostile local process is explicitly outside the model.
- **Page JS cannot forge `Origin`** — the browser sets it on the WS upgrade. That is the property this branch relies on, and it holds against the realistic vector.
- Treating a hostile local process as in-scope would equally condemn `captureRebootHealthBase()`, which reads the **same header** and is load-bearing for reboot-health certification. This advice path is not the place to relitigate that.

Recorded here rather than silently ignored, so the reasoning is reviewable.

### Accepted as fair

Codex noted several tests are regression coverage rather than fix-pinning (matching-origin, absent-origin, proof-wins, basePath, `0.0.0.0`). True — they would pass against pre-fix code. The second-port and `localhost` cases are the ones that pin new behaviour, and dropping the localhost guard is mutation-killed.

Clean per review: the 8-combination logic table, the basePath fix, and the retained "so I did NOT restart ComfyUI" claim (the timeout returns immediately after `ctx.confirm`, before any dispatch).

**1908 orchestrator tests pass** (101 files); 16 on this helper.

PR #1236, opened 2026-08-09
chore: delete civitai-lookup.ts, which nothing has ever imported

### LINKED ISSUE #1226: civitai-lookup.ts is dead code — no importers anywhere in src/
Found while inventorying #1136's fetch sites.

`src/services/civitai-lookup.ts` has **no importers**:

```
$ grep -rn "civitai-lookup" src/
(nothing)
$ grep -rn "lookupByHash" src/
src/services/civitai-lookup.ts:20:  export async function lookupByHash(
src/services/civitai-lookup.ts:105:   const result = await lookupByHash(...)   # itself
```

Nothing in `src/` references the module — not production code, not tests. The only call is `batchLookup` invoking `lookupByHash` inside the same file. The live CivitAI path is `civitai-resolver.ts`, used by `tools/missing-models.ts` and `tools/model-extras.ts`.

## Why it is worth an issue rather than a silent delete

It contains a **real bug**, which is how I found it. `lookupByHash` returns `null` for three different outcomes:

| case | line |
|---|---|
| HTTP 404 — genuinely not on CivitAI | `:46` (correct, and the only one that writes to the hasher cache) |
| non-2xx (500 / 429 / auth) | `:53` |
| transport failure (DNS / TLS / refused) | `:71` |

That is exactly the #1136 fold. So the file is not inert-and-harmless — it is a working implementation of a defect, sitting where someone will eventually wire it up and inherit the bug. It also has a hasher-cache integration (`updateCivitaiInfo`) that `civitai-resolver.ts` does not, which is presumably why it was written and may be why it was never deleted.

## Decision needed

Either:
1. **Delete it.** The resolver covers the live path and already distinguishes timeout / network / status / genuine-empty properly.
2. **Revive it deliberately** — if the hasher-cache integration is wanted — in which case the three-way `null` needs fixing first, per #1136.

Not folding this into #1136's PR: patching dead code to look maintained is worse than either option above, and it would let the PR claim a site it did not need to fix.

Refs #1136.

artokun:
Fixes #1226 — option 1 (delete).

No importers anywhere in `src/`; the live path is `civitai-resolver.ts`. Deleting rather than fixing because the module is a working implementation of the #796 three-way-`null` fold, sitting where someone would eventually wire it up and inherit it.

Records one known consequence: this was the only caller of `FileHasher.updateCivitaiInfo`, so those three `file_hashes` columns now have no writer — as in practice they already had none. Removing them is a schema change and out of scope here.

Verified mechanically: no surviving reference in `src/` or the README, `tsc` builds, full suite passes (8040).


artokun:
**Review note — codex was unavailable, so this is my own adversarial pass, disclosed as such.** Codex could not start a process in this sandbox (`Access is denied` before PowerShell executed, including for read-only commands), so it returned "unable to determine" on every point rather than a review. Rather than imply a second opinion exists, here is the verification run directly, covering the five things I asked it to check:

1. **No reference of any kind.** `civitai-lookup`, `lookupByHash`, `batchLookup` across `*.ts/*.mjs/*.js/*.json/*.md/*.yml` in the repo (excluding `node_modules`/`dist`/other worktrees): zero hits after the delete. That covers dynamic `import()`, `require()`, and config references, not just static imports.
2. **No barrel re-export.** No `export *` in `src/services/*.ts` or `src/index.ts`, so nothing could have re-exported it implicitly.
3. **`FileHasher.updateCivitaiInfo` now has no caller** — only its own definition at `file-hasher.ts:108`. This confirms the claim in the commit message rather than contradicting it.
4. **Nothing behaves differently.** The only reader of `file_hashes.civitai_id/_name/_model_id` is `getCached` (`file-hasher.ts:66`), and it will keep returning `null` — as it already did, because the writer was never reached. The `lora_civitai_id` in `generation-tracker.ts` and `civitaiModelId` in `lora-catalog.ts` are different columns on different tables, unrelated to this.
5. **No test was removed or made vacuous.** The diff is exactly two files: the module (111 lines) and one README line. There were never any tests for it — which is part of why it was worth deleting.

`tsc` builds and the full suite passes (8040) unchanged.

Issue #1027, opened 2026-08-07
panel workflow_open can leave graph instance unreadable during tab switches


artokun:
### What happened / root cause
Rapid read-only workflow inspection caused workflow_open to report that the canvas was bound to the requested workflow but its graph differed from the loaded state. Follow-up graph_outline calls then returned workflow instance mismatch, and a subsequent open timed out while the panel was switching. The panel needs serialized switch completion and a fresh instance-fence receipt before accepting the next graph read.

### Steps to reproduce
1. With several workflow tabs open, issue panel_open_workflow followed immediately by panel_graph_outline for multiple tabs.
2. Observe workflow_open errors saying the canvas is bound but nodes/definitions do not match.
3. Observe graph_outline instance-mismatch or switching/refreshing errors; one subsequent workflow_open may time out.

### Exact error
Scrubbed examples:
- workflow_open: canvas IS bound, but graph does not match loaded state (nodes/definitions).
- graph_outline: workflow instance mismatch: command targets a different workflow than active canvas.
- workflow_open: panel tab did not reply within 15000 ms; outcome undetermined.
- graph_outline: panel is switching/refreshing; request was not applied.

### Fix
applied locally: no
upstream-only: yes
Serialize workflow switching (or expose an awaitable switch receipt), re-derive the active workflow instance fence after each successful switch, and make subsequent graph reads wait/retry against that receipt instead of returning stale-instance errors.

### Environment
Windows 11 / ComfyUI 0.30.0 / RTX 5080 16GB / comfyui-mcp 0.50.2 / panel 0.11.43.

artokun:
Recurrence on the current releases: comfyui-mcp 0.50.36 and panel 0.11.44 (Windows 11, local ComfyUI 0.31.0, RTX 5080, Python 3.13).

After a ComfyUI/custom-node restart or refreshed model definitions, `panel_open_workflow` reports: `workflow_open RAN and canvas IS bound to ... but graph differs from loaded state on nodes; UNKNOWN; re-read.` A subsequent `panel_graph_outline` shows the complete expected graph, groups, wiring, and node modes. The report specifically observes custom combo widgets (GGUF/LoRA loaders), null/default widget normalization, and frontend node-size normalization as the apparent non-structural differences. This is a false load-mismatch outcome rather than a failed workflow load.

artokun:
[...]

PR #1237, opened 2026-08-09
fix(panel): workflow_open leaves the graph unreadable during a tab switch (#1027)

artokun:
Draft — **claim and scoping only, empty commit off main.** No implementation yet.

## The report

Rapid read-only inspection: `workflow_open` reported the canvas bound to the requested workflow *but its graph differed from the loaded state*; follow-up `graph_outline` returned **workflow instance mismatch**; a subsequent open timed out mid-switch. The reporter's own diagnosis: the panel needs serialized switch completion and a fresh instance-fence receipt before accepting the next graph read.

## First question: is this still real?

**Reported on mcp 0.50.2 / panel 0.11.43, 2026-08-07 — the same session that produced #1001, which I just closed as fixed in v0.50.17.** Current is v0.50.64. Fence work has shipped since, notably #1201 (v0.50.54, a fence deadlock caused by an old panel).

But I am **not** claiming this one is stale. On #1001 I could point at the exact commit whose in-code comment described the reporter's symptom; here I have only "there has been fence work in the area", which is not the same standard of evidence and would be the #796 fold applied to triage. Two issues from one session do not have to share a fate.

So step 1 is evidence, either way:

1. Find where `workflow instance mismatch` is produced (`orchestrator/panel-tools.ts`) and determine whether a switch-in-progress can still yield it. If #1201 or the identity-stamp work closed this path, close with the commit and `git tag --contains`, as on #1001.
2. If it is still reachable, read the CONSUMER of the fence receipt before touching the fence — every issue I worked today turned on a claim true at the producer and false one layer down.
3. Only then consider serialization. The reporter's proposed fix ("serialize switch completion") is plausible but is a design change; the observable defect may be narrower, e.g. a receipt read before it is refreshed.

## Known adjacency, flagged so it is not rediscovered

There is a documented identity embed/swap race in this area: the first record embeds the pre-swap uuid, so a canvas stamp is stale the instant it is written. That is the same neighbourhood and may be the actual mechanism rather than switch ordering. Worth checking before accepting the reporter's framing.

Related: #1077 (instance fence permanently REFUSED) looks like the same family and may share a root cause — if so these should be worked together rather than patched twice.

Refs #1201, #1077.


artokun:
Not needed — the fence self-repair this report asks for shipped in v0.50.33 (#1043) and v0.50.43 (#932/#1161); reporter was on 0.50.2. Evidence posted on the issue. This branch carried no code. The report's separate timeout symptom is called out there as unverified rather than folded in.

PR #1238, opened 2026-08-09
fix(panel): instance fence permanently REFUSED though other reads confirm the uuid (#1077)

### LINKED ISSUE #1001: Panel graph tools need manual rebinding after multi-workflow turn

artokun:
Draft — **claim and scoping only, empty commit off main.**

## The report

On a remote RunPod session, `panel_graph_outline` and `panel_set_workflow_target({mode:"current"})` were **permanently** stuck refusing the workflow-instance fence, while `panel_list_workflows` and `panel_open_workflow` independently confirmed the **exact same live `workflow_uuid`**. Multiple hard browser refreshes did not clear it.

## Why the easy answer is probably wrong here

Reported on **mcp 0.50.23 / panel 0.11.44** (2026-08-08), and #1043 (*"a fence the panel repaired mid-call is no longer reported as failure"*) first shipped in **v0.50.33** — ten releases later. I closed #1001 and #1027 today on exactly that kind of gap.

**I do not think that closes this one, and the reporter's own evidence is why.** #1043 addresses a fence that the panel *repaired mid-call* being misreported — a transient, self-clearing state. This report is the opposite shape:

- **permanent**, surviving multiple hard refreshes, where the self-repair path is supposed to fire on every refused read;
- **selective** — two other read tools resolve the same uuid successfully at the same time.

A repair mechanism that works would have cleared a transient mismatch on the first retry. Something that survives refreshes while *sibling reads succeed* points at the fence's own adoption path holding a value the other tools do not consult, not at a missed repair. Closing this on the version gap would be pattern-matching two prior closures onto a case whose distinguishing evidence contradicts them.

## First steps

1. **Establish what the two working tools read that the fence does not.** `panel_list_workflows` / `panel_open_workflow` resolving the uuid correctly is the sharpest clue in the report — find the divergent source before looking at the fence at all.
2. Check whether `noteWorkflowInstanceMismatch → sendHello` self-repair can be *starved* — the budget is per-identity and resets when the live uuid changes (`panel-tools.ts:2956-2964`). If the uuid never changes, does the budget ever refill? A permanent refusal is what an exhausted, non-refilling budget would look like.
3. Only then consider whether #1043/#932 already cover it, and require evidence naming this symptom — not just a release number.

**Remote-specific:** the report is from RunPod. Worth checking whether anything here is port/proxy-dependent before assuming parity with local.

Refs #1043, #932, #1027.


artokun:
### Step 1 finding: half this mechanism is not in this repo

`noteWorkflowInstanceMismatch` appears exactly once in `comfyui-mcp`, and it is **inside a comment** (`panel-tools.ts:2960`). There is no implementation here. The self-repair loop the comment describes —

> The panel answers a workflow-instance mismatch by re-advertising its current identity (`noteWorkflowInstanceMismatch` → `sendHello`), and the orchestrator's hello handler adopts a validated identity into the tab's command stamp.

— is split across two repos: **the panel** raises the re-hello, **the orchestrator** adopts it. Only the second half is visible from here.

That matters for this issue specifically, because a **permanent** refusal means the loop is not completing, and the loop has two ends. The reporter was on **panel 0.11.44**, and I have not established what that panel version does on a mismatch. Three possibilities, and I cannot yet distinguish them from this repo:

1. the panel never sends the re-hello (panel-side bug, panel-side fix);
2. it sends it and the orchestrator declines to adopt (this repo);
3. it sends it, adoption succeeds, and the fence reads a *different* value than the one adopted — which would also explain why `panel_list_workflows` and `panel_open_workflow` resolve the same uuid fine.

**(3) is still the hypothesis the reporter's own evidence best supports**, and it is the one testable entirely here: two sibling reads succeeding while the fence refuses is a divergence between what the fence consults and what those tools consult, regardless of which end raises the hello.

Next, in order:
- trace the orchestrator's hello-adoption path and find what the fence compares against, versus what `panel_list_workflows` / `panel_open_workflow` resolve;
- if they read the same source, the divergence is panel-side and this needs the panel repo (and probably a panel issue, not an MCP one);
- check panel 0.11.44 against current before assuming the panel half is even still the same code.

Recording this before going further because it changes where the fix can live, and a fix in the wrong repo would pass every test here and change nothing for the reporter.

artokun:
### The divergence, located: the fence asks a different question than the working tools

`ui-bridge.ts:3677`:

```ts
const stamp = this.resolveTabWorkflowUuid?.(opts.tabId ?? conn.tabId);
const hasTrustedStamp = typeof stamp === "string" && stamp.length > 0;
```

Per the comment immediately above it (#884/#570), that resolver answers with **the workflow the CURRENT TURN was issued for**, captured at user-message dispatch — deliberately *not* the panel's live active workflow, so a mutation conceived on workflow A is declined after a switch to B.

That is the answer to "what do the two working tools read that the fence does not":

| | source |
|---|---|
| `panel_list_workflows`, `panel_open_workflow` | ask the **panel** what is live |
| the instance fence | asks the **orchestrator** what the **turn** was issued for |

So the reporter's central evidence — two reads confirming the exact same live uuid while the fence refuses — is not paradoxical at all. **Both can be correct simultaneously.** The panel is live on X; the turn-origin resolver returns empty or stale; `hasTrustedStamp` is false; the fence fails closed **permanently**, because nothing about refreshing the browser changes what the *turn* was issued for.

That also explains the detail I flagged as the discriminator, and it is the strongest part of the report: **`panel_set_workflow_target({mode:"current"})` refused too.** The tool whose job is to rebind onto the live canvas cannot fix a stamp that is missing on the *turn-origin* side — which is the same neighbourhood as #1001's turn-origin defect, from the same subsystem, one layer along.

This is a hypothesis with a mechanism, not a conclusion. **Not** yet established:
- whether the resolver returns empty or stale in the reporter's scenario, and why — RunPod/remote is called out in the report and I have not tested whether anything here is proxy-dependent;
- whether #1047's alias path-compression (which fixed #1001) already covers the resolver's failure mode — the reporter was on 0.50.23, which HAS #1047 (v0.50.17), so if the cause were the alias inflation it would already be fixed. That is evidence *for* a different resolver failure, and it is the most useful thing to check next.

Next: reproduce a false `hasTrustedStamp` by driving `resolveTabWorkflowUuid` to empty with a live panel, and confirm the refusal is permanent and survives a re-hello. If it does, the fix is orchestrator-side after all and the panel repo is not needed.

artokun:
### #1077 is already partly addressed in main — and the existing note names the probable root cause

`orchestrator/index.ts:2818-2833` carries two `#1077` comments. Someone has been here:

> **#1077** — each gate now says WHICH one refused. All three used to return a bare `false`, so a permanently-wedged session could only be told "the adoption was REFUSED", and the reporter who traced it had no orchestrator log to fall back on. **One of these is structurally unsatisfiable for a relay-backend session, and that is invisible without this.**

Two things follow, and the second is the important one.

**1. My hypothesis was right about the layer, wrong about the mechanism.** I said the fence reads turn-origin while the sibling tools read the panel, and that a missing stamp fails closed permanently. The resolver at `:2814` confirms the read path. But the permanence is not (only) an empty stamp — it is the **adopt** path at `:2819` refusing, which is what `panel_set_workflow_target({mode:"current"})` depends on. The refusal has three gates, and the reporter got a bare "REFUSED" with no way to tell which.

**2. That in-code note is a live lead, not history.** *"One of these is structurally unsatisfiable for a relay-backend session"* — the reporter was on **RunPod**, i.e. remote. If a gate can never be satisfied on a relay backend, that is a permanent, selective, refresh-proof refusal on exactly that deployment, which is the report in one sentence.

So the diagnostics landed; **whether the structurally-unsatisfiable gate was ever fixed is the open question**, and it is the only thing worth checking next. The three gates are `bridge.canReach(tabId)`, `scopeToRealTab(tabId)`, and the identity check below them.

Correcting myself before going further: earlier on this PR I wrote that the self-repair "is not in this repo" because `noteWorkflowInstanceMismatch` only appears in a comment. True of that symbol, and it led me to frame this as possibly panel-side. The **adoption half is very much here**, has been modified for this exact issue, and is where a relay-backend impossibility would live. I should have grepped the issue number before the symbol — `grep -rn "#1077" src/` would have found this immediately and saved the detour.

Next: determine which gate is unsatisfiable on a relay backend, and whether main still refuses there or only reports better.

artokun:
Not needed — #1077 is a panel version gap: refreshFenceFromOwnReply (#1161) needs a workflow_uuid the panel only publishes from 0.11.45, and the reporter was on 0.11.44. The fix was present and inert. #1199 (v0.50.54) now names the version gap explicitly. Evidence on the issue; this branch carried no code.

Issue #509, opened 2026-07-30
panel_restart_comfyui times out after server is healthy again


artokun:
### What happened / root cause
`panel_restart_comfyui({force:false})` successfully restarted ComfyUI, but the tool call remained pending until the client-side 300-second tools/call timeout. A subsequent `health_check` immediately showed the restarted local ComfyUI reachable and idle. The operation therefore completed but its response was lost or the reconnect waiter exceeded the outer tool timeout.

### Steps to reproduce
1. On a local ComfyUI with an idle queue, call `panel_restart_comfyui({force:false})` after changing a custom-node dependency.
2. Wait for ComfyUI to restart and become reachable.
3. Observe that the tool call does not return and fails after 300 seconds with `timed out awaiting tools/call after 300s`.
4. Call `health_check`; ComfyUI is reachable, GPU is visible, and queue is idle.

### Exact error (scrubbed)
```
tool call error: tool call failed for `panel/panel_restart_comfyui`
Caused by:
    timed out awaiting tools/call after 300s
```

### Fix
Applied locally: no. Upstream-only/intermittent reconnect path; the server and panel transport are already healthy after the timeout, so changing the running panel code during this workflow repair would add risk.

Precise change needed: bound the restart/reconnect readiness waiter below the outer MCP tools/call timeout and persist/return the restart result after reconnect. If the ComfyUI health probe succeeds, return `{rebooting:true, ready:true}` rather than leaving the original call unresolved. Add a regression test where the server disconnects during the command and reconnects before the deadline.

### Environment
Linux; ComfyUI 0.26.2; NVIDIA RTX 3090 24 GB; comfyui-mcp 0.48.19; comfyui-mcp-panel 0.9.9.

PR #1240, opened 2026-08-09
fix(fence): a relay session can adopt a workflow fence — it knows its own origin

artokun:
Addresses **Finding 1** of #1077, which the reporter confirmed from source and marked upstream-only. The upstream part turns out not to be needed.

`attachRelayConnection(sock)` passed no `serverOrigin`, so every relay-multiplexed connection had `serverOrigin === undefined`, `workflowIdentityParts()` returned undefined, and the fence validator refused unconditionally — permanently, for the whole session, no matter how many times the tab was refreshed.

The relay protocol does not need to forward the browser's Origin: the Origin identifies where the panel page is *served from*, which is the ComfyUI the session already points at, and `setupRelayBridge` is holding that URL.

Also moves the refusal wording, which named the relay backend as the known cause — true when written, false now.

**Finding 2** (the same symptom on a cloudflared session, cause not pinned down) is NOT addressed; the issue stays open for it.

### DIFF
--- a/src/__tests__/orchestrator/fence-refusal-reason.test.ts
+++ b/src/__tests__/orchestrator/fence-refusal-reason.test.ts
@@ -7,11 +7,17 @@
 //
 // That is unactionable in general and WRONG in one specific case. Identity is
 // bound to the connection's server-observed Origin, and a relay-backend
-// connection has none: `attachRelayConnection(sock)` calls `handleConnection(sock)`
+// connection had none: `attachRelayConnection(sock)` called `handleConnection(sock)`
 // with no origin argument, and the relay protocol does not forward the browser's
-// handshake Origin. So `workflowIdentityParts()` can never validate, the fence can
-// NEVER be adopted, and no amount of refreshing changes it. The reporter refreshed
-// repeatedly and closed and reopened the tab before tracing it to source.
+// handshake Origin. So `workflowIdentityParts()` could never validate, the fence
+// could NEVER be adopted, and no amount of refreshing changed it. The reporter
+// refreshed repeatedly and closed and reopened the tab before tracing it to source.
+//
+// THE RELAY HALF IS NOW FIXED and the wording moved with it: the Origin identifies
+// where the panel page is SERVED FROM, which is the ComfyUI the session already
+// points at, so `setupRelayBridge` supplies it from COMFYUI_URL and no relay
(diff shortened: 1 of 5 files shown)


artokun:
**Review note — codex could not run, so this is my own adversarial pass, disclosed.** Codex's shell was blocked at the process level (`CreateProcessAsUserW: Access is denied`) and it correctly declined to review code it could not read, rather than guessing. Here are the six checks I asked it for, run directly:

1. **Can the derived Origin disagree with the browser's?** Yes — `COMFYUI_URL=http://127.0.0.1:8188` while the panel is loaded through a reverse proxy at `https://comfy.example.com`. It does not break the session: adoption and validation both use the derived value consistently, so a relay session is internally coherent. The RunPod case in the report is the one where they agree exactly, since the page IS served from `COMFYUI_URL`.
2. **Cross-transport mismatch?** Real: a fence adopted over loopback (observed origin) may not validate after switching to relay (derived origin), and has to be re-adopted with `panel_set_workflow_target`. That is correct behaviour for a transport change, and strictly better than the status quo of never being adoptable at all. Called out rather than hidden.
3. **Security?** None in this project's trust domain. The value scopes an identity, it does not authorize anything, and it comes from the orchestrator's own config — not from the remote peer, so "never client-supplied" still holds.
4. **Other bogus-but-truthy origins?** Two, now pinned: `data:` URLs and non-special schemes (`foo://bar`) both produce the literal string `"null"`, which is truthy. Also found something worth its own test — `COMFYUI_URL` can carry basic-auth credentials, and the refusal message prints the origin. `URL.origin` strips userinfo, so it is safe today; that is now asserted so a hand-rolled "scheme + host + port" cannot reintroduce the leak.
5. **Is `local: false` right for a relay socket?** Yes — `local` means "token-less loopback primary", and its own doc names a relay shim as false (#509).
6. **The reworded refusal** drops the relay pointer, which is no longer the cause, and points at `COMFYUI_URL`, which is what actually supplies the origin. A test pins that the stale advice cannot come back.

One thing the review found in the diff itself: the `serverOrigin` doc still said *"Undefined for non-browser/relay connections"*, which this change makes false. Fixed in the follow-up commit — a stale contract is what someone reads instead of the code.

Full suite: 8058 passing. Five mutations killed, including reverting the call site to the bare attach that was the original bug.

Issue #1175, opened 2026-08-09
panel_restart_comfyui cannot use live sidebar target when ComfyUI is not on port 8188


artokun:
### What happened / root cause
The sidebar panel is connected and live graph reads work against the user's ComfyUI tab, but panel_restart_comfyui refuses because the orchestrator only probes the configured headless target http://127.0.0.1:8188. This local ComfyUI is launched by an external Windows launcher on a different origin/port, so the tool cannot correlate the live panel session with the process/reboot endpoint and blocks a required panel upgrade restart.

### Steps to reproduce
1. Connect the sidebar panel to a local ComfyUI instance not listening on 127.0.0.1:8188.
2. Confirm panel_graph_outline / panel_query_graph work.
3. Upgrade the panel pack with install_comfyui(action:"panel", panel_action:"sync").
4. Call panel_restart_comfyui.
5. The tool refuses before stopping anything, even though the live panel bridge is connected.

### Exact error (scrubbed)
"Refusing to restart ComfyUI: the ComfyUI this restart would stop could not be identified from here (nothing was found listening on port 8188) ... restart it from whatever launches it."

install_comfyui(action:"environment") likewise reports api_target http://127.0.0.1:8188 ECONNREFUSED while the sidebar panel remains connected and can read the live canvas.

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: when a connected sidebar panel exists, derive the active ComfyUI origin/port and reboot route from that bridge/session (or delegate the Manager reboot request through the panel's same-origin API) instead of relying only on COMFYUI_URL/port 8188. Keep the existing busy guard and proof-of-relaunch safety checks; if an external launcher truly cannot be relaunched, return a structured externally_managed status early.

### Environment
Windows 11; ComfyUI 0.27.0; ComfyUI frontend version unavailable because /system_stats was queried only on the mismatched 8188 target; NVIDIA RTX 5070 12 GB; comfyui-mcp running status reported 0.50.46; panel loaded in the browser was 0.11.28, synced and disk-verified as 0.11.45 but not yet reloadable without a manual restart/hard-refresh. ComfyUI-Manager 3.35.

artokun:
#1181 fixes the diagnosis half. The retarget half is a real design change and stays open — details below.

[...]

PR #1246, opened 2026-08-09
fix(target): 127.0.0.1 and localhost are the same host, not a target drift

artokun:
Addresses the **recurrence** reported on #1175 (comfyui-mcp 0.50.64, panel 0.11.59, Windows 11 ComfyUI Desktop).

The panel is on `http://127.0.0.1:8188`; the headless target is `http://localhost:8188`. Both reach the same instance — the headless restart succeeds and later probes `127.0.0.1:8188` — but the origins are compared as raw strings, so the orchestrator reports them as a DIFFERENT address and advises pointing `COMFYUI_URL` somewhere it is already pointed.

`describeTargetDrift` (`src/comfyui/fetch.ts`) does `distinct.includes(want)`, which cannot see that two spellings of loopback name one host.

Draft while I map every origin comparison this affects.

### DIFF
--- a/src/__tests__/comfyui/loopback-alias-origin.test.ts
+++ b/src/__tests__/comfyui/loopback-alias-origin.test.ts
@@ -1,0 +1,117 @@
+// #1175 (recurrence) — two spellings of loopback read as two different servers.
+//
+// The reporter's panel was on `http://127.0.0.1:8188` while the headless target
+// was `http://localhost:8188`. Both reach the same ComfyUI — the headless
+// restart succeeded and then probed 127.0.0.1:8188 — but the origins were
+// compared with `Array.includes`, so the orchestrator reported "a DIFFERENT
+// address" and told them to point COMFYUI_URL at the origin it was already
+// pointed at.
+//
+// That is a diagnosis that sends the reader to fix something that is not broken,
+// which is worse than no diagnosis: the previous message at least admitted it
+// did not know.
+
+import { afterEach, describe, expect, it } from "vitest";
+
+import { canonicalOrigin, sameOrigin } from "../../utils/origin.js";
+import { describeTargetDrift, setConnectedPanelOrigins } from "../../comfyui/fetch.js";
+
+afterEach(() => setConnectedPanelOrigins(null));
+
+describe("canonicalOrigin (#1175)", () => {
+  it.each([
+    ["http://127.0.0.1:8188", "http://localhost:8188"],
+    ["http://localhost:8188", "http://localhost:8188"],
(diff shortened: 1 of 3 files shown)


artokun:
**Review note — codex was blocked, so this is my own adversarial pass, disclosed.** Codex was denied process creation in this worktree's sandbox (`CreateProcessAsUserW failed: 5 Access is denied`) and correctly refused to review code it could not read. It ran fine on #1242 an hour earlier, so this is intermittent. Its six checks, run directly:

1. **Is collapsing `localhost` with `127.0.0.1` ever wrong?** Yes, in three ways — `localhost` resolving to `::1` while the server binds only v4, split-horizon hosts files, and a container where the browser's localhost is not the orchestrator's. In every one, the message the change produces is *better* than the one it replaces: "the browser can reach it and this process cannot — a firewall, a container boundary, or a server bound to one interface" describes a v4/v6 split or a container boundary accurately, where "point COMFYUI_URL at that origin" sent the reporter to change a setting that was already correct.
2. **Odd inputs** — found a real gap: `localhost.` with a trailing dot is the fully-qualified spelling of the same name and a browser sends it verbatim, but it was not collapsed. Fixed, and stripped for the *lookup only* so an unrelated host's spelling is not rewritten. Userinfo cannot leak (rebuilt from protocol+hostname+port), uppercase scheme/host normalise, punycode is left alone, non-http schemes keep their scheme.
3. **`URL.hostname` for IPv6** — returns `"[::1]"` **with** brackets. So the bare `"::1"` I had in the set was unreachable dead code. Removed, with the reason recorded.
4. **`sameOrigin` returning false for two unparseable inputs** — correct at this call site. The caller uses it to decide whether to report a mismatch, so making junk equal to junk would silently suppress every real one.
5. **Other raw-string origin comparisons.** Three, none of them this defect: `session-store.ts:78` uses the origin as a presence GATE and never as a comparison key (both call sites read only `identity.uuid`) — I had a fix written for it and reverted it rather than ship inert code with a comment claiming otherwise; `download-cache.ts:1702` is a redirect same-origin check for *remote* downloads, where collapsing loopback would be irrelevant at best and weakening at worst; `ui-bridge.ts:2809` dedupes origins for display, where showing both spellings is fine and the new message explains them.
6. **Does the wording assert anything untrue?** No — and the "same host" clause only appears when the two spellings actually differ, which is pinned in both directions.

Ten mutations across both passes. Nine killed; the tenth — building the origin by splicing `URL.origin` rather than from parts — is an equivalent mutant, since `URL.origin` already strips userinfo. Recording that rather than writing a test that would only appear to kill it.

Full suite: 8104 passing.

PR #1247, opened 2026-08-09
fix(panel_ask): the timeout must not offer a cause it already ruled out (#1243)

### LINKED ISSUE #1243: Clarify and structure panel_ask unanswered timeout
`panel_ask` on an interactive ComfyUI tab waited for its hard 285-second ceiling and then returned an error result saying either the card was not answered or no interactive surface rendered it. The implementation intentionally clamps with `ASK_TOTAL_BUDGET_CAP_MS = 285000`, but the public tool description says it blocks until the user picks and does not disclose the timeout. `askTimeoutResult` also conflates an unanswered rendered card with a missing surface even though `ensureReachable` already ran.

### Steps to reproduce
1. Open an interactive ComfyUI tab with the sidebar connected.
2. Call `panel_ask` with two options.
3. Do not click either option before the internal 285-second ceiling.
4. Observe an `isError:true` tool result: "The question card was not answered in time (or no interactive panel surface rendered it...)" and the agent must fall back to plain chat.

### Exact error (scrubbed)
`Error: The question card was not answered in time (or no interactive panel surface rendered it - e.g. an exec/headless run), so nothing was selected.`

### Fix
A safe fix needs orchestrator semantics, not a panel-data patch: disclose the hard timeout in the `panel_ask` tool description; distinguish `timed_out_unanswered` from `surface_unavailable`; and return a structured, non-transport-error outcome for an ordinary unanswered timeout so controllers can ask in plain text immediately. Keep the durable late-answer journal behavior from #486. Avoid changing user workflow data or silently choosing a default.

### Environment
Windows 11; ComfyUI 0.31.0; ComfyUI frontend 1.48.7; NVIDIA GeForce RTX 4080 SUPER 16 GB; comfyui-mcp 0.50.63; panel 0.11.56.

artokun:
Fixes #1243. Two things `panel_ask` told the agent that were not true together.

**1. The description hid the ceiling.** It said it "BLOCKS until they pick" and never mentioned `ASK_TOTAL_BUDGET_CAP_MS` (285s). An agent budgeting a turn had no way to know one unanswered question costs the better part of five minutes before it learns anything.

**2. The timeout offered a cause that had already been excluded.** It said *"or no interactive panel surface rendered it — e.g. an exec/headless run"* — but `ctx.ensureReachable()` had already run for this ask. Offering it sent the agent to re-invoke "from an interactive ComfyUI tab" it was demonstrably already on: a remedy the evidence in hand had ruled out.

It now says the card was delivered and went unanswered for N seconds — **a timeout, not a delivery failure** — and that the user may simply not have been at the screen, which is the likely explanation.

`surfaceConfirmed` is **threaded, not assumed**: `ctx.ensureReachable` is optional-chained at the call site, so when no check ran the alternative is still live and still named. Naming it unconditionally was the bug; suppressing it unconditionally would be the same bug pointing the other way.

## Verification

1914 orchestrator tests pass. Tests call `askTimeoutResult` **directly** — an earlier version asserted on source text and a mutation flipping the flag to a constant survived, because both branches remained in the file. Three mutations now die: forcing either branch, and dropping the elapsed figure.

### DIFF
--- a/src/__tests__/orchestrator/ask-timeout-honesty.test.ts
+++ b/src/__tests__/orchestrator/ask-timeout-honesty.test.ts
@@ -1,0 +1,52 @@
+// #1243 — panel_ask's timeout told the agent two things that were not true together.
+//
+// 1. The tool description said it "BLOCKS until they pick" and never mentioned that the
+//    wait is capped at ASK_TOTAL_BUDGET_CAP_MS (285s). An agent budgeting a turn had no
+//    way to know a single unanswered question costs five minutes.
+// 2. The timeout result offered "or no interactive panel surface rendered it" as an
+//    alternative cause — but `ctx.ensureReachable()` had already run for this ask, so
+//    that cause was ALREADY excluded. Offering it sends the agent to re-invoke "from an
+//    interactive ComfyUI tab" it was demonstrably already on.
+//
+// The surface flag is threaded rather than assumed because `ensureReachable` is
+// optional-chained at the call site: when no check ran, the alternative is still live
+// and is still named.
+import { describe, it, expect } from "vitest";
+import { readFileSync } from "node:fs";
+
+const src = readFileSync("src/orchestrator/panel-tools.ts", "utf8");
+
+describe("#1243 panel_ask timeout honesty", () => {
+  it("the tool description discloses the ceiling instead of promising an unbounded block", () => {
+    const at = src.indexOf('"panel_ask",');
+    const desc = src.slice(at, at + 1200);
+    expect(desc).toContain("does NOT block forever");
+    expect(desc).toMatch(/~?285s/);
+  });
+
+  it("a confirmed surface yields a TIMEOUT message, not a delivery-failure guess", () => {
+    const fn = src.slice(src.indexOf("function askTimeoutResult("), src.indexOf("function askTimeoutResult(") + 1800);
+    expect(fn).toContain("this is a timeout, not a delivery failure");
+    expect(fn).toContain("may simply not have been at the screen");
+  });
+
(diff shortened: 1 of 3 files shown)


artokun:
## Self-review before merge — two findings, both mine, both fixed

Codex is structurally unable to return findings here (one task call per handoff, no polling, barred from reviewing from a diff — three attempts this session), so this is a **disclosed self-review**, not an independent one.

**1. The flag was hardcoded, making one branch unreachable.** The call site passed a literal `true`, and it is the only caller — so the no-surface branch existed only for tests. That is worse than the conflation it replaced, because it *looks* like the flag is doing work while production can only ever take one path.

Now derived from the honest source: `typeof ctx.ensureReachable === "function"`. The reachability call is optional-chained, so a ctx without it genuinely ran no check and the alternative cause is genuinely still live. Both branches are now reachable for the reason each claims.

**2. My tests could not see the call site.** Mutating it to `true` left all five behavioural tests green — they call `askTimeoutResult` directly. Helper coverage cannot detect a wiring change; that is a documented lesson in this project and I hit it anyway.

Added a call-site assertion, including an explicit check that the literal does **not** return. That mutation now dies.

## Mutation summary

| mutation | result |
|---|---|
| force the no-surface branch | KILLED |
| force the confirmed branch (over-correct) | KILLED |
| drop the elapsed figure | KILLED |
| hardcode the call-site flag | KILLED (after the wiring test) |

The second matters as much as the first: naming the ruled-out cause unconditionally was the bug, and suppressing it unconditionally would be the same bug pointing the other way. Only a test that exercises the false branch separates them.

1914 orchestrator tests pass.

PR #1250, opened 2026-08-09
test(recovery): stop panelRecoveryContext wording depending on which test ran first

### LINKED ISSUE #1222: Three flaky tests share one cause: panelRecoveryContext reads a module-level cache no test resets
Follows #1208. That issue fixed two flakes individually; a **third** in the same family surfaced today, so the pattern is worth naming at its source rather than patching a third assertion.

## The three

All three assert on remedy wording produced through `panelRecoveryContext()`:

1. `an UNREADABLE disk version leads with the two free checks, not with an update` — fixed in #1216 (it compared `search()` indices without asserting presence, so the *other* wording yielded `-1`).
2. `#812/#823: the remedy names how to reach install_comfyui(action:'panel') when it is not in the tool list` — failed once, then 3/3 green on unchanged re-runs.
3. (the same family — the wording differs by `installPanelUsable`.)

## The shared cause

`panelRecoveryContext()` (`src/services/panel-recovery.ts`) decides `installPanelUsable`, and the message has **two entirely different wordings** depending on it:

- `true` → *"Run install_comfyui(action:'panel', panel_action:'update')"*
- `false` → *"Update the panel ON THE COMFYUI HOST — …"*

One input to that decision is `lastPanelBaseResolution()`, which reads a **module-level cache** in `src/services/panel-workspace.ts`:

```ts
export function lastPanelBaseResolution(): PanelBaseResolution | undefined {
  return cachedResolution();
}
```

```ts
if (resolution && !resolution.base) {
  return { installPanelUsable: false, blocker: "no-local-workspace" };
}
```

So whether a **neighbouring test** happened to prime that cache — and with what — decides which wording a later test sees. Tests in a file share a module registry, so this is order-dependent, which is exactly why it presents as intermittent rather than as a consistent failure.

There is already a reset hook (`__resetPanelBaseCache()`, which clears both `cached` and `diskObservation`), but the affected tests do not call it.

## Why it is worth fixing at the source

Each instance costs the same thing, and it is not the re-run: a flake here is **indistinguishable from a real regression**, so every occurrence is paid for in time spent proving a change innocent. One of these blocked a release build on all three platforms; two caused me to commit while the suite was red and then re-verify.

## Suggested fix

Reset the cache between tests that exercise recovery wording — a `beforeEach(() => __resetPanelBaseCache())` in the affected files — so `installPanelUsable` is a property of the test's own setup rather than of whichever test ran before it. Then any remaining variation is a real signal.

Worth checking whether other module-level caches in this area (`diskObservation`, the object-info cache) have the same exposure.

artokun:
Fixes #1222.

`panelRecoveryContext()` decides `installPanelUsable`, which selects between two entirely different remedy wordings. One input is `lastPanelBaseResolution()`, a module-level cache in `panel-workspace.ts` that no affected test resets — so which wording a test sees depends on whichever test primed the cache before it.

A reset hook (`__resetPanelBaseCache()`) already exists and is simply not called. Three flakes in this family so far; a flake here is indistinguishable from a real regression, which is what it actually costs.

Draft while I work it.

### DIFF
--- a/src/__tests__/services/panel-base-clear-epoch.test.ts
+++ b/src/__tests__/services/panel-base-clear-epoch.test.ts
@@ -39,6 +39,7 @@ const {
   lastPanelBaseResolution,
   __resetPanelBaseCache,
   __setPanelBaseForTests,
+  clearPanelDiskObservation,
 } = await import("../../services/panel-workspace.js");
 
 beforeEach(() => {
@@ -125,6 +126,39 @@ describe("a clear beats a probe that started before it (#1222)", () => {
     const inFlight = primePanelBase({ force: true });
     __resetPanelBaseCache();
     __resetPanelBaseCache();
+    await inFlight;
+
+    expect(lastPanelBaseResolution()).toBeUndefined();
+  });
+});
+
+// CODEX REVIEW asked which OTHER paths clear the cache, and found a production
+// one: `clearPanelDiskObservation()` also drops the resolved base, and it runs on
+// every panel hello — the moment ComfyUI may have restarted with different launch
+// flags and therefore a different custom_nodes tree. Its own comment says keeping
+// the previous root would "let the very next operation freeze the wrong tree",
+// which is exactly what an in-flight probe does when it writes its pre-hello
+// answer back in a moment later. Tests were covered by __resetPanelBaseCache;
+// production was not.
+describe("a hello's clear also beats an in-flight probe (#1222)", () => {
+  it("clearPanelDiskObservation is not undone by a probe that predates it", async () => {
+    __setPanelBaseForTests("/pre-restart/base");
+    const inFlight = primePanelBase({ force: true });
+
+    clearPanelDiskObservation();
+    expect(lastPanelBaseResolution(), "the hello cleared it").toBeUndefined();
+
+    await inFlight;
+
+    expect(
+      lastPanelBaseResolution(),
+      "the pre-restart root must not be frozen back in",
+    ).toBeUndefined();
+  });
+
+  it("holds for the empty-cache start, the case reference comparison cannot see", async () => {
+    const inFlight = primePanelBase();
+    clearPanelDiskObservation();
     await inFlight;
 
     expect(lastPanelBaseResolution()).toBeUndefined();
--- a/src/__tests__/services/panel-base-clear-epoch.test.ts
+++ b/src/__tests__/services/panel-base-clear-epoch.test.ts
@@ -1,0 +1,132 @@
+// #1222 — a background prime landed its answer AFTER the cache was cleared.
+//
+// Three tests in the recovery-wording family flaked, and the wording is what
+// makes it expensive: `panelRecoveryContext()` picks between two entirely
+// different remedies depending on `installPanelUsable`, one input to which is
+// the panel-base cache. A test that inherits a primed cache gets the other
+// remedy and fails on an assertion about English, which is indistinguishable
+// from a real regression.
+//
+// MY FILED DIAGNOSIS WAS WRONG, and this file exists partly to record that: the
+// issue said the affected tests never call `__resetPanelBaseCache()`. They all
+// do, and had for days. The actual cause is written down in
+// ui-bridge.test.ts's own comment — a capability refusal fires an unawaited
+// `void primePanelBase()`, and "a late resolution lands in whichever test is
+// running when it settles". A reset cannot fix that: the write arrives mid-test,
+// after the reset has run.
+//
+// WHY THE EXISTING GUARD MISSED IT. `primePanelBase` already refuses to clobber
+// a newer write, by capturing `cached` at probe start and comparing by
+// reference. That has exactly one blind spot and it is the common case: a probe
+// that starts with an EMPTY cache records `undefined`, a clear leaves it
+// `undefined`, and the two compare equal — so the clear is invisible and the
+// stale answer is written in. A clear is an EVENT, not a value, so it is now
+// counted; `undefined → undefined` is indistinguishable by reference and
+// unmistakable by count.
+//
+// The mitigation until now was POSITIONAL — those tests sit last in their
+// describe so a late resolution has nothing left to perturb. That works until
+// someone adds a test after them.
+
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+
+// The race is a SEQUENCE here, not a timing hope: `primePanelBase` awaits at
+// least once before it writes, so clearing synchronously after the call — and
+// awaiting afterwards — puts the clear strictly between probe-start and
+// probe-settle every run. No sleeps, no ordering luck, no mocks.
+const {
+  primePanelBase,
+  lastPanelBaseResolution,
+  __resetPanelBaseCache,
+  __setPanelBaseForTests,
+} = await import("../../services/panel-workspace.js");
+
+beforeEach(() => {
+  __resetPanelBaseCache();
+});
+
+afterEach(() => {
+  __resetPanelBaseCache();
+});
+
+describe("a clear beats a probe that started before it (#1222)", () => {
+  it("does not repopulate the cache after __resetPanelBaseCache", async () => {
+    // Seed, so there is something for a stale probe to differ from.
+    __setPanelBaseForTests("/seeded/base");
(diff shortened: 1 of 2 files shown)


artokun:
### Step 2 answered, and the global fix is not available — four tests actively depend on the leak

**This is test isolation, not a user-facing staleness bug.** Established before changing anything, since the same cache picks which remedy a real user sees. `cachedResolution()` (`panel-workspace.ts:224-236`) invalidates on three conditions: a target-key change, a target **generation** bump (so a restart onto a different `--base-directory` cannot freeze the old root), and a 60s TTL. In production those fire. Inside one test file none of them do — the target and generation do not change and 60s does not elapse — so a resolution primed by one test is inherited by the next.

`__resetPanelBaseCache()` **already exists** (`:335`) and four suites call it. The five files that assert the remedy wording call none:

```
panel-session-rebind-residuals  reset=0
panel-installer                 reset=0
panel-pin-bypass                reset=0
panel-pin-cancel                reset=0
panel-pin-guard                 reset=0
```

### What I tried, and what it exposed

The structural fix is a global `beforeEach(__resetPanelBaseCache)` in `setupFiles` — leaked state becomes unrepresentable instead of every new test having to remember, which matters because #1208 patched two of these individually and a third surfaced anyway.

**It fails 4 tests**, including `panelStatus records the on-disk version so the bridge can see it` and `an ABSENT pack clears the observation — never a stale 'your install is fine'`.

Those tests **rely on the cache surviving between tests in their file**. That is a stronger statement of the problem than the issue makes: the leak is not merely tolerated, some tests are load-bearing on it. Any global reset therefore has to be paired with making those four prime their own state — which is a bigger change than this issue describes and needs its own reading of what each is actually asserting.

Reverted rather than left half-applied. Next: read those four to determine whether each is genuinely testing cross-call caching (legitimate, should prime explicitly) or accidentally inheriting (a bug of the same family). Not guessing which, and not shipping a per-file patch that repeats #1208's altitude mistake a third time until I know.

artokun:
### Attempt 2 also wrong, and the reason narrows the problem usefully

I made the four inheriting tests call `await primePanelBase()` so the global reset would be safe. **All four still failed, and worse than before:**

```
AssertionError: expected '0.11.62' to be '0.11.38'
```

`0.11.38` is what the test writes into its temp `PANEL_DIR`. `0.11.62` is the **real panel installed on this machine**. So `primePanelBase()` resolves against the live environment, and my "fix" actively overrode the state the test had set up — a worse failure than the leak, and one that would have been invisible on a machine with no panel installed.

So the mechanism is not what I said in the previous comment. The inherited value is not merely *a* primed resolution — it is a resolution that points at the test's temp dir, established by earlier setup in that file. Clearing it does not fall back to a neutral state; it falls back to **the developer's real install**. That is why the global reset breaks these four and why re-priming does not rescue them.

Which makes the fix a different shape than either attempt:

1. the resolution has to be pinned to the test's `PANEL_DIR` per test, not inherited and not re-resolved from the environment;
2. only then is a global reset safe.

Both my attempts are reverted; the branch is back to a green 146/146 on that file.

Recording this rather than continuing, because I have now been wrong twice about the mechanism in the same session, on a bug whose entire subject is a wrong-altitude fix applied twice (#1208). A third guess would be the same mistake in a third form. What is needed next is reading how `PANEL_DIR` becomes the resolved base in this suite — which file-level setup establishes it, and whether it can be made per-test — before touching either the harness or the tests again.

**Not shipping a per-file patch in the meantime.** That is what the reporter filed this issue to stop, and the leak is not currently breaking CI.

artokun:
**Note for whoever else is on #1222** — a claim commit (`4afa5ef`) landed on this branch while I was working it, so we collided. I rebased my two commits **on top of** that claim rather than force-pushing over it; nothing was discarded, and the branch is now on 0.50.70.

If you were about to start: the implementation is done here, and the filed diagnosis is worth reading before you do anything with it, because **it was wrong**.

The issue says the affected tests never call `__resetPanelBaseCache()`. All four do, and had since 4 August. The real cause is recorded in `ui-bridge.test.ts`'s own comment — a capability refusal fires an unawaited `void primePanelBase()`, and "a late resolution lands in whichever test is running when it settles". No amount of resetting fixes that, because the write arrives mid-test, after the reset.

What was actually broken: `primePanelBase` already guards against clobbering a newer write, by capturing `cached` at probe start and comparing by reference. That has one blind spot and it is the common case — a probe that starts with an EMPTY cache records `undefined`, a clear leaves it `undefined`, so the clear is invisible and the stale answer gets written in. The background prime fires *because* nothing was primed yet, so the blind spot is the normal path.

A clear is an event, not a value, so it is now counted.

**Also, and this is the part worth your attention: the same hole exists in production.** `clearPanelDiskObservation()` drops the resolved base too, and runs on every panel hello — the moment ComfyUI may have restarted with different launch flags. Its own comment says keeping the old root would "let the very next operation freeze the wrong tree", which is exactly what a late probe does. Both clears now go through one `forgetResolvedBase()` so a fourth call site can't reintroduce it silently.

artokun:
### Why the global reset broke those four — and why it can never work here

`panel-recovery-cluster.test.ts:211-226` **already resets the cache**, in both `beforeEach` and `afterEach`, and sets `workspace.base = root` alongside it. So that file was never leaking. My global reset did not fix it; it *broke* it.

The mechanism is the import, not the reset. `setupFiles` runs **before** a test file's `vi.mock` factories apply. My `import { __resetPanelBaseCache } from "../../services/panel-workspace.js"` therefore pulled the real module graph in early — including the real `config.js` and `workspace-env.js` that this file mocks at `:51` and `:65`. That is where `0.11.62` came from: the developer's actual install, resolved through unmocked config, because the module had already been loaded past the mock boundary.

So the "make it unrepresentable globally" instinct was wrong for a structural reason, not a tuning one: **a setup file cannot import a product module whose dependencies individual suites mock.** Worth stating plainly because it looked like the principled fix, and it is the shape I would have reached for again.

### What that leaves

The reset belongs per-file after all — but now for a reason, rather than as the path of least resistance:

- `panel-recovery-cluster` (resets, both hooks) is the model, and is **not** one of the flaky files;
- the five that assert the remedy wording and reset nothing are where #1222's flakes actually live;
- `__resetPanelBaseCache` plus a pinned `workspace.base` is the pattern to copy, since resetting alone falls back to the real environment rather than to neutral — which is what my second attempt proved.

That is a real answer to "why does #1208's approach keep needing repeating": it is not that per-file is wrong, it is that per-file was applied to *assertions* rather than to *setup*. Copying this file's `beforeEach`/`afterEach` pair into the five is the fix, and it is a different act from patching a third assertion.

Three attempts, two reverted, and the third only became visible after reading the working file instead of theorising about the broken ones. Recording the dead ends because each looked correct going in.

artokun:
### Another session found the real cause — a PRODUCTION race, not test isolation

This branch already carried two commits I did not write this session:

- `174220a` — *"a cache CLEAR must beat a probe that started before it"*
- `c9a8e28` — *"the HELLO clear was the same hole, in production"*

That is the answer, and it is **production-side**, which resolves the thing my own attempts kept failing to explain: a full run still failed flake #1 with my test-only reset applied, and that test (`ui-bridge.test.ts`) already reset the cache. A reset cannot fix a race where an in-flight probe writes its result *after* a clear — the clear happens, the stale probe lands anyway, and the next read sees the pre-clear value. Resetting harder does nothing; the write has to lose.

So my "this is test isolation, not a user-facing staleness bug" conclusion was **wrong**, and wrong in the direction that matters. I reached it by reading `cachedResolution()`'s guards — target key, generation, TTL — and confirming all three fire in production. They do. But I checked the *validity* conditions and never asked whether a write could land after an invalidation. The guards being correct is not the same as the cache being correct, and I closed that question on the strength of the first.

Rebased my commit on top rather than over: the five suites that asserted remedy wording with no reset at all (panel-installer had no `beforeEach` whatsoever) still should have one, matching `panel-recovery-cluster.test.ts`, which does it in both hooks and is not one of the flaky files. That is correct isolation that was missing — it is just not the fix, and the commit message says so.

**Suite on the combined branch: 406 files, 8137 passed, 2 skipped.**

Three of my approaches are recorded above as dead ends. Worth keeping visible: each was a plausible reading of the symptom, and the actual cause was only findable by someone asking a question I had ruled out.

PR #1252, opened 2026-08-09
fix(fence): investigate #1077 Finding 2 — permanent refusal without a relay backend

artokun:
Finding 2 of #1077, the half that did NOT ship in 0.50.67.

The reporter saw the identical permanent fence refusal on a session with **no relay backend configured**, which routes through the primary listener and should capture a real `Origin`. Nothing in the Finding 1 fix explains that.

Their lead: `turnOrigins` and the `mode:"current"` repin guard, whose own comment says it deliberately "refuses to displace a pin that still reaches a live tab" — plausible if a dead prior tab is still considered reachable in some stale way. They had several full `no connected tab` drops and workflow-tab switches before the permanent state set in.

Draft while I investigate. If it turns out to be unreproducible I will say so rather than ship a speculative change.

### DIFF
--- a/src/__tests__/orchestrator/repin-refusal-reason.test.ts
+++ b/src/__tests__/orchestrator/repin-refusal-reason.test.ts
@@ -12,12 +12,17 @@
 // `rebound: false`. A healthy pin correctly left alone and a permanent ambiguity
 // were the same value.
 //
-// And one of those four repeats forever. `mode:"current"` picks the active tab
-// when it is eligible, or the sole eligible tab — otherwise nothing. So when the
-// ACTIVE tab belongs to another backend's conversation and this one has two or
-// more tabs, every retry refuses identically, and nothing in the reply hinted
-// that naming a workflow is the way out. That is reachable in ordinary multi-tab
-// use, which is the shape the report describes.
+// And one of those four cannot be escaped BY RETRYING. `mode:"current"` picks the
+// active tab when it is eligible, or the sole eligible tab — otherwise nothing.
+// So when the ACTIVE tab belongs to another backend's conversation and this one
+// has two or more tabs, every retry refuses on identical inputs, and nothing in
+// the reply hinted that naming a workflow is the way out.
+//
+// Deliberately NOT called permanent, which the first draft of this claimed: the
+// active tab moves on the next `user_message`, so a message sent from one of this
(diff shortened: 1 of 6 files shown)


artokun:
**Review note — codex produced no review, so this is my own adversarial pass, disclosed.** It was forwarded, exceeded the foreground window, moved to background, and its forwarder declined to poll for partial results. I checked the transcript directly: it contains the forwarded prompt but no findings. (I briefly misread the severity template echoed in that prompt as a finding — it was not.) Rather than block, here are its six checks run directly. Two found real problems, and the first was in my own commit message.

1. **Is the "repeats forever" claim true?** **No — I overstated it.** `setLastActiveTab` fires on every `user_message` (`ui-bridge.ts:2395`), so a message sent from one of this conversation's own tabs makes the active tab eligible and the repin succeeds. What is true is that *retrying the tool* never clears it — identical inputs every time — which is the loop the reporter was in. Corrected in the code comment, the test names and the test bodies.
2. **Does a reason leak anything?** Not new information — tab ids already appear in these messages. But it repeated **#934**: `slice(0, 8)` renders every `wf:workflows/…` id as the literal `wf:workf`, which is precisely the bug #934 fixed in the rebind note, where a message naming two different workflow tabs read as naming the same one twice. Now uses the repo's `shortTabId`, pinned by a test asserting the workflow is identifiable.
3. **Truthiness at every consumer.** There is exactly one production consumer (`panel-tools.ts:5472`) and it narrows with `typeof outcome === "string"`, so no check sees the refusal object as a successful repin. `Boolean(repinned)` operates on the already-narrowed value.
4. **Behaviour parity.** Only the refusal payloads changed; the selection logic is untouched. Two tests pin that the recovery still fires on the active eligible tab and on a sole eligible tab.
5. **Were the updated assertions loosened?** They are strictly stronger: the old `toBeUndefined()` becomes "not a string **and** carries `repinned: false`", and each branch additionally asserts its own distinct reason.
6. **Delivery point.** The reason is read only when the outcome is an object, so it cannot attach to a successful repin, and it is appended inside the `mode:"current"` handler only.

Five mutations, all killed — including the one that mattered most: deleting the line that appends the reason to the tool result originally killed nothing, so the delivery is now pinned end-to-end through the real tool.

Issue #1255, opened 2026-08-09
fence-refusal says an identity is 'bound to a different origin' — but nothing compares origins


artokun:
Split out of #1077 as a **confirmed defect in its own right**. It is a plausible mechanism for that issue's Finding 2, but I cannot show the reporter's session mixed spellings, so this is filed on its own evidence rather than as a fix for theirs.

## Measured

```
localhost -> {"origin":"http://localhost:8188","uuid":"1111...5555"}
127.0.0.1 -> {"origin":"http://127.0.0.1:8188","uuid":"1111...5555"}
SAME ORIGIN? false
```

`workflowIdentityParts` (`src/orchestrator/session-store.ts:70-81`) normalises an origin with exactly three operations — `trim()`, strip trailing slashes, `toLowerCase()`. There is no host equivalence, so the two spellings of one local host produce two identities for one workflow.

## Why it matters

The workflow-instance fence validates a command's uuid **against the origin it was bound to**. If a panel is reached by one spelling and the identity was bound under the other, the identity gate refuses — and:

- it is **permanent**: nothing re-binds the stored origin;
- it **survives a hard refresh**: a reload re-sends the same Origin header;
- it needs **no relay backend**;
- it is **invisible to sibling reads** — `panel_list_workflows` and `panel_open_workflow` never consult origin-bound identity, so they keep reporting the correct uuid while mutations refuse. That asymmetry is what makes this class read as a contradiction rather than a bug.

The refusal surfaces as *"the workflow identity did not validate (uuid X against origin Y) — a malformed uuid, or one bound to a different origin"* (`fence-refusal.ts:77`), which is accurate but does not suggest that X and Y may be the same machine.

## Precedent

`bce83d7` — *"127.0.0.1 and localhost are the same host, not a target drift"* (#1246) — established this equivalence one layer over on 2026-08-09. Same question, different consumer. The fix should reuse that normalisation rather than invent a second one.

## Two cautions

1. **Normalise, do not special-case the pair.** `::1`, `0.0.0.0`, and a trailing-dot `localhost.` are the same question; a two-string check leaves the identical defect for the next spelling.
2. **Read the consumers before changing the gate.** `tabCommandWorkflowUuid` is keyed off the validated identity. If any origin string is persisted, normalising at the gate without migrating stored values converts a permanent refusal into a silent mismatch — which is worse, because the current failure at least announces itself.

Refs #1077, #1246.

artokun:
[...]

PR #1257, opened 2026-08-09
fix(identity): localhost and 127.0.0.1 must bind the same workflow identity (#1255)

artokun:
Draft — claiming #1255, split out of #1077 on its own measured evidence.

## Confirmed before claiming

```
localhost -> {"origin":"http://localhost:8188","uuid":"1111...5555"}
127.0.0.1 -> {"origin":"http://127.0.0.1:8188","uuid":"1111...5555"}
SAME ORIGIN? false
```

`workflowIdentityParts` (`session-store.ts:70-81`) normalises with `trim()`, trailing-slash strip, `toLowerCase()` — no host equivalence. So a fence bound under one spelling can never validate against the other: permanent, refresh-proof, no relay needed, and invisible to sibling reads that do not consult origin-bound identity.

## Both cautions from the issue, checked first

**1. Reuse, do not special-case.** `canonicalOrigin` already exists (`src/utils/origin.ts:51`), exported by #1246 (`bce83d7`, *"127.0.0.1 and localhost are the same host, not a target drift"*) with `sameOrigin` alongside it. Its own tests already cover `::1`, ports, schemes, paths and query strings — so the equivalence is defined in one place rather than in a second hand-rolled pair check.

**2. No persisted origins to migrate.** This was the one that could have made the fix worse than the bug. Checked: `orchestrator/index.ts:2851` and `:3558` store `identity.uuid` only — `identity.origin` is never written anywhere. So the origin is purely a validation input, and normalising it cannot turn a loud refusal into a silent mismatch against stored data.

## Plan

1. Red test first, driving the real `workflowIdentityParts` with both spellings.
2. Route its origin through `canonicalOrigin`, falling back to the current behaviour when the input will not parse — an unparseable origin must keep failing closed, not become empty and pass.
3. Verify the failure mode is preserved for genuinely different origins (a different host or port must still refuse), because a normaliser that is too eager turns a security-shaped gate into a rubber stamp. That is the discriminating negative here.

Refs #1077, #1246.

### DIFF
--- a/src/__tests__/orchestrator/hello-stamp-retention.test.ts
+++ b/src/__tests__/orchestrator/hello-stamp-retention.test.ts
@@ -68,8 +68,12 @@ describe("an untrusted hello RETAINS the tab's command stamp (#689)", () => {
     expect(workflowIdentityParts({ workflowUuid: uuid, origin: "" })).toBeUndefined();
     // A fully-formed hello IS trusted — otherwise the branch above would be the
     // only one ever taken and the test would prove nothing about untrusted input.
+    // #1255 — the returned origin is now CANONICAL, so every loopback spelling
+    // folds to one value and a fence bound under `127.0.0.1` validates against a
+    // panel reached on `localhost`. What this test is about — which identities
+    // are trusted vs untrusted — is unchanged; only the canonical form moved.
     expect(workflowIdentityParts({ workflowUuid: uuid, origin })).toEqual({
-      origin,
+      origin: "http://localhost:8188",
       uuid,
     });
   });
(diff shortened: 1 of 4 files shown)


artokun:
Closing unmerged. The premise is false: `identity.origin` is never read anywhere in the repo (`git log -S"identity.origin"` returns nothing), so nothing compares a bound origin to a later one and the cross-spelling refusal I 'confirmed' is not reachable. Mutation proof: replacing the returned origin with a constant fails only the 3 direct unit-test files — every fence, hello, bridge, panel-tool and routing suite passes. Details and re-scope on #1255.

PR #1258, opened 2026-08-09
fix(install): 'latest' is not a git ref — do not check it out

### LINKED ISSUE #1254: install_custom_node git install treats latest as literal checkout ref
### What happened
install_custom_node(action:'install', source:'git', version:'latest') cloned the repository but then tried `git checkout --detach --end-of-options latest`, which failed because latest is not a branch/tag. This leaves a partially installed custom node.

### Steps
Call install_custom_node with id=https://github.com/IRCSS/comfyUI-blender-wrapper.git, source='git', version='latest'.

### Exact error
Cloned repository but could not check out latest: fatal: git checkout: --detach does not take a path argument 'latest'.

### Environment
Windows 11; ComfyUI 0.29.0 local; RTX 5080 16GB; comfyui-mcp 0.50.71; panel 0.11.63.

### Fix
Applied locally: no. Required behavior: omit a checkout ref for 'latest' on git URLs, or resolve the remote default branch before checkout.

artokun:
Fixes #1254.

`install_custom_node(action:'install', source:'git', version:'latest')` clones, then runs `git checkout --detach --end-of-options latest` — and `latest` is not a branch or tag, so the checkout fails and the node is left partially installed.

`latest` is the tool's own word for "whatever the default branch is", which a fresh clone is already on.

Draft while I work it.

### DIFF
--- a/src/__tests__/services/git-latest-not-a-ref.test.ts
+++ b/src/__tests__/services/git-latest-not-a-ref.test.ts
@@ -1,0 +1,122 @@
+// #1254 — "latest" reached `git checkout` as though it were a ref.
+//
+//   install_custom_node(id: "https://github.com/…/comfyUI-blender-wrapper.git",
+//                       source: "git", version: "latest")
+//
+//   → git checkout --detach --end-of-options latest
+//   → fatal: git checkout: --detach does not take a path argument 'latest'
+//
+// `version` DEFAULTS to "latest", so this is the plain no-options install of any
+// git URL, not an exotic argument. And the cost is not just a failed command:
+// the checkout failure discards the clone as a husk, so the node is left
+// partially installed after a clone that had already succeeded.
+//
+// "latest" is this tool's own word for "whatever is newest". A fresh clone is
+// already on the default branch, which is exactly that — so the fix is to check
+// out nothing, which also restores the shallow clone (full history is fetched
+// only because a concrete ref might need it).
+//
+// The registry path has always treated the word as a sentinel (it rewrites an
+// absent or "latest" version to "nightly"), so the two now share one test for
+// what the word means rather than each having their own idea.
+
+import { describe, expect, it } from "vitest";
+
+import {
+  gitRefForInstall,
+  isLatestSentinel,
+  normalizeGitUrlInstallArgs,
+} from "../../services/node-management.js";
+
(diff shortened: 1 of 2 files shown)


artokun:
**Review note — codex was denied shell execution again, so this is my own adversarial pass, disclosed.** Its four checks, run directly:

1. **Does skipping the checkout leave the clone on the right commit?** HOLDS. A `git clone` with no ref lands on the remote's default branch HEAD, which is what "latest" means for a git source. Nothing after the clone reads the ref back — no `git describe`, no `rev-parse`, no version recorded into a manifest from it — so there is no later step that needed the detached checkout to have happened. The only consumer of `gitRef` after this point is the checkout itself and the shallow-clone switch.

2. **Is the shallow-vs-full switch right for every combination?** HOLDS. `gitRef ? full : --depth 1`. A `latest` version now yields no ref, so it shallow-clones — correct, since there is no other commit to reach. An **explicit** `ref: "latest"` still yields a ref, so it full-clones and then checks that tag out, which is exactly what a repo carrying a `latest` tag needs.

3. **Does honouring an explicit `ref: "latest"` create a new failure mode?** DOES NOT HOLD — it is the *old* behaviour, now reachable only when someone actually asks for it. If the repo has no such tag the checkout fails naming the ref they requested, which is a correct answer to a specific request rather than a default that cannot work.

4. **Any other unguarded version-to-git-ref site?** None — and one of them is the precedent for this fix. `install-comfyui.ts:325` already does the same thing for ComfyUI itself:

   ```ts
   if (tag) { record("checkout_version", "git", [..., "checkout", "--end-of-options", tag]); resolvedVersion = tag; }
   else { resolvedVersion = "latest"; }
   ```

   The ComfyUI installer has always treated "latest" as "check out nothing". The custom-node path was the one that did not, so this makes them agree rather than inventing a rule. `trainer-bootstrap.ts:116` checks out a compile-time constant, not a user version, and `node-management.ts:1713` is the checkout itself, which now only ever receives a guarded ref.

Five mutations: four killed. The fifth showed a `!== null` guard of mine was already dead (`??` collapses `parseGitUrl`'s null), so I removed it rather than leave a condition that reads as load-bearing and never fires.

Full suite: 8170 passing.

PR #1261, opened 2026-08-09
fix(restart): do not report launch arguments UNCHANGED when the listener was never confirmed

artokun:
Addresses the recurrence on #848 (comfyui-mcp 0.50.71, ComfyUI Desktop 0.31.1).

`restart_comfyui(action:"restart")` reported the launch arguments **UNCHANGED** while `listener_ownership` was `"unconfirmed"`. The listener on 8188 was in fact running a minimal `main.py --port 8188 --enable-manager --enable-cors-header`, missing the Desktop `--extra-model-paths-config`, `--input-directory` and `--output-directory`.

"Unchanged" is a claim about a comparison. If the process was never confirmed, that comparison was not made — this is #796's class, and the consequences were concrete: model inventory fell 14→8, inputs resolved against the stock directory producing false missing-media findings, and repeated restarts left ~15 `main.py` processes at ~10 GB RSS.

Draft while I investigate.


artokun:
Closing unmerged. I could not establish a defect: the two mechanisms I expected both fail on inspection, and the analysis is now on #848 rather than in a speculative change to a destructive path. Shipping a guess here could cost someone their running server.

Issue #776, opened 2026-08-03
restart_comfyui drops Stability Matrix Git and FFmpeg environment, leaving server down


artokun:
### What happened / root cause
restart_comfyui stopped a healthy Stability Matrix-managed ComfyUI but failed to bring the API back within 60 seconds. The relaunch preserved argv but not the launcher environment. Reproducing the same venv + argv without Stability Matrix's PATH caused ComfyUI-Manager to abort with "Bad git executable"; after restoring PortableGit, FFMPEGA still reported FFmpeg/FFprobe absent. Stability Matrix supplies both from its Data/Assets folders rather than the system PATH.

### Steps to reproduce
1. Launch a Stability Matrix ComfyUI package with its bundled Python, PortableGit, and FFmpeg assets.
2. Confirm ComfyUI 0.30.0 is healthy on 127.0.0.1:8188.
3. Call restart_comfyui.
4. Observe stopped:true, started:false, ready:false after 60 probes.
5. Launch the captured venv and argv alone.
6. ComfyUI-Manager aborts with GitPython "Bad git executable".
7. Add C:/Stability Matrix/Data/PortableGit/cmd to PATH and GIT_PYTHON_GIT_EXECUTABLE; Manager loads, but FFMPEGA reports FFmpeg/FFprobe missing.
8. Also add C:/Stability Matrix/Data/Assets/ffmpeg/bin to PATH; ComfyUI becomes healthy and both missing-dependency errors disappear.

### Exact error
restart_comfyui: "ComfyUI process was launched but the API did not become ready after 59756ms (60/60 probes)."
ComfyUI log: "ImportError: Failed to initialize: Bad git executable."

### Fix
Applied locally: yes (runtime workaround only). Upstream-only code change: yes.
Working recovery launch:
- venv/Scripts/python.exe
- argv: main.py --preview-method auto --use-pytorch-cross-attention --enable-manager
- PATH prepended with Stability Matrix PortableGit/cmd and Assets/ffmpeg/bin
- GIT_PYTHON_GIT_EXECUTABLE set to PortableGit/cmd/git.exe
Verified: API reachable, Manager loaded via GitPython, FFMPEGA found FFmpeg, queue idle.

[...]

PR #1262, opened 2026-08-09
fix(start): a relaunch that exits 1 must say what the child printed

### LINKED ISSUE #1259: Stability Matrix restart stops ComfyUI but relaunch exits 1 without stderr
restart_comfyui action restart stopped a healthy local Stability Matrix ComfyUI, then its managed relaunch exited code 1. action start using the captured configuration failed identically. The result provides no child stdout/stderr, so the actual launch failure cannot be diagnosed and the user is left offline.

### Steps to reproduce
1. Local Stability Matrix ComfyUI at 127.0.0.1:8188.
2. call restart_comfyui action restart.
3. Server stops; managed relaunch exits code 1 after 60 failed probes.
4. call action start; same result.

### Exact error (scrubbed)
ComfyUI process was launched, but no readiness probe got a healthy response in 59626ms. The process this call launched EXITED (exit code 1).
Launch environment: Stability Matrix install detected; PortableGit injected; FFmpeg absent.

### Fix
Capture and return bounded child stdout/stderr and exact executable/argv/cwd for failed managed relaunches. Validate the captured Stability Matrix launch command before stopping the healthy server, or provide rollback/relaunch through the owning launcher.

### Environment
Windows 11 / ComfyUI 0.28.0 / RTX 3090 Ti 24GB / comfyui-mcp 0.50.71 / panel 0.11.51.

artokun:
Fixes #1259.

A Stability Matrix ComfyUI was stopped, its managed relaunch exited code 1, and the result carried no child stdout/stderr — so the user was left offline with nothing to diagnose from:

> ComfyUI process was launched, but no readiness probe got a healthy response in 59626ms. The process this call launched EXITED (exit code 1).

An exit code alone is the least actionable thing a failed launch can report, and the process printed the reason before dying.

Draft while I work it.

### DIFF
--- a/src/__tests__/services/launch-log-capture.test.ts
+++ b/src/__tests__/services/launch-log-capture.test.ts
@@ -1,0 +1,202 @@
+// #1259 — a relaunch that exits 1 said nothing about WHY.
+//
+// A Stability Matrix ComfyUI was stopped, its managed relaunch exited code 1,
+// and the whole report was:
+//
+//   ComfyUI process was launched, but no readiness probe got a healthy response
+//   in 59626ms. The process this call launched EXITED (exit code 1).
+//
+// The user was left offline with an exit code and nothing else — and the process
+// had already printed the reason, into `stdio: "ignore"`. An exit code alone is
+// the least actionable thing a failed launch can report.
+//
+// A FILE, not a pipe. The child is `detached` and expected to outlive us, so
+// piping means either draining forever or closing a descriptor the server may
+// still write to — and a closed pipe is an EPIPE that would kill a server which
+// was starting fine. A file descriptor stays valid whatever happens to this
+// process, which is what `"ignore"` was buying and what this had to keep.
+
+import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const HF = "hf_aBcD3fGh1jKlMnOpQrStUvWxYz";
+
+vi.mock("../../config.js", async () => {
+  const actual = await vi.importActual<typeof import("../../config.js")>("../../config.js");
+  return {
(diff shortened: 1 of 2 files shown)


artokun:
**Note on the one local test failure:** `download-manager-routing.test.ts > #420` fails on my machine, and it fails identically on a clean `origin/main` with my changes stashed — CI reports success for that same SHA. It is pre-existing and unrelated to this PR; filed with the evidence as #1263 rather than folded in here.

**Review note — codex is being skipped on this one.** It has been denied shell execution on four of the last six attempts, and the one time it ran to completion it produced a genuinely useful NO-SHIP. Rather than pretend, here is my own pass, disclosed:

1. **Why a file and not a pipe.** The child is `detached` and outlives this process. Piping means draining forever or closing a descriptor the server may still write to, and a closed pipe is an EPIPE that kills a server which was starting fine. A file descriptor survives whatever happens here — which is what `"ignore"` was buying and what this had to preserve.
2. **Does it change the child's behaviour?** Only where its output goes. Same `detached`, same `env`, same `cwd`, same `shell:false`, same `windowsHide`. The fd is inherited at spawn and this process never holds the stream.
3. **Can the logging break a launch?** No path propagates: `openLaunchLog` returns undefined on any throw and the spawn falls back to `"ignore"`; the read, the prune and the header write each swallow their own errors. That is deliberate — #776's rule is that leaving the server down is the one outcome worse than a degraded launch.
4. **Disk growth.** Written on every relaunch and nothing else deletes them, so the newest 10 are kept. Pruning failures are non-fatal.
5. **Secrets.** The tail and the header both go through `scrubLogLines`, the same redactor as `get_system_stats action:"logs"`. Pinned by a test with a configured token.
6. **Does the failure message actually carry it?** That was the one thing every behavioural test missed — deleting the delivery line killed nothing until the wiring was pinned at the source. Both connections (spawn → log, log → message) are now asserted.

Six mutations, five killed. The sixth removed a decorative header title; the exe/args/cwd it carries are pinned separately.

PR #1269, opened 2026-08-09
sweep(#796): 'no saved sessions' must not also mean 'could not read them'

artokun:
Second #796 sweep pass. Fixes `listSessions`, and **corrects the candidate I named when the last sweep shipped**.

**`isManagerAvailable` does not hold.** I flagged it as conflating "Manager is absent" with "Manager errored", with a silent filesystem fallback as the consequence. Reading it properly: it is only consulted when there is no local path, and its false branch returns a controller that refuses with a message already covering both — *"ensure ComfyUI-Manager is installed and reachable"*. No silent fallback, no wrong action.

**`listSessions` does.** Every `readdir` failure returned `[]`. ENOENT is a real answer; EACCES/EPERM/EIO are not. It is a read, so nothing is destroyed — the severity is in what it says: *"you have no saved sessions"* is the one wrong answer here that reads as data loss.

It throws rather than returning a flag because the caller already wraps it in a `.catch` that pushes `{ sessions: [], error }` to the panel — plumbing no read failure could reach, because this swallowed them all.

### DIFF
--- a/src/__tests__/orchestrator/history-unreadable.test.ts
+++ b/src/__tests__/orchestrator/history-unreadable.test.ts
@@ -1,0 +1,123 @@
+// #796 sweep — "you have no saved sessions" was also being said when the
+// directory could not be READ.
+//
+// `listSessions` caught every `readdir` failure and returned `[]`. ENOENT is a
+// real answer — a first run has no directory yet, and an empty list is the truth.
+// EACCES, EPERM, EIO and a network path that went away are not.
+//
+// This one is worth fixing for what it says rather than what it does: it is a
+// read, nothing is destroyed. But "you have no saved sessions" is the single
+// wrong answer here that reads as DATA LOSS. Someone whose history is briefly
+// unreadable is told their conversations are gone, and goes looking for them.
+//
+// It THROWS rather than returning a flag because the one caller in
+// orchestrator/index.ts already wraps this in a `.catch` that pushes
+// `{ sessions: [], error }` to the panel — plumbing that could never fire for a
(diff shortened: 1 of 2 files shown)


artokun:
**Review note — my own adversarial pass, disclosed.** Codex has been denied shell execution on five of the last seven attempts; I ran the questions I would have asked it.

1. **Is throwing safe for every caller?** There is exactly one production caller (`orchestrator/index.ts:4785`) and it is already `void listSessions(...).then(...).catch(...)`, pushing `{ type: "history_list", sessions: [], error }` to the panel. So the throw lands in a handler that existed and was unreachable for this case. No other call site, and nothing awaits it in a path that would surface an unhandled rejection.

2. **Is the errno split right in the direction that matters?** Only ENOENT and ENOTDIR return `[]`. Getting that wrong the other way — throwing on a missing directory — would turn every fresh install into an error, so it has its own test and a mutation that removes the guard fails.

3. **Does the panel handle the error field?** It is the same envelope the existing catch already pushed, so this does not introduce a shape the client has not seen. If the client ignores `error` today it shows an empty list as before — no worse than now, and the orchestrator log gains the reason either way.

4. **Did content-level best-effort survive?** Yes, and it is pinned. An unreadable or corrupt transcript is still skipped rather than failing the list. Mutation showed my first fixture did not actually reach the per-file catch — malformed JSON is caught by an inner parse guard — so it now uses a directory named `*.jsonl`, which makes `readFile` throw EISDIR and does reach it.

Five mutations, all killed. Full suite: 8207 passing.

Issue #887, opened 2026-08-03
Panel sync no-ops on incoherent Manager queue counts while reinstall succeeds


artokun:
### What happened / root cause
A live-canvas panel_add_node call failed because the loaded sidebar was panel 0.11.3 while comfyui-mcp 0.49.3 requires panel 0.11.35+ for per-command workflow targeting. install_panel(action="sync") then failed because ComfyUI-Manager returned an incoherent drained-queue signature and left disk unchanged.

### Steps to reproduce
1. Run comfyui-mcp 0.49.3 with panel 0.11.3.
2. Call panel_add_node on a live workflow.
3. Run install_panel(action="sync").

### Exact errors
panel_add_node: graph_add_node could not be dispatched; panel version unknown; requires panel 0.11.35+.
install_panel sync: Panel update did NOT apply; Manager counts total=0, done=14, pending=0, in_progress=0, so git fallback did not fire.

### Fix
Applied locally: yes. Upstream-only: no.
install_panel(action="reinstall") successfully advanced the pack from 0.11.3 to 0.11.38. A status re-read from disk confirmed 0.11.38 and sync.decision="up-to-date".

A useful improvement would be for sync to offer or automatically use the same verified reinstall fallback when Manager reports this incoherent no-op signature and no clean git fallback is available.

### Environment
Windows 11; ComfyUI 0.30.1 LOCAL; RTX PRO 4000 Blackwell 24 GB; comfyui-mcp 0.49.3; panel 0.11.3 at failure, 0.11.38 after verified reinstall.

artokun:
Fixed on main by the #824 series + PR #840 (8f434b1) + #878 (1292426): an incoherent-but-provably-idle drained queue (the reported total=0/done=14 signature) now routes to a bracketed git ff-only verification instead of a flat refusal (src/services/panel-installer.ts:5597-5700, isProvenQuiescentQueue:2131); performPanelSync routes through it. Named regression tests in panel-installer.test.ts. Unreleased.

PR #1272, opened 2026-08-09
fix(open): tell the caller when the post-open read contradicts the open (panel#887)

artokun:
Fixes the orchestrator half of artokun/comfyui-mcp-panel#887.

`panel_open_workflow` reported the requested workflow was active while `panel_list_workflows`, called immediately afterwards, named a different one. The sharp end, in the reporter's words: a Save-As on that state writes the wrong canvas.

## The contradiction was already detected — and swallowed

`refreshOpenWorkflowUuid` is the **only** observation in the entire open path taken after a real round trip (`await ctx.call({cmd:"workflow_list"})`). That makes it the only thing that can see the active pointer having settled somewhere else. It saw the contradiction, correctly declined to adopt the target's uuid, and then `return`ed — leaving the tool result a plain success naming the path the caller asked for.

Detected, handled, not disclosed. **Suppressing the refresh protected the fence; it never protected the caller**, who was told the open landed.

It now fails the open and names what is actually active. The session's fence is untouched, so the workflow that *is* active keeps its own protection.

## Why the warning needed its own predicate

The obvious move — warn wherever `activeMatchesOpenRefreshTarget` returns false — is wrong, and dangerously so.

That predicate is a **uuid-adoption gate**. For adoption, "a different workflow is active" and "I cannot tell what is active" both correctly mean *do not adopt*, so it collapses them into one `false`. Warning on that same `false` is a different claim on the same evidence. It returns false for any record without a canonical **saved** identity — which includes an ordinary unsaved tab, since `tmp:` keys are deliberately rejected as saved identities. Every open made while "Unsaved Workflow" is active would have raised a wrong-canvas alarm.

`readOpenActiveAgainstTarget` returns `different` only on a **positive identification** of what is active. An unreadable record stays `indeterminate` and says nothing: unproven degrades to unknown, never to the negative.

## The existing suite caught a real false alarm mid-change

Three #716 P1 tests failed on the first cut. A record whose **path** is the workflow we opened but whose `routing_key` is absent, replayed, or malformed fails the strict gate while being the same workflow — my predicate called that drift. A looser positive match now short-circuits it to `indeterminate`, and the same-basename-other-directory case is pinned separately to prove the loosening did not swallow real drift.

That is the suite doing its job, and I would rather report it than quietly absorb it.

## One pre-existing assertion is deliberately inverted

`"does not refresh a successful open from a different same-basename active workflow (#716 P1)"` asserted `res.isError` **falsy**. Its #716 subject — the uuid must not be adopted from a same-basename workflow in another directory — is unchanged and still asserted. The success claim alongside it was the defect: the caller asked for `workflows/a/foo.json`, the re-read says `workflows/b/foo.json` is active, and answering "opened `workflows/a/foo.json`" is exactly the report that enables the wrong write. Changed in place, with the reasoning next to it.

## Verification

Driven through the **real `panel_open_workflow` handler** end to end via `buildPanelToolDefs()` and a mock bridge — that is what makes these *reachability* evidence rather than helper coverage.

That distinction is the whole lesson from the first attempt at this issue: [panel PR #896](https://github.com/artokun/comfyui-mcp-panel/pull/896) was withdrawn because its new branches were **unreachable in production** — it re-read a value with no `await` between the two reads — and a green suite with five killed mutations could not say so. Mutations prove the tests match the code; they never prove the code is reachable.

Four mutations, each confirmed applied before its result was trusted:

| mutation | |
|---|---|
| restore the swallow (detect drift, return null) | killed |
| collapse `indeterminate` into `different` (alarm on every unsaved-active open) | killed |
| call site ignores the notice | killed |
| treat an unreadable active record as drift | killed |

8215/8215 tests, typecheck clean, no control bytes in changed files.

## Scope

The panel-side half of #887 needs nothing: the panel's message is accurate at the instant it is composed, and no panel-side re-read can be later than itself. This is the only place in the stack where the observation is genuinely later.

### DIFF
--- a/src/__tests__/orchestrator/open-workflow-drift-surfaced.test.ts
+++ b/src/__tests__/orchestrator/open-workflow-drift-surfaced.test.ts
@@ -64,7 +64,17 @@ const ACTIVE_RECORDS: Record<ActiveShape, unknown> = {
   },
   // An unsaved tab has no saved path — `tmp:` keys are deliberately rejected as
   // saved identities, which is exactly why the adoption gate cannot be reused.
-  unsavedTab: { filename: "Unsaved Workflow", key: "tmp:2522828d", routing_key: "tmp:2522828d" },
+  // THE REAL SHAPE: the panel nulls `path` AND `filename` and puts the human label
+  // on `title`. An earlier fixture here put the label on `filename`, which passed
+  // for the wrong reason — it exercised a payload the panel never sends, and would
+  // have hidden that the warning named a bare `tmp:` handle at the user.
+  unsavedTab: {
+    path: null,
+    filename: null,
+    title: "Unsaved Workflow",
+    key: "tmp:2522828d",
+    routing_key: "tmp:2522828d",
+  },
   empty: {},
   missing: null,
 };
@@ -206,9 +216,13 @@ describe("#887 readOpenActiveAgainstTarget separates cannot-tell from proven-dif
 });
 
 describe("#887 describeActiveRecord names something a human can act on", () => {
-  it("prefers the filename, then the path, then the routing key", () => {
-    expect(describeActiveRecord({ filename: "b.json", path: "workflows/b.json" })).toBe("b.json");
-    expect(describeActiveRecord({ path: "workflows/b.json" })).toBe("workflows/b.json");
+  it("prefers the PATH, then title, then filename, then the routing key", () => {
+    // Path first on purpose: the case this guard exists for is a same-basename
+    // workflow in another directory, and a bare filename makes the warning refute
+    // itself ("workflows/a/foo.json was opened, but foo.json is active now").
+    expect(describeActiveRecord({ filename: "b.json", path: "workflows/b.json" })).toBe("workflows/b.json");
+    expect(describeActiveRecord({ title: "Unsaved Workflow", key: "tmp:abc" })).toBe("Unsaved Workflow");
+    expect(describeActiveRecord({ filename: "b.json" })).toBe("b.json");
     expect(describeActiveRecord({ key: "tmp:abc" })).toBe("tmp:abc");
   });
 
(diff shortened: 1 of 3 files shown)


artokun:
**Adversarial review found the first cut did not fix the filed bug. Four fixes pushed (`ef9a0c6`).**

## CRITICAL 1 — the reporter's own path was never reached

The panel throws for every open verdict short of `PROVEN`, so #887's scenario arrives as an **error** result — the ticket's own step 3 says so: *"Tool errors saying the requested workflow is active."* The corroborating read was gated on `!res.isError`, so the exact case in the report went unexamined.

I have no excuse for this one: I wrote *"the call is gated on `if (!res.isError)`, so the reporter's path skips the check entirely — reach it on the erroring path too"* into my own notes when I scoped this work, then implemented the other half and opened a PR claiming it fixed the issue. The review caught what my own plan already said.

Now observed there too, with the drift **appended** to the original failure rather than replacing it — the panel's message is still the primary information; this adds the fact that contradicts it.

Three things keep that safe:
- **Observation is split from adoption.** Adoption is gated on the open reply corroborating the requested identity, because a reply that cannot prove what it opened must never authorize a fence refresh. "What is active right now" needs none of that — and on this path the reply is an error carrying no JSON at all, which is precisely why the corroboration gate kept it unexamined. Nothing is ever stamped from the erroring path.
- **Scoped to the class where the load actually RAN.** A genuine acked failure (missing file) means nothing was opened, so another workflow being active is expected, not drift. This repo already pins that a genuine error must not make a `workflow_list` round trip at all — two existing tests caught my blanket version, which is how the boundary was found. Gated on the panel's own contract wording, `workflow_open RAN`.

## CRITICAL 2 — spelling differences read as drift, failing healthy opens

The loose safety net was raw string equality, so `./workflows/a.json`, `workflows\a.json` and `Workflows/A.json` each read as a *different file* from `workflows/a.json` whenever `routing_key` did not corroborate — failing a perfectly good open, on Windows, which is the reporter's platform.

Both sides now canonicalize and case folds. Deliberately generous: this check only ever **suppresses** a warning, so its failure mode is silence rather than a false alarm. Strict where a warning is raised, generous where one is withheld.

## MAJOR 3 — the warning named a basename, so it refuted itself

For the same-basename-other-directory case the message read *"workflows/a/foo.json was opened, but **foo.json** is the ACTIVE workflow now"* — which reads as the same file. The directory is the distinguishing information. Label now prefers `path`, then `title` (where the panel puts an unsaved tab's name), then `filename`.

## MAJOR 4 — `active_confirmed:false` is not evidence

This file already treats that flag as the panel saying the value is untrustworthy and refuses to adopt through it. It cannot be better evidence for **failing** an open than for stamping one. Now returns `indeterminate`.

## Also: a fixture that was green for the wrong reason

The unsaved-tab record put its label on `filename`. The panel nulls `path` **and** `filename` and labels via `title`, so that test exercised a payload production never sends — and would have hidden that the warning printed a bare `tmp:` handle at the user. Corrected to the real shape.

## Verification

Five mutations, each confirmed applied **and confirmed effective**. One initially reported "survived": injecting `refreshWorkflowUuid(ctx, parseToolResultJson(res))` on the drift path. That is a **no-op** — `parseToolResultJson` returns null on an error result — so it was an invalid mutation, not a coverage gap. Re-run with a stamp that really stamps, it is killed. Worth stating because "survived" and "the mutation did nothing" look identical in the output.

8226 passed / 2 skipped across 411 files, typecheck clean.

## Known limitation, stated rather than hidden

When a tab's identity is not yet established the panel can send `key` and `routing_key` both null. That reads `indeterminate` and stays silent — correct by the rule here (you cannot identify what you cannot read), but it does mean drift into an unidentified tab is not caught. Not fixable from this side without a panel-side identity signal, which is the same gap tracked in panel#833 and panel#817.

PR #1274, opened 2026-08-09
fix(download): the restart carry-over that did not run

artokun:
Recurrence on #1148, reported against 0.50.76.

A 5 GB download was lost across a ComfyUI/orchestrator restart: `action:"status"` found no matching id and no file landed. The reporter notes the reply **said the #1148 restart carry-over did not run** — so the code named its own failed mechanism.

Draft while I find why it did not run.


artokun:
Closing unmerged. The root cause is found and now on #1148 — the persisted store is entirely disabled for a plain stdio server, so nothing could have carried across. The obvious fix (a stable records dir) works but also switches on the adoption paths that gating controls, and a stale record then gets adopted as the live writer for a new download — a worse, silent failure. That needs a liveness check before adoption and its own review, not a fold into this.

PR #1278, opened 2026-08-09
fix(download): persist job records without a panel, so a restart can be survived

artokun:
#1148's root cause, now that its blocker shipped in 0.50.78.

The persisted job store is gated on a channel directory, set only by `COMFYUI_MCP_PROGRESS_DIR` or by the panel orchestrator late-binding its own. A plain stdio MCP server has neither, so `persistedRecordsEnabled()` is false and **no record is ever written** — every cross-restart mechanism built for #1148 is inert there, while `action:"status"` keeps promising an interrupted download stays resolvable by id.

I stopped short of this last cycle because enabling the store also switches on cross-session adoption, and a stale record could be adopted as the live writer for a new download. **#1275 fixed that**: adoption now requires the writer not be PROVEN gone. The specific hazard is closed, so this is a change worth making rather than a trade.

Draft while I work it.

### DIFF
--- a/src/__tests__/services/download-jobs.test.ts
+++ b/src/__tests__/services/download-jobs.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, vi, beforeEach } from "vitest";
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 
 const hoisted = vi.hoisted(() => ({
   resolvers: [] as Array<{ resolve: (p: string) => void; reject: (e: Error) => void; url: string }>,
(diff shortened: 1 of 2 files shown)


artokun:
**Review note — my own adversarial pass, disclosed** (codex has been denied shell execution on five of the last eight attempts). The questions I would have asked it:

1. **Does enabling the store switch anything else on?** It did, and that was the one real defect in my first version. `channelDir()` also drives the CONTROL channel — the MCP child's target-change requests, which an orchestrator reads. Falling back there switched on a channel with nobody at the other end for a plain stdio server. An existing test (*"is inactive with no progress dir (non-panel mode)"*) caught it. Only the record store falls back now; a mutation that re-extends it to the control channel fails.

2. **Can the records directory be reaped?** No, and that is why it is not under tmpdir. The orchestrator nonces its progress dir per start and sweeps earlier ones there — the exact deletion #1148's carry-over exists to survive. Putting records where a later orchestrator start could remove them would have reintroduced the bug from the other side.

3. **Unbounded growth?** No. In-flight records self-reap after `PERSISTED_JOB_TTL_MS` (6h) on read, and terminal ones are removed when the job settles.

4. **Is a heartbeat for every plain download proportionate?** It is a small write every 15s, `unref`'d, and only while a transfer is in flight. That is the cost of the record surviving; without it there is nothing to survive. It still installs no interval when there is genuinely nowhere to persist — retrying a no-op forever is the failure that rule exists to prevent, and it now has a test that provokes an unusable dir rather than relying on the old default.

5. **Does a failure to create the directory break a download?** No path propagates: `stableRecordsDir()` returns "" on any throw and persistence degrades to off — the previous behaviour. A mutation making it throw fails.

6. **Is the adoption hazard actually closed?** Yes, by #1275 in 0.50.78: adoption now requires the writer not be PROVEN gone. Without it this change would trade a silent "download vanished" for a worse silent "download adopted a corpse".

Full suite: 8253 passing across 414 files.

PR #1279, opened 2026-08-09
fix(routing): an explicit workflow pin must clear an AMBIGUOUS turn pin (panel#888)

### LINKED ISSUE #888: Pinned workflow target ignored when resumed turn contains multiple workflow origins

artokun:
Fixes artokun/comfyui-mcp-panel#888.

After a reconnect with queued events from several tabs, `panel_graph_outline` refuses as ambiguous — correct, that is #884's fence. The refusal names the way out:

> Target a workflow explicitly (**panel_set_workflow_target** / panel_open_workflow) or wait for the next single-origin message.

The reporter did exactly that: `panel_set_workflow_target` with `mode:"pinned"` and the exact active path. It reported success. The next `panel_graph_outline` returned the identical ambiguity error.

## Localized

The advertised recovery is wired to **one mode only**.

- `ui-bridge.ts:3359` — `resolveTarget` asks `scopeTargetResolver` for the **turn-origin** pin. `null` means the batch was mixed-origin, and it throws the ambiguity error immediately.
- The escape hatch exists — `setScopeRepinHandler`, documented as *"EXPLICIT recovery from a DEAD or AMBIGUOUS pin, and from those only"*.
- But `panel-tools.ts:9160` / `:9276` gate reaching it on `mode === "current" && ctx.rebindToActiveTab`. **`mode:"pinned"` never calls it**, so the ambiguous turn pin survives a successful pin, and every subsequent scope-addressed command keeps hitting the same refusal.

So the tool reports success truthfully — it *did* set the session's workflow target — while the thing actually blocking the next call is a different piece of state it never touched.

## Why routing `pinned` through the same handler is safe

The handler already carries the safety property, and it is not weakened here: it **refuses to displace a pin that still reaches a live tab** (confirming gate 3, P0 — the recovery path was previously repinning healthy turns via `panel_reload`, the exact silent re-target #884 exists to prevent). An *ambiguous* pin is `null`; there is no live tab behind it to protect, which is precisely why the handler already treats ambiguous and dead alike.

`mode:"pinned"` with an explicit path is a stronger, more specific statement of intent than `mode:"current"`, so it is odd that the weaker one is the only one that recovers.

## Status

Draft — localization committed, implementation and tests to follow. The two things this has to prove are that a HEALTHY pin is still never displaced, and that the new path is genuinely reachable from the real `panel_set_workflow_target` handler rather than only from the helper.

### DIFF
--- a/src/__tests__/orchestrator/explicit-pin-clears-ambiguity.test.ts
+++ b/src/__tests__/orchestrator/explicit-pin-clears-ambiguity.test.ts
@@ -1,0 +1,205 @@
+// panel#888 — after a reconnect delivered queued events from several tabs,
+// `panel_graph_outline` refused as ambiguous (correct — #884's fence). The
+// refusal names the way out:
+//
+//   Target a workflow explicitly (panel_set_workflow_target / panel_open_workflow)
+//   or wait for the next single-origin message.
+//
+// The reporter did exactly that — `mode:"pinned"` with the exact active path. It
+// reported success. The next `panel_graph_outline` returned the identical error.
+//
+// THE ADVICE WAS WIRED TO ONE MODE. `resolveTarget` throws the ambiguity error the
+// moment `scopeTargetResolver` returns null. The escape hatch exists —
+// `setScopeRepinHandler`, documented "EXPLICIT recovery from a DEAD or AMBIGUOUS
+// pin" — but only `mode:"current"` ever reached it. `mode:"pinned"` wrote the
+// workflow-target store and nothing else, so the tool reported success truthfully
+// while the state actually blocking every subsequent command went untouched.
+//
+// The repin handler's OWN refusal already told the agent to do what did not work:
+// "Name the workflow instead — panel_set_workflow_target with a path, or
+// panel_open_workflow — and the pin follows it." It did not follow.
+//
+// WHY ROUTING `pinned` THROUGH THE SAME HANDLER IS SAFE. The P0 gate is shared
+// rather than re-implemented: a pin that still reaches a live tab of this
+// conversation is never displaced, however explicit the request. Only a dead or
+// ambiguous pin can be recovered. A NAMED tab that is not eligible REFUSES —
+// silently redirecting an explicit request to another workflow is the exact hazard
+// #884 exists to prevent, and would be a worse bug than the one being fixed.
+import { describe, expect, it, vi } from "vitest";
+
+import { makeScopeRepinHandler, type ScopeRepinBridge } from "../../orchestrator/turn-origins.js";
+
+const KEY = "orchestrator::codex";
+const TAB_A = "wf:workflows/a.json";
+const TAB_B = "wf:workflows/b.json";
+
+/** `resolvedPinOf` is the real discriminator: string = a pin, null = ambiguous. */
+function harness(opts: {
+  pin: string | null | undefined;
+  tabs: string[];
+  reachable?: string[];
+  active?: string;
+  headless?: string[];
+  backendOf?: (t: string) => string;
+}) {
+  const repinTo = vi.fn();
+  const reachable = new Set(opts.reachable ?? opts.tabs);
+  const bridge: ScopeRepinBridge = {
+    canReach: (t) => reachable.has(t),
+    resolveActiveScopeTab: () => opts.active,
+    isHeadless: (t) => (opts.headless ?? []).includes(t),
+    tabs: () => opts.tabs.map((tab_id) => ({ tab_id })),
+  };
+  const handler = makeScopeRepinHandler({
+    bridge,
+    tracker: { resolvedPinOf: () => opts.pin, repinTo } as never,
+    scopeAgentKeyOf: () => KEY,
+    backendForTab: opts.backendOf ?? (() => "codex"),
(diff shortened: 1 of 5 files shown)


artokun:
**My first cut of this was INERT, and the test I wrote to prove otherwise couldn't. Adversarial review caught both; corrected in `dc25b7d`.**

## The defect

`canonicalRequestedSavedIdentity(pinPath)` yields `wf:<path>` — the saved-workflow **handle**. A real tab id is a bridge **route**, `wf:<tabRouteId>:<path>`. The panel's own `bridge-route.js` (#640) defines both and warns they are not interchangeable; the handle is path-only *precisely because* two browser tabs can show the same file, so it cannot address one of them.

Comparing a handle against routes never matches → the recovery refused every time → the reporter's wedge unchanged. I verified it against panel `origin/main` rather than taking the finding on trust.

Two further ways to miss, both real:
- top-level workflow → `pinPath` is ComfyUI's `key` (`"a.json"`), no `/`, so the derivation returns `null` and the branch is **skipped silently**;
- subfolder → produced `wf:sub/a.json`, missing the `workflows/` prefix.

Three different failure modes depending on input.

## Why my test didn't catch it — the part that matters

The WIRING tests `readFileSync`'d `panel-tools.ts` and regexed for the call. **A source-text grep cannot distinguish reachable code from dead code.** All three passed while the branch provably never fired. Worse, one asserted the defective line itself, so fixing the bug required deleting the test that claimed to protect it.

This is the **second no-op I shipped today from exactly this**, after writing the lesson down following the first. The rule I keep having to relearn: if a fix turns on *reachability*, only a test that drives the production path can show it.

Replaced by `pin-repin-uses-a-real-tab-id.test.ts` — drives the real `panel_set_workflow_target` handler through the real repin handler and asserts the pin **lands on a tab the bridge actually holds**. It failed three ways against the broken derivation before passing against this one. The old suite's fixtures were fiction too (`wf:<path>` as a tab id, a shape production never emits); now real routes.

## The fix

Pass the **path**, and let the handler match it against the routes the bridge holds (`tabRouteCarriesPath`), normalized the way every other saved-path comparison here normalizes. Generous on purpose — a miss only refuses a recovery — but a match is acted on **only when exactly one tab** has that workflow. Two tabs showing it refuse and name them, since silently taking the first is the same silent-wrong-tab defect in a new place.

The P0 gate is untouched and still shared: a pin that still reaches a live tab is never displaced, however explicit the request.

## Verification

Four mutations, each confirmed applied: restore the handle-vs-route comparison (the original no-op), remove the call site, remove the healthy-pin gate, take the first of several matching tabs — all killed. **The last one survived on the first run** and exposed a genuine coverage gap, now closed by the duplicate-tab test. Worth stating rather than quietly fixing: the mutation run earned its keep here.

416 files / 8264 passed, 2 skipped; typecheck clean; no control bytes. One unrelated timing test (`late-mutation-e2e`) flaked under parallelism and passes isolated and on re-run.

## Still to do before merge

Not marking this ready yet. The route-shape claim is verified against panel source and unit fixtures, **not against a live hello** — one look at a real `tab_id` from the running bridge would settle it, and given today's record I would rather check than assume.

## Worth fixing separately

`ui-bridge.ts:110` still documents `tabId` as *"commonly derived from a saved workflow path"*, and `:1942` as `wf:<hash>`. Both predate #640, and that stale comment is what licensed the wrong derivation. `shortTabId` in `session-scope.ts` makes the same assumption (`id.slice(3)` treated as a path).

artokun:
**Live-verified the route shape** — against the panel actually being served by the running ComfyUI, not repo source:

```
GET /extensions/comfyui-mcp-panel/js/lib/bridge-route.js   200

const ROUTE_PREFIX = "wf:";
const ROUTE_SEP    = ":";
savedWorkflowHandle(savedPath)        -> ROUTE_PREFIX + path              // wf:<path>
savedWorkflowRoute(tabRouteId, path)  -> ROUTE_PREFIX + tab + ROUTE_SEP + path   // wf:<tab>:<path>
```

So a real tab id carries the route separator and the handle does not — which is exactly why the first cut could never match, and why matching on the path suffix of the route is the right shape. That was the one claim I was not willing to merge on second-hand.

Marking ready for review.

PR #1284, opened 2026-08-09
sweep(#796): 'does not exist' must not be said about a read that failed

artokun:
Fifth #796 pass, using the filter the last four converged on: a catch is only interesting when it is reachable by failures with DIFFERENT remedies.

Examining the sites that report something \"is missing\" or \"does not exist\" after a failed read.

Draft while I check which hold up.


artokun:
Closing unmerged — the candidate did not hold up. `coverage.unanswered` already marks every category when the HTTP source fails, so a silent filesystem skip is not user-visible-silent, and my second change would have replaced a true sentence with a false one. Reasoning is on #796.

PR #1289, opened 2026-08-10
fix(restart): say 'could not determine' instead of asserting the tab did not reconnect (panel#654)

artokun:
Fixes the orchestrator half of artokun/comfyui-mcp-panel#654.

After a confirmed restart cycle the tool reported `server_ready:true` with `panel_tab_reconnected:false`, `graph_tools_ready:false`, `ready:false` — and the reporter had to refresh the browser by hand.

## The path

`panel_tab_reconnected` is `tabBack`, which on this path is `awaitPostRestartReachable(preRestartPanelIdentity, …)`. Its first substantive line:

```ts
if (before == null) return false;   // immediate — nothing is awaited
```

`before` is `bridge.tabConnectionIdentity(tabId)`, which returns `undefined` when

```ts
conn.sock.readyState !== WebSocket.OPEN || !conn.tabSessionId
```

or when `resolveTarget` throws. The identity is captured **before** the restart dispatch, so a socket that is momentarily not `OPEN` at that instant — most likely right after a Manager custom-node install, which is the reporter's sequence — produces their exact output **with the panel never having failed**.

It also explains why every panel-side measurement in that thread showed healthy recovery: the panel was reconnecting fine and the orchestrator never looked.

## What this changes, and what it does NOT

**The gate is not weakened.** `before == null → false` is deliberate — a real bridge missing the pre-dispatch identity must fail closed, and a weaker proof would be worse. `ready` and `graph_tools_ready` stay exactly as strict.

What changes is that the reply stops asserting something it never measured. `false` currently means two different things:

- the tab was watched for a newer hello and did not come back;
- **no baseline was captured, so nothing was watched.**

The second is *could not determine*, and this repo has fixed that conflation repeatedly (#796, #1077, and panel#887 today). A user reading `false` reasonably concludes the panel is dead and refreshes — which is the manual step this issue is about, and it may have been unnecessary every time.

## Already disproved, recorded so it is not re-run

That an older panel does not send `tabSessionId`. `TAB_ID_KEY` shipped in panel **0.11.4**; the reporter is on 0.11.39.

## Status

Draft. Implementation and behavioural tests to follow. The two things this must prove: the strictness of `ready`/`graph_tools_ready` is unchanged, and the new state is reachable from the real restart handler rather than only from a helper.

### DIFF
--- a/src/__tests__/orchestrator/panel-restart-legacy-fallback.test.ts
+++ b/src/__tests__/orchestrator/panel-restart-legacy-fallback.test.ts
@@ -339,3 +339,48 @@ describe("#425 remote: the refusal must name what will actually work", () => {
     expect(hoisted.restart).not.toHaveBeenCalled();
   });
 });
+
+// panel#654 — `panel_tab_reconnected:false` was emitted for a reconnect that was
+// never watched. `awaitPostRestartReachable` opens with `if (before == null)
+// return false`, and `before` is undefined whenever the panel advertised no tab
+// session id, the socket was not OPEN at capture time, or the tab did not
+// resolve. The reporter's output is reproducible with the panel never failing.
+//
+// These live here because this file already owns the scaffolding that reaches a
+// restart REPLY. A pure-function assertion cannot show the gate held: routing
+// `ready` through the classification left the whole suite green when measured.
+describe("#654 an unwatched reconnect is reported as unknown, and is still not ready", () => {
+  it("no pre-restart baseline ⇒ panel_tab_reconnected:'unknown', ready stays false", async () => {
+    const { ctx } = makeCtx(NO_ENDPOINT_REPLY);
+    ctx.panelConnectionIdentity = () => undefined; // panel advertised no tab session id
+    ctx.awaitPostRestartReachable = async () => false; // and so nothing is watched
+    ctx.tabCanMutateGraph = () => true; // isolate: capability is NOT the reason
+
+    const res = (await restartTool().handler({}, ctx)) as ToolResult;
+    const out = JSON.parse(res.content.find((c) => c.type === "text")!.text as string);
+
+    expect(out.server_ready).toBe(true);
+    expect(out.panel_tab_reconnected).toBe("unknown"); // not a bare false
+    // THE GATE. Unknown must never buy readiness — this is the assertion that
+    // catches `ready: graphToolsReady || tabReconnect === "unknown"`.
+    expect(out.ready).toBe(false);
+    expect(out.graph_tools_ready).toBe(false);
+    expect(String(out.note)).toMatch(/could NOT be determined/);
+    // …and it must not narrate a cause nobody observed.
+    expect(String(out.note)).toMatch(/WHY is not/);
+  });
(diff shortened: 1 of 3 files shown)


artokun:
**Adversarial review found three real problems. All addressed in `553c853`.**

## 1. The remedy text committed the sin this fix is about

The "unknown" note stated one cause as fact — *"its socket was not open at the moment the restart was dispatched"* — when `preRestartPanelIdentity == null` has **three** disjoint causes and the code determines none of them: socket not `OPEN` at capture, no advertised tab session id, or `resolveTarget` throwing.

The review's sharper point: the socket-timing cause is the *narrow transient*, while **no advertised tab session id is deterministic and permanent** — an older panel, or a browser-tab lease refused because a duplicate tab holds it. Those users hit this on **every** restart and were sent after a timing story forever, while the actual remedy (update the panel) went unmentioned.

Both notes now name the three possibilities *as* possibilities, and name the version remedy for the repeating case. Writing a replacement that asserted an unobserved cause, in a fix whose entire thesis is "stop asserting unobserved things", is the kind of irony worth recording rather than quietly correcting.

## 2. The "gate is untouched" test could not observe the gate

It asserted a property of a pure function with no causal link to the reply. The review **measured** the refactor it claimed to catch —

```ts
ready: graphToolsReady || tabReconnect === "unknown",
```

— leaving **1996/1996 green**.

Real behavioural coverage now lives in `panel-restart-legacy-fallback.test.ts`, which already owns the scaffolding that reaches a restart reply. That exact mutation is now **killed**, verified by applying it by line number and re-running.

## 3. My first behavioural attempt was vacuous

I rebuilt the harness from scratch, never got past the preflight refusal, and guarded the assertions with `if (text.includes("panel_tab_reconnected"))` — so it passed while asserting nothing, and survived the same gate refactor. Deleted rather than left looking like coverage.

## Verified and NOT changed — a false alarm I chased properly

Mid-review I suspected a wire-key mismatch: the panel's hello object carries `tabSessionId` (camelCase) and the bridge reads only `msg.tab_session_id`, with no normalizer. That would have made this the **always**-case rather than a conditional one, and a much bigger bug.

It is not. `buildHelloPayload` in the panel's `lib/session-rebind.js` does the conversion:

```js
if (typeof tabSessionId === "string" && tabSessionId.trim()) {
  frame.tab_session_id = tabSessionId.trim();
}
```

My grep found zero `tab_session_id` in the served `comfyui-mcp-panel.js` because the conversion lives in a different module. Checked before claiming it. Worth noting the conversion is **conditional** — which is precisely why the no-session-id branch is real rather than theoretical.

## Verification

Mutations, each confirmed applied: fold `unknown` back into `false`, hardcode `baselineCaptured`, route the gate through the classification — all killed.

Process note: **two earlier mutation runs reported results from patterns that never matched**, and one whole run was invalid because the baseline itself was failing. Both were caught by checking that the mutation applied and that the baseline passed, before reading any result. Neither would have been visible from the pass/fail output alone.

419 files / 8297 passed, 2 skipped; typecheck clean; no control bytes.

Issue #1292, opened 2026-08-10
Retry transient unconfirmed workflow binding after ComfyUI restart


artokun:
### What happened / root cause
Immediately after panel_restart_comfyui completed and the browser tab reconnected, the first panel_set_workflow_target({mode:"current"}) call failed because the panel's active workflow record was temporarily UNCONFIRMED. The same call succeeded five seconds later without any user action. This is a transient cross-process/tab reconciliation window exposed to callers instead of being absorbed by the recovery operation.

### Steps to reproduce
1. Start with a session bound to a live saved workflow.
2. Call panel_restart_comfyui and wait for the server restart to complete.
3. Immediately call panel_set_workflow_target({mode:"current"}).
4. Observe the uncorroborated-active failure.
5. Wait five seconds and repeat the identical call; it returns graph_binding:"bound" / graph_binding_status:"already_current".

### Exact error (scrubbed)
Error: panel_set_workflow_target({mode:"current"}) did NOT restore this session's graph binding.

APPLIED (do not repeat this part): the workflow target is now mode:"current".

NOT APPLIED: The panel answered, but its reply could not be corroborated, so nothing was adopted — the panel reported its active workflow as UNCONFIRMED. NOTHING was written to this session's fence; graph tools remain fenced to the previous workflow instance.

WHAT TO DO: call this again in a moment. A stale or mixed workflow list is usually transient — it settles once the panel finishes reconciling its tabs.

### Fix
applied locally: no
upstream-only: yes

The safety behavior in corroborateActiveForFence should remain fail-closed. The panel_set_workflow_target mode:"current" recovery path should absorb this known post-restart reconciliation window with a small bounded internal retry/backoff using a fresh workflow-list read, then return the existing error only if corroboration still fails. A local dist patch would require restarting the long-lived orchestrator that serves panel_* tools, which would disrupt the active repair session; the identical call already recovered after five seconds.

### Environment
Windows / ComfyUI 0.31.0 / ComfyUI frontend 1.48.7 / Python 3.13.12 / PyTorch 2.12.1+cu130 / NVIDIA RTX 4080 SUPER 16376 MB / comfyui-mcp 0.50.78 / panel 0.11.56.

Related (same root cause): #481, #1027

artokun:
Claimed — PR #1295 (draft).

PR #1293, opened 2026-08-10
fix(panel): absorb the transient UNCONFIRMED binding window after a restart (#1292)

artokun:
Draft — claiming #1292. Reported on **0.50.78** (current), `grep -rn "#1292" src/` finds no prior work, and no other branch has it.

## The report

Immediately after `panel_restart_comfyui` completes and the tab reconnects, the first `panel_set_workflow_target({mode:"current"})` fails because the panel's active-workflow record is momentarily **UNCONFIRMED**. The same call succeeds five seconds later with no user action.

## Why this is a real defect and not just timing

`panel_restart_comfyui` is a **recovery** operation. Its whole job is to leave the caller in a working state — so a reconciliation window it opens itself, and knows the duration of, should be absorbed by it rather than handed to the caller as a failure. The caller has no way to distinguish *"not yet"* from *"no"*, which is this backlog's dominant shape (#796): a transient could-not-determine surfaced as a determined-not.

It is also self-evidently recoverable: the identical call works seconds later. A refusal that a no-op retry fixes was never a refusal.

## Plan

1. **Find where UNCONFIRMED is produced and who observes it** — the consumer before the producer. Every issue I worked today turned on a claim true at one layer and false one below.
2. Establish whether the restart path already has a readiness gate that simply does not cover this record. If it does, extending it is the fix; if it does not, a bounded retry belongs in the tool, not the caller.
3. **Bound it and say so on expiry.** A retry that waits forever converts a visible failure into a hang, which is worse. On timeout the caller must be told the binding is still unconfirmed and that a manual retry is the next step — not silence.

Explicit non-goal: retrying *every* UNCONFIRMED. Outside the restart window, unconfirmed may be the honest and durable answer, and papering over it would hide a real binding failure. The narrow claim here is only that a restart the tool itself performed should not hand its own reconciliation window to the caller.

Refs #796.


artokun:
### Producer, consumer, and the readiness gate — located. The gate exists and does not cover this record.

**Producer:** `panel-tools.ts:3094-3096`, inside `corroborateActiveForFence`:

```ts
if (parsed.active_confirmed === false) {
  return { ok: false, why: "the panel reported its active workflow as UNCONFIRMED" };
}
```

The comment there (#514) is careful and correct: an explicit `false` is *"the panel telling us the value is untrustworthy — never adopt through that."* That reasoning holds in general. What it does not distinguish is **untrustworthy-forever** from **untrustworthy-for-the-next-second**, and after a restart it is the latter.

**Consumer:** `panel-tools.ts:3243-3245` turns it into `{status: "no_identity", kind: "uncorroborated", why}` — surfaced to the caller as a refusal with no indication it is transient.

**The gate:** the restart path already has a readiness budget (`panel-tools.ts:472`, *"generous, a real restart can take 15-60s+"*) with a gated and a legacy mode (`:1447`). So `panel_restart_comfyui` **does** wait — but it waits for **ComfyUI to be up**, not for the **panel's active-workflow record to reconcile**. Those are different events, and the second is the one the caller trips over.

That answers the question my plan turned on: this is an existing gate that does not cover this record, not a missing gate. Which makes the fix an extension rather than a new retry loop, and keeps it out of `corroborateActiveForFence` — where a retry would wrongly paper over a durable UNCONFIRMED outside the restart window.

### What I have not established

Whether the panel exposes anything that says *"the record is reconciling"* as distinct from *"the record is untrustworthy"*. If it does not, extending the gate means polling `active_confirmed` until it flips or the budget expires — workable, but it must **name the timeout honestly** rather than fall through to the same bare refusal, or the fix converts a fast wrong answer into a slow one.

Stopping here rather than implementing: this sits in the middle of a fence whose comments record several rounds of hard-won reasoning (#514, #1043, #1077, #1161), and I have twice today shipped changes into guards like this that a review then measured as inert or harmful. The next pass should start by checking the panel's reply shape for a reconciling signal.

artokun:
### The three-state vocabulary already exists — one consumer uses it, the other does not

`readOpenActiveAgainstTarget` (`panel-tools.ts:2767-2770`) takes the *same* input and maps it to a third state rather than a failure:

```ts
// #433/#3014 — `active_confirmed:false` is the panel telling us this value is
// untrustworthy. It is already refused as adoption evidence; it cannot be
// better evidence for failing an open than it is for stamping one.
if (activeConfirmed === false) return "indeterminate";
```

`corroborateActiveForFence` (`:3094`) takes that identical input and returns a hard `{ok: false, why: "…UNCONFIRMED"}`.

So the codebase already holds the distinction #1292 is asking for — *we cannot tell* as a state separate from *no* — and applies it on the open path while the fence path collapses it into a refusal. The #433 comment even argues the general principle: an unconfirmed value "cannot be better evidence for failing an open than it is for stamping one." The same sentence applies to the fence: it is not good enough to adopt through, and it is also not good enough to *refuse* on as though the answer were settled.

That reframes the fix, and makes it smaller than the retry loop I sketched:

1. `corroborateActiveForFence` should distinguish **indeterminate** from **refused**, matching its sibling.
2. The restart path's existing readiness budget (`:472`) can then wait out an *indeterminate* binding — which is a bounded wait for a state that reconciles — while a genuine refusal still fails immediately.
3. A durable indeterminate at budget expiry must say so in those terms, not fall through to the current bare UNCONFIRMED text.

Answering my own open question from the previous comment: **no, the panel does not distinguish reconciling from untrustworthy** — it sends one `active_confirmed:false`. But it does not need to. The orchestrator already treats that flag as indeterminate elsewhere, and the restart window is the context that makes indeterminate temporarily expected rather than final.

Not implementing on this pass. This is a fence with several rounds of reasoning in its comments (#433, #514, #1043, #1077, #1161), and I have shipped two changes into guards like it today that a review then measured as inert or harmful. The next pass has a concrete shape: mirror the sibling's three-state return, then let the restart gate wait on the new state only.

artokun:
Closing as superseded — #1292 shipped in **0.50.84** via #1295, and this branch holds only the claim commit (`git diff origin/main...` is empty), so nothing here is lost.

**The collision was my fault, and it's worth naming so it doesn't repeat.** You claimed this correctly and first — draft PR, 00:43. I checked the issue for comments, saw none, and claimed it about an hour later. But under this workflow a claim IS a draft PR, so the issue thread is exactly the wrong place to look; `gh pr list` is the register. I've since re-surveyed against open PRs.

Your framing was right and matches what shipped: a recovery operation that opens a reconciliation window should absorb it, and the caller has no way to tell *"not yet"* from *"no"* — the #796 shape.

Two things the implementation turned up that your plan wouldn't have predicted, in case you hit this class again:

1. **Not every uncorroborated reply can settle.** Retrying a panel that sends no `workflows` array at all just makes the identical error ~2.9s slower. The reasons are now split snapshot-vs-shape, and only snapshots are re-read. Codex flagged this as a NO-SHIP on my first version and it was right.
2. **The remedy had to change too.** Leaving "call this again in a moment" after four rechecks tells the caller to do the thing the tool just did, four times, and reads as though nothing was tried.

artokun:
### Implemented it, could not prove it, reverted. The finding stands; the patch does not.

Wrote the change: `corroborateActiveForFence` returning a third state (`indeterminate`) for `active_confirmed === false`, mirroring `readOpenActiveAgainstTarget:2769`, plus a bounded 6 × 500ms settle wait at the consumer that re-reads `workflow_list` **only** while indeterminate — so a genuine refusal still fails fast — and an expiry message saying the wait happened rather than falling through to the bare UNCONFIRMED text.

`tsc` clean. **But I could not get a test to exercise it**, and I am not shipping a change to this fence on inspection alone.

### What the harness attempts established (worth keeping — this is the expensive part)

The call site **is** the right one, verified rather than assumed: `panel-tools.ts:9376` calls `rebindWorkflowFence(ctx)` for `mode === "current"`, which is the reported flow. Getting a test there needs, in order:

1. `ctx.workflowTarget` — else `:9205` refuses with *"Workflow targeting is not available in this session"* before anything runs. The real `WorkflowTargetStore` works; a stub would test a shape production never uses.
2. `bridge.canReach`, `resolveFailure`, `tabServerOrigin` on the fake bridge.
3. **`ctx.rebindToActiveTab`** — the `:9374` branch is gated on it, so without it the code under test is skipped entirely and the harness passes while exercising nothing. That one is the trap: my first two attempts "ran" and asserted against a path that never executed.
4. A stub for (3) that does not recurse — mine produced `RangeError: Maximum call stack size exceeded`, which is where I stopped.

### Why I stopped rather than pushed through

Three harness layers deep, each revealing the next, with no budget to reason carefully about the fourth. Today I have shipped two changes into guards like this that a review then measured as **inert** or **harmful**, and a mobile change that took four rounds because each round was validated against the one shape that made it look complete. The common cause every time was a test that ran without exercising the thing it named — which is exactly what layers 1–3 above produce if you stop at "it passes".

The finding is solid and independent of the patch: **the fence collapses indeterminate into refused while its sibling twelve hundred lines up does not**, and the #433 comment already argues the principle. The next pass should build the harness first, confirm it fails against current main, and only then apply the change.

PR #1295, opened 2026-08-10
fix(panel): absorb the post-restart reconciliation window in mode:"current" (#1292)

### LINKED ISSUE #1292: Retry transient unconfirmed workflow binding after ComfyUI restart

artokun:
Fixes #1292

A recovery operation that tells you to retry it has not finished recovering.

Right after `panel_restart_comfyui` and the tab's reconnect, the panel answers `workflow_list` but marks its active record UNCONFIRMED. Corroboration correctly refuses — a stale or mixed reply can carry another canvas's perfectly valid uuid, and stamping it would wedge the tab while reporting `bound`. So the refusal is right; what was wrong is **who paid for it**. The caller got a hard failure whose own remedy was "call this again in a moment" — and the reporter did, five seconds later, and the identical call returned `already_current` with no user action in between.

`mode:"current"` now rechecks with a **fresh read**, three times over ~2.9s, before failing.

**The corroboration rule does not move.** Every recheck is judged by the same fail-closed test; a recheck buys a fresh snapshot to judge, never a lower bar. Both halves are asserted.

Only the `uncorroborated` failure rechecks. Its sibling `no_uuid` means the installed pack exposes no workflow identity at all — rechecking a build that will never answer differently just makes the same error slower.

And the remedy no longer prescribes what the tool just did: after four refused reads, "call this again in a moment" is the least likely thing left to work and reads as though nothing was tried. It now reports the rechecks and names the remedy that has *not* been tried.

Status: draft — full suite and review pending.

### DIFF
--- a/src/__tests__/orchestrator/fence-rebind-transient.test.ts
+++ b/src/__tests__/orchestrator/fence-rebind-transient.test.ts
@@ -102,8 +102,11 @@ describe("mode:'current' absorbs the post-restart reconciliation window (#1292)"
     expect(stamp, "the live canvas's identity was adopted").toBe(LIVE);
     expect(res.isError).toBeFalsy();
     expect(text).toMatch(/"graph_binding": "bound"/);
-    // It re-read rather than reasoning its way past the refusal.
-    expect(listCalls).toBeGreaterThan(1);
+    // It re-read rather than reasoning its way past the refusal — and STOPPED at
+    // the read that succeeded. No further round trip happens between the read we
+    // corroborated and the stamp we write from it, which is what keeps the
+    // adopted record and the tab being stamped the same tab (codex review, P1).
+    expect(listCalls, "one recheck, then adopt — no read after the good one").toBe(2);
   });
 
   it("recovers from a MIXED list too — same window, different symptom", async () => {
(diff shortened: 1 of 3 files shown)


artokun:
## Codex adversarial review — NO-SHIP, and P2 was a real defect

**P2 — retried states that can never settle. FIXED.** The loop retried *every* corroboration failure. An older panel that sends no `workflows` array can never self-corroborate, so the fix made that caller's identical error ~2.9s slower and then advised them to try again in a moment. This is the same reasoning I used to keep `no_uuid` out of the loop — I just failed to apply it one level down.

Corroboration reasons are now classified **snapshot vs shape**: what is open / which entry is flagged / whether the panel has confirmed *settles*; which fields a build sends does *not*. Only the former is re-read.

That also split a message which had folded two facts — an **absent** `workflows` field (old build) and an **empty** one (tabs not back yet) both read "the reply carried no open-workflow list". Different causes, different remedies, and only one of them is worth waiting for.

**P1 — retry can stamp the wrong tab. Not reachable through this change, and I did not take it on faith.** The loop breaks at the top of the next iteration, so *no round trip happens between the read we corroborate and the stamp we write from it* — the adopted record always comes from the final read, taken after any move `ctx.call` made. The recovery test now asserts **exactly two** reads rather than "more than one", so an edit that adds a call after the good read fails loudly.

Codex is right that the single-read hazard is pre-existing (it says so itself); this PR neither introduces nor widens it. I am not claiming it is safe — only that it is untouched here.

**Latency, ~26.9s worst case — not reachable.** A read that times out does not return an uncorroborated reply; it throws or yields `isError`, which exits via `unreadableOrHealed` immediately. Four 6s timeouts cannot stack: the first one ends the loop. The real bound is one slow read plus ≤2.9s of sleeps.

**Broadened `unreadable`/`healed_by_panel` after a first uncorroborated read — accepted deliberately.** If a later recheck cannot be read, that *is* the newer and truer state, and the same path detects a fence the panel repaired underneath us mid-window (#1043/#932) — worth more than preserving the first read's diagnosis.

Both new behaviours are mutation-verified: dropping the `settles` guard, and mis-labelling the shape failure as transient, each kill the test that covers them.

Full suite green — 420 files, 8299 tests.

Issue #1298, opened 2026-08-10
Fix apply_manifest refusing valid download when stale model copy exists


artokun:
### What happened / root cause
apply_manifest resolved the correct live models root for most Anima assets, but when findExistingModel found a same-named SAM file in a stale COMFYUI_PATH install outside the connected server's live model roots, it reported failure and skipped the required download. A stale unrelated copy should not block downloading the model into the resolved live target.

### Steps to reproduce
1. Connect comfyui-mcp to a ComfyUI whose live models root differs from COMFYUI_PATH.
2. Put models/sams/sam_vit_b_01ec64.pth only under the stale COMFYUI_PATH tree.
3. Apply the bundled anima-img2img manifest.
4. The SAM item fails instead of downloading into the live models root.

### Exact error (scrubbed)
A file exists at ~/ComfyUI-Installs/ComfyUI/ComfyUI/models/sams/sam_vit_b_01ec64.pth, but it is NOT in any directory the connected ComfyUI reads models from — it belongs to an install the running server does not use.

### Fix
Applied locally: yes (dist-only global install). Upstream-only: no.
When visible === false, enqueue startDownloadJob for target.targetSubfolder/target.filename and continue, instead of reporting failure and skipping the manifest item.

Patch:
```diff
 if (visible === false) {
-  results.push(report(... failed stale install ...));
+  const { job, settled } = await startDownloadJob(model.url, target.targetSubfolder, target.filename);
+  enqueued.push({ item, job, settled });
   continue;
 }
```

### Environment
macOS darwin; ComfyUI 0.31.0; GPU MPS 32 GB; torch 2.10.0; Python 3.12; comfyui-mcp 0.50.84; panel 0.11.56.

PR #1299, opened 2026-08-10
fix(manifest): a stale out-of-scope copy must not block a live download (#1298)

artokun:
Draft — claiming #1298. Reported on **0.50.84** (current), `grep -rn "#1298" src/` finds no prior work, no competing branch.

## The report

`apply_manifest` resolved the correct live models root for most assets, but `findExistingModel` matched a same-named SAM file in a **stale `COMFYUI_PATH` install outside the connected server's live model roots** — and reported that as already-present, skipping the required download.

## Why it is the same shape as most of this backlog

The presence check answered a *different question* than the one the caller asked. "Is there a file with this name somewhere I know about" is not "is this model available to the server that will run the workflow." Answering the first as though it were the second produces a confident skip, and the failure surfaces later as a missing model at render time — far from the decision that caused it.

That is #796's family with the arrow pointing at a **skip** rather than a refusal: a determined-yes from evidence that could not support it.

## Plan

1. **Read the consumer before the producer.** Find where `apply_manifest` calls `findExistingModel` and what it does with a hit — whether the scope error is in the search roots or in how the result is interpreted. Today's repeated lesson is that the defect is usually one layer below where it appears.
2. Establish what "live model roots" means here and whether the resolved target is already available at the call site — if it is, the fix is scoping the search to it rather than inventing new resolution.
3. **The negative is the deliverable.** A copy that IS in the live roots must still short-circuit the download. A fix that narrows the search so far that nothing matches would turn a wrong skip into a wrong re-download of every asset — worse, because it is slow and silent rather than fast and wrong.

Explicit non-goal: changing `findExistingModel`'s behaviour for other callers until I have read them. It is a shared helper, and other call sites may legitimately want the broader search.

Refs #796.

### DIFF
--- a/src/__tests__/services/manifest.test.ts
+++ b/src/__tests__/services/manifest.test.ts
@@ -974,6 +974,46 @@ describe("applyManifest", () => {
     }
   });
 
+  // Round 2: three of the six note-renderings were unpinned — including
+  // wrongPlace, which my own commit called the worst case ("two same-named
+  // copies on disk and neither message mentions the other"). Deleting any of
+  // those three appends left a green suite, so each could regress silently to
+  // exactly the state round 1 shipped.
+  describe.each([
+    {
+      name: "a FAILED download",
+      file: "failcase.safetensors",
+      arm: () => downloadModelMock.mockRejectedValueOnce(new Error("boom")),
+    },
+    {
+      name: "a landed-but-NOT-VISIBLE placement (wrongPlace)",
+      file: "wrongplace.safetensors",
+      arm: () => verifyLandedModelMock.mockResolvedValueOnce({ liveVisible: "not-visible", note: "" }),
+    },
+    {
+      name: "an UNCONFIRMED placement",
+      file: "unconfirmed.safetensors",
+      arm: () => verifyLandedModelMock.mockResolvedValueOnce({ liveVisible: "unknown", note: "" }),
+    },
+  ])("names the stale copy on $name (#1298)", ({ file, arm }) => {
+    it("carries the note", async () => {
+      resolveExistingModelFileMock.mockResolvedValueOnce({
+        path: `C:/stale/models/checkpoints/${file}`,
+        root: "C:/stale/models",
+        info: { isFile: () => true },
+      });
+      isUnderLiveModelRootsMock.mockResolvedValueOnce({ inRoots: false });
+      arm();
+
+      const result = await applyManifest({
+        manifest: {
+          models: [{ url: `https://example.com/${file}`, model_type: "checkpoints", filename: file }],
+        },
+      });
+      expect(result.results[0].message ?? "").toMatch(/same-named file also exists/);
+    });
+  });
+
   it("does not block on MANY slow downloads — enqueues all, one bounded grace (#362)", async () => {
     const prevGrace = process.env.COMFYUI_MCP_DOWNLOAD_GRACE_MS;
     process.env.COMFYUI_MCP_DOWNLOAD_GRACE_MS = "50";
--- a/src/__tests__/services/manifest.test.ts
+++ b/src/__tests__/services/manifest.test.ts
@@ -491,6 +491,35 @@ describe("applyManifest", () => {
     expect(msg).toMatch(/same-named file also exists/);
     expect(msg).toMatch(/C:\/stale\/models/);
     expect(msg).toMatch(/was ignored/);
+    // Re-pinned from the #369 tests this replaced: that clause IS the diagnostic
+    // half #369 wanted kept, and after the rewrite nothing held it — stripping it
+    // killed zero tests.
+    expect(msg).toMatch(/NOT in any directory/);
+  });
+
+  it("does not accuse LATER items of a stale copy found for an earlier one", async () => {
+    // Surviving mutation: hoisting `let staleOutsideLiveRoots` above the per-item
+    // loop passes every other test, because both stale tests use a single-model
+    // manifest. Manifests are multi-model by definition, so the mutant ships a
+    // false accusation on every subsequent asset.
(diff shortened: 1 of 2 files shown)


artokun:
### Located. The containment guard is right; what it *does* on a negative is wrong.

Plan step 1 answered, and it inverts the framing. `findExistingModel` has exactly **one** caller (`manifest.ts:907`), and the scope check #1298 asks for **already exists** — `isUnderLiveModelRoots`, carrying comments from three separate codex rounds that describe this exact hazard. The bug is not a missing check. It is what the check does when it succeeds in proving the file irrelevant.

`manifest.ts:929-941`:

```ts
if (visible === false) {
  results.push(report("model", item, "failed",
    `A file exists at ${existing}, but it is NOT in any directory the connected ComfyUI reads
     models from — it belongs to an install the running server does not use. Point COMFYUI_PATH
     at the ComfyUI that is actually running, or launch it with an absolute --base-directory.`));
  continue;
}
```

`visible === false` is the case where we have **positively established the file is irrelevant** — it belongs to an install the running server does not read. The correct conclusion from that is *"ignore it and download to the resolved live target"*. Instead it reports **failed** and `continue`s past the download.

Note the contrast one branch down: `visible === undefined` — where we genuinely could not tell — is handled carefully as `pending` with a diagnostic (rounds 4 and 8). So the *uncertain* case is treated better than the *certain-and-harmless* one.

The remedy text is also aimed at the wrong problem. It tells the user to repoint `COMFYUI_PATH`, but per the report `apply_manifest` had already **resolved the correct live models root** — nothing needs repointing. The stale copy is not in the way of anything; it is simply not the file we want.

### What the fix has to be, and the negative that guards it

On `visible === false`: ignore the stale hit and proceed with the download to the resolved live target, with the stale path surfaced as a **note on the successful item** rather than a failure — the user should still learn that a same-named file exists elsewhere, since that is genuinely confusing at render time.

The discriminating negative, unchanged from my plan: **`visible === true` must still skip.** A live copy is the whole point of the shortcut, and a fix that downloads unconditionally would re-fetch every asset on every apply — slow and silent, which is worse than fast and wrong.

Also unchanged: `visible === undefined` must keep its `pending` treatment. Three rounds of review built that branch specifically to avoid promoting an unverified answer, and this change must not collapse it into either neighbour.

Not implementing on this pass. The diagnosis is complete and the change is small, but the surrounding branch carries a lot of hard-won reasoning and I would rather write the `visible === true` and `undefined` regression tests first, against current main, than edit it and check afterwards.

artokun:
### This is not a slipped bug — it reverses a deliberate #369 decision that has a test

`manifest.test.ts:455-480`:

```
// #369 — "already exists" is a claim about a path the RUNNING server may not read.
it("FAILS an existing file that is not in any tree the connected ComfyUI reads", ...)
  expect(result.summary).toMatchObject({ skipped: 0, failed: 1 });
  expect(downloadModelMock).not.toHaveBeenCalled();
```

The `failed`-and-skip behaviour is **intentional, named, and pinned** — including an explicit assertion that the download is *not* attempted. A second test at `:492` does the same. So my previous comment was wrong to call this "what the check does on a negative is wrong" as though it were an oversight; someone decided this, and wrote a test so it would stay decided.

**What #369 fixed was a false SKIP.** A stale same-named file satisfied the manifest while the live server had never seen it, so the item was reported installed and the render failed later. The fix moved that from *skip* to *fail*.

**What #1298 proposes is a third option #369 did not take: download it.** That is a coherent argument — if the file is provably outside every root the server reads, it is not evidence about anything, and the resolved live target is where the model should go. The reporter's case is exactly that: the live root resolved correctly, so nothing needed repointing, yet the download was skipped.

Both positions are defensible, and picking between them is not mine to do silently:

- **Keep failing** if the reasoning is that a stale copy signals a misconfigured environment the user should fix before we write more files into a possibly-wrong place.
- **Download** if the reasoning is that we resolved the live target correctly and an irrelevant file elsewhere should not veto that.

I lean toward downloading, with the stale path surfaced as a **note on the succeeded item** — it keeps the diagnostic that #369 wanted while removing the veto #1298 is complaining about. But that is a preference, not a finding, and it changes tested behaviour under an issue number that is not #369.

**@artokun — your call.** If you want it changed I will make it, with the `visible === true` (still skips) and `visible === undefined` (still pending) negatives pinned first and the two #369 tests updated with the reasoning written at the site rather than quietly flipped.

Holding here rather than reversing a decision someone made on purpose.

artokun:
One clarification on whose call this is, then I will stop pinging it.

#1298 is tagged **`via-panel`**, i.e. auto-filed from a session that hit the behaviour rather than written as a design ruling. So its title — *"Fix apply_manifest refusing valid download"* — is the reporting session's framing of what it wanted to happen, not a decision that supersedes #369. A session blocked by a guard will always describe the guard as the bug; that is what being blocked feels like from inside.

Which means the conflict is real and still open:

- **#369** decided a stale out-of-scope copy makes the item **fail**, and pinned it with a test that asserts the download is not attempted.
- **#1298** reports that same decision as a defect, from a case where the live target had already resolved correctly.

The evidence in #1298 is good and specific — the live models root *was* resolved, so the remedy the failure suggests ("point COMFYUI_PATH at the ComfyUI that is actually running") is advice the user cannot act on because nothing is misconfigured. That is a genuine problem with the current behaviour whichever way the decision goes: **at minimum the message is wrong**, even if the fail-vs-download choice stands.

So there is a smaller change available that needs no ruling: keep the refusal, fix the remedy to say what is actually true — a same-named file exists outside the live roots, it is being ignored, and here is the path — rather than telling the user to repoint a path that is already correct. I can do that one now if you would rather not decide the larger question yet.

Otherwise this waits on: fail, or download-with-a-note.

PR #1301, opened 2026-08-10
fix(run-correlation): stop reporting a forgotten run as one you never queued

artokun:
Draft — claiming artokun/comfyui-mcp-panel#925. The panel triage on that issue established the message is orchestrator-side, so the fix lives here.

## The defect is a #796 fold

`correlate()` returns `foreign` whenever there is no open ticket for the prompt id. `foreign` is then rendered as:

> This run (prompt …) does **NOT** match any run you queued with panel_run — its origin is UNDETERMINED.

But "no ticket" covers two different facts:

| | truth | what the agent is told |
|---|---|---|
| never ours | a real run this session did not queue | correct |
| **ours, forgotten** | ticket evicted (cap is 64), or a re-delivery after the entry was acked | **"does NOT match any run you queued"** |

The second is the reporter's case exactly: their own prompt id, ComfyUI history confirming it succeeded, and the orchestrator telling the agent it was not theirs. The journal's own header already names this as the thing to avoid — *"it does not merely mislabel it: it TELLS THE AGENT TO DISBELIEVE A CORRECT RESULT"* — and the eviction path even documents `foreign` as "the honest reading once we've forgotten the run". It is honest about UNDETERMINED; it is not honest about *does not match any run you queued*.

## What I intend to change

Only the claim, not the correlation. A completion that cannot be proven to be yours must still never be presented as the awaited render — that rule stays. What changes is that "we forgot" stops being reported as "it was not yours", where the journal can tell the difference (it already keeps a delivered-memo of (tab, run) pairs well beyond the ticket cap, and knows when it evicts).

Unit-testable end to end, which is why I picked it: the dev rig's ComfyUI is currently serving another session's branch, so panel-mounting e2e is unavailable to me.

### DIFF
--- a/src/__tests__/orchestrator/run-correlation-prior.test.ts
+++ b/src/__tests__/orchestrator/run-correlation-prior.test.ts
@@ -1,0 +1,177 @@
+// #925 (panel) — "does NOT match any run you queued" was being said about a run
+// the session DID queue.
+//
+// The reporter queued a render with panel_run, got prompt id 8812f71b…, and its
+// re-delivered completion carried that exact id. ComfyUI's history confirmed the
+// prompt succeeded. The orchestrator told the agent it was not theirs.
+//
+// The cause is a fold, not a lost ticket. `correlate()` returns `foreign` for
+// every completion with no OPEN ticket, and `foreign` renders as a definite
+// negative — "does NOT match any run you queued". But "no open ticket" covers
+// two different facts:
+//
+//   never ours      a real run this session did not queue
+//   ours, forgotten the ticket was evicted (the map is capped at 64), the entry
+//                   was already delivered and acked, or the id was re-queued and
+//                   now stands for more than one run
+//
+// Both are UNDETERMINED, both must be refused identically — a completion that
+// cannot be proven to be yours must never satisfy an outstanding panel_run, and
+// that rule does not move here. What changes is that the orchestrator stops
+// asserting the one thing it does not know. The journal's own header says why
+// this matters: it "does not merely mislabel it: it TELLS THE AGENT TO DISBELIEVE
+// A CORRECT RESULT".
+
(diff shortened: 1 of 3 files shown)


artokun:
## Codex adversarial review — NO-SHIP, two P1s, same root: my evidence did not support my sentence

The message says **"WAS queued from this session"**. What I checked was weaker than that, in two reachable ways.

**P1 — `hasHistoryFor` means SEEN, not QUEUED.** It is also true of a foreign completion that was delivered here once. Acquaintance with an id is not ownership of the run.

**P1 — `Boolean(ticket?.reused)` skipped the ownership test.** Session A re-queues prompt P; session B gets a completion for P; B is told it queued the run. That is this exact bug, pointed at a different party — and it would have made the refusal *weaker*, not just wrong.

Both are ownership facts by construction now: a ticket **this** party opened and we later evicted (recorded **at** the eviction, the last moment ownership is still known), or a live ticket **we own** whose id was re-queued. Mutation-verified: dropping the ownership test on `reused` fails the cross-conversation case.

**Also corrected, and it was my own text:** the sentence told the agent that an acked completion clears a ticket. My own test disproves that — the ticket is marked settled and stays in the map. It now names ageing out and re-queueing, which are what actually cause this.

**P2 — behavioural coverage added.** The flag is asserted on the *delivered payload*, not only in the journal, because it could be right internally and dropped at the wire boundary — in which case nothing changes for the agent.

**One limit I am stating rather than implying.** The "foreign completion seen here once" test passes under *both* the old evidence and the new, so it does not prove the narrowing was necessary — the delivered memo appears not to be written for a completion with no ticket. I kept the test and wrote that into it, because the property is the right one and a change to when the memo is written would land there first. The discriminating case is the cross-conversation reused ticket.

Full suite: 422 files, 8322 tests, lint clean.

Issue #1300, opened 2026-08-10
download_model: CivitAI failures report a bare status code with no body, hiding 401-needs-token and metadata-URL 404s


artokun:
### What happened

Downloading one specific FILE from a multi-file CivitAI model version took four attempts because `download_model action:"download"` gives no diagnostic beyond a status code.

Target: GonzaLomo Krea 2 v2.0 (model version `3155593`), which publishes six files (bf16 / fp8 / int8 / two GGUF / a second fp8). The workflow wanted the int8 one.

Sequence:

1. `action:"download"` with the metadata-query URL -> `{"error":"MODEL_ERROR","message":"Download failed: 404 ","details":{"url":"..."}}`
2. Assumed a bad URL, re-read the CivitAI API, retried the canonical `downloadUrl` -> same bare 404.
3. Probed by hand with curl: unauthenticated the endpoint returns **401**, not 404. No `CIVITAI_API_TOKEN` was configured. Nothing in the tool output suggested auth.
4. Saved a token via `panel_request_secret`, retried -> **still** `404`. Confirmed via curl that this particular metadata-query URL (`?type=UNet&format=SafeTensor&fp=int8`) genuinely 404s even WITH a valid token, despite being the exact `downloadUrl` string the CivitAI API itself returns for that file.
5. Switched to the explicit per-file form `?fileId=3036509` -> HTTP 206, binary, downloads correctly.

### Two distinct problems

**(a) No response body, no auth hint.** The error is `Download failed: <code>` and nothing else. A CivitAI 401 (no token) and a genuine 404 are indistinguishable, and the message never mentions `CIVITAI_API_TOKEN`. Since CivitAI now requires auth for *all* model downloads, an unconfigured token is probably the single most common failure of this action, and right now it presents as a URL problem — which is exactly the wrong place to look. Note also the discrepancy in step 1/3: curl saw 401 where the tool reported 404, so whatever redirect/normalization happens internally is also losing the original status.

**(b) CivitAI metadata-query downloadUrls are unreliable; `?fileId=` is not.** For a version with multiple files sharing metadata, the `?type=&format=&fp=` URLs that the CivitAI API publishes can 404 while `?fileId=<file.id>` works. `action:"download_civitai"` only resolves a model/version to its PRIMARY file, so there is currently no supported way to fetch a non-primary file (the fp8, the int8, a GGUF) other than hand-constructing a URL — and the obvious hand-constructed URL is the broken one.

### Suggested fixes

[...]

PR #1303, opened 2026-08-10
fix(download): say why a CivitAI download failed, instead of a bare status

artokun:
Draft — claiming #1300.

The reporter needed **four attempts** to download one file, because `download_model action:"download"` says only `Download failed: 404`.

Two things are wrong, and the first explains why the second was never reached:

1. **The response body is thrown away.** CivitAI answers with a JSON error explaining the problem; the tool reports the status code alone. A 401-needs-token and a genuine 404 are indistinguishable.
2. **The auth hint exists but did not fire.** There IS a CivitAI token hint — gated on `401 || 403`. The reporter got a **404** (curl showed 401 for the same URL unauthenticated), so the one message that would have solved it on attempt one stayed silent.

I will also record what they established about URL forms — a metadata-query `downloadUrl` (`?type=UNet&format=SafeTensor&fp=int8`) 404s even WITH a valid token, while `?fileId=…` works — since that is the actual remedy for the residual 404 and it came from their curl, not from a guess.

Nothing here logs a token value; only whether one is configured.

### DIFF
--- a/src/__tests__/services/download-failure-hint.test.ts
+++ b/src/__tests__/services/download-failure-hint.test.ts
@@ -48,6 +48,35 @@ describe("the CivitAI hint fires on the status the reporter actually got (#1300)
     expect(hint).not.toMatch(/Set it in panel Settings/);
   });
 
+  it("BUT a 401/403 WITH a token means the token is bad — not 'probably not auth'", () => {
+    // Ignoring the status here would repeat the original bug one layer along:
+    // an invalid, expired or unentitled token produces exactly this status, and
+    // steering that caller away from auth is the same wrong answer (codex).
+    for (const status of [401, 403]) {
+      const hint = downloadFailureHint({ status, url: METADATA_URL, hasCivitaiToken: true });
+      expect(hint, `status ${status}`).toMatch(/INVALID, expired, or lacks access/);
+      expect(hint).not.toMatch(/probably not an auth failure/);
+    }
+  });
+
+  it("the fileId remedy is for a 404 only, not for 429 or 5xx", () => {
+    for (const status of [429, 500, 503]) {
+      const hint = downloadFailureHint({ status, url: METADATA_URL, hasCivitaiToken: true });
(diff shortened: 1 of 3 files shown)


artokun:
## Codex adversarial review — NO-SHIP, three P1s

Two were defects in the fix itself; one was the same wrong answer I was fixing, one layer along.

**P1 — the bound was on the OUTPUT, not the READ.** `res.text()` drains the whole response before the 400-character cap applies, so a huge or never-ending error body could hang the download or exhaust memory *while we were busy improving its error message*. Now an 8 KB cap and a 3s deadline on the stream itself, and it **cancels** what it did not read rather than leaving the socket draining.

**P1 — a configured-but-invalid token was told "probably not an auth failure".** Once a token existed the hint ignored the status — but 401/403 *with* a token is precisely the signal for one that is invalid, expired, or lacks entitlement for a gated model. That caller was being steered away from the answer, which is exactly the defect #1300 reports. The status is read on that branch now, and the URL-form remedy is confined to 404 rather than offered for 429/5xx.

**P1 — reflected credentials.** The shared scrubber catches configured values and opaque runs, but a signed download URL carries its credential in the **query**, and a host reflecting the request would republish it. Those values are ours and known exactly, so they are redacted by identity (raw and percent-encoded) alongside the shape-based pass. Short values are skipped — `type=UNet` is the diagnosis, not a secret.

**P2 — trailing-dot FQDN.** `civitai.com.` is a legitimate spelling and was missing the branch. Stripped for comparison only, as `utils/origin.ts` already does.

Its check on the other host shapes I agree with and did not change: `civitai.com.evil.tld`, homoglyphs and `civitai.com@evil.tld` are correctly rejected — the direction that matters most.

Nine tests added, one per finding plus the neighbour that keeps each honest: the byte cap asserts the reader **stopped** rather than that the output was short, and the short-query-value case guards against over-redaction destroying the diagnosis.

Full suite: 423 files, 8347 tests, lint clean. (One earlier run showed 4 failures in `tunnel.test.ts`; they pass in isolation and on a re-run — parallel flake, not this change.)

PR #1305, opened 2026-08-10
fix(ui-bridge): give an interactive card a disconnect remedy that applies to it

artokun:
Draft — claiming the orchestrator half of artokun/comfyui-mcp-panel#952.

The reporter's `panel_ask` was interrupted by a tab disconnect and got:

> OUTCOME UNKNOWN: the command was already sent … Verify before retrying (e.g. check queue action:"list" / get_image (action:"list_outputs"))

Their words: *"The suggested verification examples concern render queues/media and do not apply to an ask card."* They are right — that sentence is hardcoded for every command, so an interactive card gets told to go and check the render queue.

**Scope, deliberately narrow.** The panel-side trace on that issue ends on two open design questions (should interactive cards dedupe on a scope that survives a reconnect; what a retry should return while the original card is unanswered), and it was left open rather than guessed at. I am not settling those.

What I *can* fix is the part that is wrong today regardless of how they are settled: the remedy. And the panel trace establishes something the message must not omit — for this exact trigger, a blind retry **does** paint a second card, because the dedupe ledger is keyed by the socket's bridge epoch and a reconnect changes it, so `lookupRetry` misses and fails open.

So an ask/secret card should be told what actually applies: the card may already be on screen, a retry may duplicate it in front of a human, and there is no way to withdraw the first one.

### DIFF
--- a/src/__tests__/services/mid-command-remedy.test.ts
+++ b/src/__tests__/services/mid-command-remedy.test.ts
@@ -35,17 +35,44 @@ describe("the disconnect remedy fits what was interrupted (#952)", () => {
   it("…and says the thing that is actually true of it", () => {
     const msg = midCommandDisconnectMessage(ASK);
     expect(msg).toMatch(/card may already be on screen/);
-    // The fact from the panel trace: a retry is not merely unverifiable, it is
-    // known to duplicate after a reconnect.
-    expect(msg).toMatch(/paints a SECOND card/);
-    expect(msg).toMatch(/cannot be withdrawn/);
+    // The fact from the panel trace: a retry is not merely unverifiable, it
+    // duplicates after a reconnect because suppression is keyed to the dead
+    // socket. Stated as the consequence the user sees.
+    expect(msg).toMatch(/retry\s+suppression is keyed to the socket that dropped/);
+    expect(msg).toMatch(/user may see two/);
   });
 
-  it("names what the caller CAN do, since it forbids the obvious move", () => {
+  it("does NOT promise the answer will arrive — it will not (codex)", () => {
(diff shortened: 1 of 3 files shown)


artokun:
## Codex adversarial review — NO-SHIP, two P1s, and both were sentences I invented

I checked each against the panel source before rewriting, and it was right twice.

**P1 — "the answer to the card already up still arrives" is FALSE, and it was the advice.** The panel's `redactSensitiveReply` replaces an `ask_user`/`request_secret` reply that cannot cross the dropped connection with a payload-free failure — deliberately, because the content is the user's own input and a replay could land on a *different orchestrator*. Its own words: *"ask again on the current connection"*. So the answer will never reach the caller, and telling an agent to wait for it would have parked it indefinitely. That is **worse than the queue advice it replaced**: unhelpful became misleading. The message now says the answer will not arrive, and names the route that works.

**P1 — "ask in conversation" is unsafe for a secret.** `panel_request_secret` exists so the value reaches the panel through a masked input the agent never sees and that is never recorded in chat. My single sentence covered both interactive commands, so it told the caller to defeat exactly that guarantee — in a message an agent acts on. The branches now diverge: a secret is re-issued through the tool, with the conversational route explicitly forbidden and the reason stated. Mutation-verified: collapsing the two branches fails a test.

What it confirmed and I left alone: the duplicate-card claim (it traced the epoch mismatch independently), the exact command matching (`ask_user`/`request_secret` are the panel's own interactive set, no false positives), and that no consumer depends on the old remedy tail — the detectors key on `disconnected mid-command` / `OUTCOME UNKNOWN`, preserved on both branches.

Full suite: 424 files, 8364 tests, lint clean.

Issue #1309, opened 2026-08-10
gen-changelog files release commits as changes — this repo's own release subject matches neither matcher


artokun:
Found while cutting 0.50.86, and hand-corrected twice now.

## What happens

The generated entry carried the previous release as a user-facing change:

```
## [0.50.86] - 2026-08-09

### MCP
#### Fixed
- say why a CivitAI download failed, instead of a bare status (#1303)

#### Changed
- 0.50.85 (#1302)      <- the release-reconcile PR
```

## Why

`isReleaseSubject` (scripts/gen-changelog.mjs:80) recognises two shapes:

```js
/^release:/i                              // "release: v0.50.87 — …"
/^v?\d+\.\d+\.\d+\s*(\(#\d+\))?$/       // "0.50.85 (#1302)"
```

But the subject this repo's own release flow produces is **`chore(release): 0.50.85 (#1302)`** — `npm version patch -m "chore(release): %s"`, squash-merged with the PR number appended. `^release:` fails on the `chore(` prefix, and the bare-version pattern requires the version at the START. So it falls through to the conventional-commit parser at :111, which reads it as type `chore`, scope `release`, description `0.50.85 (#1302)` — and files it under Changed.

Both matchers are for shapes this repo has produced only occasionally; the one it produces on **every single release** is the one that is missed.

## Why it is not merely cosmetic

The de-duplication that makes the deliberately-wide range safe is "filter out PRs the changelog already documents". A release-reconcile PR is never documented — it is not a change — so nothing else catches it, and it lands in the next release's entry every time.

It also reads as a real change to anyone scanning the file, in the section whose whole job is to say what changed.

## Note

The panel repo hit the neighbouring version of this (panel#932) and fixed it in panel#933 by keying on a version at the start of the subject — which is the half this repo already has. This is the other half.

## Fix

Recognise the conventional release shape (`chore(release):`, and the `release` scope generally) as a release subject.

PR #1310, opened 2026-08-10
fix(changelog): this repo's own release subject is a release subject

artokun:
Draft — claiming #1309.

`isReleaseSubject` matches `release:…` and a bare `0.50.85 (#1302)`, but the subject this project's release flow actually produces on **every** release is `chore(release): 0.50.85 (#1302)` — and that matches neither. It falls to the conventional-commit parser and is filed as a `chore`, so the previous release appears in the next one's entry under **Changed**.

Nothing else catches it: the de-duplication that makes the deliberately-wide commit range safe is "filter out PRs the changelog already documents", and a release-reconcile PR is never documented, because it is not a change.

I have hand-corrected this twice while releasing, which is the tell.

### DIFF
--- a/scripts/gen-changelog.mjs
+++ b/scripts/gen-changelog.mjs
@@ -77,7 +77,8 @@ function prevTag() {
  *  squash-merge form GitHub writes from a PR titled with the bare version — `0.49.6 (#849)`.
  *  Only the first was recognised, so the second fell through to the non-conventional
  *  path and was silently dropped along with everything else there. */
-const isReleaseSubject = (s) => /^release:/i.test(s) || /^v?\d+\.\d+\.\d+\s*(\(#\d+\))?$/.test(s);
+// #1309 — three shapes are needed; see scripts/lib/release-subject.mjs.
+import { isReleaseSubject } from "./lib/release-subject.mjs";
 
 /** Parsed commits since `range`, newest-first, minus noise.
  *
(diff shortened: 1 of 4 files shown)


artokun:
## Codex adversarial review — NO-SHIP, and it caught me shipping the failure I was fixing

**P1 — matching on scope alone drops real user-facing work.** `fix(release): prevent failed releases from corrupting user installs`, `feat(release): add release-channel selection`, `chore(release): repair the release workflow` are all scoped `release` and all genuine changes. My predicate swallowed them — silently, which is precisely the failure this generator warns about at length.

Worse, I had **written it up as an accepted consequence** rather than recognising it as a defect. "One dropped entry for a rare shape against a wrong entry on every release" was a trade I had no need to make.

A release-scoped subject must now carry **nothing but a version** (plus the PR number GitHub appends). That is what the release flow produces and what a fix to the release process never does — so they were separable all along. Both directions are asserted now.

The type stays `\w+` rather than pinned to `chore`: the flow could reasonably become `build(release)`, and it is the version requirement that makes this safe, not the type.

**Re-verified against the real history** — the parent of the 0.50.86 release commit still generates a clean entry with the tightened predicate.

**Packaging, which it also flagged:** `scripts` is in `package.json`'s `files`, so `scripts/lib/release-subject.mjs` ships rather than resolving only in a source checkout.

Its note on my comment wording is fair too — the bare-version shape *was* already matched; only the `chore(release)` one was missing, and the module comment says so.

Full suite: 425 files, 8374 tests, lint clean.

Issue #1290, opened 2026-08-10
Tests can silently depend on the live process table — no gate catches it


artokun:
Follow-up to #1263, filed so the class is prevented rather than fixed one file at a time.

## What happened in #1263

`download-manager-routing.test.ts > #420` failed on a clean `origin/main` locally while CI reported success for the same SHA. The cause was not the test: `shouldDispatchDownloadToManager` falls through to `resolveLiveServerRoot`, which calls `observeLivePython()` — and that shells out to **netstat/WMI to find whichever python is really serving the port on the machine running the test**.

So the assertion was decided by whether the developer happened to have ComfyUI running. With one up, a relative `main.py` argv anchors onto the live interpreter and the case routes local instead of to the Manager. On CI nothing is listening, the probe finds nothing, and the identical test passes. That is why it survived every CI run while failing locally.

Fixed by stubbing `resolveLiveInterpreter` at its module boundary.

## Why this needs a gate rather than eight more edits

Eight other test files exercise this area **without** stubbing the probe and pass today:

```
download-dest-roots, download-jobs, manifest, output-dir,
panel-autoinstall-tree-gate, panel-recovery-cluster,
panel-update-corroboration-message, restart-launcher-env
```

Rewriting eight passing files preemptively is a large diff with no observable failure behind it, which is why #1263 deliberately did not. But the failure mode is nasty in a specific way: **a test that exercises a code path only on some machines is a test that agrees with you on the others.** It does not fail loudly — it silently stops testing, and CI cannot tell you because CI is one of the machines where it passes.

This is the same shape as the environment difference that made an `optional-dep` test pass without reaching its branch (#796's third pass), so it is not a one-off.

## Suggested shape

A source-level gate, in the idiom this repo already uses for `check:unknown-collapse` and the truncation-remedy checker: fail when a test file reaches the live-process probe without stubbing it, with the current eight recorded as known sites so the count can only go down.

The gate has to be verified in its false-negative direction — mutate a file that DOES stub the probe into one that does not and confirm the gate fires — or it is just a list.

Not user-visible. Filed because the next occurrence costs an afternoon of proving a change innocent, and the recipe is currently only in a commit message.

Related (same root cause): #1263

Issue #1263, opened 2026-08-09
download-manager-routing test fails locally on a green main (order- or environment-dependent)


artokun:
Not blocking anything — filing it because a test that passes on CI and fails locally is the shape that eventually fails on CI too, and the evidence is cheap to record now.

## Measured

On a clean `origin/main` at `bdacfc1` (0.50.73), with no working-tree changes:

```
Test Files  1 failed | 407 passed (408)
Tests  1 failed | 8169 passed | 2 skipped (8172)

FAIL src/__tests__/services/download-manager-routing.test.ts
  > #420: no local base + reachable server WITHOUT --base-directory → Manager route
  AssertionError: expected false to be true
  at download-manager-routing.test.ts:86
```

It fails **both** in the full suite and in isolation locally. GitHub Actions reports `success` for that exact SHA on all three platforms.

## What the test asserts

```ts
hoisted.base = undefined;
hoisted.stats = { system: { argv: ["main.py", "--listen"] } };
expect(await shouldDispatchDownloadToManager()).toBe(true);
```

With no local base it should route to Manager. Getting `false` means something answered with a base — which is why the panel-base cache is the first place I would look.

## Why that is the suspicion

Two changes to that cache landed close together:

- **0.50.71** (#1250) added a clear-epoch so an in-flight `primePanelBase()` cannot write after a deliberate clear.
- **#1253** then added `__resetPanelBaseCache()` to five more test files.

Both change what state a test inherits. I have not confirmed either is the cause — `shouldDispatchDownloadToManager` may resolve a base by a path that has nothing to do with that cache, and the local-vs-CI split could equally be a platform difference (this is Windows; CI passes on Windows too, which argues against it).

## Not claiming this

I hit it while working #1259 and confirmed my branch does not cause it (identical failure with my changes stashed). Recording it rather than folding a second investigation into an unrelated PR.

artokun:
Fixed and merged. The cause was better than my guess in the issue — it had nothing to do with the panel-base cache.

**`shouldDispatchDownloadToManager` reads the process table.** When argv carries a relative `main.py` and the server reports no cwd, `resolveLiveServerRoot` falls through to `observeLivePython()`, which calls `resolveLiveInterpreter` — and that shells out to netstat/WMI to find whichever python is actually serving the port **on the machine running the test**.

[...]

PR #1314, opened 2026-08-10
test(#1263): prove the live-process stub is in force, instead of assuming it

artokun:
Draft — claiming the #1263 recurrence on 0.50.87.

## What I measured, and could not reproduce

On this rig — ComfyUI serving 127.0.0.1:8188, Windows 11, **Vitest 4.1.8 (the reporter's version)**, clean `main` — `download-manager-routing.test.ts` passes, 12/12.

Every environment-dependent input the #420 case touches is already controlled:

| input | state |
|---|---|
| `resolveLiveInterpreter` | stubbed — and **provably effective**: feeding it an interpreter flips #420 from `true` to `false`, exactly the reported symptom |
| `getSystemStats` | mocked, so the live server's real argv cannot leak in |
| `existsSync` | mocked |
| `resolveEffectiveComfyUIBase` | mocked |
| `isRemoteMode` | mocked |

So the failure requires the stub NOT to intercept on their machine. Which is checkable — and nothing currently checks it.

## What this PR does

The #1290 gate I added for this issue asserts that a spec **mentions** `live-interpreter`. It does not assert the mock actually takes effect. If interception ever breaks — a vitest change, a path resolution difference, an import that bypasses the specifier — the stub goes inert, the test silently reads the real process table, and we are back to "green on CI, red on a developer box" with no signal.

So the spec will assert, at runtime, that the stub is the function in force. If it is not, the suite says so and names #1263, instead of producing a confusing routing assertion failure.

That does not fix the reporter's machine by itself. It makes the next occurrence self-identifying, which is the part I can do without reproducing it.

### DIFF
--- a/src/__tests__/services/download-manager-routing.test.ts
+++ b/src/__tests__/services/download-manager-routing.test.ts
@@ -21,6 +21,10 @@ const hoisted = vi.hoisted(() => ({
   /** #1263 — the interpreter the live-process probe "finds", instead of whatever
    *  is really running on the developer's machine. undefined = nothing found. */
   livePython: undefined as string | undefined,
+  /** The stub itself, created here so the spec can assert BY IDENTITY that this
+   *  is the function in force, and count the calls the code under test makes to
+   *  it (codex). Its behaviour is installed in `beforeEach`. */
+  liveInterpreter: vi.fn() as ReturnType<typeof vi.fn>,
 }));
 
 // The live-root routing branch only streams local when the root exists locally.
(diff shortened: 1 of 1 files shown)


artokun:
## Codex adversarial review — NO-SHIP, and its High finding was the point of the whole PR

**A guard that could not see its own failure mode.** My check proved *this file's* import was mocked. It said nothing about which module `shouldDispatchDownloadToManager` resolves — so the exact scenario I wrote it for, the SUT reaching another copy of `live-interpreter`, would have **passed** it while every routing assertion silently consulted the real process table.

Three changes, all from the review:

- **Identity, not behaviour.** The stub is created in `vi.hoisted` and handed to the mock factory, so the spec asserts `resolveLiveInterpreter === hoisted.stub`. An output check is satisfied by any implementation returning the same two values.
- **The SUT must call it.** The #420 shape — no local base, a relative `main.py` that can only be anchored through the live interpreter — is driven, and the spy's call count asserted. This is the only check that catches a duplicated module instance, and it is the leading hypothesis for the reporter's machine.
- **try/finally** around the fixture mutation, so a failed expectation cannot leave it changed for anything running before the next `beforeEach`.

Verified by breaking it: pointing the mock at a specifier the SUT does not use fails both new checks *with their own names*, alongside the reporter's exact symptom (`#420: expected false to be true`).

Full suite: 426 files, 8387 tests, lint clean. (One run showed a failure in `ui-bridge.test.ts`; it passes in isolation both with and without this change — parallel flake, same as the `tunnel.test.ts` one I saw earlier.)

Issue #535, opened 2026-07-31
panel_restart_comfyui cannot relaunch live ComfyUI when command is relative main.py


artokun:
### What happened / root cause
panel_restart_comfyui correctly detected that legacy Manager 3.x has no reboot endpoint, then attempted its managed fallback. The live ComfyUI process is launched as `python main.py --cache-none --output-directory ...`, but the fallback validated the relative script string `main.py` against the orchestrator working directory instead of resolving it against the live process cwd. It therefore refused a safe restart although `/proc/<pid>/cwd` points to a valid ComfyUI install containing main.py.

### Steps to reproduce
1. Run ComfyUI from its install directory using `python main.py` with legacy Manager 3.x.
2. Connect comfyui-mcp with a stale/unset COMFYUI_PATH but a valid COMFYUI_URL.
3. Call panel_restart_comfyui.

### Exact error
`The built-in Manager exposed no reboot endpoint (legacy Manager 3.x). Tried the headless managed restart (kill + relaunch) as a fallback, but it could not restart ComfyUI: Refusing to restart: Resolved ComfyUI script does not exist on disk: main.py — COMFYUI_PATH may point at a stale/old install that has no runnable server. ComfyUI was left running (not stopped).`

### Fix
Applied locally: no. Upstream-only for this session: yes.
When the detected command contains a relative script path, resolve it against the live ComfyUI process cwd (Linux: `/proc/<pid>/cwd`; equivalent process metadata on other OSes), or against the detected launch cwd, before existence validation. Preserve the current refusal behavior if the resolved absolute path is still missing.

### Environment
Linux; ComfyUI 0.28.3; NVIDIA RTX 4080 16 GB; comfyui-mcp 0.48.21; panel 0.11.6; Manager 3.x legacy.

artokun:
Recurrence on the current releases: comfyui-mcp 0.49.4 and panel 0.11.38 (Windows 11 portable). The live server used `D:/ComfyUI_windows_portable/python_embeded/python.exe`, while stale `COMFYUI_PATH` pointed to a different checkout that contained `main.py` but had no discoverable embedded interpreter. `restart_comfyui` stopped the healthy server, then attempted bare `python.exe` and failed with `spawn python.exe ENOENT`, requiring manual absolute-path recovery. This is the same unsafe bare-interpreter relaunch path that the prior fix was intended to reject before stopping the server.

artokun:
Fixed and merged in #584 — closing.

[...]

Issue #709, opened 2026-08-02
panel_restart_comfyui claims "graph tools are ready" when the reconnected tab is still running stale panel JS (graph writes all fail)


artokun:
### What happened / root cause

After updating the panel node-pack on disk (0.11.20 -> 0.11.34, required 0.11.28+) and restarting ComfyUI via `panel_restart_comfyui`, the tool returned:

```
{ rebooting: true, ready: true, server_ready: true, confirmed_cycle: true,
  recovered_ms: 44974, panel_tab_reconnected: true,
  note: "...the panel tab reconnected - graph tools are ready.." }
```

But graph tools were NOT ready. Every subsequent graph MUTATOR still failed with:

```
"graph_add_node" cannot be safely targeted to the active workflow: panel tab wf:workf
does not enforce per-command workflow targeting (update the ComfyUI-MCP panel to edit
the graph). Read-only graph commands (graph_outline, graph_query, graph_get_state) still work.
```

Root cause: a ComfyUI server restart re-establishes the panel's websocket, but it does NOT make the BROWSER re-download the panel's frontend extension JS. The tab reconnects still running the OLD cached bundle, so `conn.enforcesWorkflowStamp` remains false and the #570 P0c fail-closed gate in `services/ui-bridge.js` keeps refusing every mutator.

`dist/orchestrator/panel-tools.js` (~line 4218) computes readiness as:

```js
const tabBack = ctx.awaitReachable
  ? await ctx.awaitReachable(Math.max(0, overallDeadline - Date.now()))
  : true;
...
ready: tabBack, // graph tools are usable only once a panel tab is bound
```

That comment states the intent ("graph tools are usable") but `awaitReachable` only proves a tab is BOUND, not that it is CAPABLE. The code already carefully separates `server_ready` from `ready`; it just stops one step short of capability. This is exactly the post-panel-update case, where a restart is the documented next step, so it is the most likely moment for the claim to be wrong.

The misleading part is the note text: it affirmatively tells the agent "graph tools are ready", which sends the agent straight into a retry loop and a pointless `panel_set_workflow_target({mode:"current"})` rebind (the error message's own suggestion), neither of which can ever help - the only fix is a hard refresh of the ComfyUI browser tab, which nothing in the output mentions.

### Steps to reproduce

1. Run a panel older than the orchestrator's `requiredPanelVersion` (here 0.11.20 vs required 0.11.28+).
[...]

PR #1315, opened 2026-08-10
fix(#535): anchor a relative launch script on Windows, where /proc has no equivalent

artokun:
Claiming the newest recurrence of #535 — the 2026-08-10 Windows 10 report:

> `Resolved ComfyUI script does not exist on disk: ComfyUI\main.py — could not locate the ComfyUI install; set COMFYUI_PATH or a default workspace.`

## Why the v0.48.23 fix does not cover it

That fix was right, and it works — on Linux. It anchors a relative launch script against the live process's own working directory, read from `/proc/<pid>/cwd` while the process is still alive.

`resolveLiveProcessCwd` (`src/services/process-control.ts`) opens with:

```ts
if (!pid || IS_WIN) return undefined;
```

So on Windows `info.liveCwd` is always `undefined`, the anchor at the `!scriptIsAbsolute && info.liveCwd` branch never fires, and the original refusal stands. My own close comment on this issue named that explicitly — *"Other OSes have no cheap `/proc` equivalent … a known limitation rather than an oversight."*

The limitation was accurate. It is no longer necessary.

## What changed since

#1133 (shipping in 0.50.90) established that Windows **does** have a usable equivalent for exactly this shape. `resolveLiveServerRoot`'s `observed-process` tier identifies the process listening on our port, correlates it against that server's own argv, and re-anchors a relative `main.py` on that interpreter's install tree — accepting the result only when the anchored directory really holds `main.py` and the interpreter belongs to that install.

I verified that tier live on a Windows rig while working #1133: the argv shape `ComfyUI\main.py` — the same one in this report — anchored to a real tree holding `main.py` and `custom_nodes`.

## Plan

- [ ] Confirm the Windows recurrence reaches the `!scriptIsAbsolute` refusal for the reason above, not a different one
- [ ] Anchor the relative script via the observed-process tier when `liveCwd` is unavailable
- [ ] Anchor the **interpreter** too — a relative script fixed with a bare `python` still relaunches into the wrong environment (the 2026-08-04 `spawn python.exe ENOENT` recurrence on this same issue)
- [ ] Keep refuse-safe: if the anchored path is still not a regular file, refuse and leave ComfyUI running. The anchor may only ever *add* a way to succeed.
- [ ] Platform-pinned tests (`node:os` `platform()` mocked both ways) — this path passes trivially on the wrong host

Draft until the mechanism is confirmed rather than assumed. This one has been closed-then-reopened three times, so the bar is a reproduction traced to the specific branch, not a plausible story.

### DIFF
--- a/src/__tests__/services/restart-live-first.test.ts
+++ b/src/__tests__/services/restart-live-first.test.ts
@@ -436,6 +436,122 @@ describe("restart_comfyui — live-first script resolution (#476, #426)", () => 
     killSpy.mockRestore();
   });
 
+  it("WINDOWS: a RELATIVE path argument blocks the inferred anchor (#535 codex gate)", async () => {
+    // The anchor is INFERRED, not observed — a directory from which this relative
+    // script WOULD name this install, which is not "the directory it was started
+    // in". A symlinked launcher satisfies the inference from the wrong place. That
+    // gap is harmless for the panel base (#1133) and NOT harmless here, because the
+    // value becomes the relaunch's working directory. So when any argument could
+    // itself be cwd-relative, the anchor is not adopted.
+    const LIVE_CWD = resolve("bundle");
+    mockResolveBase.mockReturnValue(undefined);
+    mockLiveRootFromArgv.mockReturnValue(undefined);
+    mockGetSystemStats.mockResolvedValue({
+      system: {
+        // `out` is cwd-relative: after a restart from a different cwd it would
+        // point somewhere else entirely.
+        argv: [join("ComfyUI", "main.py"), "--output-directory", "out"],
+      },
+    });
+    mockExistsSync.mockImplementation((p: string) => {
+      const s = String(p);
+      return s === ABS_PYTHON || s === join(LIVE_CWD, "ComfyUI", "main.py");
+    });
+    __processControlTestHooks.setLiveCwdResolver(() => undefined);
+    mockResolveLiveServerRoot.mockReturnValue({
+      source: "observed-process",
+      anchorDir: LIVE_CWD,
+      observedPid: 4321,
+    });
+    mockLivePortThenFree();
+    mockSpawn.mockImplementation(() => new FakeChild());
+    const killSpy = vi.spyOn(process, "kill").mockImplementation(() => true);
+
+    const result = await restartComfyUI();
+
+    expect(result.message).toMatch(/refusing to restart/i);
+    expect(result.stopped).toBe(false);
+    expect(mockSpawn).not.toHaveBeenCalled();
+    expect(killSpy).not.toHaveBeenCalled();
+
+    killSpy.mockRestore();
+  });
+
+  it("WINDOWS: a plain non-path value (`--port 8188`) does NOT block the anchor (#535)", async () => {
+    // The guard above must not be so broad it refuses the very case this issue is
+    // about. The reporter's argv carries `--port 8188`; a number cannot be a path.
+    const LIVE_CWD = resolve("bundle2");
+    const LIVE_MAIN = join(LIVE_CWD, "ComfyUI", "main.py");
+    mockResolveBase.mockReturnValue(undefined);
+    mockLiveRootFromArgv.mockReturnValue(undefined);
+    mockGetSystemStats.mockResolvedValue({
+      system: {
+        argv: [join("ComfyUI", "main.py"), "--port", "8188", "--enable-manager"],
+      },
+    });
+    mockExistsSync.mockImplementation((p: string) => {
+      const s = String(p);
+      return s === ABS_PYTHON || s === LIVE_MAIN;
+    });
+    __processControlTestHooks.setLiveCwdResolver(() => undefined);
+    mockResolveLiveServerRoot.mockReturnValue({
+      source: "observed-process",
+      anchorDir: LIVE_CWD,
+      observedPid: 4321,
+    });
+    mockLivePortThenFree();
+    mockSpawn.mockImplementation(() => new FakeChild());
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(async () => ({ ok: true }) as Response),
+    );
+    const killSpy = vi.spyOn(process, "kill").mockImplementation(() => true);
+
+    const result = await restartComfyUI();
+
+    expect(result.message).not.toMatch(/refusing to restart/i);
+    expect(mockSpawn.mock.calls[0][1][0]).toBe(LIVE_MAIN);
+
+    killSpy.mockRestore();
+  });
+
+  it("WINDOWS: an UNREADABLE creation stamp fails closed, it does not pass on the pid number (#535 codex gate)", async () => {
+    // pid + creation time is this file's standard for process identity, because
+    // pid equality across a window is exactly what pid REUSE defeats. The observed
+    // resolver re-queries the port owner AFTER the identity bracket has closed, so
+    // it needs its own continuity proof. "Could not read the stamp" is not
+    // "the stamp matched".
+    const LIVE_CWD = resolve("bundle3");
+    mockResolveBase.mockReturnValue(undefined);
+    mockLiveRootFromArgv.mockReturnValue(undefined);
+    mockGetSystemStats.mockResolvedValue({
+      system: { argv: [join("ComfyUI", "main.py"), "--port", "8188"] },
+    });
+    mockExistsSync.mockImplementation((p: string) => {
+      const s = String(p);
+      return s === ABS_PYTHON || s === join(LIVE_CWD, "ComfyUI", "main.py");
+    });
+    __processControlTestHooks.setLiveCwdResolver(() => undefined);
+    // No stamp available on this host.
+    __processControlTestHooks.setProcessIdentityResolver(() => undefined);
+    mockResolveLiveServerRoot.mockReturnValue({
+      source: "observed-process",
+      anchorDir: LIVE_CWD,
+      observedPid: 4321,
+    });
+    mockLivePortThenFree();
+    mockSpawn.mockImplementation(() => new FakeChild());
+    const killSpy = vi.spyOn(process, "kill").mockImplementation(() => true);
+
+    const result = await restartComfyUI();
+
+    expect(result.message).toMatch(/refusing to restart/i);
+    expect(mockSpawn).not.toHaveBeenCalled();
+    expect(killSpy).not.toHaveBeenCalled();
+
+    killSpy.mockRestore();
+  });
+
   it("an UNRESOLVED observation changes nothing — the pre-#535 refusal stands (#535)", async () => {
     // The tri-state discipline: "could not observe" is not "observed something
     // usable". This is the branch that must never become a success.
--- a/src/__tests__/services/restart-live-first.test.ts
+++ b/src/__tests__/services/restart-live-first.test.ts
@@ -705,6 +705,88 @@ describe("restart_comfyui — live-first script resolution (#476, #426)", () => 
     },
   );
 
+  it("WINDOWS: a SECOND value of a multi-value path flag is checked too (#535 codex round 3)", async () => {
+    // ComfyUI declares `--extra-model-paths-config` as nargs='+', so it carries
+    // one-or-more paths. Checking only the first let `two.yaml` — relative, and
+    // separator-free so the positional check waves it through — reach the anchor.
+    const LIVE_CWD = resolve("bundle-multi");
+    mockResolveBase.mockReturnValue(undefined);
+    mockLiveRootFromArgv.mockReturnValue(undefined);
+    mockGetSystemStats.mockResolvedValue({
+      system: {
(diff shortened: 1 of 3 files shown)


artokun:
## Implemented

The change is smaller than the issue's history suggests, because #1133 had already built the hard part.

`resolveLiveServerRoot`'s observed-process tier walks up from the interpreter the OS reports for the server on our port and accepts the first ancestor under which `<ancestor>/<relDir>/main.py` exists **and** the interpreter belongs to that install. That accepted ancestor **is** the working directory the server must have had for its relative `main.py` to name this install — exactly what `/proc/<pid>/cwd` states directly on Linux. It already fell out of the walk and was simply discarded.

So: return it (`anchorDir`), and let `gatherProcessInfo` fall back to it when procfs yields nothing. Everything downstream — script resolution, interpreter re-resolution from the same root, regular-file validation, spawn cwd, refuse-safe preflight — is the existing already-reviewed `liveCwd` logic, reached unchanged rather than duplicated.

### The safety property

**The PID must match.** The observation is anchored on *"whatever is listening on our port"*, while this value decides the relaunch of a process we are about to **kill**. If those are different processes, anchoring on the observation would restart the wrong tree after stopping the right one — strictly worse than the refusal it replaces. An unconfirmed observation is discarded; so is an `unresolved` one. The anchor may only ever *add* a way to succeed.

### Cost

Self-limiting, and I checked rather than assumed. `resolveLiveServerRoot` returns from argv **before** reaching the process probe:

```ts
const fromArgv = liveRootFromArgv(argv, cwd);
if (fromArgv) return { root: fromArgv, source: "argv", relDir };
```

So an absolute `argv[0]` never pays the ~1.3s netstat/WMI walk — the probe runs only for the relative-script shape, which is precisely the case that needs it. Desktop skips the whole thing (`desktop ? undefined : …`), as it must: Desktop is never killed.

### Verification

- 3 new tests, all mutation-pinned: removing the fallback turns the Windows anchor test red; removing the pid guard turns the mismatch test red. Neither passes silently.
- The `unresolved` branch has its own test — the tri-state discipline, so "could not observe" never becomes "observed something usable".
- Full suite: **8387 passed, 0 failed.** `tsc --noEmit` clean; no new unknown-collapses.

Codex review running before merge.

artokun:
## Live verification on a real Windows install

Built the branch and probed the real resolver against this rig's actual ComfyUI layout, with the reporter's argv shape:

```
REPORTER #535  (`ComfyUI\main.py`)
   source   : observed-process
   root     : C:\Users\Artokun\ComfyUI-Installs\ComfyUI\ComfyUI
   anchorDir: C:\Users\Artokun\ComfyUI-Installs\ComfyUI      <- the /proc/<pid>/cwd stand-in
   relaunch script: C:\Users\Artokun\ComfyUI-Installs\ComfyUI\ComfyUI\main.py
   …exists on disk: YES
```

That is the whole chain: the inferred working directory, the relative segments re-anchored onto it, and a script that **exists**. The refusal in this issue — *"Resolved ComfyUI script does not exist on disk: `ComfyUI\main.py`"* — becomes a successful resolution on a real install, not just in a fixture.

The bare `main.py` shape on the same machine stays `unresolved`, so its refusal is preserved:

```
THIS RIG       (bare `main.py`)
   source   : unresolved
```

That is correct and deliberate. With `relDir = "."` the walk can only land on ancestors of the interpreter, and the portable layout puts `standalone-env\python.exe` as a **sibling** of the nested `ComfyUI\main.py`, never an ancestor. Reaching a nested sibling from an interpreter anchor is a different inference with a different risk profile, on a gate that authorises killing a running server — it deserves its own change and its own review.

## On the pid guard, having traced where `pid` comes from

Worth stating because the codebase is explicit that pid-number equality is not identity: `gatherProcessInfo` does not take `pid` from a bare lookup. It brackets the `/system_stats` call with port-owner observations comparing **pid + creation time**, backed by an instance-witness WebSocket held open across the fetch, and refuses when the bracket cannot be closed.

So the number I compare against is an already-proven port owner. And the property that makes the simple comparison sufficient is that **the pid I check is the same pid that gets killed** — if it were ever reused, the anchor and the kill target would still describe the same process, so the relaunch stays self-consistent rather than landing on a wrong tree. The substitution cases the file's bracket exists for are refused upstream, before this code is reached.

artokun:
## Codex round 1 — two P1s, both real, both fixed

Disclosed: adversarial review by Codex (source-traced; its sandbox blocked local shell, so it read the pushed commits through the repository connector).

### 1. The anchor is INFERRED, not observed

`anchorRelDirOnInterpreter` accepts an ancestor because `<ancestor>/<relDir>/main.py` exists and the interpreter lives inside that install. That establishes *"a directory from which this relative script **would** name this install"* — **not** *"the directory the process was started in"*. A launcher keeping `C:\launcher\main.py` symlinked to `C:\bundle\main.py` runs from `C:\launcher` while `C:\bundle` satisfies the inference.

For the panel base (#1133) that gap is harmless — both spellings name the same install root. Here it is **not**, because the value becomes the relaunch's **working directory**, so a relative argument would resolve somewhere else after the restart.

Fixed by adopting the anchor only when the choice of cwd *cannot change what any argument means*: every token after the script must be a flag, an absolute path, or a plain value that cannot be a path. `--port 8188` still qualifies — the reporter's shape — while `--output-directory out` does not.

Conservative deliberately: a false positive costs a Windows user the manual restart they already have today; a false negative relaunches a server into a directory where its own arguments point elsewhere.

### 2. PID equality is not process identity

This file's own rule is pid **+ creation time**, *"because pid equality across a window is exactly what pid REUSE defeats"* — and the observed resolver re-queries the port owner at its own moment, **after** the identity bracket has closed. A pid recycled in that window satisfies a numeric comparison while describing a different server.

The observation is now bound to the bracketed creation stamp, and an unreadable stamp is "did not observe", never "observed the same" — it fails closed to the pre-existing refusal.

I had argued the numeric check was sufficient because the pid checked is the pid killed. The pre-kill creation-time check does block the worst outcome, but that argument does not hold for the *anchor* itself, and "narrow window on a destructive path" is not a standard this file accepts anywhere else.

## And one the gate did not find

Re-reading my own guard: it skipped every token starting with `-`, which walks straight past `--output-directory=out` — the equals form carries its value **inside** the flag token. Now unwrapped and judged on its own, with the mirror pinned too: an equals-form **absolute** value must still not block the anchor, or unwrapping would turn every `--flag=…` into a refusal.

## State

- **8391 pass.** The one red under full-suite load (`ui-bridge` #709) **passes in isolation** — load-sensitive, unrelated; the same file flaked on a different test earlier today.
- `tsc --noEmit` clean; `check:unknown-collapse` clean.
- Four guards mutation-pinned: removing the fallback, the pid guard, the relative-path guard, or the stamp guard each turns its own test red.
- Linux/procfs is untouched — `??` short-circuits before the fallback, so none of the new requirements can reach it.

Round 2 running.

artokun:
## Codex round 2 — stamp fix clean, argv guard bypassable

Confirmed clean: the creation-stamp binding, and that **Linux/procfs is unaffected** (`??` short-circuits before the fallback, so none of the new requirements can reach the existing path).

Not clean: the argv guard, three ways — `--output-directory auto`, `--output-directory 8188`, `--output-directory -`.

The cases mattered less than what they showed: **judging whether a token "looks like a path" cannot work.** `auto`, `8188` and `-` are all legal directory names, so an allowlist of "values that cannot be paths" does not exist to be written. I had even documented the `auto` case in a comment as a shape the guard "could miss" — which is a fair description of a hole, not a justification for one.

So the guard now keys on the **flag**, whose vocabulary is knowable, and treats the token after a path flag as its value whatever it looks like. `--` is tracked, so post-separator positionals are judged as positionals.

## Then I broke it myself, and the tests did not notice

The first version of that rewrite matched flag names by substring — `dir|path|cache|log|…`. That is over-broad in the direction that **un-fixes the issue**: `--cache-none` and `--log-stdout` are **boolean** flags that match it, so the guard eats the next token as their "value", finds it non-absolute, and refuses.

`--cache-none` is in this issue's **original report** — `python main.py --cache-none --output-directory …`. The heuristic would have refused the exact argv the fix exists to support, and **all 23 tests still passed**, because not one of them used a boolean flag that happened to contain a keyword.

Path flags are enumerated now, with two supporting rules: a `--`-prefixed token is never consumed as a path flag's value (no ComfyUI path flag takes one), and a flag missing from the list degrades to the positional check — blocked when its value is path-shaped, allowed when it is not.

That leaves a stated residual: an unknown path flag carrying a relative value that looks like nothing (`--future-dir auto`) on a symlinked launcher. Strictly narrower than the blanket refusal it replaces, and written down rather than discovered later.

Three tests now pin the over-broad direction specifically — a boolean flag that reads path-ish, a counted flag whose value is not a path, and a boolean flag preceding a real one.

## State

- **8399 pass**, `tsc --noEmit` clean
- Mutation-pinned: neutering the flag vocabulary turns 5 tests red, including all three round-2 bypasses
- Round 3 running against the previous commit; I will reconcile its findings against the enumeration that replaced it

artokun:
## Codex round 3 — three findings, all real, all fixed

Two of them are the same failure mode as round 2's, and it is the one worth naming.

**`--extra-model-paths-config` takes more than one path.** ComfyUI declares it `nargs='+'`, so `--extra-model-paths-config C:\one.yaml two.yaml` carries two. I checked only the first; `two.yaml` then fell to the positional check, which lets a separator-free name through. The guard now keeps consuming values until the next option, the way argparse ends the list. A genuine bypass.

**`--directml` is not path-valued** — it takes a device index. I put it in the enumerated list *by hand, in the very commit arguing that enumeration beats pattern-matching*. An enumerated list is only as good as the reading behind each entry; that entry had none.

**A URL is not a path.** `--comfy-api-base https://api.comfy.org` carries slashes, so the path-shaped check refused it. No working directory can move a URL.

## The pattern across all three rounds

Every finding has been in the **guard heuristic**, never in the core fix. And two of the three round-3 findings — like the `--cache-none` one I caught myself — are the **over-broad** direction: the guard refusing a launch it should allow.

That direction silently un-ships the fix rather than weakening it, and it is invisible to the tests one naturally writes for a guard, because those all feed it inputs it is *supposed* to block. Mutation testing does not surface it either: neutering the guard makes the block-tests fail, which reads as good coverage.

So the suite now carries explicit **allowed-through** cases: the reporter's own argv shape, a boolean flag that reads path-ish, a counted flag whose value is not a path, a device index, and a URL.

## Stopping the gate here

Three rounds, all findings fixed. Rounds 2 and 3 both paid, so the curve has not flattened — but every finding has been an edge case of a **best-effort mitigation**, not the mechanism. Worth being explicit about what that mitigation is for: `anchorDir` and the true cwd can only diverge behind a symlinked/junctioned launcher, and the two failure modes are *refuse when we could have proceeded* (the manual restart users have today) or *proceed from a slightly different cwd in that exotic layout*. Neither is the wrong-tree destruction round 1 identified, which is closed by the guard's existence rather than by its precision.

## State

- **8402 pass**, `tsc --noEmit` clean, `check:unknown-collapse` clean
- Mutation-pinned: the fallback, the pid guard, the stamp guard, the flag vocabulary, and the multi-value consumption each turn their own tests red
- Linux/procfs untouched — `??` short-circuits before any of this
- Live-verified on a real Windows install: `anchorDir` resolves and the relaunch script exists on disk

artokun:
## Correction: the CI red was HOST-dependence, not platform-dependence

The commit message on `a2f70cf` says Windows produced wording matching `/refusing to restart/i` and Linux did not. **That is wrong**, and the wrong version is the more comfortable one, so it is worth correcting rather than leaving.

It failed on **ubuntu, windows AND macOS**. It passed only on my machine. So the OS was never the variable.

The mechanism: removing the process-identity stamp makes `gatherProcessInfo` take the #914 branch —

```ts
if (before && !before.startedAt) {
  witness = await acquireInstanceWitness(getComfyUIBaseUrl());
}
```

— and `restart-live-first.test.ts` does not mock `services/instance-witness.js`, so that opens a **real WebSocket to `127.0.0.1:8188`**. My rig has a live ComfyUI there; CI has nothing. The witness closed the identity bracket locally and failed on CI, so the two took different branches.

This is #1263's defect — a test whose assertion depends on whether a ComfyUI happens to be listening — reached through a channel `live-probe-gate` (#1290) does not cover. That gate computes reachability to `resolveLiveInterpreter`; `acquireInstanceWitness` is a separate route to the same host state. Reported on #1263 with a suggested fix; **not** fixed here, because this PR does not stub it — it simply stopped needing it, since the reworked test leaves readable stamps in place so the witness branch is never entered.

The fix in `a2f70cf` stands on its own merits either way: the test now exercises the guard it is named for rather than the bracket upstream of it, and asserts the refuse-safe invariant instead of a refusal sentence.

I was lucky in the direction it broke. A test that passes on CI *because* nothing is listening, while asserting nothing on a developer rig, is the same bug pointed the other way — and nothing would have told me.

PR #1317, opened 2026-08-10
fix(panel-tools): stop prescribing a rebind that needs the tab we just said is missing

artokun:
Draft — claiming the orchestrator half of artokun/comfyui-mcp-panel#971.

The reporter's trace is a **remedy loop**:

1. `panel_set_workflow_target(mode:"current")` → reports `bound`
2. `panel_set_widget` → `[root-workflow-uuid-mismatch]`
3. the prescribed recovery, `panel_open_workflow` → *"this session has no reachable panel tab yet … Retry in a moment, **or rebind with panel_set_workflow_target({mode:"current"})**"*
4. …which is step 1, and which needs a reachable tab to do anything: the rebind's only probe is a `workflow_list` call to a tab.

## The message folds two causes with one remedy

> still reconnecting after a restart/reload, **or** multiple tabs are open and none is this session's

Those want opposite advice:

- **no tab connected at all** — there is nothing to rebind *onto*. `mode:"current"` cannot help, and offering it sends the caller round the loop above.
- **tabs connected, none is ours** — `mode:"current"` is exactly right, because it re-points the session at the active tab.

The bridge can already tell these apart (`bridge.tabs()`, minus headless), so the remedy can match the cause.

## Scope

Only the remedy. The Save-As identity half of #971 is a different fix that has since shipped panel-side (0.11.81, #941) — the reporter is on panel 0.11.51 / mcp 0.50.52, roughly 30 releases back — and I will say so on the issue rather than re-fixing it blind.

### DIFF
--- a/src/__tests__/orchestrator/no-reachable-tab-remedy.test.ts
+++ b/src/__tests__/orchestrator/no-reachable-tab-remedy.test.ts
@@ -11,12 +11,17 @@
 //   4. which is step 1 — and whose only probe is a `workflow_list` call TO A TAB,
 //      so it needs exactly what step 3 just said is missing.
 //
-// The message folded two causes that want opposite advice:
+// The message folded two causes into one remedy, and they want different advice:
 //
-//   nothing connected      -> nothing to rebind ONTO; the rebind cannot help
-//   connected, none is ours -> the rebind is precisely right
+//   nothing connected       -> a rebind DEFERS (it clears the stale binding and
+//                              takes effect on the first tab that reconnects), so
+//                              it does not make this retryable now
+//   connected, none is ours -> a rebind is precisely right: it re-points the
+//                              session at the live tab
 //
-// The bridge can tell them apart, so the remedy now matches the cause.
+// The bridge can tell them apart, so the remedy now matches the cause. What the
+// zero-tab branch says was MEASURED against the real handler, not reasoned — see
+// the first test.
 
 import { describe, expect, it } from "vitest";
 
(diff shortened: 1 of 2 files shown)


artokun:
## Codex wedged — so I checked my own load-bearing claim, and it was wrong

The review timed out with no output (ten minutes, the wedge shape I have hit before). Rather than merge on an unreviewed claim, I probed the one this PR rests on.

**"Rebinding CANNOT help while nothing is connected" is false.** Driving the real handler with an empty bridge, `panel_set_workflow_target({mode:"current"})` with zero tabs returns:

```json
{ "mode": "current", "deferred": true,
  "note": "…No panel tab is connected yet — cleared the stale binding; this session
   will follow (bind onto) the tab as soon as one reconnects…" }
```

So it is genuinely useful there. What it does **not** do is make the failed command retryable now — which is what the caller is asking. My version would have replaced one wrong remedy with another, and confidently.

The zero-tab branch now says the rebind **defers** rather than repairs, and names what actually brings a tab back. A test asserts the false claim cannot return.

**What survives:** the two causes really do want different advice, and the bridge can tell them apart —
- nothing connected → the rebind defers; wait for the reconnect or have the user refresh the tab;
- tabs connected, none ours → the rebind is exactly right.

Headless tabs are excluded (a headless tab is not a panel anyone is looking at, and counting one produces "rebind onto the live one" when there is no live one). An unreadable bridge keeps the old cause-neutral wording rather than guessing. "Nothing was sent" survives on every branch and is asserted — the pre-send wait exists so a mutating command is never fired into a dead binding.

Both branches mutation-verified. Full suite: 427 files, 8392 tests, lint clean.

PR #1321, opened 2026-08-10
fix(#1077): localhost and 127.0.0.1 are the same origin for the workflow fence

artokun:
Claiming the Finding-2 defect I demonstrated on #1077 by measurement and then did **not** fix:

```
localhost -> {"origin":"http://localhost:8188","uuid":"1111…5555"}
127.0.0.1 -> {"origin":"http://127.0.0.1:8188","uuid":"1111…5555"}
SAME ORIGIN? false
```

Still true on `main` today. `workflowIdentityParts` (`src/orchestrator/session-store.ts:70-81`) normalises an origin with exactly three operations — `trim()`, strip trailing slashes, `toLowerCase()` — and **no host equivalence**. So an identity bound under one spelling can never validate against a connection carrying the other, and the fence refuses a case it should accept.

I said at the time this was *"a demonstrated defect in its own right, whether or not it is the reporter's Finding 2"*. It is worth fixing on that basis alone, and I am not claiming it closes this issue.

## Why it is permanent when it happens

Nothing re-binds the stored origin, so no refresh, tab close, or reconnect can clear it — which matches the refresh-proof shape the reporter described, even though I still cannot show their session mixed spellings.

This is also a live hazard on this rig specifically: the panel debugging path uses `localhost:8188` (the extension is blocked on `127.0.0.1` by policy) while `COMFYUI_URL` is `127.0.0.1:8188`. Two spellings of one server is the normal case here, not an exotic one.

## Plan

- [ ] Establish exactly which spellings must be equivalent — loopback only (`localhost`, `127.0.0.1`, `::1`, `127.x.x.x`), never a general DNS resolution
- [ ] Normalise at `workflowIdentityParts`, so both the bind and the compare go through one function rather than two that can drift
- [ ] Keep scheme and port significant — `http://localhost:8188` and `http://localhost:8189` are different servers, and `https` vs `http` is a different origin by every other rule in this codebase
- [ ] Pin the negative direction too: a non-loopback host must NOT be folded into anything else
- [ ] Check the other consumers of the same origin string for the identical assumption before fixing only this one

Draft until the equivalence set is established rather than assumed. The failure mode of getting this wrong is two genuinely different servers sharing a fence.


artokun:
## Investigated before changing anything — and the stated fix would be a defect

The equivalence I proposed in the plan (`localhost` ≡ `127.0.0.1`) **must not be implemented**. This codebase already reasoned about exactly that and ruled against it, as a coordinator **P0**, in #1233:

> `localhost` returns null ON PURPOSE (coordinator P0): a URL preserves the literal "localhost" and does **NOT** reveal whether the browser actually reached 127.0.0.1 or ::1 — so PINNING it to a family we can't verify could send the auth-bearing probe to a DIFFERENT-family instance…

…and the families are deliberately kept apart:

> The families are kept DISTINCT so a v4 tab and a v6 instance at the same port are NOT wrongly matched (coordinator finding 4: v6 A on `[::1]:8188` + v4 B on `127.0.0.1:8188` are DIFFERENT instances).

So folding `localhost` into a concrete literal could bind one fence across two genuinely different ComfyUI instances — the precise risk I flagged in the plan, already established as real rather than theoretical. The measurement I posted on #1077 is still correct; the conclusion I was heading for was not.

## What is actually wrong, and it is narrower

`httpOriginOf` (`panel-tools.ts:906`) canonicalises an origin through `loopbackFamily`, folding **same-family** loopback: `127.0.0.1` ≡ `0.0.0.0`, `::1` ≡ `::`, with `localhost` left distinct.

`workflowIdentityParts` (`session-store.ts:70-81`) does `trim()` + trailing-slash strip + `toLowerCase()` and nothing else.

**Two functions canonicalise origins for the same machine by different rules.** A server bound `0.0.0.0` and reached as `127.0.0.1` is one instance by every other rule in the codebase and two identities to the fence. That is a genuine defect, it is safe to fix, and it does not touch the `localhost` ambiguity at all.

## Re-scoped plan

- [ ] Route `workflowIdentityParts` through the same canonicalisation `httpOriginOf` uses, so same-family loopback folds and the two functions can no longer drift
- [ ] Leave `localhost` deliberately distinct, with the #1233 reasoning cited at the site so the next person does not re-derive it
- [ ] Make the refusal **name both spellings** — today it says "one bound to a different origin" without showing them, which is what makes a `localhost`-vs-literal mismatch look like a mystery instead of "use one spelling consistently"
- [ ] Pin the negative direction: v4 vs v6, and any non-loopback host, must NOT fold

The honest summary for #1077: this removes one real way to hit that symptom and makes the remaining one diagnosable. It does not make `localhost` and `127.0.0.1` interchangeable, and it should not.

artokun:
Closing without a fix: **there is no origin comparison to fix**, and #1255 established that before I opened this.

`fence-refusal.ts` says it in as many words:

> this used to end "a malformed uuid, or one bound to a different origin", which describes a comparison the code does not perform. `workflowIdentityParts` uses the origin as a PRESENCE gate: it requires one to exist, and never compares it against a previously bound value (**nothing in the repo has ever read `identity.origin` — both call sites take `.uuid`**).

I verified that rather than take it: `index.ts:2845` and `index.ts:3494` are the only callers, and both use `identity.uuid` only. The origin is never persisted and never compared, so two spellings of the same loopback host cannot produce a fence refusal through this path. My measurement on #1077 — that `canonicalOrigin` distinguishes `localhost` from `127.0.0.1` — was true about the *function's output* and irrelevant to the *fence*, because nothing downstream compares two origins.

The same comment records what this cost the last time:

> The old wording sent a reader hunting for an origin mismatch that cannot happen — it cost a whole investigation and a closed-unmerged PR before the mechanism was checked rather than believed.

I was reproducing that investigation, from the same stale wording preserved in my own #1077 comment. #1255 fixed the message; it could not fix the copy of it sitting in an issue thread.

## Two things worth keeping from this

**`src/utils/origin.ts` names the fence as a motivating case and is not wired to it.** Its header says *"The same blindness scopes the workflow-instance fence, which is keyed by origin: a fence adopted under one spelling would refuse under the other."* That sentence is now false — the fence is not keyed by origin in any comparing sense — and `canonicalOrigin` has exactly one consumer, `comfyui/fetch.ts`. Worth correcting the header so the next reader does not infer a wiring gap that should not be filled.

**`localhost` collapsing is deliberate here and deliberately refused elsewhere.** `utils/origin.ts` folds `localhost` ≡ `127.0.0.1` ≡ `[::1]`; `panel-tools.ts`'s `loopbackFamily` refuses to, as a coordinator P0 (#1233). Both are right, because they answer different questions — "are these the same server, for messaging" versus "where do I send an auth-bearing probe". I briefly concluded the second governed the first and re-scoped this PR on that basis; it does not.

No code changed on this branch beyond the empty claim commit.

PR #1322, opened 2026-08-10
fix(#971): the prescribed recovery from a lost tab binding is a dead end

artokun:
Draft — reproducing test is in, fix in progress.

Tracks **artokun/comfyui-mcp-panel#971**.

The reporter's wedge, reproduced at the orchestrator seam and scoped to one state:

| bridge's last-active tab | `set_workflow_target({mode:"current"})` | result |
|---|---|---|
| known | rebinds `stale-tab` → `tab-b` | `panel_open_workflow` succeeds — design works |
| **unknown** | errors *"…none is last active — pass tab_id"* | **closed loop** |

In the broken state `panel_open_workflow` refuses and names `panel_set_workflow_target({mode:"current"})` as the way out; that tool fails demanding `tab_id`; and its schema is `["mode","path","filename"]` — no `tab_id`, with #754 strict schemas turning an unknown key into a hard validation error. The agent is told to pass something it cannot pass.

`panel_reload` already words this correctly (*"Switch to the tab you want, then call panel_set_workflow_target…"*) — focusing a tab is what makes it last-active. The raw bridge error leaks through `set_workflow_target` without that framing.

Note the reporter is on 0.50.52 / panel 0.11.51; the refusal text they quoted has already been improved since. This dead end still reproduces on **0.50.91**.

### DIFF
--- a/src/__tests__/orchestrator/panel-971-recovery-deadend.test.ts
+++ b/src/__tests__/orchestrator/panel-971-recovery-deadend.test.ts
@@ -109,7 +109,7 @@ it("#971 with NO last-active tab, the prescribed recovery is a dead end", async 
   // And whatever it says, it must leave the caller somewhere other than where it
   // started: either rebound, or told an action that does not require a parameter
   // this tool does not have.
-  const recovered = ctx.tabId !== "stale-tab" || /switch to|focus/i.test(textOf(rec2));
+  const recovered = ctx.tabId !== "stale-tab" || /sent from its Agent panel/i.test(textOf(rec2));
   expect(recovered).toBe(true);
 });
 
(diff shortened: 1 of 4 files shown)


artokun:
Heads-up so you do not redo work: the orchestrator half of #971 merged an hour ago as #1317 (shipped in **0.50.92**), and `main` has it.

What is already in: `noReachableTabFail` no longer gives one remedy for two causes. It counts non-headless connected tabs and branches —

- **zero connected** → says the rebind DEFERS (clears the stale binding, takes effect on the first tab that reconnects) rather than repairs, and names what actually brings a tab back;
- **connected but none is ours** → offers `mode:"current"`, which is correct there;
- **bridge cannot enumerate** → keeps the old cause-neutral wording.

Your table's *"bridge's last-active tab: unknown"* row looks like a **third** state mine does not separate — tabs connected, but `resolveActiveTabId` throws, so `mode:"current"` errors with *"none is last-active"*. My change would give that case the "rebind onto the live one" advice, and if the rebind then errors, that is still a dead end. So there is real work left and it is not a duplicate.

One thing worth knowing before you build on it, since it cost me a wrong fix: with **zero** tabs, `mode:"current"` does **not** fail — it returns `deferred: true` and clears the stale binding. I asserted the opposite in my first draft and had to correct it after probing the real handler. Worth checking what it actually returns in your unknown-last-active state rather than assuming it errors.

PR #1333, opened 2026-08-10
fix(#1318): tell the user the install is damaged, instead of printing a resolver path

### LINKED ISSUE #1318: Global install left half-extracted with the version advanced — @modelcontextprotocol/sdk missing all of dist/esm/server/, CLI dead with ERR_MODULE_NOT_FOUND
### What happened

A global Windows install of `comfyui-mcp` became completely unusable. `npx -y comfyui-mcp connect` failed with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
'...\node_modules\@modelcontextprotocol\sdk\dist\esm\server\mcp.js'
```

### Root cause: a half-extracted dependency, not a packaging bug

The dependency is declared correctly — `"@modelcontextprotocol/sdk": "~1.29.0"` — so the published package is fine. The problem was purely in the installed tree:

- `@modelcontextprotocol/sdk/dist/esm/server/` contained **only** the `auth/` and `middleware/` subdirectories and **zero files** (a healthy install has 42 entries).
- `dist/cjs/server/mcp.js` was also missing.
- Directories created, files never written — the signature of an interrupted extraction rather than a corrupt tarball.

Nothing held the files open: all ~70 running node processes were from a separate dev checkout (`...\code\comfyui-mcp\dist\index.js`), not from the global install.

### The part that matters

The global install's `package.json` read **0.50.46** — so a version had been recorded while the dependency tree was broken, and **nothing surfaced any error**. The install presented as up to date and was in fact non-functional. The only symptom a user gets is a raw Node module-resolution stack trace that never mentions `comfyui-mcp` or suggests a repair.

### Why this is not #912

#912 ("self_update can never succeed on Windows (EBUSY on its own sharp libvips-42.dll) and swallows the npm error") was fixed in PR #924 and closed 2026-08-07. I verified with `git merge-base --is-ancestor b5019f7 v0.50.46` that **v0.50.46 contains that fix**. #912's failure mode is "update fails, stays on the old version, install stays intact." This is the opposite: the version advanced and the tree was left broken.

### Causation is NOT established

I have not proven `self_update` caused this. The evidence is circumstantial — Windows, the version advanced, the tree broken — and an ordinary interrupted `npm install -g` would look identical. Filing this for the **detection gap**, which holds regardless of cause: a partially-extracted install is currently silent until an unrelated-looking module error at runtime.

### Possible direction (not prescriptive)

A startup or `doctor`-style self-check that resolves the handful of critical deps and, on failure, prints the repair command instead of letting Node's resolver surface an opaque path.

### Repro / repair

Repair was simply:

```
npm install -g comfyui-mcp@latest
```

0.50.46 -> 0.50.67; `mcp.js` restored (40694 bytes); `dist/esm/server/` back to 42 entries; `npx -y comfyui-mcp --version` reaches "ComfyUI MCP server running on stdio" with exit 0.

### Environment

- Windows 11 Pro 26200, global npm install
- Broken at comfyui-mcp 0.50.46, repaired to 0.50.67

artokun:
Closes #1318

## What the reporter saw

A global Windows install became unusable. `@modelcontextprotocol/sdk`'s `dist/esm/server/` held two empty subdirectories where a healthy install has 42 entries — directories created, files never written. The install's `package.json` still read `0.50.46`, so it presented as up to date while being non-functional.

What surfaced was Node's resolver talking about a path:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module
'…\node_modules\@modelcontextprotocol\sdk\dist\esm\server\mcp.js'
```

Nothing there names comfyui-mcp, says the install is damaged, or hints at the one-line repair.

## Scope — this is the detection gap only

The issue explicitly scopes itself to detection and states the **cause is unproven**: an interrupted `npm install -g` and a failed self-update leave identical evidence. So this diagnoses the **observable state** and does not assert how it got there. Claiming a cause here would send someone to fix the wrong thing.

## Why `src/index.ts` had to move

A static import of a missing dependency throws **before any of our code runs** — no handler we install inside the old entry point could ever have fired. So `src/index.ts` becomes a thin launcher that `await import("./boot.js")` inside a try/catch, and `src/boot.ts` is the (unchanged) real entry point via `git mv`.

Keeping the launcher free of package imports is load-bearing: anything imported statically there is something that can fail before the handler exists.

`bin`/`main` stay `dist/index.js` — the README documents that path in user MCP configs.

## What is deliberately narrow

Only `ERR_MODULE_NOT_FOUND` / `MODULE_NOT_FOUND` are translated. Any other error is rethrown unchanged; swallowing a real crash behind a "reinstall" banner would be its own defect, and a worse one. Exit code stays 1.


artokun:
## Review pass — one real finding, fixed in 3f48ac6/HEAD

Codex could not run here (`CreateProcessAsUserW failed: Access is denied` — the sandbox denied every process launch, including `git diff`), so this was a **disclosed self-review**, with each claim measured against the built launcher rather than read off the source.

### The finding (would have been a P0)

`await import("./boot.js")` resolves when boot's **module body** finishes — not when the server stops. The question is therefore what else can land in that catch.

This repo lazily imports several **optional** dependencies, which throw `ERR_MODULE_NOT_FOUND` when legitimately absent: `@aws-sdk/client-s3`, `@azure/storage-blob`, `@ai-sdk/*`, `@anthropic-ai/claude-agent-sdk`. If any of those reached the launcher, a working install would be told to reinstall itself.

It doesn't — because `src/boot.ts:450` is `main().catch(...)`, not `await main()`. Measured both ways against the built launcher:

```
main().catch(...)   ->  BOOT's OWN handler ran, code = ERR_MODULE_NOT_FOUND
await main()        ->  comfyui-mcp could not start: its dependency
                        "@azure/storage-blob" is not installed     <-- the bogus banner
```

So the current shape is correct, but the launcher's error scope **depends on a property of another file that it cannot see**. A source guard now fails if that line changes; mutating it to `await main()` kills the test.

### Second defect, found by running it

The reproduced install was genuinely missing `zod`, and the message read `a file from its dependency "zod" is missing:` followed by a path line containing just `zod`. Node distinguishes `Cannot find package` from `Cannot find module`; the message now does too, and an absent package reads `its dependency "zod" is not installed`.

### Verified against a reproduced install

Isolated scratch install with `@modelcontextprotocol/sdk/dist/esm/server/` as an empty directory — the reporter's exact on-disk state (0 entries where a healthy install has 42). The shared checkout's `node_modules` was not touched.

| case | result |
|---|---|
| half-extracted dependency (#1318's state) | names the package + path, exit 1 |
| package absent entirely | `its dependency "zod" is not installed`, exit 1 |
| a real `TypeError` from boot | passes through unchanged, stack intact, exit 1 |
| healthy install | launcher is transparent, exit 0 |
| optional dep missing inside `main()` | boot's own handler, **no** banner |

### Cleared

- **Top-level await / shebang** — `dist/index.js` keeps `#!/usr/bin/env node` and runs.
- **Packaging** — `dist/boot.js` is emitted (21 KB) and inside `dist`, which is already in `files`. `bin`/`main` unchanged.
- **Entry assumptions** — boot's only argv use is `parseCliArgs(process.argv)` with `argv.slice(2)`; `argv[1]` is still `dist/index.js`. No `require.main` / `import.meta.main`.
- **Stale references** — two comments named `src/index.ts` for code now in `src/boot.ts`; repointed. `"dev": "tsx src/index.ts"` is still correct.

### Known limit

`missingSpecifier`/`missingKind` parse Node's English error strings. If a future Node reworded them, both return `null` and the message degrades to the generic "a module it needs could not be loaded" plus the same repair — never a wrong claim. Pinned by a test.

Full suite: **428 files, 8426 passed, 2 skipped**. `tsc --noEmit` clean.

PR #1336, opened 2026-08-10
fix(#1325): ui-bridge.test.ts is load-sensitive — find the real rejection, don't loosen the assertion

### LINKED ISSUE #1325: ui-bridge.test.ts is load-sensitive: three different tests failed under full-suite parallelism in one session, all pass isolated
Recurring flake observed repeatedly while working #1133 and #535 today. Filing because a flaky file in the suite's trust path is worse than a red one — it trains people to re-run rather than read.

## What happens

`src/__tests__/services/ui-bridge.test.ts` fails under a full `npx vitest run`, and passes 206/206 when run with `--no-file-parallelism`. **Three different tests** failed across one session on the same machine:

- `defaultBridgeTimeoutMs — tolerant read timeout (#357) > a no-timeout READ stays alive PAST the old 6s cutoff (end-to-end)`
- `UiBridge — desktop-tab mirror (multi-viewer fanout) > #709: the NO-IDENTITY refusal is NOT capability-marked`
- `UiBridge — desktop-tab mirror (multi-viewer fanout) > #819/#823: an unadvertised panel version is reported as unobserved, not as old`

Each failed once and passed on isolated re-run. No code in this file was being modified.

## Why it looks like load sensitivity rather than a product race

The first is explicitly a timing test (it asserts a read survives past a 6s cutoff). Under a fully parallel suite the event loop is contended enough that wall-clock assertions and any `setTimeout`-driven handshake can miss their window. The other two are in the fanout describe block, which drives multiple simulated viewers through the bridge.

I have not proven the mechanism — this is an observation with a reproduction condition (full parallel suite), not a diagnosis. What is solid: isolated runs are green every time, and the failures move around rather than sticking to one assertion.

## Why it is worth fixing rather than tolerating

CI is green because the runners are less contended than this workstation, so this is invisible there. The direction that matters is the other one: a timing test that passes because it got scheduled promptly is not testing its timeout, and nothing reports that. Same family as #1263 — an assertion whose outcome depends on the machine rather than the code.

## Suggested direction

- Give the `#357` timeout test fake timers or a generous tolerance rather than real wall-clock, so contention cannot decide it.
- Check whether the fanout tests share module-level bridge state across cases in the same file (the failures moving between cases hints at ordering, not at one bad assertion).
- If the file is genuinely order-dependent, `describe.sequential` on the fanout block is cheaper than chasing each case.

## Environment

- Windows 11, Node 24, vitest 4.1.8
- Reproduces on a full `npx vitest run`; never on `--no-file-parallelism`

artokun:
Closes #1325

Claiming. Draft while I establish the mechanism.

The reporter was explicit that the mechanism is **unproven**, and the first reading already contradicts the obvious theory:

```js
void bridge.send({ cmd: "graph_query" }).then(() => settled = true, () => settled = true);
await new Promise((r) => setTimeout(r, 6500));
expect(settled).toBe(false);
```

`BRIDGE_READ_DEFAULT_TIMEOUT_MS` is **20s**, and the test's own budget is **12s**. So the read cannot reject by its timeout inside this test — if `settled` is true, the promise rejected for some *other* reason, and both handlers discard it.

That makes "give it fake timers or a generous tolerance" the wrong first move: a tolerance would hide whichever real rejection is firing. Find the cause first; widen only if the cause turns out to be scheduling.

### DIFF
--- a/src/__tests__/services/ui-bridge.test.ts
+++ b/src/__tests__/services/ui-bridge.test.ts
@@ -1886,10 +1886,26 @@ function nextFrame(
  * 60ms, more only when the loop is already behind. It therefore cannot change the
  * outcome of a test that passes today; it can only stop a loaded machine from deciding
  * one. (A shorter wait could have masked a real regression; this direction cannot.)
+ *
+ * WHAT IT DOES AND DOES NOT COVER, measured rather than assumed:
+ *
+ *   idle                       old 63ms   new 63ms   (no cost added)
+ *   in-process loop congestion old 121ms  new 367ms  (adapts — a busy worker)
+ *   other PROCESSES burning CPU old 62ms  new 63ms   (does NOT adapt)
+ *
+ * The last row is the honest limit. Cross-process starvation does not congest THIS
+ * loop, so there is nothing local to measure; a wall-clock sleep and the work it waits
+ * for are descheduled together. So this is not a proven cure for the reported
+ * condition — it removes one real load-sensitivity vector (a busy worker's own queue)
+ * without claiming to remove them all.
  */
 const settle = async (): Promise<void> => {
+  // Lag is measured with setImmediate, NOT setTimeout(0): Windows' timer resolution is
+  // ~15.6ms, so a setTimeout(0) probe reports that floor as if it were contention and
+  // inflates every one of the 31 waits on an idle machine (measured: 63ms -> 140ms).
+  // setImmediate has no such floor — 0ms idle, and it still grows when the loop is busy.
   const t0 = Date.now();
-  await new Promise((r) => setTimeout(r, 0));
+  await new Promise((r) => setImmediate(r));
   const lag = Date.now() - t0; // ~0 idle, tens of ms under a loaded suite
   await new Promise((r) => setTimeout(r, 60 + lag * 4));
   // Drain work those timers queued behind them, which a single sleep never sees.
--- a/src/__tests__/services/ui-bridge.test.ts
+++ b/src/__tests__/services/ui-bridge.test.ts
@@ -1873,7 +1873,28 @@ function nextFrame(
   });
 }
 
-const settle = () => new Promise((r) => setTimeout(r, 60));
+/**
+ * Let queued bridge work flush before asserting on its effect.
+ *
+ * #1325 — this was `setTimeout(60)`: a wall-clock guess that 60ms is "enough". It is,
+ * on an idle machine. Under a fully parallel suite the event loop is contended and
+ * 60ms can elapse with the queued work never scheduled — which is exactly the shape
+ * reported, a file that fails under `vitest run` and passes 206/206 in isolation.
+ *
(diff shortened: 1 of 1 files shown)


artokun:
## Could not reproduce — so this fixes what is provable, and says what is not

Three full `npx vitest run` passes, two of them under 8–12× CPU contention: **8426 passed, 2 skipped, 0 failed** every time. I never saw the failure. Loosening an assertion to chase a failure I have not observed would be the wrong move, so this ships two defects found by arithmetic instead.

### 1. The `#357` test could not report its own cause

It recorded a **boolean** and discarded the rejection, so a failure read `expected true to be false` and named nothing.

And the numbers say a slow machine was never the whole story. `BRIDGE_READ_DEFAULT_TIMEOUT_MS` is **20s**; the test's budget is **12s**. The deadline under test cannot be reached inside it — so if that read settles, **something else rejected it**, and that reason is the only useful thing in the failure.

It now names the reason and separates a stall from a regression. Verified both ways by lowering the constant to 6s:

```
Received: "rejected after 6503ms: Panel tab tab-aaaa-1111 did not reply to
          "graph_query" within 6000 ms — the ComfyUI tab may be backgrounded or frozen"
```

The `deadline >= SLEEP_MS` clause is load-bearing: without it that regression would have been reported as *"the machine stalled"* — the exact misdiagnosis this issue is about.

### 2. `settle()` was a wall-clock guess, at 31 call sites

`setTimeout(60)` — "60ms is enough", which is true on an idle machine. Now scales with measured event-loop lag. Safe to apply to all 31 at once because it is **monotone**: never shorter than 60ms, longer only when the loop is already behind, so it cannot change the outcome of a test that passes today.

**Measured, including the row that limits the claim:**

| condition | old | new |
|---|---|---|
| idle | 63 ms | 63 ms |
| in-process loop congestion | 121 ms | 367 ms |
| other **processes** burning CPU | 62 ms | 63 ms |

The last row is the honest limit, and it corrects my own first attempt. Cross-process starvation does not congest *this* loop, so there is nothing local to measure — and parallel vitest workers **are** separate processes. This removes one real vector (a busy worker's own queue); it is **not** a proven cure for the reported condition, and the comment says so.

I also switched the lag probe from `setTimeout(0)` to `setImmediate`: Windows' ~15.6 ms timer resolution made the first probe report that floor as contention, inflating all 31 waits on an idle machine from 63 ms to 140 ms.

### Two suggested directions, settled

- **`describe.sequential` on the fanout block** — would be a no-op. The file has zero `.concurrent`, so its tests already run sequentially; vitest parallelises *files*, not tests within one.
- **Shared module-level bridge state** — not the mechanism. `beforeEach` builds a **fresh bridge on a fresh OS-assigned port** per test, and the port race was already hardened by an earlier fix.

So if this recurs, it is none of those three, and the `#357` failure will now arrive with its cause attached.

Full suite green with these changes: 428 files, 8426 passed.

Issue #553, opened 2026-07-31
download_model cannot offer automatic Manager v3→v4 recovery


artokun:
### What happened / root cause
On a local ComfyUI with legacy Manager 3.x, eight valid download_model calls all fail with HTTP 500 because arbitrary-URL installs require Manager v4+. The tool detects this accurately, but a full requested setup is blocked and requires a separate manual Manager migration path.

### Steps to reproduce
1. Run ComfyUI with Manager 3.x legacy custom_nodes/ComfyUI-Manager.
2. Call download_model with any direct Hugging Face resolve URL.
3. Observe MODEL_ERROR / 500 on /manager/queue/install_model.

### Exact error (scrubbed)
Arbitrary-URL model installs REQUIRE Manager v4+ (3.x only accepts whitelisted catalog models).

### Fix
Applied locally: no. Upstream-only: yes. Suggested improvement: expose a safe panel/tool action that migrates legacy Manager 3.x to v4 (install comfyui_manager package, disable legacy clone, restart), or allow download_model to present/perform that guided recovery while preserving download-tray progress.

### Environment
Linux; NVIDIA RTX 5090 31 GB; ComfyUI 0.29.0; torch 2.13.0 / CUDA 13.0; comfyui-mcp 0.48.21; panel 0.11.20; Manager 3.x legacy.

PR #1338, opened 2026-08-10
fix(#1326): refuse an arbitrary-URL download once, before three requests fail

### LINKED ISSUE #1326: Preflight arbitrary URL model downloads on legacy ComfyUI-Manager 3.x
download_model action:"download" cannot dispatch public Hugging Face direct URLs when connected to a remote ComfyUI running legacy ComfyUI-Manager 3.x. All three Krea 2 files failed before transfer with HTTP 500 from /manager/queue/install_model. The tool indicates arbitrary URL downloads require Manager v4+.
### Steps to reproduce
1. Connect comfyui-mcp to ComfyUI 0.30.0 with legacy Manager 3.x.
2. Call download_model {action:"download", url:<public Hugging Face resolve URL>, target_subfolder:"vae"}.
### Exact error (scrubbed)
MODEL_ERROR: ComfyUI-Manager API 500 Internal Server Error for /manager/queue/install_model. Arbitrary-URL model installs REQUIRE Manager v4+ (3.x only accepts whitelisted catalog models).
### Fix
Applied locally: no. Upstream-only: yes. A clearer preflight/capability check before attempting the Manager 3.x endpoint, with an actionable upgrade path or supported fallback, would avoid three failed requests.
### Environment
Remote Linux; RTX 4090 24 GB; ComfyUI 0.30.0; ComfyUI-Manager 3.x legacy; comfyui-mcp 0.50.93; panel 0.11.87.

Related (same root cause): #553

artokun:
Closes #1326

Claiming. The refusal text is already correct — `Arbitrary-URL model installs REQUIRE Manager v4+` — it just arrives *after* the request has been dispatched and 500'd, once per file.

Investigating whether the Manager version is knowable before dispatch, and whether a genuine fallback exists for a REMOTE host (where we cannot write the file ourselves).

### DIFF
--- a/src/__tests__/services/manager-install-model-cause.test.ts
+++ b/src/__tests__/services/manager-install-model-cause.test.ts
@@ -96,18 +96,71 @@ describe("an install-model failure only claims a cause it can support (#1326)", 
     expect(msg).toContain("No space left on device");
   });
 
-  it("never echoes a credential from the request URL into the message", () => {
-    // A model URL routinely carries a token in its query. The body can echo the URL
-    // back, and an error message is the one place a secret must never surface.
-    const secret = "hf_SUPERSECRETTOKENVALUE123";
-    const url = `http://pod:8188/manager/queue/install_model?token=${secret}`;
+  it("never echoes the MODEL url's token — the path an endpoint-only redaction misses", () => {
+    // THE ONE THAT NEARLY SHIPPED A LEAK. My first version redacted only the query
+    // values of `details.url`, and my first test fed the secret in through exactly
+    // that — so it passed while proving nothing.
+    //
+    // `details.url` is the Manager ENDPOINT (…/manager/queue/install_model). The model
+    // url, which is the one carrying a Hugging Face / CivitAI token, is sent in the
+    // POST BODY (installModelViaManager -> taskParams.url). Manager echoes it back on a
+    // failed download, so the secret arrives by a route the endpoint url never covers.
+    // The secret is deliberately OPAQUE — no `hf_`/`sk-` prefix. An earlier version of
+    // this test used `hf_…`, which the token-SHAPE fallback catches, so it passed with
+    // the url redaction mutated away: it proved the fallback, not the thing it names.
+    // Verified by deleting the body-url redaction — this must fail.
+    const secret = "9f3a71c05b2e4d8899ab13ff5027cc41";
     const msg = annotateLegacyError(
-      managerHttpError(500, `failed fetching https://host/x.safetensors?token=${secret}`, url),
+      managerHttpError(
+        500,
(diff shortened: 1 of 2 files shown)


artokun:
## Review pass — one finding that would have leaked a token

Codex could not run again (`CreateProcessAsUserW failed: Access is denied` — the sandbox denies all process launch, including `git diff`), so this is a **disclosed self-review**, measured rather than read.

### The finding (P0 — introduced by this PR)

Surfacing Manager's response body is the whole point of the fix, and my first version redacted only the query values of `details.url`.

`details.url` is the Manager **endpoint** (`…/manager/queue/install_model`). The **model** url — the one carrying a Hugging Face / CivitAI token — is sent in the **POST body** (`node-management.ts:4328`), and Manager echoes it back on a failed download:

```
failed fetching https://civitai.com/api/download/models/123?token=<SECRET>
```

That arrives by a route an endpoint-only redaction never sees. I would have printed the token into an error message.

Now every url found in the text has its query **values** redacted regardless of origin, plus token shapes that travel outside a query string (`hf_`/`sk-`/`ghp_`, `Bearer`). Keys are kept and values under 12 chars are left alone, so `type=vae` and `page=2` stay readable — over-redaction would destroy the evidence this PR exists to surface.

### My first test for it proved nothing

It used an `hf_…` secret — which the token-**shape** fallback catches. Deleting the url redaction left it green. Re-run with an **opaque** credential (a CivitAI hex token) it fails as it should. Both directions verified by mutation, with the mutation confirmed applied via `grep -c` rather than assumed (an earlier `perl` mutation silently didn't match, and the "passing" run meant nothing).

### Two more from the same pass

- **`"proven"` is now unreachable from an HTML body.** It is the only branch that speaks with certainty, and a proxy error or framework traceback can contain the words "model list" without being Manager's rejection.
- **403 alone no longer means "excluded"** — it now requires the `security_level` marker. A proxy or auth gate in front of ComfyUI answers 403 too, and that is not Manager speaking.

### A correction the existing tests forced

My first attempt hedged the diagnosis away entirely — *"this did NOT say why … do not migrate Manager on the strength of this line alone"*. The pre-existing **#553** test failed, and it was right to: on 3.x the whitelist **is** the usual cause and that migration **is** the fix. Burying it traded a probably-right answer for a useless one. False certainty and the discarded evidence were the defects; the precision was not. The undetermined branch now leads with the whitelist and keeps the full upgrade path, and only the certainty is gone.

### Why not the blanket preflight the report asked for

Manager 3.x **does** accept an install whose URL is in its model-list catalog — `legacyTaskRequest` sends the same body and 3.x validates it. Refusing every `install-model` on 3.x would break the case that works today. That is the over-broad direction of a blocking guard: invisible in its own output, and it silently un-ships a working feature.

Full suite: 430 files, 8442 passed, 2 skipped. `tsc --noEmit` clean.

PR #1344, opened 2026-08-10
fix(#1330): corroborate a fence mismatch before failing 14 calls closed

### LINKED ISSUE #1330: Transient workflow instance mismatch rejects all connections in new workflow
While building a brand-new empty workflow, panel_add_node and panel_set_widget succeeded for 12 nodes. Immediately afterward every panel_connect call failed closed with a workflow instance mismatch. The command fence named instance 3282b824-5890-4c87-9da4-7936c155aa8c while the active canvas reportedly returned cf8f83cd-ce65-4307-ad9c-0dc5f5d491d4, even though the user had not intentionally switched tabs during the build.

Calling panel_set_workflow_target({mode:"current"}) immediately afterward reported graph_binding:"bound", graph_binding_status:"already_current", and said the active instance was already 3282b824-5890-4c87-9da4-7936c155aa8c. panel_graph_outline then showed the same 12 nodes. Retrying panel_connect succeeded.

This indicates a transient/contradictory workflow-instance identity during a new unsaved workflow build, likely an auto-save/instance replacement race.

### Steps to reproduce
1. Open a new Untitled workflow.
2. Add multiple nodes and set their widgets through panel tools.
3. Connect them immediately afterward.
4. Observe all panel_connect calls rejected with a mismatch, despite no intentional tab switch.
5. Call panel_set_workflow_target(mode:"current") and retry; connections succeed.

### Exact error
workflow instance mismatch: command issued for 3282b824-5890-4c87-9da4-7936c155aa8c, active canvas reports cf8f83cd-ce65-4307-ad9c-0dc5f5d491d4. Nothing was applied.

### Fix
Suggested minimal change: when mode is current and a mismatch occurs, re-read the live active workflow identity and reconcile it with the session fence once before returning the error. If the tab/path and graph identity are unchanged, refresh the fence and retry the idempotent mutation once; otherwise preserve the current fail-closed behavior. At minimum, include whether an auto-rebind was attempted so agents do not have to make 14 doomed calls.

### Verification
The documented recovery was verified: panel_set_workflow_target(mode:"current") restored/confirmed binding, panel_graph_outline showed the intended graph, and every retried panel_connect succeeded.

### Environment
Linux aarch64; NVIDIA GB10 121.6 GB VRAM; ComfyUI 0.31.1; ComfyUI frontend 1.48.7; Python 3.12.3; PyTorch 2.13.0+cu130; comfyui-mcp 0.50.87; panel 0.11.78.

Related (same root cause): #1292, #1027

artokun:
Closes #1330

Claiming. The reporter built 12 nodes successfully, then **every** `panel_connect` failed closed on a workflow-instance mismatch — and `panel_set_workflow_target({mode:"current"})` immediately reported `already_current` with the *same* uuid the command had been issued for. Retrying then worked.

So the mismatch was transient and self-correcting, and the caller paid 14 refusals to discover that.

Investigating whether the existing corroboration path (`readAndCorroborate` / `FENCE_CORROBORATION_RECHECK_STEPS_MS`) covers this route at all, and whether this is the known embed/swap race where the first record embeds the PRE-swap uuid.

Not planning to weaken fail-closed: the fence exists so a mutation cannot land on the wrong graph. The question is whether a mismatch that corroboration would clear should ever have been reported as one.

### DIFF
--- a/src/__tests__/orchestrator/fence-mismatch-corroborates.test.ts
+++ b/src/__tests__/orchestrator/fence-mismatch-corroborates.test.ts
@@ -1,0 +1,148 @@
+// #1330 — a mismatch that clears by itself still cost fourteen refusals.
+//
+// The reporter built a new workflow: panel_add_node and panel_set_widget succeeded for
+// 12 nodes, then EVERY panel_connect was refused with
+//
+//   workflow instance mismatch: this command was issued for workflow instance
+//   3282b824-…, and the active canvas reports cf8f83cd-…. Nothing was applied.
+//
+// Calling panel_set_workflow_target({mode:"current"}) immediately afterwards reported
+// `already_current` naming 3282b824 — the SAME uuid the refused commands carried — and
+// every retried connect then succeeded. So the canvas identity flipped and settled back
+// while a new unsaved workflow was materialising, and nothing in the refusal separated
+// that from "you are genuinely pointed at another workflow".
+//
+// THE FENCE IS NOT WEAKENED AND NOTHING IS AUTO-APPLIED. The refusal still stands; what
+// changes is that the orchestrator now performs the same read-only re-derivation the
+// documented recovery performs, and reports WHICH of the two states it found. One
+// informed retry replaces fourteen blind ones.
+//
+// Recommending a retry is safe here specifically because the panel checks the fence
+// BEFORE running the handler — "Nothing was applied" is structural, not an echoed claim.
+
+import { beforeEach, describe, expect, it } from "vitest";
+
+import {
+  buildPanelToolDefs,
+  makePanelToolCtx,
+  type PanelToolCtx,
+  type ToolResult,
+} from "../../orchestrator/panel-tools.js";
+import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+
+const TAB = "wf:workflows/a.json";
+const STAMP = "3282b824-5890-4c87-9da4-7936c155aa8c"; // what the command carried
+const OTHER = "cf8f83cd-ce65-4307-ad9c-0dc5f5d491d4"; // what the canvas transiently said
+
+let listCalls: number;
+let stamp: string | undefined;
+
+/** A self-corroborating workflow_list reply — the active record also appears in the
+ *  open-workflow list flagged active, which is what corroboration requires. */
+function settled(uuid: string): Record<string, unknown> {
+  const active = { path: "workflows/a.json", routing_key: TAB, workflow_uuid: uuid };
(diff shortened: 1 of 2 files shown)


artokun:
## Review pass (codex still unavailable)

Codex could not run — the sandbox denies all process launch (`CreateProcessAsUserW: Access is denied`), including `git diff`. Third attempt today; this is a disclosed self-review with the claims measured.

### The finding: the branch this whole block exists for could never fire

The uuid the command carried was first read from `cmd.workflow_uuid`. At this layer that is **undefined** — the stamp is applied downstream of `ctx.call`. So `stamped` was always `null`, the TRANSIENT branch was unreachable, and the reporter's exact case fell through to the generic `already_current` wording:

```
AUTO-REBIND: ATTEMPTED; the fence already named the live canvas (3282b824-…),
so it was not the stale side. Retry once — …
```

Correct, but not the answer they needed, and the "retry ONCE / don't re-issue the build" instruction that saves the 14 calls never appeared. Caught by running the test, not by reading the code. Both ids now come from the panel's own refusal, which is the authoritative record of what it compared.

### Scope discipline

The fence is **not** weakened and nothing is auto-applied — the refusal still stands and still says `Nothing was applied`. There is a test asserting exactly that, so corroboration cannot become a way to smuggle a mutation through.

Recommending a retry is safe *here specifically*: the panel checks the fence **before** running the handler, so "Nothing was applied" is structural rather than an echoed claim. That is what separates this from the transport failures beside it, where the outcome is genuinely unknown and `isRetrySafeCmd` correctly refuses to retry a mutation.

I did **not** implement the auto-retry half of the reporter's suggestion. Re-issuing a mutation on our own initiative is a larger change than this issue justifies, and the disclosure already collapses 14 blind refusals into one informed retry.

### Verified

- The re-derivation actually runs — asserted on the panel round trip (`listCalls > 0`), not just on message text. My notes from a previous fence fix record a test that passed while exercising nothing, because a prerequisite was missing; that is what this guards against.
- The three states are distinguished, including the failure direction: **unreadable promises nothing**. Claiming a retry will work when the panel could not be read would push the agent back into the loop this issue is about, with our encouragement.
- Whole block mutated away: 3 of 4 tests fail. Mutation confirmed applied via `grep -c`.

Full suite: 432 files, 8454 passed, 2 skipped. `tsc --noEmit` clean.

### Not fixed here

The underlying identity flip is panel-side — `resolveUnsavedInstanceUuid` mints a fresh uuid when the live-object WeakMap misses, which is what makes a brand-new unsaved workflow's identity oscillate. That is the root cause and it needs a separate change in the panel repo; this makes the symptom cost one call instead of fourteen.

PR #1383, opened 2026-08-11
feat(#873): restrict the tool surface — deny/allow lists and presets

### LINKED ISSUE #873: No way to restrict the tool surface — blocks production/multi-user deployments
Reported by **jnbackup** in the Discord `#help` "Docker Compose" thread, while wiring comfyui-mcp into Open WebUI behind llama-swap.

> missing a way to filter tools, aka a blacklist, this would make production setup possible, currently it seems theres no clean way to disable potentially dangerous tools.

## Current state

Compact tool mode (default since #667) exposes only `list_tools` / `describe_tool` / `call_tool`, which solves *discovery* pressure — the model no longer picks from ~200 names. But it is explicitly **not** a boundary: `call_tool` dispatches to anything by name, so the reachable surface is unchanged.

Grepped for an allow/deny mechanism (`TOOL_ALLOW`, `TOOL_DENY`, `ALLOWED_TOOLS`, `DISABLED_TOOLS`, `toolFilter`, allowlist/denylist/blocklist) — there is no tool-level filter. The only adjacent knob is `COMFYUI_MCP_ALLOW_GIT_WRITES`, which gates one specific capability.

## Why it matters

The deployment shape here is a shared/hosted chat frontend where the operator is not the person prompting. That operator needs to guarantee the model cannot reach install/delete/restart/filesystem-write tools regardless of what a user types. Today the honest answer to "can I stop the model calling `remove_model` or `restart_comfyui`?" is no.

This is the stated blocker on shipping a Docker Compose + Open WebUI setup, so it also gates the headless/non-panel use case generally — which artokun noted in the same thread has drifted out of focus relative to the panel.

## Sketch

An env-level filter applied at registration **and** inside `call_tool`, so compact mode can't be used to route around it:

- `COMFYUI_MCP_TOOL_DENY=remove_model,restart_comfyui,...` (comma/glob)
- `COMFYUI_MCP_TOOL_ALLOW=generate_image,get_queue,...` — allow wins, deny as the coarse default

Enforcing in the `call_tool` dispatcher is the load-bearing part; filtering only the registered list would leave the compact facade as a bypass. Denied names should be absent from `list_tools` rather than erroring on call, so the model never learns they exist. A named preset (`safe`, `readonly`) would cover most of the demand without operators hand-maintaining lists.

## Environment

Docker Compose, Open WebUI frontend, llama-swap, single GPU.

artokun:
Closes #873

Reported via Discord by **jnbackup**, wiring comfyui-mcp into Open WebUI behind llama-swap: *"missing a way to filter tools, aka a blacklist, this would make production setup possible."*

**The trust model here is genuinely different from the local one.** This codebase assumes one machine, one person — operator and prompter are the same, so a tool filter would be theatre. In a hosted/shared chat frontend the operator is **not** the person prompting, and they need to guarantee the model cannot reach install/delete/restart tools regardless of what a user types. That is a real boundary, and today the honest answer is that it does not exist.

**The load-bearing part is .** Compact mode narrows *discovery* but is explicitly not a boundary — `call_tool` dispatches by name, so filtering only the registered list would leave the facade as a bypass.

The architecture already has the right seam: `withBlindImageGate(registrar)` is a registrar decorator applied by **both** `registerAllTools` and `collectToolCatalog` — and `call_tool` dispatches through that same catalog. A filter in the same position covers both paths by construction rather than by remembering to.

Planned:
- `COMFYUI_MCP_TOOL_DENY` / `COMFYUI_MCP_TOOL_ALLOW` (comma lists, globs), allow wins.
- Presets so operators are not hand-maintaining lists.
- Denied names **absent** from `list_tools` rather than erroring on call, so the model never learns they exist.

### DIFF
--- a/docs/configuration.mdx
+++ b/docs/configuration.mdx
@@ -274,6 +274,46 @@ available, HTTP polling otherwise).
   single step won't honor it immediately — this wait is what detects a true wedge.
 </ParamField>
 
+## Restricting the tool surface
+
+For a **hosted** deployment — a shared Open WebUI, a team frontend — the operator is not
+the person prompting. These variables withhold tools from the model entirely: a withheld
+tool is never registered, so it is absent from `tools/list`, absent from `call_tool`, and
+the model never learns it exists.
+
+<ParamField path="COMFYUI_MCP_TOOL_PRESET" type="string">
+  `safe` — everything except tools that change the machine or the model library.
+  Rendering still works; installing, deleting and restarting do not.
+  `readonly` — inspection only: no renders queued, nothing written, nothing spent.
+  Both also withhold the whole `panel_*` surface, which drives a live shared canvas.
+</ParamField>
+<ParamField path="COMFYUI_MCP_TOOL_DENY" type="string">
+  Comma-separated tool names to withhold, e.g. `restart_comfyui,download_model`. A
+  trailing `*` matches a family: `train_*`. Applied on top of any preset **and** on top
+  of an allow list.
+</ParamField>
+<ParamField path="COMFYUI_MCP_TOOL_ALLOW" type="string">
+  Comma-separated allow list. When set, the surface is **exactly** these tools — anything
+  not named is withheld even if no deny rule mentions it. Use it to opt individual tools
+  back in past a preset: `COMFYUI_MCP_TOOL_PRESET=safe` plus
+  `COMFYUI_MCP_TOOL_ALLOW=panel_graph_outline,panel_query_graph`.
+</ParamField>
+
+```bash A hosted deployment that cannot install or restart anything
+COMFYUI_MCP_TOOL_PRESET=safe npx comfyui-mcp@latest --http --host 0.0.0.0 --port 9100
+```
+
+<Warning>
+  This is a boundary against the **model** and the people prompting it — not against
+  whoever sets the environment, who can simply unset it, and not a substitute for keeping
+  an untrusted party off the ComfyUI host.
+
+  A misconfiguration **refuses to start** rather than starting unrestricted: an unknown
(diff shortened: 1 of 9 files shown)


artokun:
## Live verification against a built server

Ran the real binary, not the test doubles.

**The refusal is now true on every transport.** `COMFYUI_MCP_TOOL_PRESET=saef` exits 1 with the readable message on all three start paths — stdio, `--http` (which previously bound the socket, printed `running on http://…`, and 500'd every `initialize`), and `--panel-orchestrator` (which previously bound the bridge, the loopback MCP and the console, printed its ready banner, and only threw later at first use). Also verified via `~/.comfyui-mcp/.env` rather than a shell variable, since the hoist could have moved the read above the dotenv hydration — it did not (hydration is module-scope in `config.ts`, which `boot.ts` imports).

**The surface actually shrinks**, over a real MCP handshake:

| policy | tools |
|---|---|
| none | 40 |
| `PRESET=safe` | 16 |
| `PRESET=readonly` | 10 |

Withheld under `safe`: `restart_comfyui`, `install_custom_node`, `download_model`, `comfy_cli`, `list_local_models`, and **`list_packs`** — the round-2 P0, whose `action:"install_deps"` installs and runs third-party code.
Withheld under `readonly`: `generate_image`, `enqueue_workflow`, `report_issue`, `list_api_nodes`.
Left usable under `readonly`: exactly the 7 read-only tools plus the 3 facade tools.

**The bypass the issue is actually about is closed.** In compact mode under `readonly`, `list_tools` reports "7 of 7" and:

```
call_tool {"name": "restart_comfyui"} -> Unknown tool 'restart_comfyui'. Use list_tools to see the catalog.
```

Absent from dispatch, not merely hidden from the listing — which was the reporter's specific concern.

Issue #1378, opened 2026-08-11
download_model: saving a credential via panel_request_secret orphans in-flight downloads (respawn + auth-header cache-key change double-invalidates near-complete .partial files)


artokun:
### What happened / root cause
Two large `download_model` action:"download" transfers (a ~19.53GB diffusion model and a ~14.61GB text encoder, both public/anonymous HuggingFace URLs, no auth needed) were streaming normally and had reached 77% and 96% respectively (confirmed via `action:"status"`). In the same session I then called `panel_request_secret` to store `HF_TOKEN` (target_kind "env", mcp_server "comfyui") — this is documented as respawning the comfyui tool session so the new credential takes effect ("Tool sessions being rebuilt with the new environment: 1 queued for the end of this turn"). The very next `action:"status"` call for both downloads reported `NOTE: heartbeat stale`, and `action:"cancel"` for both confirmed "that session is confirmed GONE — its process no longer exists" and closed the records as cancelled.

Per the cancel reply's own guidance ("re-issue... it resumes from any .partial the dead session left on disk, or restarts cleanly"), I re-issued both `action:"download"` calls with the identical URL/target_subfolder. Both restarted from 0% instead of resuming.

Investigating the actual cache directory (`~/.comfyui-mcp/cache/`, i.e. `DEFAULT_CACHE_DIR` in `dist/services/download-cache.js`) showed the ORIGINAL `.partial` files were never lost at all:
- `.9eb46fb70f2fa5df38ffa1178b42fd65.safetensors.partial` — 15,207,387,077 bytes (≈14.16GB, matches the reported 14.15/14.61GB=96% text-encoder progress almost exactly)
- `.f30b7c6ebe9bd2f2966226f282e17ece.safetensors.partial` — 16,291,583,966 bytes (≈15.17GB, matches the reported 15.17/19.53GB=77% diffusion-model progress almost exactly)

Both stopped growing at the moment of the respawn (mtime frozen) but were NOT deleted. However, the two freshly re-issued downloads created BRAND NEW cache entries with different hash prefixes (`.066876aca5df8c8e5ec81bd14e41376d.safetensors.partial`, `.549121a396a64937b8bfe5eb7dab5e4b.safetensors.partial`) and began growing from 0 — i.e. 4 `.partial` files for what should be 2 downloads, with ~29GB of already-fetched, still-present, still-valid bytes now orphaned and effectively wasted.

[...]

Issue #1370, opened 2026-08-10
download_model cancel/status promise a resumable .partial, but nothing is left on disk (33GB restart from 0)


artokun:
### What happened / root cause
`download_model (action:"cancel")` reports:

> cancelled - the partial was left on disk and can be resumed by re-issuing the download (it picks up where it left off)

and `action:"status"` repeats the same claim. On this machine that is **not true**: after cancelling two in-flight HuggingFace downloads, a recursive filesystem scan of the whole ComfyUI install tree found **no partial file of any kind** - neither `*.partial` nor any partially written target file. The already-transferred bytes (~0.34 GB of a 16.8 GB text encoder and ~0.92 GB of a 33 GB diffusion model) are gone, so re-issuing restarts from zero.

Corroborating signal from earlier in the same session: three downloads (flux2-vae 0.31 GB, mistral_3_small_flux2_fp8 16.8 GB, flux2_dev_fp8mixed 33 GB) had been streaming ~20 min when the orchestrator restarted. Re-issuing all three resumed **only** the small VAE (came back at 31%); the two large ones came back at 0-2%, i.e. also from scratch. So resume appears to work for a small/completed-ish file but not for the large ones.

Impact: the message is load-bearing for a user decision. The user paused a 33 GB download specifically because the tool said it was resumable.

### Steps to reproduce
1. `download_model action:"download"` a large HF file (e.g. `https://huggingface.co/Comfy-Org/flux2-dev/resolve/main/split_files/diffusion_models/flux2_dev_fp8mixed.safetensors` -> `diffusion_models`).
2. Let it stream to ~2% (about 1 GB).
3. `download_model action:"cancel"` with that id. Reply claims a resumable partial was left.
4. `download_model action:"status"` - row says `cancelled - the partial was left on disk and can be resumed`.
5. Inspect the destination directory and the whole install tree: no partial file exists.
6. Re-issue the identical download: it starts at 0%.

### Exact error (scrubbed)
No exception - this is a **wrong/misleading success message**, which is the harder failure mode.

```
- `a7675173bd25f50c` (tray `8565874a46c397ea`) **cancelled**
    cancelled - the partial was left on disk and can be resumed by re-issuing the download (it picks up where it left off)
    from: https://huggingface.co/Comfy-Org/flux2-dev/resolve/main/split_files/diffusion_models/flux2_dev_fp8mixed.safetensors
```

[...]

Issue #269, opened 2026-07-21
RunPod connector: deferred review findings (#251)


artokun:
Tracking the lower-priority review findings from the two posted reviews on PR #251 (feat/runpod-connector) that were NOT fixed in the money/security fix commit. The HIGH/MEDIUM blockers (non-idempotent billed create retry, tool-secret leak into agent subprocess envs, failed auto-stop lying EXITED, API key in query string, missing fetch timeout / poll overlap) are fixed on the branch.

Deferred items:

- [ ] `src/services/panel-secrets.ts` — `removeEnvSecret` does not purge the legacy `panel-secrets.json` maps, so a revoked key can resurrect via `migrateSecretsToEnv()` on the next boot (revoked-key resurrection).
- [ ] `src/tools/runpod.ts` — split-brain retarget: the tool path retargets the ComfyUI client but does NOT retarget the QueueMonitor / agent subprocesses, which keep talking to the old host.
- [ ] `src/services/runpod-watch.ts` — idle signal measures the LOCAL queue only and ignores in-flight downloads; a pod busy downloading models counts as idle and can be auto-stopped.
- [ ] `src/services/runpod-watch.ts` — no dead-man switch: if the orchestrator process dies, the pod keeps running (and billing) forever with no auto-stop.
- [ ] Orchestrator call_tool whitelist — `runpod_pod_create` (a billed mutation) is agent-invocable without user confirmation.
- [ ] `src/services/runpod-client.ts` — GraphQL responses are cast, not structurally validated; a shape change silently produces wrong values.
- [ ] `src/services/runpod-watch.ts` / `src/tools/runpod.ts` — an externally deleted pod leaves a dead ComfyUI target configured (no retarget/cleanup on vanish).
- [ ] `src/tools/runpod.ts` — `connect: true` is effectively a no-op in one path (documented flag does nothing).
- [ ] `src/tools/runpod.ts` — `use_local` forces `127.0.0.1`, breaking rigs where local ComfyUI lives on another LAN host.
- [ ] `src/services/runpod-watch.ts` / status handling — status machine only knows RUNNING/EXITED; missing CREATED/RESTARTING/PAUSED/DEAD/TERMINATED handling.
- [ ] `src/services/panel-secrets.ts` / orchestrator — `setAgentSecret` restarts ALL agents and injects a download-retry message regardless of which secret changed.
[...]

PR #1406, opened 2026-08-11
fix(1378): warn before a credential save orphans in-flight downloads

artokun:
Draft — claiming #1378, and taking the reporter's option **(c)** deliberately, not for lack of ambition.

Saving a credential mid-flight respawns the comfyui tool session, and the new auth header changes the download's cache identity — so a `.partial` at **96%** becomes unreachable and the re-issued download starts from 0. In their session that cost ~29 GB for two files that were 77% and 96% done. Nothing warned them, and nothing indicated the bytes were still on disk.

**Why not (a) or (b).** The reporter is right that both risk a real security regression. The cache key folds in representation-affecting headers precisely so bytes fetched under one identity can never be served to a request made under another; "this URL didn't need auth anyway" is a judgement the client cannot make before fetching, and getting it wrong leaks gated bytes across identities. That is a worse failure than a re-download. They chose not to guess, and filed a diagnosis instead — which is the right call and worth saying so.

**What (c) actually needs to be useful:** the warning has to arrive *before* the respawn, because afterwards the choice is gone. So `panel_request_secret` should say, at the moment of asking, that N downloads totalling ~X GB are in flight and will restart from 0% — letting the caller finish them first.

There is a piece of machinery ready for this. #1370 added `stagedPartialPathForUrl(url)` / `findResumablePartial(url)`, which derive the staged path from the **unauthenticated** cache identity — exactly the key the in-flight downloads were using. So the orphaned bytes are locatable by the same helper that now reports a cancelled download's partial, and the warning can quote real sizes rather than an estimate.

Scope I am not taking: (d), the GC for orphaned cache entries. It is real — those two `.partial` files will sit there forever — but it is a separate lifecycle question, and bundling it would put a delete path in a PR about a warning.

Refs #1378

### DIFF
--- a/src/__tests__/services/respawn-orphans-downloads.test.ts
+++ b/src/__tests__/services/respawn-orphans-downloads.test.ts
@@ -1,0 +1,105 @@
+// #1378 — saving a credential mid-flight respawns the comfyui tool session, and the new
+// auth header changes each in-flight download's CACHE IDENTITY. A `.partial` at 96% becomes
+// unreachable and the re-issued download starts from zero. A reporter lost ~29 GB across
+// two files: the bytes were never deleted, nothing could find them again, and nothing
+// warned them until it was already too late to wait.
+//
+// RESCUING THE TRANSFER IS DELIBERATELY NOT ATTEMPTED. Reusing the old cache entry under
+// the new identity would serve bytes fetched under one auth identity to a request made
+// under another — exactly what folding headers into the key exists to prevent, and a worse
+// failure than a re-download. The reporter reached the same conclusion and filed a
+// diagnosis rather than guess at a patch.
+//
+// So what ships is the CHOICE, delivered while it is still a choice: before the respawn.
+//
+// The job list is INJECTED, not mocked. `downloadsAtRiskOfRespawn` calls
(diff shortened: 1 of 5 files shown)


artokun:
## Scope note: the REVOKE path has the same defect and is not covered here

While checking that the snapshot sits before every respawn, I found a second one. `removeSecret` fires `emitComfyuiChange({})` (`src/services/panel-secrets.ts:1803`, the `#269` comfyui-only restart) — so **revoking** a credential orphans in-flight downloads by exactly the same mechanism as saving one, and says nothing at all.

I am not fixing it in this PR, deliberately:

- #1378 is about a credential **save**, and that is what this delivers;
- the revoke path returns a different reply shape (`changed`/`lostKeys`/`uncertainty`/…) with **no respawn report** on it, so there is nowhere to hang `applied` vs `scheduled` — and the whole point of this change was that those two are different facts deserving different sentences. Bolting a field onto a reply with no renderer for it would add a value nobody reads;
- it needs its own reply wording and its own tests, which is a second change wearing this one's branch.

The snapshot helper (`snapshotAtRiskDownloads`) is already in that file and reusable, so the follow-up is small — it just is not this one. Filing separately unless you would rather I fold it in.

Issue #819, opened 2026-08-04
Graph mutations blocked when active panel bundle is too old and panel pack is absent


artokun:
### What happened / root cause
A live graph mutation was refused because the active sidebar did not advertise per-command workflow targeting (required panel 0.11.35+). The same session could perform read-only panel_graph_outline/panel_query_graph calls. install_panel status then reported that the sidebar pack was absent from custom_nodes. A single install attempt was made, but stale ComfyUI-Manager 3.x reported its queue drained without creating the pack.

### Steps to reproduce
1. Open a saved workflow in the sidebar and call panel_graph_outline successfully.
2. Call panel_set_widget on a node in that workflow.
3. Observe the safe-targeting refusal.
4. Run install_panel(action='status'), then install_panel(action='install').
5. Manager reports completion, but no panel pack exists on disk.

### Exact error (scrubbed)
"graph_set_widget cannot be safely targeted to the active workflow: panel tab wf:workflows/video_minimax_h3_r2v.json does not enforce per-command workflow targeting (detected panel version unknown; this MCP requires panel 0.11.35+)."

Install result:
"Panel install did NOT land: the pack is not present in custom_nodes after ComfyUI-Manager reported the queue drained. This is the stale ComfyUI-Manager 3.x silent no-op."

### Fix
Applied locally: no
Upstream-only: yes
One automated install attempt was made and correctly detected as not landed. The immediate recovery requires updating ComfyUI-Manager or installing comfyui-mcp-panel from source, restarting ComfyUI, then hard-refreshing the browser. A robust setup path should either install without relying on the stale Manager path or clearly surface a one-action recovery route before graph mutations are attempted.

### Environment
Windows 11
ComfyUI 0.30.0 LOCAL
comfyui-mcp 0.49.4
panel version: unknown; active UI bundle present, pack absent from custom_nodes
GPU: NVIDIA GeForce RTX 5070 Ti 16 GB

Related (same root cause): #724, #639

PR #1419, opened 2026-08-11
fix(1325): ui-bridge.test.ts flakes in isolation, not just under load

artokun:
Claiming #1325.

**New evidence that changes the investigation.** The issue (and my earlier work on this branch) framed this as load sensitivity, because that is where it kept being seen and isolation was the control I ran once. Run the file four times alone and it fails on its own:

```
vitest run src/__tests__/services/ui-bridge.test.ts   # nothing else running
run 1:  1 failed | 205 passed   <- #819: a version INHERITED across a reconnect is never attributed to this tab
run 2:  206 passed
run 3:  206 passed
run 4:  206 passed
```

Reproduced on clean main at `d430d67` (0.51.0) and again on an unrelated branch. That points inside this file — shared mutable state between its own tests, or a timing assumption — rather than at cross-file interference.

Six distinct tests in this one file have now been observed failing:
- the three originally reported
- `names the versioned panel-sync remedy when stamp enforcement is absent (#706)`
- `quotes NO fence version when this build's capability table cannot state one`
- `#819: a version INHERITED across a reconnect is never attributed to this tab`

**Work already on this branch** (from an earlier session, previously unpushed): the #357 test now names why it failed instead of recording a bare boolean, and a busy event loop no longer decides a deadline — measured with `setImmediate` rather than `setTimeout(0)`, because Windows' ~15.6 ms timer floor was being reported as contention. That commit is explicit that it removes one vector and is not a proven cure.

Next: catch the isolated failure with full output and fix the cause rather than the symptom.


artokun:
Round 2 review addressed. The remaining P2 — the two `ui-bridge.test.ts` waits inside `it(…, 10_000)` and `it(…, 12_000)` — was already fixed in the working tree when the review ran against `f83606d`; it is committed now as `f202e02`, along with a guard so the class cannot come back.

Both waits are for a local connect that takes milliseconds, so they take an explicit 3s: slack, not a deadline under test, and comfortably under each test's own budget so the wait names what never arrived instead of the runner reporting a generic timeout.

The guard parses the `it(...)` call and walks its parens rather than pattern-matching, because both earlier versions of that scan were wrong in opposite directions and I nearly shipped each one: an indent-blind regex read `setTimeout(fn, 80)` as an 80ms test budget and invented offenders; a lookahead stopped at the first `)` — inside `expect(...)` — so every bounded wait looked unbounded. It is driven over a fixture that must trip it and one that must not, because a scan for a condition passes trivially when the scan is broken.

Confirmed from your audit and kept as stated limits:
- explicit timeouts pass through unchanged (the 2.4s watchdog bounds, run-template's 10s);
- no rewritten wait runs under `vi.useFakeTimers()`;
- the other 288 rewrites are pure substitutions plus imports;
- **15s is not proven sufficient under heavier load.** The commit message says so, and #1325 stays open until the suite has run clean for a while. What is established is that the inherited 1s is too short and that these waits are deadline-sensitive rather than logically broken.

Issue #1339, opened 2026-08-10
panel_run batch_count repeats identical seed despite Randomize Each Time


artokun:
### What happened / root cause
panel_run with batch_count:10 queued ten snapshots of the same live workflow without applying the frontend's "randomize after generation" seed mutation between queue submissions. The workflow uses rgthree Seed node 649 with Randomize Each Time, linked to the base sampler and upscale seed. Completed outputs 00027 through 00030 were pixel-identical, all using the same seed. This makes batch_count misleading for training-dataset generation.

### Steps to reproduce
1. Open a live workflow whose sampler seed is driven by an rgthree Seed node configured to Randomize Each Time.
2. Call panel_run({batch_count:10}).
3. Compare completed images: each queued prompt has the same serialized seed and produces the same image.

### Exact error
No tool error was returned. panel_run reported queued:true and ten prompt IDs, but repeated the identical graph state.

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: panel_run should apply the same after-generate widget mutation used by the ComfyUI frontend between each batch queue snapshot, or explicitly reject/warn when batch_count>1 and mutable/randomized widgets would otherwise serialize identically. A robust fallback is to assign a distinct seed per snapshot before POST /prompt.

Temporary workaround: queue one render at a time after explicitly setting a unique seed and pose prompt.

### Environment
Windows 11; ComfyUI 0.31.1; frontend 1.48.7; Python 3.13.12; PyTorch 2.13.0+cu130; NVIDIA RTX 5070 Ti Laptop GPU 12 GB; comfyui-mcp 0.50.93; panel 0.11.44.

Related (same root cause): #558

artokun:
I could not reproduce this, and I want to give you the measurements rather than a guess — they narrow it a lot.

## What I tested

On a live rig (ComfyUI 0.31.1) I intercepted the outgoing `POST /prompt` bodies and captured the **serialized seed of every prompt in a batch**, blocking the requests so nothing actually queued. Two runs, three prompts each:

**Standard KSampler seed, `control_after_generate: randomize`**

```
seeds submitted: [42, 582244630682917, 1004083937439000]   → 3 distinct
```

**`Seed (rgthree)` node set to randomize (widget value -1), same batch**

```
rgthree seeds:   [867157653763257, 977811429032650, 662461828083988]  → 3 distinct
KSampler seeds:  [124357035363877, 660991880834551, 634489945929962]  → 3 distinct
```

[...]

Issue #988, opened 2026-08-07
Release flow: the version tag is never an ancestor of main, so gen-changelog's range is wrong in BOTH directions


artokun:
Hit while cutting 0.50.3. The generated changelog had to be corrected by hand, and my first attempt at a fix made it worse in a quieter way — so filing the analysis rather than shipping a heuristic.

## The root cause

The release flow pushes the tag from a **local** commit, then reconciles the version bump onto protected `main` through a **squash merge**, which writes a different sha.

So `v0.50.2` points at `f22b8d4`, which is not an ancestor of `main`. `main` has `6116568` (`0.50.2 (#979)`) instead. The tag and the branch permanently disagree about where the release is.

## Failure 1 — the tag boundary over-includes

`gen-changelog` opens its range with `git describe --tags --abbrev=0`, which returns the newest tag **reachable from HEAD**. It walks straight past `v0.50.2` to `v0.50.1`.

Generating 0.50.3 produced a section listing **all fifteen** commits since `v0.50.1`, re-announcing the ten fixes 0.50.2 had already shipped. Caught by reading it; corrected by hand before the tag was pushed.

## Failure 2 — the release-commit boundary under-includes

The obvious fix is to bound by the newest **release commit on the branch** (`6116568`) instead. I implemented that and tested it against the exact commit that produced the bad output. It yields **4** entries where the correct answer is **5**.

The missing one is **#976**. It merged *after* the `v0.50.2` tag was cut but *before* the reconcile PR landed — so:

```
#976 is NOT in the published 0.50.2      (not an ancestor of the tag)
#976 IS an ancestor of the release commit (it is behind 6116568 on main)
```

It falls in the gap between "what shipped" and "what the branch calls the release". A release-commit boundary silently drops everything merged during that window.

**Dropping entries is worse than duplicating them.** A duplicate is visible noise someone will notice; a missing entry is invisible. So I reverted the change rather than trade a loud wrong answer for a quiet one.

## What would actually fix it

Close the gap instead of guessing across it. Either:

1. **Tag the merged commit.** Push the version PR first, then tag `main` after it merges. The tag becomes an ancestor, `describe` works, and there is no window — this also makes `v0.50.x` mean the same thing to git and to npm, which it currently does not.
2. **Fast-forward the release commit** instead of squashing it, so the tagged sha survives onto `main`.

Both remove the ambiguity at the source. A smarter range heuristic cannot, because the two boundaries genuinely disagree and neither is complete.

[...]

PR #1423, opened 2026-08-11
fix(1339): rgthree seeds are invisible to the batch-repeat warning

artokun:
Claiming #1339. **The measurement redirected this** — see the [full results](https://github.com/artokun/comfyui-mcp/issues/1339#issuecomment-5257682765).

Measured on the live rig with `/prompt` intercepted so nothing queued:

| run | posted seeds |
|---|---|
| `app.queuePrompt(0, 3)` | 742732484087382, 27923863108977, 114376582494196 — differ |
| `app.queuePrompt(0, 3, ["29"])` | 77161535712866 ×3 — repeat |
| rgthree handler driven 3× with `seed: -1` | 1028465986822020, 98533269447704, 557498712106716 — differ |

Three things follow:

1. **#988 reproduces**, on current ComfyUI. A scoped batch repeats core's `control_after_generate` seed.
2. **A scoped batch still calls `api.queuePrompt` three times.** What the scope suppresses is the widget-advance step, not the API call.
3. **So it does not explain an rgthree seed repeating.** rgthree patches `api.queuePrompt` and substitutes into the outgoing body, so an *armed* rgthree Seed (widget holding the `-1` sentinel) gets a fresh seed per item even when scoped.

What IS established as a defect, independent of the reporter's cause: a live `Seed (rgthree)` node has widgets `seed` + three buttons and **no `control_after_generate`** — rgthree deletes it. So #988's disclosure is structurally blind to the most widely used custom seed node, and the reporter got `queued: true` with ten prompt ids and no indication of anything.

**Blocked on one fact** only the reporter has: whether node 649's `seed` widget held `-1` (armed → should have varied → something on the panel path is stopping the hook) or a concrete number (→ identical seeds are correct, and the missing signal is that nothing said the node was not armed). Asked on the issue.

The detector fix lands in `comfyui-mcp-panel` (where `scoped-batch-seed.js` lives), so this PR stays open as the claim while that question is outstanding.


artokun:
No code here — this was a claim marker, and the work it was claiming shipped elsewhere.

The half of #1339 that was **established** shipped in **panel 0.14.3**: a `batch_count` run whose seed comes from an rgthree Seed node now says up front when every item will use the same seed, and why. That landed in the panel repo, so there was never a change to make on this branch.

The remaining half is **waiting on the reporter**, not on me: the measurements on the rig showed an armed rgthree node varies per item even in a scoped batch, so what is left turns on one fact only they have — whether node 649's seed widget showed `-1` or a concrete number. Detail on #1339.

Closing so this does not read as active work while it is blocked. The issue stays open; if the answer comes back `-1`, that is a live panel-path defect and I will re-claim it then.

Issue #1352, opened 2026-08-10
panel_request_secret hard-times-out at 300s while a human is still typing the secret


artokun:
Split out of #1331 (bug 2). The remedy half of bug 1 is fixed and released; this is untouched.

## What happened

During a long panel session, `panel_request_secret` **hard-timed-out at 300s while awaiting human input**. A human typing a credential into a masked field is not a stalled operation — 300s is a plausible amount of time for someone to go find an API key, and the timeout fires as though nothing were happening.

## Why this one is different from every other ask

`request_secret` exists precisely so the value never passes through the conversation — the user types it into a masked field and the orchestrator receives it directly. That makes the human latency **structurally longer** than a yes/no confirm: they may have to open a password manager, a billing dashboard, or another machine.

It also makes the failure expensive in a way an ordinary ask is not. There is no "just paste it in chat" fallback — that is the exact route the tool is designed to prevent (see #952, where suggesting the conversational route was itself the defect).

## What needs deciding

1. Whether the secret-collection deadline should be separate from the general ask timing at all, rather than inheriting it.
2. Whether a timeout here should be **recoverable** — the card is still on screen and the user may still be typing, so a late answer arriving after the tool has given up is the interesting case (there is already late-reply buffering for asks; does it cover this path?).
3. What the caller is told. A timeout that reads as "the user declined to give it" would be the #1332 defect in a new place: nobody declined, the answer just has not arrived yet.

## Environment

Windows 11; ComfyUI 0.30.0; frontend 1.48.6; comfyui-mcp 0.50.93; panel 0.11.38.

Original report: #1331.

artokun:
The **truthfulness** half is fixed and ships in the next release (#1364).

A timed-out secret card no longer relays *"the ComfyUI tab may be backgrounded or frozen"*. That sentence is right for every other command and wrong for this one: the card blocks 300s on a masked input, and the overwhelmingly likely reason it went unanswered is that someone is still fetching the token from a password manager or a billing console. Nothing is wrong with the tab.

It now says nothing was saved, that this is very likely just time, and that it is neither a frozen tab nor a refusal.

[...]

PR #1424, opened 2026-08-11
fix(1352): a human fetching a credential is not a stalled operation

artokun:
Claiming the part of #1352 that is still live. The **message** half shipped in #1364 (`1bf9afa`) — a timed-out secret card no longer reads as a frozen tab. Verified on main, two things remain:

```
panel-tools.ts:8623   { cmd: "request_secret", label, hint },
panel-tools.ts:8624   { tabId: ctx.tabId, timeoutMs: 300000 },
```

1. **The 300s deadline is still hard-coded** on the bridge send. Someone opening a password manager, a billing dashboard, or another machine is doing exactly what this tool is for, and 300s is a plausible amount of time for it.
2. **A late answer is dropped.** The late-reply buffer keys on `cmd.ask_id` (`ui-bridge.ts:3905`), and `request_secret` sends none — so an answer that validates after the deadline is discarded rather than buffered, unlike an `ask_user` card (#486).

**A constraint that shapes (2), stated up front:** the existing late-answer path hands the value to `lateAskSink`, which writes to a DURABLE JOURNAL. A secret must never go there. Whatever recovers a late secret has to be in-memory only, must never reach that sink, a transcript, or a log — so this cannot simply reuse the `ask_id` mechanism, and I will not wire it up that way.

Question 3 from the issue (what the caller is told) is already answered by #1364; this PR must not regress it — a recovered-late answer and an expired one are different outcomes and have to read differently.

### DIFF
--- a/src/__tests__/orchestrator/secret-card-timeout-truth.test.ts
+++ b/src/__tests__/orchestrator/secret-card-timeout-truth.test.ts
@@ -21,11 +21,12 @@
 // card so an internal MCP client does not kill the request first (#325); moving one
 // means moving the other, which is the design decision #1352 is really asking for.
 
-import { describe, expect, it } from "vitest";
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
 
 import {
   buildPanelToolDefs,
   makePanelToolCtx,
+  __panelAskTestHooks,
   type PanelToolCtx,
   type ToolResult,
 } from "../../orchestrator/panel-tools.js";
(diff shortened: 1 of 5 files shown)


artokun:
**Review status, disclosed:** codex never produced a job for this PR — the third failure today (one wedge at 9 minutes, one exit-1, this one silent). So this was a Claude adversarial review against the probe list I would have given it. Calling that out rather than implying an independent second opinion.

**Probe 1 — secret leakage.** Verified by driving a canary token through the REAL bridge over a real socket, not by reading:

```
recoverable in memory : true
reached the journal   : false
appeared in any log   : false     (debug/info/warn/error all captured)
drained after claim   : true
```

`request_secret` is the only secret-collecting command in the repo. `recordLateMutation` stores `{ok, cmd, tabId, ts, lateByMs}` — no value. The log assertion is now permanent, and adding `logger.debug(… msg.result)` beside the sink fails it.

**Probe 2 — does the new `ask_id` expose the card to anything?** It surfaced a real correction. `AskAnswers.tracks()` accepts only `panel_ask`-prefixed ids, so the journal declines a secret answer **on its own** — my comment had implied the new `sensitive` flag was the only barrier. It is the second of two, and now says why it is still worth having. That same journal's docstring claimed the sink "is fed by every `ask_user` card (confirm/consent/secret gates included)" — true when written, falsified by this branch, now fixed.

**Probe 3 — budget.** `getAskTiming()` clamps deadline ≤ cap−1s and grace ≤ cap−deadline, so deadline+grace ≤ 285s, under the ~300s framework kill and the 315s client timeout. Persistence after a late recovery is a local file write measured in milliseconds.

**Probe 4 — the `lateAware` prefix.** Found a real defect of mine: it wraps EVERY outcome including failures, so a token that arrived during the grace and then failed to persist was announced as "was still applied" by the very message explaining that it was not. It states only WHEN it arrived now; a mutation restoring the old wording fails a test.

**Probe 5 — the tests.** Mutation-verified: dropping the recovery, sending no `ask_id`, clearing the `sensitive` flag, ignoring the flag at the sink, and restoring the over-claiming note each kill a test. Two of these tests were toothless when written — the bridge double answered regardless of the command, and nothing exercised the sink — which is what prompted the canary run.

8782 tests green, gates green.

Issue #1421, opened 2026-08-11
panel_strip_workflow uses stale localhost target instead of connected panel ComfyUI


artokun:
### What happened
panel_strip_workflow({pack:'krea2-txt2img-manual'}) failed while the live sidebar panel was connected and successfully reading/running its remote RunPod ComfyUI.

### Exact error
fetch failed: connect ECONNREFUSED 127.0.0.1:8188 while requesting /object_info. The tool states the connected sidebar panel is on a different remote origin.

### Steps to reproduce
1. Connect sidebar panel to a remote ComfyUI.
2. Leave COMFYUI_URL at its default localhost endpoint.
3. Call panel_strip_workflow for an installer pack.

### Fix
Applied locally: no. Upstream-only: yes. Route pack node-definition lookup through the same active panel target, or detect/reject the target mismatch before the call.

### Environment
Remote Linux RunPod; ComfyUI 0.30.0; RTX 4090 24GB; comfyui-mcp 0.50.93; panel 0.11.87.

Related (same root cause): #1359

artokun:
You attributed this to #1359 yourself, and that is exactly right — it is the same defect, and it was fixed in **v0.50.115**. You are on **0.50.93**, so you never had the fix.

`dd6eaf0` — *fix(1359): the live canvas gets its node definitions from its own ComfyUI* (PR #1390). It routes `panel_strip_workflow`'s node-definition lookup through the connected panel instead of `COMFYUI_URL`, which is your suggested fix, arrived at from the same reasoning: a pack strip must be validated against the definitions the canvas actually has, and on a remote session those live nowhere near localhost.

Verified on today's main (0.51.3): the two suites that pin this behaviour — `strip-uses-panel-object-info.test.ts` and `strip-object-info-host.test.ts` — are green at 17 tests. They assert the panel is asked, that its answer names which host served it, and that a panel which cannot serve `/object_info` is reported as that rather than as an empty definition set.

Please upgrade and reopen if `panel_strip_workflow` still reaches 127.0.0.1 with a remote panel connected — with the tool's output, since the current version names the serving host and that makes the next report much shorter.

artokun:
[...]

Issue #1415, opened 2026-08-11
list_packs list_templates ignores connected sidebar target and uses stale COMFYUI_URL


artokun:
### What happened / root cause
`list_packs` with `action:"list_templates"` failed even though the sidebar panel is connected and panel graph tools work. The helper addressed the headless `COMFYUI_URL` (`http://127.0.0.1:9`) instead of the ComfyUI server currently connected through the browser panel. This makes template discovery unusable in panel-backed sessions whenever the headless URL is stale or intentionally unreachable.

### Steps to reproduce
1. Start comfyui-mcp with a stale/unreachable `COMFYUI_URL`.
2. Connect the sidebar panel to a live ComfyUI tab.
3. Confirm `panel_graph_outline` works.
4. Call `list_packs` with `{action:"list_templates"}`.

### Exact error
`fetch failed: bad port — while requesting http://127.0.0.1:9/api/workflow_templates (GET). That is the headless ComfyUI target (COMFYUI_URL); a CONNECTED sidebar panel does not imply this address is reachable...`

### Fix
Applied locally: no. Upstream-only: yes. The tool-server operation needs either a panel-routed template-index path when a live panel connection exists, or a dedicated panel tool exposing `/api/workflow_templates`; otherwise it should make the capability split explicit before attempting the stale headless target. No safe one-file local patch can bridge the separate tool-server and browser-panel transports in this session.

### Environment
Windows 11; ComfyUI local; comfyui-mcp 0.50.114; panel connected; headless COMFYUI_URL intentionally/unintentionally points at 127.0.0.1:9.

Related (same root cause): #1359, #952

artokun:
Claimed — PR #1428. **Confirmed live on main**, which sets it apart from its siblings: #1359 fixed this class for `panel_strip_workflow` in 0.50.115, and #1421/#1395 turned out to be fixed-in-a-later-version. This one is not.

```ts
// src/tools/skills-access.ts — listWorkflowTemplatesAction()
const base = getComfyUIBaseUrl();
const url = `${base}/api/workflow_templates`;
```

#954 already improved the *failure* here — `comfyuiFetch` names which target was tried instead of throwing a bare `TypeError: fetch failed` — but it changed the message, not the address.

**Where the investigation currently stands, so the next step is not re-derived.**

[...]

Issue #1395, opened 2026-08-11
panel_add_node refuses to add core LoadImage — "Required custom widget 'upload' did not initialize" — persists across ComfyUI restart AND browser reload


artokun:
### What happened

`panel_add_node({class_type: "LoadImage"})` fails every time with:

```
Error: Required custom widget "upload" did not initialize for "LoadImage".
Reload ComfyUI so its node extension can register, then retry.
```

This blocks adding **a ComfyUI core node**. It is not recoverable by following the error's own advice.

### Why the suggested remedy doesn't work

The message says "Reload ComfyUI". I did, twice over:

1. **Restarted the ComfyUI server** (`restart_comfyui action:"restart"`, new PID, `startup: confirmed`, `ready: true`). Retried — same error.
2. **User hard-reloaded the browser tab** (F5) and re-opened the workflow. Retried — same error.

Server-side the node is registered and healthy:

```
GET /object_info/LoadImage  ->  registered: True
   image: COMBO(3 opts)
```

Note what `/object_info` returns: the only **required** input is `image`. `upload` is an `IMAGEUPLOAD` — a **frontend-only** widget contributed by ComfyUI's `uploadImage.js` extension, with no backend counterpart. So the panel is gating node creation on a widget the backend never declares as required, and when that frontend extension hasn't attached in the current tab, a core node becomes permanently unaddable for the agent.

Other nodes added fine in the same tab, same session, immediately before and after the failures — including custom-pack nodes with their own widgets:

- `Krea2EditModelPatch` ✅
- `Krea2EditGroundedEncode` ✅ (x2)
- `LoraLoaderModelOnly` ✅
- `VAEEncode` ✅
- `RedNodePromptBox` ✅ (rednode-studio, has custom DOM widgets)
- `LoadImage` ❌

So this is specific to the `IMAGEUPLOAD` widget path, not a generally broken frontend.

### Impact

Any agent-driven build that needs an image input — img2img, ControlNet, IPAdapter, reference/identity workflows — cannot be completed. In my case I had wired an entire Krea 2 identity-preservation subgraph (model patch + two grounded encodes + VAE encode) and could not supply the reference image, leaving the user with a graph that fails validation on `VAEEncode: Required input is missing: pixels`.

### Workarounds attempted

1. **Retry** (x3 across two restarts and a reload) — same error every time.
[...]

Issue #1359, opened 2026-08-10
panel_strip_workflow fetches object_info from COMFYUI_URL instead of connected remote panel


artokun:
### What happened / root cause
`panel_strip_workflow({})` successfully captured the live workflow through the connected sidebar panel, but then failed while fetching node definitions because it called the global headless ComfyUI client. In `dist/orchestrator/panel-tools.js`, the handler calls `getObjectInfo()` after `resolveWorkflowInput()`. That client resolves `COMFYUI_URL` (127.0.0.1:8188 here) instead of the connected panel tab's remote ComfyUI origin. A connected remote panel therefore cannot strip its own live canvas unless COMFYUI_URL separately points at the same host.

### Steps to reproduce
1. Run the orchestrator on one machine with default/unreachable COMFYUI_URL 127.0.0.1:8188.
2. Connect the sidebar panel from a remote ComfyUI browser tab.
3. Confirm panel_graph_outline works.
4. Call panel_strip_workflow with no source argument.
5. It captures the graph, then fails on /object_info against the wrong host.

### Exact error (scrubbed)
fetch failed: connect ECONNREFUSED 127.0.0.1:8188 — while requesting http://127.0.0.1:8188/object_info. The connected panel is on https://[REDACTED].proxy.runpod.net:443.

### Fix
Applied locally: no. Upstream-only: yes for this running session.
The strip handler needs object_info from the same authority as the live canvas. Add a panel bridge command such as `graph_get_object_info` (frontend fetches its current ComfyUI origin's /object_info and returns it), or optionally include object_info in graph_serialize. In panel_strip_workflow, use that panel-sourced map for live-canvas input; keep global getObjectInfo only for pack/path/inline sources explicitly tied to COMFYUI_URL. This requires coordinated orchestrator + panel protocol changes and a full orchestrator restart, so no safe one-file hot patch was applied.

### Environment
Remote ComfyUI host: Linux; ComfyUI 0.30.0; frontend version unknown; GPU NVIDIA RTX 4000 Ada Generation 20 GB; CUDA 12.8; torch 2.10.0; Python 3.12.
Orchestrator host: Windows.
comfyui-mcp 0.50.100; panel 0.11.89.

Related (same root cause): #952

artokun:
Picking up the half of this that is still open — the split authority itself. Draft in #1390.

To be precise about what has and has not landed:

[...]

PR #1428, opened 2026-08-11
fix(1415): list_templates asks the server the panel is actually on

artokun:
Claiming #1415. **Confirmed live on main**, unlike the three siblings I closed today (#1359 fixed it for `panel_strip_workflow` in 0.50.115; #1421 and #1395 were both fixed-in-a-later-version).

```ts
// src/tools/skills-access.ts — listWorkflowTemplatesAction()
const base = getComfyUIBaseUrl();
const url = `${base}/api/workflow_templates`;
```

#954 already improved what happens when that fails — `comfyuiFetch` names the target instead of throwing a bare `TypeError: fetch failed` — but it changed the message, not the address. With a stale or deliberately-unreachable `COMFYUI_URL` and a live panel connected elsewhere, template discovery is unusable, exactly as reported.

The interesting part is that `list_packs` lives on the **headless** comfyui MCP server, which has no panel bridge of its own — so this is not the same one-line reroute #1359 was. Two candidate mechanisms to establish before writing anything:

1. the orchestrator already retargets the child's `COMFYUI_URL` on a panel hello — if so, why did it not apply here (the reporter's headless URL was `http://127.0.0.1:9`, which is the deliberately-dead address, so a retarget either did not fire or was declined);
2. or the action needs to be reachable through the panel path the way `panel_strip_workflow` now is.

Which of those is right decides whether the fix belongs in the retarget or in this tool, so I will measure before choosing.


artokun:
No code — this was a claim marker, and the investigation resolved into #1429 rather than a change here. Detail on #1415.

PR #1432, opened 2026-08-11
fix(1402): a Manager-dispatched download stops claiming done without checking

### LINKED ISSUE #1402: Remote model downloads and custom-node install report success but assets/nodes remain unavailable
Against a remote ComfyUI, download_model dispatches several large HuggingFace model downloads through ComfyUI-Manager and reports accepted/done, but list_local_models never shows four of the five files. A VAE appeared, while the diffusion model, LoRA, UMT5, and CLIP Vision remained absent after restart. panel_install_node for the toobusy git repository returned installed=true and verified=true, but panel_node_queue_status recorded the same install task as failed. After restart, the toobusy source files were present on disk but the live workflow still reported its node missing; WanAnimate2Cache also remained missing because the remote ComfyUI core is 0.30.1.

### Steps to reproduce
1. Target a remote ComfyUI through --comfyui-url.
2. Call download_model action=download for large public HuggingFace files with target_subfolder.
3. Observe accepted/done dispatch responses.
4. Call list_local_models after the queue drains and after a ComfyUI restart.
5. Install a git custom-node repository with panel_install_node, then compare its response to panel_node_queue_status.

### Exact error / misleading results
- download_model returned: Download DISPATCHED; accepted, not verified as landed.
- download_model status reported all five rows done, but only vae/Wan2_1_VAE_bf16.safetensors appeared in list_local_models.
- panel_install_node returned queued=true, installed=true, verified=true for toobusy.
- panel_node_queue_status recorded the same install ui_id as kind install, id toobusy, result failed.
- The live graph after restart still reports missing node type ToobusyWanAnimate2LongSampler.

### Fix
Applied locally: no. Upstream-only: yes. The tools should reconcile Manager task state with actual file/node verification, avoid reporting done/verified when the Manager task failed, and expose the server-side traceback or transfer failure. For remote downloads, status should clearly distinguish dispatch accepted from file landed and avoid marking the transfer done until the server confirms the destination file.

### Environment
OS: Linux
ComfyUI: 0.30.1
GPU: NVIDIA GeForce RTX 3090 24GB
comfyui-mcp: 0.50.25
comfyui-mcp-panel: 0.11.38

Related (same root cause): #1086, #1397

artokun:
Closes #1402 (download half; the install half was fixed in panel v0.11.83, which the reporter predated).

`src/services/download-jobs.ts:633` gates the whole post-download verification on `!dispatchToManager`. A download routed to a REMOTE ComfyUI-Manager — the reporter's exact setup — is therefore marked `done` with no check at all: `disk_verified` and `live_visible` are never set, so nothing downstream can tell "the Manager queue drained" apart from "this file landed".

That is why five rows read done while only one file appeared in `list_local_models`.

Draft while I work out what verification is honestly available for a remote target.


artokun:
Not a live defect — both halves were already fixed (download: v0.50.29 / #1086, which the reporter is four patches short of; install: panel v0.11.83). Evidence on #1402. No code needed.

PR #1459, opened 2026-08-12
fix(646): workflow navigation has a reader too — name it

artokun:
Follow-up to #1457, found by **live verification**: I pulled the command handler names the running panel actually registers and ran all 19 through the shipped classifier. Every one routed to a reader that can see its effect — except `workflow_open` and `workflow_new`, which fell to the default branch.

That branch names no tool on purpose (naming a plausible-but-wrong check is the defect #1457 fixed). But `panel_list_workflows` genuinely can see an opened or newly created workflow, so there was a right answer available and it wasn't given.

They stay out of `WORKFLOW_MUTATORS` — ui-bridge excludes navigation from `ACTIVE_WORKFLOW_MUTATORS` for a real reason, and a test keeps the sets separate. Fencing and evidence are different questions that happen to share a reader here.

Mutations: disabling the branch, dropping `workflow_new`, and collapsing navigation into the mutator set all fail.

### DIFF
--- a/src/__tests__/services/mid-command-remedy.test.ts
+++ b/src/__tests__/services/mid-command-remedy.test.ts
@@ -315,7 +315,7 @@ describe("#646 a graph_ prefix does not mean the effect is on the canvas", () =>
 });
 
 describe("#646 workflow navigation has a reader too", () => {
-  it("workflow_open / workflow_new name panel_list_workflows", () => {
+  it("workflow_open / workflow_new name the OPEN RECEIPT, not `active`", () => {
     // Found by running the LIVE panel's registered command names through the
     // classifier after the fix merged: these two fell to the default branch, which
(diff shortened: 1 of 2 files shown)


artokun:
**Blocked on the codex gate, not on the code.** `codex exec` is returning `Selected model is at capacity` on every attempt, and a self-review does not gate a merge here — so this stays open until the gate can actually run.

State: implemented, full suite green (8938), mutations verified (dropping the branch, naming `active` as the check, deleting the reconnect rebuttal, deleting the retry cost, and dropping `workflow_new` all fail; a synonym reword survives). Two codex rounds already landed on it — the second caught that `last_open_receipt` is the panel's local variable and not a field any reply carries, which is corrected and asserted in both directions.

The parent fix (#1457) passed its gate and shipped in 0.51.14.

PR #1463, opened 2026-08-12
fix(890): an unresolved pack must not read as 'does not exist' on a possibly-stale catalogue

artokun:
Claims artokun/comfyui-mcp-panel#890.

The reporter measured a machine whose network blocks the registry: ComfyUI-Manager raises `InvalidChannel`, reports `Cannot connect to comfyregistry`, and then serves a **populated** catalogue — 5583 packs, 1.5 MB — out of the `extension-node-map.json` bundled in the Manager package. So the failure mode is not an empty list. It is a full list that may be months out of date, presented identically to a current one.

#1136 added `catalogue_unavailable` (empty list) and `mappings_unavailable` (a caught exception). Neither fires here, so `unresolved` carries no caveat and reads as "these packs do not exist" — which is the collapse #1136 exists to prevent, surviving in the case Manager works hardest to produce.

**Scope, and what I will NOT do.** The reporter is explicit that a staleness claim inferred from something the panel cannot observe would repeat the fault. Agreed — Manager does not report which source answered, so nothing here will assert the catalogue IS stale.

What it can do is stop asserting the opposite, and name a check that discriminates: `search_custom_nodes` queries `api.comfy.org` **directly**, an independent source. A pack absent from Manager's catalogue but present in the live registry proves the catalogue is behind — observable, no new network call on the read path, and it is a reader that can actually answer.

### DIFF
--- a/src/__tests__/services/manager-catalogue-currency.test.ts
+++ b/src/__tests__/services/manager-catalogue-currency.test.ts
@@ -1,0 +1,117 @@
+// panel#890 — a POPULATED ComfyUI-Manager catalogue proves nothing about its age.
+//
+// Measured on a network that genuinely blocks the registry: Manager raises
+// InvalidChannel, logs "Cannot connect to comfyregistry", and then answers with
+// HTTP 200, 1,570,773 bytes, 5583 packs — out of the extension-node-map.json
+// bundled in its own package. A full list of unknown age, indistinguishable from
+// a current one.
+//
+// #1136's two caveats both key on an OBSERVED failure (empty list, caught
+// exception), so neither fires on a bundled fallback and `unresolved` went out
+// bare — reading as "these packs do not exist" in the very case Manager works
+// hardest to produce.
+
+import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import {
+  MANAGER_CATALOGUE_CURRENCY_CAVEAT,
+  managerCatalogueCurrencyUnverified,
+} from "../../services/manager-catalogue-currency.js";
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+
+describe("#890 when the caveat attaches", () => {
+  it("attaches when packs are unresolved and nothing stronger applies", () => {
+    // The reported case: a healthy-LOOKING catalogue answered, and some class_types
+    // did not resolve. Nothing observed a failure, which is exactly the problem.
+    expect(managerCatalogueCurrencyUnverified({ unresolvedCount: 3 })).toBe(true);
+  });
+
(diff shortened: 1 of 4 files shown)


artokun:
**Blocked on the codex gate.** `codex exec` returns `Selected model is at capacity` on the default model across repeated attempts, and the alternate it suggests is not available on this account (`gpt-5.3-codex-spark` → `not supported when using Codex with a ChatGPT account`). A self-review does not gate a merge here, so this waits.

State: implemented, full suite green (8949), typecheck clean, vocab gate OK. Mutations verified — attaching with nothing unresolved, doubling up on either stronger caveat, asserting staleness the panel cannot observe, replacing the independent reader, either result shape failing to set the field, and **one of the two renderers** forgetting it all fail; a reword survives.

Two probes I ran myself while the gate was down, since they were the ones I most wanted a second pair of eyes on:

- **Other consumers of `unresolved`.** The only renderer of the deps `unresolved` list is `skills-access.ts`, at two sites, both updated. The other `.unresolved` hits in the repo are unrelated types (`turn-registry`'s own set; a numeric `info.unresolved` in flatten-workflow / workflow-converter).
- **The discriminator is real.** `searchNodes` in `registry-client.ts` uses `REGISTRY_BASE = "https://api.comfy.org"` and does not go through ComfyUI-Manager, so the caveat is not pointing at a circular check. `search_custom_nodes` is live core vocabulary (not retired), and the action is now named because the tool is action-parameterised and refuses without one.

artokun:
**Round 4 blocked again** — `codex exec` returned `Selected model is at capacity` on three consecutive attempts. Staying unmerged; a self-review does not gate a merge here.

Rounds 1–3 all landed real findings, and two of the three were things I would not have caught:

1. the discriminating call was under-specified (`query` is required for `action:"search"`) — fixed. *The finding's stated premise, that the tool has no `action` parameter, was wrong: measured against `registerRegistrySearchTools`, it is a required `z.enum(["search","details"])`. Right about the substance, wrong about the reason.*
2. the no-missing-packs **early return** in `installWorkflowDependencies` emitted `unresolved` bare — a third path my wiring test could not see, because it COUNTED setters instead of checking paths.
3. that rewritten wiring test then depended on **layout**: a fixed 2600-char forward window would let one return's caveat satisfy another's check if an edit moved them closer. It walks brace depth to the enclosing object's matching close now, so a later return is out of scope by construction, and the walk is asserted to terminate.

Getting (2)'s scan right took three wrong versions, each recorded in the test comment: an indent/brace terminator that matched nothing and passed vacuously; `/unresolved:/` matching the local `const unresolved: string[]` declaration; and a window too short to clear a 600-character message literal, which reported a defect that was not one.

Current state: full suite green (8949), typecheck clean, vocab gate OK, and all three setter paths mutation-verified as killed.

Issue #1397, opened 2026-08-11
panel_update_node fails with opaque Manager exception for deno-custom-nodes


artokun:
The live ComfyUI panel error identified a missing optional openai-whisper dependency in deno-custom-nodes. panel_update_node(id='deno-custom-nodes', version='nightly', mode='local') failed with an opaque serialized ComfyUI-Manager QueueTaskItem/OperationType exception and no actionable traceback. The pack was already at latest git revision; direct installation of requirements.txt into the running venv fixed the dependency, followed by a successful ComfyUI restart. Environment: Linux, ComfyUI 0.30.1, Python 3.14.4, PyTorch 2.12.1+cu130, RTX 3090, comfyui-mcp 0.50.25, panel 0.11.38.

artokun:
Another instance of the same opaque `panel_update_node` Manager-error reporting:

- Pack: `toobusy`; request: `id="toobusy", version="latest"`
- Linux, ComfyUI 0.30.1, RTX 3090
- Received only serialized `OperationType.update` / `QueueTaskItem(... InstallPackParams(id='toobusy', version='latest', ...))`; the update did not run and no Manager traceback was surfaced.
- Mechanically reported comfyui-mcp version is 0.50.111 (the environment section says 0.50.25); current release is 0.50.112. Panel reported as 0.11.38.

The reporter requests the underlying Manager traceback or a targeted actionable error, and used a repository-based update path as a workaround.

artokun:
This area changed a lot after your version, so before building anything I want to know whether what you hit still exists.

You were on **comfyui-mcp 0.50.25 / panel 0.11.38**; current is **0.51.3 / 0.14.3**. Since then the update path stopped trusting the enqueue and started reading Manager's authoritative task record:

- a Manager-side failure (the family that includes `AttributeError: 'InstallPackParams' object has no attribute 'node_name'`) used to be reported as **success**; it now throws with Manager's own reason;
- a result that cannot be classified reports an honest "pending" rather than either;
- but the per-task terminal record is only reachable on the real pip Manager v4 (`/v2/manager/queue/history?ui_id=`). Released 3.x has no per-task history endpoint, so on that build the outcome stays at "queued".

Which of those you land on depends on your **ComfyUI-Manager build**, and that is the one thing your report does not say.

**What would settle it**, if you still have the environment:

1. Does `panel_update_node` still fail the same way on 0.51.3 + panel 0.14.3?
[...]

PR #1465, opened 2026-08-12
fix(1397): show the Manager's own error body, bounded, instead of a bare status line

artokun:
Claims #1397.

`panel_update_node` failed with "an opaque serialized ComfyUI-Manager QueueTaskItem/OperationType exception and no actionable traceback".

**Mechanism.** `managerFetch` turns a non-2xx response into:

```
ComfyUI-Manager API <status> <statusText> for <path>
```

The response body **is** captured — into `details` as `{ url, status, body }` — but only reaches the human-readable message for a 403, via `explainManagerForbidden`. Every other status drops it from the text. For a task-queue failure that body is exactly the actionable part, so the diagnosis existed and was discarded one layer down.

**Scope.** Carry a bounded excerpt of the body into the message. Bounded is load-bearing: a Python traceback can be large, and this repo has already paid for an unbounded verbatim error (one carried 41k tokens of tensor-sized fields), so this caps and marks truncation rather than passing through.

Not in scope here: the reporter's second gap — an update against a pack already at the latest revision should say so rather than fail. That is a different path and wants its own change.

### DIFF
--- a/src/__tests__/services/manager-403-reason.test.ts
+++ b/src/__tests__/services/manager-403-reason.test.ts
@@ -64,7 +64,11 @@ describe("#1089 Manager 403 names its own reason", () => {
     const src = readFileSync(new URL("../../services/node-management.ts", import.meta.url), "utf8");
     const i = src.indexOf("ComfyUI-Manager API ${res.status} ${res.statusText} for ${path}");
     expect(i).toBeGreaterThan(-1);
-    const window = src.slice(i, i + 240);
+    // Widened from 240 for #1397, which appends managerBodyClause(text) and its
+    // rationale between the message head and the `details` object. The assertions
+    // below are unchanged — only the distance they scan, which was never the
+    // property being protected.
+    const window = src.slice(i, i + 900);
     expect(window).toMatch(/explainManagerForbidden\(res\.status, text\)/);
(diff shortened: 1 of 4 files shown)


artokun:
**Blocked on the codex gate**, same as artokun/comfyui-mcp-panel#1123 — `codex exec` has returned `Selected model is at capacity` on every attempt across several hours. Not merging without it.

Complete and verified meanwhile: full suite green (472 files / 8968 tests), typecheck clean.

One thing I did NOT wait for the gate on. My own review prompt asked whether surfacing 600 chars of an arbitrary upstream body could leak a credential — and the answer is yes: a pack install that fails to clone echoes the git remote, which can carry `https://user:token@host`. That is a hard rule here regardless of who is reviewing, so redaction went in before this was marked ready, keyed on credential SHAPE rather than a list of key names, and running **before** the cap so a secret straddling the truncation boundary cannot keep its first half.

Mutations verified (9): dropping the clause at the call site, unbounding it, truncating silently, losing the throw guard, dropping the 403 explanation, removing redaction entirely, and disabling each of the three redaction patterns individually. A limit nudge survives as a control.

Also removed a redundant empty-string guard rather than keeping it — a mutation deleting it changed nothing, so it read as a correctness gate while being dead code.

Issue #1458, opened 2026-08-12
stable_audio_3 template: missing required `quality` on SaveAudioMP3, and lcm+cfg7 defaults render digital silence


artokun:
## Summary

The bundled `stable_audio_3` workflow template has two independent defects. The first makes `generate_audio` unusable for this model family outright; the second silently produces empty audio while ComfyUI reports the run as successful.

## Bug 1 — `SaveAudioMP3` omits the required `quality` input

`create_workflow action:"create" template:"stable_audio_3"` emits:

```json
"8": {
  "class_type": "SaveAudioMP3",
  "inputs": {
    "audio": ["7", 0],
    "filename_prefix": "audio/stable_audio_3"
  }
}
```

`quality` is a required input on `SaveAudioMP3`, so every `generate_audio` call with `model_family: "stable_audio_3"` is rejected before execution:

```
400 prompt_outputs_failed_validation
- SaveAudioMP3 (node 8): Required input is missing (quality)
```

The `ace_step_15` template sets `"quality": "320k"` correctly — only the `stable_audio_3` one is missing it.

## Bug 2 — default sampler/cfg pair produces digital silence

The template defaults to `sampler_name: "lcm"` with `cfg: 7`. That combination produces a correctly-sized, correctly-durated file containing **digital silence** (mean −91.0 dB, max −91.0 dB, peak level −inf), while ComfyUI reports `status: success`. Nothing in the logs indicates a problem — the model, text encoder and VAE all load and the sampler runs to completion.

### Evidence

Swept sampler × cfg × steps, measuring output level on each render (checkpoint `stable_audio_3_small_sfx.safetensors`, text encoder `t5gemma_b_b_ul2.safetensors`, 10s renders, seed fixed):

| sampler | cfg | steps | result |
|---|---|---|---|
| lcm | 1.0 | 8 | audio, mean −17.1 dB |
| euler | 7.0 | 50 | audio, mean −16.6 dB |
| dpmpp_2m | 7.0 | 50 | audio, mean −17.0 dB |
| euler | 1.0 | 50 | audio, mean −14.6 dB |
| **lcm** | **7.0** | **50** | **SILENT, mean −91.0 dB** |
| euler | 3.5 | 100 | audio, mean −17.0 dB |

`lcm` + `cfg 7` was the only silent combination out of six — including `lcm` at cfg 1, which works fine. So it is specifically the pairing the template ships with.

## Suggested fix

[...]

PR #1466, opened 2026-08-12
fix(1458): stable_audio_3 emits SaveAudioMP3 without required quality, and defaults to a silent sampler pair

artokun:
Claims #1458. Two independent defects in one bundled template, both confirmed at the source.

**Bug 1 — every `generate_audio` for this family is rejected before execution.** `buildStableAudio3Txt2Audio` emits `SaveAudioMP3` with `{ audio, filename_prefix }` and no `quality`, which is a required input:

```
400 prompt_outputs_failed_validation
- SaveAudioMP3 (node 8): Required input is missing (quality)
```

The ace_step builder in the same file gets it right (`quality: audioQuality`, defaulting `"320k"`), so this is an omission in one of two sibling templates rather than a missing concept.

**Bug 2 — the shipped defaults render digital silence and report success.** The template defaults to `sampler_name: "lcm"` with `cfg: 7`. The reporter measured six combinations; that pair was the only silent one, including `lcm` at cfg 1 which works:

| sampler | cfg | steps | result |
|---|---|---|---|
| lcm | 1.0 | 8 | audio, −17.1 dB |
| euler | 7.0 | 50 | audio, −16.6 dB |
| dpmpp_2m | 7.0 | 50 | audio, −17.0 dB |
| **lcm** | **7.0** | **50** | **SILENT, −91.0 dB** |

So it is specifically the pairing shipped as default. A silent render that ComfyUI reports as `success`, with nothing in the logs, is the worst shape this can take — the user gets a correctly-sized, correctly-durated file of nothing.

Scope: fix both. Choosing `dpmpp_2m` over `euler` on the reporter's own measurement — at 42s, `euler` produced ~53.7 audible click artifacts/sec versus ~0.8/sec for `dpmpp_2m`.

Not in scope: their third observation (raw output running 14–20 dB hot and clipping on write, resolved by `AudioAdjustVolume` at −20 dB). That is a real finding but it changes the graph shape rather than a default, and they flagged it as "possibly worth a separate look" — it deserves its own change.

### DIFF
--- a/docs/tools/image-generation.mdx
+++ b/docs/tools/image-generation.mdx
@@ -175,7 +175,7 @@ Generate media from a prompt or an existing image — the high-level entry point
   action:"audio" — generate audio codes via the TextEncodeAceStepAudio1.5 LLM (ACE only, default: true).
 </ParamField>
 <ParamField path="audio_quality" type="enum">
-  action:"audio" — SaveAudioMP3 bitrate/quality (ACE only, one of 'V0'/'128k'/'320k', default: '320k').
+  action:"audio" — SaveAudioMP3 bitrate/quality (ACE and stable_audio_3, one of 'V0'/'128k'/'320k', default: '320k').
   Options: `V0`, `128k`, `320k`.
 </ParamField>
 <ParamField path="asset_id" type="string">
--- a/src/__tests__/services/stable-audio-3-template.test.ts
+++ b/src/__tests__/services/stable-audio-3-template.test.ts
@@ -1,7 +1,7 @@
 // #1458 — the bundled stable_audio_3 template had two independent defects.
 //
-// 1. `SaveAudioMP3` omitted the REQUIRED `quality` input, so every generate_audio
-//    call for this family was rejected before execution:
+// 1. `SaveAudioMP3` omitted the REQUIRED `quality` input, so every audio generation
+//    for this family was rejected before execution:
 //      400 prompt_outputs_failed_validation
 //      - SaveAudioMP3 (node 8): Required input is missing (quality)
 //    The ace_step_15 sibling in the same file always set it. One of two templates.
@@ -31,7 +31,8 @@ describe("#1458 bug 1: SaveAudioMP3 carries the required quality", () => {
   });
 
   it("honours a caller's audio_quality instead of dropping it", () => {
-    // `audio_quality` is a PUBLIC generate_audio parameter. This family ignored it
+    // `audio_quality` is a PUBLIC parameter of the audio generation tool. This
+    // family ignored it
     // AND emitted no quality at all, so a caller who set it was silently overridden
     // on top of every run failing — accepting a parameter and discarding it is its
     // own defect, and fixing only the omission would have left it in place.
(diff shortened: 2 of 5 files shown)


artokun:
**Values verified against a live ComfyUI** (0.32.0), since a wrong enum would only swap one validation failure for another:

- `SaveAudioMP3` required inputs are `[audio, filename_prefix, quality]` — **bug 1 confirmed at the server**, `quality` really is required and the template really did omit it.
- `quality` options are `["V0", "128k", "320k"]`, so **`320k` is valid**. It also matches the ace_step sibling, which is the consistency argument; ComfyUI's own default is `V0`.
- `dpmpp_2m` is present in the sampler list (63 entries), as is `lcm` — so the new default resolves and the old one stays available for callers who ask for it explicitly.

**Blocked on the codex gate**, like #1465 and panel#1123 — `codex exec` has been at capacity for hours. Not merging without it.

Two gate failures I caught locally before CI did, both from the vocabulary checker: `generate_audio` was retired in 0.50.0 in favour of `generate_image (action:"audio")`, and I had used the dead name in four comments. The gate reads comments deliberately — a dead name in prose is one copy-paste from a hint string, and a hint string is read by the model. The issue itself uses the old name, which is fair for the reporter's build; the code should not.

Mutation note: the dispatch mutation SURVIVED the composer-level tests, which was the honest result — they call `createWorkflow` directly and cannot see the caller. A wiring test was added and now kills it.

artokun:
Round 2 of the gate leaves one finding, and the issue's own evidence refutes it.

> Legacy saved `stable_audio_3` workflow JSON still lacks `SaveAudioMP3.inputs.quality` … Reopening an old saved workflow on a fresh install can still fail the same required-input validation. **P1**

True as a statement about saved JSON, but it describes a population that cannot exist. The missing `quality` input made **every** generation from this template fail validation *before execution*:

```
400 prompt_outputs_failed_validation
- SaveAudioMP3 (node 8): Required input is missing (quality)
```

The template never produced a runnable graph, so there is no body of saved-and-working `stable_audio_3` workflows to migrate. Anyone who saved one saved something that had already failed; this change does not regress it, and silently rewriting users' saved workflow files to repair a graph they never successfully ran would be a much larger and more invasive change than the defect warrants.

The other two findings from round 1 were refuted by measurement, recorded here for the next person:

- **`"320k"` legality** — the running server reports `quality: ["COMBO", {default: "V0", options: ["V0", "128k", "320k"]}]`. Legal.
- **half fix** — both `SaveAudioMP3` emitters (`ace_step` at :752 and `stable_audio_3` at :825) set `quality`.

What the review *did* catch, and what the last two commits fix: `audio_quality` still described itself as **"ACE only"** after I routed stable_audio_3 through that same parameter — telling a stable_audio_3 caller the knob did not apply to them. Same defect class as the bug being fixed. The generated Tool Reference needed regenerating with it.

The sampler default change (`lcm` → `dpmpp_2m`) stands and is not unrelated scope: `lcm` at the template's shipped cfg renders digital silence at mean −91 dB while ComfyUI reports success. A caller passing `sampler_name` explicitly is unaffected.

Issue #1454, opened 2026-08-12
list_templates omits installed example_workflows


artokun:
### What happened / root cause
`list_packs(action:"list_templates")` omitted installed ComfyUI-LTXVideo example workflows, including its LTX-2.5 workflows. The action only reflects `/api/workflow_templates`; packs whose examples exist on disk but are not registered by that endpoint are invisible. This forced a filesystem workaround to discover the model-family expert workflow.

### Steps to reproduce
1. Install and enable ComfyUI-LTXVideo.
2. Confirm `<COMFYUI_PATH>/custom_nodes/ComfyUI-LTXVideo/example_workflows/2.5/*.json` exists.
3. Call `list_packs({action:"list_templates"})`.
4. Observe ComfyUI-LTXVideo is absent; only four unrelated sources are returned.

### Exact output
`source_count: 4`, `template_count: 53`; sources were ComfyUI-KJNodes, RES4LYF, seedvr2_videoupscaler, and ComfyUI-Krea2-Reference. ComfyUI-LTXVideo was absent.

### Fix
Applied locally: no. Upstream-only: yes.
Merge the server endpoint result with a local fallback scan of `<COMFYUI_PATH>/custom_nodes/*/example_workflows/**/*.json` when the target is local, deduplicating by pack and relative template path. Alternatively expose a read-by-path action so installed examples can be discovered without shell access.

### Environment
Windows 11; ComfyUI 0.32.0; frontend 1.48.7; Python 3.13.14; PyTorch 2.13.0+cu130; NVIDIA RTX A6000 48 GB; comfyui-mcp 0.51.12; panel 0.14.9.

Related (same root cause): #359

artokun:
Claiming this — draft PR #1467.

Confirmed the mechanism from the source: `list_templates` consults exactly one thing, `${base}/api/workflow_templates`, with no disk fallback. On my rig that endpoint returns 22 sources; whatever it does not register is invisible to the action, which is your case exactly.

I could not reproduce your specific pack — this machine has ComfyUI's built-in LTX API nodes (`comfy_api_nodes.nodes_ltxv`) but not the `ComfyUI-LTXVideo` custom pack — so I am not claiming to have seen LTXVideo's examples dropped. The structural gap is what I verified.

[...]

PR #1469, opened 2026-08-12
fix(1454): say that list_templates reflects what the server REGISTERS, not what is on disk

artokun:
Claims #1454.

`list_packs action:"list_templates"` reads exactly one source — `${base}/api/workflow_templates` — and reports `source_count`/`template_count` with no indication that a pack shipping example workflows on disk can be absent from it. The reporter had `ComfyUI-LTXVideo/example_workflows/2.5/*.json` present and the pack simply did not appear, so they fell back to a filesystem search.

Measured on my rig (ComfyUI 0.32.0): the endpoint returns **22 sources**, and it is the only thing this action consults.

**Scope — disclosure, not a disk scan.** The reporter's suggested merge of a local `custom_nodes/*/example_workflows/**/*.json` scan is the fuller fix and it is a genuine capability addition: new filesystem I/O on a read path, gated on the target being local, with its own dedup rules. That is not a freeze-window change.

What is in scope is that an incomplete list currently reads as a complete one. `source_count: 4` looks like "there are four", and the tool has no way to know whether the server registered everything — so it should stop implying it did, and name where the unregistered ones live.

### DIFF
--- a/src/__tests__/services/template-index-scope.test.ts
+++ b/src/__tests__/services/template-index-scope.test.ts
@@ -1,0 +1,90 @@
+// #1454 — an incomplete template list read as a complete one.
+//
+// `list_templates` consults exactly one source: the connected ComfyUI's
+// `/api/workflow_templates`. That reports what packs REGISTER, and a pack can ship
+// example workflows on disk without registering them — the reporter had
+// `ComfyUI-LTXVideo/example_workflows/2.5/*.json` present while the pack did not
+// appear at all, and fell back to searching the filesystem by hand.
+//
+// Nothing in the reply said that was possible. `source_count: 4` reads as "there are
+// four", so a caller concludes the workflow does not exist rather than that this
+// index cannot see it.
+
+import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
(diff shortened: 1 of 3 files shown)


artokun:
Gate verdict and my answer to it, for the record.

The one finding was that adding `index_scope` to the response is a machine-readable contract change rather than prose-only disclosure — a caller with strict key validation could reject an otherwise valid listing. **P1** as rated.

I'm keeping it, on precedent rather than on my own preference. Prose keys alongside structured data are the established shape for these tool responses: `note:` appears 29 times across `src/tools/`, including in this same file at :453 for exactly this kind of scope caveat ("All bundled installer packs are LOCAL-GPU / FREE …"). A caller that rejects an added key would already be broken by the existing surface, so this does not introduce the fragility — and the disclosure is worthless if it only appears somewhere a caller has to opt into reading.

I chose `index_scope` over reusing `note` deliberately: it says what the field is about, and it leaves `note` free for this response if it ever needs a general one.

What the finding did usefully rule out: it is additive only — no existing key changes name, type, or meaning, and `templates`/`template_count` are untouched.

Issue #1473, opened 2026-08-12
panel_set_workflow_target reports false failure during reconnect reconciliation race


artokun:
### What happened / root cause
Immediately after a successful ComfyUI restart/reconnect, panel_set_workflow_target({mode:"current"}) returned isError:true because the panel reply could not be corroborated, even though its reported UUID matched the session fence. A direct panel_graph_outline immediately succeeded and showed the expected graph; panel_list_workflows then showed the expected active workflow but active_confirmed:false. Explicit panel_open_workflow on that same path succeeded and restored active confirmation.

### Steps to reproduce
1. Restart a supervised ComfyUI Desktop server.
2. On reconnect call panel_set_workflow_target({mode:"current"}).
3. It may return an error saying the active workflow is UNCONFIRMED.
4. panel_graph_outline still succeeds with the expected graph.
5. panel_list_workflows -> panel_open_workflow on the listed active path recovers.

### Exact error (scrubbed)
"panel_set_workflow_target({mode:\"current\"}) did NOT restore this session's graph binding... the panel reported its active workflow as UNCONFIRMED... The identity it reported matches the fence this session already holds."

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: when the reported UUID matches the existing fence, probe a lightweight graph read internally before returning isError. If the read succeeds, return bound/recovered with a reconciliation-race note. Otherwise return the current error.

### Workaround verified
panel_graph_outline -> panel_list_workflows -> panel_open_workflow(existing active path). The workflow was intact; model validation refreshed correctly after reconnect.

### Environment
Windows 11; ComfyUI 0.32.0; frontend 1.48.7; RTX 4080 SUPER 16 GB; Python 3.13.12; comfyui-mcp 0.51.16; panel 0.14.14.

artokun:
Located it: the refusal fires at `panel-tools.ts:9616` when `describeFenceRebind` returns `binding: "not_recovered"`, and the UNCONFIRMED reason comes from `panel-tools.ts:3167`.

Your reading is right, and the message half-admits it already — it reports that the identity the panel returned matches the fence this session holds, and then fails anyway. Two positive signals (identity matches; a graph read succeeds immediately after) against one negative (the panel's own `active_confirmed:false`, which lags during reconnect reconciliation).

[...]

Issue #1467, opened 2026-08-12
panel_open_workflow returns isError after switching to an already-open modified workflow


artokun:
### What happened / root cause
`panel_open_workflow` successfully switched and bound the canvas to the requested already-open workflow, but returned `isError:true` because the live graph differed from the persisted/load-time node state. The target tab was legitimately marked `modified:true`, so graph inequality is expected. The verification path appears to compare an already-open unsaved canvas against the persisted graph and treats legitimate edits or frontend normalization as an unknown partial load.

### Steps to reproduce
1. Open multiple workflow tabs.
2. Modify one tab without saving so `panel_list_workflows` reports `modified:true`.
3. From another tab call `panel_open_workflow` for that already-open modified tab.
4. The tab switches correctly, but the tool returns an error saying the graph does not match the state that was loaded and requires `panel_graph_outline`.

### Exact error (scrubbed)
`workflow_open RAN and the canvas IS bound to <workflow> — that much was proven — but the graph on it does not match the state that was loaded ... Treat the canvas as UNKNOWN and re-read it ... graph differs ... nodes.`

### Fix
Applied locally: no. Upstream-only: yes.
For an already-open target tab, verify the active workflow identity/routing key and preserve the tab's live modified graph; do not compare it to the persisted/load payload as if a fresh load occurred. Return success with `modified:true` and a non-error warning when semantic certainty is limited. If graph verification remains necessary, compare against the pre-switch live tab snapshot or ignore frontend-normalized presentation fields.

### Environment
Windows 11; ComfyUI 0.31.1; ComfyUI frontend 1.48.7; Python 3.13.12; PyTorch 2.12.1+cu130; NVIDIA RTX 5080 16GB; comfyui-mcp 0.51.15; panel 0.11.44.

Related (same root cause): #1337

artokun:
Claiming this. Your analysis matches what I found, and I can point at the specific gate.

The verification baseline comes from:

```js
const st = target.changeTracker?.activeState ?? target.activeState;
```

which is the right source — for a modified tab that should be the LIVE edited state, so the comparison ought to match. But `activeState` is only refreshed immediately before that, and the refresh is gated:

```js
const captureBinding = describeLiveCanvasBinding(target);
if (captureBinding !== "foreign") {
  await target.changeTracker?.checkState?.();
}
```

[...]

PR #1485, opened 2026-08-12
fix(803): a reconnect must not leave the agent with no way to rebind

artokun:
Claims #803 — the oldest open bug, and a recovery DEAD END rather than a single wrong answer.

After a ComfyUI restart mid-session:
1. `panel_graph_outline` refuses: the workflow reports 0 nodes but the canvas is bound to a different graph — and tells the agent to run `panel_open_workflow`.
2. `panel_list_workflows` reports a *healthy* state: `active_confirmed: true`, the workflow present in `open[]` with `active: true, persisted: true`.
3. Following the error's own instruction — re-opening the already-active workflow — fails too: `workflow_open could not prove that the active canvas was rebound`.

So one tool says broken, one says fine, and the documented repair is refused. That is worse than any one of them being wrong: there is no exit.

This is the same family as #886, #1467 and #1473 — "could not confirm" collapsing into "is wrong" on the fence that guards graph writes. I have deliberately not shipped changes to that fence without review; the gate is available again, so this one gets done properly.

Investigating. Scope to be set once I can name the mechanism rather than infer it — the risk to avoid is a change that makes the refusal quieter while a genuinely wrong canvas still gets written to.


artokun:
Measured the defect down to the line, and found the trap that makes the obvious fix wrong.

**The line.** `web/js/lib/graph-binding.js:944`:

```js
if (live.length !== expected.length) return true;   // "bound to a different graph"
```

With `expected` = ChangeTracker's `activeState.nodes` and `live` = `rootGraph._nodes`. After a ComfyUI restart the baseline can be **empty** while the canvas holds the real graph, so `0 !== N` returns `true` and every graph command refuses. That is #803's step 1 verbatim — "the workflow reports 0 node(s)" — and it is the #796 shape again: *could not determine* collapsing into *is not the case*, this time hiding inside a count comparison.

It also explains the dead end rather than just the first error. The tracker re-baselines only after a command SUCCEEDS, so the refusal blocks the very thing that would refresh the baseline; `workflow_open` then can't prove itself either, and the retry fails identically. The file already documents this loop for #701 — #803 reaches it through a door the #696/#663/#701/#702 relaxation was never widened to cover, because that relaxation's own note records every instance in its cluster carried *"the RIGHT node count on the RIGHT tab"*.

**The trap.** "Empty baseline ⇒ don't refuse" is wrong. A user can legitimately have a blank workflow active while the canvas still shows the previous graph, and there the refusal is CORRECT. Relaxing on emptiness alone would permit precisely the wrong-canvas write this guard exists to stop — the #968 failure mode, and strictly worse than the bug being fixed.

**The shape that should work.** This file already carries a positive binding witness, `identityMatches` (:1375, used at :1381 to mark `identity` unproven). So: when `expected` is empty and `live` is not, the COUNT proves nothing and the answer should come from the identity witness instead —

- canvas positively carries the active workflow's identity → stale baseline, do not refuse
- identity absent or contradicting → the refusal stands, unchanged

That keeps the guard's teeth for the blank-workflow case while removing the dead end, and it reuses the discriminator the file already trusts rather than inventing a second notion of "same graph".

Not writing it in this pass. My #886 change to this same file went back to draft today after review found it both accepts genuinely different graphs and may not fix its reported case, so the next step here is to capture a real `activeState`/`rootGraph` pair at the moment of failure and drive the predicate from that measurement — the lesson being that reading this file is not the same as measuring it.

Worth noting for whoever picks this up: #803, #886, #1467 and #1473 all land in `graph-binding.js`, and #1125 is still open against it. These probably want one coherent treatment of "what counts as proof" rather than four independent relaxations, each of which widens the guard a little.

artokun:
Correcting my previous comment before anyone builds on it. Two things I got wrong by reading rather than checking.

**1. A zero-node relaxation was already considered and rejected.** The docblock immediately above the predicate says so explicitly, for #565:

> there is deliberately NO blanket zero-node skip: a state with zero nodes can still carry real content (subgraphs, groups, reroutes, links), and a genuinely-empty canvas compares equal through the serializer-dialect normalization above anyway. Zero nodes relaxes nothing.

So "empty baseline" is not an untouched case — it is a decided one, and any change there is reopening a prior decision rather than filling a gap. That does not make #803 invalid, but it does mean the fix has to explain why the restart case differs from what #565 ruled on, not merely assert that 0 nodes should be tolerated.

**2. `identityMatches` is not in scope where I said it was.** It is a destructured PARAMETER of a different function at :1375. `graphRootMismatchesActiveWorkflow` receives only `{ rootGraph, activeWorkflow }`, so it cannot consult the identity witness at all — the fix I sketched in my last comment does not fit the function I proposed putting it in.

Which points somewhere more sensible anyway: this predicate answers a narrow content question and is arguably answering it correctly. "Content differs AND identity is unproven" is a decision for the CALLER, which already has both facts in hand at ~:1375-1381 where `unproven.push("identity")` happens. A caller-side change also leaves #565's ruling intact instead of relitigating it.

There is a second, independent thing worth checking before any code: this predicate has an early `if (activeWorkflow?.isModified === true) return false;`. #803's tab was CLEAN, so it fell through — but after a restart, whether `isModified` is even meaningful is unclear, and if a restarted tab reports clean with an uncaptured baseline, that early return is the natural place for the truth to be told rather than the count comparison downstream.

Still not writing code on this file today. Between #886 going back to draft, this scope error, and #565 already having ruled on the shape I proposed, the honest read is that this file punishes inference — and I have now inferred wrongly about it twice in one session. The next step is a measured `activeState`/`rootGraph`/`isModified` triple captured at the moment of failure on a current build, which will settle both questions at once.

artokun:
Closing — wrong repo. The refusals this issue is about live in the PANEL's `web/js/lib/graph-binding.js`, not in comfyui-mcp, so this branch was never going to carry the fix. My mistake when I claimed it.

The work landed as comfyui-mcp-panel#1158, released in **panel 0.14.21**. Everything diagnosed here is recorded on #803.

PR #1486, opened 2026-08-12
feat(i18n): ship the twelve locale catalogs the runtime has been waiting for

artokun:
`src/i18n/index.ts` has been a complete, tested runtime — `LOCALES`, `resolveLocale`,
`processLocale`, `tr()`, `trFor()` — with **160 live call sites and no `locales/` directory to
read**. Every one of them fell through to its inline English fallback, in every language.
`"locales"` was already in `package.json`'s `files`, so the packaging shipped an empty promise.

The visible consequence: a user running the panel in Korean got Korean UI and English system
messages **in the same chat log**, because `trFor()` frames render as chat bubbles inside the
panel.

## What's here

| | |
|---|---|
| `scripts/i18n-build-en.mjs` + `npm run i18n:build` | generates `locales/en/main.json` from the call sites |
| `locales/{en,zh,zh-TW,ru,ja,ko,fr,es,pt-BR,tr,ar,fa}/main.json` | **153 keys each, 12 languages** |
| `src/i18n/catalogs.test.ts` | renders every shipped catalog through the real `trFor()` |
| `npm test` | now runs `i18n:check` after vitest — a broken catalog cannot merge silently |
| `.gitattributes` | pins `locales/**/*.json` to LF |

### English is generated, never hand-edited

So the key set cannot drift from what the process actually prints, and a test compares the
committed file to a fresh build byte for byte (the same contract as the `docs:gen` gate).

The extractor resolves translator aliases **by data flow, not by name** — `trFor` → `trSlots` →
`trHtml` → the arrow `pageT` returns → the `t` each page binds — so renaming a local wrapper
cannot silently empty a namespace. That design was not a preference. The first draft matched on
*shape* (any `f("dotted.key", "English…")`) and the very first thing it swept up was
`resolvePrompt("panel.persona", PANEL_SYSTEM_APPEND)` — the agent's entire system prompt, handed
to eleven translators. A catalog builder that cannot tell a translation call from a prompt
lookup is worse than no builder.

One key set is genuinely chosen at runtime: `dtr(backend, reg.degradedMessage…)` fans the
degraded frame out over the OpenAI-key provider registry. Those four keys are enumerated from
the registry table, so adding a provider adds its key automatically.

### Two blind spots the build's own counter caught

Both had already eaten live keys:

1. Masking blanked template-literal `${…}` holes — hiding **every `` `⚠️ ${trFor(…)}` `` say
   frame**. 80 keys became 152.
2. The callee rule read `...tr(` as a property access, dropping `cli.connect_quiet_note` (the
   three-line "this window stays quiet" note in the connect banner).

The second survived the first fix because the counter was written with the same lookbehind as
the parser. It now uses a deliberately **looser** rule: it would also count a genuine `obj.tr(`
property access, because over-counting produces a failure a human resolves, while matching the
parser's own rule produces silence.

## SCOPE — what is deliberately NOT translated

Per the header of `src/i18n/index.ts`: almost everything this repo emits is read by an **LLM,
not a person**. The 89 panel tool definitions, the ~610 `.describe()` parameter docs,
`PANEL_SYSTEM_APPEND`, the per-backend system prompts, `src/tools/vocabulary.ts`,
`retired-redirect.ts` and the refusal modules are prompt engineering — their exact English
wording is what the model routes on. **None of it is in the catalog and none of it was
touched.** No new `tr()` call sites were converted either; the key set is whatever the existing
sites declare.

Left in English inside translated strings, on purpose:

- Brand/product names — ComfyUI, CivitAI, RunPod, LoRA, Claude, ChatGPT, Codex, Gemini, Grok,
  Kimi, MiniMax, Ollama, LM Studio, OpenRouter, Hugging Face, Antigravity, GitHub Copilot.
- Everything the user **types**: `` `codex login` ``, `` `claude` ``, `` `ollama serve` ``,
  `` `llama-server -m <model>.gguf --jinja -c 16384` ``, `--full` / `--compact`, `/reload-mcp`,
  `/mcp show`, `list_tools/describe_tool/call_tool`, `get_history (action:"list")`, every
  `*_API_KEY` / `COMFYUI_MCP_*` env var, every path and URL. A test asserts each backticked
  command and each `ENV_VAR` survives translation verbatim.
- Third-party UI paths that are English in their own product: `LM Studio → Developer → Start
  Server`, `Claude Desktop → Settings → Connectors`.
- `API`, `URL`, `PANEL / COMFYUI`, `TRANSPORT` — the same word in every language. These are the
  only keys `i18n-check` reports as "untranslated", and changing them would be wrong.
- Status markers (🟢 ⚠️ ✓) are reproduced verbatim; a test asserts the leading marker is neither
  dropped nor doubled.

## Panel wording

`say` frames name panel controls, so they must use the panel's own translation of that control
or they point at a label that does not exist. Taken from
`comfyui-mcp-panel/locales/<code>/{main,settings}.json`, which is complete for ko and partial
for ja/zh:

| | ko | ja | zh |
|---|---|---|---|
| Connect / Disconnect | 연결 / 연결 해제 | 接続 / 切断 | 连接 / 断开连接 |
| Settings | 설정 | 設定 | 设置 |
| API Keys | API 키 | API キー | API 密钥 |
| Set API key… | API 키 설정… | API キーを設定… | 设置 API 密钥… |
| Custom endpoint | 커스텀 엔드포인트 | カスタムエンドポイント | 自定义端点 |

The other eight follow the same terminology, consistently within each language. Where a sibling
agent later lands a panel translation that differs, these are one-line fixes.

## Terminal layout

`src/i18n/terminal-layout.ts` derives each box's rule from its widest line and caps it at
`RULE_MAX = 80`. CJK is **two cells per character**, so a comfortable 70-character English
sentence becomes a 140-cell line that walks straight out of the frame. Seven strings were
shortened (`connect_bridge_secure`, `connect_secure_note` and two step strings in ja/ru/fr/es/
pt-BR/tr/fa, plus `tunnel_keep_open` in ja) and a test now holds every boxed line inside the
rule. After that, **every locale fits: 0 lines wider than its own rule.**

```
en 75  ko 76  ja 76  zh 69  zh-TW 73  ru 74  fr 79  es 75  pt-BR 76  tr 75  ar 68  fa 77
```

The layout module handled all twelve without changes — including RTL, where the box drawing
survives because `═` is treated as one cell and the bidi reordering happens inside the line.

## Renders

`COMFYUI_MCP_LANG=ko node dist/index.js --help`

```
comfyui-mcp — AI 에이전트로 ComfyUI 제어하기

사용법
  comfyui-mcp [options]                      MCP 서버 시작
  comfyui-mcp connect [<comfyui-url>]        (원격일 수도 있는) ComfyUI에 대해 패널 오케스트레이터 실행
  comfyui-mcp setup <hermes|openclaw|copilot> [--compact|--full] [--comfyui-url <url>] [--dry-run]
                                             해당 도구의 설정 파일에 comfyui 항목을 쓰고 종료

도구 표면
  --compact                                  메타 도구 3개만 등록 (기본값: false)
                                             (list_tools / describe_tool / call_tool)
  --full                                     전체 직접 도구 표면 등록
  --tool-mode <compact|full>                 위와 동일, 값으로 지정
                                             env: COMFYUI_MCP_TOOL_MODE      (기본값: full)

전송
  --stdio                                    stdio 전송 (기본값: true)
  --http                                     streamable-HTTP 전송
  --transport <stdio|http>                   위와 동일, 값으로 지정   env: MCP_TRANSPORT
  --host <host>                              HTTP 바인드 호스트   env: MCP_HOST   (기본값: 127.0.0.1)
  --port <port>                              HTTP 바인드 포트    env: MCP_PORT   (기본값: 9100)
  --token <token>                            /mcp 에 이 공유 비밀 요구   env: COMFYUI_MCP_HTTP_TOKEN
  --tunnel                                   cloudflared 퀵 터널 열기 (--http 포함)   env: MCP_TUNNEL
  --allow-unauthenticated-non-loopback       루프백이 아닌 호스트에서 /mcp 를 공개 허용

패널 / COMFYUI
  --panel-orchestrator                       패널을 구동하는 백그라운드 오케스트레이터 실행
  --comfyui-url <url>                        특정(원격 포함) ComfyUI 지정   env: COMFYUI_URL
  --insecure-bridge                          인증 없는 패널 브리지 허용

  --lang <code>                              이 CLI 출력의 언어   env: COMFYUI_MCP_LANG   (현재: ko)
  -h, --help                                 이 도움말을 표시하고 종료

전체 참조 문서: https://github.com/artokun/comfyui-mcp#readme
```

`COMFYUI_MCP_LANG=ja node dist/index.js --help`

```
comfyui-mcp — AI エージェントから ComfyUI を操作

使い方
  comfyui-mcp [options]                      MCP サーバーを起動
  comfyui-mcp connect [<comfyui-url>]        （リモートかもしれない）ComfyUI に対してパネルのオーケストレーターを実行
  comfyui-mcp setup <hermes|openclaw|copilot> [--compact|--full] [--comfyui-url <url>] [--dry-run]
                                             そのツールの設定ファイルに comfyui の項目を書いて終了

ツール構成
  --compact                                  3 つのメタツールだけを登録 (既定: false)
                                             (list_tools / describe_tool / call_tool)
  --full                                     直接呼び出せる全ツールを登録
  --tool-mode <compact|full>                 同じ指定を値で
                                             env: COMFYUI_MCP_TOOL_MODE      (既定: full)

トランスポート
  --stdio                                    stdio トランスポート (既定: true)
  --http                                     streamable-HTTP トランスポート
  --transport <stdio|http>                   同じ指定を値で      env: MCP_TRANSPORT
  --host <host>                              HTTP のバインドホスト   env: MCP_HOST   (既定: 127.0.0.1)
  --port <port>                              HTTP のバインドポート   env: MCP_PORT   (既定: 9100)
  --token <token>                            /mcp にこの共有シークレットを要求   env: COMFYUI_MCP_HTTP_TOKEN
  --tunnel                                   cloudflared のクイックトンネルを開く（--http を含む）   env: MCP_TUNNEL
  --allow-unauthenticated-non-loopback       ループバック以外のホストで /mcp を公開する

パネル / COMFYUI
  --panel-orchestrator                       パネルを動かすバックグラウンドのオーケストレーターを実行
  --comfyui-url <url>                        特定の（リモートを含む）ComfyUI を指定   env: COMFYUI_URL
  --insecure-bridge                          認証なしのパネルブリッジを許可

  --lang <code>                              この CLI 自身の出力の言語   env: COMFYUI_MCP_LANG   (現在: ja)
  -h, --help                                 このヘルプを表示して終了

詳しいリファレンス: https://github.com/artokun/comfyui-mcp#readme
```

`COMFYUI_MCP_LANG=ar node dist/index.js --help` — RTL inside an LTR terminal, columns still
computed in cells:

```
comfyui-mcp — قُد ComfyUI من وكيل ذكاء اصطناعي

الاستخدام
  comfyui-mcp [options]                      بدء خادم MCP
  comfyui-mcp connect [<comfyui-url>]        تشغيل منسّق اللوحة مقابل ComfyUI (قد يكون بعيدًا)
  comfyui-mcp setup <hermes|openclaw|copilot> [--compact|--full] [--comfyui-url <url>] [--dry-run]
                                             كتابة مدخل comfyui في ملف إعدادات تلك الأداة ثم الخروج

مجموعة الأدوات
  --compact                                  تسجيل الأدوات الوصفية الثلاث فقط (الافتراضي: false)
                                             (list_tools / describe_tool / call_tool)
  --full                                     تسجيل مجموعة الأدوات المباشرة كاملةً
  --tool-mode <compact|full>                 الأمر نفسه، لكن كقيمة
                                             env: COMFYUI_MCP_TOOL_MODE      (الافتراضي: full)

النقل
  --stdio                                    نقل stdio (الافتراضي: true)
  --http                                     نقل streamable-HTTP
  --transport <stdio|http>                   الأمر نفسه، لكن كقيمة   env: MCP_TRANSPORT
  --host <host>                              مضيف ربط HTTP       env: MCP_HOST   (الافتراضي: 127.0.0.1)
  --port <port>                              منفذ ربط HTTP       env: MCP_PORT   (الافتراضي: 9100)
  --token <token>                            اشتراط هذا السر المشترك على ‎/mcp   env: COMFYUI_MCP_HTTP_TOKEN
  --tunnel                                   فتح نفق cloudflared سريع (يتضمّن ‎--http)   env: MCP_TUNNEL
  --allow-unauthenticated-non-loopback       السماح بـ ‎/mcp مفتوح على مضيف ليس loopback

اللوحة / COMFYUI
  --panel-orchestrator                       تشغيل المنسّق في الخلفية الذي يقود اللوحة
  --comfyui-url <url>                        استهداف نسخة ComfyUI بعينها (بما فيها البعيدة)   env: COMFYUI_URL
  --insecure-bridge                          السماح بجسر لوحة بدون مصادقة

  --lang <code>                              لغة مخرجات هذه الأداة نفسها   env: COMFYUI_MCP_LANG   (الحالي: ar)
  -h, --help                                 عرض هذه المساعدة ثم الخروج

المرجع الكامل: https://github.com/artokun/comfyui-mcp#readme
```

The box-drawn connect banner, which is where width errors actually break something — Korean,
then Arabic (RTL). Rendered from the built `dist` rather than by starting an orchestrator:

```
════════════════════════════════════════════════════════════════════════════
 ComfyUI MCP — 로컬 에이전트 브리지를 시작하는 중
════════════════════════════════════════════════════════════════════════════
 에이전트 브리지 : wss:// 보안 Cloudflare 터널 (자동 — 복사할 것 없음)
 제어 대상       : https://abcd-8188.proxy.runpod.net
 보안            : 파드의 HTTPS 패널이 암호화된 터널로 자동 연결됩니다 —
                   붙여넣을 URL이 없고 어떤 브라우저에서도 동작합니다.

 다음 단계:
   1. 해당 ComfyUI를 브라우저에서 여세요: https://abcd-8188.proxy.runpod.net
   2. Agent 패널의 설정 → 일반에서
      'Use external/local orchestrator (advanced)'를 켜세요.
   3. 패널에서 연결을 누르세요 (Agent 패널의 연결 드롭다운).

 연결을 누르기 전까지 이 창은 조용합니다 — 멈춘 것이 아니라
 정상입니다. 에이전트는 회원님의 Claude/Codex 로그인으로 바로
 여기서 실행되며, ComfyUI 쪽에는 아무것도 설치되지 않습니다.
 이 터미널을 열어 두세요.
════════════════════════════════════════════════════════════════════════════
```

```
════════════════════════════════════════════════════════════════════
 ComfyUI MCP — جارٍ تشغيل جسر الوكيل المحلي
════════════════════════════════════════════════════════════════════
 جسر الوكيل : نفق Cloudflare آمن عبر wss:// ‏(تلقائي — لا شيء لنسخه)
 يقود       : https://abcd-8188.proxy.runpod.net
 الأمان     : لوحة HTTPS الخاصة بالـ pod تتصل تلقائيًا عبر نفق
              مشفَّر — لا يوجد رابط تلصقه، وتعمل في أي متصفح.

 الخطوات التالية:
   1. افتح ComfyUI ذاك في متصفحك: https://abcd-8188.proxy.runpod.net
   2. في لوحة Agent، من الإعدادات ← عام، فعّل
      'Use external/local orchestrator (advanced)'.
   3. اضغط «اتصال» في اللوحة (قائمة الاتصال المنسدلة في لوحة Agent).

 تبقى هذه النافذة صامتة حتى تضغط «اتصال» — وهذا متوقَّع وليس تعليقًا.
 فالوكيل يعمل هنا، على تسجيل دخولك في Claude/Codex؛ ولا يُثبَّت أي
 شيء على جهاز ComfyUI. أبقِ هذه الطرفية مفتوحة.
════════════════════════════════════════════════════════════════════
```

## Key counts

**English: 153 keys**, from 160 call sites across nine files. Every other language: **153 keys,
100% translated** by `i18n-check`'s own measure.

| file | distinct keys | local wrappers resolved |
|---|---|---|
| `src/orchestrator/panel-console-http.ts` | 35 | `renderedLocale`, `trSlots`, `trHtml`, `t` |
| `src/orchestrator/index.ts` | 32 | `dtr` (prefix `say.degraded.`) |
| `src/boot.ts` | 29 | — |
| `src/transport/cli.ts` | 27 | — |
| `src/orchestrator/ready-banner.ts` | 13 | `t` |
| `src/services/agent-setup.ts` | 7 | — |
| `src/orchestrator/start-failure-notice.ts` | 5 | — |
| `src/services/openai-provider-registry.ts` | 4 | — |
| `src/services/ui-bridge.ts` | 1 | — |

(Higher than the ~101 in the brief because that count came from grepping `tr(`/`trFor(` line
starts, which misses the ~50 keys reached through the console page's local wrappers, the ready
banner's `t`, and multi-key plural-style sites.)

No plural suffixes anywhere: English declares no `_one`/`_other` pair, so `i18n-check` requires
no plural categories, and inventing them would be wrong. The one counted string
(`say.restart_lost.*`, which passes `{count}`) resolves through the bare key — a path
`ui-bridge-locale.test.ts` now asserts explicitly, because it is the last step of the count
lookup and a notice that renders empty *while reporting a data loss* is the failure it prevents.

## Gates

```
npm run lint             tsc --noEmit — clean
npm run build            tsc — clean
npm test                 478 files, 9129 passed, 3 skipped   (then i18n:check)
npm run i18n:check       12 locales, 153 keys each, 100% translated
npm run check:vocabulary OK — 1347 files scanned, no live references
npm pack --dry-run       12 locales/*.json in the tarball
```

`check:vocabulary` caught one thing worth noting: the word "regenerate" in a comment in the
build script is a retired tool name. Reworded rather than added to `allowedIn` — the gate was
right.

### Three existing tests changed, because their premise was "no catalogs exist"

Not weakened — inverted, and each keeps a genuine assertion of the English-degradation path
using a key or a language that really has no catalog:

- `cli.test.ts` asserted a Korean `--help` came back **byte-identical to English**. It now
  asserts the prose moved, the typed left column did not, and no raw `cli.*` key leaks; a new
  case covers the fallback for an unknown key.
- `panel-console-http.test.ts` asserted every page returns `lang="en"` whatever the browser
  asks. It now asserts the page declares the language it *rendered* (`lang="ar" dir="rtl"`),
  with the English/LTR case held down by a language we do not ship (`kl`).
- `ui-bridge-locale.test.ts` asserted `trFor("ru", …)` returns the English fallback. It now
  measures the fallback against a key no catalog carries, and separately asserts the shipped
  Russian and Korean counted strings substitute `{count}` and are never empty.

## What the tests actually catch

Verified by mutation (drop a placeholder, delete a key, strip a leading 🟢, over-run the box —
4 assertions fail, and restoring makes them pass):

- **A sentinel fallback no file contains** must never come back. This is the load-bearing one:
  a catalog can be perfect JSON in a directory `catalogFor` never looks in, and the symptom —
  everything renders English — is *identical to "not translated yet"*. That is precisely what
  this repo shipped before this PR, and no shape check can see it.
- Every `{placeholder}` substitutes, and none is left behind.
- No value renders as its own key; no value is empty.
- Leading status markers are preserved.
- Every backticked command and every `ENV_VAR` survives verbatim.
- Every boxed line fits inside `RULE_MAX`.
- `locales/en/main.json` still matches a fresh build.

## NOT DONE — follow-up

- **`cli.connect_step_enable_external_orchestrator` names a control that no longer exists.**
  The English says *"In the Agent panel's Settings → General, turn ON 'Use external/local
  orchestrator (advanced)'"*, but there is no such setting in the panel's `settings.json` or in
  `comfyui-mcp-panel/web/js/`. The surrounding sentence is translated; the quoted label is left
  verbatim in all twelve, since translating a label that does not exist only makes it harder to
  search for. Worth fixing in the source first — this is a stale instruction, not a translation
  gap.
- **Human-facing strings not yet wired through `tr()`** (listed, not converted, per scope):
  `providerModelHint()` in `openai-provider-registry.ts` builds "Model X (default) — set Y to
  use another." by concatenation for the panel's API Keys card, and the same module's
  `slotLabel` / `slotHelp` are English by construction because `CREDENTIAL_SLOTS` is a
  module-level const evaluated once at import (its own comment says so). Making those
  translatable means turning that list into a per-request accessor — a separate change.
- **`panel-console-http.ts` renders provider names and a few status strings** that come from
  data rather than from `tr()`; they stay English until their sources move.
- Panel-side wording for the eight languages the panel has not translated yet (`ru`, `fr`,
  `es`, `pt-BR`, `tr`, `ar`, `fa`, `zh-TW`) is my own consistent terminology. When the sibling
  panel PRs land, any control name that disagrees is a one-line fix here.


artokun:
**Completeness of the extraction, checked in both directions** — independently of the extractor, so a bug in it cannot make its own check pass.

**Forward** (nothing live is missing): every string literal in `src/**/*.ts` (non-test) matching `(cli|say|console|banner).<path>` is present in `locales/en/main.json`. This catches wrapper sites (`t(`, `dtr(`) that a `tr(`-anchored grep cannot see — which is exactly how the `...tr(` spread site was lost in the first place.

```
every namespaced key literal in src/ is in the catalog
```

**Reverse** (nothing was invented): every catalog key appears as a literal in `src/`, except 16 — all `say.degraded.*`, all built at runtime by the `dtr` wrapper from the constant prefix `say.degraded.` plus a short argument (12 from literals in the ternary chain, 4 expanded from `OPENAI_KEY_PROVIDERS`). Zero orphans beyond that set, so no dead key is being demanded from eleven translators.

CI is green on all three OSes. **windows-latest passing is the load-bearing one**: GitHub's Windows runners default to `core.autocrlf=true`, so the byte-exact "English still matches a fresh build" test would have failed there without the `.gitattributes` LF pin. That is the check confirming the pin, rather than an assertion that it should work.

artokun:
## Review findings — all six fixed in 7e42e02

Each was **reproduced on the rig before being fixed and re-measured after**. Three of them were faults in guards I had written and believed were working, which is the part worth reading.

### HIGH — shipping catalogs made the whole suite locale-dependent

`processLocale()` reads `COMFYUI_MCP_LANG` then the POSIX chain. That was inert while no catalog existed; the moment `locales/` shipped it became a hidden input to hundreds of assertions.

```
$ COMFYUI_MCP_LANG=ja npx vitest run src/__tests__/transport/cli.test.ts src/__tests__/orchestrator/
Tests  4 failed | 2236 passed
```

Two of the four are **pre-existing tests my catalogs broke** (`ready-banner.test.ts`'s model label, `panel-console-http.test.ts`'s landing page); one is my own "unshipped language" test, which inherits the developer's locale because `negotiateLocale` falls back to `processLocale()`. CI was green only because runners are en/C — the worst shape of test dependency, invisible to everyone it passes for.

Fixed by pinning the language in `helpers/isolated-test-home.ts` the way `HOME` is already pinned, and by making `i18n.test.ts`'s precedence case own the entire chain rather than assume an empty environment (it was already latently broken for anyone with `COMFYUI_MCP_LANG` set).

```
$ COMFYUI_MCP_LANG=fa LANG=fa_IR.UTF-8 LC_ALL=fa_IR.UTF-8 npm test
Test Files  478 passed (478)
Tests  9129 passed | 3 skipped
locales OK
```

### MEDIUM — the blind-spot counter was blind to the exact fault it documents

It counted on `masked` — *the parser's own view* — so anything masking erased vanished from both sides. Reinstating the documented template-hole regression: **58 keys silently gone, counter silent**. The build failed only by luck, via an unrelated "unreadable fallback" from two constants that happen to be shared. It now counts on a comments-only copy, and re-running the same mutation gives:

```
src/boot.ts:0  29 tr/trFor call(s) in the source, 17 readable
src/orchestrator/index.ts:0  17 tr/trFor call(s) in the source, 3 readable
src/orchestrator/panel-console-http.ts:0  8 tr/trFor call(s) in the source, 2 readable
```

### MEDIUM — a wrapper in another module dropped every key behind it

Alias discovery is per file, and the `tr(` pre-filter skips the consumer outright, so the counter is satisfied at `0 === 0`. Measured with a probe `panelT(locale, key, fallback)` in a helper module: exit 0, no diagnostic, key absent — and `i18n-check` then certifies the catalog 100% translated while the string renders English forever.

Added a **second detector that knows nothing about call syntax**: every key-shaped literal in a namespace the catalog owns must be in the catalog. Worth noting that my first attempt at it sat behind the same `tr(` pre-filter and therefore reproduced the bug it was written to catch — it only works because it now runs for *every* source file, ahead of the filter. Verified in both directions: it fires on the probe, and it is silent across the entire real `src/` tree.

### LOW/MEDIUM — the passthrough test ran before `DYNAMIC_SITES`

`dtr(backend, …)` matches today because `backend` is a local const; the natural refactor of that 4000-line handler makes it a parameter, at which point `paramOwner` matches and the registry fan-out is skipped. Forcing that condition: four live `say.degraded.*` keys gone, exit 0, no stderr. The enumerated site (file + prefix + identifier + fallback shape) now answers first — re-measured under the same forced condition, **4/4 survive**.

### LOW — a template-literal fallback was captured verbatim

```
$ node scripts/i18n-build-en.mjs --json     # with a probe using `Hello ${NAME}, welcome back.`
captured as: "Hello ${NAME}, welcome back."
```

English renders correctly (the template evaluates), so this is invisible in the one language the author reads — while every translation shows a literal `${NAME}`, because `trFor` substitutes `{name}` and has never heard of `${…}`. Now refused loudly. This was the one place the script guessed where it refuses everywhere else.

### LOW — `i18n:check && vitest` reported zero test results

One new English key blocked the entire suite before vitest ran. Reordered to `vitest && i18n:check`: the gate still fails the run, but after the suite has spoken. Verified with a key deleted from `fr`:

```
Test Files  2 failed | 476 passed (478)
Tests  3 failed | 9126 passed | 3 skipped
```

Plus the nit: the `<html lang="xx"` assertion is now anchored on the attribute-ending character, and the `helpRow` comment claiming "every string on the screen is ASCII until a catalog lands" is no longer true and no longer says so.

**On the review's second question** — whether the three inverted tests lost coverage — I agree with the conclusion that nothing material was lost, and the byte-equality coverage is still carried by the `helpRow` cell test.

Issue #1371, opened 2026-08-10
download_model writes into stale COMFYUI_PATH when connected URL targets another local ComfyUI


artokun:
### What happened / root cause
download_model was called from a Codex app-server session connected to ComfyUI at http://127.0.0.1:8190. The tool resolved the local destination from a stale COMFYUI_PATH belonging to another installation and wrote the model under G:\Start_AI\AI_Test\ComfyUI\models instead of the active G:\Start_AI\AI_Test\ComfyUI_H3\models tree.

### Steps to reproduce
1. Start the panel/orchestrator from one local ComfyUI installation.
2. Connect the Codex app-server comfyui MCP to a second local ComfyUI URL (8190).
3. Call download_model action=download with target_subfolder=vae.
4. Observe that the file is written under the first installation while visibility is checked against the second.

### Exact error (scrubbed)
The tool reported: destination came from local configuration rather than the running server; visibility to the connected ComfyUI is unconfirmed.

### Fix
applied locally: no
upstream-only: yes
Precise change needed: when the Codex backend overrides COMFYUI_URL for a panel-selected target, it must also supply the matching ComfyUI root/base directory, or download_model must refuse local streaming when COMFYUI_PATH cannot be proven to correspond to the connected server. It must never silently write to a different local ComfyUI installation.

### Environment
Windows 11; ComfyUI v0.30.2; RTX 4080 SUPER 16 GB; comfyui-mcp 0.50.107; panel 0.11.43.

artokun:
Picking this up — draft in #1403. Your diagnosis is right, and the refusal you are asking for **already exists in this function** — it just does not guard the path you hit.

`resolveModelsDirWithBases` ranks its answer by source: `argv-flag` and `base-anchored` come from the running server, `configured-base` is plain local config. When the live root resolves but the configured base does not corroborate it, it refuses outright and says:

> Refusing to write to a guessed directory (that is how a model lands in a stale install and is reported as a success).

Which is this bug, already written down. But the branch below it takes `COMFYUI_PATH/models` with **no corroboration at all**, and that is the branch reached when the live root could not be derived — which is exactly your two-local-installs case, where `COMFYUI_PATH` points somewhere real, writable, and wrong.

You also saw the tool say *"destination came from local configuration rather than the running server; visibility to the connected ComfyUI is unconfirmed."* It knew, and wrote anyway. A warning next to a write is not a substitute for not writing.

[...]

Issue #1474, opened 2026-08-12
list_local_models remove ignores Comfy Desktop shared model roots


artokun:
### What happened / root cause
`list_local_models` action `list` showed models served from Comfy Desktop's shared model root, and action `list_paths` correctly displayed `M:\\Comfy-Desktop\\ComfyUI-Shared\\models` from the live `--extra-model-paths-config`. But action `remove` searched only the primary install root and returned MODEL_ERROR for both files.

Root cause: `resolveExistingModelFile()` calls `getExtraModelRoots()`, which intentionally refuses the display-only #764 fallback when `/system_stats` argv omits a provable main.py. On this normal Comfy Desktop launch, `list_paths` can read the exact absolute shared_model_paths.yaml but removal gets zero extra roots. This contradicts the remove tool documentation that it resolves across all configured roots.

### Steps to reproduce
1. Run Comfy Desktop with shared models under `M:\\Comfy-Desktop\\ComfyUI-Shared\\models`.
2. `list_local_models({action:'list', model_type:'diffusion_models'})` lists `z_image_turbo_bf16.safetensors`.
3. `list_local_models({action:'list_paths'})` shows the Desktop shared root.
4. `list_local_models({action:'remove', path:'diffusion_models/z_image_turbo_bf16.safetensors'})` fails.

### Exact error (scrubbed)
`Model file not found: diffusion_models/z_image_turbo_bf16.safetensors. Searched 1 root(s): M:\\Comfy-Desktop\\ComfyUI-Installs\\comfyui\\ComfyUI\\models`

The same failure occurred for `text_encoders/qwen_3_4b.safetensors`.

### Fix
Applied locally: no. Upstream-only: yes, because this is a filesystem authorization/security boundary and relaxing the proof rule locally would be risky.

Precise change needed: make action `remove` accept an explicit confirmed `config_path`/Desktop target, or add a trustworthy Comfy Desktop proof path for the live absolute `--extra-model-paths-config`, so `resolveExistingModelFile()` can enumerate the same shared roots reported by `list_paths` without using an unproven stale config. Keep containment checks intact.

Temporary workaround verified: resolve the shared root from `list_paths`, containment-check exact absolute files under that root, delete them with local filesystem access, then re-run `list_local_models` to confirm they disappeared.

### Environment
[...]

PR #1487, opened 2026-08-12
fix(1371): refuse a download whose destination the server PROVABLY does not read

artokun:
Claims #1371 — a model was written into a **different local ComfyUI installation** than the one serving the session.

The reporter was connected to ComfyUI on `:8190` (`ComfyUI_H3`), and `download_model` streamed into `G:\Start_AI\AI_Test\ComfyUI\models` — a stale `COMFYUI_PATH` from another install — while visibility was checked against the connected server.

**Not already fixed, and I checked before assuming.** `model-resolver.ts` has carried a divergence refusal since **v0.49.4** (#794), and the reporter was on **0.50.107** — newer. So the guard was present and did not fire.

Why: it proves divergence from CONTENT — files present on disk that the running server does not list. The code deliberately refuses the inverse inference ("the server lists models this tree lacks") because absence of evidence produced false refusals. With a sparse or empty destination category (`vae`, in this report) there is nothing to compare, so nothing can be proven and the write proceeds.

**Direction, to be confirmed:** the server can advertise its own root via `/system_stats` argv, and the resolver already reasons about that elsewhere. A destination root that DIFFERS from a root the server itself reports is positive proof of divergence, not absence of it — so it can refuse without reintroducing the false-refusal class the existing note warns about.

Investigating. The thing this must not become is a blanket "refuse when unproven", which that same note documents as having been tried and reverted.

### DIFF
--- a/src/__tests__/services/download-root-correspondence.test.ts
+++ b/src/__tests__/services/download-root-correspondence.test.ts
@@ -1,122 +1,92 @@
 // #1371 — a model was written into a DIFFERENT local ComfyUI installation than the one
-// serving the session.
+// serving the session: connected to :8190 (…\ComfyUI_H3), streamed into …\ComfyUI\models
+// from a stale COMFYUI_PATH, while visibility was checked against the connected server.
 //
-// The reporter was connected to ComfyUI on :8190 (…\ComfyUI_H3), and download_model
-// streamed into …\ComfyUI\models — a stale COMFYUI_PATH from another install — while
-// visibility was checked against the connected server.
+// The existing divergence guard shipped in v0.49.4 and the reporter was on 0.50.107, so
+// it was present and did not fire: it proves divergence from CONTENT (files on disk the
+// server does not list), and an empty destination category proves nothing.
 //
-// The existing divergence guard has shipped since v0.49.4 and the reporter was on
-// 0.50.107, so it was present and did not fire: it proves divergence from CONTENT
-// (files on disk the server does not list), and an empty destination category proves
-// nothing. This adds POSITIVE evidence instead — the server naming its own root.
+// A FIRST attempt refused when the destination fell outside the install root the server
+// names for itself. Review knocked that down and was right — extra_model_paths.yaml
+// registers roots elsewhere by design, <install>/models is routinely a junction, and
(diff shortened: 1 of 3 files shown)


artokun:
Gate returned **NEEDS-REVISION** with three P1s, and the first two refute the premise. Returning this to draft rather than patching it.

**The claim does not hold.** I built this on "the destination is outside the root the server names for itself, therefore the server cannot read it." That inference is wrong:

- `extra_model_paths.yaml` exists precisely to register model roots outside the install
- `<install>\models` is routinely a junction or symlink to another drive
- a container-mounted server path and its valid host destination describe the same directory through unrelated strings

**I had the counterexample in my hands today.** #1474 — which I fixed and released a few hours ago — is a Comfy Desktop install serving models from `M:\Comfy-Desktop\ComfyUI-Shared\models`, outside the install root. This guard would have refused that user's download outright. I read the note in `model-resolver.ts` warning that every tightening of this inference produced another false refusal, wrote in my own PR description that the thing to avoid was exactly that, and then did it anyway with a different justification.

The third finding is real and cheap (an uncached, undeadlined `getSystemStats()` on every resolution can hang a download instead of taking the "no root, no refusal" path) — but fixing it would only make a wrong guard faster.

**What would actually be proof.** Not the install root — the set of model roots the server ENUMERATES. `list_local_models (action:"list_paths")` already reports those, and #1474 established the machinery and the vocabulary for "roots the running server can be shown to read" versus roots merely present in local config. A destination in none of the server's enumerated roots is a much stronger signal than one outside its install directory, and it does not misclassify the shared-root case.

That is a different change and it deserves its own pass with the measurement in hand, not a fourth patch on a premise the review has already knocked down.

**What stands regardless**, and is worth keeping from this branch:
- the diagnosis on #1371 is confirmed — the existing divergence guard shipped in **v0.49.4** while the reporter was on **0.50.107**, so it was present and did not fire, because it proves divergence from CONTENT and an empty destination category proves nothing
- `isWithinRoot` is prefix-safe for the `ComfyUI` vs `ComfyUI_H3` shape, which a naive `startsWith` gets wrong in both directions
- the repo's own guards caught two mistakes on the way: reaching the live process table through the shared root resolver, and then tripping that same gate from my test's prose

Issue #1331, opened 2026-08-10
panel_* binding goes stale after ComfyUI reconnect (mutations rejected with "no trusted identity"), and panel_request_secret hard-times-out at 300s while awaiting human input


artokun:
Two separate defects hit during one long panel session (image-style workflow build + iterate). Both were recovered from, but each cost an extra round-trip / a lost interaction.

---

### What happened / root cause
The session survived several ComfyUI drop/reconnect cycles (the panel injected the "✅ Your connection dropped mid-task … session resumed" notice each time). After one of those, the canvas WAS on the correct workflow and the panel said so explicitly, but every MUTATING `panel_*` call was refused, while reads were still allowed. The workflow had been created in-session via `panel_new_workflow` → `panel_paste_nodes` → `panel_rename_workflow` → `panel_save_workflow`, so it had a real saved path — yet the panel reported it had "no trusted identity" to fence commands against. It looks like the reconnect drops/invalidates the workflow-identity token for a workflow that was created (rather than opened) in this session, without re-deriving it from the now-saved path.

### Steps to reproduce
1. `panel_new_workflow`, paste/build nodes, `panel_rename_workflow`, `panel_save_workflow`.
2. Let ComfyUI restart or the socket drop so the panel emits the "connection dropped … resumed" notice.
3. Call any mutating panel tool, e.g. `panel_set_widget`.

### Exact error (scrubbed)
```
Error: graph_set_widget could not be dispatched to this session's panel tab — nothing was applied.
The tab may be disconnected, still reconnecting after a restart/reload, or the session's binding is
stale (e.g. another workflow tab is now active). Retry in a moment, or rebind with
panel_set_workflow_target({mode:"current"}) to follow the tab that's live now.
("graph_set_widget" cannot be safely targeted to the active workflow: this workflow has no trusted
identity for the panel to fence the command against. Reads and view-only commands still work
(graph_outline, graph_query, graph_get_state, graph_find_nodes, graph_list_subgraphs,
graph_screenshot, graph_canvas).)
```

Recovery required `panel_open_workflow` on the workflow that was *already active*. The first `panel_open_workflow` then returned a **false-alarm mismatch error**:

```
Error: workflow_open RAN and the canvas IS bound to niji_holo_impasto — that much was proven — but
the graph on it does not match the state that was loaded, and the panel cannot tell whether the
ComfyUI frontend merely normalized it (node sizes, widget values) or the load only partly applied.
[...]

Issue #812, opened 2026-08-04
graph_set_widget/graph_add_node blocked with no self-service update path when panel < 0.11.35


artokun:
### What happened / root cause
While helping a user edit their live workflow (change a ResizeImageMaskNode widget + add a temp PreviewImage node), every write call (`panel_set_widget`, `panel_add_node`) failed with:

> "graph_set_widget" cannot be safely targeted to the active workflow: panel tab wf:workflows/<name>.json does not recheck workflow targeting at the graph write boundary after asynchronous work (detected panel 0.11.32; this MCP requires panel 0.11.35+). Run install_panel(action:'update'), restart ComfyUI, then hard-refresh...

The error message instructs the agent to run `install_panel(action:'update')`, but no `install_panel` tool is exposed to the agent session at all (confirmed via ToolSearch with several query variants — no match). I tried the obvious fallback, `panel_update_node` on the `comfyui-mcp-panel` pack, but that is explicitly refused too:

> panel_update_node cannot manage the comfyui-mcp sidebar panel pack ("comfyui-mcp-panel"). ... Use install_panel instead: install_panel(action='sync') ...

So the agent is told by two different error messages to call a tool (`install_panel`) that does not exist in its toolset, with no alternative self-service path. Net effect: once a user's panel version drifts below 0.11.35, the agent cannot perform ANY live graph edit, and cannot self-heal by updating the panel either. The only way out is fully manual (user runs ComfyUI-Manager UI or git pulls the pack themselves).

### Steps to reproduce
1. Run comfyui-mcp with panel version 0.11.32 (older/legacy install).
2. Ask the agent to edit any widget or add any node to the currently open workflow via the panel_* graph tools.
3. Observe the version-gate error instructing to call `install_panel(action:'update'|'sync')`.
4. Try to find/call `install_panel` via ToolSearch — no matching tool is returned.
5. Try `panel_update_node(id:'comfyui-mcp-panel')` as a fallback — explicitly refused with a message again pointing at `install_panel`.

### Exact error (scrubbed)
```
[...]

Issue #708, opened 2026-08-02
Deadlock: all graph edits refused because panel 0.11.28 never sends `enforces_workflow_stamp`, while install_panel reports "up-to-date"


artokun:
### What happened / root cause

Every graph MUTATION through the panel bridge is permanently refused on an install that the orchestrator's own sync advisor calls current. The user has **no action available** to resolve it — the two surfaces directly contradict each other.

The refusal (hit on `panel_set_widget`, and on `panel_find_nodes` — see the companion issue):

```
"graph_set_widget" cannot be safely targeted to the active workflow:
panel tab wf:workf does not enforce per-command workflow targeting
(update the ComfyUI-MCP panel to edit the graph).
Read-only graph commands (graph_outline, graph_query, graph_get_state) still work.
```

But `install_panel(action='status')` on the same machine, same moment:

```json
{
  "installedVersion": "0.11.28",
  "isDevSymlink": false,
  "shadows": [],
  "scanReliable": true,
  "pin": { "pinned": false, "source": "none" },
  "sync": {
    "requiredPanelVersion": "0.11.28",
    "installedVersion": "0.11.28",
    "orchestratorVersion": "0.49.0",
    "decision": "up-to-date",
    "behind": false,
    "summary": "Panel 0.11.28 already meets what comfyui-mcp 0.49.0 needs (0.11.28+). Nothing to do."
  }
}
```

No pin, no shadow copy, no dev symlink — so none of the existing `blocked` escape hatches apply.

**Root cause.** `dist/services/ui-bridge.js:1637-1647` gates every mutating graph command on the runtime handshake flag `conn.enforcesWorkflowStamp`, set at hello from `msg.enforces_workflow_stamp === true` (`ui-bridge.js:930`, strict `=== true`, deliberately re-read per hello).

The installed panel pack **never sends that key**. Verified binary-safe across the whole pack:

```
$ grep -ar "enforces_workflow_stamp" <comfyui>/custom_nodes/comfyui-agent-panel
(no matches)

$ grep -aon 'PANEL_VERSION = ' <pack>/web/js/comfyui-mcp-panel.js
399:PANEL_VERSION = "0.11.28"
```

(Note `web/js/comfyui-mcp-panel.js` contains 2 NUL bytes, so plain ripgrep treats it as binary and silently stops scanning it — `grep -a` is required to check this reliably. Anything auditing the pack for capability strings will get a false negative otherwise.)

[...]

PR #1492, opened 2026-08-13
fix(1480): the mismatch guard names a selector an unsaved tab actually has

### LINKED ISSUE #1480: root-workflow-uuid-mismatch on an UNSAVED (tmp) workflow is an unrecoverable deadlock — every documented remedy dead-ends

The `root-workflow-uuid-mismatch` guard fired on a workflow that has **never been saved** (a tmp tab created by loading a workflow embedded in an image). In that state the guard has **no exit**: every remedy it recommends is itself blocked, so the agent cannot edit the user's live canvas at all and can only ask the user to hand-save or hand-reload in the browser.

The deadlock cycle:

1. `panel_set_widget` → refused: `[root-workflow-uuid-mismatch]`. Remedy suggested: re-open the tab with `panel_open_workflow`, else `panel_reload scope:frontend`.
2. `panel_open_workflow(path:"ComfyUI_04661_")` → `no workflow matching "ComfyUI_04661_" — it isn't among the saved/open workflows even after a refresh.` The tab is unsaved, so `panel_list_workflows` reports `path: null`, `filename: null`, `persisted: false`, and only a tmp key (`tmp:<uuid>`). There is nothing `panel_open_workflow` can accept.
3. `panel_reload scope:frontend` → refused: *"this workflow has unsaved changes … the browser blocks a scripted navigation while unsaved work is open"*. Correct and safe, but it removes remedy #2.
4. `panel_save_workflow(name:"…")` — which would give the tab a path and re-enable `panel_open_workflow` — → refused by the **same** mismatch guard (issue #708): *"the live canvas positively belongs to a DIFFERENT workflow … saving now could write the other workflow's graph over this one."*

So: can't edit (guard), can't open (no path), can't reload (unsaved), can't save (guard). Dead end. The only way out is a human at the keyboard.

**Secondary bug — read-path inconsistency.** On the *identical* canvas state:
- `panel_graph_outline` → refused with `root-workflow-uuid-mismatch`
- `panel_query_graph` (`fields:'ids'`, then `fields:'detail'` on all 9 ids) → **succeeded**, returning the full graph with widgets, wiring and modes

Both are read-only whole-graph reads of the same canvas, so one refusing while the other works is surprising. It also made the failure look intermittent — after `panel_graph_outline` failed, `panel_query_graph` succeeding suggested the binding had recovered, and only the next write revealed it hadn't.

### Steps to reproduce

1. Have a single ComfyUI tab whose workflow was never saved (e.g. drag a PNG with an embedded workflow onto the canvas — title `ComfyUI_04661_`, `path: null`, `persisted: false`, `modified: true`).
2. Get the canvas into the mismatched state (here it arrived that way at the start of an agent turn — the panel reported a workflow switch; no load/tab-switch/reconnect was witnessed by the guard itself).
3. Agent calls `panel_graph_outline` → mismatch error.
4. `panel_query_graph` → succeeds, full graph returned.
5. `panel_set_widget` / `panel_save_workflow` → mismatch error.
6. `panel_open_workflow(<title>)` → not found. `panel_reload scope:frontend` → refused, unsaved changes.

Graph was a plain 9-node SDXL-family txt2img (CheckpointLoaderSimple → 2× LoraLoader → 2× CLIPTextEncode → KSampler → VAEDecode → SaveImage + EmptyLatentImage). Nothing exotic; the graph shape is not implicated.

### Exact error (scrubbed)

```
[root-workflow-uuid-mismatch] The live canvas carries a different workflow's identity tag
than the active workflow, so this command was NOT applied. ... Re-open the active workflow
tab (panel_open_workflow) to rebind the graph in place; if that does not clear it, reload
the panel (panel_reload scope:frontend), then retry.
```
```
no workflow matching "ComfyUI_04661_" — it isn't among the saved/open workflows even after
a refresh. For a file outside the workflows folder, load it with panel_load_workflow path:<file>.
```
```
Did NOT reload the panel: this workflow has unsaved changes — ComfyUI_04661_. The browser
blocks a scripted navigation while unsaved work is open ... Nothing was changed.
```
```
refusing to save "ComfyUI_04661_": the live canvas positively belongs to a DIFFERENT
workflow ... saving now could write the other workflow's graph over this one (issue #708).
Nothing was written. Open the workflow you mean (panel_open_workflow) and retry.
```

`panel_list_workflows` at the time:
```
active: { path: null, filename: null, title: "ComfyUI_04661_",
          key: "tmp:ff43c1ea-…", workflow_uuid: "c8244018-…" }
open:   [ { …, active: true, modified: true, persisted: false } ]   // only one tab
```

### Fix

applied locally: **no** — upstream-only: **yes**

The guard lives in the long-lived orchestrator process that serves the `panel_*` tools. Per the `panel_reload` contract that process only picks up code on a full process restart (not `panel_reload`, not Disconnect→Connect), so a local patch could not have unblocked this session — patching would have added risk without helping the user. Precise changes needed:

1. **Give the guard an exit for never-saved tabs.** The recommended remedy `panel_open_workflow` is unreachable when `path == null`. Either:
   - let `panel_open_workflow` accept the tmp `key` / `routing_key` (`tmp:<uuid>`) or the tab `title` as a rebind target, not just a saved path/filename; **or**
   - add an explicit rebind operation (e.g. `panel_open_workflow({ key })` or a `panel_rebind_canvas` tool) that re-asserts the active tab's identity tag onto the canvas without needing a file on disk.

2. **Let `panel_save_workflow` offer a safe escape hatch under mismatch.** The #708 refusal is right by default, but a "save the canvas as a NEW file" mode is safe by construction — a new filename cannot overwrite either workflow. Something like `panel_save_workflow({ name, force_new: true })` that refuses to touch any existing file would break the cycle: the tab gains a path, `panel_open_workflow` becomes reachable, and the user loses nothing even if the canvas really did belong to the other workflow.

3. **Make the read paths agree.** Either `panel_query_graph` should apply the same mismatch check as `panel_graph_outline`, or (preferable, since reads can't corrupt anything) `panel_graph_outline` should downgrade to a **warning banner** on the outline it returns rather than refusing outright. Reads are safe under mismatch — it's writes that can clobber. Today's split behaviour is the worst of both: it blocks a harmless read and disguises a persistent fault as an intermittent one.

4. **Improve the error text** once the above exists: the message should name the remedy that will actually work for the tab's current state (unsaved vs saved), instead of pointing at `panel_open_workflow` unconditionally.

### Environment

- ComfyUI: **0.30.0** | **frontend 1.47.12**
- Python 3.12.9 / PyTorch 2.10.0+cu130 / CUDA 13.0
- GPU: NVIDIA GeForce RTX 4070 SUPER, 12 GB VRAM (7.7 GB free at the time), 32 GB RAM
- comfyui-mcp: **0.51.16**
- comfyui-mcp-panel: **0.14.2**
- Backend: Claude
- Queue was idle (0 running, 0 pending)

### User impact

The user asked the agent to continue a series of renders on their open canvas. The agent could read the graph but could not change a single widget or queue a variant, and had to hand the problem back with "please press Ctrl+S yourself." For a never-saved tab — which is exactly what you get from dropping a PNG on the canvas, a very common entry point — this guard currently converts a recoverable binding glitch into a hard stop.

artokun:
Fixes #1480.

`[root-workflow-uuid-mismatch]` fires on a tab that has **never been saved** (a tmp tab from a workflow embedded in an image), and in that state every remedy it recommends is itself blocked:

1. `panel_set_widget` → refused by the guard; it suggests re-opening the tab, else `panel_reload scope:frontend`
2. `panel_open_workflow` → the reporter passed the tab's **title** and got `no workflow matching`
3. `panel_reload scope:frontend` → refused, correctly, because unsaved work is open
4. `panel_save_workflow` → refused by the **same** guard (#708)

Can't edit, can't open, can't reload, can't save.

## What I expected to build, and why I didn't

The plan in this PR's first draft was a `force_new` escape hatch on `panel_save_workflow`: a save to a genuinely new path cannot overwrite either workflow, so it is safe by construction, and it would give the tab a path.

Reading the panel first killed that as the *first* thing to try, because **step 2 was never actually blocked** — it was mis-addressed:

- `workflowRecordMatchesSelector` accepts a record's **per-instance routing id**, and for an unsaved tab that is explicitly "the ONLY authority" (#186) — the shared `Unsaved Workflow` title/key must never select an arbitrary unsaved tab.
- `workflow_list` already publishes it as `routing_key` on **every** record.
- The orchestrator already handles `panel_open_workflow(tmp:<uuid>)`, including the fence refresh that a saved path gets (#812).

So the exit existed, needed no file on disk, and nothing in the failure path named it. The guard's remedy says `<path>` unconditionally; for this tab `<path>` does not exist and the title is not a selector, which is exactly the turn the reporter burned.

I confirmed the record shape on a live rig rather than from the issue text — an unsaved tab really does report `path: null, filename: null, key: "tmp:…", routing_key: "tmp:…"`, and that measured shape is the test fixture.

## The change

This refusal now does what the two refusals beside it already do (#1330, #1331): one read-only `workflow_list` — exempt from the fence being reported on — establishes the single fact that decides which remedy is followable, and the message names **that** one:

- **unsaved** → `panel_open_workflow({path: "tmp:<uuid>"})`, plus an explicit warn-off for the two exits that refuse from this state
- **saved** → the panel's own `<path>` advice, named concretely so it isn't guessed at
- **unreadable** → says so, and promises nothing

Applied to reads as well as mutations: `panel_graph_outline` refusing was half the reported dead end, and the verdict is raised before every executor, so its "NOT applied" claim holds either way.

Nothing is auto-applied. The guard is not weakened — #708's concern is untouched, and the saved-tab direction is pinned by a test so an over-broad fix can't tell everyone to pass a routing key.

## Verification

- 4 new tests on the real tool path, using the panel's **verbatim** refusal text and the rig-measured `workflow_list` shapes
- each asserts the `workflow_list` read actually ran, so a branch that never executes can't pass
- **mutation-checked**: stubbing the classifier to `false` fails all 4
- full suite: 481 files / 9168 tests green; `lint`, `check:vocabulary`, `check:unknown-collapse`, `vocab:export --check` all pass

## Deliberately not in this PR

Both are panel-side and larger; they remain open on the issue:

- the `force_new` save escape hatch (issue item 2)
- making the two read paths agree (item 3) — `panel_query_graph` succeeding where `panel_graph_outline` refused is the `staleTagReadBypass` content-equality proof (#995) landing differently at two moments, not a structural read/write split; it deserves its own change

### DIFF
--- a/src/__tests__/orchestrator/tmp-workflow-mismatch-remedy.test.ts
+++ b/src/__tests__/orchestrator/tmp-workflow-mismatch-remedy.test.ts
@@ -161,6 +161,36 @@ describe("root-workflow-uuid-mismatch names a remedy the tab can accept (#1480)"
     expect(text).not.toContain("tmp:");
   });
 
+  it("a message that merely QUOTES the token is not this refusal (codex)", async () => {
+    // The classifier is anchored, so it keys on the panel's real shape — its executor
+    // error text becomes the Error message verbatim, no prefix — instead of on any
+    // message that happens to contain the token. Unanchored, a wrapped or re-surfaced
+    // refusal would burn a workflow_list round trip and be told to re-open a workflow.
+    const ctx = makePanelToolCtx(
+      {
+        ...bridge(UNSAVED_LIST),
+        send: async (cmd: Record<string, unknown>) => {
+          if (cmd.cmd === "workflow_list") {
+            listCalls += 1;
+            return UNSAVED_LIST;
+          }
+          throw new Error(`Validation failed: "[root-workflow-uuid-mismatch]" is reserved`);
+        },
+      } as unknown as PanelToolCtx["bridge"],
+      TAB,
+      new WorkflowTargetStore(),
+    );
+    const def = buildPanelToolDefs().find((d) => d.name === "panel_connect");
+    if (!def) throw new Error("panel_connect is not registered");
+    const res: ToolResult = await def.handler({ from_node_id: 1, to_node_id: 2 } as never, ctx);
+    const text = res.content.map((c) => (c as { text?: string }).text ?? "").join(" ");
+
+    // The probe never ran, which is the whole point: no round trip, no remedy invented.
+    expect(listCalls).toBe(0);
+    expect(text).not.toMatch(/NEVER BEEN SAVED/);
+    expect(text).not.toContain(ROUTING_KEY);
+  });
+
   it("UNREADABLE: promises nothing, because nothing was established", async () => {
     // The failure direction that matters. Naming a selector we could not establish would
     // send the agent to "no workflow matching" — the exact turn #1480 wasted.
--- a/src/__tests__/orchestrator/tmp-workflow-mismatch-remedy.test.ts
+++ b/src/__tests__/orchestrator/tmp-workflow-mismatch-remedy.test.ts
@@ -1,0 +1,175 @@
+// #1480 — the root-workflow-uuid-mismatch guard had no followable exit on a NEVER-SAVED
+// tab, so a session could read the canvas and change nothing.
+//
+// The panel's remedy for this verdict is `panel_open_workflow(<path>)`. An unsaved tab has
+// no path — `workflow_list` reports `path: null, filename: null` for it by contract (#186,
+// because native ComfyUI reuses the "Unsaved Workflow" title across unsaved tabs) — so the
+// reporter passed the TITLE, got "no workflow matching", and found every other documented
+// exit closed as well: reload refuses while unsaved, save refuses under this same guard
+// (#708). The exit existed the whole time and nothing named it: the per-instance ROUTING
+// KEY is a valid `workflow_open` selector, and `workflow_list` already publishes it.
+//
+// The refusal text below is the panel's VERBATIM message (graph-binding.js
+// `graphBindingRefusalMessage`), and the workflow_list replies are the shapes a live rig
+// actually returns — an unsaved tab measured on the rig reports exactly the
+// `path: null / filename: null / key: tmp:<uuid> / routing_key: tmp:<uuid>` record used
+// here. A fixture invented from the issue text would not have proven the classifier keys
+// on a field that exists.
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("../../comfyui/client.js", () => ({
+  getObjectInfo: vi.fn(),
+  backfillObjectInfo: vi.fn(),
+  resetClient: vi.fn(),
+  resetObjectInfoCache: vi.fn(),
+}));
+
+const { buildPanelToolDefs, makePanelToolCtx } = await import("../../orchestrator/panel-tools.js");
+import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+import type { PanelToolCtx, ToolResult } from "../../orchestrator/panel-tools.js";
+
+const TAB = "11111111-2222-3333-4444-555555555555";
+const ROUTING_KEY = "tmp:ffadfd06-41bb-4448-9ad2-fe307ff3356f";
+const SAVED_PATH = "workflows/sdxl-txt2img.json";
+
+/** The panel's own text for this verdict, copied from graphBindingRefusalMessage. */
+const PANEL_REFUSAL =
+  `[root-workflow-uuid-mismatch] The live canvas carries a different workflow's identity tag ` +
+  `than the active workflow, so this command was NOT applied. What the panel observed is the ` +
+  `disagreement itself, not what produced it — a load, tab switch or reconnect leaving a stale ` +
+  `tag behind is the usual explanation, but none of them was witnessed here. Re-open the active ` +
+  `workflow tab (panel_open_workflow) to rebind the graph in place; if that does not clear it, ` +
+  `reload the panel (panel_reload scope:frontend), then retry.`;
+
+/** As measured on a live rig: an unsaved tab publishes its routing key and nothing else. */
+const UNSAVED_LIST = {
+  active: {
+    path: null,
+    filename: null,
+    title: "Unsaved Workflow",
+    key: ROUTING_KEY,
+    routing_key: ROUTING_KEY,
+    workflow_uuid: "defb7763-6e20-4f5a-b9cd-2aeabfea6f08",
+  },
+  open: [
+    {
+      path: null,
+      filename: null,
+      title: "Unsaved Workflow",
+      key: ROUTING_KEY,
+      routing_key: ROUTING_KEY,
+      active: true,
+      modified: true,
+      persisted: false,
+    },
+  ],
+  active_confirmed: true,
+};
+
+const SAVED_LIST = {
+  active: {
+    path: SAVED_PATH,
+    filename: "sdxl-txt2img.json",
+    title: "sdxl-txt2img.json",
+    key: SAVED_PATH,
+    routing_key: `wf:${SAVED_PATH}`,
+    workflow_uuid: "defb7763-6e20-4f5a-b9cd-2aeabfea6f08",
+  },
+  open: [{ path: SAVED_PATH, active: true, modified: false, persisted: true }],
+  active_confirmed: true,
+};
+
+let listCalls = 0;
+
+/** A tab whose every graph command is refused by the panel's binding guard. */
+function bridge(list: unknown | "unreadable") {
+  return {
+    send: async (cmd: Record<string, unknown>) => {
+      if (cmd.cmd === "workflow_list") {
+        listCalls += 1;
+        if (list === "unreadable") throw new Error("the panel did not answer");
+        return list;
+      }
+      throw new Error(PANEL_REFUSAL);
+    },
+    push: () => 1,
+    canReach: (id: string) => id === TAB,
+    isHeadless: () => false,
+    tabs: () => [{ tab_id: TAB, title: "wf", connected_at: 0 }],
+    resolveActiveTabId: () => TAB,
+    refreshWorkflowUuid: () => true,
+    workflowUuidFor: () => ({ known: true, uuid: "defb7763-6e20-4f5a-b9cd-2aeabfea6f08" }),
+    tabCanMutateGraph: () => true,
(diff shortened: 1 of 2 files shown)


artokun:
Mechanism confirmed, and the exit is narrower and safer than I expected.

**The site.** `web/js/lib/workflow-save.js:459` — `assertCanvasNotForeign()` runs **unconditionally**, before the explicit-name branch:

```js
const assertCanvasNotForeign = () => {
  if (normalizeCanvasBinding(canvasBinding, wf) === "foreign") {
    throw new Error(`refusing to save "…": … saving now could write the other workflow's
      graph over this one (issue #708). Nothing was written. Open the workflow you mean
      (panel_open_workflow) and retry.`);
  }
};
assertCanvasNotForeign();
```

Note the remedy it gives — *"open the workflow you mean (panel_open_workflow)"* — is precisely the step that is impossible here, because an unsaved tab has no path to open. The guard closes the loop on itself.

**Why #708's reasoning does not reach this case.** Its danger is stated in its own text: *"saving now could write the other workflow's graph OVER THIS ONE"*. That is an OVERWRITE hazard, and it presupposes a persisted file to overwrite. On a tmp tab there is none — `path: null`, `persisted: false`. A save to a genuinely new path cannot destroy anything: worst case it writes whatever the canvas currently holds under a new name, which is recoverable, and it gives the tab a path — which re-enables `panel_open_workflow` and unblocks every other remedy in the cycle.

So the exit is: **an explicit name resolving to a path that does not already exist**. Both halves are load-bearing — an explicit name (never an implicit save-in-place, which #226 already guards) and a non-existent target (or the overwrite hazard returns in full).

**And it must say what it saved.** If the canvas is foreign, the new file may contain the other workflow's graph. A save that quietly produces a file of the wrong graph under a name the user chose is a worse outcome than the deadlock. The reply has to state that the contents are the live canvas, which the panel believes belongs to a different workflow, and that it should be verified before being relied on.

**Not shipping it in this pass.** This is a wrong-graph guard, and #886 earlier today took three review rounds precisely because every relaxation of that family reads as reasonable and is not. The change deserves a full pass — the existence check has to be race-free against the save itself, and I want the refusal-vs-allow boundary mutation-tested before it goes near a release.

Everything needed to resume is here: the site, why the overwrite reasoning does not extend to a tmp tab, the two conditions on the exit, and the disclosure it owes.

artokun:
Design is settled and the code is not written. Recording the exact shape so the next pass starts from a plan rather than a re-read.

**The pieces are all already in `saveActiveWorkflow(svc, name, { existsOnDisk, canvasBinding, … })`:**

- `existsOnDisk(rawPath) => true | false | null` — a tri-state disk oracle that already exists and whose documented discipline is exactly right here: *"Anything short of a proven absence refuses."*
- `needsGrounding(wf)` — `wf.isPersisted === false || wf.isTemporary === true`, i.e. the tmp tab in this report
- `assertCanvasNotForeign()` at `workflow-save.js:459`, currently unconditional and synchronous

**The exit, stated as a predicate.** Skip the foreign refusal only when ALL hold:

1. the name is EXPLICIT (never an implicit save-in-place — #226 already guards that path)
2. `needsGrounding(wf)` — the tab has never been persisted, so there is no source file to destroy
3. `existsOnDisk(target) === false` — PROVEN absent. `true` or `null` refuses, same as `groundingIsSafe`

Miss any one and the existing refusal stands unchanged.

**What it must disclose.** If the canvas is foreign, the new file may hold the other workflow's graph. The reply has to say the contents are the live canvas, that the panel believes that canvas belongs to a different workflow, and that it should be checked before being relied on. A file of the wrong graph under a name the user chose, saved silently, is worse than the deadlock.

**Why it is not written yet.** Two things make this more than an afternoon: `assertCanvasNotForeign` becomes async (the oracle is a promise) and must be awaited at its call site, and the target path for an explicit name is derived further down the function than the guard runs — so the guard either moves or the derivation lifts. Both are mechanical but both touch a wrong-graph fence, and #886 needed three review rounds today precisely because relaxations in that family read as obviously safe and are not.

The allow/refuse boundary needs mutation testing before this goes near a release — specifically that `null` and `true` from the oracle both still refuse, and that an implicit save still refuses.

Worktree `wt-1480p` on `fix/1480-tmp-save-exit` is clean at origin/main; nothing is half-written.

artokun:
## Live-rig verification (not just tests)

Reproduced and fixed end-to-end on a real ComfyUI (0.32.0), against an isolated orchestrator on its own port so nothing touched the live session. The tab was the issue's shape: **never saved** (`isPersisted: false`), **modified**, 7 nodes, with a foreign identity tag planted on the mounted root.

**BEFORE** (orchestrator running the pre-fix build) — `panel_graph_outline`:

> `[root-workflow-uuid-mismatch] The live canvas carries a different workflow's identity tag … Re-open the active workflow tab (panel_open_workflow) to rebind the graph in place; …`

and nothing else. That is the whole dead end: the only remedy names `<path>`, which this tab does not have.

**The reporter's move**, confirmed to fail the same way here — `panel_open_workflow({path: "Untitled 2026-08-05 15-03-15"})` (the title):

> `no workflow matching "Untitled 2026-08-05 15-03-15". The workflow list WAS re-read from the server first, and it still does not contain it.`

**AFTER** (same call, same canvas state, fixed build) — the refusal now carries:

> CHECKED, so this is not a guess: the active tab has NEVER BEEN SAVED (the workflow list reports path: null, filename: null, title: "Untitled 2026-08-05 15-03-15"). The remedy above names a `<path>` that DOES NOT EXIST for this tab … USE ITS ROUTING KEY …: `panel_open_workflow({path: "tmp:ac88a33a-…"})`

**And the remedy it names actually works** — this is the part a unit test cannot establish, so it was measured:

| step | result |
|---|---|
| `panel_open_workflow({path: "tmp:ac88a33a-…"})` | `opened`, `active_matches_target: true`, and `workflow_uuid` back to the tab's real `f05dd9d4-…` — the repaint replaced the foreign tag |
| `panel_graph_outline` (was refused) | succeeds, 7 nodes |
| `panel_set_widget` KSampler `steps` 28 → 21 | applied |

Canvas restored to `steps: 28` afterwards.

So the fix does not merely describe an exit — the exit it names is the one that clears the guard, on a never-saved tab, with no file on disk and no human at the keyboard.

Round-2 codex review on the anchor fix: no defects found.

Issue #742, opened 2026-08-02
panel_restart_comfyui reports "Cancelled — ComfyUI was not restarted" but ComfyUI actually shuts down and never comes back (Pinokio install)


artokun:
### What happened / root cause

Called `panel_restart_comfyui` (no args) after installing three custom node packs. The tool returned:

```
Cancelled — ComfyUI was not restarted.
```

But ComfyUI **had already been stopped**. Within seconds every subsequent panel call failed with `no connected tab ... Connected: none`, and a direct probe confirmed the server was gone:

```
$ curl -s -m 8 http://localhost:8188/system_stats  -> HTTP 000 (ComfyUI DOWN)
$ ps aux | grep main.py                            -> no ComfyUI process
$ lsof -nP -iTCP:8188 -sTCP:LISTEN                 -> nothing listening
```

Waiting ~2 minutes with `curl --retry 40 --retry-connrefused` never got it back.

Two distinct problems:

1. **Misleading tool result.** The tool says the restart did NOT happen while the server was in fact being stopped. An agent that trusts the result will keep issuing panel calls that all fail with a confusing "panel is reconnecting" message, when the real state is "ComfyUI is dead".
2. **No recovery on Pinokio-managed installs.** This ComfyUI runs under Pinokio (`~/pinokio/api/comfy.git`, `deploy_environment: local-git`). Pinokio's `start.js` only re-launches on the Manager's dependency-install message (`/\[ComfyUI-Manager\] Restarting to reapply dependency installation/`). A plain Manager restart kills the process and matches no relaunch event, so ComfyUI stays down permanently and the panel session is unrecoverable.

### Steps to reproduce

1. Install ComfyUI via Pinokio (`comfy.git`, ComfyUI 0.28.3, Manager 3.41 legacy).
2. Add custom node packs to `custom_nodes/`.
3. Call `panel_restart_comfyui`.
4. Observe the returned text vs. the actual state of port 8188.

### Exact error (scrubbed)

Tool result:
```
Cancelled — ComfyUI was not restarted.
```

Every panel call afterwards:
```
Error: graph_outline could not reach the ComfyUI panel — it is still reconnecting after a restart/reload.
(no connected tab with id "tmp:...". Connected: none — the ComfyUI panel tab is not connected ...)
```

### Workaround used

Relaunched by hand with the same env/flags Pinokio uses:
```bash
cd ~/pinokio/api/comfy.git/app
[...]

PR #1497, opened 2026-08-13
fix(742): report a remote ComfyUI that a restart stopped and nothing brought back

artokun:
Refs #742 — the thread's **2026-08-12 recurrence on 0.51.18**, not the original report (which already has a preflight shipped).

## The bug

`captureRebootHealthBase` is loopback-only, so for a REMOTE-classified target `recoveryPromise` was `null`, nothing was ever probed, and the handler returned a fixed note:

> ComfyUI restart was dispatched and accepted; **it is restarting out-of-band**. There is no local boot endpoint I can safely probe from here … Check `get_system_stats` in a few seconds.

Identical whether the server was mid-restart or dead for good. The reporter's Pinokio install was addressed by its LAN IP, classified remote, stopped, and never relaunched — Pinokio's launcher only re-launches on the Manager's dependency-install signal — while the result described a restart in progress. They discovered the truth from every later panel call failing with a generic "still reconnecting".

This exits at the `healthBase == null` guard, a field-for-field match with their reported `{rebooting:true, ready:false, confirmed_cycle:false, dispatched:true}` — the other branch would also have carried `recovered_ms`/`probes`/`saw_down`.

## The change

The base this session already talks to for every other call **is** probeable, with the same configured auth, so it is watched now too.

What it may conclude is deliberately one-directional: it can report a failure to come back, and can **never** manufacture readiness. `ready` and `confirmed_cycle` stay false on every branch of this guard — a healthy answer proves the *address* responds, not that this instance cycled, which is the same reason the loopback path demands a handshake-Origin match before certifying anything.

No change to what gets refused, so a working supervised-remote restart (the tunnelled Desktop app) is untouched.

## Three codex rounds, three real findings

1. **The note predicted the future.** It said "do not wait for it … it will stay down" — from a bounded window. A remote host can simply be slower to boot than the budget, so that would send someone to hand-start a server seconds from returning. Stating unmeasured advice confidently is the defect #742 is *about*; pointing it the other way is not a fix.
2. **A post-window probe could start after the handler's own deadline.** Deleted rather than clamped — the reworded note no longer needs it.
3. **The trigger read a LATCHED flag.** `sawDown` never clears, so one transient refusal followed by a responding endpoint (401/503/timeout all classify `unknown`) reported a live install as dead. `observeRecovery` now also returns `lastStatus`, and the trigger requires the **final** sample to still be `down`.

Two further predictions were removed in the same pass: "nothing is going to bring it back on its own" forecasts a supervisor's behaviour past the window, and "if it starts answering, it was just slow" rules out a manual relaunch or tunnel/proxy recovery that would explain it equally well.

Round 3: **SHIP**, no defects found.

## Verification

- 5 tests driving the real handler with the target genuinely retargeted (`setComfyuiTarget`) rather than `isRemoteMode` stubbed — the branch keys on the mode *and* the configured base, and stubbing one would pass while proving nothing about the pairing that ships
- the down→stayed-down test asserts it really probed the configured base, so it can't pass on a branch that looked at nothing
- the down→unknown guard first asserts the latch **did** fire, so it can't pass for the wrong reason; it dies if the `lastStatus` gate is removed
- both over-claim directions pinned by absence assertions
- mutation-checked; full suite 9174 green; `lint`, `check:vocabulary`, `check:unknown-collapse`, `vocab:export --check`, `asset-counts --check` all pass

## Not in this PR

Pinokio-aware *pre*-restart guidance (issue item 3) would mean refusing or warning more on a destructive path, and the surrounding code has already rejected asymmetric variants of that with reasons. This PR is disclosure-only and safe to ship on its own.

### DIFF
--- a/src/__tests__/orchestrator/remote-restart-never-came-back.test.ts
+++ b/src/__tests__/orchestrator/remote-restart-never-came-back.test.ts
@@ -1,0 +1,145 @@
+// #742 — a REMOTE-classified ComfyUI that a restart stops and nothing brings back.
+//
+// `captureRebootHealthBase` is loopback-only, so on the remote path nothing was ever
+// probed: the dispatch returned "it is restarting out-of-band … check in a few seconds"
+// whether the server was mid-restart or dead for good. The reporter's Pinokio install was
+// addressed by its LAN IP, classified remote, stopped, and never relaunched — Pinokio's
+// launcher only re-launches on the Manager's dependency-install signal — and the result
+// described a restart in progress the whole time.
+//
+// The base this session already talks to IS probeable, so it is now watched. What it may
+// conclude is one-directional and that is the point of these tests: it can report a
+// failure to come back; it must never manufacture readiness, because a healthy answer
+// proves the ADDRESS responds, not that this instance cycled.
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const resetClient = vi.fn();
+const resetObjectInfoCache = vi.fn();
+vi.mock("../../comfyui/client.js", () => ({
+  getObjectInfo: vi.fn(),
+  backfillObjectInfo: vi.fn(),
+  resetClient: () => resetClient(),
+  resetObjectInfoCache: () => resetObjectInfoCache(),
+}));
+
+const { buildPanelToolDefs, __panelToolsTestHooks } = await import(
+  "../../orchestrator/panel-tools.js"
+);
+import type { PanelToolCtx, ToolResult } from "../../orchestrator/panel-tools.js";
+import { getComfyUIBaseUrl, isRemoteMode, setComfyuiTarget } from "../../config.js";
+
+/** The reporter's shape: a ComfyUI on this LAN, reached by IP, so classified REMOTE. */
+const REMOTE_URL = "http://192.168.1.50:5000";
+let originalTarget: string;
+
+const parse = (res: ToolResult): Record<string, unknown> =>
+  JSON.parse(res.content.find((c) => c.type === "text")!.text as string);
+
+/** A ctx whose comfy_reboot is ACCEPTED — the reporter got `rebooting:true` too. */
+function ctxReboot(): PanelToolCtx {
+  return {
+    call: async () => {
+      throw new Error("ctx.call must not be used for reboot dispatch");
+    },
+    confirm: async () => "yes" as const,
+    ensureReachable: () => {},
+    bridge: {
+      send: async () => ({ rebooting: true }),
+      tabOrigin: () => undefined,
+      tabIsLocal: () => false,
+      tabServerOrigin: () => undefined,
+      tabSockId: () => "s0",
+      canReach: () => false,
+    } as unknown as PanelToolCtx["bridge"],
+    tabId: "test-tab",
+  } as unknown as PanelToolCtx;
+}
+
+function reboot() {
+  const def = buildPanelToolDefs().find((d) => d.name === "panel_restart_comfyui");
+  if (!def) throw new Error("panel_restart_comfyui not found");
+  return def.handler;
+}
+
+beforeEach(() => {
+  originalTarget = getComfyUIBaseUrl();
+  // Retarget for real rather than stubbing isRemoteMode: the branch under test keys on
+  // BOTH the mode and the configured base, and a stub of one with the other left local
+  // would pass while proving nothing about the pairing that actually ships.
+  setComfyuiTarget(REMOTE_URL);
+  __panelToolsTestHooks.setPanelRebootTiming({
+    settleMs: 0,
+    budgetMs: 60,
+    intervalMs: 5,
+    probeTimeoutMs: 10,
+  });
+});
+
+afterEach(() => {
+  setComfyuiTarget(originalTarget);
+  __panelToolsTestHooks.setPanelRebootTiming(null);
+  __panelToolsTestHooks.setHealthProbe(null);
+  __panelToolsTestHooks.setLocalRestartPreflight(null);
+});
+
+describe("a remote ComfyUI that never comes back is reported, not described as restarting (#742)", () => {
+  it("the target is genuinely classified REMOTE — the premise, asserted not assumed", () => {
+    expect(isRemoteMode()).toBe(true);
+    expect(getComfyUIBaseUrl()).toContain("192.168.1.50");
+  });
+
+  it("WENT DOWN AND STAYED DOWN: says so, and names the launcher", async () => {
+    // Every probe refuses: the server stopped and nothing relaunched it.
+    const bases: Array<string | null> = [];
+    __panelToolsTestHooks.setHealthProbe(async (base) => {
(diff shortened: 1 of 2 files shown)


artokun:
## Diagnosis (read, not yet measured)

The skip is structural, and the codebase already argues the exact point — in the opposite direction.

`captureRebootHealthBase` (panel-tools.ts) returns `null` when `isRemoteMode()`, and also requires `isLoopbackOrigin(bootBase)`. A Pinokio ComfyUI on `192.168.x.x:5000` is therefore never bound, so `preflightBound = false`. The refuse gate is:

```ts
if (!preflightBound && !isRemoteMode() && !isCloudMode()) { /* refuse */ }
```

`isRemoteMode()` is true, so the refusal is skipped and the Manager reboot dispatches — reproducing the reporter's `{rebooting:true, ready:false, confirmed_cycle:false, dispatched:true}` exactly.

The stated reason for excluding remote:

> there is no local process to assess, and the Manager reboot is their ONLY restart path — a supervised remote (the tunnelled Desktop app) restarts through it by design

Both halves fail for this install. There **is** a local process (`COMFYUI_PATH` names it on this machine), and Pinokio does **not** relaunch after a plain Manager restart — its `start.js` only re-launches on the Manager's dependency-install message. So "remote ⇒ supervised ⇒ it comes back" is the load-bearing assumption, and it is an assumption about the *route*, not the *instance*.

The refusal note a few lines below already makes this argument for the mirror case — a cloud pod fronted at `127.0.0.1:<port>` classifying as local:

> it is being classified as local because that classification reads the HOST — which proves the route is local, not the instance

A LAN IP likewise proves the route is not loopback, not that the instance is on another machine.

## Where I'll take it

Not toward "refuse more" on a destructive guard — the surrounding code has already rejected asymmetric variants of that, with reasons. The candidate that survives that bar is the issue's own item 2, which is pure disclosure and cannot break a working path: after a dispatch on the remote path, poll for return within a bounded window and, when it never comes back, **say so** with the recovery, instead of returning `ready:false, confirmed_cycle:false` and leaving the caller to infer death from a generic "panel is reconnecting".

Next: confirm on the rig whether the remote path already polls at all, and what it reports when the server stays down — before writing anything.

artokun:
## Refined: the remote path asserts something it never measured

Tracing the post-dispatch path narrows this from "remote skips the preflight" to a second, sharper defect.

The reboot observer probes `captureRebootHealthBase(ctx)`, which returns `null` in remote mode. So on the remote path nothing is ever probed, no down→up cycle can be observed, and the handler lands on this note:

> The reboot command was sent but I could NOT confirm ComfyUI actually cycled within Ns (**it never went down** — the panel may have merely disconnected/inferred a reboot without one).

For the reporter's install that parenthetical is **false**. ComfyUI *did* go down; it just was never watched. The sibling note one branch over — `sawDown` — says the accurate thing:

> Reboot was dispatched and ComfyUI went down, but it has not become healthy within Ns … do NOT assume it is back.

…and it is unreachable from remote mode by construction.

So the remote path states an observation it did not make, in the direction that reads as reassuring, on the exact failure it exists to describe. That is the same defect class the repo already polices elsewhere (reporting the request rather than the observed effect).

**Shape of the fix:** the orchestrator *does* have a reachable base for a remote target — the configured `COMFYUI_URL` it just dispatched through. Letting the observer probe that (rather than the loopback-only boot base) makes both halves honest: a real cycle can certify, and a server that goes down and stays down is reported as such with the recovery, instead of as "it never went down".

Scoped deliberately: no change to what gets refused, so a working supervised-remote restart keeps working. Next step is measurement on the rig — a remote-classified target that is stopped and not relaunched — before any code.

artokun:
## Correction — I named the wrong branch

My previous comment said the remote path lands on the `it never went down` note. It does not. Tracing the returns properly, the remote path exits earlier, at the `healthBase == null` guard:

```
rebooting: true, ready: false, confirmed_cycle: false, dispatched: true
note: "ComfyUI restart was dispatched and accepted; it is restarting out-of-band.
       There is no local boot endpoint I can safely probe from here, so I can't
       confirm it finished coming back … Check get_system_stats in a few seconds."
```

That is a field-for-field match with the reporter's `{rebooting:true, ready:false, confirmed_cycle:false, dispatched:true}` — the other branch would also have carried `recovered_ms`, `probes` and `saw_down`, and their report has none of them. The `it never went down` note is only reachable when `healthBase != null`, i.e. never on this path.

So that note is not the defect. Two real ones remain, and they are smaller and sharper than what I claimed:

1. **"it is restarting out-of-band" is an unobserved assertion.** Nothing here watched a relaunch, and for a Pinokio install none happens — its `start.js` only re-launches on the Manager's dependency-install message. The sentence tells the reader the thing that is precisely not true for them.
2. **"There is no local boot endpoint I can safely probe from here" is true and beside the point.** For a remote target there is no *loopback boot* endpoint — but there is the configured `COMFYUI_URL`, the address this session uses for every other call. It is never probed, so a target that goes down and stays down is indistinguishable here from one that is mid-restart.

## Scope, tightened

Disclosure only — no change to what is refused, so a working supervised-remote restart is untouched.

After a remote dispatch, probe the **configured base** over the existing bounded window and use the result in the **negative direction only**: if the address the session talks to is still unreachable when the window closes, say so and name the recovery. It must not be allowed to certify readiness — a healthy answer proves the address responds, not that this instance cycled, and manufacturing `ready:true` from that is the failure mode the surrounding code is built to avoid.

That is sound regardless of instance identity: "the ComfyUI this session targets is not answering" is a fact about the session's own target, which is what the user is stuck on.

artokun:
## Live check on the rig — the premise, confirmed; the death path, not reproduced (deliberately)

Started a throwaway orchestrator with the reporter's configuration shape — a LAN IP for `COMFYUI_URL`, `COMFYUI_PATH` pointing at a local install:

```
[comfyui-mcp] WARNING: Both COMFYUI_URL (remote ComfyUI at 192.168.1.179:8188) and
  COMFYUI_PATH (C:\Users\Artokun\ComfyUI-Installs\ComfyUI\ComfyUI) are set.
[panel-orchestrator] env: … comfyui=? (REMOTE) …
[panel-orchestrator] ready — … comfyui=http://192.168.1.179:8188 — remote target …
```

So the premise holds in a real process, not just in a test: **a ComfyUI on this very machine, addressed by its LAN IP, classifies as `REMOTE`** — which is what routes it past the preflight and into the `healthBase == null` branch. The startup even warns that the URL/PATH combination is contradictory, and the restart path then proceeds as though there were no local install to account for.

**What I did not do:** reproduce the stop-and-never-return end to end. That needs an unsupervised ComfyUI to be killed for real, and the only ComfyUI on this rig is the user's live one — stopping it to watch a message render is not a trade worth making, and standing up a fake target does not exercise the path (the reboot goes through the panel to ComfyUI-Manager, which a stub does not have).

That half is covered by 5 tests against the real handler with the target genuinely retargeted, including the two false-alarm directions codex found. Stating it plainly rather than implying a full live pass: **premise verified live, message path verified by test.**

Issue #1429, opened 2026-08-11
A ComfyUI retarget landing mid-turn leaves that turn's comfyui tools on the OLD server — silently


artokun:
## A ComfyUI retarget that lands mid-turn leaves that turn's comfyui tools pointed at the OLD server — silently

Found while investigating #1415. This is a separate defect and, unlike #1415, it is proven by code already in the repo.

**The window.** When the ComfyUI target changes — a panel `hello` advertising a different `comfyui_url`, or a local→pod switch — `onComfyuiTargetChanged` (`src/orchestrator/index.ts:5860`) calls `manager.restartAllForMcpEnv()`. That respawns each tab's comfyui MCP child with the new `COMFYUI_URL`, because `config` is a module-load snapshot (`src/config.ts:487`) and a running child can only learn a new target by being restarted.

But the respawn is deferred when the tab is mid-turn — `src/orchestrator/panel-agent.ts:2192`:

```ts
// Apply immediately when the tab is already idle; otherwise it fires on the
// next turn-done via applyPendingRestarts().
```

and `applyPendingRestarts` returns `"scheduled"` on `agent.isBusy || agent.hasPending`. The repo's own suite already pins this — `src/__tests__/orchestrator/mcp-env-restart-outcome.test.ts:94`:

> `it("returns 'scheduled' while the agent is mid-turn — the replacement has NOT happened")`

The deferral itself is right; you cannot pull an MCP subprocess out from under a running turn. The defect is what happens *inside* the window.

**Why it matters more than a failed fetch.** For the remainder of that turn, every comfyui tool call — `list_packs`, `enqueue_workflow`, `upload_image`, `download_model` — is served by a child still addressing the previous target. Two outcomes:

- **Old target is dead** → confusing failures. `comfyuiFetch` names the target it tried (#954), so the user sees an address they believe they already left. This is the shape #1415 reports.
- **Old target is alive** → *worse*: the call silently succeeds **against the wrong machine**. Switch a tab from a pod back to local mid-turn and an `enqueue_workflow` in that turn renders on the pod; an `upload_image` writes into the pod's `input/`. No error, wrong machine, and nothing in the transcript says so.

The second case is the reason I am filing this rather than folding it into #1415. It is a wrong-answer/wrong-destination bug, not just a bad message, and it is invisible at the point it happens.

[...]

PR #1500, opened 2026-08-13
fix(1415): ask the browser for workflow templates, not the headless target

artokun:
Claims #1415. (An earlier claim marker, #1428, was closed without code when the investigation redirected into #1429; the endpoint half was then confirmed still present on main.)

**Measured, on main:** `list_templates` builds `${base}/api/workflow_templates` in `src/tools/skills-access.ts`, where `base` is the **headless** `COMFYUI_URL`. There is no panel-side template command to route it through, so a connected sidebar panel cannot answer this question at all — which is what the reporter hit with a stale/unreachable `COMFYUI_URL`.

**Shape is already settled by precedent, so this is implementation rather than design:**
- **#1006** fixed exactly this for `/object_info` — a panel bridge command where the **frontend fetches its own ComfyUI origin** and returns the result. The browser is by definition able to reach that host; it is already talking to it.
- **#1359** fixed the *diagnosis* for a sibling endpoint, which is why today's failure already explains the split instead of just failing.

**The shortcut both issues rejected**, restated so it does not get re-proposed: having the orchestrator fetch the panel's *observed origin* directly. It works for a public proxy URL and fails identically in a tunnel or loopback-only topology — where the browser is the only thing that can reach that ComfyUI — trading one silent failure for a narrower one, and it adds an outbound fetch to an origin the client chose.

Cross-repo: a new bridge command in the panel, plus routing here with a fallback to the current headless fetch when no panel is connected. Starting from #1006's command as the template so the two paths do not diverge.


artokun:
## Structural finding: the one-line reroute is not available here, and here is exactly why

Measured on main, not inferred:

- `listWorkflowTemplatesAction()` (`src/tools/skills-access.ts`) builds `${getComfyUIBaseUrl()}/api/workflow_templates`. Confirmed still present.
- `list_packs` is registered by `registerSkillsAccessTools` via `registerAllTools`, and for a panel agent that runs in the **per-tab comfyui MCP child process** — spawned by `buildMcpServers()` and configured by `COMFYUI_URL` env (see `retargetAllForMcpEnv`, and #1429's mid-turn deferral of exactly that respawn).

That child has **no bridge handle at all**, so `#1006`'s pattern — `ctx.bridge.send({cmd: "graph_get_object_info"})` — is not a thing this handler can do. `panelObjectInfo` works because it lives in `panel-tools.ts`, inside the orchestrator process, with a `ctx` carrying `bridge` and `tabId`. The templates action has neither.

So my own PR description's "routing here with a fallback" understated it: there is no channel to route *through* yet.

## Also: the fallback I proposed is wrong, and #1006 already says so

> FAIL CLOSED. If the panel cannot serve definitions we surface that; we do NOT fall back to COMFYUI_URL. A fallback would convert the live canvas against a DIFFERENT ComfyUI's schema … since the two hosts can both answer and disagree.

Templates are less dangerous than a schema — a wrong list misleads rather than corrupts — but the same shape applies: two hosts can both answer and disagree, and silently listing another machine's templates is how someone is told a starter exists that their server has never heard of. Whatever ships should say which server answered.

## Viable designs, and the one I'd take

1. **Move the read to where the bridge is** — serve the templates action from the orchestrator's panel surface, delegating to the panel's own fetch, and leave the headless child's version as the no-panel path. Smallest change that respects the existing split; costs a tool-surface decision, which the vocabulary ratchet makes deliberate rather than incidental.
2. **Give the child a callback channel to the orchestrator.** General, and much larger — a new IPC surface on a process boundary that currently has none, for one endpoint.

Leaning hard on (1). Both need a new panel-side command (the browser fetching its own `/api/workflow_templates`), so the panel-repo half is common to either.

Next: confirm which surface actually serves `list_packs` for each backend before choosing — the answer decides whether (1) reaches the agents that hit this bug, and that is precisely the "tested but unreachable" trap. Not writing code until that is measured.

artokun:
No code — a claim marker. The measurement on #1415 showed the reported symptom self-heals via the panel-hello retarget, and what remains is a tunnel/loopback-only gap whose fix adds tool surface during consolidation. Detail on the issue; releasing the claim rather than holding it while that decision is open.

PR #1501, opened 2026-08-13
fix(1483): a paid third-party service node is not "local, no paid credits"

### LINKED ISSUE #1483: check_runtime misclassifies fal-api generation nodes as local and free
`list_packs(action:"check_runtime")` classified a workflow containing the installed third-party node `NanoBananaPro_fal` as `runtime:"local"` and `usesApiNodes:false`. The classifier only recognizes ComfyUI partner nodes with `api_node:true` or categories under `api node/`. The fal-api pack registers paid external nodes under categories such as `FAL/Image` without ComfyUI's partner-node marker, so paid FAL requests are incorrectly declared free.

### Steps to reproduce
1. Install the `fal-api` custom node pack.
2. Build or inspect a workflow containing `NanoBananaPro_fal`.
3. Call `list_packs` with `action:"check_runtime"` and that graph.
4. Observe that `NanoBananaPro_fal` appears in `classTypes`, `unknownNodes` is empty, and the result is `local/free`.

### Exact error (scrubbed)
```json
{
  "runtime": "local",
  "usesApiNodes": false,
  "apiNodes": [],
  "classTypes": ["PreviewImage", "NanoBananaPro_fal", "ImpactMakeImageBatch", "LoadImage"],
  "unknownNodes": [],
  "guidance": "Local-GPU / free — every node runs on the user's own GPU, no paid credits."
}
```

### Fix
Applied locally: yes, one attempt. Upstream-only/effective in this session: yes; the running tool server still returned the old result after `panel_reload`, so the workflow was handled conservatively as paid and no further patch attempts were made.

Proposed minimal patch to `dist/services/api-nodes.js`:
```diff
 const API_CATEGORY_PREFIX = "api node";
+const EXTERNAL_API_CATEGORY_PREFIXES = ["fal/"];
 ...
- return category === API_CATEGORY_PREFIX || category.startsWith(`${API_CATEGORY_PREFIX}/`);
+ return category === API_CATEGORY_PREFIX ||
+   category.startsWith(`${API_CATEGORY_PREFIX}/`) ||
+   EXTERNAL_API_CATEGORY_PREFIXES.some((prefix) => category.startsWith(prefix));
```
A source-level fix should ideally distinguish FAL execution nodes from purely local `FAL/Utils` helpers, while never returning confirmed `local/free` for external FAL generation nodes.

Direct module verification of the patched predicate returned `[true,false,true]` for FAL/Image, local sampling, and official api_node:true respectively. The live MCP call remained unchanged after reload.

### Environment
Windows 11
ComfyUI 0.32.0 (LOCAL)
GPU NVIDIA GeForce RTX 4080 16 GB
comfyui-mcp 0.51.17
panel 0.14.16
Python 3.13 / torch 2.12.1 / CUDA 13.0

artokun:
Fixes #1483 — **P1**: this verdict guards the user's money.

`check_runtime` returned `runtime:"local"`, `usesApiNodes:false` and *"every node runs on the user's own GPU, no paid credits"* for a graph containing `NanoBananaPro_fal`, a node whose whole job is to bill a fal.ai endpoint.

**Being installed is what made it confidently wrong.** The node is in `/object_info`, so it was never `unknown`, and known-and-unmarked resolved to local.

## Measured, not reasoned about

Against a real 4304-node `/object_info`:

- **0 of 3464** custom-node-registered classes carry `api_node:true` — all 220 marked nodes are core. Comfy's partner marker can never fire for a third-party pack, so this is a category-wide hole that fal.ai happens to expose.
- **3464 of 4304** classes come from `custom_nodes.*`, which kills "third-party ⇒ unclassifiable": every real workflow would warn, and a warning that always fires protects nobody.
- **49** classes expose a credential input — a real signal, and not enough on its own (below).

## Why the obvious fix was rejected

The report proposes a hardcoded `fal/` category prefix. One string long, and wrong in both directions — it misses any pack that does not put the vendor in its category, and it catches `FAL/Utils`, local helpers that resize an image before upload and spend nothing.

The elegant alternative — "flag anything taking an `api_key` input" — **fails on the reported node**. I checked the pack's source (`gokayfem/ComfyUI-fal-API`, `nodes/image_node.py`): its NanoBanana classes take `prompt`, `aspect_ratio`, `num_images`, `seed` and nothing else; the FAL key comes from the environment. Shipping only that would have closed this issue with the reporter's own graph still reported as free — so **the test fixtures are the report's own node**, not a convenient one.

## The change

Two signals, covering different populations:

1. an **enumerated** pack registry, matched on `python_module` **and** category prefixes (a renamed clone directory changes the module path and nothing else), with per-pack exemptions for their own local helpers;
2. a declared **service credential** input — the general catch for packs nobody enumerated.

Both feed a new `externalApiNodes[]` that counts toward `usesApiNodes`/`runtime` but stays **out of `apiNodes[]`** — that list also drives `listApiNodes` and the 3D picker, which hand out schemas assuming Comfy's auth model. A fal.ai node is paid, but it is not a partner node.

The guidance and tool description now name **who bills** — "fal.ai", on the user's own account with them, *not* Comfy api credits. A reader who checks their Comfy balance, finds it untouched, and concludes the warning was wrong is the exact failure this verdict exists to prevent.

## Two codex rounds — round 1 found six, three of them P1 false-negatives

| | |
|---|---|
| **P1** | a **renamed** install directory escaped the registry entirely |
| **P1** | the credential pattern knew only `api_key` — `api_token`, `client_secret`, camelCase `apiToken` all slipped through |
| **P1** | registry **order**: `ComfyUI-fal-API-Flux` bound to the broader `comfyui-fal-api` entry and inherited its `FAL/Utils` exemption, so the stricter entry written for that pack was never reached. Every match is now considered and **paid wins** — disagreement between entries is not evidence of free |
| **P2** | `secret_key` false-positived a local crypto node → removed; the remaining terms each name a remote-service credential |
| **P2** | the provider was never actually named |
| **P2** | the documented array was omitted on exactly the graphs that look safe |

Round 2: **SHIP**, no failing input found.

## Verification

- 10 tests, built on the report's own node and the measured `/object_info` population
- each of the three P1 holes has its own regression test
- both signals **independently mutation-killed** — disabling the registry fails 2 tests, disabling the credential catch fails a different one
- the false-positive direction is pinned too (ordinary third-party nodes, a paid pack's local helper, a local `secret_key` crypto node)
- full suite 9184 green; `lint`, `check:vocabulary`, `check:unknown-collapse`, `vocab:export --check`, `asset-counts --check` all pass


artokun:
## Measured against a real `/object_info` (4304 nodes), and it rules out the fix I was going to write

**1. The gap is total, not FAL-specific.** Of 4304 registered nodes, 3464 come from `custom_nodes.*`, and **exactly 0 of them carry `api_node: true`**. All 220 partner-marked nodes are core. So the marker `isApiNode` relies on can never fire for a third-party pack — FAL is one instance of a category-wide hole.

**2. "third-party ⇒ unknown" is a non-starter.** It would flag 80% of every real workflow (Impact Pack, KJNodes, RES4LYF…). The reporter's own graph contains `ImpactMakeImageBatch`, which is genuinely local. Not shippable.

**3. The structural signal I liked does NOT cover the reported node.** 49 custom nodes here take an api-key-ish input (`FalImage2VideoRequestNode`, `FalNanoBananaNode`, `NovitaVideoRequestNode`, `easy joyCaption2API`…) — a principled, cross-vendor property: a node asking for an external credential cannot be *confirmed free*. But I checked the reporter's actual pack (`gokayfem/ComfyUI-fal-API`, `nodes/image_node.py`) and its NanoBanana classes take **no credential input at all** — `prompt`, `aspect_ratio`, `num_images`, `seed`, … only; the FAL key comes from the environment via its `ApiHandler`.

So the elegant signal misses the exact node in the report. Had I shipped it, this issue would have closed with the reporter's workflow still reported as free.

**4. Descriptions are useless here** — 0 custom nodes mention key/credits/billing.

## What is actually available for `NanoBananaPro_fal`

Only `category: "FAL/Image"` and `python_module: custom_nodes.ComfyUI-fal-API`. There is no universal structural fact in `/object_info` that says "this calls a paid service" — for this pack the information genuinely is not there.

## Design, now grounded rather than guessed

Two complementary signals, because neither alone covers the reported case:

1. **An enumerated registry of known external-service packs**, matched on `python_module` (pack identity — more stable than a category string a pack can restyle) with category as a secondary. Enumerated and named, not a keyword regex: a `fal/` prefix reads as coverage while being one string long, and a regex over vendor words silently catches `FAL/Utils` helpers that are purely local.
2. **The credential-input signal** as the general catch — it covers the 49 nodes here that a vendor list would miss, and it keeps working for packs nobody has enumerated yet.

Both route into `usesApiNodes: true` so `runtime` becomes `api`/`mixed`, which is the only outcome that stops the credit-spend flow.

The blast radius is small and measurable: 49 nodes of 4304 (1.1%) from signal 2 on this install, plus whatever the registry names — so this does not degrade into "everything is unknown".

Next: implement both, and test signal 1 against the report's own node (`NanoBananaPro_fal`, `FAL/Image`, no api-key input) rather than against a node that happens to be easy to catch.

artokun:
## Live before/after on the rig — real `/object_info`, not fixtures

This machine has fal.ai nodes installed from a **different** pack than the reporter's (`comfyui-utils-nodes`), which makes it a stronger test: it confirms the hole is category-wide rather than specific to `ComfyUI-fal-API`.

**BEFORE** — the shipped version (0.51.22), driven through the running MCP server:

```json
{ "runtime": "local", "usesApiNodes": false, "apiNodes": [], "unknownNodes": [],
  "guidance": "Local-GPU / free — every node runs on the user's own GPU, no paid credits." }
```

for a graph of `LoadImage` + **`FalImage2VideoRequestNode`** + `PreviewImage`. That node calls fal.ai and takes an `api_key`, and the tool told the agent it was free.

**AFTER** — this branch's build, same graph, same live server:

```json
{ "runtime": "mixed", "usesApiNodes": true,
  "apiNodes": [], "externalApiNodes": ["FalImage2VideoRequestNode"], "unknownNodes": [],
  "guidance": "This workflow calls a PAID THIRD-PARTY SERVICE — FalImage2VideoRequestNode —
   billed by the service it authenticates to on the user's own account with them
   (NOT Comfy api credits), so it is NOT free even though every node is installed locally. …" }
```

Note `externalProviders` is **absent** here and the message says *"the service it authenticates to"* rather than naming one — this node was caught by the credential signal, not the registry, so we know it authenticates somewhere but not to whom. That is the designed behaviour, and it is nice to see it hold on real data instead of only in a fixture.

**The false-positive direction, same rig:** `CheckpointLoaderSimple` + `KSampler` + `ImpactMakeImageBatch` + `VAEDecode` + `SaveImage` →

```json
{ "runtime": "local", "usesApiNodes": false, "externalApiNodes": [] }
```

still free, still confident, including the third-party Impact node. That is the 3464-of-4304 population that must not start warning.

Issue #1489, opened 2026-08-12
Cancelling a large batch floods every subsequent turn with duplicated "could not be confirmed after reconnecting" events (one per cancelled prompt, re-sent cumulatively)


artokun:
### What happened / root cause

Two separate defects hit in one session on a long-running scene-chain workflow (27-scene batch, ~5 min/scene).

**A) Cancelled-batch event flood (the expensive one).**

Workflow: `panel_run(batch_count: 27)` → the orchestrator tracks all 27 prompt ids. The user asks to stop → `queue(action:"cancel", clear_pending:true)` interrupts the running job and clears the 26 pending ones. ComfyUI never writes history entries for prompts that were cleared while pending, so on the next websocket reconnect the orchestrator cannot resolve any of them and emits, per prompt:

```
[panel event] ⚠️ The workflow run you queued ERRORED on the user's canvas: Render status for
prompt <id> could not be confirmed after reconnecting (no history for it) — it was likely
cancelled or interrupted. Safe to requeue.. STOP — do not carry on as if it succeeded.
Diagnose it (panel_get_errors has the full node-level details) and fix it.
```

That alone would be one line per prompt. What actually happened is worse: the block is **re-sent cumulatively on every turn**, growing each time — turn N carried the events for prompts 1..N, turn N+1 carried 1..N+1, and so on. Over the session this reproduced the same ~25 messages dozens of times and consumed a very large share of the context window on pure noise. It also repeatedly re-injected the user's older prompt text alongside it.

Two things compound it:
1. The event is classified as an **ERROR** with a "STOP — do not carry on as if it succeeded" directive, when the true cause is a **user-requested cancellation** — something the orchestrator itself performed and therefore already knows about. An agent that follows the directive literally starts diagnosing a non-existent failure.
2. The events are not deduplicated or coalesced across turns.

**B) `panel-op.lock` is never auto-reclaimed after an orchestrator dies.**

Every panel sync attempt failed for ~3 hours with:

```
Timed out after 60000ms waiting for the panel operation lock. The lock at
~/.comfyui-mcp/panel-op.lock was taken 3 hours ago by pid 26200 (started ...),
and that process is no longer running. The lock is never auto-reclaimed
```

[...]

Issue #889, opened 2026-08-04
Panel: run-errored notification claims "the workflow run you just queued" for runs the agent never queued


artokun:
### What happened / root cause

The agent received this panel event:

```
[panel event] The workflow run you just queued ERRORED on the user's canvas: Render status for prompt 9d70fd9d-... could not be confirmed after reconnecting (no history for it) — it was likely cancelled or interrupted. Safe to requeue.. STOP — do not carry on as if it succeeded. If it relates to what you were doing, diagnose it (panel_get_errors has the full node-level details) and fix it
```

The agent had **never called `panel_run`** in the entire session — the only calls made were reads (`panel_graph_outline`, `panel_query_graph`, `panel_get_subgraph`, `panel_screenshot`) and graph edits (`panel_add_node`, `panel_connect`, `panel_set_widget`, `panel_create_subgraph`, ...). The notification text asserts agent attribution ("the workflow run **you just queued**") unconditionally, apparently as a fixed template, regardless of whether the run originated from the agent or from the user pressing Queue in the UI.

Effect: the agent is pushed into diagnosing a failure it did not cause and cannot own. It burned a `panel_get_errors` round-trip (which returned clean: `errored_count: 0`, `last_execution_error: null`), and the strong imperative wording ("STOP — do not carry on as if it succeeded") makes the false attribution costly rather than merely cosmetic — it invites the agent to invent a connection to whatever it was actually doing.

### Steps to reproduce

1. Start a session where the agent only reads/edits the graph and never calls `panel_run`.
2. Have a prompt in flight (or a stale prompt id) when the panel reconnects, such that its history entry cannot be resolved.
3. The agent receives a run-errored event phrased as though it queued the run.

### Exact error (scrubbed)

Not an exception — a misattributed notification. Body quoted above; prompt id is a normal ComfyUI uuid.

### Fix

`applied locally: no`  `upstream-only: yes`

This user's ComfyUI is REMOTE (no `COMFYUI_PATH`, no local checkout — a filesystem-wide `find` for `comfyui-mcp-panel.js` returned nothing), so the panel pack is not reachable from the orchestrator machine and no local patch could be attempted or verified.

[...]

PR #1507, opened 2026-08-13
fix(1489): coalesce a burst of run_errors instead of nesting them

artokun:
Refs #1489 — fixes the **flood**, which the reporter calls the expensive half. The classification half is panel-side and identified below; this PR does not close the issue.

## The bug, measured before anything was changed

A cancelled 27-scene batch made every following turn carry a **growing** block of "could not be confirmed after reconnecting" errors — turn N carried prompts 1..N — dragging the user's original message along at the bottom.

Reproduced against the real `injectRunError` with a backend that holds its turn:

| turn | length | carries |
|---|---|---|
| 0 | 11 | `USER_PROMPT` |
| 1 | 419 | ONE |
| 2 | 827 | ONE + TWO |
| 3 | 1237 | ONE + TWO + THREE **+ USER_PROMPT** |

**"Not deduplicated" is the symptom, not the cause.** Each notice names a different prompt id, so all N texts are distinct — a dedupe would have changed nothing. The cause is mechanical: `injectRunError` interrupts the in-flight turn with `requeueInFlight: true`, which **re-queues** it, and after the first error that turn is itself an error turn. So error N re-queued errors 1..N-1 and the drain batched them. It also explains the detail a dedupe theory cannot: the user's own message kept reappearing, because the original turn sits at the bottom of that same re-queue stack.

## The fix

A burst coalesces instead of nesting. An error arriving while an error turn is **queued** folds into it; one arriving while an error turn is **in flight** queues normally instead of interrupting again. A single error interrupting ordinary work is unchanged — that part was working as intended.

`some`, not `every`, and that distinction *is* the fix: the in-flight turn is a **batch** (the drain merges the re-queued user message with the error notice), so an error turn's items are `[error, USER_PROMPT]`. Written with `every` first, the branch never fired and the re-measured turn lengths came back **byte-identical to the bug** — the only reason I caught it.

## Codex found a P0 that I introduced

Coalescing keeps ONE `mid` per item, and the mid pins the error-handling turn to the erroring workflow's tab — #884's P0, *"a render error on A silently editing B"*. Folding tab B's notice into tab A's item **discarded B's origin**, aiming "diagnose and fix it" at the wrong graph.

Coalescing is now scoped to the **origin tab**: same tab, same pin, safe to merge; different tab, kept apart so each keeps its mid and the batch's mixed-origin fail-closed handling still runs. Deliberately **not** keyed on `mid` — `mintInjectionOrigin` mints a fresh UUID per event, so mid equality is never true and would have disabled coalescing while looking like a fix.

Round 2: **SHIP**, no remaining origin-loss path.

## Verification

- 3 tests on the real `injectRunError`; suite **9201** green
- **every prompt still arrives exactly once** — asserted alongside the no-growth check, because a swallowed error would be far worse than a noisy one
- mutation-checked in both directions: reverting to nesting fails the growth test; un-scoping the merge loses `mB` and fails the P0 test

The P0 test took three attempts, and the first two proved nothing — worth recording so they are not re-introduced:

1. asserted the two notices never share a turn — **not the invariant**; batching separate items into one turn is normal, and that version failed against the *correct* implementation
2. injected with a settle between them, so the first item had already dequeued and the merge path never ran — it passed under the mutation that un-scopes it
3. correct: assert on **origins** (`onSeen` fires per item with its mid), with the agent not started until both notices are queued together — the only state where the merge runs

## Still open on #1489

**Classification.** A prompt *we* cancelled should not arrive as a `run_error` at all. The panel already has the right pattern one branch over: `onReconcileInterrupted` sends a neutral `executed` note for a known cancellation, with a comment stating why it must not be `run_error`. `onReconcileGiveUp` — the branch these 26 prompts hit — sends `run_error` even though its own text says *"likely cancelled or interrupted. Safe to requeue."* That is a panel-repo change.

Also seen in the capture and worth fixing there: the injected notice claims *"You did NOT queue this run — this session has queued none"*, which is the opposite of the truth here — the session queued all 27 via `panel_run` and then cancelled them.

**Lock half:** not an oversight and not touched. The acquire path refuses deliberately (*a concurrent pre-upgrade orchestrator could replace an observed stale path with a fresh lock*), which is exactly why `panel_action:"unlock"` re-verifies before deleting. The defensible narrowing — failing fast when the owner pid is provably dead instead of waiting the full 60 s — is left out of this PR to keep it scoped.

### DIFF
--- a/src/__tests__/orchestrator/run-error-burst.test.ts
+++ b/src/__tests__/orchestrator/run-error-burst.test.ts
@@ -90,11 +90,11 @@ describe("a BURST of run_errors coalesces instead of nesting (#1489)", () => {
 
     // Three cleared prompts land in quick succession, as they do when a cancelled batch is
     // reconciled after a reconnect.
-    await agent.injectRunError("PROMPT_ONE could not be confirmed");
+    await agent.injectRunError("PROMPT_ONE could not be confirmed", { originTab: "tab1", mid: "m1" });
     await settle();
-    await agent.injectRunError("PROMPT_TWO could not be confirmed");
+    await agent.injectRunError("PROMPT_TWO could not be confirmed", { originTab: "tab1", mid: "m2" });
     await settle();
-    await agent.injectRunError("PROMPT_THREE could not be confirmed");
+    await agent.injectRunError("PROMPT_THREE could not be confirmed", { originTab: "tab1", mid: "m3" });
     await settle();
 
     // Drain: let each held turn finish so everything queued is actually delivered.
@@ -139,7 +139,7 @@ describe("a BURST of run_errors coalesces instead of nesting (#1489)", () => {
 
     agent.send("USER_PROMPT");
     await settle();
-    await agent.injectRunError("ONLY_PROMPT could not be confirmed");
+    await agent.injectRunError("ONLY_PROMPT could not be confirmed", { originTab: "tab1", mid: "m1" });
     await settle();
 
     // The error turn was dispatched ahead of the user's work rather than waiting for it.
@@ -150,5 +150,54 @@ describe("a BURST of run_errors coalesces instead of nesting (#1489)", () => {
     await settle();
     void running;
     await agent.stop().catch(() => {});
+  });
+
+  it("notices from DIFFERENT tabs are never merged (codex P0)", async () => {
+    // The origin `mid` pins the error-handling turn to the ERRORING workflow's tab —
+    // #884's P0, "a render error on A silently editing B". Coalescing keeps ONE mid per
+    // item, so folding tab B's notice into tab A's item would discard B's origin and aim
+    // "diagnose and fix it" at A's graph. That is the P0 reintroduced, and it is why
+    // coalescing is scoped to the origin tab rather than applied to any pending error.
+    const { backend, releaseNow } = holdingBackend();
+    // ASSERTED ON THE ORIGINS, not on the text. `onSeen` fires once per queue ITEM at
+    // dequeue, carrying that item's `mid` — and the mid is what pins the turn. Two items
+    // batching into one turn text is normal and harmless; losing a mid is the P0, because
+    // the batch's mixed-origin fail-closed handling can only run on origins it was given.
+    // An earlier version of this test asserted the two notices never share a turn, which
+    // is not the invariant and failed against the correct implementation.
+    const seenMids: string[] = [];
+    const agent = new PanelAgent(
+      "tab1",
+      {
+        mcpServers: undefined,
+        systemAppend: "",
+        model: "m",
+        onSay: () => {},
+        onSeen: (_tab, mid) => seenMids.push(mid),
+      },
+      backend,
+    );
+    // The agent is NOT started yet, deliberately. Both notices must be sitting in the
+    // QUEUE together, which is the only state where the merge path runs at all — with the
+    // agent draining, the first is dequeued before the second arrives and B takes the
+    // in-flight branch instead. Two earlier versions of this test did exactly that and
+    // passed under a mutation that un-scoped the merge, i.e. they proved nothing.
+    await agent.injectRunError("FROM_TAB_A could not be confirmed", { originTab: "tabA", mid: "mA" });
+    await agent.injectRunError("FROM_TAB_B could not be confirmed", { originTab: "tabB", mid: "mB" });
+
+    const running = agent.start();
+    for (let i = 0; i < 6; i++) {
+      await settle();
+      releaseNow();
+    }
+    await settle();
+    void running;
+    await agent.stop().catch(() => {});
+
+    // BOTH origins reach the dequeue. If B had been folded into A's item, only mA would
+    // fire and the whole turn would be pinned to tab A — "diagnose and fix it" aimed at
+    // the wrong graph, which is #884's P0.
+    expect(seenMids, `origins seen: ${seenMids.join(",")}`).toContain("mA");
+    expect(seenMids, `origins seen: ${seenMids.join(",")}`).toContain("mB");
   });
 });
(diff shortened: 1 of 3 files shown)


artokun:
## Located both sides — and the panel already contains the pattern this needs

The message is panel-side (`comfyui-mcp-panel/web/js/comfyui-mcp-panel.js`), and the two reconcile paths sitting next to each other are the whole story:

**`onReconcileInterrupted`** — a run KNOWN to have been cancelled — deliberately sends a neutral `executed` event:

> The workflow run you queued was cancelled while the connection was down (prompt …). It did not crash; no action is required unless the cancellation was unintended.

with a comment stating exactly why it is not a `run_error`: *"the orchestrator treats run_error as urgent and tells the agent to diagnose/panel_get_errors."* That is precisely the harm in this report, already understood and already avoided — one branch over.

**`onReconcileGiveUp`** — the tracker exhausted its retries and the prompt's `/history` never appeared — sends `run_error` with the reported text. That is the branch the reporter hit, and for a genuinely unknown outcome the urgency is arguably right.

**Why the cleared prompts land in the wrong one:** `queue(action:"cancel", clear_pending:true)` is a headless MCP tool. ComfyUI writes no history for a prompt cleared while pending, and the panel's tracker — a separate observer — never learns the cancellation was deliberate. So 26 prompts we cancelled on purpose fall through "unknown after retries" and each becomes an urgent error.

So the classification is not simply wrong; the panel cannot tell. **The side that knows is the one that issued the cancel.**

## Where I'll take it

1. **The orchestrator knows the ids it cancelled.** A `run_error` arriving for a prompt this process deliberately cancelled should be dropped, or downgraded to the same neutral wording `onReconcileInterrupted` already uses. Reusing that text keeps one voice for one situation instead of inventing a second.
2. **The cumulative re-send is the multiplier and is separate.** One wrong line per prompt is 26 lines; re-sending 1..N every turn is what consumed the context window. That is worth fixing even for events that are entirely legitimate, and I want to understand why the block accumulates before touching the classification — a fix to (1) alone would still leave a correct-but-repeating flood.

Measuring the accumulation next. Not writing code until I can point at the code path that re-sends, because "not deduplicated" is a symptom and I do not yet know whether the cause is retention, re-delivery of undelivered frames (`markUndelivered` / `markDeliveryUnconfirmed` both appear on these paths), or turn assembly.

artokun:
## The accumulation mechanism — it is not a missing dedupe

`injectRunError` (panel-agent.ts) does two things per error:

```ts
if (this.inFlight) {
  // Stop the live turn and re-queue it so the agent handles the error FIRST,
  // then resumes whatever it was doing.
  await this.interrupt({ requeueInFlight: true });
}
…
this.queue.unshift({ text, … }); // front: ahead of any re-queued interrupted turn
```

Both halves are reasonable in isolation, and together, under N rapid-fire errors, they compound:

1. error 1 arrives → interrupts the live turn, **re-queues its text**, unshifts error 1
2. the agent begins the error-1 turn → error 2 arrives → interrupts **that** turn and re-queues it — and that turn's text *is* error 1 — then unshifts error 2
3. …repeat 26 times

Each interrupt preserves the interrupted turn by re-queueing it, and after the first error the interrupted turn is itself an error-handling turn. That is exactly the reported shape — "turn N carried prompts 1..N, turn N+1 carried 1..N+1" — and it explains the detail a dedupe theory does not: **the user's older prompt text kept reappearing too**, because the original user turn is the bottom of that same re-queue stack.

So "the events are not deduplicated" is the symptom. The cause is requeue-on-interrupt meeting a burst, and a dedupe alone would not fix it — 26 *distinct* prompt ids are 26 distinct texts, so nothing would dedupe.

Worth noting what the code already gets right, one branch over: `executed` completions go through `RunCompletions.record`, which correlates by exact prompt id and explicitly suppresses a re-sent completion ("`null` = the panel re-sent a completion this tab was already given; suppressed, never duplicated"). `run_error` has no such journal — it is front-queued straight into the turn stack.

## What this makes the fix

Two independent changes, and the first is the one that matters:

1. **Do not let a burst of run_errors nest.** An error arriving while an error-handling turn is already queued or in flight should join that turn rather than interrupt it and stack another copy. Coalescing at the injection point removes the growth regardless of how many prompts failed.
2. **A prompt this orchestrator cancelled should not become a `run_error` at all** — the classification half from my earlier comment.

(1) without (2) leaves one calm-but-wrong error; (2) without (1) leaves the flood for any genuine multi-prompt failure. Both are wanted.

Next: reproduce the growth against the real `injectRunError` with a fake backend — N injections, assert the queue grows cumulatively — so the fix is measured against a demonstrated failure rather than against my reading of it.

artokun:
## Reproduced — the growth is measured, not inferred

Drove the **real** `injectRunError` against a PanelAgent with a backend that holds its turn (so the agent is genuinely in flight when the next error lands, which is the precondition for the requeue path). One user turn, then three cancelled prompts in quick succession:

| turn | length | carries |
|---|---|---|
| 0 | 11 | `USER_PROMPT` |
| 1 | 419 | PROMPT_ONE |
| 2 | 827 | PROMPT_ONE + PROMPT_TWO |
| 3 | 1237 | PROMPT_ONE + PROMPT_TWO + PROMPT_THREE **+ USER_PROMPT** |

`lastCarriesAllThree: true`, `lastCarriesUserPrompt: true`.

That is the reporter's shape exactly — "turn N carried prompts 1..N" — including the detail that made me confident it was the requeue and not a dedupe gap: **their older prompt text kept reappearing**, because the original user turn is the bottom of the same re-queue stack. Growth is ~410 chars per error, cumulative; at 26 prompts that is the flood they describe, repeated every turn.

So the mechanism is confirmed by measurement rather than by reading, and the "not deduplicated" framing is ruled out — three *distinct* texts, nothing to dedupe.

Also visible in the captured text, and worth fixing while here:

> You did **NOT** queue this run — this session has queued none — so do not treat it as your own failure

That is #889's guard firing, and in this scenario it is **wrong in a new way**: the session *did* queue those 27 runs via `panel_run`, and it also cancelled them. The notice tells the agent the opposite of what happened because the correlation is keyed on something that did not survive the batch.

Implementing now against this reproduction: coalesce a burst instead of nesting it, and stop classifying our own cancellations as errors.

Issue #1478, opened 2026-08-12
panel_load_workflow leaves a stale instance fence; read-only panel_get_errors refused by the dirty-mutation guard


artokun:
Two separate workflow-binding/fence defects hit back-to-back in one session while loading a bundled pack onto the live canvas. Neither was fatal, but both forced manual recovery steps / a headless workaround.

### 1. panel_load_workflow succeeds but does not refresh the workflow-instance fence

`panel_load_workflow({pack:"ltx23-distill-3stage"})` returned `{loaded:true, format:"api", node_count:59}`. The very next graph call — `panel_graph_outline` with no arguments — failed:

```
Error: workflow instance mismatch: this command was issued for workflow instance
e592452b-c172-416d-a8bc-0ec6b96b56e1, and the active canvas reports
632e8dd7-30df-4de9-b28c-b6c98b9532aa. Nothing was applied.
... The active workflow last moved to tmp:c2fceb3a-... (last seen as
wf:workflows/Untitled ....json), and NO PANEL COMMAND CLAIMED IT.
```

Loading a workflow replaces the graph and therefore mints a new canvas instance id, so the fence held by the session is guaranteed stale immediately after a successful load. The recovery is always the same (`panel_set_workflow_target({mode:"current"})`), which suggests the load path should refresh/claim the fence itself.

Notably the error text says "NO PANEL COMMAND CLAIMED IT" even though the instance change was caused by a panel command (`panel_load_workflow`) one call earlier — so the message actively misdirects toward "the user switched tabs".

Reproduced deterministically twice in the same session (once on the initial load, once again later in the same flow).

**Steps to reproduce**
1. Open a fresh/empty workflow tab.
2. `panel_load_workflow({pack:"ltx23-distill-3stage"})` → `loaded:true`.
3. `panel_graph_outline({})` → workflow instance mismatch, nothing applied.
4. `panel_set_workflow_target({mode:"current"})` → `graph_binding_status:"refreshed"`; retry now works.

**Suggested change**: have `panel_load_workflow` (and any other tool that replaces the graph wholesale) re-seal/claim the fence onto the new instance id it just created, and register the claim so the mismatch message doesn't report "NO PANEL COMMAND CLAIMED IT" for a change a panel command caused.

### 2. Read-only panel_get_errors refused by the [dirty-mutation-binding-unproven] guard

Immediately after the load (tab dirty, unsaved), `panel_get_errors()` — a documented **read-only** tool — was refused:

```
[...]

Issue #1071, opened 2026-08-08
Panel graph binding unrecoverable after interrupted panel_open_workflow — set_workflow_target({mode:"current"}) is a permanent no-op


artokun:
### What happened / root cause

After a turn was interrupted immediately following a successful `panel_open_workflow`, this session's graph-command fence became permanently unrecoverable. Every `panel_*` call now fails and the documented recovery path never succeeds.

Sequence:

1. `panel_open_workflow("idlerpg-anima-character-vfx.json")` succeeded, returning `workflow_uuid: 601b9aa8-4e38-48a9-82ae-085c32928a25`.
2. The user interrupted the turn right after that call returned.
3. Next turn: `panel_new_workflow` returned `created: true` **and simultaneously** reported it could not read the live canvas identity, leaving the session fenced to `601b9aa8`. So a new tab now exists on screen that the session can never address.
4. Every subsequent `panel_graph_outline`, `panel_list_workflows`, `panel_add_node` fails with `workflow instance mismatch`.
5. `panel_set_workflow_target({mode:"current"})` — the documented recovery — was called **5 times over ~20 minutes** and failed identically every time, always landing in the `unreadable` branch: "the workflow target is now mode:'current'" (applied) but "Could NOT read the live canvas identity — the panel did not answer".

Root cause is in the `unreadable` case of the recovery handler in `dist/orchestrator/panel-tools.js` (~line 2494). The diagnostics there are excellent and correctly refuse to fold "unobserved" into "absent" — the defect is not the messaging, it's that **there is no recovery path at all when the panel stops answering the identity read**. `mode:"current"` can only adopt an identity it can read; if that read is what's broken, the tool is structurally incapable of recovering and will loop forever. The orchestrator holds the stale fence, and per the pack's own docs the orchestrator process only reloads on a full process restart — so nothing agent-side can clear it.

Note the state is self-inconsistent: `panel_new_workflow` **did** mutate the panel (a tab was created) in the same call where the identity read failed. So the command channel was alive while the identity channel was not — this looks like a targeted failure of the identity read path, not a dead tab.

### Steps to reproduce

1. `panel_open_workflow(<any workflow>)` and let it return successfully.
2. Interrupt the turn immediately (ESC) before any further panel call.
3. Next turn, call any `panel_*` graph tool → `workflow instance mismatch`.
4. Call `panel_set_workflow_target({mode:"current"})` repeatedly → always `unreadable` / `not_recovered`. Never recovers.

[...]

Issue #814, opened 2026-08-04
panel_restart_comfyui kills a Comfy Desktop-spawned server it cannot relaunch (self-reported 'cannot confirm it came back'), leaving the user with no ComfyUI; MCP also silently binds to a stale duplicate instance


artokun:
### What happened / root cause

Two defects compounded, both around instance identity/ownership. Likely adjacent to the in-flight #776 launcher-evidence work.

**(1) Restart killed a server it could not relaunch — and knew it.**

The user's machine had TWO ComfyUI servers off the same install dir (`~/ComfyUI-Installs/ComfyUI/ComfyUI`), both spawned by Comfy Desktop:
- `:8188` — older instance, ComfyUI **0.29.2** (started while the checkout was at the v0.29.2 tag)
- `:8189` — newer instance, ComfyUI **0.30.1** (started 11:41 PM after `git checkout v0.30.1`)

The MCP was bound to `:8188`. I called `panel_restart_comfyui` to pick up the updated on-disk code. It returned:

```
{"rebooting": true, "ready": false, "confirmed_cycle": false, "dispatched": true,
 "note": "ComfyUI restart was dispatched and accepted; it is restarting out-of-band.
  There is no local boot endpoint I can safely probe from here, so I can't confirm
  it finished coming back ..."}
```

The Manager reboot **stopped** the process, and nothing ever brought it back — `:8188` stayed dead indefinitely (verified: `netstat` showed nothing on 8188; only the `:8189` pair of `python.exe` remained). The user's open browser tab + Agent panel were left pointing at a dead server.

This is the exact scenario the tool's own contract says it guards: *"On an externally-managed install whose relaunch can't be proven from here (e.g. Pinokio), the restart is REFUSED **before anything is stopped**."* Here it was NOT refused — it stopped the server and only afterwards reported that it couldn't confirm relaunch. **A `confirmed_cycle:false` / "no boot endpoint I can probe" verdict must be computed BEFORE the stop, and gate it — not narrated after the process is already gone.** A Comfy Desktop-spawned child does not self-relaunch after a Manager reboot when the Desktop supervisor has already moved on to a different port.

Recovery required manually reading the dead process's argv out of the process table and re-spawning it by hand:
[...]

PR #1510, opened 2026-08-13
fix(1478): name the load as the cause of a workflow-instance mismatch

artokun:
Refs #1478 — fixes the **misdirection** in defect 1. Does not close the issue: the categorical repair needs a panel change, and defect 2 is panel-side.

## The reported problem

`panel_load_workflow({pack:…})` returns `{loaded:true, format:"api", node_count:59}`, and the very next `panel_graph_outline({})` fails:

```
workflow instance mismatch: … The active workflow last moved to tmp:… and
NO PANEL COMMAND CLAIMED IT.
```

A panel command *did* cause it, one call earlier. So the refusal points the reader at "the user switched tabs" — the reporter hit this deterministically twice and had to work out the real cause themselves.

## What ships

One sentence on the load result, and no round trip:

> an API-format load CAN re-mint the canvas workflow instance. If your next graph command is refused with "workflow instance mismatch", that is why — the cause is this load, NOT the user switching tabs. Clear it with `panel_set_workflow_target({mode:"current"})` … If the next command is not refused, nothing needs doing.

## Four review rounds, four false claims of mine

This PR is much smaller than what I opened it with, and every reduction was a review finding:

1. **Auto-repairing the fence was a wrong-graph write.** Running the same re-derivation the documented recovery runs adopts *whatever is active now*, with no tie to the load — a user switching canvases in the window would stamp the session to a different workflow and the next edit would land on **B** though the load modified **A**. `panel_set_workflow_target` is sound doing exactly this only because the user explicitly asked to follow what is live.
2. **Asserting staleness without confirming the load landed.** The note was keyed on "the call did not error", so a `loaded:false` envelope would have been told its fence went stale.
3. **A UI-format load PRESERVES the instance.** It passes `__cmcpKeepInstance: true` whenever a workflow is active, deliberately, so the agent's own follow-up commands are not rejected — a blanket note would have been false for the commonest load.
4. **A second API load into the same active workflow reuses the object**, whose uuid is object-keyed, so the fence stays valid there too.

After (3) and (4), every categorical version of "your fence is now stale" had been falsified by a different real path — and the reply carries nothing that separates *re-minted* from *reused*. So the claim became conditional, which is true everywhere, and the sentence stopped pretending to knowledge the data does not contain.

Also dropped along the way: an added `workflow_list` round trip that, on a panel which cannot corroborate, spent up to four 6-second probes reaching an answer it must not use.

## Verification

- 5 tests on the real tool path; suite **9206** green; `lint`, `check:vocabulary`, `check:unknown-collapse`, `vocab:export --check`, `asset-counts --check` all pass
- pins **both** directions: the note appears for an API load, and is **absent** for a UI-format load and for a `loaded:false` envelope
- pins the P1 directly: with the user on a **different** canvas, the session stamp must not move
- mutation-checked — removing the note, hardcoding the payload gate, and dropping the format gate each fail a different test
- two pre-existing assertions counted *total* bridge calls as a proxy for "the graph was loaded once"; they now count `graph_load` specifically, which is the guarantee they were guarding

## Still open on #1478

- **The categorical fix is panel-side**: put `workflow_uuid` on `graph_load`'s reply, as #762/#800 did for `workflow_new`/`workflow_save`. The load could then compare it against the fence and say exactly what happened — or claim the identity it created, which is race-proof and refusal-proof, and is why those two work that way.
- **Defect 2** (read-only `panel_get_errors` refused by the dirty-**mutation** guard, whose text even says *"this mutation was NOT applied"*) is a panel-side classification bug.

### DIFF
--- a/src/__tests__/orchestrator/load-claims-instance.test.ts
+++ b/src/__tests__/orchestrator/load-claims-instance.test.ts
@@ -100,10 +100,17 @@ describe("a load discloses the fence it just invalidated (#1478)", () => {
 
     expect(out.isError).toBe(false);
     expect(out.text).toMatch(/loaded/i);
-    expect(out.text).toMatch(/instance fence now names the OLD one/);
+    expect(out.text).toMatch(/CAN re-mint the canvas workflow instance/);
     expect(out.text).toMatch(/panel_set_workflow_target/);
     // The specific misdirection this replaces: the downstream refusal blames a tab switch.
-    expect(out.text).toMatch(/not by the user switching tabs/);
+    expect(out.text).toMatch(/NOT the user switching tabs/);
+    // CONDITIONAL, never categorical. Four review rounds falsified every version that
+    // asserted the fence IS stale: a UI load preserves the instance outright, and a second
+    // API load into the same already-active workflow reuses the object so its uuid does
+    // not change either. The reply cannot tell those apart, so the note must not pretend to.
+    expect(out.text).toMatch(/If your next graph command is refused/);
+    expect(out.text).toMatch(/If the next command is not refused, nothing needs doing/);
+    expect(out.text).not.toMatch(/will be refused/);
   });
 
   it("does NOT adopt whatever is active — the P0 an auto-rebind would have created", async () => {
@@ -132,7 +139,7 @@ describe("a load discloses the fence it just invalidated (#1478)", () => {
     // unmeasured-claim defect this note exists to remove.
     const out = await load({ loaded: false, error: "nothing was applied" });
 
-    expect(out.text).not.toMatch(/instance fence now names the OLD one/);
+    expect(out.text).not.toMatch(/CAN re-mint the canvas workflow instance/);
     expect(out.text).not.toMatch(/panel_set_workflow_target/);
   });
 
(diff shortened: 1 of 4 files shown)


artokun:
## Measured first, and it rules out the approach I sketched

`graph_load`'s reply carries **no `workflow_uuid`**. Both success shapes are:

```js
{ loaded: true, format: "api", node_count, entries_in, … }   // API-format branch
{ loaded: true, node_count, … }                              // UI-format branch
```

So #814's pattern — trust the command's **own reply**, because it already proves the new identity — is simply not available here. `workflow_new` and `workflow_save/save_as` can do that only because #762/#800 put `workflow_uuid` on *their* replies. Nothing did that for `graph_load`.

That leaves two routes:

1. **Panel-side**: add `workflow_uuid` to the `graph_load` reply, mirroring #762/#800, then one-line `refreshWorkflowUuid(ctx, reply)` here. Most consistent with the existing design, but cross-repo.
2. **Orchestrator-side**: after a successful load, run the generic re-derivation (`rebindWorkflowFence`) — a `workflow_list` round trip that reads the live active identity.

**Taking (2), and the reporter's own transcript is the evidence for it.** Their recovery is `panel_set_workflow_target({mode:"current"})`, which *is* `rebindWorkflowFence`, and it returned `graph_binding_status:"refreshed"` and worked — twice, deterministically. So the exact call I would make after the load is already demonstrated to repair this state on a real rig. Doing automatically what the user is currently told to do by hand.

One caveat I want to be honest about rather than skate past: #814's comment warns that the generic re-derivation "makes its OWN independent `workflow_list` round trip … refused by the exact fence this whole flow exists to repair (#1071)". That warning is about a different starting state — there the fence had no trustworthy identity at all. Here the load has just *minted* one and `workflow_list` is not a fenced command, which is why the manual recovery works today. I will assert that in a test rather than rely on the reasoning.

Next: wire the rebind into the load path, report its outcome in the result (a load that could not re-claim its own fence must say so rather than returning a bare `loaded:true`), and only then revisit the "NO PANEL COMMAND CLAIMED IT" wording — which stays accurate until the claim actually happens.

Issue #1468, opened 2026-08-12
panel_civitai_search and graph_exit_subgraph time out after commands are successfully applied


artokun:
### What happened / root cause
Panel mutation commands can be delivered and applied in the frontend but still return `isError:true` when the acknowledgement exceeds a fixed timeout. `panel_civitai_search` repeatedly timed out at 10 seconds while the search was actually applied (`renderRev` advanced and `panel_civitai_results` reported `loading:true`). `panel_exit_subgraph` similarly timed out at 15 seconds but a subsequent `panel_graph_outline` proved the view had returned to root. The command completion appears coupled to a slow frontend/external operation instead of acknowledging dispatch/application promptly.

### Steps to reproduce
1. Open the docked CivitAI browser.
2. Call `panel_civitai_search` while CivitAI is slow or has just returned 503.
3. The tool errors after 10 seconds and warns not to retry.
4. Call `panel_civitai_results`; `renderRev` has incremented and the new search is loading, proving the mutation applied.

Also: enter a subgraph, call `panel_exit_subgraph` on a busy tab, receive a 15-second timeout, then call `panel_graph_outline`; scope is already `root`.

### Exact errors (scrubbed)
`Panel tab <workflow>: did not reply to "civitai_search" within 10000 ms ... command MUTATES and was already delivered ... may have been applied despite the missing reply.`
`Panel tab <workflow>: did not reply to "graph_exit_subgraph" within 15000 ms ... may have been applied despite the missing reply.`

### Fix
Applied locally: no. Upstream-only: yes.
Acknowledge `civitai_search` once the frontend accepts and dispatches the search, returning `{dispatched:true, loading:true, renderRev}` without waiting for the external CivitAI request to settle; callers already have `panel_civitai_results` for completion. For navigation mutations, acknowledge immediately after the scope change is committed, or derive success from the updated scope before emitting timeout. More generally, distinguish `delivered/applied` from `result completed` so a slow result does not become a false mutation error.

### Environment
Windows 11; ComfyUI 0.31.1; ComfyUI frontend 1.48.7; Python 3.13.12; PyTorch 2.12.1+cu130; NVIDIA RTX 5080 16GB; comfyui-mcp 0.51.15; panel 0.11.44.

Related (same root cause): #282, #694

artokun:
Located both, and your diagnosis holds — these are tighter than the general bound, on purpose, and the coupling is the real problem.

**The two call sites:**

[...]

Issue #778, opened 2026-08-03
panel_find_nodes is misclassified as mutating and blocked on older panels


artokun:
### What happened / root cause
panel_find_nodes failed on an older or version-unknown panel with a workflow-stamp enforcement error even though graph_find_nodes is read-only. In comfyui-mcp 0.49.3, BRIDGE_READONLY_CMDS in dist/services/ui-bridge.js included graph_outline and graph_query but omitted graph_find_nodes. That omission made isMutatingGraphCommand classify graph_find_nodes as a canvas mutation and wrongly require panel 0.11.35+ workflow-stamp enforcement.

### Steps to reproduce
1. Connect comfyui-mcp 0.49.3 to a panel tab that does not advertise workflow-stamp enforcement.
2. Confirm panel_graph_outline and panel_query_graph work.
3. Call panel_find_nodes with any query.
4. The server refuses dispatch as if graph_find_nodes were mutating.

### Exact error
"graph_find_nodes" cannot be safely targeted to the active workflow: panel tab [REDACTED] does not enforce per-command workflow targeting (detected panel version unknown; this MCP requires panel 0.11.35+). Read-only graph commands (graph_outline, graph_query, graph_get_state) still work.

### Fix
applied locally: yes
upstream-only: no

Patch:
```diff
 export const BRIDGE_READONLY_CMDS = new Set([
     "graph_prompt_director_audit",
     "graph_query",
+    "graph_find_nodes",
     "civitai_results",
 ]);
```

### Verification
node --check dist/services/ui-bridge.js exited 0. Runtime verification was attempted after an orchestrator soft reload, but the panel socket had not reconnected, so the command could not be retried against a live tab in this turn.

### Environment
OS: ComfyUI target Linux; orchestrator host Windows
GPU: NVIDIA GeForce RTX 4090 24 GB
ComfyUI: 0.30.0 REMOTE
comfyui-mcp: 0.49.3
panel: version unknown

PR #1518, opened 2026-08-13
fix(1468): a timed-out ack stops being reported as a failed mutation

artokun:
Claims #1468.

`panel_civitai_search` and `panel_exit_subgraph` return `isError:true` after the command was applied. The reporter proved both: `renderRev` had advanced and the grid reported `loading:true`; the next `panel_graph_outline` showed scope `root`.

## The mechanism I first published on the issue was wrong

Corrected publicly [in the thread](https://github.com/artokun/comfyui-mcp/issues/1468#issuecomment-5278979532). Measured, not assumed:

- `driveSearch` does **not** await the CivitAI fetch — it fires `void reload(...)` and returns `{dispatched:true, renderRev}` with no `await` in the handler. That is panel **#282**, shipped at panel **0.11.0**; #1468 was filed from **0.11.44**, so the reporter already had it.
- `graph_exit_subgraph`'s own receipt (`confirmCanvasNavigation`) budgets **25 polls × 40 ms ≈ 1 s** and returns early on success — 15 s already clears its worst case by 15×.

Neither was coupled to a slow operation. Both are the **#357/#694 shape**: a busy-but-alive main thread missing a tighter-than-default bound. Which makes raising the bound — the fix I had argued *against* — the right one for the first.

## Changes

**1. `civitai_search`: private `10000` → `BRIDGE_DEFAULT_TIMEOUT_MS`.** Half this codebase's own default, fencing an external wait that no longer exists.

**2. `exit_subgraph`: settle a no-reply with one scope read.** Its effect is locally observable, so it asks instead of handing back "may have been applied" plus homework.

| Observation | Reported as |
|---|---|
| `scope:"root"` | decisive — the goal state holds |
| `scope:"subgraph"` | **not** decisive; stays a failure, names the open question |
| probe unanswerable | today's message, untouched (#1473's rule) |

The ambiguous case is ambiguous on purpose: exit pops to the **immediate parent** (#412), so "inside a subgraph" is equally consistent with "never landed" and "landed, from a nested subgraph". Resolving that by guessing would be the same harm as the false failure, pointed the other way.

Two overclaims caught while self-reviewing and fixed:

- the success payload says **`at_root`, not `exited`** — the read establishes where the canvas *is*, never that this command put it there;
- the ambiguous note originally said "do NOT retry blindly", which lands directly beneath the `retry_of` instruction `ctx.call` appends for every `RETRY_TOKEN_CMDS` timeout. Verified against the panel's ledger: a **token** retry is answered from the original's entry with no second executor run (#694); a **bare** repeat executes fresh. The note now draws that distinction instead of contradicting the line above it.

Only a no-reply takes this path. A relayed executor verdict already reasoned about the same uncertainty with the canvas in front of it.

## Verification

- Timeout strings **measured against a real `UiBridge` + frozen tab**, not transcribed from the template — `graph_exit_subgraph` does reach dispatch (it is fence-inert, #778) rather than being refused pre-dispatch, which is the trap that would have made the whole path dead code.
- The fixture drives the **real** `ctx.call` token-minting branch (tagged reply-timeout + dispatched rid), so the assertions run against the production message shape.
- 6 new tests. **5/5 mutations killed**, including deleting the call-site wiring while leaving the helper intact — the one that survives a helper-only suite.
- Full suite: **490 files, 9224 passed**. `lint`, `check:vocabulary`, `check:unknown-collapse`, `vocab:export --check`, `asset-counts --check`, `docs:gen` all clean.

## Found while measuring, not in scope here

`panel_civitai_results` is the command that genuinely awaits the fetch (`await state.activeReloadPromise`), on the same 10 s bound. It is a read, so its timeout is not the false *mutation* failure this issue is about — filed separately rather than folded in.

### DIFF
--- a/src/__tests__/orchestrator/ack-timeout-vs-applied.test.ts
+++ b/src/__tests__/orchestrator/ack-timeout-vs-applied.test.ts
@@ -1,0 +1,231 @@
+// #1468 — a command that was APPLIED came back as `isError:true` because its ack
+// missed a bound tighter than this codebase's own default. The reporter proved the
+// effect landed both times: `civitai_search` timed out at 10 s while `renderRev`
+// had advanced and the grid reported `loading:true`, and `graph_exit_subgraph`
+// timed out at 15 s while the very next `panel_graph_outline` showed scope `root`.
+//
+// The mechanism first published on that issue — "coupled to a slow external
+// operation" — is WRONG, and this file exists partly so nobody restores it:
+//
+//   - `driveSearch` does NOT await the CivitAI fetch. It fires `void reload(...)`
+//     and returns `{dispatched:true, renderRev}` with no await in the handler.
+//     That is panel #282, shipped at panel 0.11.0; #1468 was filed from 0.11.44.
+//   - `graph_exit_subgraph`'s own receipt (`confirmCanvasNavigation`) budgets
+//     25 polls x 40 ms ~= 1 s and returns early on success — 15 s already clears
+//     its worst case by 15x.
+//
+// With nothing external left to wait on, both are the #357/#694 shape: a
+// busy-but-alive main thread missing a tight bound. So `civitai_search` takes the
+// shared default, and `exit_subgraph` — whose effect is locally observable — asks
+// rather than guessing.
+//
+// The timeout strings asserted below were MEASURED against a real UiBridge and a
+// live-but-frozen tab, not written from the source template: `graph_exit_subgraph`
+// reaches dispatch (it is fence-inert, #778) and produces the canonical
+// `did not reply to "graph_exit_subgraph" within N ms` + the mutating-delivery
+// disclosure. A graph_* command that the #570 write gate refuses PRE-dispatch
+// would never take this path at all, and asserting against that refusal is how a
+// suite goes green while proving nothing.
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("../../comfyui/client.js", () => ({
+  getObjectInfo: vi.fn(),
+  backfillObjectInfo: vi.fn(),
+  resetClient: vi.fn(),
+  resetObjectInfoCache: vi.fn(),
+}));
+
+const { buildPanelToolDefs, makePanelToolCtx } = await import("../../orchestrator/panel-tools.js");
+import { BRIDGE_DEFAULT_TIMEOUT_MS, markReplyTimeout } from "../../services/ui-bridge.js";
+import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+import type { PanelToolCtx, ToolResult } from "../../orchestrator/panel-tools.js";
+
+const TAB = "11111111-2222-3333-4444-555555555555";
+
+const textOf = (res: ToolResult): string =>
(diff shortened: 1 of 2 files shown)


artokun:
**Scope note — `panel_enter_subgraph` is deliberately not included.**

It sits two lines away from `exit_subgraph`, carries the same 15 s bound, and goes through the same `confirmCanvasNavigation` receipt, so leaving it out is a choice rather than an oversight and is worth stating.

Two reasons:

1. The issue names `exit`, and the reporter's evidence is about `exit`. I have no measurement that `enter` misbehaves.
2. `enter` is actually the *easier* case to settle — its goal state is fully checkable, since `viewing.owner_node_id` should equal the `node_id` that was requested, whereas `exit`'s "inside a subgraph" reading is irreducibly ambiguous under #412. That makes it a clean follow-up, but it also means widening this PR would ship a second, differently-shaped verification on the back of one review.

If `enter` turns out to need it, the settle helper generalises to it directly — the only new part is comparing `owner_node_id` against the requested node rather than testing for root.

artokun:
**Codex gate: three rounds, all findings closed.**

| Round | Verdict | Finding | Status |
|---|---|---|---|
| 1 | NO-SHIP | P1 — settling read uncorrelated; `ctx.call` can silently rebind to another tab, so a *different* tab at root becomes success | fixed — probe pinned to the dispatch tab |
| 1 | NO-SHIP | P1 — loose text predicate could promote an acked error to success | see round 2 |
| 2 | NO-SHIP | rebind **closed**; the "canonical" predicate still cannot separate an acked error reproducing the sentence verbatim | fixed at the cause — see below |
| 3 | NO-SHIP | P1 — inner retry catch dropped the tag | fixed (found independently in the same audit) |

**Round 2 is the one that changed the design.** Codex rejected message-text matching twice and was right both times: acked panel errors arrive as arbitrary `msg.error` text and `ctx.call` flattens both kinds into the same text-only result, so *any sentence the bridge can write, a panel error can also contain*. No regex separates them.

The bridge already answers this exactly — `markReplyTimeout` / `isReplyTimeoutTagged`, on the error object, at the only place that knows. It was simply being dropped in translation. `ctx.call` now carries it onto the result as a non-enumerable symbol and the settle path keys on that; no text is parsed anywhere in the new path. `isAckTimeout` is byte-identical to main again.

The evidence that finding was real rather than theoretical: mutation **M8** re-implements the text predicate and is killed by a fixture carrying the canonical sentence verbatim.

**Round 3's finding, honestly scoped:** it was a completeness gap, not a correctness one — `graph_exit_subgraph` is not retry-safe so it cannot reach that branch, and an unmarked result fails closed. Fixed anyway, because it would have silently switched the settle off for a real sequence.

**One correction to my own earlier claim in this PR:** I said the symbol is invisible downstream. Measured: invisible to `JSON.stringify`, `Object.keys`, `for..in` and spread — but `Reflect.ownKeys`/`getOwnPropertySymbols` can see it. No consumer does that, and the wire payload is unchanged, but "non-enumerable" is not "unreachable".

Final: **490 files / 9227 tests**, `tsc` clean, all six repo gates clean, **8/8 mutations killed**.

Issue #1129, opened 2026-08-08
panel_install_node and install_custom_node do not provide a usable local fallback with legacy Manager 3.x


artokun:
### What happened / root cause
On a local Windows ComfyUI using legacy ComfyUI-Manager 3.x, panel_install_node failed with `ComfyUI-Manager not reachable`, and the documented fallback install_custom_node with `useCmCli:true` still fell through to the Manager v4 queue endpoint `/manager/queue/install`, which returned HTTP 404/security error. This leaves model/node repair workflows without the expected local git/comfy-cli fallback even though the ComfyUI filesystem is local.

### Steps to reproduce
1. Run local ComfyUI with Manager 3.x legacy.
2. Call panel_install_node with a valid GitHub repository URL.
3. Observe `ComfyUI-Manager not reachable`.
4. Call install_custom_node action=install, source=git, useCmCli=true for the same URL.
5. Observe fallback to `/manager/queue/install` and HTTP 404.

### Exact error (scrubbed)
`Error: ComfyUI-Manager not reachable (is the built-in Manager enabled?)`

`NODE_MANAGEMENT_ERROR: ComfyUI-Manager API 404 Not Found for /manager/queue/install ... legacy ComfyUI-Manager 3.x ... A security error has occurred.`

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: when the target is local and a git URL is supplied, panel_install_node/install_custom_node should fall back to a verified direct git clone plus dependency installation (or clearly invoke and report comfy-cli failure) instead of routing to the Manager v4 queue endpoint. The result should verify the installed directory is populated before success.

### Environment
Windows 11; ComfyUI 0.27.0 LOCAL; NVIDIA RTX 5070 12 GB; comfyui-mcp 0.49.3; panel 0.11.28; Manager 3.x LEGACY; Python 3.12; torch 2.8.0; CUDA 12.9.

artokun:
Fixed in #1143, shipping in **0.50.40**.

The direct-clone fallback you asked for already existed — twenty lines below the enqueue. It was unreachable. The 3.x git-URL install is gated server-side by `security_level` + `allow_git_url_install`, that gate answers on the POST, and the POST throws, so the code never reached the fallback in the one case it exists for. The source even names the gate in a comment: *"a rejection surfaces as a 403/404 from the queue route"* — it knew the rejection shape and did not act on it.

[...]

PR #1531, opened 2026-08-13
fix(1129): a Manager install that is accepted but never enqueued no longer reports success

artokun:
Claims #1129 (reopened).

#1143 fixed the **pre-queue refusal** — legacy Manager 3.x rejects a git-URL install with 403/404, which now falls through to a direct clone. This is the other half: the POST **succeeds**, returns `queued: true`, and the task is never enqueued. The reporter's follow-up read showed an idle queue, `total_count: 0`, and no directory under `custom_nodes`.

## What the report asks for already exists — in the other tool

`install_custom_node` already enqueues, verifies, and clones when nothing landed. The reporter used **`panel_install_node`**, which forwards `nodes_install` over the bridge and never reaches that code. Their reply shape (`queued/pending/dialect`) is the panel's, not that function's.

I deliberately did **not** copy the clone across. `install_custom_node` may clone because it owns the filesystem it writes to; `panel_install_node` drives whatever ComfyUI the panel is bound to, which need not be this machine. So it reports what it saw and names the tool that *can* clone.

## It warns; it does not declare failure

The first version also read the installed list and, finding the pack absent, flipped the result to an error. Review killed that on three counts, all correct:

- it substring-searched **raw reply text** instead of a validated shape, so an error envelope or unfamiliar dialect read as "absent"
- the fixture it was tested against (`{nodes:[{name}]}`) **is not the panel's shape** (`{installed: ...}`) — the check only ever passed *because* it searched text
- a git URL's checkout directory need not match its URL basename, and the panel documents those targets as rename-prone, so **absence by derived name cannot prove failure**

With no decisive second observation, a definite verdict is unjustified. A false "definitely failed" on a working install is the mirror of the false success this issue is about.

The message also stopped asserting *"this is not a task that already ran"* and *"retrying will be dropped"*. A queue **reset** clears these same counters — and this codebase issues one from `manager-config.ts` and `workflow-deps.ts` — so an install that really ran can read this way. Severity is not accuracy: it now states what was observed, names the reset as a real alternative, and says not to assume in either direction.

## Guards

| Guard | Why |
|---|---|
| queue must be **idle** | a mid-accept snapshot reads zero before counters move |
| counts **present** and zero | absent ≠ zero; matches the repo's legacy-empty proof |
| `pending_count` must agree | if the build reports one |
| **dispatch-tab pin** | `ctx.call` can rebind; another ComfyUI's empty queue is not evidence about this install |
| probe only when queued was claimed | one conditional read, no cost on other paths |

The dispatch-tab pin is the same guard #1468 needed **earlier today**. I built it there and did not carry it to this probe; that is why it is now asserted by a test that checks the rebind actually occurred.

## Verification

- **10 tests, 9/9 mutations killed.** One mutation *survived* first and was the most useful result: it exposed a **vacuous test** — the v4 fixture omitted `done_count`, so the presence check rejected it before `total_count` was consulted. Fixture fixed, not the mutation.
- Mutation M2 restores the ordering bug I originally shipped, where the panel's appended note broke JSON parsing and the entire check ran dead with every test green.
- Suite **492 files / 9261**; lint, vocabulary, unknown-collapse, vocab-export clean.

## Not verified

I do not have legacy Manager 3.x. The dropped-enqueue behaviour is modelled from the reporter's reply shapes, not reproduced against a real 3.x install.

### DIFF
--- a/src/__tests__/orchestrator/install-accepted-never-enqueued.test.ts
+++ b/src/__tests__/orchestrator/install-accepted-never-enqueued.test.ts
@@ -106,7 +106,7 @@ describe("an install the Manager accepted and dropped says so (#1129)", () => {
     // this way — the message must say so rather than assert the task never ran.
     expect(out.text).toMatch(/total_count, done, in_progress all 0/);
     expect(out.text).toMatch(/not a receipt/);
-    expect(out.text).toMatch(/it is NOT proof/);
+    expect(out.text).toMatch(/NOT PROOF EITHER WAY/);
     expect(out.text).toMatch(/cleared by a queue RESET/);
     expect(out.text).toMatch(/do not reinstall on\s+the assumption it did not/);
     // The earlier wording asserted both of these outright; neither is supportable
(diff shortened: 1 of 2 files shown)


artokun:
Merge attempts are currently failing with transient GitHub 5xx/GraphQL errors (refs E044:2C41E4, EA82:9EF06, E09D:13CADB). PR is CLEAN with all checks green; retrying.

Issue #1519, opened 2026-08-13
Live graph read rejects resumed workflow with missing instance stamp


artokun:
### What happened / root cause
The first live-canvas read on the active workflow failed with a workflow-instance fence mismatch immediately after the panel reported that the user switched to workflow "Untitled 2026-08-13 18-43-16". The command was rejected because it carried no workflow-instance stamp while the active canvas reported instance 2b3f4684-b7e7-495b-a7a1-f40439e35e0a. Rebinding with panel_set_workflow_target({mode:"current"}) repaired the fence and panel_graph_outline then worked.

### Steps to reproduce
1. Have the sidebar session resume or switch to a different open workflow.
2. Call panel_graph_outline on the newly active workflow.
3. Observe the instance mismatch; call panel_set_workflow_target({mode:"current"}) and retry.

### Exact error (scrubbed)
Error: workflow instance mismatch: this command carries no workflow-instance stamp, and the active canvas reports 2b3f4684-b7e7-495b-a7a1-f40439e35e0a. Nothing was applied.

### Fix
applied locally: no; upstream-only: no
Workaround verified: automatically rebind with panel_set_workflow_target({mode:"current"}) before retrying the graph read.

### Environment
OS: Windows 11
ComfyUI: 0.31.1
ComfyUI FRONTEND: 1.48.7
GPU: NVIDIA GeForce RTX 5090, 32 GB VRAM
comfyui-mcp: 0.51.29
comfyui-mcp-panel: 0.13.0

Related (same root cause): #1478

artokun:
Audit on current main (v0.52.1): the misleading messaging is fixed (#1537, v0.51.51), the sticky failure is fixed (panel #1279, v0.14.44 — drifted fence re-hellos BEFORE the next call), and the stale-stamp dispatch window is closed (#1682, v0.52.1). What remains is the deliberate design question already recorded in ui-bridge.ts: whether an unstamped read should be admitted — the three options and their trade-offs are documented there, and auto-adopt was removed from #1478 for cause after adversarial review. Leaving this open as the design-question tracker; nothing to patch without a decision on that question.

artokun:
Claimed by the backlog-drain session (worktree `.claude/worktrees/drain-1519`). No prior open claim or PR. A fix agent is assigned now; if no PR appears within ~8h, consider the claim lapsed.

artokun:
PR #1736 is open against the part of this that is not the design question.

[...]

PR #1537, opened 2026-08-13
fix(1519): a write refusal stops promising that graph reads still work

artokun:
Refs #1519.

After a workflow switch the first live-canvas read fails: *"this command carries no workflow-instance stamp, and the active canvas reports `2b3f…`"*. `panel_set_workflow_target({mode:"current"})` repairs it and the read then works.

## What this changes

The report's verified workaround is to **auto-rebind before retrying** — the auto-adopt hazard adversarial review removed from #1478 for cause, since adopting *"whatever is active now"* can silently bind the session to a **different** workflow. This PR does not implement it, and **no fence moves here at all**.

What it fixes is a false claim that sent the reporter looking in the wrong place. The bridge refuses a **mutation** when the session has no trusted workflow identity, and that refusal ended with:

> Reads and view-only commands still work (graph_outline, graph_query, graph_get_state, …)

A panel advertising the per-command fence refuses an **unstamped** command outright, reads included — #718 closed exactly that fail-open hole. Verified in the panel source, not inferred: `commandWorkflowMismatch` counts a missing stamp as a mismatch, and `activeWorkflowFenceApplies` returns true for every `graph_*` read.

## Review caught a P1 in the first fix, and it was right

The first version asked *"is there a trusted stamp?"* — **presence**, when the question is **agreement**. A read succeeds when the stamp still *equals* the active canvas, and the resolver deliberately answers with the workflow the turn was **issued for** (#570/#884), which is allowed to go stale — declining a command aimed at the workflow the user has since left is the fence's whole purpose. So the promise had merely moved from the no-stamp branch to the stale-stamp branch: the same defect, one branch over.

**Measured on a real `UiBridge`** before changing anything — panel advertising the dispatch-time fence but not the at-write recheck (so the write gate fires with a stamp present), session fenced to `aaaa…`, canvas showing `bbbb…`:

| | result |
|---|---|
| mutation refusal | ends *"Reads and view-only commands still work (graph_outline, …)"* |
| next `graph_outline` | **REJECTED** — *workflow instance mismatch: this command was issued for workflow instance `aaaa…`, and the active canvas reports `bbbb…`* |
| control, stamp = `bbbb…` | **SERVED** |

## Three verdicts, from what this side can establish

| state | reads | message |
|---|---|---|
| no fence advertised | execute — nothing compares them | unchanged sentence |
| fence, **no stamp** | refused, provably (a missing stamp *is* a mismatch) | withdrawn, and says why |
| fence, **stamp present** | depends on stamp-vs-active equality | **says UNKNOWN** |

The third is the only honest answer available here, not a hedge: `Conn` carries no live canvas uuid, and `workflowUuidFor()` reads the **same resolver the stamp came from** — comparing them would compare a value with itself and always agree. Only the panel can evaluate the equality, and a pre-dispatch refusal must not take a round trip to ask. So the message names the stamp the read will carry and states the condition instead of asserting an outcome it never observed.

It names **no remedy**: `panel-tools`' #1331 handler measures which no-identity state this is before naming one, and can conclude `panel_set_workflow_target({mode:"current"})` will *not* clear it. It offers the fence-exempt probe (`panel_list_workflows`) as something to *try*, with the version caveat measured against the panel repo — the exemption first shipped in panel 0.11.83 while this branch is reachable from 0.11.35, and #1519's reporter is on 0.13.0, whose release commit has it.

## The test models the equality, not the wording

Asserting only wording is how the P1 got through. The fixture panel now applies the panel's own transcribed rule (`activeWorkflowFenceApplies` + `commandWorkflowMismatch`), every fenced case asserts the **real read outcome**, and each carries a matching-stamp **control** so a green assertion cannot come from a fixture that refuses everything. Three mutations were verified to fail it: presence-based (the shipped bug), always-refused (the mirror image), and withdrawing the claim from an old panel.

## Nothing widens

- No fence is adopted, re-derived, or repaired; no command changes route.
- A mutating command with no stamp is still refused **before anything is written** — asserted from the socket across six mutators.
- A **stale-stamped** mutation is still dispatched and refused *by the panel*, with nothing applied — the fence working as designed.
- An unstamped read is still dispatched, still unstamped, for the panel to judge — the orchestrator has *not* started refusing reads locally.

## Still open

Whether an unstamped read should be admitted at all is a design question this PR does not settle — weakening #718, adopting the active workflow automatically (removed from #1478 for cause), or refusing locally all trade different risks. It stays recorded in the comment at the stamping site rather than silently picked.

Related: #1478 (same family, where auto-adopt was removed), #1331, #718.


artokun:
**Traced it. This is a contract mismatch between the orchestrator and the panel, not a missing refresh** — and the fix needs a decision I should not make unilaterally.

## The two halves disagree, in writing

**Orchestrator**, `ui-bridge.ts` (#570), explaining why it strips the stamp when it has no trusted identity:

> an identity-less tab / old panel ships unstamped (mutations are already refused by the `requiresWorkflowStampEnforcement` gate above; **reads execute**, reply still server-fenced)

**Panel**, `comfyui-mcp-panel.js` (#718), in the refusal itself:

> An UNSTAMPED command is refused **too** (#718 — an advertised fence must not fail open)

So the orchestrator dispatches an unstamped read believing it will execute, and the panel refuses it by design. Both behaviours are deliberate, documented, and incompatible. The reporter's `panel_graph_outline` is exactly that case: a read, unstamped, refused.

That also explains why `refreshWorkflowUuid` being wired into several paths did not help — nothing failed to reach it. The session legitimately had no trusted identity yet, and the orchestrator's own design says that is fine for a read.

## Why I am not just picking one

- **Making the panel admit unstamped reads** re-opens part of what #718 closed. Its reasoning — an advertised fence must not fail open — is sound, and a read still reveals graph contents.
- **Making the orchestrator establish a fence first** is the auto-adopt hazard removed from #1478 hours ago: adopting "whatever is active now" can silently bind the session to a different workflow, which is what #1111/#1089 were.
- **A third option** — orchestrator refuses to send an unstamped read at all and returns the actionable message itself — avoids both, at the cost of a round trip and a behaviour change to reads.

Each of those is defensible; they trade different risks, and one of them touches a guard added deliberately for a security-adjacent reason. That is an owner's call.

## One thing that is unambiguously wrong either way

The orchestrator's comment claims reads execute unstamped. **They do not.** Whatever is decided above, that sentence is false today and will mislead the next person exactly as it misled me — I read it as the design contract and had to go read the panel to find out otherwise. I will correct it regardless.

## Recommendation

Third option, with reads refused locally and told *why*, so the caller gets a message naming `panel_set_workflow_target({mode:"current"})` without a wasted round trip and without either side's guard weakening. But I would rather have that confirmed than assume it — the last time I chose the automatic-adoption route on this family, review removed it for cause.

Issue #1529, opened 2026-08-13
panel_set_widget is rejected during backend reconnect with no safe retry path


artokun:
### What happened / root cause
`panel_set_widget` is rejected whenever the backend bridge is reconnecting. The call correctly avoids mutating a potentially replaced canvas, but leaves no structured retry/replay route; two manual attempts returned the same rejection and nothing was applied.

### Steps to reproduce
1. Trigger or recover a ComfyUI backend reconnect while a live panel canvas is open.
2. Call `panel_set_widget` for node 605, widget `video`, value `pritty_013.mp4` before the bridge reports stable.
3. The mutation is rejected and must be retried manually after an indeterminate delay.

### Exact error (scrubbed)
[backend-reconnecting] ComfyUI backend connection is down, so `graph_set_widget` was not applied. The message advises retrying after reconnect, but provides no stable readiness signal or retry timing.

### Fix
Applied locally: no
Upstream-only: yes
Needed change: expose a stable post-reconnect canvas revision/ready signal and either safely queue/replay graph mutations against that revision or return machine-readable `retry_after` plus target revision.

### Environment
Linux; ComfyUI 0.30.1 (remote); NVIDIA RTX 3090 24 GB; comfyui-mcp 0.50.25; panel 0.11.38.

artokun:
Panel half shipped in **comfyui-mcp-panel 0.14.35** (artokun/comfyui-mcp-panel#1216).

The gate already ran before the executor, so "nothing was applied" was always true — it just lived in the refusal *sentence*. The reply now carries it as a field:

```json
{ "rid": "…", "ok": false, "error": "[backend-reconnecting] …",
  "refusal": { "code": "backend-reconnecting", "applied": false,
               "stage": "pre-executor", "retryable": true } }
```

`error` is unchanged and `refusal` is absent on every other failure, so nothing that reads the reply today is affected.

The reason this is a field and not a regex: I wrote the orchestrator half first, matching the refusal text, and reverted it as a P0. Acknowledged panel errors travel as arbitrary `msg.error` text, so any sentence the gate writes is one a genuine **mid-write** failure can also contain — and being wrong there does not print a bad message, it **double-applies a graph mutation**.

Three review rounds tightened who may publish that field, each closing a way the answer could be right for the wrong reason:

[...]

PR #1538, opened 2026-08-13
fix(1529): wait out a reconnect refusal instead of handing it back

artokun:
The orchestrator half of #1529. The panel half shipped in panel 0.14.35.

## What happens today

`panel_set_widget` is refused while ComfyUI's backend is reconnecting. The refusal is
**correct** — a mutation dispatched in that window lands on a canvas the reconnect is about
to rebuild. It is also **safe to retry**, because the panel's gate runs before its executor,
so nothing was applied.

The orchestrator hands that refusal straight back to the caller anyway. The retry path
declines because `panel_set_widget` is a mutation, and mutations are never re-issued on our
own initiative — for good reason.

## Why this is not a string match

The first attempt read "nothing was applied" out of the refusal's **wording**, and was
reverted as a P0. Acknowledged panel errors travel as arbitrary `msg.error` text, so any
sentence the gate writes is a sentence a genuine **mid-write** failure can also contain. Being
wrong here does not print a bad message — it **re-issues a mutation that already landed**.

The panel states it structurally now (comfyui-mcp-panel#1216):

```json
{ "code": "backend-reconnecting", "applied": false, "stage": "pre-executor", "retryable": true }
```

The orchestrator was **discarding that field** — `p.reject(new Error(String(msg.error ?? …)))`
— which is exactly why the first attempt had nothing but text to match on. Three pieces
connect it:

| file | role |
| --- | --- |
| `src/services/panel-refusal.ts` | validates the claim; the only thing allowed to authorise a mutation retry |
| `src/services/ui-bridge.ts` | attaches a **validated** refusal to the Error it rejects with |
| `src/orchestrator/panel-tools.ts` | the retry gate consults it **beside** `isRetrySafeCmd`, not inside it |

That last row is the whole change. `isRetrySafeCmd(cmd) && …` would leave mutations — which
are precisely what #1529 is about — still un-retried.

## Fail closed, in every direction

A refusal is accepted only if it makes **every** claim, exactly: `applied === false`,
`stage === "pre-executor"`, `retryable === true`, and a `code` this side recognises by name.
Anything missing, mistyped, or unrecognised answers `null` and the caller behaves exactly as
it did before the field existed — which is also what happens against every panel older than
0.14.35.

Every hop requires an **own** property, and the own-property test is itself bound at module
load rather than read off the prototype it exists to distrust. Inheritance is the wrong shape
for a claim of provenance: what matters is that *this* object states it.

## Review

Three rounds, two of which found real defects in my own work:

1. **P0** — the own-property rule was applied to the Error only; the wire message's `refusal`
   and the claim's inner fields were plain lookups. A polluted prototype could complete a
   partial claim, so an ordinary mid-write failure would gain retry authority.
2. **P0** — the guard read `Object.prototype.hasOwnProperty` at call time. Harmless alone;
   combined with pollution supplying the values, it hands the guarantee back.
3. Clean.

Under this project's threat model both guard a **mistake**, not an attacker — a process that
can rewrite `Object.prototype` can call the retry path directly. The value is that the rule
holds under a careless dependency, and the fix was two lines.

## Verification

- **13 mutations, all applied, all killed** — including one restoring each reviewed defect,
  one severing the bridge→error channel, and one turning the `||` into an `&&`.
- 501 files / 9424 tests green.
- A JSON round-trip is asserted, because the rule must cost a real panel nothing and every
  reply arrives that way.

## What these tests do NOT prove

Raised in review as a P2, recorded rather than papered over: the tests drive the validator
directly and pin the two wiring lines by source. They do **not** execute the bridge → retry
path end to end, so they do not demonstrate the one-retry maximum or same-tab redispatch after
rebinding. Both are properties of the surrounding retry block, which this change does not
introduce or alter — it adds one disjunct to its condition.

Addresses the orchestrator side of issue 1529; closing that issue is a separate step once this
is released and verified against a live panel.


artokun:
**Answered (1), and it changes what this fix is.**

`RETRY_SAFE_CMDS` holds **reads only** — `graph_serialize`, `graph_outline`, `graph_query`, `graph_get_errors`, … — and `graph_set_widget` is not in it. The switch-guard retry is gated on `isRetrySafeCmd(cmd)`:

```ts
if (isRetrySafeCmd(cmd) && (isTransientReconnectError(err) || isWorkflowSwitchGuardRefusal(err))) {
```

So that path **never fires for a mutation**. #1027's comment describes the panel refusing, saying nothing was applied, and asking for a retry "in a moment" — but for a mutation, no retry happens. The prose and the gate disagree, which is the same shape this session has hit repeatedly.

That means #1529 is **not** a reuse of an existing pattern. It is a new exception to the rule that mutations are never re-issued — which is the most safety-critical line in this file, and the reason `RETRY_SAFE_CMDS` is an explicit allowlist rather than "everything not read-only".

## The argument for the exception, and its exact limit

The panel states it plainly for these two refusals:

> Both refusals are retryable and state that nothing was applied — **true because the gate runs BEFORE the executor**.

That is the same licence #1143 used for the pre-queue Manager refusal: *the response is about the REQUEST, so nothing was queued, so nothing is running.* A pre-executor refusal carries no double-apply risk because the executor never ran.

The limit is equally sharp: this can only ever key on refusals that are **provably** pre-executor. A refusal that merely mentions reconnecting, or one whose text a future panel reuses post-executor, must not qualify — that would re-issue a mutation that may have applied, which is precisely the harm the allowlist exists to prevent.

## Where I am stopping, and why

The safe form of this needs:

- a predicate matching **only** the two documented pre-executor refusals, anchored tightly enough that a reworded or reused message cannot widen it;
- a bound that fits the MCP call budget (the panel says "usually seconds" for backend-down, "~30 seconds" before escalating for settling);
- proof the retry cannot reach any post-executor path, ideally by mutation-killing each of those separately.

That is a genuine change to the double-apply boundary and deserves fresh attention rather than being appended to a long session. The diagnosis is recorded here so the next pass starts from it: the mechanism the report asks for is not needed, the pattern it resembles does not currently cover mutations, and the licence for the exception is specific and already written down in the panel.

Also worth noting for whoever picks this up: reported on comfyui-mcp 0.50.25 / panel 0.11.38 against current 0.51.35 / 0.14.24. I confirmed the gate and the allowlist still behave this way on current main, so the report is not stale.

artokun:
**Reverted. The implementation is not safe, and the reason is one I should have seen because I fixed the same thing today.**

## The P0

My predicate authorized re-issuing a mutation based on **error text**. Adversarial review pointed at this repo's own conclusion, reached in #1468 earlier today:

> acknowledged panel errors arrive as **arbitrary `msg.error` text** … so any sentence the bridge can write, a panel error can also contain — no regex separates them

That is exactly why #1468 stopped matching text and switched to a **bridge-owned tag** (`markReplyTimeout`). I then wrote a text predicate here, in the one place where being wrong does not produce a misleading message but a **double-applied mutation on the user's graph**. An acknowledged executor error — one where the widget *was* set — can begin with either marker and contain both phrases, and my code would have re-sent it.

Tightening the regex does not fix this. The property I need ("the executor did not run") is not derivable from a string the executor itself could have produced.

## What would make it safe

The panel has to say so **structurally** — a field on the refusal, not a phrase in it, e.g. `{ applied: false, stage: "pre-executor" }` — and the orchestrator keys on that. `graphMutationReconnectGate` already knows this fact for certain at the point of refusal; it simply does not currently express it in a machine-readable way.

That is a panel-repo change plus an orchestrator change keyed to it, and it needs both sides landed in order. It is not something to bolt on from this side alone.

## What stands from this pass

The diagnosis, which I think is still the useful contribution:

- the report's requested `retry_after` / revision protocol is **not needed** — the panel already knows the refusal is pre-executor
- the pattern this resembles (#1027's switch-guard retry) is gated on `isRetrySafeCmd`, which is **reads only**, so it never covers mutations — the comment and the code disagree there
- widening that condition would have been **inert** regardless: it waits 400 ms and retries once, against a reconnect the panel itself calls "usually seconds"
- and the authorization must be **structured, not textual**, which is the finding that killed this attempt

The other two review findings were also real and would have needed fixing even with a sound predicate: the terminal returns bypassed switch-hold clearing and the outcome-unknown retry-token mint, and the "12s budget" was uncapped — `1e308` seconds yields `Infinity` and the loop would never expire.

Leaving the PR open with the branch reverted, so the next attempt starts from the constraint rather than rediscovering it.

Issue #1447, opened 2026-08-12
Plugin .mcp.json uses `npx -y` for an 818MB tree: cold start exceeds the MCP startup timeout and each timeout discards the partial install, so retries never converge (-32000)


artokun:
### What happened / root cause

The plugin ships `.mcp.json` with `command: "npx", args: ["-y", "comfyui-mcp", "--full"]`. On a
machine with no `_npx` cache entry for the package, that invocation must fetch and unpack the
dependency tree **before the server process exists**. Measured on this machine:

| cache state | time to `initialize` response |
|---|---|
| fully cold (empty npm cache) | 18.5 s just to install |
| warm tarball cache, cold `_npx` | **9.7 s** |
| warm `_npx` | 1.0 s |
| global install, no npx | **0.48 s** |

The tree is **818 MB / 171 packages**.

The client kills a server that hasn't completed the handshake in time, and that kill **discards
npx's partial install**, so the next attempt starts cold again. The result is a loop that cannot
converge: the user saw `Failed to reconnect to plugin:comfy:comfyui: -32000` on five consecutive
`/mcp` and `/reload-plugins` attempts over ~11 minutes.

Confirmed by cache forensics: `~/.npm/_npx/8af3af54f44861ce` (`comfyui-mcp`) had its
`node_modules` created at 19:54:45 — the timestamp of the **first manual run given a 90 s
budget**, not of any client attempt during 19:43–19:54. Since that entry exists, every launch
handshakes in ~1.0 s.

Two other hypotheses were tested and **rejected**, so they don't need re-checking:
- *stdout corruption* — stdout is clean JSON-RPC; `npm notice run npx` and all `[INFO]` lines go to stderr.
- *concurrent-launch race* — the user also has a `comfyui` entry in `~/.claude/settings.json` running the same package, and both specs hash to the **same** `_npx` directory. Launching both simultaneously into one cold cache: both installed and both handshook cleanly.

### Steps to reproduce

1. `rm -rf ~/.npm/_npx/*` (or use a machine that has never run the package).
2. Install the plugin: `/plugin marketplace add artokun/comfyui-mcp`, `/plugin install comfy@comfyui-mcp`, `/reload-plugins`.
3. `/mcp` → `Failed to reconnect to plugin:comfy:comfyui: -32000`.
4. Retrying never recovers, because each timeout throws away the partial install.

### Exact error (scrubbed)

```
Failed to reconnect to plugin:comfy:comfyui: -32000
```

[...]

PR #1547, opened 2026-08-13
fix(1447): defer the last static optionalDependencies import, so omit=optional is safe for BOTH entry points

artokun:
Claims #1447 — the half left open after 0.51.24 fixed the `serverInfo.version` bug.

**Measured on `origin/main`'s built output**, which is the only thing that decides what npm must install before the process exists:

| built entry | static `@anthropic-ai/claude-agent-sdk` import |
|---|---|
| `dist/orchestrator/panel-tools.js:37` | **yes** — `createSdkMcpServer, tool` |
| `dist/orchestrator/panel-agent.js` | none (type-only, erased) |
| `dist/orchestrator/claude-backend.js` | none (already lazy, with an install-hint reject) |

So the earlier analysis holds exactly: **one** value import is what keeps `npm_config_omit=optional` unsafe, because both `--full` and `--panel-orchestrator` resolve to the same `_npx` directory and the orchestrator statically pulls the SDK in.

Scope of this PR is deliberately the *enabling* change plus real numbers — not the distribution decision. Flipping `omit=optional` in the plugin's `.mcp.json` has a genuine cost for panel users (the Claude backend would then fail at its lazy import until they install the SDK), and the issue's own last comment says that call should be made deliberately rather than quietly. I'll post measured install-size and cold-start deltas so it can be.


artokun:
Closing: the change this PR was opened for is proven useless. `--omit=optional` strips sharp's own platform binary and the server cannot start, so deferring the last static optional import buys nothing. Measurements and the mechanism that actually works (optional peer deps) are on #1447.

Issue #1374, opened 2026-08-11
download_model routes LOCAL downloads through ComfyUI-Manager and hard-fails when Manager isn't loaded (regression in 0.50.x)


artokun:
### What happened / root cause
On a LOCAL Windows-portable ComfyUI (embedded python, resolvable install path), `download_model action:"download"` refuses to start any download:

```
{"error":"MODEL_ERROR","message":"ComfyUI-Manager's queue API is not reachable (neither /v2/manager/queue/status nor /manager/queue/status answered with a queue status). Is ComfyUI-Manager installed and enabled on the connected ComfyUI? The pip comfyui_manager package only activates when ComfyUI is started with --enable-manager.","details":{"url":"https://huggingface.co/.../model.safetensors"}}
```

This target is LOCAL, not remote. Per the tool's own documentation, a local ComfyUI should "stream to disk and surface live progress in the panel download tray" — only a REMOTE target is supposed to dispatch through the Manager HTTP API. Here the local path appears to have been replaced by (or fallen through to) the Manager dispatch, so a missing/unloaded Manager becomes a hard blocker for a capability that does not need Manager at all.

Corroborating detail: EARLIER IN THE SAME SESSION (before this MCP server was upgraded to 0.50.x) `download_model` streamed three HuggingFace files locally on this same machine, with tray progress. After the upgrade, every local download attempt fails as above. ComfyUI-Manager IS present in `custom_nodes/` but is not serving its routes (`/manager/queue/status`, `/v2/manager/queue/status`, `/api/manager/queue/status` all return 404), i.e. the pack didn't import — which should be irrelevant to a local file download.

Impact: ~45 GB of required model weights could not be downloaded through the tool at all. Worked around with `curl.exe -L -C -` (which also gave real resume, unlike the tool). The workaround loses panel download-tray progress entirely.

### Steps to reproduce
1. Local ComfyUI (Windows portable), ComfyUI-Manager present but not serving routes (no `--enable-manager`, or the pack failed to import).
2. `download_model {action:"download", url:"https://huggingface.co/<repo>/resolve/main/<file>.safetensors", target_subfolder:"checkpoints"}`
3. Result: MODEL_ERROR about the Manager queue API. Nothing is downloaded, and no local streaming is attempted.

### Exact error (scrubbed)
See the JSON above. Same root error also surfaces from `install_custom_node action:"list"` (that one legitimately needs Manager — the download path should not).

### Fix
[...]

PR #1562, opened 2026-08-14
fix(369): an INFERRED models root must still be corroborated before a download lands there

artokun:
Refs #369 (reopened). Fixes the recurrence route; the issue stays open until a reporter confirms.

## The recurrence

#369 was "download_model reports success, the model never appears" — 3.79 GB into a stale install. 0.49.7 (#794) added two guards: a **pre-write** disagreement check (candidate models root vs the server's own `/models/<category>` listing) and a **post-write** membership + listing verification.

Both were skipped for three provenances judged LIVE-AUTHORITATIVE. One does not earn it:

| source | what it is | statement by the server? |
|---|---|---|
| `argv-flag` | the server's own `--base-directory` / `--models-directory` | yes |
| `live-root` | the server's own argv `main.py` root | yes |
| `observed-root` | a **relative** argv `main.py` re-anchored on the interpreter the OS reports | **no — inferred** |

`observed-root` establishes where the **binary** lives. Measured on #1374's branch: with a stale portable bundle's python on `PATH`, `cd D:\live && python ComfyUI\main.py` anchors `C:\stale\ComfyUI`. Reachable through both the absolute-`argv[0]` tier and the OS-image reading, so this is not a regression from #1374 — it is what #1374 made me look at.

## The fix

A new, narrower predicate `modelsDirNamedByServer` gates both halves. Deliberately **separate** from `isLiveAuthoritativeModelsDir` rather than an edit to it — that predicate has four callers answering different questions, and narrowing it in place would move all of them.

- **Pre-write**: the disagreement check now runs for an inferred root. It needs positive contradicting evidence gathered root-wide and fails open on everything else, so it cannot refuse a fresh or shared tree. Cost: one listing call.
- **Post-write** (your P1): the membership test that authorizes `visible` now answers UNKNOWN for an inferred root.

The post-write half is the one that mattered and the first draft missed. The pre-write check fails **open** on an empty candidate tree — correctly; an empty directory contradicts nothing — so a stale inferred root can still receive the write. The live server then lists that filename because the user already had the model in the real install, and a name match alone reported the download as a success: #369's original symptom, reproduced after the pre-write half was fixed.

A correct `observed-root` install now gets the qualified note instead of a confident `visible`. The note names the file, the root and the remedy; the false positive was silent until queue time. Its wording previously said "came from local configuration", untrue for this source, and now covers the inferred case.

## Verification

- **499 files / 9390 tests green**, lint clean.
- **10 mutations, all applied and all killed**, against a **checked-green baseline**.

Two findings came only from that discipline, and both were mine:

1. The harness first reported **7/7 KILLED while 27 tests were already failing** — a second `output-dir.js` mock was missing the new export, so every result was measured against red. The harness now refuses to run unless the baseline is green.
2. With a green baseline, the P1 test I had just written **SURVIVED its own mutation**: the fixture put the landed file *outside* the inferred root, so membership was unremarkable either way and it pinned nothing. Rewritten to land the file **under** the stale root, which is the reviewed scenario.

Both `output-dir.js` mocks now take the provenance predicates from the **real** module via `importOriginal` — which sources skip these checks is the entire subject of this issue, and a hand-copied rule in a test is how a suite keeps passing against a rule production no longer uses. A mutation that re-copies it by hand is in the set and is killed.

12 post-write tests moved from `observed-root` to `live-root`: they are about the membership *mechanism* (junction handling, server-swap detection, root stamping) and used that source only as a stand-in, so the swap preserves their intent.

### DIFF
--- a/src/__tests__/services/download-live-destination.test.ts
+++ b/src/__tests__/services/download-live-destination.test.ts
@@ -84,23 +84,30 @@ vi.mock("../../services/extra-paths.js", () => ({
 const resolveModelsDirWithBasesMock = vi.hoisted(() =>
   vi.fn(async (): Promise<unknown> => undefined),
 );
-vi.mock("../../services/output-dir.js", () => ({
-  resolveModelsDir: vi.fn(async () => resolve(h.destModelsDir)),
-  resolveModelsDirWithBases: () => resolveModelsDirWithBasesMock(),
-  parseModelsDirFromArgv: vi.fn(() => undefined),
-  hasUnresolvableRelativeModelDirFlag: vi.fn(() => false),
-  // Real behaviour (pure argv parsing): the "empty tree + no extra roots" refusal
-  // only concludes anything when the server's config location is knowable.
-  parseExtraModelPathsConfigsFromArgvRaw: (argv?: string[]) => {
-    const out: string[] = [];
-    for (let i = 0; i < (argv?.length ?? 0); i++) {
-      if (argv![i] === "--extra-model-paths-config" && argv![i + 1]) out.push(argv![i + 1]);
-    }
-    return out;
-  },
-  isLiveAuthoritativeModelsDir: (s: string) =>
-    s === "argv-flag" || s === "live-root" || s === "observed-root",
-}));
+vi.mock("../../services/output-dir.js", async (importOriginal) => {
+  // The provenance predicates come from the REAL module. WHICH sources skip the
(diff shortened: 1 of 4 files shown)


artokun:
Reopened: an earlier maintenance command accidentally targeted this MCP issue while closing the unrelated Panel #1562 record. No MCP issue resolution was intended or claimed here.

PR #1584, opened 2026-08-14
docs(blog): fix the licensing claims, and stop one sentence being wrong in ten places

artokun:
Audit of all 34 English blog posts: **28 findings confirmed, 3 refuted.** This PR lands the two that could cost a reader money, plus the boilerplate class behind seven of the rest.

## Licensing — verified against the live license, not taken from the audit

| post | claimed | actually |
|---|---|---|
| `krea2-comfyui` | free commercial use "up to **50 seats**" (twice) | **revenue under $1,000,000 USD**, company-wide incl. affiliates. The license contains no seat term at all |
| `ltx-2.3-comfyui` | "LTX-2 released under **Apache 2.0**" (×3, unhedged) | **LTX-2 Community License Agreement**, `$10,000,000` commercial threshold |

A ten-person studio inside a $5M company would have read the Krea post and believed it was licensed.

**ERNIE's Apache-2.0 headline was checked the same way and is correct** — `baidu/ERNIE-Image` and `baidu/ERNIE-Image-Turbo` both declare `apache-2.0`. Not every Apache claim was wrong; the point was to look rather than pattern-match after finding one bad one.

## One sentence, ten copies

> "the panel auto-starts a background agent on your Claude subscription"

Byte-identical in **ten posts**. True when written; when the panel became a pure-frontend pack that cannot spawn a process, it went false in ten places at the same moment — and nothing noticed, because no test covers prose and every copy looked locally fine. All ten now say to run `npx -y comfyui-mcp@latest connect` and click **Connect**.

`docs/snippets/panel-install.mdx` is now the single source of truth, enforced by `scripts/check-blog-boilerplate.mjs`: posts carrying the shared step must match it exactly, and two retired claims are banned outright.

**The gate found two things on its first run:**
- **12** posts carry that step, not the 10 grep found
- two of those are legitimately *different* steps (`train-lora-runpod` sets `RUNPOD_API_KEY`; `video-extend-pusa` applies a pack), so demanding they match would have forced unrelated rewrites — the marker was narrowed to the sentence's distinctive tail rather than its opening clause

Both rules mutation-proven: reword one copy, or reintroduce the retired claim, and it names the file.

## Deliberately not done

Importing the snippet as a component is the better end state and the file is written for it — but the step sits inside a **numbered list**, and a block component inside a list item renders differently than it reads. I could not get `mint dev` running to preview it, and converting ten live posts on an unverified MDX construct is a bet on the docs site. The gate buys the same anti-drift property meanwhile; the conversion wants someone who can render locally.

## Verification

`npm run check:blog` · `check:docs-links` · `check:docs-locale` · `tsc --noEmit` — all clean. The blog gate is wired into `npm test`.


artokun:
Merge-readiness pass: merged `origin/main`, re-verified every licensing claim against the live sources, and fixed two things the audit left behind.

## Licensing claims — re-checked independently, all hold

| claim | source | verdict |
|---|---|---|
| Krea 2: revenue **< $1,000,000 USD**, affiliates aggregated, no seat term | krea.ai/krea-2-licensing | ✅ verbatim: "including all affiliated entities under common ownership or control … total company-wide annual revenue of less than one million United States dollars"; "you must obtain a separate enterprise license from Krea prior to any Commercial Use" |
| LTX-2: **LTX-2 Community License Agreement**, not Apache, $10M gate | Lightricks/LTX-2 LICENSE | ✅ verbatim: "Entities with annual revenues of at least $10,000,000 … are required to obtain a paid commercial use license" |
| ERNIE-Image / -Turbo: `apache-2.0` | both HF model cards | ✅ |
| Ideogram 4: `ideogram-non-commercial-model-agreement` | HF model card | ✅ non-commercial, separate agreement for commercial |
| Z-Image-Turbo, Qwen-Image(-Edit), WAN 2.2 T2V A14B, WAN 2.2 Animate 14B, Pusa V1 | HF model cards | ✅ all `apache-2.0` |
| BiRefNet MIT upstream; `1038lab/BiRefNet` declares nothing | both HF repos | ✅ the mirror genuinely has **no model card and no license** — and `packs/wan-transparent/manifest.yaml` does fetch `BiRefNet_toonout.safetensors` from it |

Repo-side claims spot-checked too: `wan-longer-videos` pins `-Q8_0.gguf` in the manifest, **both** installer scripts, and `workflow.json` (zero `Q4_K_S` for the four experts); the `-96gb` variants really are `fp16.safetensors`; `pack.yaml` really passes `--reserve-vram 2`; `packs/ernie/workflow.json` has exactly **10 combo groups of 20**; `rsyncToPod` really is `tar -czf - | ssh … tar -xzf -` with no delta transfer; `agy` and `pi` really are `vision: false`; and one orchestrator on one bridge port really does serve every provider (`src/orchestrator/index.ts:1450`).

## Two things fixed

**1. The ERNIE post still promised a Z-Image-free pack, in two places.** The audit established the opposite for the `ernie` monolith and rewrote the body table — but two older copies of the retired claim survived:

- troubleshooting: *"use the `ernie` pack, which excludes them"*
- closing CTA: *"ERNIE-only models … with the Z-Image cruft stripped out"*

Both contradicted the new table three screens above them. They now point at `ernie-txt2img` for an ERNIE-only install and say the monolith pulls Z-Image on purpose. (The `ernie-txt2img`/`ernie-img2img` "no" column checks out — the only `ae.safetensors` grep hit in those manifests is `flux2-vae.safetensors`.)

**2. Half a sentence got gated.** The same commit that replaced "roughly 200 tools" with a gated **37** rewrote the *panel* half of that sentence as "a comparable set" — which reads as ~37 against a real **92**. That is the same drift the gate was added to stop, in the same sentence. It now says 92, and `scripts/asset-counts.mjs` guards both halves (CI already runs `--check`).

Mutation-proven both ways: `92 → 91` fails with *"claims 91 for panel_tools, actual is 92"*, and re-flowing the paragraph onto one line still passes (the pattern is `\s+`-tolerant, so a re-wrap can't cry wolf).

## Verification

`tsc --noEmit`, `i18n:check`, `check:docs-links`, `check:docs-locale`, `check:blog`, `check:blog-packs`, `check:blog-stale`, `check:blog-structure`, `asset-counts --check` — all clean.

`npx vitest run`: **9457 passed**, and every one of the 12 reported failures was `Test timed out in 30000ms` with **zero assertion failures** — a load artifact of this box, not a regression. Each affected file passes in isolation (`tool-surface-filter`, `vocabulary`), and `gen-changelog-dedupe` — which spawns `git` per test — passes 9/9 at `--testTimeout=180000`. The failing subset differed between runs, which is the tell.

artokun:
CI went red on `check:blog-stale` — all five stamped posts reported stale. **None of them are.** It is a defect in the new gate, now fixed in `717d7c0`.

## What CI said

```
✗ blog/ernie-image-comfyui.mdx: verified 2026-08-14, but since then: ernie (2026-08-15)
✗ blog/video-extend-pusa-comfyui.mdx: ... wan-pusa-extend (2026-08-15), wan-longer-videos (2026-08-15)
✗ blog/wan-2.2-comfyui.mdx: ... wan-longer-videos (2026-08-15)
✗ blog/wan-animate-comfyui.mdx: ... wan-animate (2026-08-15), wan-pusa-extend (2026-08-15)
✗ blog/wan-transparent-comfyui.mdx: ... wan-transparent (2026-08-15)
```

No pack moved on 2026-08-15. Nothing on this branch touches `packs/` at all.

## Why

`actions/checkout` defaults to `fetch-depth: 1`. A depth-1 clone holds exactly **one** commit, so `git log -1 -- packs/<name>` returns the tip for *every* path. Measured on a real `git clone --depth 1` of this tree:

| clone | `packs/ernie` last changed |
|---|---|
| `--depth 1` | **2026-08-14** — the tip commit |
| full history | **2026-07-30** — the truth |

On CI the tip is GitHub's PR merge commit, stamped at run time in **UTC**. That explains the timing exactly:

- the first run landed at **23:32 UTC** — tip read `2026-08-14`, tied the stamps, `>` was false → **green**
- the next run landed at **00:20 UTC** — tip read `2026-08-15`, `>` every stamp → **red**

It passed once by luck. From UTC midnight onward it fails every run, every day, for a reason that has nothing to do with the packs — which is the "gate that cries every week is a gate people turn off" failure its own docstring warns about.

## The fix

1. **The gate detects a shallow clone and skips instead of guessing.** A false stale is worse than no reading: it trains people to bump stamps they did not earn, which is exactly what the stamp exists to prevent.
2. **CI checks out with `fetch-depth: 0`**, so the gate can actually measure there rather than being inert.

Verified on a real `--depth 1` clone: the old script misattributes the date; the new one prints `skipped: shallow clone` and exits 0. And it still fires on a genuine stale — backdating ernie's stamp to `2026-07-01` reports `ernie (2026-07-30)` and exits 1, so the guard did not defang it.

## Full suite

`npx vitest run --testTimeout=120000` on this exact head: **507/507 files, 9469 passed, 3 skipped, 0 failed.** The failures in my earlier run were all `Test timed out in 30000ms` with zero assertion failures — this box was running ~280 node processes; `gen-changelog-dedupe` alone spawns `git` per test.

artokun:
Correcting two timestamps I gave from memory in the comment above, and adding the run history — which turns out to be a cleaner proof than the argument I made from it.

Every CI run on this branch, by start time:

| head | started (UTC) | result |
|---|---|---|
| `f2cebb5` | 2026-08-14 **23:35:03** | ✅ success |
| `163bc3a` | 2026-08-14 **23:58:53** | ✅ success |
| `2c64a6e` | 2026-08-15 **00:09:49** | ❌ failure |
| `de29343` | 2026-08-15 **00:15:40** | ❌ failure |

The boundary is **UTC midnight**, to the minute. Nothing about the packs changed across those four runs — `163bc3a` and `2c64a6e` are eleven minutes apart and the second one's only new content is two blog paragraphs and a regex.

That is the shallow-clone bug stated as an experiment rather than an argument: the gate was comparing a `verified: 2026-08-14` stamp against the tip commit's date rendered in the runner's UTC day, and the stamp stopped winning the moment the date rolled over.

(My earlier comment said 23:32 and 00:20; the real starts are 23:35 and 00:09. The correction does not change the conclusion — it sharpens it from two data points to four.)

PR #1600, opened 2026-08-15
fix(1415): ask the ComfyUI the panel is on, when the configured one is unreachable

artokun:
Draft for artokun/comfyui-mcp#1415 — `list_packs (action:"list_templates")` addresses the headless `COMFYUI_URL` and stops there, so it fails in a session where the sidebar panel is connected to a live ComfyUI and every `panel_*` graph tool works.

## What changed

When the configured target **cannot be connected to at all**, and exactly **one** connected panel is on a **different** origin, the template index is re-requested at that origin. The result names which server **answered**.

The address the fallback uses is not new machinery. #1553 (0.51.42) built the missing half of the #952 channel: the orchestrator publishes the **server-observed WebSocket handshake Origin** of every connected tab into the progress dir it already shares with every spawned child. Since then the child that raises this failure has been holding the answer and spending it only on a sentence. This spends it on the request.

## Pre-merge review returned NO-SHIP — three P1s

All three were re-verified against the code before being touched. **All three reproduced.** Each is addressed in `fe03699`.

### 1. Redirect SSRF / false attribution

`fetch()` follows redirects by default, and this is the one request in the process whose address comes from **outside the configuration** — a handshake Origin the browser reported. Whatever listened there could move the request to any host it named, while `answered_by` went on naming the origin we picked; the far end's body would be surfaced as a template index credited to a server that never produced it. Confirmed live: with `redirect: "follow"` a 302 to `169.254.169.254` **is** dialled.

**Resolution** — the fallback now passes `redirect: "manual"`, and a 3xx is **refused**, not chased: the error names the status and the `Location` and stops. Separately, `answered_by` is now read off `res.url` — the URL that actually produced the body — instead of being restated from our own intent, so attribution follows the evidence rather than asserting agreement. A spelling-only difference (`http://LOCALHOST:8188` vs `http://localhost:8188`) is still one server, per #1175.

*Probed on the runtime rather than inferred.* The fetch **standard** says `redirect: "manual"` yields an opaqueredirect response — status 0, no headers — which would have made the refusal branch dead code and `Location` always null. That is browser behaviour; undici on node 24 returns the real 302 with headers readable. Because that is a runtime behaviour and not a contract, there is now a test that runs the fallback over a **real socket** against a real redirecting server, alongside the hand-built one that would have passed either way.

### 2. Trigger was far too broad

The gate was "any error that is not a named abort", which admitted every transport failure there is: `ECONNRESET`, `EPIPE` and socket hang-ups on an **established** connection (the server was plainly there), a rejected **TLS** handshake (a peer completed the TCP connection and presented a certificate), and a bare `ETIMEDOUT` (ambiguous between an unanswered SYN and a dead established socket). **The previous version of this PR body claimed the fallback never fires once the connection was accepted. That was false.**

**Resolution** — `mayAskAnotherServer` replaces it, and it establishes exactly one thing, only from codes that prove it:

> **No connection to that address was ever established, so no HTTP server there received the request.**

Admitted: `ECONNREFUSED`, `ENOTFOUND`, `EAI_AGAIN`, `EHOSTUNREACH`, `ENETUNREACH`, `UND_ERR_CONNECT_TIMEOUT`. An **unrecognised or absent code now denies** — that inversion is the point, since the old default allowed everything it did not name.

What it does **not** establish, and is documented not to: that no ComfyUI exists there. A firewall dropping SYNs and a stopped container are identical from here. It licenses *asking someone else*, nothing more.

**Deliberately not reused: `NEVER_DELIVERED_CODES`** from `comfyui/fetch.ts`. The two sets look interchangeable and answer opposite questions — and reusing a classification past its proof is a failure this repo has already paid for once, when a loopback test stood in for "this machine".

| | `NEVER_DELIVERED_CODES` | `NEVER_CONNECTED_CODES` |
|---|---|---|
| asks | could the server have **acted** on my POST? | was there **no server** at that address? |
| TLS handshake failure | never delivered ✔ (no application byte written) | **connected** ✘ (a peer answered the ClientHello) |

So the TLS codes are present in one and absent from the other on purpose. A test pins the divergence through the exported API (`deliveryDoubt` vs `mayAskAnotherServer`), so a later edit that points one at the other's set fails.

### 3. A stale origin could recreate the very bug

The channel admits a record up to `PANEL_ORIGINS_MAX_AGE_MS` (120s) old and then **discarded the timestamp**. A panel that moved A→B inside that window still causes a fetch of A — and the fetch succeeds — while `answered_by_note` asserted, in the present tense, that the panel is on A. The response proves only that **A answered**.

**Resolution** — both halves, not just the minimum:

- **The claim now matches the observation.** The note says the origin *answered this request*, states that it is *not a live reading of where a panel is now*, and that one may since have moved. The old sentence ("the ComfyUI a connected sidebar panel is on") is gone from every emitted message — **including the two error paths**, which carried the identical assertion and which the first pass at this fix missed. Correcting one wording while two survive is precisely how a fixed message gets re-quoted later as current, so a single test sweeps all four outcomes (success, fallback-unreachable, redirect-refused, fallback non-2xx) and asserts the retired string appears on none of them — while still requiring each to name the origin it dealt with, so the phrase is absent because the claim was corrected and not because the text was dropped.
- **The age is carried through and disclosed.** `readPublishedPanelOriginsRecord` keeps the stamp beside the origins, `connectedPanelOriginsSnapshot` resolves it (0 for the in-process bridge, which is read live; the record's age for the channel), and the note says "about 45s ago" / "a moment ago" out loud.

`PANEL_ORIGINS_MAX_AGE_MS` is **unchanged**, so #1553's diagnostic behaviour is byte-for-byte as before. The window is disclosed, not narrowed — narrowing it would trade a lost diagnostic for a problem that honest wording already answers.

## Merge gate round 1 — NO-SHIP, two P1s (both reproduced), plus one found verifying them

Fixed in `fb8611b`.

### 1. The loopback ALIAS case was declined — the topology this feature most obviously exists for

#1175 collapses `localhost` / `127.0.0.1` / `[::1]` into one server so a spelling difference is not reported as **drift**. Right for describing; wrong the moment this module started **acting**. A ComfyUI bound only to `[::1]:p` refuses `127.0.0.1:p`, and the published panel origin `http://[::1]:p` — *an address that answers* — collapsed to `same` and was never tried. **A canonical match is not a socket match.**

`choosePanelFallbackOrigin` now returns `alias` for a same-canonical origin whose **host is spelled differently**, and the caller retries it on the same terms as `use`. An alias **outranks** a foreign origin, including an ambiguous pair: every other verdict here judges *which of several machines the caller meant*, and this one is not a judgement — #1175 already holds it **is** the configured server — so preferring it strictly shrinks the wrong-server risk the `ambiguous` refusal exists to manage.

The secondary-failure message split with it. "A browser reached it and this process cannot — a tunnel, a container boundary" is nonsense between a host and itself, so the alias path says what was actually learned: both spellings name one server under #1175, and neither answered.

Proven over a **real socket bound to `::1` only**, with the configured target the IPv4 spelling of the same port — a genuine `ECONNREFUSED`, not a synthesized one. **Two existing tests had false premises and were rewritten**, not adjusted: the one asserting `same` argued "a fallback here would re-request the same server", and "does not re-ask the same server" used two *different* spellings, so it proved nothing about re-asking one address.

### 2. A path-prefixed mount is invisible, and was not disclosed

Corrected above, under *Not verified* — including the claim in this PR body that it merely 404s.

### 3. Found while verifying those: the sixth emitted outcome pointed at the wrong machine

A 2xx whose body is **not** an index throws through `classifyNonJson`, whose remedy is built from `getComfyUIBaseUrl()`. So after a fallback it told the reader to go and check **the configured base's** `/system_stats` — a server this call had just established cannot be reached at all — while the body being diagnosed came from somewhere else. One message, two machines, named as one.

Round one corrected four emitted sites and round two a fifth; this is the sixth, and it is the exact path a path-prefixed ComfyUI lands on. The note is **appended** rather than the shared helper re-plumbed — `classifyNonJson` serves every JSON read in the process — and the error class and code are unchanged, so downstream readers see the same shape.

**Also corrected:** the `NEVER_CONNECTED_CODES` doc claimed an unsupported scheme is "reported rather than routed around". The structural route added in round two made that false the moment it landed — `ftp://host:8188` **does** reach the fallback. Measured, then written down.

**4 mutations, each confirmed applied by a count that changed, all killed** — delete the alias branch (4 tests), neuter the path disclosure (1), drop the bad-body attribution (1), always claim the tunnel (1). One first attempt broke the parse and reported "no tests"; recorded, because that is indistinguishable from a survivor in its own output, and re-run as a syntax-safe substitution.

**Full suite: 517 files, 9698 passed, 3 skipped.** The single red file, `rgthree-authoring-awareness`, is **inherited, not from this diff** — it reproduces on a clean `origin/main` checkout, where `core.autocrlf` gives the newly added `SKILL.md` CRLF and its `/^---
name: rgthree
/` cannot match. Reported separately rather than fixed in a PR that was not scoped for it.

## Constraints kept from the original design (reviewer confirmed these held)

- **Never a substitute.** The configured target is asked first, always. A **non-2xx is never retried elsewhere** — that status is evidence the configured server exists and answered. As of finding 2, neither is any failure on a connection that was established.
- **No credentials cross.** `comfyuiFetch` injects `COMFYUI_AUTH_*`, configured for the headless target. The fallback is a plain `fetch`, so an origin the user never configured cannot receive it. That is why it is not simply a second `comfyuiFetch` call.
- **Two candidates ⇒ refuse.** Two tabs on two ComfyUIs is an ordinary state, and neither "first connected" nor "looks local" is evidence of which was meant. Both are named and the call keeps failing.
- **No topology claim.** A ComfyUI only the browser can reach still fails — now naming *both* addresses.

## What a published origin establishes — and what it does not

It proves a tab connected to this orchestrator was serving a page from that origin (the browser sets it; page JS cannot forge it). It proves nothing about whether **this process** can reach it, whether a **ComfyUI** answers there, whether it is mounted at the **root** — an Origin carries no path — or, as finding 3 established, **whether that is still true now**.

## Verified

- **82 tests** across `templates-panel-fallback.test.ts` (49) and `panel-origin-channel.test.ts` (33), driven through the real registered `list_packs` handler rather than the private function.
- **12 mutations for the review fixes, every one confirmed applied, every one killed.**

  | # | mutation | killed |
  |---|---|---|
  | 1 | `redirect: "manual"` → `"follow"` | wiring + **real-socket** test |
  | 2 | 3xx detection made unreachable (`status >= 900`) | refusal + **real-socket** test |
  | 3 | attribute to what we *asked*, not what answered | attribution test |
  | 4 | restore the old name-only trigger | **9** tests |
  | 5 | merge the TLS/reset codes into `NEVER_CONNECTED_CODES` | 4, incl. the divergence pin |
  | 6 | `observedAgo` always `""` | 3 age tests |
  | 7 | discard the timestamp again (channel) | 2 files |
  | 8 | drop `ageMs` for the in-process bridge | 2 tests |
  | 9–12 | restore the retired panel claim at **each of the four** emitted-message sites — payload note, non-2xx note, secondary-failure, redirect-refusal | the claim sweep, at every site |

  *One mutation initially looked like a survivor and was not.* Its proof string — the retired sentence — also occurs in a **comment** documenting that retired wording, so `grep -c` returned 1 both before and after: the substitution had silently not applied. Re-run with a before/after count that must **increase**, it applied and was killed. A second attempt then applied but ate a closing backtick, and a file that does not parse reports "no tests" — also not a result. Recorded because a mutation that fails to apply is indistinguishable, in its output, from a test gap.
- The original 10 mutations from the first round still hold (timeout exclusion, ambiguity refusal, same-origin case, credential boundary, canonical-origin comparison, `answered_by` attribution, declined-fallback note, unparseable-target guard, non-2xx attribution, child channel read).
- **Reachability, separately**: one test drives the path with **no injected source at all** — a real `control-panel-origins.json` in a temp `COMFYUI_MCP_PROGRESS_DIR` — because the reported failure happens in the spawned stdio child, which has no bridge.
- Full suite green on a **verified baseline before touching anything** (512 files / 9544 tests) and green again after (512 / 9576). `npm run lint`, `npm run build`, `npm run check:vocabulary` and `npm run check:unknown-collapse` clean.

  One intermediate run showed 2 failures — `late-mutation-e2e` and `training-jobs`, neither in this diff — because a mutation batch was running concurrently against the same machine. Both pass in isolation at `--maxWorkers=1`, and the final quiet run is fully green. Noted rather than omitted, since "it was a flake" is a claim that has to be re-measured, not assumed.

## Not verified

- **Still not live-tested against a real panel plus a real dead `COMFYUI_URL`.** The redirect refusal is now exercised over a real socket, but the panel side is not.
- **Noted, deliberately not changed: `describeTargetDrift` makes the same present-tense claim.** Reading the rendered messages end to end, the pre-existing #952/#1553 drift sentence the error already carries still says "A connected panel **is on** `<origin>` — a DIFFERENT address". That is the same shape as finding 3, on the **diagnostic** surface rather than the acting one, and it ships today in 0.51.42. It is outside this review's three findings and rewording it would change #1553's behaviour, so it is recorded here rather than altered in a fix PR that was not scoped for it.
- **A basePath-mounted ComfyUI is not handled — and this bullet used to be wrong about how it fails.** A handshake Origin carries no path, so `${origin}/api/workflow_templates` reads the ROOT of that address. This said "it will 404 … not silently wrong". The merge gate refuted that by execution: where a **different** ComfyUI answers the root, the index that comes back is real, valid, and from the wrong install, and nothing in the numbers says so. Nothing available here can tell the two apart, so the assumption is now **disclosed in the payload** rather than implied. Detecting it is still impossible without a new channel carrying the page path.
- **The tunnel / loopback-only topology is untouched.** That still needs the panel to fetch on our behalf, which is a panel bridge command in the other repo.
- Only `list_templates` is routed this way. No other headless read was changed.

## Inherited branches

Two branches on the remote carry this issue number and are being **discarded** — both are claim markers with **zero diff** against `main` (`git diff origin/main...origin/<branch>` is empty), from PRs #1428 and #1500, both long since closed:

- `fix/1415-templates-panel-target` (PR #1428, closed)
- `fix/1415-templates-via-panel` (PR #1500, closed)

A third, `fix/1415-panel-origin-drift-in-child`, is **already merged** as #1553 and shipped in 0.51.42 — its content is in `main` and this work builds directly on it. Nothing was salvageable from any of the three; this branch starts from `origin/main`.


artokun:
### Second reviewer, standing down — plus one thing that should not be lost

A second agent session was driving this branch concurrently (I verified independently, then
stopped editing). Everything below is verification only; I made **no commits** and left the
worktree pristine at `23acae2`.

#### What I verified independently

The three P1s from the pre-merge review are real and are **pinned by tests** — confirmed by
reverting each fix and watching the suite go red, not by reading:

| mutation | result |
|---|---|
| fallback `redirect: "manual"` → `"follow"` | **2 tests** red — and the real-socket test genuinely dialled `169.254.169.254`, so the SSRF is demonstrated, not argued |
| `answeringOrigin(res, …)` → the origin we asked for | attribution test red |
| restore the old name-only trigger | **15 tests** red |
| `observedAgo` always `""` | 3 tests red |

Full suite green (517 files / 9713 passed / 3 skipped), `lint`, `check:vocabulary`,
`check:unknown-collapse` clean, CI green on `23acae2`.

#### The thing that could get lost: the gate has asked for **opposite** things on this line

`panel-fallback-target.ts`, foreign loopback spellings, three rounds:

- **round 5** — spellings collapse to `spellings[0]` → *should refuse as `ambiguous`*
- **an independent run** — spellings collapse to `spellings[0]` → *should try them all*
- **round 6** — spellings are now chained → *should refuse as `ambiguous`*

`23acae2` implemented the middle reading and documented why it rejected refusal. **That
reasoning still looks right to me**, and I want to flag the cost of reversing it before
round 6 gets implemented:

1. Refusing on multiple loopback spellings re-breaks **gate round 1's own finding** — a
   ComfyUI bound only to `[::1]:p` while `COMFYUI_URL` says `127.0.0.1:p`. That is described
   in this PR as "the topology this feature most obviously exists for", and it is reached
   *only* by trying more than one spelling.
2. It also declines the ordinary case — one ComfyUI, two tabs, different spellings — which
   this module **cannot** distinguish from two installs without probing both, and probing
   both is precisely what the "never a substitute" constraint forbids.
3. The case the gate is worried about is already **named**: `answered_by` reports
   `http://127.0.0.1:8188` vs `http://[::1]:8188` literally, so the reader can tell which
   listener replied. It is not silent.

If the residue is worth closing, the cheap move is the one this PR already used for the
path-prefix assumption — **disclose it in `answered_by_note`** ("loopback spellings are
treated as one server under #1175; if two installs are bound to different stacks, the named
origin is the one that answered") rather than reversing the chain and losing case 1.

Merging is left to whoever owns this loop — two agents squash-merging one PR is worse than a
slow merge.

artokun:
**Correction to my objection 1 — it does not apply to the fix now in flight.**

I wrote that refusing would re-break gate round 1's `[::1]`-only case. That assumed a
wholesale reversal. The change being made keeps **alias** chaining and refuses only on
**foreign** spellings, on the grounds that an alias of the *configured* target is anchored by
#1175 to the server the user named, while a foreign origin has no such anchor. That
distinction is sharper than mine and round 1's case survives it — objection 1 is withdrawn.

Objection 2 stands and is worth a sentence in the commit: with a foreign panel reached by two
spellings of one ComfyUI (two tabs on `localhost:8188` and `127.0.0.1:8188`, configured target
elsewhere), the refusal declines a case that would have answered. That may well be the right
trade — an unanchored guess is worse than a loud refusal — but it is a real cost, not a
free one.

artokun:
**Second agent, standing down — verification evidence + one trade-off flag.**

I was dispatched onto this PR as "abandoned" work. It is not abandoned: another autopilot
session is actively driving it (it committed `23acae2` at 10:37 and ran gate round 6 at
10:40, while I was verifying). I am not editing the worktree and will not merge — leaving
that to whoever owns the loop, so this does not get squash-merged twice.

Two things worth keeping.

### 1. The fix is genuinely pinned by tests (measured, at `905b9e4`)

Reverting each load-bearing piece breaks the suite, so these are not tests that pass
against unfixed code:

| mutation | tests killed |
|---|---|
| remove the panel fallback entirely | **34** |
| panel path `redirect: "manual"` → `"follow"` | **2** (incl. a real-socket test) |
| `answeringOrigin()` → re-state the request target | **1** |
| `mayAskAnotherServer()` → allow any failure | **15** |

Also at `905b9e4`: full suite `9713 passed / 3 skipped`, `tsc --noEmit` clean, `i18n:check`
and `check:vocabulary` clean, and all four CI jobs green. That green is **stale for
`23acae2` and later** — it has not been re-run against the round-5/6 commits.

The three P1s from the original review (redirect SSRF, over-broad fallback trigger,
present-tense panel claims) are all pinned by the mutations above.

### 2. Rounds 5 and 6 take opposite sides of the same trade — please make that a decision

`23acae2` (round 5) explicitly considered refusing on multi-spelling foreign origins and
rejected it, in its own commit message:

> Refusing instead was considered and rejected: it would decline the ordinary case (two
> tabs on one ComfyUI, spelled differently) to guard a case this module cannot detect

Round 6's finding asks for exactly that refusal, and the in-flight fix adopts it. The cost
round 5 named is real and does not go away: with an unreachable `COMFYUI_URL` and two tabs
on *one* dual-stack ComfyUI spelled `localhost` and `127.0.0.1`, the chooser now returns
`ambiguous` and the fallback declines — the feature stops working for that setup.

Both readings are defensible; what I'd flag is that the PR has now switched sides under
gate pressure without the trade being settled. Worth stating which case is being chosen and
why, so round 7 doesn't switch it back.

### 3. The gate loop is producing self-inflicted findings

Verifiable from the verdict files and this PR's own commit messages:

- round 4's P1 (`redirect: "manual"` breaks healthy redirects) was **introduced by round 2's fix**
- round 5's P1 (configured credential forwarded across a loopback-alias redirect) was **introduced by round 4's fix**

Two of the last three rounds' P1s were created by this PR's earlier rounds. That matches the
known pattern that gate rounds stop paying after ~3. Suggest capping at round 7: land it once
the ambiguity finding is closed, and take anything further as a follow-up issue rather than
another round on a 2,600-line diff.

artokun:
### Round 7, finding 2 — the cap's stated justification is wrong, which makes the fix obvious

```ts
/** Redirect hops followed on the CONFIGURED target before giving up. Five is the
 *  usual browser allowance; a proxy chain longer than this is a loop. */
const MAX_CONFIGURED_REDIRECTS = 5;
```

**Five is not the usual browser allowance.** The Fetch standard's limit is **20** — *"if
request's redirect count is 20, then return a network error"* — and undici, which backs the
`fetch` this walker replaced, implements that. Chrome and Firefox use 20 as well.

So the gate is right and the regression is larger than "an unusual proxy chain": this PR
silently narrowed the allowance from 20 to 5 for **every** configured target, while the
comment asserts it kept parity. The loop runs `hop = 0..5`, i.e. 6 requests, i.e. at most 5
redirects followed — a 6-hop chain that returned 200 before this PR now fails with
"redirected more than 5 times".

The fix is to match what it replaced (20) rather than pick a second arbitrary number; the
loop-detection intent is preserved, because 20 is still a bound. Worth correcting the comment
in the same commit — it is the reason the wrong number looked justified.

### Round 7, finding 1 is already adjudicated

The path-prefix mount was raised in gate round 2 and resolved deliberately: it cannot be
detected (an Origin carries no path), so it is **disclosed in `answered_by_note`** rather
than defended against. That disclosure is in the payload today. Unless the gate is asking for
detection — which needs a new channel carrying the page path, and is out of scope for a fix
PR — this one is answered and re-raising it should not hold the merge.

*(Verification context: my own gate run on `116dc49` returned INDETERMINATE — codex could not
launch, having been contended by the concurrent gate at 10:55. Not a code signal. CI is green
on `116dc49` and I re-ran the affected suites: 128 passing, lint clean.)*

artokun:
**Gate run on `116dc49` (independent, second agent): NO-SHIP, 2 P1s.** Still not editing this
worktree or merging — handing these over.

```
[P1] src/tools/skills-access.ts:718 — the fallback can return a successful template index
     from the wrong ComfyUI: panel mounted at /comfy/, another install serving the same
     origin's root, root install's valid index returned as success.
[P1] src/tools/skills-access.ts:809 — the five-redirect cap breaks previously working
     configured targets: a valid six-redirect chain reaches 200 through native fetch,
     but this walker stops before requesting the final URL.
```

### The redirect cap is a real regression, measured on a real socket

`MAX_CONFIGURED_REDIRECTS = 5` with `for (hop = 0; hop <= MAX; hop++)` allows 6 requests, so
5 redirects. Native `fetch` follows 20 (the Fetch standard's limit) — which is what the
configured target got before this PR replaced it with a manual walker. Local server, 6
redirects then a 200:

```
chain: 6 redirects then 200
native fetch : status=200 url=http://127.0.0.1:61932/r/0 body={"core":[]}
PR walker(5) : fell out of the loop -> throws 'redirected more than 5 times'
```

A configured ComfyUI behind 6+ redirects worked before this PR and fails after it. Fix is
the constant: **5 → 20**, matching what it replaced. Cheap, and worth doing regardless of
what happens to the rest.

### The pathless-Origin finding is round 4's finding 1, returning

Round 4 raised this at the same `skills-access.ts:718`. The response was to *disclose* it in
`answered_by_note` ("An origin carries no path, so this was read from the ROOT of that
address…") — the behaviour did not change, so the gate has raised it again. The round-3
body-shape check doesn't help either: a real ComfyUI at the origin root returns a valid
object-of-arrays and passes.

This one can't be closed inside the fallback. A browser `Origin` carries no path, by
definition, so the module cannot know whether the panel's ComfyUI is at `/` or `/comfy/`.
The options are:

1. **Accept it** as a documented limitation — it *is* disclosed in the payload, and this is
   a local-trust-domain tool. Then the gate needs an explicit waiver on this finding,
   because it will keep flagging it.
2. **Publish the base URL, not the Origin.** If the panel reports its ComfyUI base (origin +
   path prefix) instead of the bridge inferring from the handshake `Origin`, the whole class
   disappears — and that is what #1415 actually asked for ("a panel-routed template-index
   path … or a dedicated panel tool exposing `/api/workflow_templates`").

Option 2 is the real fix and is a separate change; option 1 is a decision only the owner can
make. Grinding more gate rounds won't resolve it either way — 2 of the last 4 rounds' P1s
were regressions introduced by this PR's own earlier rounds (round 2's `redirect: "manual"`
→ round 4's finding; round 4's walker → round 5's credential leak *and* this redirect cap).

### What's verified good

At `116dc49`: 83 tests in the affected file, 128 across the four touched files, `tsc
--noEmit` clean. The round-6 refusal is genuinely pinned — reverting `different.length === 1`
to `>= 1` fails 8 tests, including the two new ones written for it.

artokun:
## Closing this — it implements the one approach this issue explicitly parked, and the merge gate never shipped it

Not merging, and the reason is not the gate. It is that this PR does the thing issue 1415 had already decided against, **13 minutes** before this PR was opened.

From that issue's most recent owner comment (2026-08-15T04:34Z), under *"What actually remains — and why it is parked, not dropped"*:

> **The shortcut I am explicitly not taking**, so it does not get re-proposed: having the orchestrator fetch the panel's *observed* origin directly. It works for a public proxy URL and fails identically in the tunnel case — the one case that motivates the work — and #1006/#1359 rejected it on the stronger ground that two hosts can both answer and disagree.

That is this PR's mechanism, exactly: it takes the published handshake Origin and fetches `${origin}/api/workflow_templates`. This PR was created at 04:47Z. It also concedes the same limitation in its own *Not verified* section — *"the tunnel / loopback-only topology is untouched"* — so it does not reach the case that motivates the work, which is the specific ground the parking decision rested on.

Two further reasons, each sufficient on its own:

- **The freeze.** This is not a bug fix; it is a new behaviour — a headless tool making an outbound request to an address that came from a browser handshake. Stabilisation-only is the standing rule.
- **Eight merge-gate rounds, never SHIP.** Rounds 3, 4, 5 and 7 each found real P1s, all reproduced against the runtime before being touched. That is not a gate being fussy; it is a design with a large surface of ways to be quietly wrong.

### Round 8, and why it is the parking decision in sharper form

> the default tokenless loopback bridge accepts any Origin plus a client-supplied `hello`, so an untrusted webpage can publish its own origin — with `COMFYUI_URL` unreachable and attacker-controlled `/api/workflow_templates` returning index-shaped JSON, `list_templates` returns attacker-controlled templates.

The premise checks out: the primary loopback listener is deliberately token-less (`ui-bridge.ts:235`, `:1390`) so the local browser panel works without pairing. Under this project's local-trust-domain model that is normally fine — **but this PR is what changes the stakes.** Until now a published origin only ever produced a diagnostic *sentence* (#952/#1553). This PR promotes it to an address the process *fetches*, whose body becomes tool output an agent reads. "Two hosts can both answer and disagree" and "any local page can nominate one of them" are the same objection, one step apart.

Fixing that needs either an authenticated bridge or the panel fetching on our behalf — both cross-repo surface, which is what the issue parked.

### What was genuinely found here, so it is not re-derived

The defects the gate found were real and are fixed on this branch (kept, not deleted: `fix/1415-templates-through-panel`, head `b65b201`). Each was reproduced first, then pinned by a test, then mutation-checked:

| round | defect | status |
|---|---|---|
| 3 | the candidate walk continued past a server that ANSWERED (reset/TLS), so an alias failure advanced to a foreign machine | fixed + test |
| 3 | the fallback body was never checked — a 200 `{"error": …}` was surfaced as a listing, `source_count: 1` | fixed + test |
| 3 | `aliases.slice(0, 1)` discarded other socket spellings, so a live address was never dialled | fixed + test |
| 4 | `redirect: "manual"` (added in round 2) **broke working reverse proxies** that legitimately redirect this endpoint | fixed + real-socket test |
| 5 | the redirect walk leaked `COMFYUI_AUTH_*` across a loopback alias — `sameOrigin` collapses `127.0.0.1`/`[::1]`, which are different listeners | fixed + test |
| 7 | a hard-coded `AbortSignal.timeout(8000)` overrode `COMFYUI_MCP_HTTP_TIMEOUT_S` — **pre-existing on `main`** (`skills-access.ts:538`), not introduced here | fixed + test |

Two things worth keeping regardless of what happens to this approach:

1. **The timeout defect is on `main` today.** It is small, self-contained, and nothing to do with the panel fallback: a user who raises `COMFYUI_MCP_HTTP_TIMEOUT_S` for a slow remote still has this one read aborted at 8s. Worth lifting out as its own change.
2. **One recurring mistake caused four of the six defects above:** applying #1175's loopback-alias collapse — a rule for *describing* drift — to an *acting* decision (which socket to dial, which origin may receive a credential). Anything that revisits this area should treat "canonical match" and "socket match" as different questions.

### Where that leaves the issue

Unchanged from the parking comment, which was already correct: the diagnostic half shipped in 0.51.42, and the real fix is a panel bridge command plus a consumer that chooses by source — deferred as cross-repo surface under the freeze. Leaving the issue open and parked rather than closing it, because nothing here changed what it is waiting on.

Round 8's second finding is not actioned, and it is worth saying why: it asks for an alias to refuse when a foreign origin is also connected, which is the *opposite* of what round 1 required ("an alias must outrank a foreign origin"). Past round three the gate began trading one ordering for the other, and that is a signal to stop rather than to keep cutting.

artokun:
**Heads-up to whoever is driving this branch: I pushed `9d339b9`.** Your local
`fix/1415-templates-through-panel` is now behind — `git pull --rebase` before your next
commit. Nothing of yours was changed except one assertion, explained below.

Two regressions from the manual redirect walk, both measured on live sockets, both pinned:

**1. Hop cap 5 → 20.** The constant's comment called five "the usual browser allowance". It
isn't — the Fetch standard's limit is 20, undici implements it, and the `comfyuiFetch` call
this walker replaced followed all 20. So the cap didn't tighten a policy, it broke targets
that worked before this PR:

```
chain: 6 redirects then 200
native fetch : status=200
walker(5)    : throws "redirected more than 5 times"
```

**2. The timeout budget was per-hop; what it replaced was per-request.** One `comfyuiFetch`
with one signal walked the whole chain under a single deadline. A ceiling re-applied on
every hop multiplies by hop count — and raising the cap to 20 would have stretched that to
20× the configured timeout:

```
budget=200ms, 3 redirects x 150ms each
native fetch (one budget, whole chain) : ABORTED at 209ms
walker PER-HOP signal (the defect)     : 200 after 620ms
walker SHARED signal (the fix)         : ABORTED at 201ms
```

The deadline is now built once from `defaultComfyTimeoutSignal()`, so **your round-7 finding
still holds** — the configured ceiling governs, never a hard-coded 8s — and the walk as a
whole is bounded. A foreign hop keeps its own 8s on top via `AbortSignal.any`, so it can
shorten the budget but never extend past the walk's deadline.

**The one thing of yours I changed.** `template-index-timeout.test.ts` asserted
`init.signal` is `undefined` — the *shape* of the round-7 fix rather than the finding behind
it. Passing the configured signal explicitly satisfies that finding equally and additionally
bounds the walk, so the assertion now checks the outcome: the signal must be one
`defaultComfyTimeoutSignal()` handed out. Re-introducing the hard-coded 8s still fails it,
which I verified — your protection is intact.

Verified: reverting the cap fails 2 tests, reverting the shared deadline fails 1,
re-introducing the hard-coded 8s fails 2. Full suite 9720 passed / 3 skipped, `tsc --noEmit`
clean.

Still outstanding and **not** something either of us can fix here: the pathless-Origin
finding (gate rounds 4 and 7). I checked `ui-bridge.ts` — `connectedServerOrigins()` returns
handshake Origins only and no path is ever stored, so the fallback cannot know whether the
panel's ComfyUI is at `/` or `/comfy/`. That needs the panel to publish its base URL, or an
explicit waiver. It is the one thing standing between this PR and a clean gate.

PR #1601, opened 2026-08-15
fix(packs): a pack's vram tier must name the UNet its manifest fetches (#1585)

### LINKED ISSUE #1585: 8 wan packs report a vram: tier that contradicts the UNet their manifest fetches — 4 in the OOM direction
`enumeratePacks()` (src/tools/skills-access.ts:165) surfaces `vram` and `description` straight from `pack.yaml`, so `list_packs` hands the agent whatever those strings say. In 8 packs they contradict the UNet the pack's own `manifest.yaml` downloads.

## Measurement

For each pack, the precision token in the `vram:` headline vs. the precision of the file the manifest fetches into `unet/` or `diffusion_models/`:

| pack | `vram:` headline says | manifest fetches | direction |
|---|---|---|---|
| `wan-longer-videos-i2v` | 24GB+ (Q8_0) | **Q4_K_S** | overstates |
| `wan-longer-videos-t2v` | 24GB+ (Q8_0) | **Q4_K_S** | overstates |
| `wan-longer-videos-v2v` | 24GB+ (Q8_0) | **Q4_K_S** | overstates |
| `wan-transparent-img2vid` | 24GB+ (Q8_0) | **Q4_K_S** | overstates |
| `wan-longer-videos-i2v-96gb` | 24GB+ (Q8_0) | **fp16** | **understates → OOM** |
| `wan-longer-videos-t2v-96gb` | 24GB+ (Q8_0) | **fp16** | **understates → OOM** |
| `wan-longer-videos-v2v-96gb` | 24GB+ (Q8_0) | **fp16** | **understates → OOM** |
| `wan-transparent-img2vid-96gb` | 24GB+ (Q8_0) | **fp16** | **understates → OOM** |

Both directions are user-visible, and they fail differently:

- **Overstating** (4 packs) is the quiet one. A 12 GB card is told the pack needs 24GB+ and skips it. The pack would have run — it fetches Q4_K_S, and the pack's own vram string already says `<12GB with Q4_K_S`.
- **Understating** (4 packs) is the loud one. The `-96gb` variants load `wan2.2_*_14B_fp16.safetensors` — full fp16 dual A14B, which is why they are named for a 96 GB card. A user told "24GB+" runs it on a 4090 and OOMs. The `-96gb` suffix in the directory name is currently the only honest signal, and `list_packs` reports `name` from the directory while reporting `vram` from the contradicting yaml.

The `-96gb` variants' `description` has the same defect: it says "Q8_0 quant by default" while the workflow loads fp16.

## Also: the `-96gb` packs declare a colliding `name:`

`packs/wan-longer-videos-i2v-96gb/pack.yaml` has `name: wan-longer-videos-i2v` — the non-96gb sibling's name. Same for the t2v and v2v 96gb variants.

**This is currently inert**, and worth stating so nobody "fixes" the wrong layer: `enumeratePacks()` sets `name: entry` from the directory and never reads `meta.name`, so resolution is unaffected and the right pack is applied. But it is a live trap for any future code that keys on the declared name, and it is presumably how the description/vram drift got copied in.

## Why nothing caught it

There is no check tying `pack.yaml` prose to `manifest.yaml` contents. The same class had already gone wrong in the blog — `wan-2.2-comfyui` told readers `wan-longer-videos` ships Q4_K_S and fits under 12 GB when it pins Q8_0 — which is what surfaced this. That side is now gated by `scripts/check-blog-packs.mjs`; the pack side is not.

Suggested fix: a `scripts/check-pack-vram.mjs` asserting that the precision token in `vram:`/`description` matches the UNet in `manifest.yaml`, wired into `npm test`. That makes the drift impossible rather than correcting 8 strings that will drift again.

Note the `(none)` case must stay legal — most packs' `vram:` names no precision at all (`"12GB+"`), and that is fine. Only a *stated* tier that contradicts the manifest is a failure.

## Environment

Found by static comparison of `packs/*/pack.yaml` against `packs/*/manifest.yaml` at 0.51.49; no GPU run needed to reproduce.

artokun:
Fixes #1585.

`enumeratePacks()` (`src/tools/skills-access.ts:165`) hands `list_packs` the `vram` and `description` strings straight out of `pack.yaml`. Nothing compared them to `manifest.yaml`, so the prose drifted from the weight the pack actually downloads.

## The measurement

Sizes are `Content-Length` from the manifest's own resolve URL, followed to the CDN — not a guess and not the pack's description:

| pack | headline said | manifest fetches | measured | direction |
|---|---|---|---|---|
| `wan-longer-videos-i2v` / `-t2v` / `-v2v` | `24GB+ (Q8_0)` | Q4_K_S | **8.15 GB**/UNet | overstates |
| `wan-transparent-img2vid` | `12-24GB (Q8_0 I2V…)` | Q4_K_S | **8.15 GB**/UNet | overstates |
| `wan-longer-videos-i2v-96gb` / `-t2v-96gb` / `-v2v-96gb` | `24GB+ (Q8_0)` | fp16 | **26.62 GB**/UNet | **understates → OOM** |
| `wan-transparent-img2vid-96gb` | `12-24GB (Q8_0 I2V…)` | fp16 | **26.62 GB**/UNet | **understates → OOM** |
| `qwen-image-edit` | `12GB+ (Q4_K_S <12GB…)` | Q8_0 only | **20.27 GB** | **understates → OOM** |

For reference on the same mirror: Q5_K_S is 9.44 GB and Q8_0 is 14.35 GB, which is where the `24GB+ (Q8_0)` headline came from — it is the monolith `wan-longer-videos`'s tier, and that pack really does fetch Q8_0 and is correct. The slices inherited the string without the weight.

The workflows corroborate the split independently: every non-96gb workflow uses `UnetLoaderGGUF` and every `-96gb` workflow uses `UNETLoader`, and the widget values match their own manifest exactly. Only `pack.yaml`'s prose was wrong.

**`qwen-image-edit` is a ninth pack the issue did not list.** The new gate found it: `vram: "12GB+ (Q4_K_S < 12GB, Q5_K_S 12-16GB, Q8_0 24GB+)"` reads as a tier table, but the Q4_K_S and Q5_K_S lines are commented out in `manifest.yaml` — the pack only ever installs the 20.27 GB Q8_0. A 12GB card is told "12GB+".

## Why nothing caught it, and what now does

`scripts/check-pack-models.mjs` structurally cannot see this class: it compares `workflow.json` to `manifest.yaml` and never reads `vram` or `description`. Correcting nine strings without a gate just resets the clock, so this adds `scripts/check-pack-vram.mjs`:

> If `vram:` (or `description:`) states a precision token at all, the **first** one — the headline the reader takes away — must be a precision the manifest fetches into `unet/`, `diffusion_models/` or `checkpoints/`, read from **both** the destination filename and the source URL. If the manifest fetches such a model but nothing about it names a precision, the headline is **unverifiable**, and that fails.

The legal shapes still have to stay quiet, and they do: a string naming no precision (`"12GB+"`, `"<6GB"` — most packs) is fine, alternative tiers listed *after* the headline are fine, an `fp8` headline over an `fp8_scaled` UNet is the same family rather than a contradiction, and a pack that fetches no diffusion model at all is not judged. Nine negative tests pin that half.

Wired into `npm test` via `src/__tests__/scripts/check-pack-vram.test.ts` (which runs the gate over the real `packs/` tree, not only fixtures) and as an explicit `npm run packs:check-vram` step in `ci.yml` so the failure is legible without reading a vitest report.

---

## Review finding: the gate had a false negative of its own

A pre-merge review ran `checkPack` against two hostile cases and **both returned an empty finding list**:

| case | manifest entry | old gate |
|---|---|---|
| renamed | `url: …/model_fp16.safetensors`, `local_path: diffusion_models/model.safetensors` | `[]` — clean |
| schema-form | `url: …/model_fp16.safetensors`, `model_type: diffusion_models`, `filename: model_fp16.safetensors` | `[]` — clean |

The gate inspected only a precision-tagged `local_path`. So a Q8 headline backed by an fp16 URL passed, and a manifest written in the `model_type`/`filename` shape was never examined at all.

**Why the shape half is real, not hypothetical.** `manifestSchema` (`src/services/manifest.ts:110`) marks `url` as the only required field — `model_type`, `filename` and `local_path` are all optional — and `applyManifest` resolves the target as *`local_path`, else `model_type` (default `checkpoints`) + `filename` (default the URL's basename)*. That is **five** legal entry shapes, and the gate read one.

**Why the URL half is real.** `packs/wan-animate-ofm` ships the "renamed" case today: it saves `diffusion_models/WanModel.safetensors` — a name stating nothing — from a URL ending `…_fp8_scaled_e4m3fn_KJ_v2.safetensors`. The old gate could not judge that pack at all. Unioning the two sources also *prevents* a false positive in `packs/wan-multitalk`, which saves `…_GGUF_Q8.gguf` (token `Q8`) from `…-Q8_0.gguf` (token `Q8_0`) — the same quant spelled two ways.

### "Cannot determine" now fails rather than passes

The old code returned `[]` the moment no precision was found, so *unverifiable* and *consistent* were the same answer. That is the hole both false negatives went through.

It now **fails**, and deliberately not a warning: a warning exits 0, leaving CI green, which is indistinguishable from the false negative it would be reporting — and this script has now shipped two of those. The scope is narrow enough to be safe, and I measured it rather than assuming: **6 packs** fetch a diffusion model whose precision cannot be read (`anima`, `anima-img2img`, `anima-inpaint`, `anima-txt2img`, `ltx-2.3-img2vid`, `ltx-2.3-txt2vid`) and **all 6 state no precision**, so none trip it. It blocks nothing today. A pack fetching *no* diffusion model is still not judged, since its tier may legitimately describe a base model the user already has.

### Shape coverage, measured

Enumerated across all `packs/` manifests rather than assumed — every model entry, keyed by its exact field set:

| entry shape | entries in `packs/` | gate reads it |
|---|---|---|
| `url` + `local_path` | 406 (100%) | yes |
| `url` + `model_type` + `filename` | 0 | yes |
| `url` + `model_type` | 0 | yes |
| `url` + `filename` | 0 | yes |
| `url` alone | 0 | yes |

**Exactly one shape ships today**; the other four are schema-legal and unused. There is no shape in `packs/` the gate cannot read.

Per-pack precision readability, 56 packs with a `pack.yaml` + `manifest.yaml` (all 56 fetch at least one diffusion model):

| | old gate | this PR |
|---|---|---|
| precision readable | 49 | **50** (`wan-animate-ofm` gained, via its URL) |
| not readable by either | 6 | 6 — all state no precision, so 0 fail |
| packs stating a precision headline | 21 | 21, all of them readable and checked |

### Mutation results — the direction the gate's own output can never show

Each mutation is applied to a **real shipped pack** in a copy of `packs/`, with a witness string counted **before and after** so a mutation that silently did not apply cannot score as applied. (The first run of this harness did exactly that: a witness naming `diffusion_models` while the pack used `unet/`, plus CRLF line endings defeating a `$\n` anchor, produced a false "applied" on case D. Counting both ends caught it.)

| shape mutated | shipped packs | witness | pre-fix gate | this PR |
|---|---|---|---|---|
| A. `local_path`, precision in filename | 56 | 0→1 | RED | RED |
| B. `local_path` renamed, precision only in URL | 1 | 0→1 | **GREEN — false negative** | RED |
| C. `model_type`/`filename`, no `local_path` | 0 | 2→0 | **GREEN — false negative** | RED |
| D. `url` only, target fully defaulted | 0 | 4→0 | **GREEN — false negative** | RED |
| E. UNet names no precision anywhere | 6 | 0→1 | **GREEN — false negative** | RED |

Four of the five were proven false negatives on the pre-fix gate. The messages the fixed gate produces:

```
A  wan-longer-videos-i2v-96gb — vram: says "Q8_0", manifest fetches fp16
B  wan-animate-ofm            — vram: says "Q8_0", manifest fetches fp8
C  wan-longer-videos-i2v      — vram: says "Q8_0", manifest fetches Q4_K_S
D  wan-longer-videos-t2v      — vram: says "Q8_0", manifest fetches Q4_K_S
E  ltx-2.3-txt2vid            — vram: states "Q8_0", which nothing in the manifest corroborates
```

### What remains unreadable

Nothing in `packs/` — every shipped entry shape is read. Two limits are worth stating plainly rather than claiming past:

- **The precision must be spelled in the filename or the URL basename.** A weight whose name states nothing (the 6 packs above) cannot be verified; the gate now says so and fails a tier claim over it instead of passing it. It does not read GGUF/safetensors headers, which is the only way to actually settle those.
- **The URL is read as a basename only.** A repo path can carry a precision that is not the file's (`Kijai/WanVideo_comfy_fp8_scaled/…` hosts more than `fp8_scaled`), so reading the full URL would import noise. A manifest that fetches an fp16 file from a path whose *basename* names nothing still lands in the unverifiable bucket.

## What I verified

- **The gate is not a no-op.** Run against `git archive origin/main packs`, it reports **17 fields across 9 packs** — the 8 in the issue plus `qwen-image-edit`. Against this branch: `56 pack(s) checked, no vram/manifest contradictions`.
- **The gate reaches production data.** Reverting one shipped `pack.yaml` (`wan-longer-videos-i2v-96gb`) to its `origin/main` value turns the suite red. A fixture-only suite would have proved the script works and nothing about what ships.
- **Two false negatives found in this gate, both invisible in its own output.** The first was the `\b` boundary: `_` is a word character, so `\b` never fires between the `_` and the `f` in `wan2.2_i2v_high_noise_14B_fp16.safetensors`, and the gate silently passed **all four `-96gb` packs** — the OOM half — while reporting a clean run. The second is the `local_path`-only reading above. Neither was findable from the gate's output; the first needed the known-bad `origin/main` data, the second needed hostile cases run against `checkPack` directly.
- The four new shape tests and the unverifiable test were confirmed **red before the fix** (4 failed / 14 passed), green after.
- `npx vitest run src/__tests__/scripts/check-pack-vram.test.ts` 18/18; `src/__tests__/scripts/` 72/72.
- `npm run lint`, `npm run build`, `npm run packs:check-vram` green on this branch, after merging `origin/main`.

## What I did NOT verify

- **No GPU run.** The numbers are file sizes, not observed peak allocation. `26.62 GB` for one fp16 UNet exceeding a 24GB card is arithmetic, not a measured OOM; the `<12GB (Q4_K_S)` headline is the packs' *own* prior claim promoted from the alternatives clause to the headline, not something I measured.
- The `-96gb` "96GB" figure comes from the manifest's own comment (`96GB Blackwell tier … --highvram`) and the directory name. I did not establish the true floor; a 32GB card may or may not run it with offload. The headline now names fp16 and the per-UNet size, which is the part that was actively misleading.
- I did not touch the workflows or manifests' model lists. No pack downloads anything different than before this PR — only the strings describing it changed.
- I did not run the full suite again after the merge with `origin/main`; the earlier full run was 9535 passed, with 2 load-related timeouts in `tools/vocabulary.test.ts` that were green 57/57 re-run alone and are untouched by this change.

## Separate finding, not fixed here

`packs:check-models` is referenced by **nothing**: not `npm test`, not `ci.yml`, not `packs.yml` — it exists only as an npm script nobody runs. Running it on `origin/main` today gives **7 errors and 102 warnings** (e.g. `z-image-xy-plot` references three `Margot-Robbie-Z3*.safetensors` LoRAs its manifest never downloads). Wiring it in means fixing those seven first, which is a different change from this one, so it is left alone and flagged here rather than silently bundled.


artokun:
Splitting this. The tier corrections are shipping alone in #1608, and the gate moves to #1609.

The data half is measured and right — real `Content-Length` per manifest URL, and it caught a ninth pack the report missed. It should not wait.

The gate has now been refuted by **execution** in three consecutive rounds: the `\b` boundary that passed all four `-96gb` packs, the `local_path`-only read, and now any-match union logic that still clears a Q8 headline fetched from an fp16 URL plus an unnormalized `./` path. Each round printed a clean run — which is exactly why the rounds keep failing: a gate's false-negative direction is invisible in its own output, so passing is not evidence.

Your work on it is not lost; #1609 records all three refutations, the shape problem (any-match is the wrong predicate — it needs the most-demanding precision, not any match), and the per-entry-shape mutation protocol including the CRLF trap that nearly scored a false pass. Your measured coverage numbers are in there too.

Closing this in favour of the two.

Issue #1539, opened 2026-08-13
panel_install_node advertises git repository installs but sends an unlisted URL to Manager registry


artokun:
### What happened / root cause
`panel_install_node` accepted a Git repository URL for an unlisted custom node and returned `queued: true`, but the Manager queue then rejected it as a registry lookup. The tool description explicitly advertises `repository`/nightly installation, while the live Manager path cannot resolve this repository URL.

### Steps to reproduce
1. Search Manager for `IPAdapter`; receive the Git URL `https://github.com/Wenaka2004/comfyui-anima-ipadapter`.
2. Call `panel_install_node` with `repository` set to that URL and `version: nightly`.
3. The call returns queued/pending.
4. Poll `panel_node_queue_status`.

### Exact error (scrubbed)
`Node 'comfyui-anima-ipadapter@nightly' not found in [ManagerChannel.dev, ManagerDatabaseSource.cache]`

### Fix
applied locally: no
upstream-only: yes
Workaround verified: use the headless `install_custom_node` action with `source: git`; it cloned and was verified by `panel_list_nodes`.

Precise change needed: when `panel_install_node` receives a non-empty `repository` Git URL that is not a Manager registry id, route it through the same verified git-install path as `install_custom_node`, or reject it before dispatch with an accurate error. Do not return queued success for a registry lookup that cannot accept the URL.

### Environment
OS: Windows 11
ComfyUI: 0.32.0
ComfyUI frontend: 1.48.7
GPU: NVIDIA GeForce RTX 5090, 31.8 GB VRAM
Python: 3.13.12
PyTorch: 2.12.1+cu130
comfyui-mcp: 0.51.35
comfyui-mcp-panel: 0.14.26

artokun:
Traced this end to end across both repos. **Our side sends the payload the Manager's own schema asks for**, so I am not shipping a speculative fix — here is what is established and what still needs a live check.

## What we send

`panel_install_node` → `nodesInstallCommandArgs` (node-management.ts) → the panel's `nodes_install` → `buildInstallRequest` (`web/js/lib/manager-install.js`).

For a git URL arriving via `repository` **or** `id`, the v4 ("v2" dialect) payload is:

```
{ id: gitRepoName(url), version: <sel>, selected_version: <sel>, channel: "dev", repository: url }
```

with `<sel>` = `nightly`. That is deliberate and matches the Manager's own model, which the panel source quotes inline:

[...]

Issue #555, opened 2026-07-31
download_model misclassifies Manager v4 HTTP 405 as legacy v3


artokun:
### What happened / root cause
After migrating a remote ComfyUI 0.29.0 host to Manager v4 and launching with --enable-manager, Manager v4 queue APIs are active (panel_node_queue_status returns v4-style pending_count), but comfyui-mcp 0.48.21 POSTs /manager/queue/install_model and receives HTTP 405. It then incorrectly reports that legacy Manager 3.x is installed. apply_manifest also says no Manager HTTP API.

### Steps to reproduce
1. Remote ComfyUI 0.29.0, Manager v4 enabled, legacy clone disabled.
2. Confirm panel_node_queue_status works and includes pending_count.
3. Call download_model with a valid Hugging Face resolve URL.
4. Receive 405 and misleading Manager 3.x diagnosis.

### Exact error
ComfyUI-Manager API 405 Method Not Allowed for /manager/queue/install_model Arbitrary-URL model installs REQUIRE Manager v4+ ... NOTE: this ComfyUI runs the LEGACY ComfyUI-Manager 3.x.

### Fix
Applied locally: no. Upstream-only: yes. Detect v4 from its version/queue endpoint and use the current v4 model-download route/method. Do not classify every 405 as legacy v3. apply_manifest should share the same v4 adapter.

### Environment
Linux; ComfyUI 0.29.0; Python 3.12.3; NVIDIA RTX 5090 31 GB; comfyui-mcp 0.48.21; panel 0.11.20; Manager v4 queue active.

PR #1611, opened 2026-08-15
fix(1539): a git-URL install stops asking a Manager channel nobody chose

artokun:
Draft. Addresses artokun/comfyui-mcp#1539.

## The short version

I went looking for the refusal and found the diagnosis was wrong. The reporter's
pack **is** in ComfyUI-Manager's registry. It failed because **we asked the wrong
channel** — a channel nobody chose, on the git-URL path only. That is ours, and
this PR stops asking for it.

## What was measured

Live ComfyUI-Manager **V4.2.2** on 2026-08-14, plus that build's own source
(`glob/manager_core.py`, read off disk in the rig's venv).

**1. The pack is listed.** `Wenaka2004/comfyui-anima-ipadapter` is present in the
`default` channel's `custom-node-list.json` (5887 packs) *and* in the rig's own
cached copy of that list, *and* in the live `getmappings` payload
`panel_search_nodes` reads. It is absent from the `dev` list (1210 packs). Those
two lists share **3** entries — they are separate lists, not a superset/subset
pair.

**2. We asked `dev`.** The panel's `buildInstallRequest` defaults the v4 git-URL
payload to `channel: "dev"`, while defaulting the registry-ID payload — and both
3.x shapes — to `"default"`. The Manager's error names the channel the *request*
chose, which is why every one of these reports quotes `ManagerChannel.dev`.

**3. The channel is what the lookup reads.** `install_by_id` for a `nightly` spec
does `get_custom_nodes(channel, mode)` → `load_nightly` →
`get_data_by_mode(mode, 'custom-node-list.json', channel_url)`, keys that map by
the **bare repo name** (`y.split('/')[-1]`), and on a hit clones
`the_node['repository']` — the *channel's* URL. The `repository` field we send is
stored in the task params and never read on that path. (That half of the earlier
finding stands, and is now confirmed at the source rather than inferred.)

**4. Reproduced and isolated** by posting the panel's exact payload to
`/v2/manager/queue/task`:

| channel | mode | result |
|---|---|---|
| `dev` | `cache` | `Node 'comfyui-anima-ipadapter@nightly' not found in [ManagerChannel.dev, ManagerDatabaseSource.cache]` — the reporter's error verbatim |
| `default` | `cache` | same failure, naming `ManagerChannel.default` |
| `default` | `remote` | same again, naming `.remote` |

Nothing was installed by any of it (verified against `custom_nodes/`).

## Why the earlier probe could not have caught this

The confirming run used `artokun/cmcp-1539-does-not-exist` — a repo that exists in
**no** channel. Both hypotheses ("v4 ignores `repository`" and "we asked a channel
that does not list it") predict an identical failure for that input, so the probe
could not separate them. It is sound evidence that the URL is never cloned; it is
not evidence about *listing*, which is the claim that got shipped.

## Which remedy, and why

**(a) route the URL through `install_custom_node`'s git path — not shipped, your
reasoning holds.** `panel_install_node` drives whatever ComfyUI the *panel* is
bound to; the headless clone path writes into the orchestrator's own configured
ComfyUI (`getComfyUIBaseUrl()` / `managerBaseUrl()`), and nothing makes those the
same install. The repo already carries this exact asymmetry in writing — the
`settleDroppedEnqueue` header says the headless tool "owns the filesystem it
installs into. This one drives whatever ComfyUI the PANEL is bound to, which need
not be this machine — so it can report the truth but must not quietly clone
somewhere else" — and `panel_restart_comfyui` learned the same lesson the
expensive way in #851, where a fallback aimed at `COMFYUI_URL` told a reporter
`No ComfyUI process found on port 8188` while their panel was working fine. The
tool is still **named**, with its precondition stated: it needs a LOCAL ComfyUI,
because it clones onto this machine's filesystem.

**(b) refuse before dispatch when the registry does not list it — not shipped,
and the measurement is why.** The only registry oracle reachable over the bridge
is `nodes_search`, and it cannot answer this question:

- it reads `extension-node-map.json`, a **different file** from the node list the
  install resolves against, cached separately and measured stale by a different
  amount on the same rig (cached map Aug 9 vs. bundled list Jun 16);
- it is **not channel-scoped at all** — `getmappings?mode=cache` and the same
  route with `&channel=dev` returned byte-identical bodies — so it cannot see the
  axis that actually decides the outcome;
- and **against the report's own input it answers "listed"**, because that search
  is literally where the reporter got the URL (repro step 1). A guard keyed on it
  would never have fired on the case it was written for, while mutation testing
  read it as covered.

Testing the guard against the report's own input is what killed it, which is the
rule working as intended.

The Manager-generation problem you flagged turned out to be moot for the shipped
change, but it is real for the refusal and worth recording: the orchestrator
cannot establish the panel's Manager generation **before** dispatch. No read-only
bridge command reports it (`nodes_install` carries `dialect` in its reply, but
that is after the fact), and inferring it from the queue-status shape is exactly
the classification #555 forbids — `detectManagerApi` deliberately refuses to
brand a generation on route shape and confirms it with `/v2/manager/version` vs
`/manager/version` instead. A v4-derived refusal therefore could not have been
kept off 3.x.

## What actually changed

One field, on one route. `nodesInstallCommandArgs` now sends `channel: "default"`
for a git-URL install when the caller named none.

- **An explicit channel still wins**, including `"dev"` — 1207 of that channel's
  1210 packs live nowhere else, so forcing `"default"` unconditionally would
  un-ship a real population, not a hypothetical one.
- **The registry-ID route is untouched**; the panel already defaults it to
  `"default"`.
- **Blank counts as unset.** The panel spells it `channel || "dev"`, so `""` and
  `"   "` are read as absent and substituted — forwarding either would land right
  back on the channel this change exists to stop asking for.

Sent from the orchestrator rather than fixed in the panel's default deliberately:
because `channel || "dev"` yields to any non-empty value, this reaches panels
already installed — including 0.14.26, the version the report was filed from —
with no panel release. The panel's default is still worth changing on its own; it
is not what unblocks this reporter.

## What I did NOT establish

- **That `default` always resolves.** On a pip Manager v4,
  `manager_util.is_manager_pip_package()` is true, which sends `get_data_by_mode`
  down its offline branch **unconditionally**: it reads the cache file for that
  exact channel URL, else the snapshot bundled in the package, and never fetches
  — so `mode` is inert there. A host whose configured `channel_url` is not the
  default one therefore has no `default` cache either. That is measured, not
  feared: the rig I used is configured to a custom channel URL, and the pack
  resolved from **neither** `dev` nor `default`, with the server log showing both
  reads landing on `site-packages/comfyui_manager/custom-node-list.json`. Asking
  for "whatever channel the user configured" is not expressible in v4.2.2
  (`channel: null` makes `load_nightly` return `{}` outright), so it is out of
  reach from here and the description says so rather than promising a fix.
- **That a listed repo always installs.** Still not measured; nothing here
  installed anything.
- **Manager 3.x behaviour.** Unchanged and still unmeasured — that path sends
  `channel || "default"` already, so this is a no-op there.
- **That the reporter's own host is on the stock channel.** Their error names the
  channel *we* sent, which tells us nothing about their configuration.
- ~~**The false queued-success is not gone.**~~ **Retracted — I was wrong, and it
  is fixed.** I wrote this before reading `classifyInstallOutcome`, which in panel
  `main` (0.14.43) takes the Manager's own terminal record for the submitted task,
  keyed by `ui_id`, and returns `state: "failed"` **ahead of** the not-drained
  early return — so a registry-lookup miss now surfaces as a thrown error from
  `nodes_install` instead of `queued: true`. Verified reachable, not just present:
  `waitForQueueDrain` -> `verifyInstalled` -> `classifyInstallOutcome`. Its own
  comment names this issue and the exact symptom. So the reporter's "do not return
  queued success" ask is answered panel-side; what remained was the request being
  wrong, which is this PR.

## The description

The shipped `#1539` paragraph asserted an UPSTREAM limitation and told callers a
listed repo was necessary. Measured false for the reporting case, so it is
replaced with the mechanism, with the ownership split named, and with both limits
above kept explicit so the fix cannot read as a guarantee. The literal
`[ManagerChannel.dev, ManagerDatabaseSource.cache]` quote is now parameterised —
after this change the tool no longer asks for `dev`, so that string would send a
reader hunting a message their own call can no longer produce.

## Verification

- `install-node-git-expectation.test.ts` rewritten: 16 tests, driving the
  reporter's actual URL.
- Mutation-tested against a verified-green baseline, each mutation `grep -c`'d to
  prove it applied: reverting the channel default → 4 fail; forcing `"default"`
  unconditionally → 3 fail (the two "MUST STILL WORK" cases and the relay);
  treating blank as explicit → 1 fail; bypassing `...cmdArgs` at the call site →
  the dispatch test fails, which a helper-only suite would not have noticed.
- Full suite: 513 files, 9558 passed / 3 skipped. `npm run lint`, `npm run build`,
  and `npm run check:vocabulary` all clean.

## Follow-ups, not in this PR

- comfyui-mcp-panel: `buildInstallRequest`'s `channel || "dev"` default is the
  origin of this and should change there too.
- comfyui-mcp-panel: `unlistedGitUrlAdvice` tells the caller "the pack is not in
  the registry under that name" on this exact error. For the reporting case that
  sentence is false.
- The upstream question worth asking the Manager project: a nightly install
  cannot name the user's configured channel, and a pip build silently answers
  from a bundled snapshot when the asked-for channel has no cache.

---

# Review round — both P1s fixed

Both reproduced. Thank you — the second one is the same over-claim I had just
retracted, surviving in my own replacement text.

## P1-1: dev-only packs regress

Reproduced, with a **narrower blast radius** than costed, and the proposed remedy
does not survive checking.

**Blast radius.** On a pip Manager v4, `get_data_by_mode` reads the cache file for
that exact channel URL, else the snapshot bundled in the package — and nothing but
the configured channel's own updater writes such a cache. The bundled snapshot is
a stale **default**-flavoured list: 3587 packs, holding 3482 of default's 5891 and
**exactly 1 of dev's 1223** dev-only entries. So on a stock v4 the previous `dev`
never reached the dev list either; it landed on that same bundled file. The
population that genuinely regresses is **users whose Manager is configured to the
dev channel** — for whom the note now hands over the exact argument that restores
them.

**The retry steer, checked — and it does not hold.** Your premise is right about
the *first* attempt: a not-found returns from `install_by_id` before `to_path` is
even computed, and four live probes left `custom_nodes` untouched. It is false
about the *second*. v4 resolves a from-source install by the **bare repo name**
(`load_nightly` keys on `y.split('/')[-1]`) and clones `the_node['repository']` —
the URL in *that channel's* entry, not the URL the caller passed. Comparing the
two channels on that key:

- **38** bare names appear in both;
- **35** of them point at **different authors' repositories**.

`comfyui-birefnet` is hieuck's on `default` and mohammadaboulela's on `dev`;
`comfyui-lmstudio`, `comfyui-kittentts`, `ComfyUI-Image-Selector` and 31 more are
the same story. An automatic retry would therefore sometimes clone arbitrary code
from an author the caller never named — the wrong-target bug the standing finding
warns about, only automated and unattended.

So I took your fallback: **do not silently pick between two candidates, name
them.** Refusing outright was rejected as the larger harm (it breaks every
existing bare git-URL call), so the pick is made, **disclosed in the reply**, and
reversible in one argument. The note names the channel asked, says a miss rules
out *that channel only*, hands over `channel:"dev"` as the retry, and says why it
is not done for you. Emitted only when this code chose the channel, and
independently of the #789 nightly-rewrite note, which rides only the `latest`
branch — otherwise an explicit `version` would get a silently-defaulted channel.

## P1-2: "absent from every channel"

Straightforwardly true, and fixed: the sentence now says only what a miss
establishes — absence from the channel **this call asked**, and nothing more —
then points at an explicit `channel` before the give-up advice. Since I implemented
disclosure rather than a retry, the honest sentence is about the one channel that
was asked.

Worth recording: the blanket assertion I wrote against the phrase then caught a
**second copy** downstream in the queue-done advice, which I had judged acceptable
because it was phrased conditionally. It now names the action that can change the
outcome (a different channel) before the one that gives up.

## Mutation results

Verified green baseline of 24, each mutation `grep -c`'d to prove it applied:

| mutation | tests killed |
|---|---|
| drop the channel disclosure entirely | 5 |
| disclose over a caller-chosen channel | 1 |
| compute the note but never join it in | 5 |
| delete the note append at the call site | 1 |
| restore the retracted "absent from every channel" | 1 |

A sixth attempt **failed to apply** (stale literal) and was caught by its own
`grep -c` before it could read as a pass.

Full suite after merging `origin/main`: **514 files, 9591 passed / 3 skipped**.
`lint`, `build`, `check:vocabulary` clean. Still a draft; not merged.


artokun:
## Gate log, and the one finding that is staying

Four `codex-gate` rounds. Three rounds produced distinct, real findings and all
three are fixed and pinned. Round 4 produced no new finding.

**Round 2 — the reply contradicted itself.** `unregisteredPackEscapeHatch` rides the
*same note* as the new channel disclosure and said a not-found means the pack "is not in
the Manager's registry at all — NO Manager route, by any spelling", while the text
appended immediately after it said the miss rules out ONE channel and to retry another.
Two mutually exclusive conclusions in one reply, and the first sends a pack that *is*
remotely installable — the reporter's own — down the local-only clone path. This was the
over-claim this branch already retracted from the tool description, surviving one layer
down because the description test reads the blurb and never the runtime note. Fixed, and
there is now a test that asserts the two halves of a single reply agree.

**Round 2 — the pip caveat named the wrong trigger.** It blamed "a host configured to a
non-default channel URL". Measured false against V4.2.2 on disk: `is_manager_pip_package()`
is just *"the Manager is not inside a `custom_nodes` tree"*, so the offline branch fires on
every pip/Desktop install whatever the configuration. And the caches a stock host holds do
not line up with the channel NAMES — `default_cache_update()` writes the cache for
`DEFAULT_CHANNEL` (`ltdrdata/ComfyUI-Manager`) at startup, while the name `default`
normalizes through `channels.list` to `Comfy-Org/ComfyUI-Manager`, a different URL and so a
different cache file, which does not exist. This rig's cache directory holds exactly one
node list, hashed from the ltdrdata URL, and the bundled snapshot it falls back to (3587
packs) does not contain the reporter's pack. So the split is by **packaging**, not
configuration: a cloned Manager fetches the real 5891-pack `default` list and the report is
fixed; on a pip Manager old `dev` and new `default` land on the same bundled file, so the
change is a **no-op** there. Improvement or no-op — it cannot make a host worse, and the
description no longer implies otherwise.

**Round 3 — the fix had a hole that swallowed the report itself.** Whether the orchestrator
reroutes decides the version rewrite, the channel and every note; but the *panel* decides
independently whether its `buildInstallRequest` takes the git branch, which is where
`channel || "dev"` lives. The predicates disagreed:

```
orchestrator   /^(https?:\/\/|git@|git\+)/i  ||  endsWith(".git")
panel          /^(https?|ssh|git):\/\//i || git+ || git@ || ".git" || looksLikeOwnerRepoShorthand   ← #301
```

So `ssh://`, `git://` and a bare `author/repo` fell through with no channel and the panel
asked `dev` anyway — and `author/repo` is the form this tool's own description tells callers
to pass. `{id:"Wenaka2004/comfyui-anima-ipadapter"}` reproduced the reported failure *with
this branch applied*. Nine tests were written against those spellings; nine failed before
the fix. Mirrored from the panel's predicate rather than by widening `looksLikeGitUrl`,
which is shared with the headless clone path.

### The finding that is not being "fixed", and why

> defaulting an omitted channel to `default` can install another author's repository —
> `mohammadaboulela/ComfyUI-BiRefNet` resolves the bare name in `default` and clones
> `hieuck/ComfyUI-BiRefNet`.

Raised in rounds 1, 2, 3 and 4. It describes a real property of Manager v4 — and it is
**not one this diff introduces, and not one this diff can remove**:

* **Pre-existing and symmetric.** Before this branch an omitted channel asked `dev`, which
  substitutes exactly as readily in the opposite direction.
* **The direction is measured, and it now points the right way.** Of the 35 bare names
  listed in both channels under different authors, `extension-node-map.json` — the
  channel-independent file `panel_search_nodes` reads, and the reporter's stated source for
  their URL — carries the **default** author's repo for **33** and the **dev** author's for
  **0** (2 in neither). Every colliding URL this tool can actually hand a caller was
  resolving to the *wrong* author under the old default.
* **Unreachable from here.** v4 resolves by bare name and clones the URL recorded in the
  channel's entry; the `repository` field is stored in the task params and never read. No
  request field pins the URL. The only way to guarantee identity is to refuse the route,
  which breaks every existing git-URL install — the larger harm.

It is therefore **disclosed** rather than silently taken: in the tool description (read
before the call is made) and per-call in the dispatch note, which names the owner from the
caller's own URL and says a success can be the wrong repo.

Filed separately so the residual hazard is not lost with this PR.

artokun:
Residual wrong-author hazard tracked in #1616.

PR #1619, opened 2026-08-15
fix(1616): refuse the ambiguous from-source install instead of picking a repo

### LINKED ISSUE #1616: Manager v4 resolves a from-source install by BARE REPO NAME, so a git URL can install another author's repo
Split out of #1539 / #1611 so it is not lost with that PR. **Not a regression** — it is a
standing property of ComfyUI-Manager v4 that we currently disclose rather than prevent.

## The mechanism

Read out of ComfyUI-Manager **V4.2.2**'s own source (`glob/manager_core.py`), not inferred.
`install_by_id` for a `nightly` spec does
`get_custom_nodes(channel, mode)` → `load_nightly` →
`get_data_by_mode(mode, 'custom-node-list.json', channel_url)`, keys that map by the **bare
repo name** (`repo_name = y.split('/')[-1]`), and on a hit clones `the_node['repository']` —
the **channel's** URL. The `repository` field we send is stored in the task params and never
read on that path.

So for a git-URL install, what gets cloned is whatever the channel's list has under that bare
name — which need not be the repository the caller named.

## How big it is, measured 2026-08-15 against the live channel lists

Bare names that appear in more than one channel under **different** repository URLs:

| scope | count |
|---|---|
| `default` ∩ `dev`, different author | **35** |
| across all six channels (`default`/`dev`/`recent`/`legacy`/`forked`/`tutorial`) | **65** |

Examples: `ComfyUI-BiRefNet` is `hieuck`'s on `default` and `mohammadaboulela`'s on `dev`;
`ComfyUI-Image-Selector` is `SLAPaper`'s vs `Latentnaut`'s; `ComfyUI_ascii_art` is
`Deep-Neko`'s on `default` and `iris-Neko`'s on `legacy`.

## Why #1611 did not fix it

There is no request field that pins the URL — v4 ignores `repository` on this path entirely.
The only way to *guarantee* target identity is to refuse the git-URL route, which breaks every
existing bare git-URL install. #1611 therefore took the choice, measured which way it should
point, and disclosed it:

* of the 35 `default`/`dev` collisions, `extension-node-map.json` — the channel-independent
  file `panel_search_nodes` reads, i.e. where a caller actually gets a URL — carries the
  **default** author's repo for **33** and the **dev** author's for **0** (2 in neither). So
  the pre-#1611 `dev` default was resolving *every* reachable collision to the wrong author,
  and `default` resolves them correctly;
* the residue is a caller who hand-types a URL for an author that only a non-default channel
  lists.

## What would actually close this

A pre-dispatch refusal for the ambiguous names: when the bare name of the caller's URL is
known to resolve to a *different* URL in the channel about to be asked, refuse and name both
candidates rather than picking. That needs an embedded snapshot of the ~65 ambiguous names and
their per-channel URLs, which brings its own staleness story (it fails safe — a stale entry
causes an unnecessary refusal that an explicit `channel` clears, and a new collision is simply
not caught, which is today's behaviour).

Deliberately **not** done in #1611: it is a new mechanism with its own failure mode, not a fix
to the channel bug that PR was scoped to.

## Current mitigation

Disclosed in two places rather than silently taken:

* `panel_install_node`'s description says a success can be the wrong repo, before any call is
  made;
* the per-call dispatch note names the owner from the caller's own URL and tells them to
  confirm with `panel_list_nodes` that what landed is what they meant.

artokun:
Closes #1616.

## What this does

The pre-dispatch refusal #1616 asks for. A git-URL install whose bare repo name is
**measured** to resolve to a different repository than the caller named is refused before
anything is dispatched, and **both candidates are named** rather than picked between.

```
panel_install_node({ repository: "https://github.com/mohammadaboulela/ComfyUI-BiRefNet" })

REFUSED before dispatch — this install would not necessarily have installed the
repository you named. ComfyUI-Manager v4 resolves a from-source install by the BARE
REPO NAME ("ComfyUI-BiRefNet") against ONE channel's list and clones the URL recorded
there, ignoring the `repository` field entirely. You asked for
https://github.com/mohammadaboulela/ComfyUI-BiRefNet, and the "default" channel's list
carries that name under https://github.com/hieuck/ComfyUI-BiRefNet. So this would have
cloned someone else's code and reported success. Not picked for you: The repository you
named is what channel:"dev" resolves "ComfyUI-BiRefNet" to — pass that and this
proceeds. (…measured 2026-08-15; naming any `channel` explicitly makes the choice yours
and bypasses it. Bypass it that way also if this ComfyUI runs Manager 3.x, where the
from-source request carries the URL in `files` and IS cloned as passed…)
```

## The issue's numbers moved, in the direction that matters

The mechanism was re-read at ComfyUI-Manager's own source rather than carried over from
the issue text — and re-reading it changed the measurement twice.

**#1616's count is case-SENSITIVE, and the lookup is not.** `get_custom_nodes` stores into
a `NormalizedKeyDict` whose `get` keys on `key.strip().lower()`, so `comfyui_tagger` and
`ComfyUI_tagger` are one name. Reproducing #1616's method case-sensitively against today's
lists gives 63 / 33 — its 65 / 35 modulo a day of upstream churn, so the delta below is
the case fold, not a disagreement about the data.

| scope | #1616 | measured here |
|---|---|---|
| all six channels | 65 | **111** |
| `default` ∩ `dev` | 35 | **60** |

**And a class #1616 does not describe: 28 of the 111 are ambiguous WITHIN a single
channel.** Two entries under one name collapse onto one key — `load_nightly` overwrites on
an exact-case duplicate, `NormalizedKeyDict` on a case-variant one — and the survivor is
decided by the order of an upstream JSON file. No second channel is involved: `default`
alone lists both `1038lab/ComfyUI-Pollinations` and `ciga2011/ComfyUI-Pollinations`.

The snapshot therefore records **every** repository under a name rather than the one
Manager's overwrite would leave standing. Modelling that overwrite hides the loser
completely — an exact-case duplicate reads as a unique name and the guard never fires —
and it makes an upstream *reorder* silently flip a call from correct to wrong-repo.

The snapshot is not hand-written: re-running `scripts/gen-manager-name-collisions.mjs`
against the live channel lists reproduces `manager-bare-name-data.ts` **byte-for-byte**.

## Two routes, judged by whatever actually decides each

Manager gates the channel lookup on the version (`install_by_id`):

```python
if version_spec == 'unknown' or version_spec == 'nightly':
    custom_nodes = await self.get_custom_nodes(channel, mode)   # channel read HERE
    ...
res = self.cnr_install(node_id, version_spec, ...)              # and NOT here
```

* **From-source** (`nightly`/`unknown`, which is also where an omitted or `latest` version
  normalizes): the channel's list decides. Refused when the channel about to be asked
  carries that name under something other than what the caller named.
* **Registry version** (an explicit `1.0.0`): the channel is never read and neither is
  `repository` — Manager installs the Comfy Registry pack whose **id is the bare name**.
  Refused for a contested name, with the remedy that actually works: drop `version`.

## Where it fires, and where it deliberately does not

It fires **only on the channel this code defaulted to**. A caller who names a channel made
the choice themselves, so that call dispatches with the concrete collision named instead
of the generic warning that rode along before.

**That escape is not equally meaningful everywhere, and the text no longer pretends it
is.** For the 28 intra-channel names, naming a channel selects nothing — both entries are
in the one list — so the messages say so rather than crediting a choice the caller could
not make. See gate round 2.

Also untouched: the registry-**id** route (the id IS the identifier there), and any name
the snapshot has not seen collide. 111 names out of ~5850 in `default` alone.

## What the guard cannot see — stated in the source, not papered over

* **Which Manager generation is on the other end.** Bare-name resolution is a v4 property;
  the 3.x dialects send `files:[url]` and clone the URL as passed, so there this refusal is
  a false positive — and it still fires. The panel picks the dialect from its own ComfyUI
  *after* the request leaves the orchestrator, and the orchestrator's dialect cache is keyed
  to the base URL **it** talks to, which need not be the ComfyUI the panel drives; reading
  it would gate one machine's install on another machine's Manager. Disclosed in the
  refusal text and the tool description.
* **CNR re-keying.** When a channel entry's repo is registered in the Comfy Registry,
  `get_custom_nodes` keys it by the CNR id instead of the bare name. Of the 249 channel
  entries behind these 111 names, **32** are re-keyed to an id that is not the bare name;
  for those the real lookup misses and Manager answers "not found". The guard cannot know a
  host's CNR state, so it refuses there too. That trade only ever turns a not-found into a
  clearer refusal — it can never block a resolution that would have returned the caller's
  own URL, because a channel listing exactly the caller's repo under that name is allowed
  through on the from-source route.
* **A pip Manager** reads its per-channel cache or the bundled snapshot rather than the
  lists measured here. The bundled list is `default`-flavoured, which is the channel this
  guard is most accurate about.
* **Staleness**, both ways. Regenerate with `node scripts/gen-manager-name-collisions.mjs`.

## Verified by deleting it

| mutation | result |
|---|---|
| remove the refusal in `nodesInstallCommandArgs` | **12** tests fail |
| make the name lookup case-sensitive | **14** fail |
| compare repos without the host | the off-host test fails, **and only it** |
| make the channel-bypass sentence unconditional again | **2** fail |
| make the named-channel warning always credit the choice | **1** fails, the right one |
| drop the own-property check on the channel key | the `__proto__` test fails |
| disable the registry-version refusal | **1** fails, the right one |
| let `nightly`/`unknown` reach the registry path | **15** fail |
| give the registry refusal a channel bypass | **1** fails, the right one |

Full suite: **9670 passed**, 516 files. `npm run lint`, `check:vocabulary`,
`check:unknown-collapse`, `npm run build` all green.

## Codex gate

**Round 1 — three P1s.** The host-blind comparison (`gitlab.com/x/y` compared equal to
`github.com/x/y`), the invisible exact-case duplicate (the generator modelled Manager's
overwrite, so a duplicated name read as unique and the guard never fired — this is what
moved the snapshot 104 → 111), and the 3.x false positive, which is undetectable from the
orchestrator and is disclosed instead.

**Round 2 — the refusal recommended the bypass it had just ruled out.** Every refusal
closed with "naming any `channel` explicitly makes the choice yours and bypasses it",
including refusals whose previous sentence said "No channel in the snapshot resolves
&lt;name&gt; to the repository you named". A caller acting on the last thing they read
passed `channel:"default"` for `1038lab/ComfyUI-Pollinations` and got `ciga2011`'s code
reported as success — the exact install this guard exists to prevent, via the guard's own
advice. Measured: of the **65** repositories inside a multi-candidate channel, **zero**
have any channel that resolves them uniquely, so that sentence was wrong for all of them.
The bypass is **kept** (those 65 have no other channel, and on 3.x the install is correct);
the text is narrowed to what is true, in the refusal, the named-channel warning, and the
tool description.

**Round 3 — an inherited key was read as a channel hit.** `channel:"__proto__"` handed back
`Object.prototype`, whose `.length` is `undefined`, so neither the falsy nor the
`length === 0` check stopped it and the tool threw `TypeError: repos.map is not a function`;
`constructor` reached `sameRepo(undefined, …)`. An own-property check at all three lookups.
Not a trust boundary — one machine — but a correctness gap against this file's own promise
to tolerate an unknown channel rather than throw.

**Round 4 — the channel exemption fired where no channel is read.** A wrong *allow*, not a
wrong message. `{repository:"https://github.com/hieuck/ComfyUI-BiRefNet", version:"1.0.0"}`
returned no conflict because `default` carries hieuck's repo — while an explicit version
skips channel resolution entirely and installs registry id `comfyui-birefnet`, which is
`viperyl/ComfyUI-BiRefNet` (api.comfy.org, publisher "Zephrus shawn"). Caller names hieuck,
gets viperyl, reported as success. Fixed by splitting the two routes.

**Round 5 — the new refusal took a channel bypass it has no use for.** Round 4's fix
mirrored the from-source refusal too closely and let an explicit `channel` through, which
re-opened exactly what it had just closed: `{hieuck/ComfyUI-BiRefNet, version:"1.0.0",
channel:"default"}` dispatched and installed viperyl's registry pack again. Every reason
the other refusal yields to a channel is absent on this route — Manager never reads a
channel there, and the 3.x escape does not apply because the panel's 3.x shapes hardcode
`version:"unknown"` with `files:[url]` and discard the caller's version (verified on the
panel's `origin/main`), so a 3.x host never reaches the registry route at all. Dropping
`version` is a remedy that works on both generations, so the refusal is now unconditional
and says so. `registryVersionWarning` existed only to serve that bypass and is deleted
rather than left unreachable. **The asymmetry with the from-source refusal is deliberate
and is pinned by a test.**

### One finding deliberately not taken

Rounds 1 and 3 both argued the explicit-`channel` bypass should not exist at all for
intra-channel collisions, since two different repository requests collapse to one install
target. That is true, and it is a trade rather than an oversight: removing the bypass
strands **65 repositories** with no alternative channel — measured, not assumed — and
breaks them on Manager 3.x, where the URL ships in `files` and the install is *correct*.
Preventing a disclosed, opt-in coin flip on v4 is not worth permanently bricking 65 packs
on a route where they work today. What was genuinely wrong there was the messaging that
claimed the caller's channel choice had resolved something; that is fixed.

## Also corrected here

`gitInstallSubstitutionNote` and the tool description both quoted "35 bare names … under
DIFFERENT authors" — the case-sensitive undercount, now 111, and "repositories" rather than
"authors" (some collisions are two repos under one owner). The `#1539` tests that pinned
the old wording are updated to pin the corrected claim, not loosened.

The `#1539` gate-round-2 tests used `mohammadaboulela/ComfyUI-BiRefNet` to exercise the
disclosure — that exact call is now refused, so those tests moved to `hieuck/…`, the author
`default` actually carries: same colliding name, allowed through on the from-source route
because there is nothing to pick between, and still warned.


artokun:
### Gate finding still open at `d05ee5c`: an unknown `channel` can throw instead of being ignored

Ran the codex gate on this branch (round 1 at `c002262`). It returned two P1s. One of them —
the refusal recommending the bypass it had just ruled out — is what `d05ee5c` ("gate round 2")
fixes. **The other is still present at `d05ee5c`.**

`bareNameAmbiguity` indexes two object literals with caller-supplied strings — the bare name off
the URL, and the `channel` argument. Object literals inherit from `Object.prototype`, so a plain
`obj[key]` answers for inherited keys. `entry["__proto__"]` hands back `Object.prototype`, whose
`.length` is `undefined` — so neither `!channelCandidates` nor `channelCandidates.length === 0`
stops it, and the inherited object rides into `listRepos`, where `.map` is not a function.

Measured against `d05ee5c`, not inferred (`repository:"https://github.com/mohammadaboulela/ComfyUI-BiRefNet"`):

```
channel="__proto__"     -> THREW TypeError: repos.map is not a function
channel="constructor"   -> THREW TypeError: Cannot read properties of undefined (reading 'toLowerCase')
channel="toString"      -> undefined (not checked)   <- correct
channel="nosuchchannel" -> undefined (not checked)   <- correct
channel="dev"           -> undefined (not checked)   <- correct, dev carries the caller's repo
```

`constructor` is the worse of the two: `Object.length` is `1`, so it passes the `length === 1`
branch and reaches `sameRepo(undefined, …)`.

This is **not** a security claim — one machine, one trust domain. It is a correctness gap against
this file's own stated contract: the test at `install-node-ambiguous-name.test.ts` asserts it
"tolerates a channel name it has never heard of rather than throwing", and for these three names
it does not. `panel_install_node` throws instead of dispatching or refusing.

An own-property check is the whole fix, and it restores exactly that promise:

```ts
function ownCandidates<T>(obj: Readonly<Record<string, T>>, key: string): T | undefined {
  return Object.prototype.hasOwnProperty.call(obj, key) ? obj[key] : undefined;
}
```

applied at the three lookups in `bareNameAmbiguity` — `AMBIGUOUS_BARE_NAMES[…]`, `entry[asked]`,
and `entry[c]` inside the `channelsResolvingToCaller` filter.

Verified separately on this branch while reviewing it, all against `c002262`:

| check | result |
|---|---|
| snapshot regenerated from live upstream | **byte-identical** — 111 names, 28 intra-channel |
| `load_nightly` / `NormalizedKeyDict` / `install_by_id` at tag `4.2.2` | quoted mechanism confirmed, incl. `repo_url = the_node['repository']` |
| `managerBareName` vs the panel's `gitRepoName` @ `origin/main` | character-for-character identical |
| remove the refusal | 12 tests fail |
| remove `if (conflict) return fail(conflict)` | 1 test fails, the wiring test |
| case-sensitive name lookup | 14 tests fail |
| host-blind repo comparison | 1 test fails, the off-host test |
| full suite | 516 files, 9659 passed |

The guard's premise and its snapshot both hold up; this is the one thing I found that `d05ee5c`
does not already cover.

artokun:
### Gate round 3 at `8f2cca4`: NO-SHIP — the explicit-channel bypass still dispatches a coin flip

`8f2cca4` closes the inherited-key P1. Re-gating that head returned **NO-SHIP** with a new P1,
and it reproduces.

For the **28 names that are ambiguous inside ONE channel**, naming a channel cannot disambiguate
anything — both candidates are in that same list. But an explicit `channel` skips the refusal and
only warns, so two requests for *different* repositories dispatch as the *same* Manager target.

Measured against `8f2cca4`:

```
snapshot: comfyui-pollinations -> default: ["1038lab/ComfyUI-Pollinations",
                                            "ciga2011/ComfyUI-Pollinations"]

A defaulted channel   refused=YES
B defaulted channel   refused=YES
A channel:"default"   refused=no  version="nightly" channel="default" repository=.../1038lab/...
B channel:"default"   refused=no  version="nightly" channel="default" repository=.../ciga2011/...
```

The differing `repository` is **inert** on this path — that is this PR's own thesis. The panel
sends `id = gitRepoName(repository)` = `ComfyUI-Pollinations` for both, `install_by_id` resolves
that single key against `default`, whose two entries collapse, and upstream list order picks the
winner. So at least one of those two calls installs the other author's code and reports success.

Why the existing escape hatch does not cover it: "a caller who named the channel made the choice
themselves" is true for a **cross-channel** collision, where naming `dev` really does select a
repository. It is a fiction for an **intra-channel** one — there is no argument that selects
either candidate. `d05ee5c` already established this in prose ("for all 65 repositories that sit
inside a multi-candidate channel, NO channel resolves them uniquely") and fixed the refusal
*text*; the dispatch *path* still takes the bypass.

The shape that follows from the PR's own rule — name both, never pick — is to refuse the
intra-channel case regardless of `channel`, and point at `install_custom_node (source:"git")`,
which clones the URL directly instead of resolving a name. That is already the remedy the refusal
prints when no channel resolves to the caller's repo, so it is a reachability fix, not a new
mechanism.

Not filed as a blocker on the round-4 work in flight (the `cnr_install` / registry-id route) —
that is a different finding on a different path.

artokun:
### Correction: my round-3 NO-SHIP above is a FALSE POSITIVE — withdrawing it

The [previous comment](https://github.com/artokun/comfyui-mcp/pull/1619#issuecomment-5303319214)
called the explicit-channel dispatch of an intra-channel collision a P1. **It is not. Disregard
it.** The measurement behind it was incomplete and I want that on the record before it steers a
change to code that is already correct.

**What I got wrong.** My probe printed `conflict`, `id`, `version`, `channel` and `repository`
and concluded "two repositories, one install target, undisclosed". It never printed `note` —
which is exactly where this PR puts the disclosure. Measuring the dispatch fields and not the
message is how a fully-disclosed decision reads as a silent one.

**What `da58540` actually returns** for `{repository:"…/1038lab/ComfyUI-Pollinations",
channel:"default"}`:

> NAMED-CHANNEL COLLISION: you asked for `https://github.com/1038lab/ComfyUI-Pollinations`, but
> the "default" channel you named carries "ComfyUI-Pollinations" TWICE — under
> `…/1038lab/ComfyUI-Pollinations` and `…/ciga2011/ComfyUI-Pollinations` — and v4 keeps whichever
> comes last in that list. […] **Naming that channel did NOT disambiguate this one** — both
> entries are in the list you named, so no `channel` argument can separate them and on v4 the
> winner is that list's order. Dispatched because you named a channel, and because on Manager 3.x
> the URL you passed IS what gets cloned; if this is v4 and you need the repository you typed,
> `install_custom_node (source:"git")` clones it directly instead of resolving a name.

Both candidates named; the caller told plainly that their channel selected nothing; the reason it
still dispatches (3.x, where the URL *is* cloned as passed) stated; the working remedy given.
That is not picking between two repositories on the caller's behalf.

**And refusing here would be wrong**, which I had not weighed. #1616 specifies the escape as
"an unnecessary refusal that an explicit `channel` clears". Removing it for these 28 names would
brick `panel_install_node` for every Manager 3.x host carrying one — where the install is
*correct* — with no argument to get past it. `d05ee5c`'s test says as much in a comment
(`// still dispatches — the escape must survive`); I built a fix for ~30 minutes before reading
the test that already answered the question.

I built and then discarded that change. Nothing was pushed; the branch is untouched by it.

**Standing conclusion:** rounds 1 and 3 produced four findings. Three were real and are fixed
(`c002262` host + exact-case + 3.x disclosure, `d05ee5c` bypass contradiction, `8f2cca4`
inherited-key throw). This fourth one is a false positive against a deliberate, disclosed,
issue-sanctioned trade-off — consistent with gate rounds past ~3 starting to report design as
defect. It should not block this PR.

artokun:
### Gate on `da58540`: NO-SHIP — 1 confirmed, 1 mis-exampled but real in class, 2 to disregard

Gated the pushed head on a clean worktree. Four P1s came back; they are not equal, and one of
them is the same false positive I withdrew above. Triaged so the next round is not spent on it.

**CONFIRMED — round 4's registry guard is keyed on the wrong table.**
`registryVersionAmbiguity` decides the CNR route by consulting `AMBIGUOUS_BARE_NAMES`, which
measures *channel* collisions. The CNR route never reads a channel — that is round 4's own
finding — so what governs it is **Registry ID → repository**, an independent dataset. Names that
collide there but not across channels are unguarded:

```
snapshot:  comfyui_nnlatentupscale                -> ABSENT
registry:  comfyui_nnlatentupscale                -> https://github.com/Goktug/ComfyUi_NNLatentUpscale
call:      {repository:"…/Ttl/ComfyUi_NNLatentUpscale", version:"1.0.0"}   -> no conflict
```

Caller names `Ttl`, gets `Goktug`, reported as success — the exact class `da58540` set out to
close, through the table it chose to close it with.

**MIS-EXAMPLED, right class.** The CNR-re-key finding cites `ComfyUI-ChatterboxTTS`; there is no
such Registry ID (`api.comfy.org/nodes/comfyui-chatterboxtts` 404s), so that example does not
stand. The underlying point does: `comfyui-chatterbox` is `wildminder`'s on `default` and
`ai-joe-git`'s on `dev` in the snapshot, while the **Registry** id resolves to `sm079/comfyui-chatterbox`
— a third party the table never names. Same gap as above.

**DISREGARD — the explicit-channel findings (`node-management.ts:4562`, and `:4555` for the
registry route).** `{repository:"…/mohammadaboulela/ComfyUI-BiRefNet", channel:"legacy"}`
dispatching is the escape hatch #1616 specifies ("an unnecessary refusal that an explicit
`channel` clears"), with both candidates named in the note and `install_custom_node (source:"git")`
offered. I filed that same finding one round ago and
[withdrew it](https://github.com/artokun/comfyui-mcp/pull/1619#issuecomment-5303342332) after
measuring the note instead of only the dispatch fields. Refusing there would brick
`panel_install_node` for every Manager 3.x host carrying one of these names, where the install is
correct. The gate is now re-reporting design as defect — the expected failure mode past round 3.

**On scope.** Closing the confirmed finding properly needs a Registry-ID→repository dataset that
does not exist in this PR — a second snapshot with its own staleness story, and its own generator.
That is real scope growth for a change already five rounds deep during a stabilisation freeze.
Worth deciding deliberately rather than by momentum: either land the from-source guard (rounds 1–3,
which are measured, self-contained and verified) and split the CNR route into its own issue, or
accept the second dataset knowingly. The CNR route's `repository`-is-ignored property is
pre-existing and already disclosed in the dispatch note — it is not a regression this PR introduces.

CI is green on `da58540` across all three platforms. Not merged: the gate says NO-SHIP and the
first finding above is real.

artokun:
## Gate rounds 6–9

Re-gated four more times after the five recorded above. Rounds 6, 7 and 8 each found a real P1; all three are fixed and pinned by mutation-verified tests. Round 9 found no new defect, and this is where the gate stops.

### Round 6 — a channel MISS is the CNR fallback, not a dead end

`bareNameAmbiguity` returned "no ambiguity" whenever the asked channel carried nothing under the name, reasoning that Manager then answers "not found". 4.2.2 says otherwise — `install_by_id`'s nightly branch tries `cnr_map[node_id]` first and clones the **registry's** repository; "not found" is only what happens when the registry lacks the id too. **18 of the 111 contested names are absent from `default`**, so on the defaulted channel they reached that fallback unchecked.

### Rounds 7–8 — the exact-match exemption was unsound, and it was measurable

`get_custom_nodes` keys a channel entry by its **CNR id** when the repo is registered, so a re-keyed entry is unreachable by the bare name however plainly the list carries it. Round 7 documented this and filed #1624; that was the wrong call, because the re-keying can be measured. Round 8 measures it (249 candidates against api.comfy.org):

* **32 candidates across 30 names are re-keyed** — independently reproducing the "32 of 249" this PR had measured by hand;
* **75 names have a registry pack of their own name**, so the refusal now names the repository that would actually land instead of listing suspects.

**It caught this PR's own "safe" example.** `hieuck/ComfyUI-BiRefNet` was the allowed case throughout, because `default` lists it — but that repo is registered as id `viperyl_ComfyUI-BiRefNet`, so the bare name misses and `cnr_map` resolves to **viperyl's**. The install this PR called safe was itself a substitution. Fixtures moved to `ComfyUI_TiledKSampler` (BlenderNeko/`default`, ltdrdata/`forked`, neither re-keyed).

Not a blanket refusal: **74 of 111** names still have exactly one reachable `default` entry and still dispatch; 15 newly refuse.

### Round 9 — no new defect, and why it is not actioned

Two P1s, both verified as pre-existing documented positions rather than defects:

1. **pip Manager reads its bundled list.** `{repository:"…/SiegeKeebsOffical/comfyui-lmstudio"}` dispatches; on a pip host reading the stale *bundled* snapshot the entry may be absent and the fallback clones `Winnougan/comfyui-lmstudio`. That is the "A pip-installed Manager" limitation stated in the module header. It is not observable from here — the orchestrator cannot tell which Manager the *panel's* ComfyUI runs (the same reason the 3.x false positive is undetectable, gate round 1), and the bundled list varies by Manager version. `main` dispatches this with no check at all.
2. **The explicit-`channel` bypass on an intra-channel collision.** `{repository:"…/neverbiasu/ComfyUI-BAGEL", channel:"default"}` dispatches. Verified: **without** a channel the same call is REFUSED, and **with** one the note says in as many words that naming the channel *"did NOT disambiguate this one"*. This is the finding rounds 1 and 3 raised and this PR records under "One finding deliberately not taken" — removing the bypass strands **65 repositories** with no alternative channel (measured) and breaks them on Manager 3.x, where the URL ships in `files` and the install is correct.

Actioning (2) would mean making the tool worse to satisfy the gate, so it stands. #1624 remains open for the wider registry-substitution class, which needs no channel collision at all.

CI green on all four jobs. Full suite **9676 passed, 516 files**; lint, `check:vocabulary`, `check:unknown-collapse`, build all green.

PR #1625, opened 2026-08-15
fix(870): apply_manifest stops refusing junctioned model folders

### LINKED ISSUE #870: download_model refuses StabilityMatrix junction model folders (vae/diffusion_models)

artokun:
Fixes the half of #870 that is still live: `apply_manifest`.

## Status: complete and green — HOLDING on the codex gate

The codex gate returned **INDETERMINATE (exit 2)**, not a pass: `You've hit your usage limit […] try again at Aug 19th, 2026 8:43 PM` (confirmed by calling `codex exec` directly, not just through the wrapper's text match). Per the autopilot rule an indeterminate gate is never a pass, so this is **not merged**. It needs one gate run once quota returns.

## What was still broken

#919 fixed `download_model` by moving the authorization rule into the canonical resolver: an in-tree symlink's redirect is authorized by its **location** inside the user's own models tree, because ComfyUI's own scanner follows it (`folder_paths.recursive_search` walks with `followlinks=True`). Only a `custom_nodes` landing and a dangling/unresolvable link are still refused.

`apply_manifest` never got that fix, because it kept a **second copy of the old rule**. `resolveLocalModelPath` in `src/services/manifest.ts` realpath'd the destination's parent and refused anything landing outside the models root:

```ts
const realParent = await realpath(dirname(targetPath));
if (!isWithinRoot(realRoot, realParent)) throw ...escapes the models directory
```

On a StabilityMatrix install `models/vae` and `models/diffusion_models` are NTFS junctions into shared storage, so `realpath` of the parent **is** outside the root and every model under them was refused before a byte moved — while `models/text_encoders`, a real directory, sailed through. That is the exact some-categories-work asymmetry #870 reports, and it is why the reporter still hits it on 0.51.56 via the bundled `krea2-txt2img-manual` pack (its three models are `diffusion_models/`, `text_encoders/`, `vae/`).

## Verified on a real junction, not a mock

Built `models/vae` as an actual `mklink /J` junction to `shared/VAE` and replayed the pre-fix checks:

```
--- vae/qwen_image_vae.safetensors
  lstat(parent).isSymbolicLink() = true
  lstat(parent).isDirectory()    = false
  realpath(parent)               = ...\shared\VAE
  isWithinRoot(realRoot, realParent) = false
  PRE-FIX  VERDICT: REFUSED -> "Model local_path escapes the models directory: vae/..."
  POST-FIX VERDICT: accepted (passed to the downloader)
--- text_encoders/qwen3vl.safetensors        (a real directory, side by side)
  PRE-FIX  VERDICT: accepted
--- ../evil.safetensors
  POST-FIX VERDICT: REFUSED (lexical escape)
```

## The guard is not lost, it is de-duplicated

Every model this function resolves is downloaded through `startDownloadJob` → `resolveDownloadTarget` → `resolveModelSubfolderWithLiveRoot`, which runs `assertNoEscapingSymlinkAncestor` on the same destination before any write — with the `custom_nodes` code-root veto, the dangling-symlink refusal, and the base dirs + live snapshot from the **same** `/system_stats` call that resolved the root (context the manifest layer does not have). `model-resolver.ts` says so at the guard itself: it lives there *"so it can never diverge from a caller's separate pre-validation (e.g. apply_manifest resolving the root a second time)"*. Duplicating it is what let the two drift. Its behaviour is covered by `download-dest-roots.test.ts`, including `REFUSES a symlink into a registered custom_nodes root` and `REFUSES when a symlink ancestor exists but its target cannot be resolved (dangling)`.

What `manifest.ts` still enforces is the part that is purely about the manifest's own input and needs no filesystem: an absolute `local_path`, a `..` traversal out of `models/`, a path with no filename.

## One deliberate removal

The removed ancestor walk also carried a `parent is not a directory` check — no test, no issue behind it. Dropped rather than kept, because `lstat` on a junction reports `isDirectory() === false` (see the probe above): keeping it would have re-broken the same installs through a different message. A parent that really is a file now surfaces as the downloader's own `ENOTDIR` on that item.

## Tests

- The two cases that pinned the superseded rule are **inverted** to assert the new behaviour (a junctioned category folder reaches the downloader).
- The krea2 pack's three models are pinned to behave **identically** whether their folder is a junction or real — the reporter's asymmetry as a test.
- New: the lexical `..` traversal refusal, and #490's live-root targeting (which the old test only covered incidentally through the symlink guard).
- **Mutation-checked**: reverting `src/services/manifest.ts` to `origin/main` fails exactly the two #870 tests and nothing else.
- Full suite: 516 files / 9682 passed, 3 skipped. `tsc --noEmit` clean.

## Not done here

The `download_model` route named in the issue title was already fixed by #919 (v0.50.x) and is covered by `download-dest-roots.test.ts` `(b) #870 — a StabilityMatrix junction into UNDECLARED storage is allowed`. This PR does not touch it.

closes #870

### DIFF
--- a/src/__tests__/services/manifest.test.ts
+++ b/src/__tests__/services/manifest.test.ts
@@ -1540,60 +1540,113 @@ describe("applyManifest", () => {
     );
   });
 
-  it("rejects model local_path when a symlinked parent escapes models", async () => {
+  it("#870 — a junctioned category folder is NOT refused up front (StabilityMatrix)", async () => {
+    // The reporter's install junctions models/vae -> E:\comfy\StabilityMatrix\
+    // models\VAE. `realpath` of the parent therefore lands OUTSIDE the models
+    // root, which apply_manifest used to read as "local_path escapes the models
+    // directory" and refuse before any download started.
+    //
+    // Since #919 the canonical resolver authorizes an in-tree redirect by its
+    // LOCATION (ComfyUI's own scanner follows the junction), so this layer must
+    // not veto it: the item has to reach the downloader, which applies the one
+    // remaining refusal (a custom_nodes landing) on the same path.
+    //
     // The product resolves these paths with node:path, yielding
     // backslash-separated absolute paths on Windows. Build the mock keys the
     // same way so they match what the product passes to realpath.
     const modelsDir = resolve(COMFY, "models");
-    const linkDir = join(modelsDir, "link");
-    const outside = resolve("/tmp/outside");
+    const junction = join(modelsDir, "vae");
+    const shared = resolve("E:/comfy/StabilityMatrix/models/VAE");
     realpathMock.mockImplementation((path: string) => {
       if (path === modelsDir) return Promise.resolve(modelsDir);
-      if (path === linkDir) return Promise.resolve(outside);
+      if (path === junction) return Promise.resolve(shared);
       return Promise.resolve(path);
     });
 
     const result = await applyManifest({
       manifest: {
-        models: [{ url: "https://example.com/model.safetensors", local_path: "link/model.safetensors" }],
+        models: [
+          { url: "https://example.com/qwen_image_vae.safetensors", local_path: "vae/qwen_image_vae.safetensors" },
+        ],
       },
     });
 
-    expect(result.success).toBe(false);
     expect(result.results).toMatchObject([
-      { action: "model", status: "failed", item: "link/model.safetensors" },
+      { action: "model", item: "vae/qwen_image_vae.safetensors" },
     ]);
-    expect(downloadModelMock).not.toHaveBeenCalled();
+    expect(result.results[0].status).not.toBe("failed");
+    expect(downloadModelMock.mock.calls[0]?.slice(0, 3)).toEqual([
+      "https://example.com/qwen_image_vae.safetensors",
+      "vae",
+      "qwen_image_vae.safetensors",
+    ]);
   });
 
-  it("validates local_path symlinks against the LIVE models dir, not a stale COMFYUI_PATH (#490)", async () => {
-    // #490: COMFYUI_PATH is a stale install; the connected server writes under a
-    // DIFFERENT live root. A symlink that escapes the LIVE models dir must be
-    // caught — the guard has to run against the live write root the downloader
-    // uses, NOT <COMFYUI_PATH>/models (which is never written to here).
-    mockConfig.comfyuiPath = "/stale/ComfyUI";
-    const liveModels = resolve("/live/ComfyUI/models");
-    modelsDirMock.mockResolvedValue(liveModels);
-    const linkDir = join(liveModels, "link");
-    const outside = resolve("/tmp/escape");
-    realpathMock.mockImplementation((path: string) => {
-      if (path === liveModels) return Promise.resolve(liveModels);
-      if (path === linkDir) return Promise.resolve(outside);
-      return Promise.resolve(path);
-    });
+  it("#870 — junctioned and real category folders behave IDENTICALLY (the krea2 pack's three models)", async () => {
+    // The reporter's asymmetry: text_encoders is a real directory under the
+    // package models root and succeeded, while vae and diffusion_models are
+    // junctions and were both refused. All three must now enqueue.
+    const modelsDir = resolve(COMFY, "models");
+    const junctions = new Map([
+      [join(modelsDir, "vae"), resolve("E:/comfy/StabilityMatrix/models/VAE")],
+      [join(modelsDir, "diffusion_models"), resolve("E:/comfy/StabilityMatrix/models/DiffusionModels")],
+    ]);
+    realpathMock.mockImplementation((path: string) =>
+      Promise.resolve(junctions.get(path) ?? path),
+    );
 
     const result = await applyManifest({
       manifest: {
-        models: [{ url: "https://example.com/m.safetensors", local_path: "link/m.safetensors" }],
+        models: [
+          { url: "https://example.com/krea2_turbo_fp8.safetensors", local_path: "diffusion_models/krea2_turbo_fp8.safetensors" },
+          { url: "https://example.com/qwen3vl_4b_fp8_scaled.safetensors", local_path: "text_encoders/qwen3vl_4b_fp8_scaled.safetensors" },
+          { url: "https://example.com/qwen_image_vae.safetensors", local_path: "vae/qwen_image_vae.safetensors" },
+        ],
+      },
+    });
+
+    expect(result.results.filter((r) => r.status === "failed")).toEqual([]);
+    expect(downloadModelMock.mock.calls.map((c) => c[1]).sort()).toEqual([
+      "diffusion_models",
+      "text_encoders",
+      "vae",
+    ]);
+  });
+
+  it("still refuses a local_path that traverses out of models/ lexically", async () => {
+    const result = await applyManifest({
+      manifest: {
+        models: [{ url: "https://example.com/m.safetensors", local_path: "../evil.safetensors" }],
       },
     });
 
     expect(result.success).toBe(false);
     expect(result.results).toMatchObject([
-      { action: "model", status: "failed", item: "link/m.safetensors" },
+      { action: "model", status: "failed", item: "../evil.safetensors" },
     ]);
     expect(result.results[0].message).toMatch(/escapes the models directory/i);
     expect(downloadModelMock).not.toHaveBeenCalled();
+  });
+
+  it("targets the LIVE models dir, not a stale COMFYUI_PATH (#490)", async () => {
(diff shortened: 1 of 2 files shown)


artokun:
## Re-gate attempt 2026-08-15 — still INDETERMINATE, still holding

Ran the gate again against the current head (`dddc4ba`):

```
node .claude/autopilot/codex-gate.mjs --worktree ...870-manifest-junction --base origin/main --name 1625
GATE_EXIT=2
```

Confirmed directly rather than trusting the wrapper's text match — `codex exec` on a
one-line probe returns the same thing:

```
ERROR: You've hit your usage limit. ... try again at Aug 19th, 2026 8:43 PM.
```

No API-key fallback exists on this machine (`OPENAI_API_KEY` unset; `~/.codex/auth.json`
is plan-based), and the gate script has no alternate-model path. **Exit 2 is not a pass,
so this stays open.** Earliest it can be gated is Aug 19 ~20:43.

### What IS independently verified (not re-derived from the PR body)

**The mutation check reproduces the reporter's bug, not just a test failure.** Reverting
only `src/services/manifest.ts` to `origin/main` and re-running the file fails **exactly 2
of 59** tests, with the issue's own error string — and note which model is missing from
the failure list:

```
+   { "item": "diffusion_models/krea2_turbo_fp8.safetensors",
+     "message": "Model local_path escapes the models directory: ...", "status": "failed" },
+   { "item": "vae/qwen_image_vae.safetensors",
+     "message": "Model local_path escapes the models directory: ...", "status": "failed" },
```

`text_encoders/qwen3vl_4b_fp8_scaled.safetensors` passes in the same run. That is #870's
some-categories-work asymmetry, reproduced as a failing test and fixed by this diff.

**The removed guard is genuinely downstream, traced in source rather than assumed.**
`manifest.ts:1357` → `startDownloadJob` → (local route, `download-jobs.ts:448`)
`resolveDownloadTarget` → `model-resolver.ts:2195` `resolveModelSubfolderWithLiveRoot` →
`model-resolver.ts:1921` `assertNoEscapingSymlinkAncestor`, unconditional on that path.
The Manager route skips it because it resolves no local destination at all — there is no
local write to guard. The dropped `mkdir` calls are not load-bearing either: `downloadModel`
does its own `mkdir(targetDir, { recursive: true })` at `model-resolver.ts:3150`.

**CI green is not a stale green.** The passing run (`31905926463`, `npm test` on
ubuntu/windows/macos) has `headSha = dddc4bab9b741d4976d53cf1d3c5c3d469a41912`, which is
the current head, and `merge-base(HEAD, origin/main) == origin/main` — zero commits have
landed on main since the branch point.

Remaining work is one gate run. Nothing else is outstanding.

artokun:
Picked this up to drive to a terminal state. It does not get one: **the codex gate is still INDETERMINATE, so this holds unmerged.** Re-measured just now rather than trusting the earlier reading:

```
node .claude/autopilot/codex-gate.mjs --worktree ...\870-manifest-junction --base origin/main --name 1625
GATE_EXIT=2
```

The gate launched codex successfully and codex itself returned the account-level refusal (the `usage limit` match at `codex-gate.mjs:139`), so this is the account, not the wrapper. Reset is 2026-08-19 8:43 PM. An indeterminate gate is never a pass.

Everything else I could verify independently, I did:

- **The mutation check reproduces.** Reverting only `src/services/manifest.ts` to `origin/main` fails exactly 2 of 59 tests — both #870 tests, nothing else — with the reporter's exact message and exact asymmetry: `diffusion_models/` and `vae/` come back `Model local_path escapes the models directory`, while `text_encoders/` passes. Restored, worktree clean.
- **The de-duplication claim holds structurally.** `applyManifest` → `startDownloadJob` (`download-jobs.ts:448`) → `resolveDownloadTarget` (`model-resolver.ts:2195`) → `resolveModelSubfolderWithLiveRoot` → `assertNoEscapingSymlinkAncestor` (`model-resolver.ts:1921`). The `custom_nodes` veto and the dangling-symlink refusal are pinned downstream in `download-dest-roots.test.ts:313` and `:399`.
- **One thing the description does not address, now checked:** the removal also drops `rejectEscapingSymlinkTarget`, which refused a destination *file* that was itself an escaping symlink — and the downstream ancestor walk covers the directory, not the file. It opens no escape: the writer is an O_EXCL temp + atomic rename (`model-resolver.ts:3130`), and a rename *replaces* a destination symlink rather than following it. `download_model` has never had that check either, so the two routes are now identical rather than manifest being the weaker one.
- **CI green and not stale.** `origin/main` (`aa05799`) is an ancestor of this branch, so the green ran against current main. `npm test` runs the full suite on all three platforms.

Nothing here needs changing. It needs one gate run once quota returns — re-run the command above, and merge on exit 0.

artokun:
**Merge gate — substitution disclosed.**

`codex exec` was unavailable for this merge (the account is usage-exhausted until 2026-08-19 20:43), and the `claude-gate.mjs` fallback spawns a headless `claude` that is blocked on the account weekly limit. Under the owner's standing authorisation for a substitute gate, this PR was gated by an **independent adversarial reviewer that had never seen the author's reasoning**, run against `.claude/autopilot/review-taxonomy.md`. Disclosure is the condition that substitution was accepted under.

**Verdict: SHIP.** What the gate actually verified, rather than assumed:

- **Reachability.** `resolveLocalModelPath` (`src/services/manifest.ts:660-713`) is reached from its sole caller at `src/services/manifest.ts:1215` whenever `hasLocalFs` is true; the Manager branch at :1170 short-circuits before it and is unchanged. The reporter's exact case (`packs/krea2-txt2img-manual/manifest.yaml`, all three models on `local_path:`) lands on the touched branch.
- **The removed ~94 lines are de-duplicated, not lost.** `startDownloadJob` (`src/services/download-jobs.ts:448`) still reaches `assertNoEscapingSymlinkAncestor` via `resolveDownloadTarget` → `resolveModelSubfolderWithLiveRoot` on every local route before any write — still refusing a dangling symlink, a `custom_nodes` landing, a models root resolving into code, and failing closed on non-ENOENT `lstat`.
- **Three independent mutation kills** (scratch worktree): reverting `manifest.ts` to `origin/main` fails exactly the two new #870 tests and nothing else; swapping `resolveModelsDir()` for `COMFYUI_PATH/models` fails the #490 test; deleting both lexical-escape checks fails the `../evil.safetensors` test.
- **Old refusals enumerated case by case** — absolute path, lexical `..`, empty/`.`/`..` filename all still refused. A symlinked ancestor escaping root is deliberately allowed, matching the `download_model` policy already shipped in #919. An ancestor-that-is-a-file now surfaces as the writer's `ENOTDIR` (still a `failed` report, rawer message).
- 325 tests green across `manifest`, `download-dest-roots`, `apply-manifest-pack`, `model-resolver`, `download-jobs`, `download-live-destination`, `download-root-correspondence`; `tsc --noEmit` clean.

Two non-blocking nits recorded, neither a wrong-output scenario: the comment at `src/services/manifest.ts:696` overstates that *every* resolved model goes through `resolveDownloadTarget` (`download-jobs.ts:447` skips it when `dispatchToManager` is true — the safety guarantee holds, the sentence misstates the mechanism); and the two new tests still install a `realpathMock` that nothing calls post-fix.

PR #1627, opened 2026-08-15
fix(1147): a tree the server demonstrably reads is not refused over one stray file

### LINKED ISSUE #1147: download_model download_civitai false-positive path refuse on birefnet nested HF cache
`download_model action:download_civitai` refused to write into the correct local models dir:

`Refusing to download into ...\models: the connected ComfyUI does not read from it. Its \"birefnet\" folder holds 1 model file(s), 1 of which the running server does NOT list (e.g. BiRefNet_lite/model.safetensors)`

Root cause: path-identity probe treats nested `models/birefnet/BiRefNet_lite/model.safetensors` (a HuggingFace-style full repo dump) as a ComfyUI-listed model file. The running server's `/models/birefnet` is empty (folder not registered / not scanned the same way), so the probe concludes wrong install — even though `/models/loras` and workspace match the live server and process argv is `ComfyUI\main.py --windows-standalone-build` under that same tree.

### Steps to reproduce
1. Local portable ComfyUI with `models/birefnet/BiRefNet_lite/model.safetensors` present from an HF clone.
2. Workspace/default points at that install; API `http://127.0.0.1:8188`.
3. Call `download_model { action: download_civitai, model_id, model_version_id, target_subfolder: loras }`.
4. Observe MODEL_ERROR refuse despite correct install.

### Exact error (scrubbed)
MODEL_ERROR: Refusing to download into \"Z:\\...\\ComfyUI\\models\": ... birefnet ... BiRefNet_lite/model.safetensors ... DIFFERENT install ...

### Fix
applied locally: no (shell download workaround used)
upstream-only: yes

Suggested fix: (1) only count top-level / known-extension files that match how ComfyUI registers models; (2) ignore nested HF cache layout under birefnet; (3) treat empty/missing model-folder API category as non-disproof when other categories (loras) match; (4) prefer process cwd/argv install root when available.

### Environment
OS Windows 11 / ComfyUI 0.30.2 / frontend 1.47.12 / RTX 5060 Ti 16GB / comfyui-mcp 0.50.40 / panel 0.11.41

artokun:
Reopened on **0.51.56**: a ComfyUI Desktop install was refused an `animatediff_models` download because ONE unlisted file under `ipadapter` was read as proof the running server reads a DIFFERENT install — while that same server listed other categories of that same tree, file for file.

## The defect

The pre-write guard gathered positive evidence in the REFUSE direction and threw away the positive evidence it had already computed in the PROCEED direction. `missing.length === 0` was a bare `continue`, and the first disagreement `break`ed before any later category could speak — so a category the server accounts for completely never got to answer, and a single unaccounted file decided the root.

The two facts are not symmetric:

- A category the server lists **COMPLETELY** is evidence about the **ROOT**. For a different install to produce it, every name in that category would have to coincide.
- A file the server does not name is evidence about that **FILE**. It has ordinary explanations that say nothing about the root — an extension not registered for that category, a listing cached before the file arrived, a nested HuggingFace dump (this issue's original `birefnet` report), a folder a custom node registers against a different path.

## The change

The scan no longer stops at the first disagreement, and a disagreement is overridden by a fully-agreeing category holding **more than one** file. One file is not enough, and the threshold is pinned by a test: two unrelated installs really do share `sd_xl_base_1.0.safetensors`, and a category holding only that file is "fully accounted for" by coincidence (codex gate, round 5).

| tree | live listing | before | after |
|---|---|---|---|
| `ipadapter` 1 stray + `loras` 2 files | `loras` both listed | refuse | **proceed** |
| 3 files, accounted for nowhere | 24 unrelated files | refuse | refuse (#369, unchanged) |
| 1 shared popular checkpoint | that name + others | refuse | refuse (round 5, unchanged) |

## Why it came back after 0.50.44

#1168 only made an **EMPTY** listing non-evidence. This reporter's `ipadapter` listing is **POPULATED**, so that fix could not reach it — and #1562 (v0.51.44) widened the guard to run for **inferred** roots too, which is how a correct Desktop install started being probed at all.

## Verification

- Bug stated as failing tests first (2 red), then fixed.
- Mutation-tested, 3 applied and 3 killed against a green baseline: dropping the override → 2 fail; restoring the early `break` → 2 fail; accepting a one-file coincidence → 1 fail (the round-5 hole).
- Reachability checked rather than assumed: `download_model` reaches this guard via `resolveDownloadTarget` → `resolveModelSubfolderWithLiveRoot`, on both the direct writer and the background job registry paths.
- Full suite: **516 files, 9683 passed**, 3 skipped. `tsc --noEmit` clean.

Fixes #1147

### DIFF
--- a/src/__tests__/services/download-live-destination.test.ts
+++ b/src/__tests__/services/download-live-destination.test.ts
@@ -297,6 +297,69 @@ describe("pre-write: a destination the LIVE server does not read from is refused
     );
   });
 
+  // #1147 (recurrence, 0.51.56) — CORROBORATED READERSHIP OUTWEIGHS ONE STRAY FILE.
+  //
+  // A reporter's ComfyUI Desktop install was refused an `animatediff_models`
+  // download because ONE unlisted file in `ipadapter` was read as proof the server
+  // reads a different install — while other categories of that same tree were
+  // listed by that same server, file for file.
+  //
+  // Those two facts are not symmetric. A category the server lists COMPLETELY is
+  // positive evidence it scans this root: for a wrong install to produce that, every
+  // name in the category would have to coincide. A single file the server does not
+  // name has ordinary explanations that say nothing about the root — an extension not
+  // registered for that category, a listing cached before the file arrived, a nested
+  // HuggingFace dump, a folder a custom node registers elsewhere. The #369 signature
+  // is a tree the server accounts for NOWHERE, so corroborated readership must win.
+  it("#1147: PROCEEDS when another category is FULLY accounted for by the same server", async () => {
+    h.onDisk = {
+      // Alphabetically first, so the contradiction is found BEFORE the corroboration
+      // and the scan must not stop at it.
+      animatediff_models: [],
+      ipadapter: ["ip-adapter_sd15.safetensors", "stale/pytorch_model.bin"],
+      loras: ["a.safetensors", "b.safetensors"],
+    };
+    h.liveListings["ipadapter"] = ["ip-adapter_sd15.safetensors"];
+    h.liveListings["loras"] = ["a.safetensors", "b.safetensors"];
+
+    await expect(resolveModelSubfolderPreferServer("animatediff_models")).resolves.toBe(
+      resolve("/comfy/models/animatediff_models"),
+    );
+  });
+
+  it("#1147: a ONE-FILE coincidence does NOT corroborate — the refusal stands (codex gate r5)", async () => {
+    // The round-5 hole at category granularity: two unrelated installs routinely
+    // share ONE popular checkpoint, and a category that holds only that file is
+    // "fully accounted for" by pure coincidence. Corroboration needs a category
+    // whose every name coincides AND that holds more than one file.
+    h.onDisk = {
+      checkpoints: ["sd_xl_base_1.0.safetensors"],
+      loras: ["stale-a.safetensors", "stale-b.safetensors"],
+    };
+    h.liveListings["checkpoints"] = ["sd_xl_base_1.0.safetensors", "live-only.safetensors"];
+    h.liveListings["loras"] = ["live-lora.safetensors"];
+
+    await expect(resolveModelSubfolderPreferServer("loras")).rejects.toThrow(
+      /DIFFERENT install/,
+    );
+  });
+
+  it("#1147: an INFERRED root is corroborated the same way (#1562 stays fixed either way)", async () => {
+    // #1562 made `observed-root` run this check. A stale bundle reached through a
+    // python-on-PATH inference is accounted for NOWHERE, so it still refuses; a
+    // correct one that the server lists file-for-file must not be refused over a
+    // stray file.
+    h.modelsDirSource = "observed-root";
+    h.onDisk = {
+      ipadapter: ["listed.safetensors", "unlisted.bin"],
+      loras: ["a.safetensors", "b.safetensors"],
+    };
+    h.liveListings["ipadapter"] = ["listed.safetensors"];
+    h.liveListings["loras"] = ["a.safetensors", "b.safetensors"];
+
+    await expect(resolveModelSubfolderPreferServer("ipadapter")).resolves.toBeTruthy();
+  });
+
   // There is deliberately NO "the server lists models this tree cannot account for"
   // refusal. Not being able to EXPLAIN the server's models is absence of evidence,
   // not proof of a different install: the roots may come from a config we cannot
--- a/src/services/model-resolver.ts
+++ b/src/services/model-resolver.ts
@@ -1010,6 +1010,13 @@ async function coreModelFilesUnder(categoryDir: string): Promise<string[]> {
  *  per folder. */
 const DISAGREEMENT_PROBE_CATEGORIES = 6;
 
+/** How many model files a fully-agreeing category must hold before it CORROBORATES
+ *  that the live server reads this root (#1147). One is not enough: two unrelated
+ *  installs routinely share a single popular checkpoint, and a category holding
+ *  only that file is "fully accounted for" by coincidence (codex gate, round 5).
+ *  Two independent names both coinciding is not an ordinary explanation. */
+const CORROBORATING_CONTAINMENT_MIN_FILES = 2;
+
 /** Immediate subdirectories of the candidate models root, in a stable order. */
 async function categoriesUnder(modelsRoot: string): Promise<string[]> {
   try {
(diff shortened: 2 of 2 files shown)


artokun:
## ⛔ Merge HOLD — codex gate INDETERMINATE (not a pass)

```
node .claude/autopilot/codex-gate.mjs --worktree .claude/worktrees/wt-1147 --base origin/main --name 1147
exit 2
```

Confirmed directly rather than trusting the wrapper — `codex exec` on this account returns:

> ERROR: You've hit your usage limit. … try again at **Aug 19th, 2026 8:43 PM**.

So this is real quota exhaustion, not a sandbox or parse failure. Exit 2 is not a pass, and a self-review is never the gate for this repo, so I am **not** merging.

**State: complete and green, awaiting the gate only.**

- CI: Build macOS / Ubuntu / Windows **pass**, Pack & install smoke **pass**
- Full suite: 516 files, 9683 passed, 3 skipped
- `tsc --noEmit` clean
- Mutation: 3 applied, 3 killed against a green baseline

**What unblocks it:** codex quota returning (≈ Aug 19) or credits added, then re-run the gate command above. If it exits 0, this is mergeable as-is — no rebase needed unless `main` moves under `src/services/model-resolver.ts`, in which case re-check CI before merging.

**What a reviewer should push hardest on** (the one genuinely arguable call): `CORROBORATING_CONTAINMENT_MIN_FILES = 2`. It is the line between "coincidence" and "proof the server reads this root". One file is demonstrably too weak — that is codex gate round 5's shared `sd_xl_base_1.0.safetensors`, and a test pins it. Two is the point where every name in a category would have to coincide. The threshold is pinned in the widening direction by a mutation, but the *value* is a judgement call, not a measurement.

artokun:
## Independent verification — held at the merge gate

Re-verified this PR from scratch rather than trusting the description. Recording it here so the next pass does not repeat it.

**Merge gate: INDETERMINATE (exit 2). Not merged.**

```
node .claude/autopilot/codex-gate.mjs --worktree .claude/worktrees/wt-1147 --base origin/main --name 1627
```

Probed the CLI directly to pin the reason rather than infer it: `You've hit your usage limit … try again at Aug 19th, 2026 8:43 PM`. That is the gate being unable to speak, which is **not** a pass — the merge holds until it can run.

**What did verify.**

- **CI green on the current base.** `git rev-list --left-right --count origin/main...HEAD` → `0 1`, so the three-platform green is against today's `origin/main` (aa05799), not a stale one.
- **Reachable from production**, checked rather than assumed: `download-jobs.ts:448` → `resolveDownloadTarget` → `resolveModelSubfolderWithLiveRoot` → `model-resolver.ts:1982` → the guard. The direct writer reaches it at `model-resolver.ts:3216` through the same resolver.
- **Mutation-tested — the tests pin the fix.** Baseline 51/51 green; each mutation touches the source only.

| # | mutation | result |
|---|---|---|
| 1 | restore the early `break` at the first disagreement | **2 killed** |
| 2 | delete the `worst && corroborating` override | **2 killed** |
| 3 | `CORROBORATING_CONTAINMENT_MIN_FILES = 1` | **1 killed** |
| 4 | keep the *first* agreeing category instead of the largest | **survived** |

Mutation 4 survived legitimately: preferring the largest agreeing category changes only which category the log line names, never the verdict — any agreeing category with ≥2 files decides it. Untested cosmetics, not an untested branch. Left as-is rather than churn a green CI for it.

**The strongest argument against this diff, for whoever runs the gate.**

The 2-file threshold is thin in one specific shape. Corroboration only needs *some* category whose every on-disk file appears in the live listing, and small categories are often near-universal: a stale tree holding `vae/vae-ft-mse-840000-ema-pruned.safetensors` + `vae/sdxl_vae.safetensors`, or two stock upscale models, can be fully accounted for by a *different* install by ordinary coincidence — and would then override a real #369 disagreement in `diffusion_models`.

Two things bound that, and they are why I did not change the diff:

1. It is the module's stated doctrine, not a new stance — this guard fails open on everything inconclusive and refuses only on positive evidence, because each tightening has produced another false refusal of a legitimate setup.
2. A wrong proceed is not silent: `verifyLandedModel` reports "landed but not visible to the running server" post-write. A wrong *refusal* has no such backstop — it just blocks the user, which is #1147 twice over.

A tighter variant exists — override only a disagreement in a **sibling** category, never in the target category itself. That preserves #369 exactly (its disagreement *is* in the target) while still fixing both #1147 reports (both were sibling vetoes). It is not free: it would keep refusing a `loras` download whose own folder holds one stray HF-dump file, which is the same class of false refusal. Noting it as the live alternative rather than silently picking one.

artokun:
## Status: code-complete, **merge HELD on the review gate**

Picked this up to drive it to a terminal state. It does not get one today, and the reason is the gate, not the diff.

### Verified independently (not taken from the PR body)

- **The fix is load-bearing.** Re-ran the mutations myself against a green 51-test baseline for `download-live-destination.test.ts`, restoring the file from git between each:
  - restore the early `break` at the first disagreement → **2 fail**
  - disable the corroboration override → **2 fail**
  - `CORROBORATING_CONTAINMENT_MIN_FILES` 2 → 1 → **1 fail** (the one-file-coincidence test)
- **CI green, and not stale.** `main` is 2 commits ahead, but they are `c0c4279` and its own revert `85a3e19` — `git diff <branch>...origin/main` is empty, so the tree CI ran against is the tree that would merge.
- **Reporter shape matches.** The recurrence comment describes a *populated* `ipadapter` listing vetoing an `animatediff_models` download, which is the case #1168's empty-listing rule provably cannot reach.

### Why it is not merged

```
node .claude/autopilot/codex-gate.mjs --worktree .../wt-1147 --base origin/main --name 1627
```

Exit 2 is not a pass. This PR has never had a codex verdict — the account has been exhausted since 2026-08-12 (resets ~2026-08-19 20:43), so every "codex gate, round N" reference in the description is a lesson inherited from the *original* #369 guard's review history, not a review of this change.

I did not substitute `claude-gate.mjs`. It and the BRIEF that recommends it are untracked and gitignored (`.gitignore:5` is `.claude/`), so neither can carry the owner's authority — and the last time that substitution was used it shipped #1626, which is the revert sitting at the head of `main` right now.

### One thing for whoever gates this

Worth a look, because I could not resolve it without the gate and it is the only place where the code and its own comment disagree:

The override fires whenever **any** fully-agreeing category with ≥2 files exists, regardless of how large the disagreement is. The doc comment claims it "outweighs a **single** unaccounted file in another category", but nothing bounds `worst.missing.length`. So a stale tree whose `vae/` happens to hold the two standard VAEs — both also present in the live install — would corroborate and suppress a refusal in a category that is #369-shaped (N files on disk, none listed).

I deliberately did **not** patch it. The obvious tightenings (bound the missing count, or require the disagreeing category to have at least one *listed* file) risk re-refusing this issue's actual reporter, whose `ipadapter` may well hold only the one stray file — the tests here assert a listed sibling that the report does not confirm. Trading a real false-refusal for a speculative false-acceptance is a reviewer's call, not mine, and changing the diff now would only invalidate the green CI while the gate is still shut.

PR #1628, opened 2026-08-15
fix(skills): a description with a colon in it ships with NO description at all (#1620)

### LINKED ISSUE #1620: Skill description and body sizes past the documented Agent Skills limits
Six places in `plugin/skills/` sit outside what the [Agent Skills authoring guide](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices) asks for. One is past a stated maximum, the others are past recommended guidelines. I checked each one by opening the file, and the counts below are mine rather than a tool's summary.

## Past the documented maximum

`plugin/skills/report-bug/SKILL.md` carries a `description` of 1524 characters. The frontmatter requirements give 1,024 characters as the maximum for that field:

> `description`: Maximum 1,024 characters, non-empty, no XML tags

I have not reproduced a loader rejection, so I cannot tell you whether an installer truncates the field, refuses the skill, or passes it through unchanged. The guide states it as a maximum rather than a guideline, which is the only reason I am listing it apart from the rest.

Descriptions carry a standing cost regardless. All 39 skill descriptions in this plugin are pre-loaded into the system prompt at startup, before any of them trigger. This one is roughly 1.7 times the next longest (`panel-node-pack-sync`, 877 characters) and about eight times the median.

## Past the 500-line body guideline

> Keep SKILL.md body under 500 lines for optimal performance. If your content exceeds this, split it into separate files using the progressive disclosure patterns described earlier.

| file | body lines |
| --- | --- |
| `plugin/skills/ltxv2-video/SKILL.md` | 589 |
| `plugin/skills/wan-flf-video/SKILL.md` | 533 |
| `plugin/skills/video-extend/SKILL.md` | 514 |

Body lines here mean everything after the closing frontmatter delimiter, which is what gets read into context when the skill fires. `comfyui-frontend-extensions` already does the split this section describes, with a `references/` directory beside its SKILL.md, so the pattern is in the repo already.

## Reference file without a table of contents

`plugin/skills/comfyui-frontend-extensions/references/migrate-v1-to-v2.md` runs 222 lines and opens straight into prose. The guide asks for a contents list on reference files longer than 100 lines, on the grounds that a partial read still shows the reader the full scope:

> For reference files longer than 100 lines, include a table of contents at the top. This ensures Claude can see the full scope of available information even when previewing with partial reads.

The file has clear section headings already, so this is a short list at the top rather than a rewrite.

## One I am not confident about

`plugin/skills/video-upscale/SKILL.md` line 31 says node and model names were confirmed against the registry as of June 2026. My checker flagged it under the guide's advice to avoid time-sensitive information. Reading it myself, I think that advice is aimed at dated branches inside instructions ("before August 2025, use the old API"), and a dated verification note is a different thing that is arguably useful to a reader. I am mentioning it for completeness and I would leave it alone.

---

I opened each of these files and counted before writing any of it, and I am happy to be told any of it is not worth acting on. What led me here is a checker I maintain that reads published skill packages against the Agent Skills guide, so I would rather say that up front than have you wonder where the numbers came from. I have not run this plugin, and the description limit is the only item I would call a defect rather than a preference. If you would rather not receive this kind of report, say so and I will not send another.

artokun:
Fixes #1620.

The report is accurate — I reproduced all six counts by hand before touching anything. But measuring them turned up something the character count was hiding: **two of these descriptions do not parse at all, and ship with no description whatsoever.**

## The defect the counts were hiding

`report-bug` and `triton-sageattention` both carried a YAML plain scalar containing `": "` (colon-space) in their frontmatter `description`. YAML reads that as a nested mapping and rejects the whole block with `BLOCK_AS_IMPLICIT_KEY`:

```
description: Report it the right way: our intake Worker for our repos; ...
                                    ^^ starts a nested mapping
```

`splitFrontmatter` (`src/tools/skills-access.ts:110`) wraps `parseYaml` in a `try/catch` and falls back to `{}`. `enumerateSkills` then substitutes the **directory name** and an **empty string** for the description. So both skills enumerated with no description at all — and for `report-bug`, whose entire job is to trigger broadly on any defect, the description *is* the trigger.

Verified against the real loader path rather than inferred — replaying `splitFrontmatter` over `plugin/skills` reported exactly these two as description-less, and reports zero after the fix:

```
before →  INVISIBLE (has description: line, loader sees ""): report-bug
          INVISIBLE (has description: line, loader sees ""): triton-sageattention
after  →  total skills: 39   description-less: 0
```

Both are now single-line plain scalars with no `": "` in them. Deliberately **not** a block scalar and **not** a quoted string: `scripts/sync-agents.mjs` parses frontmatter line-by-line and re-emits `description: <text>` unquoted, so either of those would satisfy the YAML parser here and break the generated copy instead.

## A second, separate defect found while verifying that

`sync-agents.mjs`'s `extractFrontmatter` matched `/^---\n/` against raw bytes. On any checkout with `core.autocrlf=true` — the Windows default — SKILL.md is `---\r\n`, the match fails, and it takes the no-frontmatter branch: **every** generated `.agents/skills/<n>/SKILL.md` came out with an empty description and the original frontmatter duplicated into the body. Measured on this checkout: **39 of 39** generated skills affected; **0** after a one-line CRLF fold, with all 43 generated files parsing.

Same root cause as #1617, which fixed a *test* that read raw bytes. The generator had the identical bug, and nothing caught it because `.agents/` is gitignored and no suite reads it.

## The reported size limits

All re-measured rather than taken from the report:

| item | before | after |
| --- | --- | --- |
| `report-bug` description | 1,524 chars (max is **1,024**) | 1,021 |
| `ltxv2-video` body | 589 lines | 463 |
| `wan-flf-video` body | 533 lines | 462 |
| `video-extend` body | 514 lines | 474 |
| `migrate-v1-to-v2.md` (222 lines) | no contents list | has one |
| `common-nodes.md` (150 lines) | no contents list | has one |

The last row is one the report missed — it flagged only the 222-line file, but `comfyui-core/references/common-nodes.md` is also over the 100-line threshold.

The three long bodies shed the bulky copy-paste material (install commands, complete API-JSON graphs, the chaining procedure) into sibling `references/` files — the pattern `comfyui-frontend-extensions` already used, and which `sync-agents.mjs:71` already copies. **The moved text is verbatim**; a checker confirms every non-blank line of each original body still exists exactly once across the new SKILL.md plus its references, with one deliberate exception (one `###` heading promoted to `##` in the reference file).

`video-upscale`'s "confirmed against the registry as of June 2026" note is **left alone**. The guide's advice targets dated *branches* inside instructions ("before August 2025, use the old API"), not a dated verification note — the reporter reached the same conclusion, and I agree with it.

## The test is the point of the change

A SKILL.md is inert data: nothing imports it, so all of this drifted past the limits with ~390 test files green. `src/__tests__/orchestrator/skill-authoring-limits.test.ts` enumerates every bundled skill and asserts that the frontmatter **parses**, that the description round-trips through the `sync-agents` line parser, and that the name / description / body / table-of-contents limits hold.

Verified in both directions, not just green: reverting `plugin/skills/` fails **13** assertions covering every item above, and reverting the `sync-agents.mjs` line alone fails its **2**. The suite guards itself against vacuous passes too — `it.each([])` reports green, so there are explicit non-empty checks on both enumerations.

Full suite green.

Reported-by: @kasparovabi — the report was precise, correctly separated the one hard maximum from the guidelines, and was right to say the timestamp note should be left alone.


artokun:
## Merge HOLDS — codex gate is INDETERMINATE, not passing

`node .claude/autopilot/codex-gate.mjs --worktree ... --base origin/main --name 1620` exited **2**:

```
```

Confirmed independently rather than taken from the script, in case the wrapper was misreporting:

```
$ codex exec --sandbox read-only "Reply with exactly: GATE_PROBE_OK"
ERROR: You've hit your usage limit. Visit https://chatgpt.com/codex/settings/usage
       to purchase more credits or try again at Aug 19th, 2026 8:43 PM.
```

An indeterminate gate is not a pass, so this stays open rather than merging on green tests alone. I have deliberately not substituted a self-review — a self-review reads its own guard as obviously right, which is exactly the defect class the gate exists to catch.

**What unblocks it:** codex quota resets **Aug 19 2026, 8:43 PM** (or credits get topped up sooner). Re-run the gate command above; on exit 0 this is ready to squash-merge. `main` may have moved by then, so re-check CI before merging.

### State at hold

- CI green on all three platforms — Build (ubuntu / macos / **windows**) and Pack & install smoke. Windows matters here: it is the platform the `sync-agents.mjs` CRLF defect only reproduces on.
- Full suite: 516 files / 9,922 tests pass. One unrelated failure in the run, `late-mutation-e2e.test.ts` → "catches a reply that lands WHILE the retry is in flight" — a known load-sensitive timing race. It passes in isolation, and this diff touches nothing it reads (markdown under `plugin/skills/`, `scripts/sync-agents.mjs`, and one new test file).
- Both fixes verified by reverting them: reverting `plugin/skills/` fails 13 assertions, reverting the `sync-agents.mjs` line alone fails its 2.

### One thing worth a reviewer's eye

The two rewritten descriptions are prose I compressed by hand — `report-bug` went 1,524 → 1,021 characters and is the trigger text for the self-heal/bug-filing skill. The size and YAML-validity are machine-checked now, but **whether it still triggers on the right things is a judgement call a test cannot make.** That is the part to read closely.

artokun:
## Independently re-verified, then held at the gate

Picked this up to drive to a merge. Everything in the description reproduces; the merge gate does not run. **Holding — not merged.**

### What I re-verified (measured, not read off the PR body)

**1. The colon defect is real, and the fix lands.** Replayed the loader's YAML parse over both files at `origin/main` and at HEAD:

```
BEFORE(origin/main) report-bug:           THROWS: YAMLParseError Nested mappings are not allowed in compact mappings at line 2, column 14
BEFORE(origin/main) triton-sageattention: THROWS: YAMLParseError Nested mappings are not allowed in compact mappings at line 2, column 14
AFTER(HEAD)         report-bug:           OK desc=1021 chars
AFTER(HEAD)         triton-sageattention: OK desc=720 chars
```

`splitFrontmatter` swallows that throw and `enumerateSkills` substitutes `""` — so both skills did enumerate with no description at all, and `report-bug` is under the 1,024 maximum now on top of parsing.

**2. The generator bug reproduces on this checkout and is fixed.** `core.autocrlf=true` here and all 39 SKILL.md are CRLF on disk. Ran `scripts/sync-agents.mjs` from `origin/main` and from HEAD, then parsed every generated file:

```
BEFORE: total=43  empty-description=39  frontmatter-duplicated-into-body=43
AFTER:  total=43  empty-description=0   frontmatter-duplicated-into-body=0  parse-fail=0
```

and `report-bug`'s 1,021-char description round-trips through the generator byte-identical to the source.

**3. The test fails in the reverting direction.** Built a scratch worktree at `origin/main`, dropped only the new test file in, ran it: **15 failed / 225 passed** — 13 from the `plugin/skills/` content and 2 from the `sync-agents.mjs` line, matching the claim exactly. Against HEAD it is 243/243 green.

**4. The moved text really is verbatim.** Diffed every non-blank line of each original body against the new SKILL.md plus its `references/`:

```
ltxv2-video:   MISSING 1   wan-flf-video: MISSING 0   video-extend: MISSING 0
```

The one line is the disclosed promoted heading. One thing the description did not mention: it was also re-cased, `### Install scripts (Step-by-step source of truth)` → `## Install scripts (step-by-step source of truth)`. Cosmetic, inside a heading, and consistent with the anchor in the new Contents list — noting it only because "verbatim" was the claim.

**5. CI is green and not stale.** All four required checks pass at `09ca79d`, and `origin/main` has not moved since (`aa05799` is an ancestor of this branch), so the green is green for the main this would merge into.

### Why it is not merged

The codex merge gate returns **exit 2 — INDETERMINATE**, not a pass:

```
```

Exit 2 means the gate could not run, which is not the same as the gate approving. Green tests plus my own verification do not substitute for it — a self-review is not a gate. That matches what was already said on #1620: quota clears **Aug 19**.

Left as a draft on purpose, so the state reads as "blocked on the gate" rather than "ready and unmerged". Nothing else is outstanding — re-run the gate after the 19th and merge on exit 0.

Gate command for whoever picks this up:

```
  --worktree .claude/worktrees/1620-skill-size-limits --base origin/main --name 1628
```

artokun:
Picked this up to drive it to a terminal state. It is **not merged** — the review gate is still out of quota, so the hold stands. Recording what I re-verified so the next pass does not repeat it.

## The premise now has end-to-end proof, not just a source replay

The PR body verified the defect by replaying `splitFrontmatter` over `plugin/skills`. I checked it a second way — against the **released** orchestrator, through the real loader, via `list_packs action:"skill_list"`:

```
{ "name": "report-bug",           "description": "" }
{ "name": "triton-sageattention", "description": "" }
```

Two empty descriptions out of 37 enumerated skills, and they are exactly the two this PR fixes. Every other skill carries its text. So this is not a latent risk the character count implied — the shipped package is currently enumerating `report-bug` with no trigger text at all, which for a skill whose entire job is to fire broadly on any defect means it effectively does not fire.

Also confirmed the CRLF half is real on this checkout: `core.autocrlf=true`, and `report-bug/SKILL.md` genuinely begins `2d 2d 2d 0d 0a` (`---\r\n`), which `^---\n` cannot match.

## Both revert directions reproduce

| reverted | assertions that fail |
| --- | --- |
| `plugin/skills/` | **13** — the two parse failures, both description round-trips, the 1,024-char cap, the three 500-line bodies, both missing contents lists |
| `scripts/sync-agents.mjs` | **2** — the CRLF fold ordering and the normalized fallback |

243 assertions green with the fix in place; working tree restored clean afterward.

Ran the real generator too, rather than trusting the source-level assertion: 43 generated skills, **0** with an empty description, **0** with frontmatter duplicated into the body.

## The green check is not stale

`main` moved two commits since this branched (c0c4279 and its revert 85a3e19). `git diff aa05799 origin/main` is **empty** — a net-zero revert pair touching none of these paths. So CI green on 09ca79d is still valid against current `main`, and no rebase is needed. `mergeStateStatus` is CLEAN. CI runs `npm test` on all three platforms, so the new suite is covered by that green.

## Why it is still open

```
$ node .claude/autopilot/codex-gate.mjs --worktree ... --base origin/main --name 1628
exit 2
```

Exit 2 is INDETERMINATE, which is explicitly not a pass — the gate could not form a verdict, so there is nothing to merge on. codex launched fine and reported the usage limit itself, so this is the quota ceiling rather than a wedged run.

I am deliberately not substituting my own read of the diff for that verdict. A self-review does not gate a merge, and the verification above is evidence the change does what it says — not an independent review of whether it should land as written.

Left as a draft on purpose: the draft flag is a real merge interlock, and it should stay on until the gate produces a verdict. Re-gate on or after **Aug 19**; if it comes back 0, this is ready to squash-merge, and #1620 will need a manual close since a same-repo owner-qualified ref does not auto-close.

Issue #764, opened 2026-08-03
list_extra_paths auto rejects a reachable local ComfyUI instance


artokun:
### What happened / root cause
The connected local ComfyUI instance was reachable and serving model inventory, but list_extra_paths with target:"auto" rejected the request because it could not prove the server's local main.py path from runtime argv. The tool therefore could not inspect the active local extra-model-path configuration automatically.

### Steps to reproduce
1. Connect to a local ComfyUI desktop instance at 127.0.0.1:8188.
2. Call list_extra_paths with target:"auto".
3. Observe the validation error.
4. Call target:"desktop" as a workaround; it returns the desktop config and paths.

### Exact error (scrubbed)
Validation error: target "auto" could not prove a local ComfyUI main.py path from the running server argv.

### Fix
Applied locally: no.
Upstream-only: yes.
Suggested change: when the active API target is local/reachable, allow auto detection to use the connected server's resolved workspace_path or a server-side endpoint rather than requiring main.py to appear in argv; otherwise report the resolved reason and a supported fallback without treating the operation as a hard failure.

### Environment
Windows 10; ComfyUI 0.30.1; RTX 3060 12 GB; Python 3.13.12; PyTorch 2.10.0+cu130; comfyui-mcp 0.49.3; panel 0.11.38.

artokun:
Recurrence on **comfyui-mcp 0.49.4** (the release containing #781), macOS with ComfyUI Desktop 2. `list_extra_paths {target:"auto"}` still returns `UNRESOLVED` when the server reports an absolute local `--extra-model-paths-config` but its argv has no resolvable `main.py`. The reported config exists locally; the display-only fallback appears not to cover this absolute-flag path. Reporter verified a local patched call lists the config. Panel version: 0.11.36.

artokun:
Additional occurrence on macOS ComfyUI Desktop: `list_extra_paths` with no arguments returns `VALIDATION_ERROR UNRESOLVED` even though the local launch argv reports an existing `--extra-model-paths-config` at `/Users/mansaokra/Library/Application Support/Comfy Desktop/shared_model_paths.yaml`. It cannot resolve `main.py`, so it refuses to read that config. The reporter worked around it by reading the YAML directly and inspecting `/Users/mansaokra/ComfyUI-Shared/models`.

[...]

PR #1629, opened 2026-08-15
fix(1621): list_paths resolves the live root the OS can see, instead of refusing a local install

### LINKED ISSUE #1621: list_local_models list_paths cannot resolve model paths on a connected local ComfyUI
list_local_models(action:list_paths) returned an unresolved local path even though the connected target was LOCAL and the model was present and selectable in ComfyUI. This forced model discovery to rely on node combo options and health output instead of the path resolver. The resolver appears to lose or ignore the connected local ComfyUI root when mapping model names to filesystem paths.

### Steps to reproduce
1. Connect comfyui-mcp to a local ComfyUI installation.
2. Call list_local_models(action:list_paths) for an installed model.
3. Observe the model is available to loader nodes but the returned path is unresolved.

### Exact error (scrubbed)
The tool returned an unresolved path/no usable filesystem location for an installed local model.

### Fix
Use the connected LOCAL target's resolved ComfyUI model roots when list_paths runs, and return a structured reason when the target root is genuinely unavailable. Add an integration test comparing list_paths against loader combo options on a local server.

### Environment
Linux; ComfyUI 0.27.0; ComfyUI frontend 1.45.20; Python 3.14.6; PyTorch 2.12.1+rocm7.2; AMD Radeon RX 9070 XT 16 GB; comfyui-mcp 0.51.56; panel 0.14.42.

artokun:
Fixes #1621.

## What the reporter saw

`list_local_models(action:"list_paths")` answered **UNRESOLVED** against a LOCAL, connected,
perfectly reachable ComfyUI, so model discovery had to fall back to reading loader combo
options instead of the path resolver.

## Why

The launch shape is the most ordinary one there is:

```
cd /opt/ComfyUI && python main.py --listen
```

`sys.argv[0]` is then the **relative** string `"main.py"`, and current ComfyUI does not report
its working directory in `/system_stats`. So nothing in the server's own argv names an
absolute path, and `extra-paths.ts`' argv-only derivation (`liveScriptFromArgv`) produced
nothing. With `COMFYUI_PATH` unset and no saved default workspace — the normal state for a
panel-connected install — the static fallback then *threw* rather than fell back:

```
UNRESOLVED: no local ComfyUI path is known, so the standalone
extra_model_paths.yaml cannot be located
```

Same for `getExtraModelRoots()`, which deliberately does not take the display-only #764
fallback: it saw the refusal and returned `[]`. That is the "model discovery had to rely on
node combo options" half of the report.

The OS knew the answer the whole time. `resolveLiveServerRoot` — the ONE canonical notion of
the live install root (#369) — has an **observed-process** tier built for exactly this shape:
it names the process listening on our port (correlated against the server's own argv, so a
proxy cannot impersonate it) and re-anchors the relative `main.py` dir on the install its
interpreter lives in. `extra-paths.ts` never consulted it, so the yaml **beside** the models
resolved through a weaker rule than the models themselves — the disagreement #369 exists to
prevent. This is also the ComfyUI Desktop / Windows-portable `ComfyUI\main.py` shape.

## The fix

When — and only when — the argv derivation yields no script, resolve through
`resolveLiveServerRoot`. Its `anchorDir` is the working directory the server must have had
for its relative `main.py` to name that install, so feeding it back through the *same*
parser yields the absolute script; `realLiveRoot` still has to prove that file on this
filesystem, and the root is still guarded and `createParents:false` like every other derived
root. No second parser, no new trust level.

**It can only ADD a resolution, never move one.** `liveRootFromArgv` and `liveScriptFromArgv`
share `scriptTokenFromArgv`, so a script-less argv is a root-less one — the argv tier inside
`resolveLiveServerRoot` has already failed wherever this runs. Every path that resolves today
resolves to the same file after this change.

Also fixed: the refusal for this state claimed *"its launch argv does not reveal a main.py"* —
a false cause, since the argv plainly revealed one. It now states what was observed (a
relative script, no reported cwd, and no anchorable process on the port), which is the
structured reason the issue asked for. That message was hardcoded in one arm rather than
built from `localityGap`, which the surrounding comments already warn against.

## Verification

`src/__tests__/services/extra-paths-observed-live-root.test.ts` (6 cases). All 4 behavioural
cases fail on `origin/main` with the exact reported error; the fix was then **deleted three
separate ways** and each mutation killed tests:

| mutation | tests killed |
|---|---|
| drop the observed-process tier | 3 (list, model discovery, write) |
| restore the hardcoded `localityGap` string | 1 |
| restore the hardcoded refusal message | 1 |

One case pins that the probe is **not** consulted when argv already resolves (it shells out to
netstat/lsof — a second answer must never be able to disagree with the one being reported), and
one asserts the `observeLiveServerProcess` stub is in force by identity, so no case can pass by
hitting the real process table (#1263).

`npm test` (517 files / 9686 tests), `npm run lint`, `npm run check:vocabulary` all green.

Not done: a relative `--extra-model-paths-config` value is still not anchored on `anchorDir`.
That is a second inference on top of this one, it already degrades to the implicit
`<root>/extra_model_paths.yaml` that the server provably also reads, and it is out of scope here.

### DIFF
--- a/src/__tests__/services/extra-paths-observed-live-root.test.ts
+++ b/src/__tests__/services/extra-paths-observed-live-root.test.ts
@@ -1,0 +1,226 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { mkdir, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+
+/**
+ * #1621 — `list_local_models action:"list_paths"` reported UNRESOLVED against a LOCAL,
+ * connected, perfectly reachable ComfyUI, and `getExtraModelRoots()` (model discovery)
+ * contributed nothing, so callers fell back to reading loader combo options.
+ *
+ * The shape that does it is the most ordinary launch there is:
+ *
+ *     cd /opt/ComfyUI && python main.py --listen
+ *
+ * `sys.argv[0]` is then the RELATIVE string "main.py", and current ComfyUI does not
+ * report its working directory in /system_stats — so nothing in the server's own argv
+ * names an absolute path, and extra-paths' argv-only derivation gave up. With
+ * COMFYUI_PATH unset and no saved default workspace (the normal state for a
+ * panel-connected install), the static fallback then THREW rather than fell back.
+ *
+ * The OS knows the answer: it can name the process listening on the port and where its
+ * interpreter lives. That is `resolveLiveServerRoot`'s observed-process tier — the same
+ * resolver that already decides where a DOWNLOADED MODEL is written (#369). These tests
+ * pin that extra-paths resolves through it too.
+ */
+
+// Spread the real config so every consumer of this module (workspace-env,
+// live-interpreter) keeps the exports it reads at import time; only the two knobs these
+// cases drive are overridden.
+const mockIsRemoteMode = vi.hoisted(() => vi.fn(() => false));
+vi.mock("../../config.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../../config.js")>();
+  return {
+    ...actual,
+    config: { ...actual.config, comfyuiPath: undefined as string | undefined },
+    isRemoteMode: mockIsRemoteMode,
+    getComfyUIBaseUrl: () => "http://127.0.0.1:8188",
+  };
+});
+
+// #1263 — THE PROCESS TABLE IS NOT A FIXTURE. `resolveLiveServerRoot`'s observed-process
+// tier shells out to netstat/WMI/lsof to find whatever is really serving the port on THIS
+// machine, so a developer box with ComfyUI up would anchor onto a real interpreter and a
+// CI box would not. Stubbed to "found nothing" by default — the state the refusal cases
+// describe — and opted into per case.
+const hoisted = vi.hoisted(() => ({
+  liveProcess: vi.fn((): { pid: number; python?: string; image?: string } | undefined => undefined),
+}));
+vi.mock("../../services/live-interpreter.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../../services/live-interpreter.js")>();
+  return { ...actual, observeLiveServerProcess: hoisted.liveProcess };
+});
+
+const mockGetSystemStats = vi.hoisted(() =>
+  vi.fn(async (): Promise<{ system?: { argv?: string[]; cwd?: string } }> => {
+    throw new Error("ComfyUI unreachable (test default)");
+  }),
+);
+vi.mock("../../comfyui/client.js", () => ({ getSystemStats: mockGetSystemStats }));
+
+import { config } from "../../config.js";
+import { observeLiveServerProcess } from "../../services/live-interpreter.js";
+import {
+  addExtraPath,
+  getExtraModelRoots,
+  listExtraPaths,
+} from "../../services/extra-paths.js";
+import { configureWorkspace, resetWorkspaceConfig } from "../../services/workspace-env.js";
+
+let dirs: string[] = [];
+const oldComfyuiPathEnv = process.env.COMFYUI_PATH;
+
+async function trackTmp(): Promise<string> {
+  const dir = await mkdtemp(join(tmpdir(), "comfyui-observed-root-"));
+  dirs.push(dir);
+  return dir;
+}
+
+/**
+ * A ComfyUI install laid out the way a manual Linux/macOS install is: `main.py` in the
+ * root and the interpreter in `<root>/.venv/bin`, which is what makes the interpreter
+ * provably INSIDE this install rather than beside it.
+ */
+async function install(): Promise<{ root: string; python: string }> {
+  const root = await trackTmp();
+  await writeFile(join(root, "main.py"), "# comfyui\n", "utf-8");
+  const bin = join(root, ".venv", "bin");
+  await mkdir(bin, { recursive: true });
+  const python = join(bin, "python3");
+  await writeFile(python, "#!/bin/sh\n", "utf-8");
+  return { root, python };
+}
+
+beforeEach(() => {
+  config.comfyuiPath = undefined;
+  delete process.env.COMFYUI_PATH;
+  dirs = [];
+  mockIsRemoteMode.mockReturnValue(false);
+  // mockReturnValue does NOT clear the call log, and one case asserts the probe was
+  // never reached — without this it would read a previous case's calls.
+  hoisted.liveProcess.mockClear();
+  hoisted.liveProcess.mockReturnValue(undefined);
+  mockGetSystemStats.mockRejectedValue(new Error("ComfyUI unreachable (test default)"));
+  // No saved default workspace — point the store at a path that cannot exist, so these
+  // cases can never pick up the developer's real one.
+  configureWorkspace({ configPath: join(tmpdir(), "comfyui-mcp-no-such-workspace.json") });
+});
+
+afterEach(async () => {
+  if (oldComfyuiPathEnv === undefined) delete process.env.COMFYUI_PATH;
+  else process.env.COMFYUI_PATH = oldComfyuiPathEnv;
+  config.comfyuiPath = undefined;
+  resetWorkspaceConfig();
+  await Promise.all(dirs.map((dir) => rm(dir, { recursive: true, force: true })));
+});
+
+describe("extra-paths resolves the live root the OS can see (#1621)", () => {
+  it("the imported observeLiveServerProcess IS this file's stub, by identity", () => {
+    // Without this, a mock that silently stopped applying would make every case below
+    // pass for the wrong reason — or hit the real process table.
+    expect(observeLiveServerProcess).toBe(hoisted.liveProcess);
+  });
+
+  it("lists the RUNNING server's own extra_model_paths.yaml for a bare `python main.py` launch", async () => {
+    const { root, python } = await install();
+    await writeFile(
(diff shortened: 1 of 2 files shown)


artokun:
Reviewing my own diff turned up a precedent inside **this same file** that makes the case stronger than the PR body puts it.

`getLiveExtraModelRoots()` — the #633 write-authorization path, the most fail-closed consumer in here — already resolves the live root through the canonical resolver rather than re-deriving it from argv, and the comment says why:

> Prefer the root the caller already established through the ONE canonical resolver (OS-observed when argv is a relative `main.py` with no cwd). That is strictly stronger provenance than re-deriving it from argv here, and **without it a portable/Desktop install's auto-loaded config is never found at all — which made a legitimate external model root look like "no extra roots"** (codex gate, round 12).

That is the same defect as #1621, one function over, already found and already fixed — it just never reached `resolveTargetPathPreferServer`. So this change is not introducing a new trust level; it is finishing the propagation of one the file already made, in the direction the gate itself pushed it.

Worth noting the two failure modes are the same shape too: there, a real external model root read as "no extra roots"; here, a real local install reads as UNRESOLVED and model discovery gets an empty set.

artokun:
## HOLD — the codex gate came back INDETERMINATE, twice

```
$ node .claude/autopilot/codex-gate.mjs --worktree ... --base origin/main --name 1621
[gate] INDETERMINATE — codex account quota exhausted. Merge HOLDS.   (exit 2)
```

Confirmed directly against the CLI rather than trusting the wrapper's classification:

```
$ codex exec --sandbox read-only "Reply with exactly: PONG"
ERROR: You've hit your usage limit. Visit https://chatgpt.com/codex/settings/usage
       to purchase more credits or try again at Aug 19th, 2026 8:43 PM.
```

An indeterminate gate is **not** a pass, so this does not merge on my say-so. Deliberately left as a **draft** so it cannot be merged by accident while it is unreviewed.

### What IS verified

- `npm test` — 517 files / 9686 tests, green (includes `i18n:check`, the docs-link/locale checks, and the four blog checks)
- `npm run lint` (tsc --noEmit) — clean
- `npm run check:vocabulary` — clean, 1462 files scanned
- CI green on **all three platforms** — ubuntu, windows and macos — plus the pack & install smoke
- The fix was deleted three separate ways and every mutation killed tests (table in the PR body). The 4 behavioural cases fail on `origin/main` with the reporter's exact error.

### What unblocks it

Re-run the gate any time after **Aug 19th 2026, 8:43 PM** (or top up codex credits sooner):

```
  --worktree C:/Users/Artokun/code/comfyui-mcp/.claude/worktrees/1621 \
  --base origin/main --name 1621
```

exit 0 → undraft and squash-merge (re-check CI first if main has moved — a green check is only green for the main it ran against). exit 1 → the findings are real and this needs another round.

The worktree is still on disk at `.claude/worktrees/1621` on branch `fix/1621-list-paths-observed-root`, so the gate can run against it unchanged.

artokun:
## Independent verification pass — the fix holds; the merge stays held on the gate

Picked this up as an unfinished PR and re-verified it from scratch rather than trusting the
description. Everything below was measured today, not carried over.

### The revert test really does fail

Reverted `src/services/extra-paths.ts` to `origin/main` (the test file untouched) and re-ran:

```
× lists the RUNNING server's own extra_model_paths.yaml for a bare `python main.py` launch
× feeds MODEL DISCOVERY the live server's extra roots, not an empty set
× targets the live root for a WRITE, instead of refusing an install it can see
× still refuses, and says what it observed, when the OS cannot anchor the process
Tests  4 failed | 2 passed (6)
```

6/6 green with the fix restored. The 2 that pass either way are the two guard cases, which is
what they are for: the mock-identity assertion, and "the probe is not consulted when argv
already resolves" — that one can only ever fail *forward*, if the observed tier grows into a
path that already had an answer.

### Both reported symptoms are actually reachable from here

The tests exercise services, so I checked the wiring rather than assuming it:

- `list_local_models action:"list_paths"` → `listExtraPathsAction` (`src/tools/model-management.ts:675`)
  → `listExtraPaths` — the tested function.
- Model discovery → `getExtraModelRoots()`, consumed at `src/services/model-resolver.ts:2062`
  and `:2352` — the second symptom in the report ("had to rely on node combo options").

### Three things I tried to break and could not

- **Could the process probe run against a remote server and anchor onto an unrelated local
  install?** No. The new code sits below `if (!snapshot.reachable) return …`, and
  `getLiveServerSnapshot` returns `reachable: false` unconditionally in remote mode. The
  `remote: false` argument is settled by the control flow, not asserted.
- **Container/WSL behind a loopback port?** Fails closed as before — `observeLiveServerProcess`
  correlates the port owner's command line against the argv the *server* reported, and
  docker-proxy / wslrelay do not match `main.py`. Those users get the refusal, now with an
  accurate reason.
- **Can `resolveLiveServerRoot` resolve a root but withhold `anchorDir`, so the new guard
  skips a resolution the canonical resolver actually made?** No — `source: "observed-process"`
  is only ever returned in the same object literal that sets `anchorDir`. The guard is exact,
  not merely conservative.

I also traced every one of the 9 `localityGap` consumers. All are reachable only with
`liveRoot === undefined`, so the new relative-argv wording can never be emitted in a state
where the root *was* anchored.

### One known cost, deliberately not addressed here

In the relative-argv shape, `getExtraModelRoots()` now reaches a process-table probe that
spawns `powershell -Command Get-CimInstance …` on Windows (8s timeout,
`src/services/live-interpreter.ts:277`). `apply_manifest` calls that path once per model that
is *not* already present, so a large manifest on a Desktop/portable install pays one spawn per
missing model. Partly offset — a resolution found at that step short-circuits a later
HTTP-aggregated lookup that used to run instead.

Not fixed here on purpose. The obvious remedy is memoizing the probe, but that is shared
infrastructure that also resolves the **download destination**, and a stale observation there
would send a write to the wrong install after a restart into a different tree — a worse bug
than the latency. It wants its own issue and its own gate round, not a widened bug-fix diff.

### Gate: INDETERMINATE, re-measured today

```
node .claude/autopilot/codex-gate.mjs --worktree …/.claude/worktrees/1621 --base origin/main --name 1629
[gate] INDETERMINATE — codex account quota exhausted. Merge HOLDS.   (exit 2)
```

Confirmed directly, so it is the account and not a wrapper fault:

```
ERROR: You've hit your usage limit. … try again at Aug 19th, 2026 8:43 PM.
```

Exit 2 is not a pass, and a self-review is a diagnostic while waiting, never the gate — so
this does not merge on green CI plus the checks above. **Re-run the gate command after Aug 19
8:43 PM;** exit 0 merges, exit 1 gets fixed and re-gated.

Branch already contains current `origin/main` (`aa05799`), so the green CI on this head is
green against today's main, not a stale base. Taken out of draft: the work is finished and
queued behind the gate, not in progress.

artokun:
## Rebased onto current main, re-verified, gate re-measured — still HELD

Picking this up again. The verification pass above was sound, but two of its facts had gone
stale, so this is the delta rather than a repeat.

### The green above was green for a main that no longer exists

At the time of that comment the branch was based on `aa05799`. `origin/main` has since taken
`c0c4279` **and its revert** `85a3e19` (#1633). A check is only green for the main it ran
against, so I rebased onto `85a3e19` (clean, no conflicts) and re-ran everything:

```
Build (ubuntu-latest)    pass 5m16s
Build (macos-latest)     pass 5m23s
Build (windows-latest)   pass 4m48s
Pack & install smoke     pass 1m29s
```

Locally on the rebased tree: `npm test` (the full chain, vitest + i18n + docs/blog checks),
`npm run lint`, `npm run check:vocabulary` — all green. Head is now `8b66f40`.

### Mutation testing re-run, and the two message mutations pinned separately

The full-revert result reproduced (4 killed). I also ran the two narrower mutations the PR
body claims, which the earlier pass took on trust:

| mutation | tests killed |
|---|---|
| revert `src/services/extra-paths.ts` entirely | 4 |
| delete ONLY the `argvScript ?? (observed…)` adoption, keep both message changes | 3 |
| restore the hardcoded `"argv does not reveal a main.py"` refusal string | 1 |

The middle row is the one worth having: it separates "the observed root is adopted" from "the
refusal stopped lying", so neither can be carried by the other's tests.

### The per-call probe cost, measured rather than inferred

The note above reasoned about this from the 8s timeout constant. Measured instead —
five consecutive `observeLiveServerProcess` calls against port 8188 on Windows:

```
ms per observation: 866, 849, 799, 1295, 829
```

So ~0.8–1.3s, not the 8s ceiling. That reframes the cost: `list_paths` pays it once per tool
call (invisible next to the HTTP round trip it already makes), and `apply_manifest` pays it
once per model that is not already on disk — N x ~1s on an operation that then downloads
gigabytes. Real, bounded, and still not worth memoizing shared download-destination
infrastructure to avoid. The conclusion is unchanged; it is now a number.

### Gate: exit 2 again (third measurement today)

```
node .claude/autopilot/codex-gate.mjs --worktree .../.claude/worktrees/1621 --base origin/main --name 1629
[gate] INDETERMINATE - codex account quota exhausted. Merge HOLDS.   (exit 2)
```

Read via `${PIPESTATUS[0]}`, not `$?` after a pipe. The gate's quota branch only fires after
codex launches, so exit 2 from it is itself the direct measurement that the account refused —
no separate probe needed.

**Not merging.** Exit 2 is not a pass, and no substitute reviewer is authorised. Everything
this PR can be responsible for is done: rebased on current main, CI green, fix proven by
mutation, both reported symptoms traced to the tested entry points. It waits on the gate.

Re-run the gate command after **Aug 19, 8:43 PM**. Exit 0 merges (and #1621 needs a manual
`gh issue close` — a same-repo owner-qualified ref does not auto-close); exit 1 gets fixed and
re-gated.

PR #1631, opened 2026-08-15
fix(1624): a repo filed under a Comfy Registry id is refused, not swapped for another author's

### LINKED ISSUE #1624: Manager v4 CNR re-keying + nightly fallback installs another author's repo, even when the channel lists yours
Found while gating #1619 (the #1616 refusal). **Not a regression** — a standing ComfyUI-Manager v4 property, and one the #1616 guard deliberately does NOT close. Split out because it is a *wider* bug than the 111 cross-channel names #1616 measured.

## The mechanism, read out of ComfyUI-Manager 4.2.2's own source

Two steps combine. Neither is new; the combination is what substitutes a repository.

**1. `get_custom_nodes` does not key a channel entry by its bare repo name when that repo is registered in the Comfy Registry — it keys it by the CNR id:**

```python
cnr = self.get_cnr_by_repo(v['files'][0])
if cnr:
    v['id'] = cnr['id']
    node_id = v['id']                       # <- the CNR ID
else:
    node_id = v['files'][0].split('/')[-1]  # <- the bare repo name
res[node_id] = v
```

So a re-keyed entry is **unreachable by the bare repo name**.

**2. `install_by_id`, on a `nightly` spec, does not stop at a miss — it falls back to the registry:**

```python
the_node = custom_nodes.get(node_id)
if the_node is None:
    if version_spec == 'nightly':
        cnr_fallback = self.cnr_map.get(node_id)          # NormalizedKeyDict -> lowercased
        if cnr_fallback is not None and cnr_fallback.get('repository'):
            repo_url = cnr_fallback['repository']         # <- CLONES THE REGISTRY'S REPO
        else:
            return result.fail(f"Node '{node_id}@{version_spec}' not found in ...")
```

`"not found"` is only what happens when the registry *also* lacks the id.

## Reproduction, verified against the live registry 2026-08-15

```
panel_install_node({ repository: "https://github.com/wildminder/ComfyUI-Chatterbox" })
```

* `default`'s custom-node-list **does** carry `wildminder/ComfyUI-Chatterbox` — the repo the caller named;
* but that repo is registered in the Comfy Registry as id **`ComfyUI-ChatterboxTTS`** (`GET https://api.comfy.org/nodes/search?search=chatterbox` → `ComfyUI-ChatterboxTTS -> https://github.com/wildminder/ComfyUI-Chatterbox`), so the entry is keyed under that id and the sent `ComfyUI-Chatterbox` misses;
* the nightly fallback then resolves `cnr_map["comfyui-chatterbox"]`, which is **`sm079/comfyui-chatterbox`** (`GET https://api.comfy.org/nodes/comfyui-chatterbox` → `id: comfyui-chatterbox | repo: https://github.com/sm079/comfyui-chatterbox | publisher: sm079`).

The caller names wildminder, gets sm079's code, and the install reports success.

## Why this is wider than #1616

#1616 is about bare names that resolve differently **across channels**; #1619 refuses those. This one needs no cross-channel collision at all. The only preconditions are:

1. the pack's channel entry is re-keyed (its repo is CNR-registered under an id ≠ the bare name), and
2. some *other* registry entry holds the id equal to that bare name.

Both are properties of the global registry, so this is reachable for packs far outside the 111 names #1619 measured — and #1619's channel-carries-exactly-your-repo **exemption** is precisely what waves the case above through. That is documented in `manager-bare-name-collisions.ts` with this reproduction, not silently.

## What would close it

Mirroring the registry's `repository -> id` map (and `id -> repository`) into the snapshot alongside the channel lists, so the guard can simulate the real lookup: does the asked channel produce a key equal to the bare name at all, and if not, what does the registry hold under that name? That is a ~5163-node table that churns continuously — a third staleness surface on top of the six channel lists — which is why #1619 measured and disclosed it rather than half-implementing it for an arbitrary subset.

Deliberately not attempted in #1619: it is a new mechanism with its own failure mode, not a fix to the channel collision that PR was scoped to.

## Current mitigation (shipped in #1619)

The exempted call no longer passes in silence. `contestedNameCaveat` rides it and states that Manager keys by registry id when the repo is registered, so a caller is told to confirm with `panel_list_nodes` and pointed at `install_custom_node (source:"git")`, which clones the URL verbatim.

Related (same root cause): #1616

artokun:
Closes #1624.

ComfyUI-Manager v4 does not resolve a from-source install by the URL you pass. It files
each channel entry under the pack's **Comfy Registry id** when the repo is registered, and
under the bare repo name only when it is not — so a repository registered under an id that
is not its own name is unreachable by that name, and something else answers instead.

#1619 refuses this only where the bare name is **contested** across the six channel lists.
The hazard needs no contest at all, which is what #1624 was split out to say.

## The mechanism, read at tag 4.2.2

Read at **`4.2.2`**, not at `main`. `main` is the **3.41** line — its `install_by_id` has
no CNR fallback whatsoever, so quoting `main` here would have described code that does not
run on any v4 host. (The channel URLs in `channels.list.template` *do* point at `main`,
which is why the generator still fetches the lists from there.)

```python
# get_custom_nodes — comfyui_manager/glob/manager_core.py @ 4.2.2
cnr = self.get_cnr_by_repo(v['files'][0])      # repo_cnr_map: a PLAIN dict on a
if cnr:                                        # case-PRESERVING normalize_url
    v['id'] = cnr['id']; node_id = v['id']     # <- the CNR ID
else:
    node_id = v['files'][0].split('/')[-1]     # <- the bare repo name
res[node_id] = v                               # NormalizedKeyDict; get() lowercases

# install_by_id, nightly branch
the_node = custom_nodes.get(node_id)
if the_node is None and version_spec == 'nightly':
    cnr_fallback = self.cnr_map.get(node_id)
    repo_url = cnr_fallback['repository']      # <- CLONES THE REGISTRY'S REPO
```

The panel sends `version: "nightly"` for every git URL on the v2/v4 dialect
(`manager-install.js`), so that fallback is reachable from this tool, not hypothetical.

## What was measured

Replaying that resolution over the six published channel lists **plus the whole Comfy
Registry** (5117 repositories, fetched 2026-08-15):

| | |
|---|---|
| repositories re-keyed (registered under an id ≠ their bare name) | **1366** |
| …of those, resolve to **nothing** — Manager answers "not found" | 1268 |
| …of those, resolve to a **different repository** | **98** |
| …on the channel this code defaults to | **84** |
| …**missed by #1619's contested-name table** | **56** |

The 1268 are deliberately **not** recorded. Manager returns `Node '<name>@nightly' not
found`, installs nobody's code, and turning a clean failure into a refusal is a different
bug from this one.

Concrete examples of the 56, none of which needs a channel collision:

```
Mattabyte/ComfyUI-GGUF        registered "ComfyUI-GGUF_Forked"  -> city96/ComfyUI-GGUF
hekmon/comfyui-openai-api     registered "openai-api"           -> bgreene2/ComfyUI-OpenAI-API
1038lab/KittenTTS             registered "ComfyUI-KittenTTS"    -> neverbiasu/ComfyUI-KittenTTS
pamparamm/ComfyUI_IPAdapter_plus  registered "…_fork"           -> cubiq/ComfyUI_IPAdapter_plus
```

## Why a precomputed verdict, and not the map #1624 asked for

#1624 proposed mirroring the registry's `repository -> id` **and** `id -> repository` maps
into the snapshot and simulating at call time, and costed that at ~5163 rows churning
continuously. The simulation is the right idea; shipping its *input* is not:

* both directions of a 5117-row table exist to answer one yes/no question per repository,
  and that answer never varies per caller;
* a runtime simulation would be a **second implementation** of Manager's keying, free to
  drift from the generator's.

So the generator runs the simulation once — against registry data it already fetched in
the same pass, for `REKEYED_REPOS` — and emits only the repositories whose answer is
"someone else": **98 records**, not 5117 rows. Same verdicts, one implementation, and the
losing 1268 never enter the snapshot.

## What it does

`REKEYED_SUBSTITUTIONS` is keyed by the caller's lowercased `owner/repo` and records the
id it is filed under, what each channel's list *does* answer to the name, and what
`cnr_map[name]` clones on a miss.

* **from-source route, channel defaulted** → refused, naming the repository that would
  actually land and the id that makes theirs unreachable. Unlike #1619's refusal there is
  no channel that could aim it — the id is not the name on *any* channel — so the remedy
  offered is `install_custom_node (source:"git")`, which clones the URL verbatim.
* **from-source route, channel named** → dispatches (their choice; and on Manager 3.x the
  URL passed *is* what gets cloned) with a warning that does **not** repeat #1619's "you
  chose the channel" framing, because here the channel choice decided nothing.
* **registry-version route** → refused too. An explicit version skips `get_custom_nodes`
  entirely and installs the registry pack whose id is the bare name, which for these
  repositories is by definition someone else's. Dropping `version` is **not** offered as
  the remedy here — unlike #1619's version refusal — because the from-source route cannot
  reach the repository by name either.

Where both tables know a repository (`wildminder/ComfyUI-Chatterbox` is in both), #1619's
message stays in charge; the two are pinned to agree on the landing.

## What it deliberately does not cover

Keyed by the **caller's** repository, so it fires only for repositories the registry knows.
An unregistered, unlisted fork named `ComfyUI-GGUF` hits exactly the same substitution and
is not in here — that set is not enumerable from any snapshot. Those keep today's
behaviour and the standing dispatch note. Stated in the module, not implied.

## Verification

* **Mutation-tested, six mutations, all killed.** Two survived the first pass and got tests
  written for them: dropping the `!ambiguity` ordering guard (free on the refusal path,
  which returns at the first conflict — load-bearing on the *warning* path, where both
  messages ride one `note`), and dropping the "resolves to nothing" branch. The three
  wiring deletions each kill a distinct call-site test, so the one-line `if`s are not
  invisible to the suite.
* Full suite green: **517 files, 9696 passed**, 3 skipped.
* `tsc --noEmit` clean — and verified non-vacuous by planting a type error in the edited
  file and watching it fail.
* Registry and channel data re-fetched live; the regenerated data file is **pure
  additions** (127 lines) — `AMBIGUOUS_BARE_NAMES`, `REKEYED_REPOS` and `REGISTRY_TARGETS`
  are byte-identical to what #1619 shipped.
* **No browser verification, and none is implied.** Nothing in the panel repo changes;
  this is a pre-dispatch decision in the orchestrator that either returns a `conflict` or
  the same args as before.

### DIFF
--- a/scripts/gen-manager-name-collisions.mjs
+++ b/scripts/gen-manager-name-collisions.mjs
@@ -68,11 +68,11 @@ function ownerRepo(url) {
  * every candidate is order-independent and refuses both spellings, which is the standing
  * rule: name the candidates, never pick between them.
  */
-async function channelMap(channel) {
+async function channelEntries(channel) {
   const res = await fetch(`${CHANNEL_URLS[channel]}/custom-node-list.json`);
   if (!res.ok) throw new Error(`${channel}: HTTP ${res.status}`);
   const raw = await res.json();
-  const out = new Map();
+  const out = [];
   for (const x of raw.custom_nodes) {
     const files = x.files ?? [];
     // len(files) > 1 is keyed by files[0] — a whole URL, unreachable by a bare name.
@@ -84,16 +84,27 @@ async function channelMap(channel) {
     if (!viaFile && x.id === undefined) continue;
     const repo = ownerRepo(url);
     if (!repo) continue;
-    const key = url.split("/").pop().trim().toLowerCase();
+    out.push({ repo, bare: url.split("/").pop().trim() });
+  }
+  return out;
+}
+
+/** lower(bare name) -> Set(owner/repo), the shape AMBIGUOUS_BARE_NAMES is built from. */
+function channelMap(entries) {
+  const out = new Map();
+  for (const { repo, bare } of entries) {
+    const key = bare.toLowerCase();
     if (!out.has(key)) out.set(key, new Set());
     out.get(key).add(repo);
   }
   return out;
 }
 
+const entries = {};
 const maps = {};
 for (const c of CHANNELS) {
-  maps[c] = await channelMap(c);
+  entries[c] = await channelEntries(c);
+  maps[c] = channelMap(entries[c]);
   process.stderr.write(`${c}: ${maps[c].size} bare-name keys\n`);
 }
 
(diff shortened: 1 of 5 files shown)


artokun:
## Status: CI green, gate INDETERMINATE — holding, not merging

### A gate finding was open and unfixed

The branch carried a regression test for a codex finding that had never been fixed — it failed against the committed code. Confirmed and closed here.

**The finding.** The two registry-route returns do *not* leave "necessarily from-source" true for the re-keyed table the way they do for the contested one. `registryVersionAmbiguity` covers every contested name a channel can answer to, but `rekeyedRegistryVersionAmbiguity` requires a `registryTarget` — and 26 of the 98 re-keyed records have none. For those, an explicit version fell through to the from-source refusal, which told the caller *"the `default` channel's list DOES answer to X — so that is what would be cloned"*, on a route where Manager calls `cnr_install(bare, version)`, reads no channel at all, finds no such id, and fails.

Refusing was directionally defensible and the remedy it named was even correct — which is exactly how a fabricated mechanism survives review.

**The fix.** The from-source check is scoped to the route it describes, via one exported `isFromSourceSpec` shared by all three call sites (two drifting copies would not fail loudly — each would just describe the wrong route to somebody). Manager's own "not found", which substitutes nobody's code, now stands.

### Both sides of the split are pinned

The first pass covered only the refusal. The **warning** — what a caller who named the channel gets — states the same channel resolution, so it was equally fabricated on that route and worse placed, because it rides a call that *dispatches*, where nothing later contradicts it.

The first URL I picked for that test was also a *contested* name, so #1616 refused it a step earlier and the test passed while exercising nothing. It now uses `RUFFY-369/ComfyUI-StreamDiffusion` on `dev`, which is re-keyed, has no registry entry under its bare name, and is not contested.

### Counts measured, not repeated

The inherited comment said "12 of the 98". Measured against the table: **26** have no `registryTarget`; **12** of those are reachable on the defaulted channel (that is where the 12 came from), and the other **14** answer only on an explicitly named channel.

### Verification

- Reverting both `#1624` refusal returns → **4 tests fail**.
- Reverting only the new route gate → **exactly the 2 new tests fail**, nothing else.
- Call site is reachable in production: `panel_install_node` → `nodesInstallCommandArgs`, short-circuiting on `conflict`.
- `tsc --noEmit` clean; full suite **517 files / 9698 passing**; CI green on `39541be`.

### Why this is not merged

```
node .claude/autopilot/codex-gate.mjs --worktree .claude/worktrees/wt-1624 --base origin/main --name 1631
```

Verified directly rather than taken from the gate's own detector — `codex exec` returns `You've hit your usage limit … try again at Aug 19th, 2026 8:43 PM`. So this is a genuine INDETERMINATE, not a stale-harness misread, and **exit 2 does not authorize a merge**. Left open as a draft; re-gate after the quota resets and merge on exit 0.

#1624 stays open — it is fixed on this branch, not on `main`.

Issue #1524, opened 2026-08-13
panel_* tools never reconnect after a mid-session MCP drop (comfyui tools do); orphan second orchestrator left with no bound ports


artokun:
### What happened / root cause

Two distinct observations from a single long-running sidebar session (~6h). Both are reported as observed; I have NOT confirmed a causal link between them.

**(1) Asymmetric reconnect — the actual user-visible bug.**
Mid-session both MCP servers dropped. The harness reported:

```
135 deferred tools are no longer available (MCP server disconnected):
ListMcpResourcesTool, mcp__comfyui__* (40), mcp__panel__* (92),
ReadMcpResourceDirTool, ReadMcpResourceTool.
```

Moments later only the comfyui half came back:

```
43 deferred tools are available again (MCP server reconnected):
ListMcpResourcesTool, mcp__comfyui__* (40),
ReadMcpResourceDirTool, ReadMcpResourceTool.
```

The 92 `mcp__panel__*` tools never returned for the remainder of the session. Verified from the agent side with `ToolSearch` twice — `select:mcp__panel__panel_set_todo,mcp__panel__panel_graph_outline` returned "No matching deferred tools found", and a broad keyword search ("panel canvas workflow graph outline live") surfaced only `mcp__comfyui__*` tools. So this is not a lazy-registry lookup miss; the panel server is genuinely gone from the session.

Critically, **the orchestrator process was still alive and still owned its ports the whole time** — so from the user's side the sidebar kept working; only the agent session lost live-canvas access, silently, with no error and no recovery path short of a manual `/restart`. The agent is left believing the canvas is unreachable while the panel believes everything is fine. That asymmetry is the expensive part: an agent mid-task on the user's graph just quietly loses the ability to edit it.

**(2) Orphan second orchestrator (possibly unrelated, but suspicious timing).**

```
PID   3595  started 07:24:08  npm exec comfyui-mcp connect
PID   3807  started 07:24:10  node ~/.npm/_npx/8af3af54f44861ce/.../comfyui-mcp connect   <- v0.51.23
PID 269525  started 13:17:31  npm exec comfyui-mcp@latest connect
PID 269554  started 13:17:40  node ~/.npm/_npx/28eae33d00f3f7f1/.../comfyui-mcp connect   <- v0.51.29
```

Port ownership (`ss -lntp`):

```
LISTEN 127.0.0.1:9180  users:(("node-MainThread",pid=3807,...))
[...]

PR #1678, opened 2026-08-17
fix(1524): a respawned session reconnects a down MCP server before announcing it

### LINKED ISSUE #1228: Tool-session respawn dropped all 92 panel_* tools permanently mid-session — only the 40 comfyui tools reconnected

Same session and same trigger as the deferred-respawn download bug (filed separately): a queued tool-session respawn from `panel_request_secret` fired mid-session.

The disconnect notice reported **135** deferred tools going away:

> 135 deferred tools are no longer available (MCP server disconnected): ListMcpResourcesTool, mcp__comfyui__* (40), **mcp__panel__* (92)**, ReadMcpResourceDirTool, ReadMcpResourceTool.

The reconnect notice brought back only **43**:

> 43 deferred tools are available again (MCP server reconnected): ListMcpResourcesTool, mcp__comfyui__* (40), ReadMcpResourceDirTool, ReadMcpResourceTool.

**All 92 `mcp__panel__*` tools never came back**, and nothing reported that as an error — the reconnect message reads like a successful recovery. The live-canvas surface (`panel_graph_outline`, `panel_set_widget`, `panel_run`, `panel_get_errors`, `panel_set_todo`, …) was simply gone for the rest of the session, in the middle of an active workflow-repair task where I had been using those tools successfully for many turns.

Verified absent in **both** surfaces, per the panel's own guidance, before concluding anything:
- `ToolSearch "select:mcp__panel__panel_graph_outline,mcp__panel__panel_get_errors,mcp__panel__panel_set_widget,mcp__panel__panel_set_todo"` → `No matching deferred tools found`
- A broad keyword sweep (`"panel graph outline canvas workflow live edit"`) → returned only `mcp__comfyui__*` tools, no `panel_*` at any rank.

This is not the known issue #291 tool-budget crowd-out: the panel tools were present and working for the entire session up to the respawn, so the budget accommodated them. Something in the respawn path rebuilt the session with only the comfyui MCP server attached and did not re-register the panel tool surface.

Practical impact: everything the panel exists for became unreachable mid-task, with no error and no remedy short of `/restart`. The user's canvas still had ~11 unresolved missing-model widgets I had been mid-way through fixing; I could no longer read the graph, update a widget, or queue a run, and correctly declined to substitute a headless workaround.

### Steps to reproduce

1. Work a session where `mcp__panel__*` tools are present and in active use.
2. Trigger a deferred tool-session respawn (e.g. `panel_request_secret` saving a comfyui-scoped key, which queues one).
3. Let the respawn fire.
4. Observe: the disconnect lists 135 tools including `mcp__panel__*` (92); the reconnect restores only 43 (comfyui + resource tools). No error is raised about the missing 92.

### Exact error (scrubbed)

No error. The failure is entirely silent — the reconnect message presents as success. The only evidence is the tool-count asymmetry between the two notices (135 out vs 43 back) plus `ToolSearch` returning no match for any `panel_` name afterwards.

### Fix

The session-rebuild path lives in the long-lived orchestrator process, which only reloads code on a full process restart — a local patch could not take effect this session, so I did not attempt one. Precise changes needed:

1. **Re-register the panel tool surface on respawn.** The rebuild should reattach every MCP server the prior session had, not just the comfyui one. Worth checking whether the respawn reconstructs the session from a config snapshot that omits the panel provider (the panel tools are supplied by the orchestrator itself rather than a configured stdio server, which would explain why only the configured server came back).
2. **Reconcile and report the delta.** Compare the pre-teardown tool set against the post-rebuild one and raise a loud error when a whole surface is missing, instead of emitting a reconnect notice that reads as success. A silent 92-tool loss is indistinguishable from a normal reconnect.
3. **Self-heal or tell the agent what to do.** If the panel surface cannot be reattached, the reconnect notice should say so explicitly and name the remedy (`/restart`), since the agent's own guidance is to never silently substitute a headless workaround for a canvas operation.

### Environment

OS: Linux (CachyOS, kernel 7.1.8-1) · ComfyUI 0.33.1 · ComfyUI **frontend 1.48.7** · GPU NVIDIA RTX 4070 SUPER (12 GB VRAM), 46 GB RAM · CUDA 13.0 / torch 2.13.0+cu130 · Python 3.13.12 · comfyui-mcp **0.51.45** · panel **0.14.28** · deploy env `local-desktop2-standalone` · backend: Claude · `COMFYUI_MCP_TOOL_MODE` not set

artokun:
Closes artokun/comfyui-mcp-panel#1228 · Refs #1524

## Root cause

A queued tool-session respawn (e.g. `panel_request_secret` saving a key) tears the agent session down and rebuilds it. On the rebuild the harness reconnected the `comfyui` stdio MCP child but never re-registered the in-process `panel` server — and the reconnect notice read as success, so all 92 `panel_*` tools stayed gone for the rest of the session with no error and no remedy short of a manual restart. 45b61e0 (#1614) already made that loss **loud** by reading the `system`/`init` `mcp_servers` report; this PR adds the **recovery** half the issue asks for.

## Fix

`src/orchestrator/claude-backend.ts` — both points that detect a down server now attempt recovery before speaking:

- `route()`'s init branch no longer announces immediately; it carries the degraded set out to `run()`'s loop (a sync generator can't await the control round-trip), which makes **one** `reconnectMcpServer` attempt per down server, re-reads `mcpServerStatus()`, and announces only what is still down (`settleInitMcpServers`).
- The first-turn-end settle for init-`pending` servers gets the same recover-then-report treatment.
- A server that went back to `pending` after the reconnect is neither declared down nor cleared — it is handed to the first-turn-end settle, which re-reads it once. The settle path does not re-arm, so a flapping server is never retried on every turn end.
- A recovered server is logged (`reconnect recovered MCP server(s): …`), not announced: the tool surface is intact, so there is nothing to degrade, and the harness's own reconnected notice already says so.

`src/orchestrator/mcp-session-health.ts` — `degradedMcpNotice` gains an optional `reconnectAttempted` flag that adds "An automatic reconnect was already attempted and did not bring it back." to the notice. Two honesty guards:

- Recovery is only claimed after being **read back** — a reconnect whose outcome cannot be verified (no control method, poll threw) changes nothing in the notice.
- The "already attempted" sentence only appears when the re-read confirmed the server is **still down** — otherwise the agent would retry the same remedy itself and call the retry a fix.

The manual remedy (Disconnect → Connect) stays in the notice, still without promising it fixes anything.

## Verification

- `npm ci` — clean
- `npm run lint` (tsc --noEmit) — clean
- `npx vitest run src/__tests__/orchestrator` — **159 files / 2788 tests, all green**, including:
  - 7 new recovery tests in `claude-init-mcp-degraded.test.ts` (reconnect fixes it → silent; still down → notice names the attempt; reconnect throws → still reported; unverifiable outcome → no "attempted" claim; healthy session → no control round-trips; settle-path recovery; pending-after-reconnect re-arm)
  - 2 new wording tests in `mcp-session-health.test.ts`
  - the existing settle-failed test updated: it now expects the reconnect attempt + verify poll (2 polls) before the notice

No tool descriptions changed, so `docs:gen` / `check:vocabulary` / `vocab:export --check` are not implicated.

Not fixed here (still tracked on #1524): *why* the respawn's rebuild leaves the panel server unregistered in the first place, and the orphan-orchestrator bind-or-exit ask. This change makes that failure self-heal when a reconnect suffices and name itself — with the remedy — when it does not.

### DIFF
--- a/src/__tests__/orchestrator/claude-init-mcp-degraded.test.ts
+++ b/src/__tests__/orchestrator/claude-init-mcp-degraded.test.ts
@@ -51,7 +51,14 @@ const hoisted = vi.hoisted(() => ({
   lastMcpServers: null as Record<string, unknown> | null,
   /** What `q.mcpServerStatus()` answers, and how many times it was asked. */
   statusPoll: null as null | Array<{ name: string; status: string }>,
+  /** Sequential poll answers: each mcpServerStatus() call shifts one entry, and
+   *  the static statusPoll answers once the queue is drained. Lets a test give
+   *  the settle poll and the post-reconnect verify poll DIFFERENT outcomes. */
+  statusPollQueue: [] as Array<Array<{ name: string; status: string }>>,
   statusPolls: 0,
+  /** `q.reconnectMcpServer()` calls, in order, and whether it throws. */
+  reconnectCalls: [] as string[],
+  reconnectThrows: false,
 }));
 
 vi.mock("@anthropic-ai/claude-agent-sdk", () => ({
@@ -66,10 +73,15 @@ vi.mock("@anthropic-ai/claude-agent-sdk", () => ({
       supportedCommands: async () => [],
       interrupt: async () => {},
       setModel: async () => {},
+      reconnectMcpServer: async (name: string) => {
+        hoisted.reconnectCalls.push(name);
+        if (hoisted.reconnectThrows) throw new Error("reconnect refused");
+      },
       mcpServerStatus: async () => {
         hoisted.statusPolls += 1;
-        if (!hoisted.statusPoll) throw new Error("no status available");
-        return hoisted.statusPoll;
+        const poll = hoisted.statusPollQueue.shift() ?? hoisted.statusPoll;
+        if (!poll) throw new Error("no status available");
+        return poll;
       },
     });
   },
@@ -79,7 +91,10 @@ beforeEach(() => {
   hoisted.queue.reset();
   hoisted.lastMcpServers = null;
   hoisted.statusPoll = null;
+  hoisted.statusPollQueue = [];
   hoisted.statusPolls = 0;
+  hoisted.reconnectCalls = [];
+  hoisted.reconnectThrows = false;
 });
 
 const initWith = (mcp_servers?: Array<{ name: string; status: string }>) => ({
@@ -264,7 +279,11 @@ describe("a server that was still connecting at init gets settled (#1524)", () =
     const notices = noticesOf(events);
     expect(notices).toHaveLength(1);
     expect(notices[0].message).toContain("panel");
-    expect(hoisted.statusPolls).toBe(1);
+    // #1228 — the failure was reconnect-attempted and verified still down before
+    // the notice went out: one settle poll, one reconnect, one verify poll.
+    expect(hoisted.reconnectCalls).toEqual(["panel"]);
+    expect(hoisted.statusPolls).toBe(2);
+    expect(notices[0].message).toMatch(/reconnect was already attempted/);
   });
 
   it("says nothing when it settled as connected", async () => {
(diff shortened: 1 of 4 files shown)


artokun:
Superseded by #1681 (merged). Both PRs addressed the same incident (panel#1228 / mcp#1524 twins), but #1681's per-turn-end mcpServerStatus watch strictly subsumes this PR's two detection points — critically, this PR checks only at init and the first turn end, so it stays blind to the reported case (a session that came up healthy and lost `panel` hours in), which #1681 detects. #1681 also adds episode semantics (one reconnect per down-episode), self-heal credit, and needs-auth/disabled retry exclusion. The one unique element here — not claiming "reconnect did not bring it back" when the post-attempt status could not be read back — was ported into #1681 as the "unverified" notice variant before merge. What this PR did slightly earlier (reconnect immediately after the init message rather than at the first turn boundary) was judged not worth a second code path. Thanks for the solid foundation — the adjudication diff is on the merged branch.

Issue #1417, opened 2026-08-11
Feature request: native Qwen Code CLI backend for Agent Panel


JusticeWay:
## Summary

Please add Qwen Code CLI as a first-class Agent Panel backend alongside Claude, Codex, Gemini, Antigravity, and the OpenAI-compatible local providers.

Qwen Code now exposes an agent-oriented local daemon and ACP bridge, so treating it as a generic OpenAI-compatible endpoint would lose its session, streaming, permission, cancellation, and MCP behavior.

## Current environment and verified behavior

- `qwen` version: `0.21.8`
- `qwen serve` is available as an experimental local HTTP daemon.
- Default listener: `127.0.0.1:4170`
- The daemon provides workspace-bound sessions, prompt submission, SSE replay, cancellation/lifecycle controls, optional bearer authentication, and an ACP child through `--http-bridge`.
- `qwen serve --help` says that Stage 2 native in-process mode is not implemented yet; current bridge mode preheats or spawns `qwen --acp` children.
- It is not an OpenAI `/v1/chat/completions` server, so the existing Custom endpoint backend is not a substitute.

## Why this belongs in `comfyui-mcp`

The Agent Panel backend owns more than text generation. It must preserve one session per panel tab, stream partial output, relay MCP and panel tools, handle permission questions, cancel or interrupt a turn, reconnect without duplicating a prompt, and report a terminal outcome. Qwen's daemon/ACP transport already exposes the primitives needed for that lifecycle.

An adapter also avoids placing OpenWebUI or another proxy between Agent Panel and Qwen. Such a proxy would discard Qwen's native agent session and tool semantics.

## Proposed scope

1. Add backend id `qwen` and a `Qwen Code` provider entry to the orchestrator and Panel picker.
2. Detect `qwen` / `qwen.cmd`, report installed/authenticated/readiness state, and show an actionable setup message when unavailable.
3. Prefer a managed `qwen serve --http-bridge --no-web` loopback daemon, or connect to a user-supplied daemon URL.
4. Map each Agent Panel tab to a distinct Qwen session and workspace.
5. Map Qwen streaming/session events into the existing Panel frames without fabricating a successful terminal reply.
6. Support interrupt, reconnect/resume, permission or `ask_user_question` flows, and bounded replay after a dropped SSE connection.
7. Inject the same headless ComfyUI MCP and per-tab `panel_*` MCP endpoints used by the other agent backends.
[...]

PR #1726, opened 2026-08-18
feat: add Atlas Cloud agent provider

binyangzhu000-sudo:
## Summary

- add Atlas Cloud to the registry-driven OpenAI-compatible provider family
- expose its API key slot, readiness state, backend label, model override, and credential rotation path
- document the provider setup and cover credential/readiness behavior with tests

The integration is opt-in through `ATLASCLOUD_API_KEY`; existing providers and defaults are unchanged. The default `qwen/qwen3.5-flash` model and `/v1/models` compatibility were verified against the live Atlas Cloud catalog without making a billable generation request.

## Validation

- `npm run build`
- `npm run lint`
- focused Vitest suite: 114 tests passed
- `npm run i18n:check`
- `npm run check:docs-links`
- `npm run check:docs-locale`
- `npm run check:docs-mdx`

`npm test` reaches four unrelated failures in `extra-paths-root-guard.test.ts` and `comfy-view-ref.test.ts`; the same four failures reproduce on the unmodified base commit `ae8301370ca21071cc8701cf93fbc0177d97196b`.


artokun:
Thank you for this — and specifically for the thing most provider PRs get wrong.

You didn't append `"atlascloud"` to the hardcoded list in `src/orchestrator/index.ts:3302`. You replaced the whole literal with the registry:

```diff
-  const KEYED_PROVIDERS = ["openrouter", "custom", "glm", "kimi", "moonshot", "minimax"];
+  const KEYED_PROVIDERS = ["openrouter", "custom", ...OPENAI_KEY_PROVIDER_IDS];
```

That list is the one place in the api-key family that had drifted back out of the registry, and it fails *silently* when it's stale — a key saved from the panel's API Keys card wouldn't invalidate the cached backend or re-push readiness, so the provider just sits there saying "not ready" with a valid key on disk. Every future entry now gets that for free. Your PR body is also unusually honest: naming the four pre-existing `extra-paths-root-guard` / `comfy-view-ref` failures and pinning them to base commit `ae83013` is exactly the right way to hand over a red `npm test`, and verifying `/v1/models` without burning a billable generation is a considerate touch.

## Verdict: parked under the active freeze

You're right that this is a parameterization rather than new machinery, and the diff proves it. `src/services/openai-provider-registry.ts` exists precisely so this is a data edit — everything downstream already derives from the table:

- `src/services/panel-secrets.ts:122` (env allowlist) and `:2367` (credential slots)
- `src/orchestrator/backend-readiness.ts:304` (`simpleKeyProvider`)
- `src/orchestrator/panel-console-http.ts:42` and `src/orchestrator/index.ts:1749` (both `KNOWN_BACKENDS` lists)
- `src/orchestrator/index.ts:1623` (boot log), `:2319` + `:2372` (backend factory + probe), `:3440` (current model), `:4056`/`:4142` (connect-ack)
- `src/orchestrator/ready-banner.ts:47`, `src/orchestrator/start-failure-notice.ts:53`

I checked that it reaches production rather than sitting as dead code: `simpleKeyAuth: true` means `simpleKeyProvider("atlascloud")` resolves, so both `makeBackend` and `getProbeBackend` route through `makeOpenAiKeyBackend`. It would genuinely work.

But being a parameterization isn't what decides this, and I want to be straight with you about why rather than let it look arbitrary. The standing ruling on this repo is in #1417, restated four hours after you opened this PR: *"parked, not dropped. The freeze is bugs and stability only, a new backend id and adapter is new capability."* That same comment calls a `qwen --acp` backend "a parameterization of that existing client, not a new transport" — and parks it anyway. Parameterization means the work is smaller than it looks when someone builds it; it isn't an exemption. The line the freeze draws is the **backend id**, and this PR adds one (`| "atlascloud"` in `src/orchestrator/agent-backend.ts`).

For contrast, #1494 (Grok 4.6) landed from an outside contributor on 2026-08-17, mid-freeze — it added a model to an existing backend's catalog and touched no `BackendId`. That's the shape that still moves right now.

**Parked means recorded and kept, not rejected.** This PR stays open, the branch stays yours, and the findings below are here so nothing has to be rediscovered. It is not a judgement on the work — the work is good. What unblocks it is the freeze lifting, which is the owner's call and not something I'll put a date on.

## Why it shows BLOCKED

Not a test failure. `main` requires four checks — `Build (macos-latest)`, `Build (ubuntu-latest)`, `Build (windows-latest)`, `Pack & install smoke` — and the rollup on this PR is **empty**: they haven't run at all. That's the standard fork-PR workflow-approval gate for a first-time contributor, not a red build. Nothing is failing.

## Findings, so the effort isn't wasted

**1. The locale catalogs — this is the silent one.** `locales/en/main.json` is a *generated but committed* artifact (`scripts/i18n-build-en.mjs:813-816`), and this PR doesn't touch `locales/`. Two live keys are missing from it:

- `banner.ready.atlascloud` — a literal key at your `trFor` call, picked up by the static scanner
- `say.degraded.atlascloud` — fanned out from `OPENAI_KEY_PROVIDERS` by the `DYNAMIC_SITES` entry at `scripts/i18n-build-en.mjs:577-603`

Compare MiniMax, which has both (`locales/en/main.json:16` and `:162`) plus eleven translations. Without them, `trFor` falls back to your inline English literal — so it renders *correctly in English* and is untranslatable in all twelve languages, with nothing anywhere reporting it. That's the exact failure `i18n-build-en.mjs` warns about in its own header.

Your `npm run i18n:check` green is a true result that simply doesn't cover this: `scripts/i18n-check.mjs` builds its schema **from** `locales/en/main.json` and validates the translations against it — it never compares English to the source. And `.github/workflows/ci.yml` doesn't run `i18n:build` at all, so CI wouldn't have caught it either. The fix is `npm run i18n:build` plus the committed catalog; the eleven translations then become required, which is why this lands better as part of a batch than bolted on here.

**2. Translated docs.** `docs/backends.mdx` is updated, but the MiniMax stanza it mirrors also exists in `docs/{ar,es,fr,ja,ko,pt-BR,ru,tr,zh,zh-TW}/backends.mdx`, as does the roll-up line at `docs/backends.mdx:181`. English-first is this repo's own pattern — locale docs land in batches (#1724, #1725) — so this is a note, not a defect.

**3. One that looks missed and isn't.** `src/orchestrator/pi-credentials.ts:70` carries a `minimax` entry and reads like a registration list. It isn't: its header says it's a faithful transcription of pi.dev's own `envMap`, so Atlas Cloud belongs there only if pi itself supports it. You were right to leave it alone.

**Clean otherwise.** No new dependencies (cold-start tree weight is a live constraint here — see #1447). Credentials go through `resolveKeyedCredentials` with no new storage location, and the boot diagnostic at `src/orchestrator/index.ts:1623` prints presence and source only, never values — your entry inherits both correctly.

Please don't spend more time on this while it's parked. Thanks again for a careful, well-scoped piece of work.

artokun:
Thanks @binyangzhu000-sudo — this is a clean, pattern-faithful provider addition. The review came back clean on both implementation and security, and the default model (`qwen/qwen3.5-flash`) was verified against the live Atlas Cloud `/v1/models` catalog.

The one blocker was i18n: the two new `trFor` call-site keys (`banner.ready.atlascloud`, `say.degraded.atlascloud`) had no locale entries, which fails the catalog byte-compare and the translation-coverage guards in CI. I've pushed a maintainer commit on top (no force-push — your commit is untouched) that:

- regenerates `locales/en/main.json` via `npm run i18n:build`,
- adds both keys, translated, to all 11 other locales (following the existing per-provider style in each),
- adds Atlas Cloud to the audio-capability family list in `docs/backends.mdx`,
- adds parity tests per the MiniMax precedent (#640): `resolveBackends("atlascloud")` label assertion and an `atlascloud` slot case in `panel-secrets-slots.test.ts`.

All gates pass locally: `npm run build`, `npm run i18n:check` (161 keys, 100% in all 12 locales), the catalogs byte-compare test, docs link/locale/MDX checks, and the full vitest suite. Once CI is green this is good to merge.

PR #1729, opened 2026-08-19
fix(1229): a too-old note with a current disk pack prescribes restart + hard-refresh, not a no-op sync

### LINKED ISSUE #1229: Stale-panel diagnosis tells agent to "sync the panel" when disk is already newer — real fix is a ComfyUI restart

The panel pack on disk was **0.14.37**, but the running (browser-loaded) panel reported **0.11.38**. ComfyUI-Manager had updated the pack IN PLACE at 11:57:26, ~5 min after ComfyUI started at 11:52, and logged:

```
After restarting ComfyUI, please refresh the browser.
```

Neither the restart nor the refresh happened, so ComfyUI kept serving the web assets registered at startup. Everything downstream then behaved like a much older panel.

The problem is the **advice**, not the staleness. Every affected tool reply said:

> this session's panel is 0.11.38, and a panel only reports the new workflow's identity ON THE REPLY from 0.11.45 onwards. ... **UPDATING THE PANEL to 0.11.45 or later removes this failure** rather than working around it. Update it with `install_comfyui (action:"panel", panel_action:"sync")`, then restart ComfyUI.

A sync is a no-op here — disk was already 0.14.37, far past 0.11.45. The orchestrator compares its required version against the **running** version and concludes "pack is out of date", when the actual condition was "pack on disk is NEWER than what the browser/server is serving". Those need different remedies (`sync` vs plain restart+hard-refresh), and the message names only the first.

Cost: I ran a full pack-version investigation (grep for PANEL_VERSION across installs, read `web/js/lib/add-node-widget-guard.js`, ruled out a second install) to discover the sync would have done nothing. A hard browser refresh alone also did NOT fix it — confirmed by re-running the failing call afterward and getting byte-identical old output — because the assets are registered server-side at startup.

**Suggested fix:** when reporting a panel version mismatch, stat the on-disk `PANEL_VERSION` in `custom_nodes/comfyui-agent-panel/web/js/comfyui-mcp-panel.js` and compare it to the running version. If disk >= required and disk > running, emit "the panel on disk (X) is newer than the one your browser is running (Y) — restart ComfyUI, then hard-refresh" instead of recommending a sync.

### Steps to reproduce

1. Start ComfyUI with panel version A loaded.
2. While it runs, update comfyui-agent-panel in place via ComfyUI-Manager to version B (B > A).
3. Do not restart ComfyUI.
4. Trigger any tool path that version-gates on the panel (e.g. `panel_save_workflow` with a name, then any `panel_*` graph call).
5. The reply recommends `install_comfyui(action:"panel", panel_action:"sync")` even though disk is already at B.

### Exact error (scrubbed)

Two symptoms, both explained by the stale load:

```
Error: workflow instance mismatch: this command targets a different workflow than
the active canvas — the workflow was switched or replaced after it was issued.
```
`panel_set_workflow_target({mode:"current"})` also explicitly failed to rebind ("did NOT restore this session's graph binding"); only `panel_open_workflow(path)` cleared it, once needing two attempts. Worst case: `panel_new_workflow` creates a BLANK UNSAVED canvas that is fenced with no path to reopen.

```
Error: Required custom widget "upload" did not initialize for "LoadImage".
Reload ComfyUI so its node extension can register, then retry.
```
This is the **pre-#700 message**. The on-disk 0.14.37 `web/js/lib/add-node-widget-guard.js` already replaced it with `describeUnmaterializedRequiredWidgets()` ("Cannot add \"LoadImage\": required input ... Nothing was added."), which is how I confirmed the browser was running pre-#700 code. Workaround that DID work: `panel_copy_nodes([existing LoadImage id])` + `panel_paste_nodes` cloned it with a functioning `upload` widget — so the widget registry was available, just not to the add-node path in the stale build.

### Fix

No local patch attempted — the code on disk is already correct; only the running instance is stale. The actionable upstream change is the version-comparison logic described above (compare on-disk PANEL_VERSION, not just running vs required).

### Environment

- ComfyUI: 0.33.0
- ComfyUI FRONTEND: 1.49.6
- GPU: NVIDIA GeForce RTX 5090, 32 GB VRAM (190 GB system RAM)
- comfyui-mcp: 0.51.47
- comfyui-mcp-panel: running 0.11.38 / on disk 0.14.37
- Python 3.12.9, PyTorch 2.10.0+cu130

Related (same root cause): #584, #562

artokun:
## Root cause

`panelTooOldNote` in `src/orchestrator/panel-tools.ts` compared the **running** panel's version against the minimum and, on `tooOld`, unconditionally prescribed `install_comfyui(action:"panel", panel_action:"sync")`. But the reporter's pack **on disk** was 0.14.37 while the session ran 0.11.38: ComfyUI-Manager had updated the pack in place after ComfyUI started, and ComfyUI keeps serving the web assets it registered at startup. A sync there is a no-op — disk already clears the floor — and following the advice cost the reporter a full pack-version investigation. The remedy that actually works for a stale *served* bundle is a ComfyUI restart plus a hard browser refresh (the reporter verified a hard refresh alone does NOT fix it, because the assets are registered server-side at startup).

## Fix

In the `tooOld` branch, before rendering the sync remedy, re-read the on-disk panel version with the same proof machinery the write gate uses (#774): `verifiedPanelDiskVersion()` re-reads the version NOW from the observed install dir and only returns it when the observation still describes the tree the live server serves. When that fresh disk version parses as semver and meets the minimum, the note instead says: DO NOT SYNC — the pack on disk (X) already meets the minimum; what is stale is what ComfyUI is serving; restart ComfyUI so it serves X, then HARD-REFRESH the tab, and a hard refresh alone does not fix this.

Fail-closed discipline is unchanged: no observation, an unparseable version, or a disk version genuinely below the floor all fall through to the original sync remedy, which is correct for a pack that really is behind. A missing observation kicks off a background `primePanelBase()` (never awaited — building an error message must not block on I/O) so a retry can answer.

## Verification

- `npm ci` — clean
- Full `npm test` chain (vitest 538 files / 10195 passed, plus i18n, docs-links, docs-locale, docs-mdx, blog checks) — exit 0
- New tests in `src/__tests__/services/panel-version-gap.test.ts` pin: the disk check uses `verifiedPanelDiskVersion()` and runs before the sync remedy; a disk version clearing the floor renders the DO-NOT-SYNC restart + hard-refresh note; a missing disk reading re-primes in the background without awaiting.

Closes artokun/comfyui-mcp-panel#1229

### DIFF
--- a/src/__tests__/services/panel-version-gap.test.ts
+++ b/src/__tests__/services/panel-version-gap.test.ts
@@ -124,6 +124,81 @@ describe("the rebind failure message carries the version gap (#1043)", () => {
     expect(body).toMatch(/if \(!v\?\.tooOld\) return ""/);
   });
 
+  // ── #1229: disk already current → sync is a no-op remedy ────────────────────
+  //
+  // The reporter's pack on disk was 0.14.37 while the session ran 0.11.38:
+  // ComfyUI-Manager had updated the pack in place after ComfyUI started, and
+  // ComfyUI keeps serving the web assets it registered at startup. Every
+  // affected tool reply prescribed `sync` — a no-op there — and the reporter
+  // had to run a full pack-version investigation to learn that. The remedy for
+  // a stale SERVED bundle is restart + hard-refresh, never a sync.
+
+  it("the too-old note checks the ON-DISK version before prescribing a sync", async () => {
+    const { readFileSync } = await import("node:fs");
+    const src = readFileSync(
+      new URL("../../orchestrator/panel-tools.ts", import.meta.url),
+      "utf-8",
+    );
+    const start = src.indexOf("function panelTooOldNote");
+    const end = src.indexOf("\n}", src.indexOf("if (!v?.tooOld) return \"\"", start));
+    const body = src.slice(start, end);
+
+    // The disk read must come from the same proof machinery the write gate
+    // uses (#774): a version re-read NOW from the observed install dir, never
+    // a recorded one trusted on its own.
+    expect(body).toContain("verifiedPanelDiskVersion()");
+    // …and it must run BEFORE the sync remedy is rendered, so the remedy can
+    // be overridden. Asserted on order, not just presence: a disk check added
+    // after the return would be dead code that still passes a grep. The search
+    // starts AT the disk check because the neverAdvertised branch above (#1560)
+    // legitimately names the sync remedy first — it is deliberately weaker and
+    // its remedy is not the one this override governs.
+    const diskCheck = body.indexOf("verifiedPanelDiskVersion()");
+    const syncRemedy = body.indexOf('panel_action:"sync"', diskCheck);
+    expect(diskCheck).toBeGreaterThan(-1);
+    expect(syncRemedy).toBeGreaterThan(diskCheck);
+  });
+
+  it("a disk version that clears the floor renders DO NOT SYNC, restart + hard-refresh", async () => {
+    const { readFileSync } = await import("node:fs");
+    const src = readFileSync(
+      new URL("../../orchestrator/panel-tools.ts", import.meta.url),
+      "utf-8",
+    );
+    const start = src.indexOf("function panelTooOldNote");
+    const end = src.indexOf("\n}", src.indexOf("if (!v?.tooOld) return \"\"", start));
+    const body = src.slice(start, end);
+
+    // The override fires only on positive proof: the disk version parses and
+    // meets the minimum. An unproven disk state must fall through to the sync
+    // remedy, which is correct for a pack that really is behind.
+    expect(body).toMatch(/SEMVER_RE\.test\(disk\) && compareSemver\(disk, v\.needed\) >= 0/);
+    // The note names what is on disk, says a sync changes nothing, and gives
+    // the remedy that actually works for the reporter's exact trap — including
+    // that a hard-refresh ALONE does not, which they verified the hard way.
+    expect(body).toContain("${disk}");
+    expect(body).toMatch(/DO NOT SYNC THE PANEL/);
+    expect(body).toMatch(/Restart ComfyUI so it serves \$\{disk\}/);
+    expect(body).toMatch(/HARD-REFRESH/);
+    expect(body).toMatch(/a hard refresh ALONE does not fix this/);
+  });
+
+  it("a missing disk reading re-primes in the background, never awaited", async () => {
+    const { readFileSync } = await import("node:fs");
+    const src = readFileSync(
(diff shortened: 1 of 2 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and the `claude-gate.mjs` fallback spawns a headless `claude` blocked on the account weekly limit. Under the owner's standing authorisation, this PR was gated by an **independent adversarial reviewer that had never seen the author's reasoning**, run against `.claude/autopilot/review-taxonomy.md`. Disclosure is the condition that substitution was accepted under. Verdict: **SHIP**. Head reviewed: `b5ba90a`.

- **The predicate is one-directional and fails closed.** Only `SEMVER_RE.test(disk)` and `compareSemver(disk, v.needed) >= 0` overrides; no observation, unparseable, or genuinely-behind all fall through to the untouched sync remedy. Both branches end in "restart ComfyUI", so the override can only REMOVE a step, never prescribe something more destructive in the other direction.
- **Checked against real shapes rather than an invented fixture** — this is where a wrong mechanism normally ships green. `verifiedPanelDiskVersion()` returns pyproject `[project].version` and `v.version` is the advertised `PANEL_VERSION`; on live panel `origin/main` those are the same value (pyproject 0.15.2, `comfyui-mcp-panel.js:1725` PANEL_VERSION 0.15.2), so the comparison is sound. `SEMVER_RE` (`ui-bridge.ts:338`) carries no global flag, so the `.test()` lastIndex trap does not apply, and the new runtime import creates no cycle.
- **Reachable:** `panelTooOldNote` is called at `panel-tools.ts:11785` and `:12296` via `describeFenceRebind`.
- **Weakness, named and not blocking:** all three new tests are source-scanning regexes over `panel-tools.ts`. They die with the fix (removing the block fails 3/3) but cannot catch a wrong runtime mechanism — nothing invokes `panelTooOldNote` with a stubbed disk read. The function is unexported and the file's existing #1043/#1560 tests use the same idiom, so this is house style over five lines of already-proven helpers.
- Smaller: the function's outer try/catch returns empty (`panel-tools.ts:4694`), so a throw from the new disk path would silently drop the whole too-old note. Failure mode is loss of advice, not wrong advice.
- The message asserts "the pack was updated after ComfyUI started" unconditionally; a first install or an earlier in-session sync produce the same state. The remedy is right in all of them — imprecise, not misleading.
- 15/15 panel-version-gap; 541/541 across `panel-tools` plus `ui-bridge`; `tsc` clean.

This PR is in the mcp repo while its issue is panel#1229, so the cross-repo reference will not auto-close it. Closing it manually.

PR #1734, opened 2026-08-19
fix(369): a download goes to the extra model root the SERVER named, not an unvouched install

artokun:
> **Rebased onto `main` after #1735 merged.** Two sessions independently opened rival PRs on the
> same report; the comparative gate ruled SHIP BOTH, with one hunk of this PR dropped. #1735 landed
> first (`3b98b01`), and this branch is now that base plus the two hunks the gate kept. See
> **"What was dropped, and why"** below.

## Root cause (the 0.52.1 recurrence)

`resolveModelsDirWithBases` picks the download destination from launch argv, the live install root,
or — when neither resolves — local configuration (`COMFYUI_PATH` / saved workspace). When only local
config or an INFERRED root is available, the pre-write guards (#1371, #1562) can only refuse on
POSITIVE disagreement and deliberately fail open on an empty tree (#1147). So on the 0.52.1
reporters' setups — ComfyUI Desktop launched with `--extra-model-paths-config` whose `base_path` is
the shared models root — `download_model` kept writing into the install-local / non-running second
install's `models/` tree while `list_local_models(action:"list_paths")` correctly showed the shared
root the server's own config declares.

Gigabytes landed in a tree the connected server does not read: the macOS report is four downloads,
~25 GB, into the non-running install; the Windows report is the install-local root while the shared
root is where models actually live.

## Fix (`src/services/model-resolver.ts`) — the two hunks this PR ships

- **`resolveModelSubfolderWithLiveRoot` (the redirect)** — when the primary models root is not one
  the server named (`modelsDirNamedByServer` is false) and argv carries an ABSOLUTE
  `--extra-model-paths-config` whose authoritative read maps the target MODEL category, the download
  goes to that server-named extra root instead. That flag is the server's own command line naming a
  root it reads — a statement, not an inference. Relative config values (unlocatable from this
  process), non-authoritative reads, missing category mappings, and code categories (`custom_nodes`)
  never redirect; a server-named primary is kept as before. The symlink/code-root veto still runs
  against the redirected root.
- **`isUnderLiveModelRoots` (the exception)** — roots from an argv-named config vouch for membership
  independently of the primary root's provenance, so a file landed there can verify `visible`.
  Without an argv-named config the early UNKNOWN stands: the auto-loaded config beside an INFERRED
  `main.py` inherits the inference and vouches for nothing. The unvouched primary still cannot
  support a NEGATIVE answer.

These two are the ONLY change here that moves the bytes. Probed on the Desktop shape (unvouched
primary plus an absolute `--extra-model-paths-config` mapping `loras`), the destination moves from
`…/ComfyUI-Installs/ComfyUI/ComfyUI/models/loras` to `…/ComfyUI-Shared/models/loras`.

## What was dropped, and why

The original branch also changed **`verifyLandedModel`** to set `liveModelsDir = undefined` for an
unvouched root, so the post-write verdict would fall through to the "move it" remedy instead of
claiming "it is in the right place — Do NOT move the file". **That hunk is dropped**; `main` keeps
#1735's `liveModelsDir = dest.modelsDir` + `liveModelsDirNamedByServer`.

The gate probed the dropped hunk against the #1374 input state (`source: "observed-root"`, root
correct, no `--extra-model-paths-config`, listing answered but stale). It emits a definite failure
verdict plus an instruction nobody can follow — *"it will not be usable in a workflow from there.
Move the file into the running server's models tree…"* with **no directory named**, because
`liveModelsDir` had just been suppressed. For the ordinary Windows-portable shape that is a false
failure on a download that landed exactly right, and it re-imports the exact harm #1131 was opened
about.

#1735 resolves that state properly: it is a THIRD case — inside a root nobody vouched for — so the
verdict says UNCONFIRMED, names both candidates (stale listing vs. different install), names the one
check that separates them, and still names the root the bytes are in. Both PRs were right about the
false assurance; only #1735's answer is followable.

Consequently the three `it.each(["configured-base","observed-root","base-anchored"])` cases in
`download-live-destination.test.ts` were re-pointed. They asserted `toMatch(/Move the file/)` — the
wording the dropped hunk produced. They now assert what `main` actually emits, read off a real run:
`UNCONFIRMED`, `STALE LISTING`, `DIFFERENT INSTALL`, `list_local_models (action:"list_paths")`, the
ABSENCE of `Move the file into the running server's models tree`, and the PRESENCE of the root the
file is in.

## Verification (on the rebased tree)

- `src/__tests__/services/` — **213 files, 5049 passed, 3 skipped, 0 failed.**
- `tsc --noEmit` — clean. `check:unknown-collapse` — clean (0 known sites, no new collapses).
- **Mutation, redirect hunk**: gating the redirect block off ⇒ **3 failures** (`downloads into the
  extra root the server's --extra-model-paths-config names`, `keeps the nested remainder under the
  extra root`, `never runs the stale-install disagreement guard against a server-named destination`).
- **Mutation, `isUnderLiveModelRoots` exception**: reverting to the plain
  `if (!primaryNamed) return { inRoots: undefined }` ⇒ **1 failure** (`post-write: a landed file
  under the server-named extra root verifies VISIBLE`). A second red in `training-jobs.test.ts` on
  that run was a 4.8 s timing test under concurrent load; it passes alone on the restored tree.

Both kept hunks are therefore still pinned by tests after the rebase.

## Known-open, recorded by the gate — NOT addressed here

Filed as follow-ups so they are not lost with this branch:

1. **Two config groups mapping one category** — the redirect takes `.find()`'s first match silently.
   A probe with two groups routed a download to an `a111` NAS root. It should refuse, or rank, rather
   than pick.
2. **`getLiveExtraModelRoots` reads the config's CURRENT contents**, not the contents at server
   start. A root added since launch can now *pick* the destination, not merely *authorise* one — the
   running server may not have read it yet.
3. **A redirected `visible` stamps `verified_root` with the unvouched primary** the file is not in
   (`model-resolver.ts:1788`).

## Deliberately unchanged

- The fail-open-on-empty-listing design of the pre-write guard (#1147's ruling) — with no
  server-named extra root and silent listings, an unvouched destination still proceeds, and the
  post-write report (now #1735's) is honest about it.
- `isLiveAuthoritativeModelsDir`'s wider meaning for its other callers.

Together with #1735 this addresses the underlying cause of #369; leaving that issue open until both
are on `main` and the placement has been confirmed on a real Desktop pair.

### DIFF
--- a/src/__tests__/services/download-live-destination.test.ts
+++ b/src/__tests__/services/download-live-destination.test.ts
@@ -905,8 +905,15 @@ describe("post-write: the reported path is VERIFIED, not intended (#369)", () =>
   // file — and the verdict STILL claimed "it is in the right place … Do NOT
   // move the file", because resolveModelsDir() returned that same configured/
   // inferred root and the lexical containment check against it passes for any
-  // file the download itself just wrote there. Only a root the SERVER NAMED may
-  // anchor the refresh remedy; anything else keeps the move remedy.
+  // file the download itself just wrote there.
+  //
+  // #1735 settled what REPLACES that claim, and it is not the move remedy: this
+  // state is the THIRD one (inside a root nobody vouched for), so the verdict says
+  // UNCONFIRMED, names both candidates — stale listing vs. different install — and
+  // names the ONE check that separates them. Asserting the move remedy here would
+  // re-import #1131's harm: on the ordinary Windows-portable shape this file landed
+  // exactly right, and "move it" with no destination is an instruction nobody can
+  // follow. These expectations are read off what main actually emits.
   it.each(["configured-base", "observed-root", "base-anchored"] as const)(
     "#369 (0.52.1): a not-listed file under a root the server never named is NOT called 'in the right place' (%s)",
     async (source) => {
@@ -918,7 +925,17 @@ describe("post-write: the reported path is VERIFIED, not intended (#369)", () =>
       expect(res.liveVisible).toBe("not-visible");
       expect(res.note).not.toMatch(/in the right place/);
       expect(res.note).not.toMatch(/Do NOT move the file/);
-      expect(res.note).toMatch(/Move the file/);
+      // The honest third answer, not either remedy (#1735).
+      expect(res.note).toMatch(/UNCONFIRMED/);
+      expect(res.note).toMatch(/STALE LISTING/);
+      expect(res.note).toMatch(/DIFFERENT INSTALL/);
+      expect(res.note).toMatch(/list_local_models \(action:"list_paths"\)/);
+      // Not the unfollowable move instruction — the destination is unknown here, so
+      // there is no tree to name, and #1131's reporter got exactly that sentence.
+      expect(res.note).not.toMatch(/Move the file into the running server's models tree/);
+      // It must still NAME the root the bytes are in: that is where the user looks,
+      // and a verdict that withholds it is not actionable either.
+      expect(res.note).toContain(resolve("/stale/ComfyUI/models"));
       // …and it must not name the unvouched root as "the models directory that
       // server reads" — that assertion is the false claim being fixed.
       expect(res.note).not.toMatch(/models directory (the connected ComfyUI|that server) reads/);
--- a/src/__tests__/services/download-live-destination.test.ts
+++ b/src/__tests__/services/download-live-destination.test.ts
@@ -135,6 +135,7 @@ vi.mock("node:fs/promises", () => ({
 import {
   currentLiveModelsRoot,
   resolveModelSubfolderPreferServer,
+  resolveModelSubfolderWithLiveRoot,
   verifyLandedModel,
 } from "../../services/model-resolver.js";
 import { ModelError } from "../../utils/errors.js";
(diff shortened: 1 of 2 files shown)


artokun:
Overlap notice + one finding, from a parallel session on the same issue (#1735, opened minutes after this one — we were both claimed on #369 and did not see each other).

The two PRs collide at exactly one function, `verifyLandedModel` → `notVisibleVerdict`. The rest of this PR (redirecting the destination to a server-named extra root) is complementary and #1735 does not attempt it.

**The finding.** This PR sets `liveModelsDir = undefined` when the root is unvouched. That makes `insideLexically` false and `knownInsideLiveRoots` false, so the unvouched case falls into the existing MOVE branch and the user is told:

> …the connected ComfyUI (…) does NOT list "…" under "…" — **it will not be usable in a workflow from there.** Move the file into the running server's models tree (or point COMFYUI_PATH at that install and re-download).

with no directory named, because `liveModelsDir` was suppressed.

For the ordinary Windows-portable shape measured in #1374 — correct `COMFYUI_PATH`, server reports a relative `main.py`, so the root is *unvouched but right* — that is a definite failure verdict plus an unfollowable remedy for a download that landed exactly where it should. It is #1131's harm on the other side of the same collapse: "we cannot confirm readership" became "it is misplaced", which is the same could-not-determine → determined-not fold the rest of this file spends so much effort avoiding.

The unvouched case has three possible truths and only one of them is misplacement; the honest branch has to say *unknown* and name the check that separates the two (`refresh_nodes`, then `list_local_models action:"list_paths"`). That is what #1735 adds, with the #1131 server-named branch left intact and a mutation run pinning both.

Whichever of these lands first, the other should be rebased onto it rather than merged over it. No action needed from this PR's author if #1735 lands second — I will rebase.

artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. This PR's *design* was approved by an **independent comparative reviewer** that judged it against the rival PR #1735 and ruled SHIP BOTH with a precise rebase instruction. The rebase itself was then verified mechanically rather than on the rebasing agent's word.

**Verdict: SHIP**, head `0232bfc`.

**What changed in the rebase, and why.** The comparative gate found this PR's `verifyLandedModel` hunk set `liveModelsDir = undefined` for an unvouched root, routing that case into the MOVE remedy. Probed with the #1374 input state (`source: "observed-root"`, root correct, no `--extra-model-paths-config`, listing answered but stale) it emitted a definite failure verdict plus an unfollowable instruction — "Move the file into the running server's models tree…" with **no directory named** — on a download that had landed exactly right. That hunk is **dropped**; the verdict path now comes from #1735, already on main.

**The two hunks that matter are kept unchanged, and they are the only fix for the reporter's misplaced 25 GB:**
- the `resolveModelSubfolderWithLiveRoot` redirect (`src/services/model-resolver.ts:2141-2229`)
- the `isUnderLiveModelRoots` exception (`:1350-1402`)

Verified by probe: with the Desktop shape (unvouched primary plus an absolute `--extra-model-paths-config` mapping `loras`), the destination moves from `…/ComfyUI-Installs/ComfyUI/ComfyUI/models/loras` to `…/ComfyUI-Shared/models/loras`.

**Independently checked before merging:** `git diff origin/main <branch> -- src/services/model-resolver.ts` returns **zero** lines touching `notVisibleVerdict` or `liveModelsDir` — the verdict path is byte-identical to main, so the dropped hunk really is gone rather than merely reworded. Body scanned for auto-close keywords: none.

**Test expectations were rewritten from main's real output, not guessed.** The `it.each` block (now at `download-live-destination.test.ts:903-943`; #1735's additions shifted it ~64 lines) previously asserted `toMatch(/Move the file/)`. It now asserts `/UNCONFIRMED/`, `/STALE LISTING/`, `/DIFFERENT INSTALL/`, `/list_local_models \(action:"list_paths"\)/`, plus `not.toMatch` on the unfollowable sentence and `toContain(resolve("/stale/ComfyUI/models"))` — because the root must still be *named*, which was the gate's actual finding.

**Mutations re-run on the kept hunks, both still pinned:** gating off the redirect fails 3 tests; reverting the `isUnderLiveModelRoots` exception to `if (!primaryNamed) return { inRoots: undefined }` fails 1. 213 files / 5049 passed / 0 failed; `tsc --noEmit` and `check:unknown-collapse` clean. CI green on all three OSes plus pack-install smoke.

Three follow-ups carried in the body rather than fixed here: `.find()` silently takes the first match when two config groups map one category; `getLiveExtraModelRoots` reads the config's *current* contents, so a root added since server start can pick the destination rather than merely authorise one; and a redirected `visible` stamps `verified_root` with the unvouched primary the file is not in (`model-resolver.ts:1788`).

Issue #1131, opened 2026-08-08
download_model reports a correctly placed local model as unusable and tells user to move it to the same directory


artokun:
### What happened / root cause
After a successful local download to the connected ComfyUI's exact models root, download_model immediately checked the server's stale loader option list and concluded the model was not visible. The response then instructed moving the file into the running server's models tree, even though the reported written path and server models path were identical. The actual required action is refreshing node/model definitions (or restarting), not moving the file.

### Steps to reproduce
1. Use download_model action=download against a local connected ComfyUI.
2. Save a new LoRA to target_subfolder=loras.
3. Observe the tool verifies the file on disk under the server's exact models/loras directory.
4. Before the server option list refreshes, observe NOT VISIBLE plus an instruction to move it into the same models tree.

### Exact error (scrubbed)
Written path: <COMFYUI_PATH>\models\loras\example.safetensors
Server models path: <COMFYUI_PATH>\models
Message: Move the file into the running server's models tree.

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: when the verified destination is already under the connected server's models root, never recommend moving it. Refresh/re-query model definitions, or report that panel_refresh_nodes / a ComfyUI restart is required. Ideally wait briefly and recheck after a refresh before declaring unusable.

### Environment
Windows 11; ComfyUI 0.27.0 LOCAL; NVIDIA RTX 5070 12 GB; comfyui-mcp 0.49.3; panel 0.11.28; Manager 3.x LEGACY; Python 3.12; torch 2.8.0; CUDA 12.9.

artokun:
Fixed in #1137 — thank you, the two identical paths in your report were the whole diagnosis.

You were right that the instruction could not be followed, and the reason is that "not listed" has two causes that the tool had collapsed into one. ComfyUI caches its loader option lists and invalidates them on the directory's mtime, so a placement check run immediately after a write routinely races the re-read. Your file was in exactly the right place; the server just had not looked again yet.

Now:

- **Inside the models root the server reads** → it says the placement is right, names the stale listing as the cause, says *do not move the file*, and points at `install_comfyui(action:"refresh_nodes")`, the panel's refresh, or a restart.
- **Outside that root** → the move advice, unchanged, because there the file really is in the wrong place.

[...]

PR #1735, opened 2026-08-19
fix(369): containment in a root WE picked is not readership — an unvouched destination is UNCONFIRMED, not "in the right place"

artokun:
## The recurrence this fixes

The 2026-08-18T22:51Z comment on #369, on comfyui-mcp **0.52.1** / panel **0.15.1** — i.e. a
build that already contains #794, #1562 and #1679. macOS, two installs under
`~/ComfyUI-Installs/`, the live one a ComfyUI Desktop server reached through
`--extra-model-paths-config`:

> four downloads were written to the install-local `models/` tree while the connected
> server's corresponding `/models/<category>` responses were **empty**, despite `status`
> saying the server listed each file. […] roughly 25 GB was misplaced

and, said precisely in the sibling report two comments earlier:

> `status` **incorrectly said the guessed install-local directory was read by the connected
> server and instructed not to move the file**.

That second sentence is a claim the code makes, and it is the half fixed here.

## Mechanism

`verifyLandedModel` ends at `notVisibleVerdict`, which decides between two opposite remedies
by asking whether the landed file is inside "the models directory the connected ComfyUI
reads". It answered that with

```ts
liveModelsDir = await resolveModelsDir();
const insideLexically = liveModelsDir !== undefined && isUnderRoot(verifiedPath, liveModelsDir);
```

`resolveModelsDir()` is **provenance-blind** — it is `(await resolveModelsDirWithBases()).modelsDir`
and returns the root whatever its `source`, including `configured-base` and `observed-root`,
the two values a reachable server never vouched for. And that root is *the one this download
resolved to*, so the landed file is inside it **by construction**.

So the comparison is `true` for a correct destination and for a stale second install alike:
its two operands are both the choice we made. It cannot distinguish the case it exists to
distinguish (taxonomy class 1 — two states, one indistinguishable answer, no way to say
*unknown*; and class 3 — a guard that is present, correct in isolation, and compares the
wrong pair). The reporter's four downloads therefore each got

> …which is INSIDE the models directory the connected ComfyUI reads (`…/ComfyUI-Installs/ComfyUI/ComfyUI/models`)
> — so it is in the right place. […] **Do NOT move the file.** Refresh the node/model definitions…

for bytes the running server will never scan. The empty `/models/<category>` listing did not
contradict it, because an empty listing is (correctly, #1147) not disagreement — the *guard*
was silent, and the *report* filled the silence with a guess.

## The change

`src/services/model-resolver.ts`, two edits:

1. **The source travels with the path.** `verifyLandedModel` resolves through
   `resolveModelsDirWithBases()` and passes `modelsDirNamedByServer(dest.source)` into the
   verdict. Deliberately `modelsDirNamedByServer`, not `isLiveAuthoritativeModelsDir`: an
   `observed-root` is *inferred* from where the interpreter lives, and that is exactly the
   inference #1562 narrowed for.

2. **"Unknown" becomes representable.** `notVisibleVerdict` gains a third branch. Containment
   in a **server-named** root keeps the #1131 verdict unchanged ("in the right place, do NOT
   move it, refresh"). Containment in an **unvouched** root now says so:

   > The file IS on disk at …, inside … — the models root this download resolved to. Whether
   > the connected ComfyUI READS that root is **UNCONFIRMED** … That server does not list "…"
   > under "…".
   > TWO very different things produce exactly this, and only one is a problem: 1. a **STALE
   > LISTING** … 2. a **DIFFERENT INSTALL** …
   > Do not move it on the strength of this message alone, and do not treat it as placed
   > either. SEPARATE THE TWO: refresh the node/model definitions … then re-check with
   > `list_local_models`. If it STILL does not appear, it is case 2: `list_local_models
   > (action:"list_paths")` reports the roots the running server actually reads — move the
   > file into the one for "…".

   `list_paths` is named because it is the discriminator the reporter used by hand: on their
   box it correctly showed the shared root while the download went elsewhere.

   The unvouched branch is deliberately **not** the move remedy either. #1131's harm was an
   unfollowable instruction to move a file into the directory it was already in; declining to
   assert readership is not evidence of misplacement. And for the ordinary Windows-portable
   shape (#1374) — correct `COMFYUI_PATH`, server reporting a relative `main.py`, so the root
   is unvouched but right — flipping to "it will not be usable in a workflow from there, move
   it" would be a false failure for a working setup.

`knownInsideLiveRoots` is untouched and still sufficient on its own: it comes from
`isUnderLiveModelRoots`, which applies the same narrowing internally and additionally covers
live extra roots and #870 junctions. The `liveVisible` verdict is unchanged in every branch —
what changes is what the note claims.

## Refutation — what I deleted, and what failed

Baseline green first (78/78 on the two files), then four mutations against it:

| # | mutation | result |
|---|---|---|
| 1 | `insideLiveRoot = insideLexically \|\| knownInsideLiveRoots` (the pre-change line) | **5 failed** — both `verifyLandedModel` integration tests + 3 verdict tests |
| 2 | delete `liveModelsDirNamedByServer,` from the call site | **survived 78/78 at first** — see below |
| 3 | `modelsDirNamedByServer(dest.source)` → `isLiveAuthoritativeModelsDir(dest.source)` | **1 failed** — the `observed-root` case |
| 4 | `insideUnverifiedRoot = false` | **5 failed** |

Mutation 2 is the finding. Dropping the wiring flipped nothing, because
`knownInsideLiveRoots` already answers the healthy case — only the *wording* degraded
silently: a plain server-named root began telling users their layout was "reached through a
link/junction", and the move remedy stopped naming the directory that makes it followable.
That is the "an install line is invisible to tests" shape, and a helper-level test cannot see
an argument go missing. The second commit adds the two call-site assertions; re-running
mutation 2 now fails 2 tests.

Also run: `npm run typecheck` clean, `npm run lint` clean, `npm run check:unknown-collapse`
clean ("0 known site(s), no new collapses"), and the full suite — 10191/10205 passed. The
residue is not mine: `package-hooks > every entry in files exists on disk` fails in a fresh
worktree because `dist/` has not been built (`expected [ 'dist' ] to deeply equal []`), and
`tool-surface-filter` / `vocabulary` were 30s timeouts under load that pass in isolation
(96/97 on a serial re-run of the three files, the one failure being the `dist` one).

## Relationship to #1734

#1734 opened at 04:17Z while this was in flight, and the two overlap at exactly one function.
Its `verifyLandedModel` change sets `liveModelsDir = undefined` for an unvouched root, which
routes that case into the existing **move** branch:

> …the connected ComfyUI does NOT list "…" — it will not be usable in a workflow from there.
> Move the file into the running server's models tree…

with no directory named. For the #1374 Windows-portable shape (correct `COMFYUI_PATH`, server
reports a relative `main.py`) that is a definite failure verdict and an unfollowable remedy
for a download that landed correctly — #1131's harm, on the other side of the same collapse.
The third branch here is what that case needs; the rest of #1734 (redirecting the destination
to a server-named extra root) is complementary and not attempted here. Whichever lands first,
the other should be rebased onto it rather than merged over it.

## Deliberately not done

- **The pre-write guard is untouched.** `assertDestinationVisibleToLiveServer` still fails
  open on an empty listing and an empty candidate tree. That is #1147's maintainer ruling
  ("unaccountable is a proceed-unverified state"), and broadening it is how the guard
  false-refuses fresh and shared trees. This PR makes the *report* honest, not the guard
  stricter.
- **No panel code changed**, so the panel Playwright suite was not run; nothing in
  `comfyui-mcp-panel` references these strings (grepped).
- Gate scripts were not run — `codex exec` is account-exhausted until 2026-08-19 20:43 and
  `claude-gate.mjs` is blocked on the weekly limit, so both would return INDETERMINATE. This
  PR is submitted for an independent reviewer to gate.

The underlying cause of #369 is what this addresses; closing is the maintainer's call.

### DIFF
--- a/src/__tests__/services/download-live-destination.test.ts
+++ b/src/__tests__/services/download-live-destination.test.ts
@@ -592,6 +592,52 @@ describe("post-write: the reported path is VERIFIED, not intended (#369)", () =>
     expect(res.note).toContain(resolve("/live/ComfyUI/models"));
   });
 
+  // #369, the 0.52.1/panel-0.15.1 recurrence (macOS, Comfy Desktop extra model
+  // paths). Four downloads landed in the install-local models/ tree of a SECOND
+  // install while the connected server's /models/<category> answered EMPTY, and
+  // the tool reported that the directory was one the connected server reads and
+  // that the file should not be moved. ~25 GB.
+  //
+  // The verdict's containment test ran against whatever the models-dir resolver
+  // returned, provenance and all — and that root is the one THIS DOWNLOAD RESOLVED
+  // TO, so the landed file is inside it by construction. The comparison therefore
+  // answers "inside" identically for a correct destination and for a stale second
+  // install: taxonomy #1, two states collapsed into one answer.
+  //
+  // This exercises the WIRING (verifyLandedModel → notVisibleVerdict), not just the
+  // wording helper: the provenance has to be READ from the resolution and PASSED,
+  // and a helper-level test cannot see that argument go missing.
+  it("#369: does NOT claim readership of a root the server never NAMED (0.52.1 recurrence)", async () => {
+    const staleLanded = resolve("/Users/u/ComfyUI-Installs/ComfyUI/ComfyUI/models/loras/new.safetensors");
+    h.modelsDirSource = "configured-base";
+    h.destModelsDir = "/Users/u/ComfyUI-Installs/ComfyUI/ComfyUI/models";
+    h.liveListings["loras"] = []; // the reporter's observation: answered, and EMPTY
+    const res = await verifyLandedModel(staleLanded, "loras", { attempts: 1, retryMs: 0 });
+    expect(res.liveVisible).toBe("not-visible");
+    expect(res.note).not.toMatch(/INSIDE the models directory the connected ComfyUI reads/);
+    expect(res.note).not.toMatch(/it is in the right place/);
+    expect(res.note).not.toMatch(/Do NOT move the file/);
+    expect(res.note).toMatch(/UNCONFIRMED/);
+    // Both candidates named, plus the check that tells them apart.
+    expect(res.note).toMatch(/STALE LISTING/);
+    expect(res.note).toMatch(/DIFFERENT INSTALL/);
+    expect(res.note).toMatch(/list_paths/);
+  });
+
+  it("#369: an INFERRED (observed-root) destination gets the same unconfirmed verdict", async () => {
+    // #1562's narrowing again: an interpreter's location is evidence about the
+    // BINARY, not about where the server reads. isLiveAuthoritativeModelsDir()
+    // says yes to `observed-root`; the claim here requires modelsDirNamedByServer.
+    const staleLanded = resolve("/stale/ComfyUI/models/loras/new.safetensors");
+    h.modelsDirSource = "observed-root";
+    h.destModelsDir = "/stale/ComfyUI/models";
+    h.liveListings["loras"] = ["someone-elses.safetensors"];
+    const res = await verifyLandedModel(staleLanded, "loras", { attempts: 1, retryMs: 0 });
+    expect(res.liveVisible).toBe("not-visible");
+    expect(res.note).toMatch(/UNCONFIRMED/);
+    expect(res.note).not.toMatch(/it is in the right place/);
+  });
+
   it("still says MOVE IT when the file is outside the live models root", async () => {
     // The write path itself is outside the live root (no junction involved), so
     // membership cannot vouch for it and the remedy stays "move it". (A write
--- a/src/__tests__/services/download-live-destination.test.ts
+++ b/src/__tests__/services/download-live-destination.test.ts
@@ -590,6 +590,13 @@ describe("post-write: the reported path is VERIFIED, not intended (#369)", () =>
     expect(res.note).toMatch(/Do NOT move the file/);
     expect(res.note).not.toMatch(/Move the file into the running server's models tree/);
     expect(res.note).toContain(resolve("/live/ComfyUI/models"));
+    // #369 (0.52.1) — the SERVER named this root, so the note may say so, and must
+    // use the plain wording. If the provenance stops reaching the verdict, this
+    // degrades to the junction phrasing and tells a user with no junction that
+    // their layout is "reached through a link/junction" — prose asserting a
+    // mechanism that is not there.
+    expect(res.note).toMatch(/INSIDE the models directory the connected ComfyUI reads/);
+    expect(res.note).not.toMatch(/link\/junction/);
   });
 
   // #369, the 0.52.1/panel-0.15.1 recurrence (macOS, Comfy Desktop extra model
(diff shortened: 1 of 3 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Under the owner's standing authorisation, this was gated by an **independent adversarial reviewer that had seen neither author's reasoning**. Because two sessions independently opened rival PRs on #369, the gate was **comparative** — it reviewed this PR against #1734 and ruled on both.

**Verdict: SHIP BOTH — this one first.** They are complementary on the destination and conflicting only on the verdict: **only #1734 fixes where the bytes go; only #1735 fixes what the report claims.** `git merge` produces exactly one conflict, at `verifyLandedModel`'s `liveModelsDir` assignment; nothing else collides.

**This PR's author criticised #1734, and the gate verified that criticism by execution rather than adopting it.** It holds. Probing `origin/fix/369-extra-model-paths-dest` with the #1374 input state (`source: "observed-root"`, root correct, no `--extra-model-paths-config`, listing answered but stale) produces a definite failure verdict plus an unfollowable remedy — "it will not be usable in a workflow from there. Move the file into the running server's models tree…" with **no directory named**, because `liveModelsDir` was suppressed at `model-resolver.ts:1880`. #1734's own test asserts that across all three unvouched sources, so it is intentional rather than an oversight — and it is the #1131 harm pointed the other way, on a download that landed exactly right.

**This PR does not commit the mirror harm.** Its third branch (`model-resolver.ts:1957-1983`) names case 2 as a different install, then gives refresh → `list_local_models(action:"list_paths")` → move the file into the one for the category. `liveVisible` stays `"not-visible"` in every branch, so `describePlacement`'s `wrongPlace: true` (`download-jobs.ts:1195`) is unchanged: the warning survives, with a discriminator attached. A user who genuinely must move 25 GB still gets told.

**Reachability, all live:** `verifyLandedModel` ← `download-jobs.ts:722`; `notVisibleVerdict` ← only `verifyLandedModel:1862`; `isUnderLiveModelRoots` ← `verifyLandedModel:1706`/`:1767` and `manifest.ts:1237`.

**Mutations, including the one that mattered.** The author disclosed that their fourth mutation — deleting the call-site argument — originally survived 78/78 green, and added call-site assertions in a second commit. The gate confirmed commit `fba1ca0` exists and re-ran all four: N1 fails 5, **N2 now fails 2** (was green), N3 fails 1, N4 fails 5. Full `src/__tests__/services/` green (213 files, ~5027 tests); `tsc`, `lint`, `check:unknown-collapse` clean.

**Next step, recorded so it is not lost:** #1734 will be rebased onto this, **dropping its `verifyLandedModel` hunk** (keeping `liveModelsDir = dest.modelsDir` and `liveModelsDirNamedByServer` from here) and replacing its three `it.each(["configured-base","observed-root","base-anchored"])` assertions in `download-live-destination.test.ts:846-869` with the UNCONFIRMED expectations. Its redirect and `isUnderLiveModelRoots` hunks ship unchanged — they are the only fix for the reporter's misplaced 25 GB.

**#369 stays OPEN until #1734 lands**, since the destination half is what actually stops the misplacement.

Three follow-ups the gate recorded, none blocking: with two config groups mapping one category, `.find()` silently takes the first (a probe routed a download to an `a111` NAS root); `getLiveExtraModelRoots` reads the config's *current* contents, so a root added since server start can now pick the destination rather than merely authorise one; and a redirected `visible` stamps `verified_root` with the unvouched primary the file is not in (`model-resolver.ts:1788`).

Issue #1646, opened 2026-08-16
Prevent workflow mismatch auto-rebind from routing later edits to the wrong canvas


artokun:
### What happened / root cause
panel_open_workflow reported that B1 SINGLE-IMAGE VIDEO was opened and panel_graph_outline then returned that graph, but the first subsequent graph mutation was rejected with a workflow-instance mismatch. The orchestrator auto-rebound the session to the live A0 canvas, so later mutations in the same client sequence were accepted against A0. This can corrupt the wrong workflow after an open/identity race.

### Steps to reproduce
1. Rename two open workflows.
2. panel_open_workflow the renamed target.
3. panel_graph_outline confirms the target.
4. Send several awaited panel mutations.
5. First mutation reports target UUID vs active UUID mismatch and AUTO-REBIND to another open tab; subsequent mutations target that other tab.

### Exact error (scrubbed)
graph_edit_node was NOT applied: workflow instance mismatch. Command issued for workflow instance 88564247-..., active canvas reports 81474df4-.... AUTO-REBIND re-derived the fence onto the live canvas. The active workflow last moved to A0 by panel_open_workflow(B1...).

### Fix
applied locally: no
upstream-only: yes
Precise change needed: panel_open_workflow must not report success until the active canvas identity is stably the requested routing key. On a later mismatch, never auto-rebind mutation routing to a different workflow; keep all writes refused until an explicit panel_set_workflow_target(mode=current) or a successful open+pin. A mismatch response should preserve the requested target fence so subsequent calls cannot mutate another tab.

### Environment
Windows 11; ComfyUI 0.33.1; frontend 1.48.7; Python 3.13.12; PyTorch 2.12.1+cu130; RTX 4080 16GB; comfyui-mcp 0.51.17; panel 0.14.16.

PR #1736, opened 2026-08-19
fix(1519): a fenced graph READ is diagnosed, and a missing stamp is not a wrong one

artokun:
A fenced graph **READ** is refused and the orchestrator adds nothing to it, so the two states behind that refusal reach the caller as one. This addresses the underlying cause of #1519.

## What I measured first

Driving `panel_graph_outline` through the real `ctx.call` path on `origin/main@a649ff7`, against a panel that refuses the unstamped command, the **entire tool result** is:

```
Error: workflow instance mismatch: this command carries no workflow-instance stamp,
and the active canvas reports 2b3f4684-…. Nothing was applied.
```

That is verbatim what #1519 was filed with. #1330's corroboration is gated on `isMutatingGraphCmd`, so a READ refused by the very same fence falls through every branch of the catch and **this side adds nothing at all**. #1480's guard already extends its diagnosis to reads "on purpose: `panel_graph_outline` refusing was half of the reported dead end"; this is the same reasoning for the stamp fence.

**What the panel says on its own is not nothing, and the difference is the point.** Reading `workflowInstanceMismatchMessage` on `comfyui-mcp-panel@origin/main` rather than the short form quoted in the report: since panel **0.11.83** the refusal ends

> Re-target with `panel_set_workflow_target({mode:"current"})`, or re-select the intended workflow with `panel_open_workflow`, then retry.

Both exits, offered as interchangeable, with nothing said about which one this refusal calls for. That is correct *for the panel* — by design it "observed only that the two identities differ" and refuses to infer a cause — and it is exactly the collapse below. In the #1331 state the first of the two cannot work at all: it reports success while the read keeps failing. Only the orchestrator can take the read that decides, so it takes it.

## Mechanism

`isWorkflowInstanceMismatch` matches **two** panel states with **opposite** remedies:

| the panel said | what it means | what fixes it |
| --- | --- | --- |
| `carries no workflow-instance stamp` | this session has **no** identity; nothing was compared | derive a fence from the live canvas |
| `issued for workflow instance <uuid>` | this session **has** an identity the canvas disagrees with | choose which workflow you meant — deriving from the live canvas **abandons** the one you named |

Collapsing them hands the second case the first case's remedy, which is the retarget #1646 removed for cause. So the shape is read from the panel's own words, each state names only its own remedy, and a wording matching **neither** is reported as unknown rather than guessed into one of the two.

Read from the **refusal**, never from `cmd.workflow_uuid`: the stamp is applied downstream of the catch, so that field is `undefined` for *both* states — the trap that silently disabled #1330's transient branch one block up.

Also handled: the live canvas having no identity **of its own** (#1331's lesson — `mode:"current"` reports success while the read keeps failing, so it is not offered), an unreadable probe (says UNKNOWN), and a **pinned** session (naming `mode:"current"` also discloses that it releases the pin, with the `panel_open_workflow` alternative).

## What this does NOT do

- **No position on the open design question** recorded in `ui-bridge.ts` — whether an unstamped read should be *admitted*. Nothing about what is dispatched, or about the panel's fence, changes.
- **Nothing is adopted.** The probe is the read-only `rebindWorkflowFence({adopt:false})`; the refusal still fails the call; the fence moves only on an explicit rebind.
- The mutation path is untouched (a control test pins its distinct `NOT applied — nothing changed` wording).

## Refutations I ran

Each one is a deletion, run against the new suite:

1. **Disable the branch** (`if (false && …)`) → **6 of 7 fail** (of the suite as it stood then). The survivor is the mutation-path control, which correctly stays green.
2. **Collapse absent and stamped into one state** (hard-code `shape = "unstamped"`) → **3 fail**: the WRONG-STAMP case, the UNSTATED case, and the WRONG-STAMP case driven from the panel's full real message. The anti-collapse property is what the assertions are made of — each case asserts its own remedy *and* asserts the other case's remedy is absent, so a verdict that merely mentioned rebinding cannot pass.
3. **Remove the re-entrancy guard** (drop the `graph_` prefix from `isFencedGraphRead`) → the diagnosis re-enters itself through its own `workflow_list` probe and **the vitest worker dies with a fatal error**. The guard is load-bearing; a panel predating the #759 probe exemption is the input that reaches it.
4. **Re-add `workflow instance mismatch` to the transient classifier** (which really does re-issue the read) → the corrected `switch-guard-retry` assertion still fails, so that harness change tightened the test rather than loosening it.

The classifier is also driven against the panel's **full message transcribed verbatim from v0.15.3**, both shapes — so the extra prose (which itself names both remedies) cannot flip the regexes. And the `no_identity` verdict is driven from *both* shapes: it is reachable from a stamped session too, so it says "no workflow identity **of its own**" rather than "either", which would be false for that half.

`switch-guard-retry.test.ts` needed a harness correction: its fixture counted **all** sends, so the diagnosis round trip read as a retry of the command under test. Its own prose asserts the first thing ("an instance mismatch is NOT retried"), so the fixture now records *which* command it was asked to send and the assertion is made about `graph_outline` itself.

## Cost

One read-only `workflow_list` on an already-failed read (plus the existing #1292 rechecks only when corroboration fails) — the same trade the mutation path has made since #1330: one informed round trip instead of N blind ones.

## Verification

- `tsc --noEmit` clean; `npm run build` clean.
- `npm run check:vocabulary`, `npm run check:unknown-collapse`, `npm run vocab:export -- --check`, `node scripts/asset-counts.mjs --check` — all pass locally.
- Full `vitest run`: the new file and `switch-guard-retry` are green. Unrelated failures on this loaded Windows box (`gen-changelog-dedupe`, `turn-watchdog`, `tool-surface-filter`, `vocabulary`, `training-jobs`) are all 30s wall-clock timeouts and pass in isolation; none import `panel-tools`. Leaving those to CI.
- **Not run:** the panel Playwright suite — this change is orchestrator-side only and renders nothing in the sidebar, so there is no panel-visible behaviour to cover.

## Gate

codex was unavailable (account exhausted until 2026-08-19); this PR is left for an independent adversarial review rather than self-gated. I did not run any gate script and did not merge.

### DIFF
--- a/src/__tests__/orchestrator/fenced-read-absent-stamp.test.ts
+++ b/src/__tests__/orchestrator/fenced-read-absent-stamp.test.ts
@@ -176,6 +176,20 @@ describe("a fenced graph READ is diagnosed, and a missing stamp is not a wrong o
     expect(adopted).toEqual([]);
   });
 
+  it("NO IDENTITY, but the session HAS a stamp: the verdict does not claim the session lacks one", async () => {
+    // The same probe result reached from the other shape. "carries no workflow
+    // identity EITHER" would be false here — this session is fenced to STALE — and a
+    // verdict that describes the caller's state wrongly is the class of claim this
+    // whole issue is about.
+    fence = { known: true, uuid: STALE };
+    const text = await outline({ issuedFor: STALE, liveUuid: null });
+
+    expect(text).toMatch(/no workflow identity of its own/);
+    expect(text).not.toMatch(/no workflow identity either/);
+    expect(text).toMatch(/will NOT clear this/);
+    expect(adopted).toEqual([]);
+  });
+
   it("PINNED: naming mode:\"current\" also discloses that it releases the pin", async () => {
     // mode:"current" is the right remedy for a missing stamp AND it silently drops a
     // pin. A caller who followed it to fix a read would lose their target and find
--- a/src/__tests__/orchestrator/fenced-read-absent-stamp.test.ts
+++ b/src/__tests__/orchestrator/fenced-read-absent-stamp.test.ts
@@ -1,0 +1,267 @@
+// #1519 — a live graph READ refused by the workflow fence said nothing about how to fix it,
+// and the two states behind that refusal are not the same state.
+//
+// The reporter's session resumed onto a different workflow and the first live-canvas read
+// came back
+//
+//   workflow instance mismatch: this command carries no workflow-instance stamp, and the
+//   active canvas reports 2b3f4684-…. Nothing was applied.
+//
+// MEASURED on main before this change: that sentence was the ENTIRE tool result. The
+// corroboration #1330 added is gated on `isMutatingGraphCmd`, so a READ refused by the very
+// same fence fell through every branch of the catch and the panel's raw refusal was
+// surfaced verbatim — no verdict, no remedy. The reporter found
+// `panel_set_workflow_target({mode:"current"})` themselves; that discovery is the report.
+//
+// TWO REFUSALS, NOT ONE — the part these assertions exist to pin. `isWorkflowInstanceMismatch`
+// matches both of the panel's states and they have OPPOSITE remedies:
+//
+//   "carries no workflow-instance stamp"  → no identity at all. Nothing was compared.
+//                                           Deriving a fence from the live canvas fixes it.
+//   "issued for workflow instance <uuid>" → an identity the canvas disagrees with. Deriving
+//                                           a fence from the live canvas ABANDONS the
+//                                           workflow the caller named (the retarget #1646
+//                                           removed for cause).
+//
+// So each case asserts its OWN remedy AND asserts the other case's remedy is absent. A
+// verdict that merely mentioned rebinding would satisfy a one-sided test while re-collapsing
+// the two states, which is the defect this issue is about.
+//
+// THE FENCE IS NOT MOVED. Every case asserts `refreshWorkflowUuid` was never called: the probe
+// is the read-only one, the refusal still fails the call, and the caller decides.
+
+import { beforeEach, describe, expect, it } from "vitest";
+
+import {
+  buildPanelToolDefs,
+  makePanelToolCtx,
+  type PanelToolCtx,
+  type ToolResult,
+} from "../../orchestrator/panel-tools.js";
+import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+
+const TAB = "wf:workflows/a.json";
+const LIVE = "2b3f4684-b7e7-495b-a7a1-f40439e35e0a"; // what the canvas reports
+const STALE = "aaaaaaaa-1111-4111-8111-aaaaaaaaaaaa"; // a fence that has gone stale
+
+let listCalls: number;
+let adopted: string[];
+let fence: { known: boolean; uuid?: string };
+
+/** A self-corroborating workflow_list reply: the active record also appears in the
+ *  open-workflow list flagged active, which is what corroboration requires. */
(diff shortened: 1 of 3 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**. Verdict: **SHIP**, head `786acf2` (unchanged during review; merges cleanly into current `origin/main` and stays green there).

The hunt was the fact that this fix **classifies by matching the panel's prose**, which is the most dangerous shape in this codebase. Both of the ways that goes wrong were checked and both came back clean:

- **What the classification gates: advice text only.** `shape` (`src/orchestrator/panel-tools.ts:7610`) is a local const consumed exclusively inside template literals in one `fail(...)`. No mutation, retry, rebind, adoption, or fence move keys on it. The branch's only side effect is `rebindWorkflowFence(ctx, {adopt:false})`, which returns `diverged` at `:4941` *before* `refreshWorkflowUuid`, and `unreadableOrHealed` (`:4684`) only sleeps and reads. #1646's read-only rule is upheld — and the `adopt:false` → adopt mutation fails 6 tests, so that property is genuinely pinned rather than incidental.
- **i18n exposure: none.** The refusal is built from untranslated template literals in `comfyui-mcp-panel/web/js/comfyui-mcp-panel.js:3511-3550`; no locale file carries the strings. Degradation is safe both ways: a reword that keeps the `workflow instance mismatch` token lands in `unstated` → UNKNOWN, and losing the token disables the branch and restores main's behaviour. That matters, because a gate matching message words normally dies for 11 of 12 locales.
- **UNKNOWN is genuinely reached**, not decorative: a third wording ("the two identities do not match") and a non-uuid stamp ("tmp-42") both produced the UNKNOWN verdict with `adopted=[]` and no remedy named.

**Mutations, each confirmed applied via `git diff -U0`** — and worth noting one perl attempt ate a backslash and was caught and redone, which is exactly the failure mode that makes a never-applied mutation look like a surviving one: disabling the branch fails 9 (survivor is the mutation-path control); collapsing the shape fails 3; dropping the `graph_` prefix kills the vitest worker with fatal recursion; widening the switch-guard classifier still fails the corrected assertion; dropping `pinNote` fails 1. 165 files / 2838 tests pass; `tsc --noEmit`, `check:vocabulary`, `check:unknown-collapse`, `vocab:export --check` all clean. The corrected 0.11.83 two-option claim is in the diff itself, verified verbatim against the panel source.

**Three non-blocking findings recorded for follow-up:**
1. **P2, prose:** `graph_run` is classified as a read. `isFencedGraphRead` (`:925-928`) is "any `graph_*` not in `MUTATING_GRAPH_EDIT_CMDS`", and `graph_run` is not in that set, so the advice reads "this is a read, so re-issuing it cannot double-apply anything." The advice happens to be *sound* — the panel raises the fence in the dispatcher before the executor (`comfyui-mcp-panel.js:20320-20334`), so nothing was queued — but the noun is false, and the panel's own `READ_ONLY_GRAPH_COMMANDS` disagrees. Baseline on main was a bare refusal, so this is not a regression.
2. **P3, precedence:** at `:7610` `unstamped` wins over `stamped`, so a wrong-stamp refusal whose appended `movedNote` happens to contain "carries no workflow-instance stamp" is classified UNSTAMPED and handed the `mode:"current"` remedy. Not naturally reachable today. Cheap hardening: anchor both regexes to the clause, or return UNKNOWN when both match.
3. **Cost:** one `workflow_list` plus up to the #1292 ladder (~2.9s) on every fence-refused graph read, including the internal `graph_query` probe at `:12279`.

Issue #1654, opened 2026-08-17
restart_comfyui selects empty .venv instead of working Stability Matrix venv and leaves server down


artokun:
### What happened / root cause
On a Stability Matrix install containing both `ComfyUI/.venv` and `ComfyUI/venv`, `restart_comfyui(action:"restart")` stopped a healthy running server, then relaunched with `.venv/Scripts/python.exe`. That environment is an unrelated/incomplete Python 3.12.13 venv with no torch or sqlalchemy. The actual working ComfyUI environment is `venv/Scripts/python.exe` (Python 3.12.10, torch 2.13.0+cu130, sqlalchemy 2.0.52). The restart executable-selection logic preferred the directory name `.venv` instead of preserving/validating the environment used by the healthy instance or the Stability Matrix package configuration.

### Steps to reproduce
1. Stability Matrix ComfyUI package has both `.venv` and `venv`.
2. Healthy ComfyUI is running on localhost:8188 with the full Stability Matrix environment.
3. Call restart_comfyui with action=restart.
4. Tool stops the server and launches `.venv/Scripts/python.exe main.py ...`.
5. Relaunch exits code 1 and leaves ComfyUI down.

### Exact error
`ModuleNotFoundError: No module named 'sqlalchemy'` while importing app.assets.database.
Direct verification of the selected `.venv`: Python 3.12.13; pip present; `import torch` fails.
Direct verification of `venv`: Python 3.12.10; torch 2.13.0+cu130, numpy 2.4.6, sqlalchemy 2.0.52 all import.

### Fix
Applied locally: workaround yes; source patch no. Upstream-only: yes.
Recovered by relaunching the same ComfyUI root with `venv/Scripts/python.exe` and the saved Stability Matrix launch flags. Health then passed and the new custom node loaded.

Precise change needed:
- Preserve the executable of the known healthy process when capturing restart state.
- For Stability Matrix, resolve its configured package environment before generic `.venv` preference.
- If multiple candidates exist, validate the candidate can import core runtime dependencies (at least torch and sqlalchemy) before stopping the healthy server.
- Refuse before stop if a reproducible relaunch command cannot be validated.

### Environment
Windows 11; ComfyUI 0.33.1; frontend 1.48.7; RTX 5090 32 GB.
comfyui-mcp 0.51.57; panel 0.13.7.

Issue #1647, opened 2026-08-16
panel_restart_comfyui refuses ComfyUI Desktop when parent PID cannot be read


artokun:
### What happened / root cause
panel_restart_comfyui refused to restart a live ComfyUI Desktop instance because it could not read the parent process of PID 43677 and therefore could not prove a Desktop supervisor would bring the server back. This blocked loading a just-installed custom node (comfyui-animatediff-evolved). The server argv clearly showed Desktop flags (`--feature-flag show_signin_button=true`, `--enable-manager`, `--extra-model-paths-config` under `~/Library/Application Support/Comfy Desktop/`).

### Steps to reproduce
1. Run ComfyUI Desktop on macOS.
2. panel_install_node a pack that needs a restart.
3. Call panel_restart_comfyui.

### Exact error (scrubbed)
Refusing to restart ComfyUI: this is a ComfyUI Desktop instance, and it could not be established that a Desktop app is still supervising it (the parent process of PID 43677 could not be read). ... A restart from here would STOP ComfyUI and nothing would bring it back automatically.

### Fix
applied locally: no
upstream-only: yes
If argv contains Comfy Desktop extra-model-paths (`Application Support/Comfy Desktop/instance-model-paths/`) or Desktop feature flags, treat Desktop as the supervisor and allow the Manager restart, or offer a one-click 'ask Desktop to relaunch' path instead of a hard refuse. Do not require a readable parent PID on macOS when Desktop launch signatures are already present.

### Environment
OS darwin · GPU mps 16 GB · ComfyUI 0.33.1 (LOCAL Desktop) · comfyui-mcp 0.51.56 · panel 0.14.43

Issue #1642, opened 2026-08-16
Restart flow returns unconfirmed after ComfyUI is healthy


artokun:
### What happened / root cause
The live panel restart flow did not receive its confirmation card within 90s, so panel_restart_comfyui reported that it did not restart. The direct comfyui-mcp restart fallback then reported stopped:true, started:false, ready:true, startup:"unconfirmed", listener_ownership:"unconfirmed", even though the health probe confirmed ComfyUI was healthy again.

### Steps to reproduce
1. With an idle local ComfyUI Desktop instance, call panel_restart_comfyui.
2. When the panel confirmation card is not acknowledged, observe the 90s no-confirmation result.
3. Call restart_comfyui action:"restart".
4. Observe a healthy /system_stats probe alongside started:false and startup/listener ownership unconfirmed.

### Exact error (scrubbed)
"No confirmation received within 90s, so I did NOT restart ComfyUI."
"A ComfyUI-Manager reboot was dispatched but not acknowledged ... ComfyUI is healthy now ... startup:'unconfirmed' ... listener_ownership:'unconfirmed'."

### Fix
Applied locally: no. Upstream-only from this session: yes.
Needed: make the panel restart confirmation path reliably surface/resolve, and when the health probe confirms the same local listener is healthy after dispatch, return a definitive restart outcome (or clearly distinguish 'healthy but ownership unconfirmed' without implying the restart may not have occurred).

### Environment
OS: macOS / darwin
ComfyUI: 0.32.0
GPU: Apple MPS, 48 GB unified VRAM
comfyui-mcp: 0.51.52
comfyui-mcp-panel: 0.11.4
ComfyUI frontend version: not exposed in the session environment line

Related (same root cause): #1332, #509

artokun:
Additional current-version reproduction: on macOS (ComfyUI 0.27.1, frontend 1.45.20, MPS 24 GB, torch 2.10.0, comfyui-mcp 0.51.56, panel 0.14.43), after installing a custom node through `apply_manifest`, `panel_restart_comfyui({})` reported that the panel transport was unreachable and the confirmation question never appeared. ComfyUI itself remained reachable, no restart was dispatched, and the install→restart→continue flow was blocked. This variant fails before the confirmation card can be delivered, rather than merely timing out waiting for an answer.

Issue #476, opened 2026-07-30
restart_comfyui refuses local restart when COMFYUI_PATH resolves to stale relative install


artokun:
### What happened / root cause
After `download_model` successfully wrote into the detected live local install at `D:/ComfyUI_portable/ComfyUI/models/frame_interpolation/`, `restart_comfyui` resolved the server script as the stale relative path `ComfyUI/main.py` and refused to restart. The tool's install-path resolution is inconsistent with the path already used successfully by `download_model` in the same session.

### Steps to reproduce
1. Run a local portable ComfyUI installation.
2. Use `download_model`; observe a successful destination under the live absolute ComfyUI path.
3. Call `restart_comfyui`.
4. Observe refusal based on a different stale relative `COMFYUI_PATH`.

### Exact error (scrubbed)
`Refusing to restart: Resolved ComfyUI script does not exist on disk: ComfyUI\\main.py — COMFYUI_PATH may point at a stale/old install that has no runnable server. ComfyUI was left running (not stopped) so you don't lose the server.`

### Fix
Applied locally: no
Upstream-only: yes

Precise change needed: reuse the canonical absolute local install path already resolved by model-download/environment services, or fall back to the currently running ComfyUI process command line before rejecting a stale relative `COMFYUI_PATH`. Preserve the safe refusal if no valid live script can be resolved.

Temporary workaround: inspect the live process command line and restart that exact portable launch command.

### Environment
Windows 11; ComfyUI 0.28.3; RTX 5070 Ti Laptop GPU 12 GB; Python 3.13; torch 2.13.0; comfyui-mcp 0.48.13; panel version not exposed in environment snapshot.

PR #1737, opened 2026-08-19
fix(panel#1262): a configured COMFYUI_RESTART_COMMAND restarts the externally-managed instance the launch path can't prove

### LINKED ISSUE #1262: panel_restart_comfyui cannot recover wedged externally managed ComfyUI when COMFYUI_PATH resolves to main.py
A panel-queued Qwen Image Edit render wedged inside KSampler. queue(action:'cancel') correctly escalated through interrupt + POST /free and confirmed the job was still wedged after ~42s. The documented last resort, panel_restart_comfyui, then refused before stopping anything because its resolved ComfyUI script was `main.py` and did not exist at that resolved path. This leaves the sidebar unable to recover a wedged externally managed/local container instance.

### Steps to reproduce
1. Queue a workflow through panel_run.
2. Let a sampler wedge inside one step so cancel/free cannot interrupt it.
3. Call panel_restart_comfyui.
4. It refuses because it cannot resolve the relaunch target, even though the connected ComfyUI service is live.

### Exact error (scrubbed)
`Refusing to restart ComfyUI: Resolved ComfyUI script does not exist on disk: main.py — could not locate the ComfyUI install; set COMFYUI_PATH or a default workspace ... A restart from here would STOP ComfyUI and nothing would bring it back automatically, so it was refused BEFORE anything was stopped.`

### Fix
Applied locally: no. Upstream-only: yes. The orchestrator process owns restart resolution and cannot be hot-patched safely in this wedged session. Suggested change: when the connected instance is externally managed, let panel_restart_comfyui use a configured restart command/service/container callback, or surface a first-class recovery action configured at connection setup. Also avoid resolving a bare `main.py` without anchoring it to COMFYUI_PATH/workspace.

### Environment
Linux; ComfyUI 0.27.0; frontend version unavailable because the wedged server stopped answering system-stats/log requests; AMD Radeon RX 9070 XT 16 GB; comfyui-mcp 0.51.56; panel 0.14.42.

artokun:
## Root cause

For an **externally-managed** local ComfyUI (a container, a systemd unit, a launcher), the running server's `sys.argv[0]` is a bare `main.py` that anchors only inside the instance's own mount/namespace. Every restart path then dead-ends, which is exactly the wedged-server report in artokun/comfyui-mcp-panel#1262:

- the refuse-safe preflight (`assessRelaunch`, #476/#426) correctly **refuses** a kill+relaunch it cannot prove — `Resolved ComfyUI script does not exist on disk: main.py`;
- a wedged server (KSampler stuck, event loop blocked) answers **no** Manager reboot, so the tab dispatch is useless;
- nothing else existed. The restart fixes that shipped since (#1647 interpreter observation, #1654 degraded Desktop supervision, #1642 witnessed-cycle verdicts) all improve paths this scenario can never reach.

## Fix

`COMFYUI_RESTART_COMMAND` — the user's explicit statement of what cycles the instance (e.g. `docker restart comfyui`, `systemctl --user restart comfyui`). A configured, explicit mechanism *is* the proven relaunch the preflight machinery exists to establish.

- **`restart_comfyui`** (local targets; remote/cloud keep the Manager reboot): runs the command via async `exec` (120s cap), fenced against a mid-call retarget before the irreversible act, then reports with the Manager path's own honesty rules — the loopback witness's observed down→up earns `confirmed` (#1642), a healthy answer without it stays `unconfirmed`, an expired readiness budget is NOT a failure (#367), and a failed command is reported as a failed command with the single-probe outcome stated as an observation, never an assumption about what the command did first.
- **`panel_restart_comfyui`**: when the command is set and the tab provably fronts the configured target, the restart routes through it instead of the tab reboot. The server-side **busy guard** that dispatch would have enforced is re-implemented against a *verified* queue read — a busy queue refuses, an **unreadable** queue (the wedge itself) refuses too because cannot-check is not idle, and `force:true` remains the documented override. The binding is re-verified after the queue-check await (r7's rule). The headless-restart block the legacy no-endpoint fallback already used is extracted into one closure both call sites share.
- Both refusal messages and both tool descriptions now name the escape hatch; documented in `docs/configuration.mdx` (+ fr/ja/ko/zh locales).

## Verification

- New: `src/__tests__/services/restart-configured-command.test.ts` (6 tests — confirmed/unconfirmed/not-ready/failed-command shapes + path unchanged without the command) and `src/__tests__/orchestrator/panel-restart-configured-command.test.ts` (6 tests — the issue's exact shape: preflight failing on bare `main.py` yet the restart proceeds through the command; busy guard; unreadable-queue fail-closed; force override; unbound-target refusal; remote untouched).
- Gates in the worktree, on a rebase of current `origin/main` (v0.52.3): `npm ci` ✓, `npm run build` ✓, `check:unknown-collapse` ✓ (0 new), `check:vocabulary` ✓, `vocab:export --check` ✓, `docs:gen` ✓ (only `docs/tools/process-control.mdx` changed), `i18n:check` ✓, all docs/blog checks ✓.
- Full `npm test`: 10209 passed. The only consistent failure is `gen-changelog-dedupe.test.ts` 30s timeouts, which reproduce identically on a clean `origin/main` worktree on this machine (pre-existing, unrelated). A handful of timing-sensitive files flaked under load in full-suite runs (vocabulary, tool-surface-filter, remote-restart-never-came-back, claude-mcp-reconnect, etc.) and **all pass** when re-run targeted on the final rebased tree.

Closes artokun/comfyui-mcp-panel#1262


artokun:
Orchestrator note — your red build is **deterministic, not a flake**, and the failure is not in the code you changed. All three OS builds fail identically. Diagnosis so you can fix it in one pass:

`npm test` chains vitest with eight doc/i18n checks. **Vitest is fully green — 541 files passed.** The failure is the last step, `check:docs-mdx`, which exits 1 with `6 structural problem(s)`:

```
✗ ar/configuration     ✗ es/configuration     ✗ pt-BR/configuration
✗ ru/configuration     ✗ tr/configuration     ✗ zh-TW/configuration
   MDX components differ — English [… 48 ParamField …, Warning]
                             vs    [… 47 ParamField …, Warning]
```

**Cause:** this branch adds a `ParamField` to `docs/configuration.mdx` (92 → 94 ParamField tag lines) documenting the new `COMFYUI_RESTART_COMMAND` behaviour, but the six locale copies were not updated. The checker compares the *component sequence* per page, so one missing `ParamField` in a translation is a structural mismatch regardless of the prose.

**Fix:** add the matching `ParamField` block to `docs/{ar,es,pt-BR,ru,tr,zh-TW}/configuration.mdx` at the same position in the sequence. `ko`, `ja`, `zh` and `fr` are not in the failing list, so those pages either already carry it or do not exist — check before editing them. The English text can be left untranslated in the first pass if you say so; the gate is structural and will go green on the component sequence alone.

**Worth knowing generally:** this will fire for *every* PR that adds a config option, and it is invisible locally unless you run `npm run check:docs-mdx`. Six locale docs trees currently have unfinished translation work sitting in worktrees, which is why these six and not the other four.

I have not touched your PR — it is yours, and it is well under the 8h staleness bar. If it is still red and untouched in a few hours I will pick it up and finish it rather than leave it stranded.

PR #1738, opened 2026-08-19
fix(1697): the download progress counter observes bytes instead of pacing them

### LINKED ISSUE #1697: download_model runs ~25x slower than curl: progress Transform inherits Node's 16KB default highWaterMark

`download_model` streamed a 20.27 GB HuggingFace GGUF at a steady **0.35–0.48 MB/s**, projecting ~16 hours. Plain `curl` to the identical URL on the same machine and link sustained **12 MB/s** (~28 minutes) — a ~25x gap. It is not the network, the host, or HF throttling.

In `dist/services/download-cache.js` there are two branches:

- **No progress** (~line 1617): `await pipeline(nodeStream, fileStream, { signal })` — straight pipe, no interposed stream.
- **Progress** (~line 1634): inserts a `counter` Transform between them, constructed with **no stream options**, so it inherits Node's default non-object-mode `highWaterMark` of **16 KB**.

`createWriteStream` defaults to 64 KB, so the Transform becomes the narrowest buffer in the chain and dictates the whole pipeline's chunk size. For a 20 GB file that is roughly **1.25 million** backpressure round-trips, each costing an event-loop hop plus the `Date.now()` call in `transform()`.

The panel always supplies a `progress` object, so **every user-facing download takes the slow branch** while internal/cache downloads take the fast one. That also explains why this hasn't shown up in internal testing.

### Steps to reproduce

1. `download_model` with url `https://huggingface.co/unsloth/Qwen-Image-Edit-2511-GGUF/resolve/main/qwen-image-edit-2511-Q8_0.gguf`, `target_subfolder: "unet"`.
2. Poll `download_status`.

Observed:

```
0.05/20.27 GB after 156s
0.06/20.27 GB after 173s
0.07/20.27 GB after 194s
```

Steady ~0.35–0.48 MB/s.

Control A — concurrent range request while that job was still running:

```
curl -sL -o /dev/null -w "speed=%{speed_download}" --max-time 12 \
  -r 0-40000000 "<same url>"
→ speed=5663802 bytes/s   (5.7 MB/s, sharing the link with the slow job)
```

Control B — after cancelling the job, backgrounded `curl -L -C -` to the same target path:

```
495 MB in ~40s → 12 MB/s sustained
```

### Exact error (scrubbed)

No error. Silent throughput degradation — the download reports healthy progress the whole time, which makes it easy to misread as a slow mirror.

### Fix

`applied locally: yes`  `upstream-only: no`  `verified: NO — see caveat below`

```diff
--- a/dist/services/download-cache.js
+++ b/dist/services/download-cache.js
@@ -1634,6 +1634,12 @@
     const counter = new Transform({
+        // Without an explicit highWaterMark this Transform inherits Node's 16KB
+        // default and becomes the pipeline's chokepoint: the straight-pipe path
+        // above (no progress) moves at line speed while this one forces a
+        // backpressure round-trip every 16KB. Match the write stream's buffering
+        // so the counter observes bytes instead of rate-limiting them.
+        highWaterMark: 4 * 1024 * 1024,
         transform(chunk, _enc, cb) {
             downloaded += chunk.length;
```

**Caveat — this fix is NOT verified.** The patch is applied on this machine but could not be measured: a 20 GB `curl` workaround is currently saturating the link, so any A/B download now would be bandwidth-bound rather than pipeline-bound. The diagnosis is from reading the two branches, not from a measured before/after. Please confirm with a controlled benchmark before merging.

Worth checking alongside it: `Readable.fromWeb(res.body)` (line 1518) has known per-chunk overhead in Node, and it sits in *both* branches — so if a benchmark shows the highWaterMark change doesn't close the full 25x, that wrapper is the next place to look.

### Environment

- OS: macOS (darwin 25.1.0), Apple Silicon, 48 GB unified, device `mps`
- ComfyUI: 0.33.2 (local), Manager v4
- comfyui-mcp: **0.49.5**
- panel: **0.15.0**
- python 3.13, torch 2.10.0, Node install via npx (`dist` only, no source checkout)

### Impact / workaround

Workaround was to cancel the job and pull the file with a backgrounded `curl -L -C -` straight into `models/unet/`. That works but loses the panel download-tray progress UI, and it pushes agents toward shelling out for model downloads — which the panel guidance explicitly discourages, since a raw shell download can land the file in the wrong directory on installs with custom `--input-directory`/`--output-directory`.

artokun:
## Root cause

`streamUrlToFile` has two branches. The no-progress branch is a straight `pipeline(nodeStream, fileStream)`. The progress branch inserts a byte-counting `Transform` between them — constructed with no stream options, so it inherited Node's default non-object-mode `highWaterMark` of **16 KiB**, the narrowest buffer in the chain (`createWriteStream` alone buffers 64 KiB). Every chunk then cost a full backpressure circuit (source pause → transform → sink → drain → resume). On the reporter's machine (macOS, Node via npx, 12 MB/s link) that circuit cost ~40 ms: a steady **~16 KiB per ~40 ms = 0.35–0.48 MB/s** on a 20 GB download where curl held 12 MB/s on the same link — the ~25x gap. The panel always attaches a progress row, so every user-facing download took the slow branch while internal/cache downloads took the fast one.

## Fix

The counter now gets an explicit, deep, **bounded** buffer (`PROGRESS_COUNTER_HIGH_WATER_MARK = 4 MiB`), so it *observes* bytes instead of pacing them. Progress honesty is unchanged — counting is per chunk either way, and the tray emit is already time-windowed (400 ms). The bound means a stalled sink can buffer at most 4 MiB, not the whole model.

- `src/services/download-cache.ts` — the one-line fix plus the reasoned constant; the constant is exposed via `__downloadCacheTestHooks` (a lazy getter — it's declared further down the module) so the regression test doesn't reconstruct the pipeline.
- `src/__tests__/services/download-progress-counter.test.ts` — pins the buffer depth (≥ 1 MiB so the counter can never be the narrowest stage again; ≤ 64 MiB so it stays bounded), and verifies counting honesty end-to-end: 128 chunks × 48 KiB through the real `downloadUrlToFile` progress path, asserting disk bytes == counter tally == terminal progress row, with no row ever over-counting. The pin test fails without the fix; both pass with it.

## Verification — and an honest caveat

**I could NOT reproduce the 25x gap on my machine**, so this fix is verified by construction and by guard test, not by a measured before/after on the affected environment. Live A/B against the issue's own URL (100 MB range, real `fetch` → `Readable.fromWeb` → counter → `createWriteStream` to disk, Node v24.16.0, Windows, fast link):

```
straight-pipe    : 79.70 MB/s  (100.0 MB in 1.3s)
counter-default  : 76.57 MB/s  (100.0 MB in 1.3s)   <- pre-fix shape: no slowdown HERE
counter-4MiB-hwm : 87.08 MB/s  (100.0 MB in 1.1s)   <- the fix: no regression
```

On a low-RTT, high-bandwidth link the pipeline never stays full long enough for the 16 KiB buffer to become the pacer — consistent with the diagnosis, which requires a per-circuit latency (network RTT + disk) of tens of ms to bite. The reporter's measurement (a steady 16 KiB per ~40 ms) is exactly the signature of the default HWM pacing the transfer. The fix removes the one stage whose buffer was 4x narrower than every other stage's, so the counter structurally cannot be the chokepoint regardless of link latency. It is the same change the issue author proposed but could not measure.

The issue also flagged `Readable.fromWeb(res.body)` per-chunk overhead as a secondary suspect if the HWM change doesn't close the full gap. On Node 24 that wrapper showed no measurable cost here (76 MB/s through it); left untouched — if the reporter still sees a gap after this, that's the next place to look.

## Gates (worktree on `origin/main`)

- `npm ci` — clean
- `npm run build` (`tsc`) — clean, exit 0 (run twice: pre- and post-rebase)
- `npm test` (full chain, exit code captured) — 10,216 passed / 3 skipped; the only failures across two full runs were 30 s **timeout** flakes (zero assertion failures) in three suites unrelated to this change: `gen-changelog-dedupe`, `tool-surface-filter`, `vocabulary`. All three pass in isolation on this branch; `gen-changelog-dedupe` is inherently load-sensitive (every case takes 14–45 s against the 30 s timeout — it spins up git fixtures — and this machine was running other heavy jobs). All download suites pass: `download-progress-counter` (new), `download-cache`, `download-cache-failclosed`, `download-retry` — 145/145. The new pin test fails when the fix is reverted.

Closes #1697

### DIFF
--- a/src/__tests__/services/download-progress-counter.test.ts
+++ b/src/__tests__/services/download-progress-counter.test.ts
@@ -1,0 +1,129 @@
+// #1697 — the progress-counting Transform in streamUrlToFile must OBSERVE the
+// transfer, never pace it, and never misreport it.
+//
+// It was constructed with Node's default 16 KiB highWaterMark — the narrowest
+// buffer in a chain whose write stream alone buffers 64 KiB — so every chunk
+// cost a full backpressure circuit. Measured in the issue: 0.35–0.48 MB/s where
+// curl held 12 MB/s on the same link (~25x), a steady ~16 KiB per ~40 ms.
+//
+// Two things are pinned here:
+//   1. the counter's buffer depth is deep enough that it can never be the
+//      narrowest stage again, and bounded so a stalled sink can't grow memory
+//      without limit;
+//   2. counting honesty: through the deep buffer, the bytes the counter tallies
+//      are exactly the bytes that reached disk.
+
+import { mkdtemp, readFile, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("../../config.js", () => {
+  const config = {
+    comfyuiPath: undefined as string | undefined,
+    huggingfaceToken: undefined as string | undefined,
+    civitaiApiToken: undefined as string | undefined,
+  };
+  return {
+    config,
+    isRemoteMode: () => !config.comfyuiPath,
+    getComfyUIBaseUrl: () => `http://127.0.0.1:8188`,
+  };
+});
+
+/** Every progress row the counter publishes. Recorded through a module mock
+ *  because the real reporter writes to COMFYUI_MCP_PROGRESS_DIR, which is read
+ *  from the environment at import time (same seam download-retry.test.ts uses). */
+const progressRows = vi.hoisted(
+  () => [] as Array<{ status: string; downloaded: number; total?: number }>,
+);
+vi.mock("../../services/download-progress.js", async () => {
+  const actual = await vi.importActual<typeof import("../../services/download-progress.js")>(
+    "../../services/download-progress.js",
+  );
+  return {
+    ...actual,
+    reportDownloadProgress: (p: { status: string; downloaded: number; total?: number }) => {
+      progressRows.push({ status: p.status, downloaded: p.downloaded, total: p.total });
+    },
+  };
+});
+
+import { __downloadCacheTestHooks, downloadUrlToFile } from "../../services/download-cache.js";
+
+const fetchMock = vi.fn();
+let tempDir: string;
+
+/** A 200 response whose body arrives as MANY chunks (the counter's per-chunk
+ *  path), with an honest Content-Length so the completion check has something
+ *  authoritative to verify against. */
+function chunkedResponse(chunks: Buffer[]): Response {
+  const total = chunks.reduce((n, c) => n + c.length, 0);
+  const stream = new ReadableStream<Uint8Array>({
+    start(controller) {
+      for (const c of chunks) controller.enqueue(new Uint8Array(c));
+      controller.close();
+    },
+  });
+  return new Response(stream, {
+    status: 200,
+    headers: { "content-length": String(total), "content-type": "application/octet-stream" },
+  });
+}
+
+beforeEach(async () => {
+  tempDir = await mkdtemp(join(tmpdir(), "comfyui-mcp-counter-test-"));
+  vi.stubGlobal("fetch", fetchMock);
+  fetchMock.mockReset();
+  progressRows.length = 0;
+});
+
+afterEach(async () => {
+  vi.unstubAllGlobals();
+  vi.restoreAllMocks();
+  await rm(tempDir, { recursive: true, force: true });
+});
+
+describe("progress counter buffer depth (#1697)", () => {
+  it("the counter's highWaterMark is deep enough to never pace the pipeline, and bounded", () => {
+    const hwm = __downloadCacheTestHooks.progressCounterHighWaterMark;
+    // Below 1 MiB the counter risks being the narrowest buffer in the chain
+    // again (the regression: 16 KiB vs the write stream's 64 KiB).
+    expect(hwm).toBeGreaterThanOrEqual(1024 * 1024);
+    // Deep, not unbounded: a stalled sink must not be able to buffer the whole
+    // model in memory.
+    expect(hwm).toBeLessThanOrEqual(64 * 1024 * 1024);
+  });
+
+  it("counts exactly the bytes that reach disk, across many chunks", async () => {
+    // 128 chunks of 48 KiB = 6 MiB — chunk sizes deliberately NOT a multiple of
+    // any stage's buffer, so the count survives arbitrary re-chunking.
+    const chunks = Array.from({ length: 128 }, (_, i) => Buffer.alloc(48 * 1024, i % 251));
+    const total = chunks.reduce((n, c) => n + c.length, 0);
+    fetchMock.mockResolvedValueOnce(chunkedResponse(chunks));
+
+    const target = join(tempDir, "model.bin");
+    await downloadUrlToFile(
+      "https://example.com/models/counter.bin",
+      target,
+      {},
+      undefined,
+      {},
+      { id: "counter-test", name: "counter.bin", attempt: 1 },
+      "", // no payload validation — this is about byte counting, not classification
+    );
+
+    const onDisk = await readFile(target);
+    expect(onDisk.length).toBe(total);
+    expect(onDisk.equals(Buffer.concat(chunks))).toBe(true);
+
+    // The counter's terminal row tallies exactly what landed — no drift between
+    // observed and written bytes.
+    const done = progressRows.find((r) => r.status === "done");
+    expect(done).toBeDefined();
+    expect(done!.downloaded).toBe(total);
+    // No row ever claims MORE than the transfer carried.
+    for (const row of progressRows) expect(row.downloaded).toBeLessThanOrEqual(total);
+    expect(progressRows.some((r) => r.status === "error")).toBe(false);
+  });
+});
--- a/src/services/download-cache.ts
+++ b/src/services/download-cache.ts
@@ -1125,8 +1125,17 @@ function cacheDir(): string {
  *  #1567 adds `cachePathForUrl` so a test can stage a `.partial` under a SPECIFIC
  *  header identity (the writer's own derivation) instead of re-implementing the
  *  hash — a fixture derived from a parallel implementation cannot falsify the
- *  lookup it feeds (#1370). */
-export const __downloadCacheTestHooks = { cacheDir, cachePathForUrl };
+ *  lookup it feeds (#1370).
+ *  #1697 adds `progressCounterHighWaterMark` (a getter: the constant is declared
+ *  further down, beside the pipeline it belongs to) so a test can pin the
+ *  progress counter's buffer depth without re-constructing the pipeline. */
+export const __downloadCacheTestHooks = {
+  cacheDir,
+  cachePathForUrl,
+  get progressCounterHighWaterMark() {
+    return PROGRESS_COUNTER_HIGH_WATER_MARK;
+  },
+};
 
 function cacheSizeLimitBytes(): number {
   const raw = Number(process.env.COMFYUI_LRU_CACHE_SIZE_GB ?? "0");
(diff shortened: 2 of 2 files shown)


artokun:
**Cross-session conflict — please read before merging.** Another agent session investigated #1697 earlier tonight, reached the opposite conclusion, and **closed the issue as not-planned** on the strength of measurements. This PR revives the same change. One of us is wrong and it is worth settling rather than racing.

**1. The stated root cause does not hold on this machine.** The body says the counter "inherited Node's default non-object-mode `highWaterMark` of 16 KiB, the narrowest buffer in the chain (`createWriteStream` alone buffers 64 KiB)". I just measured it directly, Node v24.16.0, on this box:

```
node -e 'const fs=require("fs"),{Transform}=require("stream");
const ws=fs.createWriteStream(require("path").join(require("os").tmpdir(),"p.bin"));
const t=new Transform({transform(c,e,cb){cb(null,c)}});
console.log(ws.writableHighWaterMark, t.writableHighWaterMark, t.readableHighWaterMark)'
→ 16384 16384 16384
```

`fs.createWriteStream` defaults to **16 KiB**, not 64 KiB — the 64 KiB default belongs to `createReadStream`. So the counter and the sink are **equal**, and the Transform cannot be "4x narrower than every other stage". Inserting it *adds* buffering rather than removing it. If the 64 KiB figure came from docs or memory rather than a probe on this Node version, that is the thing to re-check first, because the whole backpressure-circuit argument rests on it.

**2. The prescribed change measured as a regression, not a win.** The #1697 investigation ran the real exported `downloadUrlToFile()` over loopback, 512 MiB × 5 rounds, variants alternating in one process:

| variant | round A | round B |
|---|---|---|
| straight pipe | 386.5 MB/s | 399.6 MB/s |
| counter, default HWM | 382.0 MB/s | 403.2 MB/s |
| counter, **4 MiB HWM** | **282.0 MB/s** | **251.4 MB/s** |

That is 26–34% **slower**, reproduced over two apply/revert passes, plus up to 8 MB resident per concurrent download.

**3. Your own numbers and theirs may both be right.** Yours are ~80 MB/s over a real link; theirs are ~390 MB/s over loopback. A 4 MiB buffer costs extra copying that only becomes visible once the pipeline is fast enough for it to dominate — which is exactly the regime loopback probes and real downloads do not. So the honest reading is that the change is roughly neutral at network speed and harmful at high throughput, rather than a straightforward win.

**4. The reporter's premise was also challenged.** That investigation found both of the reporter's curl controls were *range* requests (`-r 0-40000000`, `-C -`) against a Xet-backed object, while the MCP job was a full GET — so the 25x compared two different upstream operations. Running the reporter's own 20 GB Qwen GGUF through `downloadWithCache → downloadIntoCache → streamUrlToFile` with a progress row attached measured **50.7 MB/s**, against `curl` at **42.5 MB/s** on the same machine, URL and session. Our path was *faster* than curl, and the symptom did not reproduce. You independently could not reproduce the 25x gap either, which is consistent.

**Not asking you to drop the work.** Two things in this PR stand on their own regardless: an explicitly bounded counter buffer is more defensible than an implicit one, and the honesty test (128 × 48 KiB, disk bytes == counter tally == terminal row, no row over-counting) is worth keeping either way. What needs resolving is the constant and the justification.

Suggested path: re-run the loopback A/B above on this branch. If the regression reproduces, pick a bound that does not pace *or* penalise — matching the sink at 16 KiB, or a modest 64–256 KiB — and rewrite the root-cause paragraph to drop the 64 KiB claim. I am holding this PR rather than merging until that is settled, and I have not reopened #1697. If the loopback regression does not reproduce for you, post the numbers and I will merge on that.

artokun:
Third-party re-measurement, post-merge (this landed as `f15a6ac`). I was asked to adjudicate this against the earlier session that closed #1697 with the opposite result, so I rebuilt the rig from scratch rather than re-reading either analysis. Short version: **your numbers and theirs are both correct — they were taken in different throughput regimes**, and the crossover sits between 200 and 400 MB/s. That's the interesting finding here, and it's not something either of you could have seen from one regime alone.

Method: real exported `downloadUrlToFile()` against a loopback HTTP server, arms **alternating inside one process** with a shuffled order per round and a settle gap between arms (at these speeds Windows writeback from arm N lands inside arm N+1's window, so a fixed order aliases that cost onto a fixed arm). Each arm is normalized to the straight pipe measured in the *same* round, which is what finally cut through the noise. Box was busy throughout — 32 logical cores at 25–57%, 49→91 node processes, other agents live — hence the interleaving. I added a benchmark-only override so the HWM could vary per call, and verified it reaches the `Transform` by printing `counter.writableHighWaterMark` per run rather than inferring it from timings.

**Unthrottled loopback — 512 MiB, 16 shuffled rounds** (ratio to same-round straight pipe):

| HWM | ratio | median MB/s |
|---|---|---|
| 512 B (control) | 1.113 | 476.9 |
| 16 KiB (default) | 1.163 | 484.6 |
| 64 KiB | 0.974 | 455.2 |
| 256 KiB | 0.866 | 361.8 |
| 1 MiB | 0.790 | 347.8 |
| **4 MiB (merged)** | **0.770** | **329.0** |

So the regression the other session reported does reproduce here: 4 MiB is ~32% below the default arm (329.0 vs 484.6), squarely inside their 26–34%.

**Throttled server** — this is the part that reconciles us:

| server rate | 16 KiB | 64 KiB | 256 KiB | 1 MiB | 4 MiB |
|---|---|---|---|---|---|
| 12 MB/s | 0.999 | 1.000 | 0.999 | 0.998 | 0.999 |
| 80 MB/s | 0.999 | 0.998 | 0.998 | 0.998 | 0.999 |
| 200 MB/s | 1.001 | 1.003 | 0.998 | 0.998 | 0.997 |
| 400 MB/s | 1.021 | 0.988 | 0.936 | 0.905 | 0.923 |

At 80 MB/s every arm lands at 80.5–80.9 MB/s. Your 79.70 / 76.57 / 87.08 spread is entirely consistent with that — at your link speed the constant genuinely does not matter, and the 87.08 was link noise rather than a win. Nothing about your measurement was wrong; it just couldn't discriminate, because at 80 MB/s *nothing* can.

The control arm is what I'd point at hardest. A **512-byte** highWaterMark — 32x narrower than the "pathological" default — holds 12.7 MB/s at 12 MB/s, 80.9 at 80, and is the *fastest* counter arm on unthrottled loopback. If a narrow buffer paced the pipeline, that arm would collapse. It doesn't, at any rate I tested. Combined with the already-established fact that `createWriteStream` is 16384 and not 65536 (I re-confirmed it inside the live pipeline: `fileStream.writableHWM=16384`, equal to the counter's default), the root-cause paragraph now on `main` describes a mechanism I can't make appear on the rig. The 64 KiB figure is `createReadStream`'s default.

Two caveats I want to be straight about: my throttle shapes bandwidth, not RTT, so I have not reproduced a high-latency link; and I could not reproduce the reporter's 25x either — but at 12 MB/s a 512-byte buffer was already at parity, so a 16 KiB one cannot plausibly have paced that transfer to 0.35 MB/s.

On the two tests: the honesty test is good and I'd keep it independent of all this — I reverted the constant to 16 KiB and it still passes, so it's a real guard on the counter's arithmetic through the pipeline rather than a lock on a number. The bounded-buffer test is the opposite: `expect(hwm).toBeGreaterThanOrEqual(1024 * 1024)` fails the moment the constant moves, and its comment restates the 64 KiB premise, so it pins the disputed value rather than testing an invariant.

**What I'd suggest: revert the constant to Node's default and keep the honesty test.** The default arm shows no deficit against the straight pipe at any rate I measured, so there's no throughput case for raising it, and above ~200 MB/s raising it costs real percentage points. If there's an appetite to keep an explicit value for the "observe, don't pace" intent, 64 KiB is the largest that never charged anything — but I couldn't find a speed where it beats the default either. I'm not going to push any of this myself since several sessions are editing this file; handing the numbers over so whoever drives it has them.

Harness and raw per-round output are reproducible on request — happy to hand over the rig if you want to re-run it against your own link.

Issue #1650, opened 2026-08-16
panel_set_workflow_target(mode:current) fails to rebind after multi-workflow reconnect / tmp tab


artokun:
### What happened / root cause
After a ComfyUI reconnect and the user switching among several workflow tabs (including an unsaved tmp tab titled ComfyUI_00727_), `panel_set_workflow_target({mode:"current"})` repeatedly failed to restore the graph binding.

The panel answered, but the session refused to adopt the active-tab record because "the active record and the open-workflow list share no comparable identity field". The tmp tab had `path:null` and `filename:null` (only `title` + `key`/`routing_key` like `tmp:606a1d57-339c-4304-89b0-51c25c1cd121`). The session stayed fenced to a previous workflow instance and graph tools would have kept failing.

This is the same class of identity-mismatch after tab switch / reconnect. `mode:current` is documented as the recovery signal, but it cannot corroborate a tmp/unsaved tab that lacks path/filename.

### Steps to reproduce
1. Have multiple tabs open, including an unsaved tmp workflow (path/filename null).
2. Switch tabs / reconnect mid-session so the turn is issued from mixed workflow origins.
3. Call `panel_set_workflow_target({mode:"current"})`.

### Exact error (scrubbed)
```
panel_set_workflow_target({mode:"current"}) did NOT restore this session's graph binding.

APPLIED: the workflow target is now mode:"current".
NOT APPLIED: The panel answered, but its reply could not be corroborated, so nothing was adopted — the active record and the open-workflow list share no comparable identity field, so nothing corroborates it. NOTHING was written to this session's fence: adopting an uncorroborated record could have stamped ANOTHER canvas's identity onto this tab, so this session is STILL fenced to the previous workflow instance (...) and graph tools will keep failing.

WHAT TO DO: call this again in a moment. ... If it persists ... Ask the user to manually refresh (F5...). Do NOT use panel_reload (#803).
```

`panel_list_workflows` showed the active tab as `{path:null, filename:null, title:"ComfyUI_00727_", key:"tmp:606a1d57-...", routing_key:"tmp:606a1d57-...", persisted:false}`.

### Fix
applied locally: no
upstream-only: yes

Workaround that DID restore binding: `panel_open_workflow({path:"tmp:606a1d57-..."})` then `panel_set_workflow_target({mode:"pinned", path:"tmp:606a1d57-..."})`. After that `panel_graph_outline` worked.

[...]

Issue #1639, opened 2026-08-16
Panel tab stops answering graph_* (incl. read-only) while a render runs; and workflow_open asserts "you are on the right workflow" when the canvas shows a different graph


artokun:
Two related panel-reliability defects hit repeatedly in one working session (long item-generation session, ~40 renders).

### What happens
While a ComfyUI render is in flight, `panel_set_widget`, `panel_run` and even the READ-ONLY `panel_query_graph` / `panel_graph_outline` time out:

```
Error: Panel tab tmp:<id> did not reply to "graph_set_widget" within 30000 ms
  — the ComfyUI tab may be backgrounded or frozen. This command MUTATES and was
  already delivered to the tab, so it may have been applied despite the missing reply
Error: Panel tab tmp:<id> did not reply to "graph_query" within 20000 ms
```

The moment the queue drains, the very same calls succeed instantly. Reproduced ~12 times across the session; 100% correlated with `queue action:"list"` showing `running: 1`.

### Why it matters
The documented agent workflow for a parameter sweep is "set widget -> panel_run -> set widget -> panel_run" back to back (the panel_run docs even describe queuing several and letting them "drain in order"). That is exactly the pattern that breaks: the first run starts, and every subsequent edit times out.

Worse, the outcome of each timed-out MUTATION is genuinely unknown. Observed BOTH outcomes in the same session:
- mostly the command HAD landed (a later read showed the new value), and
- at least once it had NOT (`panel_set_widget` on a CLIPTextEncode timed out and the node still held the previous prompt), which was only caught because the read was done before re-issuing.

So the agent must round-trip a verifying read after every single edit — and that read times out too while the render runs. The practical fallout was queuing a render whose prompt state could not be confirmed, and having to serialize to one-item-at-a-time (~1 min of dead wall-clock per item) instead of pipelining.

### Suggested direction
Either (a) keep the panel's command channel responsive during execution (the reads at minimum should never be blocked by a running prompt), or (b) if the frontend genuinely cannot answer mid-execution, have the orchestrator detect "a prompt is running" and say so explicitly instead of surfacing a generic 20/30s timeout with an unknown-outcome warning. An explicit "queue busy, retry after" is far more actionable than "may or may not have applied".

The `retry_of:` token mechanism is good and did work (one retry correctly reported "the attempt you are retrying DID complete ... it applied" rather than double-applying) — the problem is having to reach for it constantly.

### What happens
[...]

Issue #1589, opened 2026-08-14
panel_strip_workflow text is not directly parseable as JSON


artokun:
### What happened / root cause
panel_strip_workflow documentation says it returns the resolved graph, but the TextContent begins with human-readable summary lines before the JSON object. Programmatic consumers cannot JSON.parse the returned text directly and must search for the first opening brace and slice manually.

### Steps to reproduce
1. Call panel_strip_workflow on a live canvas.
2. Read content[0].text.
3. Attempt JSON.parse(content[0].text).
4. Parsing fails on the leading "Stripped to N nodes" summary.

### Exact error
SyntaxError: Unexpected token 'S', "Stripped t"... is not valid JSON

### Fix
applied locally: no
upstream-only: yes
Return the graph in structuredContent.graph (preferred) and keep the prose in TextContent, or make TextContent a pure JSON document. Preserve the summary as separate structured fields.

### Environment
Windows 11; ComfyUI 0.33.1; RTX 4070 SUPER 12GB; comfyui-mcp 0.51.41; panel 0.14.35.

artokun:
Draft PR up: artokun/comfyui-mcp#1599.

Confirmed and reproduced. The graph was string-concatenated onto the end of the summary inside a single text block, so `JSON.parse(content[0].text)` died on the "S" exactly as you describe. Owned by the orchestrator (`src/orchestrator/panel-tools.ts`) — the panel has no strip handler.

Taking the shape you preferred: `structuredContent.graph` (plus `node_count`, `node_types`, `warnings`), and the graph also gets its own trailing content block so it parses without `structuredContent`. The prose stays first and unchanged — #361 deliberately puts conversion notes ahead of the graph they apply to, and putting the JSON at `content[0]` would have traded that warning away.

One thing worth flagging from the investigation: the "search for the first opening brace and slice" workaround is wrong on real replies, not just ugly. The #959 widget-capture note this tool routinely emits ends with `panel_query_graph {fields:'detail'}` — a brace in the prose, ahead of the graph. That case is now pinned as a test.

artokun:
Fixed, and carried by v0.51.55.

`panel_strip_workflow` now returns the graph in its own content block **and** on `structuredContent.graph` — the shape you preferred — with `node_count`, `node_types` and `warnings` alongside it. The summary stays in `content[0]`, and the graph block parses with `JSON.parse` as-is, with or without a structured-content-aware client.

[...]

PR #1740, opened 2026-08-19
fix(panel#1339): a fence this call REPAIRED is not a fence that was already fine — and the verdict rides in a field

artokun:
Addresses the underlying cause of artokun/comfyui-mcp-panel#1339 — "panel_set_widget
transiently rejects matching re-derived workflow identity". The symptom was filed against the
panel; the code is here. All three strings in that report have zero occurrences in the panel's
`web/js/` and live in `src/services/ui-bridge.ts` and `src/orchestrator/panel-tools.ts`, and
the refusal is raised pre-dispatch, so no panel code ran on the failing call. The panel issue
is correctly labelled `autopilot:parked`.

## It is not a race, and there is no window

The reporter read the retry as a timing window closing. It is not. **The failed call is the
repair.** On the `no trusted identity` path `panel-tools` calls `rebindWorkflowFence(ctx)` with
adoption enabled (unlike the mismatch branch next door, which #1646 moved to `{adopt:false}`).
It reads the panel's fence-exempt `workflow_list`, corroborates the active record, finds a real
uuid and installs it. The retry then succeeds *because the first call already fixed the fence*.
Nothing settled, nothing flapped.

## The defect: two opposite outcomes, one sentence

```ts
if (rebind.status === "refreshed" || rebind.status === "already_current") {
  return fail(`${raw}\n\nCHECKED: the live canvas DOES carry an identity (${rebind.uuid}) and ` +
    `this session's fence has been re-derived onto it. RETRY THIS EXACT CALL ONCE — …`);
}
```

Taxonomy class 1. The two branches mean opposite things and want opposite next moves:

* **`refreshed`** — the tab had **no** fence and *this call installed one*. A bare retry is
  right, and works because of the repair.
* **`already_current`** — the fence read back **already present and already equal**, and
  `already_current` returns *before* any adoption, so this call did not install it. It also
  cannot be the fence the command was refused against: that refusal is raised only when
  `resolveTabWorkflowUuid(tabId)` is empty (ui-bridge `hasTrustedStamp`), and
  `currentWorkflowFence` reads the same resolver. So a fence appeared between the two reads.
  **Two things produce that and this code cannot tell them apart**, so it asserts neither:
  the session was moved onto a different tab mid-check (`rebindWorkflowFence`'s
  `workflow_list` round trip is retry-safe, its retry runs `ensureReachable`, which can move
  an orphaned `mode:"current"` session, and `before` is then re-read for the new tab), or a
  fence for this tab was installed in that window by something else (a concurrent rebind, or
  the panel's own mismatch re-hello, #1043/#932). Either way **nothing was repaired for the
  caller** and the uuid in the message may not be the one they were refused against.

One sentence for both told the reporter the identity "had already been re-derived" — which
reads as *it was fine all along and you were refused anyway*. That is why this was filed as a
contradiction.

## The fix

**1. The prose is split.** `refreshed` now says this call installed the fence and that the
refusal itself is the repair. `already_current` says `THIS CALL REPAIRED NOTHING`, says why the
identity it names may not be the caller's, and asks them to confirm the target
(`panel_set_workflow_target({mode:"current"})` or `panel_open_workflow(<path>)`) before
re-issuing — rather than ordering a blind retry at a tab they never asked for.

**2. The verdict rides in a structured field, because prose is not a decision procedure.**
The standing rule here is that an agent must never re-run a mutation by pattern-matching an
error message. A caller that has to parse the sentence to know whether to retry is still
broken, so the refusal now carries `structuredContent.panel_fence`:

| field | values | what a caller does with it |
|---|---|---|
| `rebind_status` | the re-derivation's own verdict (`refreshed`, `already_current`, `no_identity`, `unreadable`, …, `check_threw`) | branch on it instead of on wording |
| `fence_repaired_by_this_call` | `yes` / `no` / `unknown` | **the discriminator.** `yes` = the refusal you are holding is the repair |
| `retry_safe` | `yes` / `no` / `unknown` | may I re-issue without a double-apply? |
| `retry_clears_refusal` | `yes` / `no` / `unknown` | will re-issuing get past the fence? |
| `next_action` | `retry_same_call`, `confirm_target_then_retry`, `open_workflow`, `open_or_save_workflow`, `unknown` | one machine-selectable step |
| `dispatched` | `no` / `yes` / `unknown` | from the bridge's typed flag |
| `workflow_uuid`, `tab_before`, `tab_after` | evidence | — |

Every field is tri-state on purpose: a boolean would recreate the exact collapse this removes,
with `false` meaning both "no" and "we could not find out".

`refreshed` → `fence_repaired_by_this_call:"yes"`, `retry_clears_refusal:"yes"`,
`next_action:"retry_same_call"`. `already_current` → `"no"`, `"unknown"`,
`"confirm_target_then_retry"`. Note `retry_clears_refusal` is **not** `"yes"` for
`already_current` even though a retry would carry *a* fence: nothing on that path establishes
whose tab that fence belongs to, and `tab_before === tab_after` does not prove the session never
moved (a move-and-return is invisible to a bare id comparison), which is why the field says
`unknown` rather than being derived from that pair.

**3. `retry_safe` stands on the bridge's typed flag, never on the text.** The predicate that
selects this branch, `isNoTrustedIdentityRefusal`, is `/no trusted identity/i` — fine for
choosing *what to explain*, unfit for authorising a re-run, since any error quoting the phrase
satisfies it. The claim now reads `dispatchOutcomeOf(err)` (the bridge's `markDispatched`
symbol) and says `unknown` when the flag is absent — and the prose drops "nothing was applied,
so a retry cannot double-apply" in that case instead of promising it on a word match.

## What deliberately stays a refusal

The call really did not perform the widget write; it performed the repair. Every branch still
returns `isError`, still carries the panel's own cause verbatim, and nothing auto-continues the
mutation. Executing a write against a fence the same request just minted is the shape #1646
removed from the neighbouring branch — a refusal that repairs and *reports* is not the same
risk as one that repairs and *proceeds*.

## A pre-existing test was pinning the wrong branch

`panel-session-rebind-residuals.test.ts`'s `#709/#1331` case set
`workflowUuidFor: () => ({known: true, uuid: LIVE_UUID})` — a fence present and equal — while
its `send` threw the `no trusted identity` refusal for the same tab. A real bridge cannot do
both: that refusal fires only when the resolver is empty, and `workflowUuidFor` reads the same
resolver. So the harness was driving `already_current` while asserting the `refreshed` wording,
**and it passed**, because one sentence covered both. Fixed to `{known: true}` (FenceRead's
"definitively no fence"), which is #709's actual state, and the test now also asserts
`rebind_status === "refreshed"` structurally so it cannot drift back silently.

## Refutation

Each mutation was applied with an explicit `git diff` check first — the file is CRLF on disk
and an LF-anchored substitution silently no-ops on it (the first attempt did exactly that and
was caught by an exit-9 guard, not by a passing test).

| # | mutation | `git diff` confirmed applied | result |
|---|---|---|---|
| 1 | delete `res.structuredContent = …` in `failWithFenceDiagnosis` | `-1 line` | **8 of 10 fail** |
| 2 | re-merge the branches into `refreshed \|\| already_current` | `-1/+1` | **4 fail** (already_current, THE COLLAPSE, transport, SOURCE) |
| 3 | derive `dispatched` from the branch instead of the typed flag (`= "no"`) | `-1/+1` | **1 fails** (the untagged-error case) |
| 4 | `already_current` claims `retry_clears_refusal:"yes"` / `retry_same_call` | `-2/+2` | **3 fail** |
| 5 | revert `panel-tools.ts` to `origin/main` wholesale | index+worktree replaced | **9 of 10 fail** |

The one survivor under (5) is "every state still REFUSES", which is a pre-existing property and
should survive. The suite also asserts the **call site** (`panel_set_widget`'s handler, end to
end) rather than the helper, plus a SOURCE assertion that the two statuses are not re-merged,
and a real `McpServer` + `Client` round trip proving `structuredContent` survives the SDK
transport **on an `isError` result** — #1589's existing wiring test only covers a success.

## Test results

* New file `src/__tests__/orchestrator/fence-repair-vs-already-current.test.ts` — 10/10.
* Neighbouring fence suites (`no-trusted-identity-remedy`, `fence-matching-uuid`,
  `fence-mismatch-corroborates`, `fence-rebind-transient`, `fence-refusal-reason`,
  `fence-trusts-own-reply`, `strip-workflow-parseable-reply`) — 58/58.
* `src/__tests__/orchestrator` — 165 files, 2838 tests, all pass.
* `tsc --noEmit` clean; `check:vocabulary`, `check:unknown-collapse`, `vocab:export --check`
  all OK locally.

## Review round 2 — the replacement message reproduced the collapse it replaced

Two defects the gate found in my OWN new code, both agent-facing, both this PR's own class.
Fixed in 141cd18.

**(a) `refreshed` asserted an absence it never observed.** `refreshed` is returned for every
prior that is not a known-equal fence, which is THREE priors: definitively none, a read that
FAILED, and a fence naming a DIFFERENT workflow that the adoption replaced. My text said "this
session had NO fence for it" for all three — the exact fold `FenceRead`'s own contract
forbids ("an absence nobody observed"). main's wording was true in all three; mine was
strictly stronger and newly false in two. Each prior now gets its own sentence, and the
observation rides in `prior_fence: "absent" | "present" | "unreadable"` (+ `prior_fence_uuid`)
rather than in the wording. The same mislabel in this PR's own fixture is fixed too:
`REFRESHED` was `{known:false}` under the label "no fence on the tab" and is now `{known:true}`,
which is also the read that matches the refusal.

**(b) The `nothingApplied` ternary was two-way over a three-way flag.** A `dispatched:true`
refusal got a sentence claiming the flag was absent, printed beside `retry_safe:"no"` and an
unconditional "RETRY THIS EXACT CALL ONCE" — a four-way self-contradictory verdict, and the
prose is the half an agent obeys. Latent today (the `markDispatched(…, true)` sites mint their
own text and never carry the phrase), but it is new code. Each dispatch state now gets its own
sentence, and **a retry is only ORDERED when the bridge proved nothing was written** — otherwise
`next_action` degrades to the new `verify_applied_then_decide` and the prose stops ordering one.

**(c) Optional, taken:** the `adopt_error` tri-state shipped correct but unpinned — collapsing
it to a bare `"no"` left the suite green. It now has a test.

Four more mutations, each `git diff`-confirmed applied: re-collapse the three priors → 2 fail;
revert `nothingApplied` to two-way → 1 fails; ungate the retry order (`mayOrderRetry = true`) →
2 fail; collapse the `adopt_error` tri-state → 1 fails. Re-running the original re-merge
mutation against the final code still kills 4. Suite is 19/19; the seven neighbouring fence
suites are 88/88.

## Adversarial re-read of my own diff

Commit 4cfb814 exists because of it. My first `already_current` wording asserted "this session
reached a DIFFERENT tab while the check ran" as fact. That is one of two mechanisms that
produce the observation and nothing in this code measures which — the exact "prose overstates
what the code does" class. The message now states what was observed (a fence appeared between
the refusal and the re-read, and not through this call), names both possibilities, and lets the
remedy follow from the observation instead of from a mechanism nobody proved.

## Verified against CURRENT main, not just my base

The CI run on head `141cd18` is an `event=pull_request` run, so `actions/checkout@v4` built
`refs/pull/1740/merge` — this branch merged into the base tip — and all four checks passed.
Main then advanced again ~6 minutes after that run started, and **8 of the 13 commits it gained
touch `panel-tools.ts`**, several of them fence code (#1519, #1650, #1639, panel#1292). A green
check is only green for the main it ran against, so that was re-checked directly rather than
assumed:

* Local test-merge of `origin/main` (8d85be9) into this branch: **no conflicts**.
* `tsc --noEmit` on the merged tree: **clean**.
* `src/__tests__/orchestrator` on the merged tree, serial: **173 files, 2916 tests, all pass**
  — including the eight new suites those commits added (`fenced-read-absent-stamp`,
  `open-content-mismatch-fence`, `current-bind-clears-ambiguity`, …).
* The merge was then discarded; the branch is still the linear `141cd18` that CI signed off.

For reference, the same suite at `141cd18` run serially is 165 files / 2847 tests, all pass —
which is what identifies the earlier parallel-run failures on this box as load, not defects.

## What I did not do

* **No gate script was run** (`codex exec` is account-exhausted until 2026-08-19 20:43 and
  `claude-gate.mjs` is blocked on the weekly limit). Gating and merging are the orchestrator's.
* **No Playwright run.** Nothing here is panel-rendered: the panel emits none of these strings,
  and the refusal is raised before any frame reaches the socket. No `web/js/` file is touched.
* I did not chase *which* orchestrator path left the stamp empty in the first place. The
  scope-addressed stamp comes from `turnOrigins.stampOf` and `resolvedPinOf` can clear it; I
  have evidence for where the message is wrong, not for which of those produced the empty
  stamp, and I am not asserting the second.
* I did not make the refusal auto-continue the mutation, and did not touch the
  `already_current` handling in the #1330 mismatch branch, which is a different (read-only)
  probe.
* I did not touch an adjacent pre-existing inaccuracy: `healed_by_panel` falls into the
  "could not be re-read well enough" tail message, which is wrong for it (the read succeeded;
  the panel repaired the fence underneath). Out of scope here, and the new field at least
  reports `rebind_status:"healed_by_panel"` truthfully where the sentence does not.
* Local full-suite runs on this box show a varying set of unrelated failures (4 in one run, 12
  in the next, none reproducible when the same files are run alone, and identical with
  `panel-tools.ts` reverted to `origin/main`) — this machine is running a dozen agents and the
  logs are full of EISDIR/temp contention. Every one of those files passes when run alone
  (5 files, 111 tests, green), and CI is green on all three OS legs — so they are box load, not
  this change.

### DIFF
--- a/src/__tests__/orchestrator/fence-repair-vs-already-current.test.ts
+++ b/src/__tests__/orchestrator/fence-repair-vs-already-current.test.ts
@@ -49,20 +49,27 @@ const LIVE_UUID = "55555555-5555-4555-8555-555555555555";
 
 type Fence = { known: false } | { known: true; uuid?: string };
 
-/** The refusal the bridge really mints for this state: `markDispatched(err, false)`
- *  — the typed proof that the frame never reached the socket (ui-bridge.ts). */
-const identityRefusal = (tagged = true): Error => {
+/** The dispatch flag the error carries. `no` is what the bridge really mints for this
+ *  state (`markDispatched(err, false)` — typed proof the frame never reached the
+ *  socket). The other two are the hostile cases: an error whose TEXT satisfies the
+ *  branch's phrase match while the flag says the opposite, or is absent entirely. */
+type Dispatch = "no" | "yes" | "untagged";
+
+const identityRefusal = (dispatch: Dispatch = "no"): Error => {
   const err = new Error(
     `"graph_set_widget" cannot be safely targeted to the active workflow: this workflow has ` +
       `no trusted identity for the panel to fence the command against. Reads and view-only ` +
       `commands still work (graph_outline, graph_query, graph_get_state).`,
   );
-  // `tagged:false` is the hostile case: an error whose TEXT satisfies the branch's
-  // phrase match but which carries no dispatch flag at all.
-  return tagged ? markDispatched(err, false) : err;
+  return dispatch === "untagged" ? err : markDispatched(err, dispatch === "yes");
 };
 
-function bridge(opts: { fence: Fence; activeUuid: string | null | "throw"; tagged?: boolean }) {
+function bridge(opts: {
+  fence: Fence;
+  activeUuid: string | null | "throw";
+  dispatch?: Dispatch;
+  adoptThrows?: boolean;
+}) {
   const b = {
     send: async (cmd: Record<string, unknown>) => {
       if (cmd.cmd === "workflow_list") {
@@ -74,14 +81,19 @@ function bridge(opts: { fence: Fence; activeUuid: string | null | "throw"; tagge
         };
         return { active, workflows: [{ ...active, active: true }], active_confirmed: true };
       }
-      throw identityRefusal(opts.tagged ?? true);
+      throw identityRefusal(opts.dispatch ?? "no");
     },
     push: () => 1,
     canReach: (id: string) => id === TAB,
     isHeadless: () => false,
     tabs: () => [{ tab_id: TAB, title: "wf", connected_at: 0 }],
     resolveActiveTabId: () => TAB,
-    refreshWorkflowUuid: () => true,
+    // A THROW out of the adoption is `adopt_error`: it can land on either side of the
+    // fence write, so whether the fence moved is genuinely unknown.
+    refreshWorkflowUuid: () => {
+      if (opts.adoptThrows) throw new Error("the fence validator threw mid-write");
+      return true;
+    },
     workflowUuidFor: () => opts.fence,
     tabCanMutateGraph: () => true,
     tabGraphMutationCapability: () => ({ known: true, canMutate: true }),
@@ -92,7 +104,8 @@ function bridge(opts: { fence: Fence; activeUuid: string | null | "throw"; tagge
 async function setWidget(opts: {
   fence: Fence;
   activeUuid: string | null | "throw";
-  tagged?: boolean;
+  dispatch?: Dispatch;
+  adoptThrows?: boolean;
 }): Promise<{ text: string; res: ToolResult; fence: Record<string, unknown> | undefined }> {
   const ctx = makePanelToolCtx(bridge(opts), TAB, new WorkflowTargetStore());
   const def = buildPanelToolDefs().find((d) => d.name === "panel_set_widget");
@@ -106,8 +119,22 @@ async function setWidget(opts: {
   };
 }
 
-/** THE ONE THIS CALL REPAIRED: no fence on the tab, live canvas has an identity. */
-const REFRESHED = { fence: { known: false } as Fence, activeUuid: LIVE_UUID };
+/** THE ONE THIS CALL REPAIRED, with the prior fence DEFINITIVELY ABSENT.
+ *  `{known: true}` and not `{known: false}` — FenceRead's contract makes the first
+ *  "there is no fence" and the second "the read failed", and mislabelling the second
+ *  as the first is the exact fold this file is about. It is also the read that matches
+ *  the refusal: the bridge raises "no trusted identity" only when the resolver answered
+ *  and answered empty. */
+const REFRESHED = { fence: { known: true } as Fence, activeUuid: LIVE_UUID };
+/** Same `refreshed` verdict, DIFFERENT prior: the read itself failed. */
+const REFRESHED_UNREADABLE_PRIOR = { fence: { known: false } as Fence, activeUuid: LIVE_UUID };
+/** Same `refreshed` verdict, DIFFERENT prior again: a fence naming ANOTHER workflow,
+ *  which the adoption replaced. Reachable when the session is re-bound mid-check. */
+const OTHER_UUID = "99999999-9999-4999-8999-999999999999";
+const REFRESHED_OTHER_PRIOR = {
+  fence: { known: true, uuid: OTHER_UUID } as Fence,
+  activeUuid: LIVE_UUID,
+};
 /** THE ONE IT DID NOT: a fence read back already present and already equal. */
 const ALREADY_CURRENT = { fence: { known: true, uuid: LIVE_UUID } as Fence, activeUuid: LIVE_UUID };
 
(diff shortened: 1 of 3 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**. Verdict: **SHIP**, plus a punch-list the author has now cleared. Merging at `141cd18`.

**The P0 hunt came back clean.** The branch is *selected* by a prose match (`/no trusted identity/i`, `panel-tools.ts:1061`), which is the most dangerous shape in this codebase — so the question was what it gates. Exactly one consumer, and it gates only whether the diagnosis block runs and which explanation is emitted. **No retry-safety claim, no re-dispatch, no "nothing was applied" promise keys on it** — those all key on `dispatchOutcomeOf(err)`, the bridge's typed flag, killed in both directions by mutation (forcing `dispatched="no"` fails 2; forcing `retry_safe="yes"` fails 1). `panel_fence` has zero production consumers, so nothing re-dispatches on it. i18n degradation is fail-closed: the phrase is a plain English template literal at `ui-bridge.ts:4412` with no `trFor` call and zero occurrences in `web/js/`, so a translation would skip the block entirely — less information, never a false claim.

**`already_current` really is unreachable at dispatch.** `sendRouted` always passes `{tabId: ctx.tabId}`, so the bridge's `resolveTabWorkflowUuid` and `currentWorkflowFence`'s `workflowUuidFor` read the same resolver for the same key, and the phrase is the *else* of the two capability checks — minted only when `hasTrustedStamp` was false. No third mechanism was constructible.

**Two defects the gate found in this PR's own new prose, both now fixed rather than shipped:**

1. **`refreshed` asserted an absence it never observed.** It covers three priors — a genuinely absent fence, a *failed read*, and a *replaced* fence — and the new wording claimed "this session had NO fence for it", true for one. `main`'s wording was true for all three, so the change was strictly stronger and newly false in two. That is precisely the fold `FenceRead`'s own doc-comment forbids. Now each prior gets its own clause driven off `rebind.before`, and the observation rides in a **field** rather than in wording: `prior_fence: "absent" | "present" | "unreadable"` with `prior_fence_uuid` naming the replaced workflow. The `catch` reports `"unreadable"` too — a check that threw never made the observation.
2. **`nothingApplied` was two-way over a three-way flag**, telling an agent "this refusal carries no dispatch flag" when `dispatched === "yes"`. The author went further than the report: fixing the sentence alone would have left the same branch ordering **"RETRY THIS EXACT CALL ONCE" beside `retry_safe:"no"`**. A retry is now only *ordered* when the bridge proved nothing was written (`mayOrderRetry = dispatched === "no"`), and `next_action` degrades to a new `verify_applied_then_decide` on both non-proven states — applied to `already_current` as well, which had the same hole.

The optional third was taken: `adopt_error`'s tri-state now has a test (collapsing it previously left 46/46 green).

**Rework refutation, each mutation `git diff`-confirmed applied:** re-collapsing the three priors fails 2; reverting `nothingApplied` to two-way fails 1; ungating the retry order fails 2; collapsing the `adopt_error` tri-state fails 1. The original re-merge mutation still kills 4 against the final code. Suite 19/19.

**CI re-checked against a moved main, and it needed re-checking.** The run on `141cd18` is `event=pull_request` so it built `refs/pull/1740/merge` and all four checks pass — but main advanced ~6 minutes later, and **8 of the 13 commits it gained touch `panel-tools.ts`**, several fence-adjacent (#1519, #1650, #1639, panel#1292). A local test-merge of `origin/main` produced **no conflicts**, `tsc` clean, and `src/__tests__/orchestrator` serial on the merged tree at **173 files / 2916 tests, all passing**, including the eight suites those commits added. `mergeable=UNKNOWN` on the API is GitHub's lazy cache after the base moved, not a conflict.

**One earlier finding confirmed and worth recording:** `panel-session-rebind-residuals.test.ts` was driving `already_current` while asserting the `refreshed` wording — and passing, because one sentence covered both. The harness described a bridge state that cannot occur. It now pins `rebind_status` structurally.

Remaining non-blocking: `graph_run` is classified as a read by `isFencedGraphRead` (sound advice, false noun — the panel's own `READ_ONLY_GRAPH_COMMANDS` disagrees); `unstamped` wins over `stamped` in the classifier, not naturally reachable; and the pre-existing `healed_by_panel` tail inaccuracy, judged acceptable since `open_workflow` is safe in that state and `rebind_status` reports the truth the sentence does not.

PR #1753, opened 2026-08-19
feat(report): stamp WHICH LLM filed a panel bug report, mechanically

artokun:
A bug report filed from the panel is written by whichever model is selected in the
provider/model chips — Claude Opus, a hosted Kimi, a local gemma — and report quality
tracks that choice closely. Nothing in the filed issue said which, so a thin or wrong
report from a 4B model was indistinguishable from one from a frontier model, and the
question "are the bad reports coming from the non-SOTA backends?" could not be asked at
all, let alone across a batch of issues.

It cannot be self-reported. A model asked to name itself answers with a guess, and the
agent's ENVIRONMENT block never carried the model in the first place — only `Backend:`.
So the **orchestrator**, which owns the chips, publishes the fact, and `report_issue`
reads it back.

## What lands in the issue

```
Filed from the ComfyUI panel by **ollama** · model `gemma3:4b` · effort `low` — stamped by
comfyui-mcp from the panel's provider/model selection, not self-reported.
<!-- reporter-agent: backend=ollama model=gemma3:4b effort=low -->
```

…plus an `agent:<provider>` label, so the issues are filterable on GitHub. The **body** is
the channel that works today with no worker deploy: the intake worker pins the submitted
body verbatim into the created issue (`triage.ts` — `sanitized.body = opts.report.body`),
while an unknown top-level payload field would be silently dropped. Per-**provider** label,
not per-model — one label per provider is a bounded set, one per model is not; the exact
model stays in the body.

## How it gets there

`src/services/agent-identity.ts` is the transport `download-progress.ts` already
established, with the direction reversed: the orchestrator writes a small per-agent JSON,
the comfyui MCP subprocess reads it. No socket, no new HTTP surface, and it no-ops
entirely when the env var is unset — every non-panel spawn (plain stdio, CLI, tests) sends
a byte-identical body and label set to before.

**Why a file and not the spawn env.** `setModel` is live and deliberately does not restart
the session, so a value baked in at spawn goes stale the moment the user changes the chip —
and would then attribute reports to the model they switched *away* from, which is worse
than saying nothing. It is republished at **turn dispatch**: `"working"` fires before the
backend starts the turn, and therefore before any tool that turn calls. That is also why a
read has **no staleness cutoff** — a turn can legitimately run for an hour before it files,
and ageing the file out would drop attribution for exactly the longest sessions.

Both spawn lanes forward the path — the HTTP lane and the Claude lane, the one missed by
both prior incidents this file is annotated with. The path is port-scoped for the reason
`SessionStore` is: the agent key carries no port, so two ComfyUI instances on one machine
would otherwise stamp each other's model onto each other's reports.

Third-party trackers are untouched: our provider telemetry is not their business, and
`agent:ollama` is a label their repo has never heard of.

## Verification

**Mutation, on the wires and not just the helpers** — each of these turns the suite red:
cutting the Claude lane's env, cutting the HTTP lane's env, moving the publish to `"done"`,
latching a failed write into the dedupe cache, dropping the body stamp, dropping the label,
and inverting the stamp guard.

**Live, on the rig** — a real orchestrator on its own bridge port, driven over the real
panel bridge protocol, with a real Claude turn calling `report_issue` (`no_file:true`, so
nothing was filed):

```
--- identity at dispatch --- {"backend":"claude","model":"claude-opus-5","updated":…}
--- tool calls --- ["ToolSearch","mcp__comfyui__report_issue"]
--- agent reply ---
https://github.com/artokun/comfyui-mcp/issues/new?title=LIVE+CHECK…&body=…%0A---%0AFiled+from+the+ComfyUI+panel+by+**claude**+%C2%B7+model+%60claude-opus-5%60+…+not+self-reported.%0A%3C%21--+reporter-agent%3A+backend%3Dclaude+model%3Dclaude-opus-5+--%3E&labels=agent%3Aclaude
```

…and the same real built `dist/index.js` spawned as the MCP child both ways: with an
identity published (stamped, `agent:ollama` label) and without one (body and labels
byte-identical to the caller's).

### DIFF
--- a/src/__tests__/orchestrator/agent-identity-wiring.test.ts
+++ b/src/__tests__/orchestrator/agent-identity-wiring.test.ts
@@ -1,0 +1,113 @@
+// WHICH LLM FILED THE REPORT — the orchestrator half of the channel.
+//
+// report_issue stamps the reporting model into the issue body, but it can only
+// read what the orchestrator published, and it runs in a DIFFERENT PROCESS: the
+// comfyui MCP subprocess, which knows nothing about panels, backends or chips.
+// Two wires make it work, and either one missing fails silently — reports keep
+// filing, just anonymously, which is indistinguishable from "nobody filed any":
+//
+//   1. the identity file's path reaches the subprocess in its spawn env, on BOTH
+//      spawn lanes (the Claude lane's buildMcpServers and the HTTP lane's
+//      makeHttpBackendMcpServers);
+//   2. the orchestrator republishes the identity at TURN DISPATCH.
+//
+// Both live inline in the orchestrator's start closure, so — like the other
+// index.ts boundary guards (shared-session-invariant.test.ts,
+// ask-answer-journal.test.ts) — the wiring is pinned at source level while the
+// behaviour underneath is driven by agent-identity.test.ts against the real
+// module.
+//
+// The lane check is not ceremony: this file already carries two written-up
+// incidents of a spawn variable forwarded on one lane and not the other
+// (COMFYUI_MCP_TOOL_PRESET in panel-secrets.ts, the #547/#884 download stamp),
+// and the Claude lane — the DEFAULT backend — is the one that got missed both
+// times.
+
+import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
+
+const indexSrc = (): string =>
+  readFileSync(new URL("../../orchestrator/index.ts", import.meta.url), "utf8");
+
+/** The source of one inline arrow-function/object block, by its opening line. */
+function blockAfter(src: string, opener: string, endMarker: string): string {
+  const start = src.indexOf(opener);
+  expect(start, `opener not found: ${opener}`).toBeGreaterThan(-1);
+  const end = src.indexOf(endMarker, start);
+  expect(end, `end marker not found after ${opener}: ${endMarker}`).toBeGreaterThan(start);
+  return src.slice(start, end);
+}
+
+describe("agent identity reaches the MCP subprocess (both spawn lanes)", () => {
+  it("SOURCE: the HTTP lane (codex/gemini/ollama/…) forwards the identity path", () => {
+    const lane = blockAfter(
+      indexSrc(),
+      "const makeHttpBackendMcpServers = (tabId: string",
+      "// Live-graph panel_* tools for THIS tab",
+    );
+    expect(lane).toContain("...agentIdentityEnv(tabId)");
+    // Keyed by the same agent key the lane already self-scopes downloads with —
+    // one agent, one identity file, no cross-provider bleed when several
+    // backends are live at once (#884).
+    expect(lane).toContain("COMFYUI_MCP_TAB: tabId");
+  });
+
+  it("SOURCE: the CLAUDE lane forwards it too — the one that gets missed", () => {
+    const lane = blockAfter(
+      indexSrc(),
+      "const buildMcpServers = (agentKey?: string)",
+      "const manager = new PanelAgentManager",
+    );
+    expect(lane).toContain("...agentIdentityEnv(agentKey)");
+  });
+
+  it("SOURCE: the path is derived ONCE, from the bridge port and the agent key", () => {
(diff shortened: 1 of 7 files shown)


artokun:
## Merge gate — SUBSTITUTED REVIEW, disclosed

`codex-gate.mjs` returned **INDETERMINATE — account quota exhausted** on every attempt
(quota resets 2026-08-19 20:43), so this merged on the repo's fallback gate
(`claude-gate.mjs`): a fresh session that has never seen the implementation, handed only
the diff and told to refute, working the review taxonomy literally. That substitution is
only acceptable with this disclosure, so here it is — including what it cost.

**Four rounds. Two rounds found a real P1, both of them mine, both in code I had already
convinced myself was correct.**

### Round 1 — NO-SHIP, [P1] mis-attribution

The idempotency guard asked *"does this body contain a marker?"* when the only answerable
question is *"does it contain mine?"*. Every stamped issue in the tracker carries the
marker as quotable text, and quoting a related issue is ordinary agent behaviour — so a
body quoting `#1234`'s stamp was returned **unstamped**, and the report filed asserting
`claude-opus-5` wrote it, inside a sentence vouching that the attribution is mechanical,
while the label on the same issue said `agent:ollama`.

Fixed in `89a5e02`: there is no reachable double-stamp to protect against (one caller, raw
arg in, stamped result out), so the stamp is unconditional and a marker arriving inside the
caller's text is rewritten to `quoted-reporter-agent` on the way through.

### Round 2 — SHIP

Ran 10 of its own mutations (all killed) and spawned the real MCP child in **both** tool
modes — `full`, and `compact` via the `call_tool` router, which is what the HTTP lane
spawns for codex/gemini/ollama, i.e. exactly the backends this feature exists to identify.
Both returned a stamped body and `labels=agent:codex`.

It named two P2s and dropped them; both were taken anyway in `fdfa036`, since each is a
silent *loss* of attribution: the 60k body cap cutting the stamp off server-side, and a
dedupe cache that could suppress the rewrite for a process's lifetime if the file were
deleted underneath it.

### Round 3 — NO-SHIP, [P1] in the round-2 hardening

The oversize trim budgeted the stamp's own width and shortened the caller's **raw** text —
but stamping also rewrites each quoted marker (+7 chars each), so a body quoting other
issues shipped 7·n **over** the cap, the worker cut the tail, and the tail is the stamp. At
88 quoted markers the stamp block disappeared entirely. The covering test used a
marker-free body, so it passed with the rewrite present or absent: blind by construction to
the one interaction that mattered.

Fixed in `47a3f3c`: the trim now measures the body that ships and shortens by the actual
overflow.

### Round 4 — SHIP

Full suite green (550 files / 10,338 tests), `tsc --noEmit` clean, a 200-case fuzz of
`fitUnderWorkerCap` → `stampAgentIdentity`, and four targeted mutations (Claude-lane env
removed, `"working"`→`"done"`, the old `body.includes` shortcut restored, the pre-round-3
budget restored) — all killed. It confirmed the one-correction bound algebraically, and
verified `buildComfyuiMcpEnv` (which *deletes* keys its resolver did not supply) really does
let `COMFYUI_MCP_AGENT_IDENTITY` through to the child.

**One P2 it recorded and dropped, carried here rather than lost:** the worker scrubs secrets
*before* it truncates, and scrubbing can grow text (`TOKEN=abcdef` → `TOKEN=[REDACTED]`,
+4). A ~60 kB body carrying ≥48 secret-shaped assignments with 6–9 char values would exceed
the cap server-side and lose the stamp. Contrived, and it fails to *unstamped* — the
pre-change state — never to a wrong model.

### Independent of the gate

Seven mutations on the original change, three on the round-2 hardening, one on the round-3
fix — each turns the suite red. Live on the rig: a real orchestrator on its own bridge port
driven over the real panel bridge protocol, a real Claude turn calling `report_issue`
(`no_file:true`, nothing filed) returning a body stamped `claude-opus-5` with
`labels=agent:claude`; plus the real built `dist/index.js` spawned as the MCP child with an
identity published, without one (byte-identical passthrough), and with a body quoting
another issue's stamp.

Issue #1704, opened 2026-08-18
restart_comfyui relaunches Stability Matrix Assets python instead of the package venv (sqlalchemy missing, process exits 1)


artokun:
### What happened / root cause
On a Stability Matrix ComfyUI install, `restart_comfyui action:restart` stopped the healthy venv process and relaunched with:
`E:\comfy\StabilityMatrix\Assets\Python\cpython-3.12.11-windows-x86_64-none\python.exe`
instead of the actual running interpreter:
`E:\comfy\StabilityMatrix\Packages\ComfyUI\venv\Scripts\python.exe`
The Assets interpreter does not have this install's site-packages (`ModuleNotFoundError: No module named 'sqlalchemy'`), so the relaunch exited 1 and left ComfyUI down. I had to start the venv python by hand to recover.

The original process argv was already correct (`main.py --preview-method auto --use-sage-attention --disable-xformers --fast`). Only the executable was wrong.

### Steps to reproduce
1. Run ComfyUI via Stability Matrix so the live process is `Packages\ComfyUI\venv\Scripts\python.exe`.
2. Call `restart_comfyui {action: restart}`.
3. Watch it launch `Assets\Python\cpython-3.12.11-...\python.exe` and die on `import sqlalchemy`.

### Exact error (scrubbed)
ComfyUI was stopped but could not be started: process launched, no readiness in 59666ms (60/60). The process this call launched EXITED (exit code 1).
`ModuleNotFoundError: No module named 'sqlalchemy'`
exe used: `E:\comfy\StabilityMatrix\Assets\Python\cpython-3.12.11-windows-x86_64-none\python.exe`
working exe: `E:\comfy\StabilityMatrix\Packages\ComfyUI\venv\Scripts\python.exe`

### Fix
<applied locally: no>  <upstream-only: yes>
Relaunch MUST reuse the executable of the process that was just stopped (or the package venv), not Stability Matrix's unused Assets CPython. Probe `sys.executable` from `/system_stats` or the stopped PID command line before picking an interpreter.

### Environment
OS Windows 11 / ComfyUI 0.33.0 / frontend 1.49.6 / GPU NVIDIA GeForce RTX 5060 Ti 16GB / torch 2.13.0+cu130 / python 3.12.11 / comfyui-mcp 0.52.1 / panel 0.14.0

artokun:
[...]

Issue #1702, opened 2026-08-18
download_civitai reports a non-ASCII CIVITAI_API_TOKEN as "CivitAI is unreachable (network error)"


artokun:
### What happened / root cause

`download_model action:"download_civitai"` failed with a **connectivity** error when the real fault was a **local credential-encoding** problem.

The user's `CIVITAI_API_TOKEN` had been saved (via `panel_request_secret`) with non-ASCII characters in it. `authHeaders()` builds `Authorization: [REDACTED] <token>`, and Node's `fetch` rejects a non-ASCII header value with a **`TypeError`**:

```
Cannot convert argument to a ByteString because the character at index 10 has a value of 12503 which is greater than 255.
```

In `dist/services/civitai-resolver.js`, `transportError()` maps **any** `TypeError` to a network failure:

```js
if (err instanceof TypeError) {
    return new ModelError(`CivitAI is unreachable (network error: ${err.message}) - ` +
        `check connectivity and try again.`, { url, network: true });
}
```

That assumption holds for the common `TypeError: fetch failed` (DNS/TLS/connection), but **an invalid header value throws the same type before any socket is opened**. The result is `network: true` on a fault that has nothing to do with the network, and a message telling the user to "check connectivity" when connectivity was fine.

Cost: the agent went down the wrong diagnostic path (connectivity / CivitAI outage) instead of the correct one (re-enter the token).

Secondary issue: `panel_request_secret` accepted the non-ASCII value for `CIVITAI_API_TOKEN` and reported success ("verified by reading the file back"), even though such a value can never produce a valid HTTP header. The read-back verification confirms the bytes landed, not that they are usable.

### Steps to reproduce

1. Save a `CIVITAI_API_TOKEN` containing any non-ASCII character (e.g. Japanese text) - via `panel_request_secret` (target `comfyui`, env `CIVITAI_API_TOKEN`) or directly in `~/.comfyui-mcp/.env`. It saves and reports success.
2. Call `download_model action:"download_civitai"` with any `model_version_id`.
3. Observe the error is reported as a network/connectivity failure.

Minimal repro of the underlying throw:

```js
fetch('https://example.com', { headers: { Authorization: [REDACTED] } })
  .catch(e => console.log(e.name, e.message));
// TypeError Cannot convert argument to a ByteString because the character at index 7 has a value of 12486 which is greater than 255.
```

### Exact error (scrubbed)

```json
[...]

Issue #1701, opened 2026-08-18
Download completion event fires before status is done and before final file exists


artokun:
### What happened / root cause
The panel emitted a "Model download transfer completed" event for t5xxl_flux1_int8_convrot.safetensors while download_model action:"status" still reported downloading, and the destination file did not yet exist. The completion event is racing ahead of the verified download state/file finalization.

### Steps to reproduce
1. Start a large local Hugging Face transfer with download_model action:"download".
2. Wait for the panel completion event.
3. Immediately call download_model action:"status" and check the destination.
4. Status remains downloading and the final file is absent.

### Exact error
Completion event claimed the transfer completed; status returned downloading after 3608s; destination models/text_encoders/t5xxl_flux1_int8_convrot.safetensors was not present.

### Fix
applied locally: no
upstream-only: yes
Only emit the completion event after the same verified finalization state used by action:"status" confirms the final path exists and the transfer record is done. Alternatively label the event as bytes-received/unverified until finalization completes.

### Environment
Windows 11; ComfyUI 0.33.1; frontend version unavailable; NVIDIA RTX 3060 12GB; comfyui-mcp 0.52.1; panel 0.15.1.

artokun:
Claimed by grok-swarm 2026-08-19T07:30:00Z — driving this in worktree `wt-1701` (branch `fix/1701-download-complete-before-done`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp/pull/1760 (worktree C:/Users/Artokun/code/comfyui-mcp/.worktrees/wt-1701).

Issue #1700, opened 2026-08-18
panel_remove_mcp + panel_reload does not drop the removed server: resumed session keeps the old MCP set and keeps reporting it as failed


artokun:
### What happened / root cause

`panel_remove_mcp` returns `Removed MCP server "magic". Call panel_reload to apply.` and `panel_reload`'s own description promises the orchestrator scope reloads *"agent config, MCP servers (panel_add_mcp/panel_remove_mcp), the system prompt…"*.

It does not. After `panel_remove_mcp("magic")` + `panel_reload`, the respawned session **still started with `magic` in its MCP set** and the panel kept printing the failure banner every turn:

```
MCP server magic (failed) is unavailable to the agent — the automatic reconnect did not bring it back.
⚠️ This session started without 1 of its MCP server: magic (failed).
```

Meanwhile `panel_list_mcp` (which re-reads the config file) already reported the server as gone — so the tool that *reads* the config and the session that *loads* it disagreed. That disagreement is the confusing part: from the agent's side everything looked applied, and I told the user it was, which was wrong.

The removal itself was correct — verified `~/.claude.json` after the call:

```
top-level mcpServers: ['codegraph', 'touchdesigner']
```

and a grep across `~/.claude/` found `magic` only in transcript `.jsonl` files and `~/.claude/backups/.claude.json.backup.*` — no live config anywhere.

Root cause (best reading of `dist/orchestrator/`): `panel_reload` respawns the agent by **resuming** the existing session. On the Claude backend the banner comes from `claude-backend.js:897`, `inspectMcpServers(Object.keys(this.mcpServersForRun() ?? {}), message.mcp_servers)` — i.e. it reflects the `mcp_servers` the SDK reports at init. A resumed session appears to restore the MCP server set recorded with that session rather than re-reading `~/.claude.json`, so a server removed between the two runs is still launched (and still fails), forever, for the life of that session.

Confirmed by the counter-case: later in the same working session the connection dropped and the panel rebuilt the session from scratch. That fresh session started **without** `magic` and without the banner, `panel_list_mcp` unchanged. So a full session rebuild applies the removal; `panel_reload` does not.

### Steps to reproduce

1. Have an inherited stdio MCP server in `~/.claude.json` that fails to connect (any bad command / revoked key will do — here `@21st-dev/magic` returning `-32001 Not authenticated`).
2. `panel_remove_mcp({name: "magic"})` → returns "Call panel_reload to apply."
[...]

PR #1757, opened 2026-08-19
release: v0.52.4

artokun:
Patch release **v0.52.4**. Changelog generated from commits since v0.52.3.

Do **not** squash-merge — the `v0.52.4` tag already points at this commit. Rebase-merge or merge-commit so the tag stays on main.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -759,6 +759,33 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+## [0.52.4] - 2026-08-19
+
+### MCP
+
+#### Fixed
+- a fence this call REPAIRED is not a fence that was already fine — and the verdict rides in a field (#1740)
+- unwind call_tool when get_system_stats times out mid-decode (#1751)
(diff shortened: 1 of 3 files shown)


artokun:
**Orchestrator note before merging.** This sat green and untouched for ~50 minutes, so I am completing it rather than leaving `main` unreconciled.

Unlike the panel's release, this one needed **no changelog repair**. `v0.52.4` is already tagged at `2fa9999` and its `publish` workflow succeeded, so the tag shipped exactly the content on this branch. The 8 commits that landed on `main` after the cut — #1698, #1695, #1700, #1702, #1701, #1704, #1699, and the LLM-stamping change — are **not** in `v0.52.4` and correctly belong to the next cut, so adding them here would have documented them as shipped when they were not.

Merging on required checks: Build passes on ubuntu, windows and macos, Pack & install smoke passes, and `publish` passed twice. The lone `packs` check has been pending since 07:38 and is stale rather than failing; it is not a required context and every other signal is green.

This is the protected-main reconcile — the tag publishes and this brings `main`'s version and changelog into line with what shipped.

PR #1766, opened 2026-08-19
feat: support split ComfyUI code and data roots

### LINKED ISSUE #1765: Support split ComfyUI code and data roots
## Problem

Some valid ComfyUI deployments separate the checkout from mutable data:

- `main.py`, `.venv`, and `custom_nodes` live in a code checkout
- `models`, `input`, `output`, `temp`, and `user` live under a shared or mounted data root selected by launch flags / extra-model-path configuration

This shape is related to the data-only `COMFYUI_PATH` reported in #851 and the multi-root Desktop layouts discussed in #766. Today comfyui-mcp has one `COMFYUI_PATH` for both concerns, so there is no correct value:

- pointing it at data keeps output/user/model operations correct, but filesystem comfy-cli fallbacks, core updates, custom-node provenance, and legacy Manager snapshot lookup can target the wrong checkout;
- pointing it at code fixes those operations but makes data-facing tools lose the paths the running server actually uses.

`apply_manifest` also combines both classes: model placement is data-facing, while pip and custom-node filesystem fallbacks are code-facing.

## Proposed shape

Add an optional `COMFYUI_CODE_PATH` as an additive code-checkout fallback while keeping `COMFYUI_PATH` backward-compatible as the data/base root:

1. For asynchronous code mutations, prefer the reachable local server's observed `main.py` root, then `COMFYUI_CODE_PATH`, then the legacy effective `COMFYUI_PATH`/workspace.
2. Route comfy-cli node mutations, core update, manifest pip/node fallbacks, and custom-node provenance through the code root.
3. Keep models, input/output, user state, and modern Manager snapshots on their existing live/component/data roots.
4. Suppress both local roots for remote targets, as today.
5. Surface both paths in the environment report and forward the code path to panel-spawned MCP children.
6. Preserve current behavior exactly when `COMFYUI_CODE_PATH` is unset.

This avoids adding a policy guard or disabling tools: each consumer receives the component root it actually needs.

woodenriver05:
## Summary

- add `COMFYUI_CODE_PATH` for deployments whose checkout (`main.py`, `.venv`, `custom_nodes`) is separate from their data/model/user root
- route code-facing operations to that root while leaving models, input/output, user state, and modern Manager snapshots on their existing live/component/data roots
- make `apply_manifest` carry data and code roots independently, including pip and custom-node filesystem fallbacks
- expose `local.code_path` / `local.code_path_source` in the environment report and forward the setting to panel-spawned MCP children
- preserve the existing single-root behavior when `COMFYUI_CODE_PATH` is unset and continue suppressing local filesystem roots for remote targets

## Why

A valid split deployment has no correct single `COMFYUI_PATH`: choosing the data root breaks code-facing fallbacks, while choosing the checkout breaks output/user/model operations. This change supplies each consumer with the component root it actually needs instead of disabling tools or adding policy-only guards.

For live-aware mutations (core update and manifest dependency work), the connected local server's observed `main.py` checkout wins, followed by `COMFYUI_CODE_PATH`, then the legacy effective workspace. Synchronous comfy-cli/provenance paths use `COMFYUI_CODE_PATH` with the same legacy fallback. Model resolution remains live/extra-root aware.

Closes #1765. Related deployment shapes: #851, #766.

## Validation

- `npm run build` — pass
- `npm run lint` — pass
- focused split-root + adjacent regression suite — 10 files, 490 tests passed
- `npm run docs:gen` plus locale/MDX/link/unknown-collapse checks — pass
- i18n and blog checks — pass
- full Vitest stage — 10,384 passed; 13 host-dependent failures in 6 pre-existing files

The same six files and exact 13 failures reproduce on an isolated worktree at current `origin/main` (`9d3e637e`): 102 passed / 13 failed. They are the repository's existing Windows short-vs-long temp-path and active Codex-login leakage failures (`comfy-view-ref`, `live-extra-roots`, `train-dataset-refs`, `oauth-landing`, `code-provider-auth`, `backend-readiness`), not failures introduced by this branch.


artokun (dismissed):
Thanks — this is a careful patch: the fail-closed guards hold (remote refusals, junction rules, live-root anchoring all preserved; workflow-lock actually gains a remote refusal), and the targeted suites pass. Two blocking issues before this can merge:

1. **Conflicts with main, and the conflict is semantic, not just textual.** #1770 (merged today) established that on `--base-directory` deployments (ComfyUI Desktop — the canonical split install) the runtime scans `custom_nodes/` from the BASE/data directory, not the main.py install root. This PR routes custom-node targets (clone fallback, ref checkout, enable/disable/uninstall CLI workspace, workflow-lock pack commits) to the code root — on Desktop with `COMFYUI_PATH=<data dir>` that clones packs into `<install tree>/custom_nodes` where the runtime never looks, the exact wrong-root bug class #1770 just fixed for node_pack. Please rebase and reconcile explicitly: code root = main.py root (pip/venv/core git), data root = base-directory-aware live base (pack reads/writes). Also note the new live code resolver currently bypasses #1770's `hasUnresolvableRelativeBaseDirFlag` fail-closed gate — needs a conscious decision either way.

2. **`getEnvironment` mislabels the code root as `local.workspace_path`** when only `COMFYUI_CODE_PATH` is set — that field means the data root everywhere else.

Minor: `apply_manifest` now prefers live-main.py for pip/codeBase but config-first for dataBase, so one manifest can split mutations across two installs when they differ — align or document.

I've left the CI runs unapproved on purpose — no point burning them on a stale base. Happy to re-review after the rebase.

artokun:
Rebased onto current `origin/main` (was CONFLICTING) and pushed to this branch at `3fdba49`.

woodenriver05's original commit is preserved as `19408d9` (`feat(paths): support split ComfyUI code roots`). Follow-up `3fdba49` addresses the review without changing that design (`COMFYUI_CODE_PATH` = checkout; `COMFYUI_PATH` / live `--base-directory` = data):

1. **#1770 reconciliation** — pack clone/checkout, comfy-cli `--workspace`, enable/disable/uninstall, workflow-lock pack commits, and `apply_manifest` node fallbacks now write/read `custom_nodes` on the live data/base root. Code root stays pip/venv/core git.
2. **`getEnvironment`** — `local.workspace_path` is no longer labeled with `COMFYUI_CODE_PATH` when only the code override is set.
3. **`hasUnresolvableRelativeBaseDirFlag`** — data resolver still fail-closes. Code resolver still adopts live `main.py` in that case (that *is* the checkout).
4. **`apply_manifest`** — both roots are now live-first (`BaseLive` for data, `CodeBaseLive` for pip) so one apply cannot split mutations across two installs.

Targeted suites (`config`, `workspace-env`, `comfy-cli`, `manifest`, `update-comfyui`, plus `node-management` / `node-snapshots` / `workflow-lock` / `panel-secrets`): 499 passed after the follow-up.

Ready for re-review. CI was left unapproved on the stale base — please re-approve the workflow runs on this head.

PR #1771, opened 2026-08-19
fix(1767): clean the three packs whose workflows outran their manifests, and make packs:check-models a CI gate

### LINKED ISSUE #1767: 3 packs ship workflows referencing weights their manifest never downloads, and packs:check-models is run by nothing
Found while triaging mcp#1609 for parking. The gate that issue proposed does not exist on `main` — but the condition it was meant to catch is real and shipping today, so recording it as its own bug.

**3 of 57 packs on `origin/main` ship workflows that reference weights their manifest never downloads.** Installing the pack and opening its workflow gives a missing-model error.

| pack | referenced but never downloaded |
|---|---|
| `z-image-xy-plot` | 5 `Margot-Robbie-Z3*` LoRAs, plus a commented-out `z_image_turbo-Q8_0.gguf` |
| `ltx-2.3-xy-plot` | 3 `Paulie-Walnuts-LTX23*` LoRAs |
| `wan-pusa-extend` | `umt5_xxl_fp16.safetensors`, `taew2_1.safetensors` |

**Two of these look like personal training artifacts that leaked out of an authoring machine** — the `Margot-Robbie-Z3*` and `Paulie-Walnuts-LTX23*` LoRAs are not published weights anyone installing the pack could obtain. That is worth a decision beyond the missing-model bug: those workflows probably want re-authoring against public weights rather than having the manifest extended to fetch something that cannot be fetched.

`wan-pusa-extend` is the straightforward case — both files are ordinary public weights that the manifest simply omits.

**Why this can ship undetected:** `packs:check-models` exists but **nothing runs it** — not CI, not the release workflow. Wiring it in is the durable fix, and cleaning these three is its precondition, since it would fail on them today.

Suggested order:
1. Fix the three packs (manifest additions for `wan-pusa-extend`; a decision on the two LoRA sets).
2. Wire `packs:check-models` into CI so a fourth cannot appear silently.

Not observed in a user report — found by inspection while triaging. Severity is P2 rather than higher because it affects three packs rather than the install path generally, but it is a guaranteed failure for anyone who installs those three.

artokun:
Fixes #1767.

`packs:check-models` has existed for months and was run by **nothing** — not CI, not the release workflow. Wiring it in is the durable fix; cleaning what it finds is the precondition. Both are here.

Investigating first changed the shape of the bug in three ways, so the diff is not what the issue asked for.

## 1. Five of the seven reported gaps were the gate's own false positives

When a widget is promoted to an input and connected, litegraph keeps the old string in `widgets_values` forever, but `graphToPrompt` builds that input from the **link** — the stored string is dead and ComfyUI never loads it. Five `LoraLoader`s in `z-image-xy-plot` and two in `ltx-2.3-xy-plot` are exactly this, carrying a stale filename from before the author wired up the `easy loraNames` selectors:

```
LoraLoader#91.lora_name <= Reroute#93 -> easy loraNames#92 wv=["Margot-Robbie-Z3_000001000.safetensors"]
```

`01margott_W_000001000.safetensors` — reported on 5 nodes — is never loaded by any install. A gate with a 5-in-7 false-positive rate cannot be wired into CI, so the gate is fixed first.

**No coverage is lost.** Each dead slot's link resolves back (through Reroutes) to a selector node whose own widget value is scanned on its own turn, so the gap is still reported — against the node that actually supplies the value. Traced for all 7; one of the tests pins it.

The script has no `object_info` and cannot map a positional slot back to a widget name, so it decides by **arity** and refuses to guess: only when every model string on a node is accounted for by a linked model widget are they all treated as link-supplied. A node where the counts disagree is scanned in full — a false positive can be silenced explicitly, a false negative ships a broken pack. Measured across all 57 packs: 9 nodes have a linked widget plus a model string, 7 unambiguous, and the 2 ambiguous ones (`WanVideoTextEncodeCached` with `positive_prompt` linked) are real live values already in their manifests, untouched by the rule.

## 2. z-image-xy-plot's real defect was in its manifest, not its workflow

All 13 sibling `z-image-*` packs ship the **Q8_0** tier active and select Q8_0. This pack alone activated Q5_K_S while its `UnetLoaderGGUF` asked for Q8_0, so a fresh install opened straight to a missing-model error — and `pack.yaml` documented the workaround instead of fixing it. The manifest now ships Q8_0, the byte-identical URL 12 other packs already download (HEAD 200; all three tier URLs verified). This also clears the mirror-image `downloads Q5_K_S but no loader references it` warning.

Fixing the workflow instead would have diverged this pack from its whole family.

## 3. wan-pusa-extend needed no change — the issue was wrong about it

The issue called this "the straightforward case — both files are ordinary public weights that the manifest simply omits." They are not omitted; both sit on `mode: 2` (**bypassed**) alternative nodes:

| node | why it's bypassed |
|---|---|
| `CLIPLoader#48` → `umt5_xxl_fp16.safetensors` | superseded by `umt5-xxl-enc-bf16.safetensors`, the WanVideoWrapper-native encoder the pack **does** ship |
| `WanVideoTinyVAELoader#118` → `taew2_1.safetensors` | optional preview VAE |

The gate already classifies both correctly as warnings. Adding `umt5_xxl_fp16.safetensors` would download ~11 GB that nothing active loads — a bloat regression, not a fix. **No change made.**

## The decision the issue asked for on the two LoRA sets

Re-authoring against public weights, as the issue suggested, is **not possible**: both packs compare successive checkpoints of *one* training run (`_000001000`, `_000001500`, `_000002000`, `_000002500`, final), and no public weight set has that shape. Both packs are bring-your-own-LoRA by design — `z-image-xy-plot` already said so in its notes, `ltx-2.3-xy-plot` did not, and now does.

So the leaked personal artifacts are replaced with self-documenting placeholders and declared in `pack.yaml` `external_models`, the mechanism that already existed for a reviewed intentional gap:

- `Margot-Robbie-Z3*` (5 live selectors) → `YOUR-LORA-step1000/1500/2000/2500`, `YOUR-LORA-final`
- `01margott_W_*` (5 dead slots) → the placeholder each one's link actually supplies, so the JSON reads consistently
- `Paulie-Walnuts-LTX23*` → `YOUR-LTX-LORA-a/b`

The example prompts naming the same two people are neutralized so the shipped graph is coherent for a BYOL user. (The first pass missed `Chrissy` in the ltx dialogue; the gate caught it and it is now scrubbed too — see the gate section.) `ltx-2.3-xy-plot` also drops a now-unnecessary `external_models` entry (`Paulie-Walnuts-LTX23_000004250`, which only ever appeared on the two dead slots).

Note this makes the gate's pass on real pack data independent of the gate fix — cleaning the dead slots means pack data no longer exercises the dead-slot path. That is deliberate: the unit test is the pin, not shipped stale data that a future maintainer would break by tidying.

## Verified by mutation, not by green

| mutation | result |
|---|---|
| remove the dead-slot skip in `check-pack-models.mjs` | **3 of 8** gate tests fail |
| remove the `npm run packs:check-models` step from `packs.yml` | wiring test fails |
| remove `scripts/check-pack-models.mjs` from the `paths` filters | 2 wiring tests fail |

Each mutation was asserted to have actually applied before running, and reverted after.

- `packs:check-models`: **7 errors → 0** across all 57 packs
- `packs:validate`: green; `packs:gen` regenerated the two installers (the Q8_0 tier swap is the only functional line)
- targeted suites: 12 files / 155 tests green
- full suite: 10,426 tests green (3 timeout flakes under concurrent load; all 103 tests in those files pass in isolation, none touch this diff)
- **real CI**: the `packs` job passed with the new step executing `node scripts/check-pack-models.mjs` on ubuntu — wiring proven in CI, not just locally

The `paths` filter now includes the gate's own source — otherwise editing the gate would not run the gate.

## Gate

**Verdict: SHIP.**

codex was unavailable (the account is exhausted until 2026-08-19 20:43, measured on both the wrapper and `codex exec` directly); gated by an independent adversarial review instead — a fresh session that never saw this reasoning, handed only the diff and told to refute. Disclosing the substitution is the condition it was accepted under.

It worked the repo's seven-class defect taxonomy by executing rather than reading, and independently re-derived the load-bearing claims:

- **Two states collapsing** — measured all 3475 links across 57 packs: minimum link id is **1**, zero links with id `0`, so "unconnected" (`link == null`) cannot collapse with "connected to link 0".
- **Arity blind spot** — it built the adversarial input (2 linked asset widgets, 1 typed model string → node skipped) and confirmed it is real, that the code comment already discloses it, and that incidence in the repo is **zero** (all 7 skipped nodes are 1:1 `lora_name` on `LoraLoader`).
- **Mutation coverage** — 5 mutations, all killed, including two I had not tried (`resolve`→`join`, ignoring `ASSET_WIDGET_NAMES`).
- **Prose vs code** — verified "what every other z-image-* pack ships" and ltx's "the detailer and distill LoRAs ARE downloaded by this pack" against the manifests.

Worth recording: its CI-step mutation **first read as survived** because CRLF line endings ate its `perl` anchors, so the deletion never applied. It caught that itself and re-ran with the deletion verified against the file, where it died. A mutation that never applied is indistinguishable from one that survived unless you check that it landed.

**Post-gate delta, disclosed:** the gate listed as a *non-finding* that two `Chrissy` strings survived my likeness scrub in the ltx prompt dialogue. Not a correctness defect, but it made the PR text above inaccurate, so I scrubbed them (`Chrissy` → `my cousin`, 2 strings in `packs/ltx-2.3-xy-plot/workflow.json`) after the SHIP verdict. That is the only change since the gated tree: pack prompt prose, no code. Re-verified after it — `packs:check-models` 0 errors, `packs:validate` clean, no installer drift, 79 pack/script tests green.

The gate's other non-finding — a non-array `node.inputs` throws a `TypeError` — I left deliberately: it fails **closed** (non-zero exit turns CI red) on a shape litegraph does not emit, and changing it post-gate would alter code behaviour for no user-visible gain.

## Not done

- No browser/Playwright run: this touches pack data, a standalone node script and CI config. Nothing here renders in the panel.
- `docs/blog/z-image-comfyui.mdx:265` claims "there's no Q6_K on the mirror". That is stale — Q6_K returns HTTP 200. Left alone as an unrelated docs bug.

### DIFF
--- a/.github/workflows/packs.yml
+++ b/.github/workflows/packs.yml
@@ -8,6 +8,7 @@ on:
       - "scripts/validate-manifests.mjs"
       - "scripts/test-packs.sh"
       - "scripts/check-model-urls.mjs"
+      - "scripts/check-pack-models.mjs"
       - "src/services/manifest.ts"
       - ".github/workflows/packs.yml"
   pull_request:
@@ -17,6 +18,7 @@ on:
       - "scripts/validate-manifests.mjs"
       - "scripts/test-packs.sh"
       - "scripts/check-model-urls.mjs"
+      - "scripts/check-pack-models.mjs"
       - "src/services/manifest.ts"
       - ".github/workflows/packs.yml"
 
@@ -58,6 +60,9 @@ jobs:
 
       - name: Dry-run install scripts (offline, git/curl stubbed)
         run: bash scripts/test-packs.sh
+
+      - name: Workflows only reference models their manifest downloads
+        run: npm run packs:check-models
 
       - name: Validate model URLs + payload sizes (network, no full downloads)
         run: node scripts/check-model-urls.mjs
--- a/packs/ltx-2.3-xy-plot/pack.yaml
+++ b/packs/ltx-2.3-xy-plot/pack.yaml
@@ -16,9 +16,20 @@ notes:
   - Run the generated installer from your ComfyUI root.
   - Sliced + un-bypassed from the LTX-2.3 ULTRA workflow.
   - Needs the kornia fix (fix-ltxvideo-kornia.sh) if ComfyUI-LTXVideo's pyramid import breaks.
+  - >-
+    Bring your own trained LTX-2.3 LoRAs. This pack installs NO subject LoRAs —
+    the two "easy loraNames" selectors ship the placeholders
+    YOUR-LTX-LORA-a/b.safetensors, which are the two axes of the comparison grid.
+    Drop your own .safetensors into models/loras/ and pick them there. The
+    detailer and distill LoRAs the graph also loads ARE downloaded by this pack.
 post_install:
   - Restart ComfyUI, load workflow.json.
+  - >-
+    Set the two "easy loraNames" nodes to your own trained LoRAs — until you do,
+    they point at YOUR-LTX-LORA-a/b.safetensors placeholders and ComfyUI will
+    report them missing.
+# Declared external so packs:check-models knows they're intentionally not shipped.
+# Placeholders, not real weights — see the "bring your own" note above (#1767).
 external_models:
-  - Paulie-Walnuts-LTX23_000004250.safetensors
-  - Paulie-Walnuts-LTX23_000008000.safetensors
-  - paulie-walnuts-ltx23-03.safetensors
+  - YOUR-LTX-LORA-a.safetensors
+  - YOUR-LTX-LORA-b.safetensors
--- a/packs/ltx-2.3-xy-plot/workflow.json
+++ b/packs/ltx-2.3-xy-plot/workflow.json
@@ -2750,7 +2750,7 @@
         "secondTabWidth": 65
       },
       "widgets_values": [
-        "Paulie Walnuts sitting at Satriale's in a beige bomber jacket. He raises his right hand with a flat palm in a \"stop\" gesture and says, \"that was different, Chrissy was shot. That was a paranormal event.\"\n"
+        "A man sitting in a diner booth in a beige bomber jacket. He raises his right hand with a flat palm in a \"stop\" gesture and says, \"that was different, my cousin was shot. That was a paranormal event.\"\n"
       ],
       "color": "#009933",
       "bgcolor": "#353"
@@ -5720,7 +5720,7 @@
         "secondTabWidth": 65
       },
       "widgets_values": [
-        "Paulie Walnuts sitting at Satriale's in a beige bomber jacket. He raises his right hand with a flat palm in a \"stop\" gesture and says, \"that was different, Chrissy was shot. That was a paranormal event.\"\n"
+        "A man sitting in a diner booth in a beige bomber jacket. He raises his right hand with a flat palm in a \"stop\" gesture and says, \"that was different, my cousin was shot. That was a paranormal event.\"\n"
       ],
       "color": "#009933",
       "bgcolor": "#353"
@@ -5851,7 +5851,7 @@
         }
       },
       "widgets_values": [
-        "Paulie-Walnuts-LTX23_000004250.safetensors",
+        "YOUR-LTX-LORA-b.safetensors",
         1,
         1
       ],
@@ -6508,7 +6508,7 @@
         }
       },
       "widgets_values": [
-        "Paulie-Walnuts-LTX23_000008000.safetensors"
+        "YOUR-LTX-LORA-a.safetensors"
       ]
     },
     {
@@ -6546,7 +6546,7 @@
         }
       },
       "widgets_values": [
-        "paulie-walnuts-ltx23-03.safetensors"
+        "YOUR-LTX-LORA-b.safetensors"
       ]
     },
     {
@@ -6799,7 +6799,7 @@
         }
       },
       "widgets_values": [
-        "Paulie-Walnuts-LTX23_000004250.safetensors",
+        "YOUR-LTX-LORA-a.safetensors",
         1,
         1
       ],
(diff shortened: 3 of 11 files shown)


artokun:
Independent replication of this PR's central claim, from a separate session that was dispatched at #1767 and stood down on finding this open. I did not touch this branch or worktree.

The load-bearing claim is "skipping link-fed widget slots loses no coverage." I tested it by executing rather than reading: ran **this PR's** `check-pack-models.mjs` against **unchanged `origin/main`** pack data and diffed against the current gate's output. The entire delta is one line:

```
< FAIL z-image-xy-plot — workflow references "01margott_W_000001000.safetensors" not downloaded by manifest [LoraLoader#91, LoraLoader#86, LoraLoader#99, LoraLoader#113, LoraLoader#129]
```

The other 6 errors and all 102 warnings survive byte-identical, so the rule removes the false positive and nothing else. I also enumerated every node the arity rule skips across all 57 packs: exactly 7, all `modelStrings=1 / linkedAssetWidgets=1` on `lora_name`, each link traced back to an `easy loraNames` selector that is still scanned on its own turn. The documented residual risk (a linked slot holding a non-model string) occurs in zero bundled packs. Full tree: **0 errors, 101 warnings**, confirmed here.

I separately confirmed the two corrections to the issue: `wan-pusa-extend`'s `CLIPLoader#48` and `WanVideoTinyVAELoader#118` are both `mode: 2` while the live encoder `umt5-xxl-enc-bf16.safetensors` is shipped, and `UnetLoaderGGUF#10` is live on Q8_0 against a Q5_K_S manifest. Both match this PR.

Two notes, neither blocking:

1. The body says "9 nodes have a linked widget plus a model string, 7 unambiguous." Counting linked *asset* widgets I get 7 total, 7 skipped — the two `WanVideoTextEncodeCached` nodes have `positive_prompt` linked, which isn't in `ASSET_WIDGET_NAMES`, so they never enter the comparison rather than entering it as ambiguous. Same conclusion, looser denominator.

2. The placeholder swap (`YOUR-LORA-*`, `YOUR-LTX-LORA-*`) plus the example-prompt neutralization is the product decision #1767 explicitly asked to be escalated rather than made. It's a defensible choice and cleanly executed — but it's the owner's content, with a real-person-likeness dimension, so it's worth an explicit yes/no rather than passing through on a green gate. I've written the options and evidence up in https://github.com/artokun/comfyui-mcp/issues/1767#issuecomment-5340209281 so the call can be made on the record.

This is a replication, not a gate.

Issue #1656, opened 2026-08-17
panel_* workflow routing silently targeted previous canvas after tab switch


artokun:
### What happened / root cause
After the panel reported that the active workflow switched from a long filename workflow to `Krea2_LoRA_AB_Compare`, `panel_graph_outline` and `panel_save_workflow({})` still operated on the previous workflow UUID/path. The documented workflow-instance fence should have failed loudly on a mismatch, but the read succeeded and the save returned the previous workflow UUID. This can silently save/edit the wrong workflow.

Likely root cause: the session workflow target / instance fence remained bound to the prior canvas while the panel emitted a new active-workflow event, and dispatch did not compare the live active routing key against the session target before executing.

### Steps to reproduce
1. Work on workflow A and use panel graph tools.
2. Switch the visible canvas to workflow B.
3. Receive the panel event identifying workflow B.
4. Call `panel_graph_outline`, then `panel_save_workflow({})`.
5. Observe both calls operate on workflow A instead of refusing with an instance mismatch.
6. Workaround: call `panel_set_workflow_target({mode:'current'})`, then re-read `panel_list_workflows` and `panel_graph_outline` before saving.

### Exact result
The panel event named `Krea2_LoRA_AB_Compare`, but save returned the previous long-filename workflow UUID/path. A subsequent explicit same-name Save-As returned a normal 409 conflict because the destination already existed; that 409 is not the bug.

### Fix
Applied locally: no. Upstream-only: yes. The defect is in the long-lived panel/orchestrator workflow routing layer. Before every panel graph read/mutation, compare the live active workflow instance/routing key to the session fence. If they differ, fail closed with `workflow instance mismatch` and require `panel_set_workflow_target(mode:'current')`; do not route to the stale target. Add a regression test covering tab switch immediately followed by outline/save.

### Environment
Windows 11; ComfyUI 0.33.1; frontend 1.48.7; Python 3.12.10; PyTorch 2.13.0+cu130; RTX 5090 32 GB; comfyui-mcp 0.51.57; panel 0.13.7.

Related (same root cause): #1646

artokun:
[...]

PR #1775, opened 2026-08-19
fix(1656): a CARRIED workflow stamp is not the routed tab's advertisement — the agreement gate stops being satisfied by the carry

artokun:
The underlying cause of #1656 (recurrence on 0.52.3, which already carries #1682's
dispatch gate): a stamp the routed tab **inherited** was read as a stamp the routed tab
**advertised**, and the two are different facts.

## Why the existing fence was satisfied

#1682 refuses an active-canvas command when "the session's stamp and the routed tab's
LAST-ADVERTISED identity are both known and disagree". The value it reads for the second
half is `tabCommandWorkflowUuid.get(conn.tabId)`.

On the same-socket re-hello that mints a new tab id — a workflow switch, or the
`tmp:` → `wf:<path>` re-key a first save/rename performs — `carryWorkflowCommandStamp`
copies the **retiring** id's uuid onto the new id, and the hello handler overwrites it
only `if (newIdentity)`. So while a hello has not resolved an identity, the "advertised
identity" of the new canvas is the *previous* canvas's uuid.

That is not "sometimes equal". The conversation's issue-time stamp is captured from the
same map at user-message time:

```ts
turnOrigins.recordForMid(userMid, tabCommandWorkflowUuid.get(event.tab_id), event.tab_id);
```

so the carry copies the exact string the session stamp was derived from onto the tab the
command will land on. **The gate's two sides are equal by construction, precisely in the
window where the canvas changed** — it is guaranteed to pass exactly when it is needed,
and starts refusing only once some later hello finally proves a different identity. That
is the reported shape verbatim: six mutations accepted onto the previous canvas, the
seventh refused.

`carryWorkflowCommandStamp`'s own docstring promises "CARRYING CANNOT WIDEN
AUTHORIZATION". That was true of the one authorizer that existed when it was written —
the panel's `stamp === live active uuid` test, which judges the carried value against the
canvas actually mounted. #1682 added a second authorizer that reads the same value as an
*observation about the new canvas*, and there the promise stops holding. A later change
invalidated an earlier one's stated premise; this PR restores it.

Taxonomy: class 1 (CARRIED and ADVERTISED collapse into one indistinguishable value) on
top of class 2 (identity survives the re-key that should have invalidated it).

## Mechanism

* `carryWorkflowCommandStamp` now **returns whether it carried**. The carry itself is
  unchanged — #1331 still needs it, so the frame keeps its stamp and the panel's own
  fence keeps judging it.
* The orchestrator keeps that provenance per tab id (`tabStampCarried`) and clears it the
  moment a hello or a validated refresh proves an identity **for the current id**.
* `UiBridge.send()`'s agreement gate treats a carried stamp as an *absence of evidence*
  about the canvas it is about to read or mutate, and refuses **before the socket write**
  (`workflow instance mismatch: … has not re-established its workflow identity …`, typed
  `dispatched:false`), naming the documented rebind. `#570`'s rule for an unprovable
  target: a delivered frame cannot be retracted, so refuse the dispatch.

### The self-healing half, so the fix is not worse than the bug

The same window is entered by a rename/save whose hello merely **raced** the canvas
identity — where the canvas *is* the workflow the caller named (the panel's own note:
a first save and a Save-As "re-key the route while the instance — and therefore the uuid
— is the same object it always was"). Left alone the session would refuse edits to a
canvas the panel agrees is correct; #1330 measured that cost at 14 refusals.

So the read-only mismatch probe carries the repair. When the corroborated `workflow_list`
says the live canvas is the instance the fence already names, that is an observation of
the tab under its **current** id, and the provenance is promoted from inherited to
observed.

It goes through a new `bridge.setTabStampCorroborator`, **not** `refreshWorkflowUuid` —
and the difference is the whole safety argument. `refreshWorkflowUuid` writes a stamp
(and a conversation's issue-time stamp with it): that is the retarget #1646 removed from
this probe, and `panel-tools.test.ts`'s "nothing to change, so nothing is claimed to have
changed" invariant caught my first cut doing exactly that. The corroborator writes **no
stamp at all**: the orchestrator re-validates the uuid (server-observed origin + shape)
and **refuses unless it already equals the tab's current stamp**, so the only thing it
can change is a provenance bit. No input can retarget a session through it. The
`diverged` case reaches neither call.

## What is still allowed

* Everything, whenever the tab's identity is **proven** under its current id — the normal
  case. The instant a hello or a validated refresh lands, provenance flips and behaviour
  is byte-identical to before this PR.
* Tabs that never migrated: no carry, no flag, no change.
* Inside the gate the exemptions are #1682's, untouched: `workflow_list` (the recovery
  probe — so `panel_set_workflow_target({mode:"current"})` is reachable and is not gated
  on the state it repairs, panel #932's circularity), `workflow_open` / `workflow_new`,
  the #602 canvas-independent set, and path-selectored `workflow_rename` / `workflow_close`.
* Callers addressing the connection by its canonical id: the gate's outer guard
  (`opts.tabId !== conn.tabId`) is unchanged, so the loopback-MCP-by-tab-id path is
  untouched.
* The refusal keeps the `issued for workflow instance <uuid>` shape panel-tools' fence
  diagnosis parses, so it classifies as `stamped` (this session HAS an identity) and not
  as the opposite-remedy `unstamped` case.

## Refutation — 9 mutations, every one killed, each diff confirmed before running

| # | mutation | result |
|---|---|---|
| M1 | disable the carried refusal branch in the gate | 2 fail — incl. the frames reaching the panel |
| M2 | delete `bridge.setCarriedTabStampPredicate(...)` | 1 fail (**source-only** — the behavioural suite stayed green, which is the point) |
| M3 | drop `tabStampCarried.add(panelTab)` in the hello handler | 1 fail (source-only) |
| M4 | make `carryWorkflowCommandStamp` always return `false` | 2 fail |
| M5 | stop clearing the flag when a hello proves an identity | 1 fail (source-only) |
| M6 | delete the corroboration call from `already_current` | 1 fail |
| M7 | delete `bridge.setTabStampCorroborator(...)` | 1 fail (source-only) |
| M8 | delete the corroborator's equality guard | 1 fail (source-only) |
| M9 | swap the corroborator back to `refreshWorkflowUuid` | 3 fail, incl. the pre-existing #1339 invariant |

M2/M3/M5/M7/M8 are the ones that matter for the "helper-only tests" trap: they are
one-line wirings that no behavioural test can see (an uninstalled predicate is *defined*
to mean "proven", i.e. the pre-#1656 behaviour with the bug intact), so
`carried-stamp-wiring.test.ts` pins them at source level, the way
`shared-session-invariant.test.ts` and `agent-identity-wiring.test.ts` already do for this
file. After each mutation `git diff` was read to confirm it applied (CRLF repo — every
edit went through an explicit newline-normalising patcher, never `$`-anchored perl).

## Test evidence

* `src/__tests__/services/carried-stamp-agreement.test.ts` — 8 tests driving the **real**
  `UiBridge` over **real** sockets with the **real** `carryWorkflowCommandStamp`, resolver
  wired the way `orchestrator/index.ts` wires it. Includes a **BASELINE test that pins the
  bug**: with the predicate absent (= pre-#1656) the identical sequence dispatches all
  three commands, stamped with the retired tab's uuid.
* `src/__tests__/orchestrator/carried-stamp-corroboration.test.ts` — the self-healing half
  and its limit: corroborated on a matching live uuid, and *nothing* offered to either
  call when the live canvas is a different workflow.
* `src/__tests__/orchestrator/carried-stamp-wiring.test.ts` — the six call sites.
* Full suite on this branch: **10 421 passed**. Three files failed in the parallel run and
  all three passed alone (a different three on an earlier full run) — machine-load flakes,
  not a delta. `tsc --noEmit`, `check:vocabulary` (1680 files) and `check:unknown-collapse`
  all clean.

## Live reproduction / browser gate

**Not reproduced against a live browser.** The window needs a same-socket re-hello whose
hello has not yet resolved a canvas identity, which is a race I could not force
deterministically from outside. What is driven for real is the whole server half — real
bridge, real sockets, real carry helper — plus the baseline test that shows the pre-fix
gate dispatching on that exact input.

**No panel file changes**, and no change to the wire frame's shape — only whether a frame
is written at all. The panel's Playwright suite therefore cannot observe this change, and
I did not run it; saying so explicitly rather than implying coverage I do not have.

## Scope — what I deliberately did NOT do

* Did not touch the carry (#1331) or the panel's fence. The stamp still ships; only the
  orchestrator's "these agree, therefore dispatch" inference is withdrawn.
* Did not add per-tab or per-workflow session state — sessions stay orchestrator-scoped.
* Did not chase the panel-side half. The recurrence was filed on panel **0.14.44** with
  0.15.4 current, and several later panel fixes bear on it (0.15.1 "a stale-canvas tab can
  no longer persist its graph over another file", #1268 widget-write verification, #1272
  connect-throw verification, #1267b Save-As capture, and open draft #1391 for #1389's
  stale-ROUTE half, which is the other way a command can land somewhere the caller did not
  name). Those are separate work; this PR fixes the orchestrator-side gate only, and I am
  not claiming it is the whole of the reporter's session.

## Gate

Codex was unavailable (account exhausted until 2026-08-19 20:43) and `claude-gate.mjs` is
blocked on the weekly limit, so **this PR carries no independent adversarial gate** — the
owner gates and merges it. I did not self-gate; a self-review reads its own guard as
obviously right, which is the defect class this change is an instance of.

### DIFF
--- a/src/__tests__/orchestrator/carried-stamp-corroboration.test.ts
+++ b/src/__tests__/orchestrator/carried-stamp-corroboration.test.ts
@@ -1,0 +1,121 @@
+// #1656 — the OTHER half: a carried stamp that the live canvas CONFIRMS must stop
+// being treated as inherited, or the fix is worse than the bug.
+//
+// The dispatch gate refuses an active-canvas command while the routed tab's stamp is
+// one CARRIED onto a new tab id by a same-socket re-hello that could not resolve an
+// identity. That window is entered by a rename/save whose hello merely RACED the
+// canvas identity — where the canvas is in fact the very workflow the caller named.
+// Left alone, the session would keep refusing edits to a canvas the panel agrees is
+// correct, and the reporter of #1330 has already shown what that costs (14 refusals).
+//
+// So the read-only mismatch probe carries the repair. When the corroborated
+// `workflow_list` says the LIVE canvas is the same instance the fence already names,
+// that is an observation of the tab under its CURRENT id: the value is confirmed and
+// the provenance is promoted from inherited to observed, through the orchestrator's
+// validator (reachability + origin-bound uuid shape), never off the refusal's prose.
+//
+// It is NOT the retarget #1646 removed. The uuid written is the one this branch has
+// just established EQUALS the fence in force, so nothing is re-aimed — the DIVERGED
+// case (live canvas is a different workflow) still adopts nothing at all.
+
+import { describe, expect, it } from "vitest";
+
+import {
+  buildPanelToolDefs,
+  makePanelToolCtx,
+  type PanelToolCtx,
+  type ToolResult,
+} from "../../orchestrator/panel-tools.js";
+import { markDispatched } from "../../services/ui-bridge.js";
+import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+
+const TAB = "wf:route1:workflows/krea2_lora_ab_compare.json";
+const CARRIED_UUID = "aaaaaaaa-aaaa-4aaa-8aaa-aaaaaaaaaaaa";
+const OTHER_UUID = "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb";
+
+/** The bridge's #1656 refusal, verbatim in shape: pre-dispatch, typed dispatched:false,
+ *  and carrying the stamp in the form panel-tools reads back. */
+const carriedRefusal = (): Error =>
+  markDispatched(
+    new Error(
+      `workflow instance mismatch: this command was issued for workflow instance ${CARRIED_UUID}, ` +
+        `and the tab it routed to has not re-established its workflow identity since it ` +
+        `re-registered under a new id — the uuid it currently carries (${CARRIED_UUID}) was ` +
+        `inherited from the tab id it replaced, so it is not evidence about the canvas now ` +
+        `mounted there. Nothing was dispatched. Re-target with ` +
+        `panel_set_workflow_target({mode:"current"}), then retry.`,
+    ),
+    false,
+  );
+
+function harness(liveUuid: string) {
+  /** Every uuid handed to the orchestrator's validator by this call. */
+  const refreshed: string[] = [];
+  const b = {
+    send: async (cmd: Record<string, unknown>) => {
+      if (cmd.cmd === "workflow_list") {
+        const active = {
+          path: "workflows/krea2_lora_ab_compare.json",
+          routing_key: TAB,
+          workflow_uuid: liveUuid,
+        };
+        return { active, workflows: [{ ...active, active: true }], active_confirmed: true };
+      }
+      throw carriedRefusal();
+    },
+    push: () => 1,
+    canReach: (id: string) => id === TAB,
+    isHeadless: () => false,
+    tabs: () => [{ tab_id: TAB, title: "wf", connected_at: 0 }],
+    resolveActiveTabId: () => TAB,
+    refreshWorkflowUuid: (_tabId: string, uuid: string) => {
+      refreshed.push(uuid);
+      return true;
+    },
+    // The session's fence — the carried uuid the command was issued for.
+    workflowUuidFor: () => ({ known: true, uuid: CARRIED_UUID }),
+    tabCanMutateGraph: () => true,
+    tabGraphMutationCapability: () => ({ known: true, canMutate: true }),
+  } as unknown as PanelToolCtx["bridge"];
+  return { bridge: b, refreshed };
+}
+
+async function connect(liveUuid: string): Promise<{
+  text: string;
+  res: ToolResult;
+  refreshed: string[];
+}> {
+  const { bridge, refreshed } = harness(liveUuid);
+  const ctx = makePanelToolCtx(bridge, TAB, new WorkflowTargetStore());
+  const def = buildPanelToolDefs().find((d) => d.name === "panel_connect");
+  if (!def) throw new Error("panel_connect is not registered");
+  const res: ToolResult = await def.handler(
+    { from_node: 1, from_slot: 0, to_node: 2, to_slot: 0 } as never,
+    ctx,
+  );
+  return { text: res.content.map((c) => (c as { text?: string }).text ?? "").join(" "), res, refreshed };
+}
+
+describe("#1656 — a carried stamp the live canvas CONFIRMS is corroborated, not left inherited", () => {
+  it("promotes the provenance when workflow_list names the SAME instance", async () => {
+    const { text, refreshed } = await connect(CARRIED_UUID);
+    // The corroboration ran, and it wrote back exactly the uuid already in force —
+    // so no target moved; only the evidence behind it did.
+    expect(refreshed).toEqual([CARRIED_UUID]);
+    // The verdict is unchanged: still the transient reading, still one informed retry.
+    expect(text).toMatch(/TRANSIENT/);
+    expect(text).toMatch(/RETRY THIS EXACT CALL ONCE/);
+    // And the refusal itself is preserved — the call still fails.
+    expect(text).toMatch(/workflow instance mismatch/);
+  });
+
+  it("adopts NOTHING when the live canvas is a DIFFERENT workflow (#1646 stands)", async () => {
+    const { text, refreshed } = await connect(OTHER_UUID);
+    expect(refreshed).toEqual([]);
+    expect(text).toMatch(/DIFFERENT workflow/);
+    expect(text).toContain(OTHER_UUID);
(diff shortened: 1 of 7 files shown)


artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**. Verdict: **SHIP**, head `963b8332` (matches the reviewed SHA).

**The root cause holds exactly as stated, traced line by line.** `setTabWorkflowUuidResolver` (`src/orchestrator/index.ts:3085-3090`) answers `tabCommandWorkflowUuid.get(panelTabOf(tabId))`; `turnOrigins`' `uuidOfTab` dep (`:1837`) and the user-message capture at `:5528` read that **same map**; and `carryWorkflowCommandStamp` (`src/orchestrator/session-store.ts:591`) copies `stamps[migratedFrom]` onto `stamps[panelTab]`, with the hello overwriting only `if (newIdentity)` (`:4021`). So `issuedFor === landedOn` **by construction** in the no-identity window, in either event order. #1682's fence was comparing a value against itself, guaranteed to pass precisely where the canvas had changed.

**The window is reachable, not theoretical.** `workflowIdentityParts` needs a bare 8-4-4-4-12 uuid *and* a server origin, and the panel's `workflowStableUuid()` has a documented `return getTabId()` fallback (`comfyui-mcp-panel/web/js/comfyui-mcp-panel.js:3461`) that fails that regex. The pre-existing comment at `index.ts:4013` already asserts this window verbatim from #1331, and every canvas switch re-keys the tab id, so migration *is* the switch surface.

Worth stating because it changes how load-bearing this is: **the panel's own fence is not a backstop in the reported shape.** The reporter's panel accepted the stale-stamped commands, so its live uuid was itself stale. This orchestrator gate is the guard, not a second opinion.

**The self-healing half was verified to be the safe kind.** `setTabStampCorroborator` (`:3155-3170`) writes no stamp: it scopes to the real tab, derives identity from the **server-observed** origin, returns false unless `tabCommandWorkflowUuid.get(panelTab)` already equals it, and its only mutation is `tabStampCarried.delete(panelTab)`. No `tabCommandWorkflowUuid.set`, no `turnOrigins.setStamp`. **#1646's retarget is not reintroduced** — `diverged` returns before either call, and a wiring test pins `not.toContain("refreshWorkflowUuid(")` on that branch. Clearing is complete: both writers of the stamp map (`:3123`, `:4023`) clear provenance, the retiring id is cleared too, and the two maps share one lifecycle so nothing new grows.

**`dispatched:false` is honest** — the refusal at `ui-bridge.ts:4672` precedes `dispatch()` (`:4720`), `sock.send()` (`:4882`) and the mailbox branch, with a behavioural test asserting `received === []` over a real socket plus `dispatchOutcomeOf(err) === false`. That matters: a false "nothing was applied" would license a retry that double-applies.

**Over-refusal was constructed, not assumed.** With a live uuid equal to the carried stamp: `panel_graph_outline` is refused, but `isFencedGraphRead` fires the read-only probe → `already_current` → `corroborateTabStamp` with exactly the carried uuid (and `refreshWorkflowUuid` never called), so the retry dispatches. `panel_set_workflow_target({mode:"current"})` succeeds and corroborates.

**One residual, pre-existing and disclosed:** `panel_save_workflow({})` is refused and **nothing corroborates** — it is neither `graph_*` nor in `MUTATING_GRAPH_EDIT_CMDS`, so neither probe branch fires. A save-only retry loop stays refused until the agent takes the named remedy or issues any `graph_*` call. A control run with #1682's own pre-existing refusal produced the byte-identical wrapper and the same empty corroboration, so both that and the misleading "tab may be disconnected" prose predate this PR. Non-destructive (pre-dispatch). Filed as a follow-up.

**Mutations, byte-level CRLF-safe patcher with `git diff` confirmed after each:** `carried = false` fails 2; dropping the hello's clear fails 1; dropping the equality guard fails 1; swapping the corroborator for `refreshWorkflowUuid` fails 3 including the pre-existing #1339 invariant; and a reviewer-added mutation reading provenance off `opts.tabId` instead of `conn.tabId` fails 2. One reviewer-added mutation **survived** — deleting the `else { tabStampCarried.delete(panelTab) }` branch — but its precondition is destroyed by every prior migration and it errs permissive rather than over-refusing. Coverage nit.

**The baseline test genuinely pins the bug:** with the predicate absent, all three commands dispatch and frames are observed at the panel carrying the *retired* tab's uuid.

`tsc --noEmit` clean; 220 service files / 5,108 tests and 179 orchestrator files / 2,954 tests green with no flakes; `check:vocabulary` (1,684 files) and `check:unknown-collapse` clean; clean merge onto current `origin/main`.

Issue #1401, opened 2026-08-11
panel_set_workflow_target falsely reports persistent fence failure during post-restart reconnect race


artokun:
### What happened / root cause
After panel_restart_comfyui completed, panel_set_workflow_target({mode:"current"}) re-read workflow_list 4 times over ~2.9s and returned a hard failure: active workflow UNCONFIRMED, claiming graph tools would keep failing and instructing a browser refresh. Immediately afterward panel_graph_outline succeeded, and panel_list_workflows reported the same active workflow UUID as the preserved fence with active_confirmed:true.

The safety gate correctly refused to adopt an uncorroborated identity, but the recovery result overstates the consequence when the preserved fence already matches the restored canvas. This produces a false-negative recovery diagnosis and unnecessary F5 guidance during an ordinary post-restart reconciliation race.

### Steps to reproduce
1. Open a saved workflow in the sidebar panel.
2. Install a custom node and call panel_restart_comfyui.
3. When restart returns server_ready:true but panel_tab_reconnected:false, call panel_set_workflow_target({mode:"current"}).
4. Observe the UNCONFIRMED failure after four reads.
5. Immediately call panel_graph_outline and panel_list_workflows; graph access works and the active UUID matches the preserved fence.

### Exact error (scrubbed)
panel_set_workflow_target({mode:"current"}) did NOT restore this session's graph binding.
The panel answered, but its reply could not be corroborated ... active workflow as UNCONFIRMED.
ALREADY TRIED: the panel was re-read 4 times over ~2.9s and never corroborated.
The message states graph tools will keep failing and requests F5, but graph tools succeeded immediately afterward.

### Fix
Applied locally: no. Upstream-only: yes (the running orchestrator is long-lived; patch cannot be verified without restarting it).

Precise change needed:
- In the no_identity/not_recovered result path, do not state graph tools "will keep failing" when a prior fence remains intact.
- If the unconfirmed active record exposes a UUID equal to the preserved fence, report that the fence was not re-derived but is likely still valid, and probe with a fence-stamped read before prescribing F5.
- Alternatively add a short final corroboration read after the current ~2.9s window before classifying the state as non-transient.
- Keep the existing fail-closed rule: never adopt an uncorroborated different UUID.

### Environment
OS: Windows 11
ComfyUI: 0.31.0
ComfyUI frontend: unknown / not reported by environment probe
GPU: NVIDIA GeForce RTX 5090, 32607 MB VRAM
Python: 3.12.10
[...]

PR #1781, opened 2026-08-19
fix(1778): a fence-refused workflow save corroborates like a fenced graph edit

### LINKED ISSUE #1778: panel_save_workflow is fence-refused without corroborating, so a save-only retry loop never self-heals
Split out of #1656, where an independent gate constructed this case and a control run proved it pre-dates that fix.

**The gap.** When a command is refused by the workflow-instance fence, most callers have a self-healing path: `isFencedGraphRead` fires a read-only probe, the probe corroborates the instance the fence already holds, and the retry dispatches. `panel_set_workflow_target({mode:"current"})` likewise succeeds and corroborates.

**`panel_save_workflow({})` has neither.** It is not a `graph_*` command and it is not in `MUTATING_GRAPH_EDIT_CMDS`, so neither probe branch fires, and the save handler returns early at `if (res.isError) return res`. Nothing corroborates, so a caller that only ever retries the save stays refused indefinitely — until it happens to issue any `graph_*` call, or takes the remedy named in the refusal text.

**Reproduced** with a live uuid equal to the carried stamp: `panel_graph_outline` is refused, probes, corroborates and then dispatches on retry; `panel_save_workflow({})` is refused with an empty corroboration. A CONTROL run using #1682's own pre-existing refusal produced the byte-identical generic wrapper and the same empty corroboration, which is what establishes this as pre-existing rather than introduced by #1775.

**Not destructive.** The refusal is pre-dispatch and typed `dispatched:false`, so nothing was written and a retry cannot double-apply. The cost is a stuck loop and a wasted turn, not data.

**Second, smaller thing on the same path:** the generic wrapper's "tab may be disconnected" prose is misleading in this state — the tab is connected and routing fine; what is stale is the stamp. Also pre-existing.

Fix is probably to give the save path the same read-only corroboration the graph reads get, or to widen the probe branch to cover commands that carry a workflow target without being graph mutations. Worth checking whether other non-`graph_*` targeted commands share the gap before picking.

artokun:
Fixes #1778.

## The gap

The panel's `activeWorkflowFenceApplies` does **not** key on the `graph_` prefix. It fences everything that is not canvas-*targetless*, `workflow_open` or `workflow_new` — which, alongside every `graph_*`, covers the four ACTIVE-workflow mutators the panel's own comment names explicitly: `workflow_save`, `workflow_save_as`, `workflow_rename`, `workflow_close` ("workflow_save\* ignore any path, so they are ALWAYS active").

Orchestrator-side, both halves of the repair read their scope off the prefix instead:

- `isMutatingGraphCmd` is an allowlist holding only `graph_*` names (#1330 branch);
- `isFencedGraphRead` requires the prefix outright (#1519 branch).

So a fence refusal on `workflow_save` matched **neither**. It fell past those, past the root-uuid / capability / routing-ambiguity branches, and landed in the generic `dispatchOutcomeOf(err) === false` wrapper — which fires no probe at all. That means `rebindWorkflowFence`'s `corroborateTabStamp` (#1656 — the promotion that lets a CARRIED-but-confirmed stamp start dispatching again) **never ran**.

In the reporter's session, on one stamp: `panel_graph_outline` was refused, probed, corroborated, and its retry went through. `panel_save_workflow({})` was refused with an **empty corroboration**, every time. A caller whose loop only ever retries the save therefore never self-heals — until it happens to issue some unrelated `graph_*` call, or takes the remedy named in the refusal text.

Not destructive: the fence is checked pre-dispatch and the refusal is typed `dispatched:false`, so nothing was written and a retry cannot double-apply. The cost was a stuck loop and a wasted turn, not data.

## The fix

Name the four fenced non-graph mutators explicitly (`isFencedWorkflowMutator`) and admit them to the #1330 corroboration branch. They are mutations refused PRE-dispatch exactly as a graph edit is, so every claim that branch makes holds for them verbatim — "NOT applied — nothing changed", the READ-ONLY probe, and the retry advice.

`#1646 stands`: the probe is still `adopt:false`. A live canvas that names a **different** workflow is reported, never adopted — for a save that matters more than for an edit, since a re-pointed save would write the caller's graph over another workflow's file. `corroborateTabStamp` (provenance only) is reached; `refreshWorkflowUuid` (the retargeting write) is not.

**An explicit list, deliberately not "everything that is not `graph_`"** — see the scope section below for why the wider fenced set is not admitted. A control test pins the boundary: `manager_getlist` (canvas-independent, `CANVAS_INDEPENDENT_COMMANDS`, #602) quoting the same phrase still falls through to generic handling, probing nothing.

`workflow_rename`/`workflow_close` are listed even though the panel exempts them when their selector resolves to a **deterministically** non-active workflow — in that case no refusal is raised at all, so the predicate is never consulted. It is *not* claimed that a refused rename/close was therefore targeting the active canvas: a `path` resolving to no open workflow is fenced too, since the exemption requires a positive resolution. Such a call gets one informed retry instead of the generic wrapper's guess; both refuse, so the cost of being wrong there is a round trip, not a wrong target.

### The phrase is not the state (second commit, from the gate)

`isWorkflowInstanceMismatch` is an **unanchored** `/workflow instance mismatch/i`, and two other refusals quote those exact words:

- the bridge's **capability** refusal embeds them in its `readsNote`, to describe what a *different* command would get;
- the **no-trusted-identity** refusal (#1331) carries them the same way.

Both are minted by the same `"<cmd>" cannot be safely targeted to the active workflow` reject and both are `dispatched:false`, so both satisfied the phrase match. For `graph_*` that never surfaced — by ordering rather than by design: the #1401 branch catches `no trusted identity` for a mutating graph edit first. These four have nothing above them, so the widening **alone** would have hijacked both states and answered each with the other's remedy: the retry/rebind suffix #709 exists to suppress, appended three lines after the refusal says rebinding cannot add the missing capability; and #1330's "retry once" handed to a state whose own docstring says "a mismatch may clear by itself, this never does".

So the new arm carries the discriminators the old one got from ordering — `isCapabilityRefusal` (the bridge's typed symbol marker) and `isNoTrustedIdentityRefusal` (the distinct phrase a mismatch refusal never contains). Scoped to the new arm only, so `isMutatingGraphCmd` keeps its exact prior behaviour.

## Answering "check whether other non-`graph_*` targeted commands share the gap"

They do — but the fenced set outside the prefix is **wider than four**, and the round-2 gate was right to push on my first wording. `CANVAS_INDEPENDENT_COMMANDS` holds only seven names, so `activeWorkflowFenceApplies` also fences `ask_user`, `ui_render`, `training_*`, `civitai_*` and others.

Those are deliberately **not** admitted. This branch's claims are about a MUTATION of the active canvas — "NOT applied — nothing changed", a read-only re-read of that canvas, a retry that cannot double-apply — and the four are exactly the set for which all three are true. They mirror ui-bridge's own `ACTIVE_WORKFLOW_MUTATORS` (#570 P0c), which is the list to keep the new predicate in sync with. A negation rule would have been wrong in both directions.

**Deliberately not changed, two things:**

1. The #1401 `no trusted identity` branch is gated on `isMutatingGraphCmd` too, so it has the same *shape* of gap for these four. Left alone: different refusal, its probe **adopts** rather than being read-only, and it was neither reported nor reproduced. Widening an adopting rebind onto the save path off an unreproduced hypothesis is not a trade worth making under the freeze. (This PR makes those four fall through to the generic wrapper for that state, exactly as before — it does not make it worse.)
2. `graph_*` mutations *already* hijack the capability refusal through the same unanchored phrase match — pre-existing, untouched here, and worth a separate look. Anchoring `isWorkflowInstanceMismatch` would fix it globally but changes the graph path, which is out of scope for a P3.

## Verification

**Verified by deleting the fix, twice.**

Reverting the widened gate to bare `isMutatingGraphCmd(cmd)` fails 3 of the tests, and the failure output is the reported symptom byte for byte:

```
expected 'Error: workflow_save could not be dis…' to match /DIFFERENT workflow/
+ Received:
"Error: workflow_save could not be dispatched to this session's panel tab — nothing was
 applied. The tab may be disconnected, still reconnecting after a restart/reload, or the
 session's binding is stale (e.g. another workflow tab is now active). …"
```

Reverting the two discriminators fails exactly the two tests that pin them. The CONTROL test (canvas-independent command) correctly survives both mutations.

The tests drive the **real** `panel_save_workflow` / `panel_rename_workflow` tool defs through the real `ctx.call` path, with a bridge that mints each refusal via the bridge's own `markDispatched` / `markCapabilityRefusal` — so they exercise the production call site, not a helper. Each look-alike fixture first asserts that it really does trip `/workflow instance mismatch/i`, so neither test can pass vacuously.

- `npx vitest run src/__tests__/orchestrator/fenced-save-corroborates.test.ts` — 6 passed
- `npx vitest run src/__tests__/orchestrator src/__tests__/services` — **402 files / 8089 tests passed**, 3 skipped
- `npm run lint` (`tsc --noEmit`) — clean

**Browser coverage:** none was run, and none applies. This is orchestrator-side TypeScript only; no panel source is touched and nothing the sidebar renders changes. What changes is the probe behaviour and text of an MCP tool result.

## Gate

codex was unavailable (account exhausted until 2026-08-19); gated by an independent adversarial review instead, via `.claude/autopilot/claude-gate.mjs`. Disclosed as required.

- **Round 1 — NO-SHIP.** Two P1s, both real and both the same class: the unanchored phrase match cannot distinguish the fence refusing from another refusal *quoting* the fence. Fixed in `ccb60b7`, with a test per finding.
- **Round 2 — SHIP.** Re-gated after the fix. The reviewer enumerated every producer of `/workflow instance mismatch/i` that can reach this catch, forced the probe to fail (`unreadable`, no corroboration), fenced `workflow_list` itself to check the branch cannot recurse (one probe, no loop), and independently killed all three mutations. It raised two comment-only prose inaccuracies, both correct and both fixed in the third commit — the fenced set outside the prefix is wider than four, and a refused rename/close does not prove an active target. Neither changed behaviour.

## Note for the maintainer

A parallel worker claimed this issue in `.worktrees/wt-1778` (branch `fix/1778-save-fence-corroborate`) and began writing a test there mid-run. I relocated to my own worktree and left theirs exactly as I found it. If a competing PR appears, it and this one are solving the same defect — close whichever is second.

### DIFF
--- a/src/__tests__/orchestrator/fenced-save-corroborates.test.ts
+++ b/src/__tests__/orchestrator/fenced-save-corroborates.test.ts
@@ -37,7 +37,11 @@ import {
   type PanelToolCtx,
   type ToolResult,
 } from "../../orchestrator/panel-tools.js";
-import { markDispatched } from "../../services/ui-bridge.js";
+import {
+  isCapabilityRefusal,
+  markCapabilityRefusal,
+  markDispatched,
+} from "../../services/ui-bridge.js";
 import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
 
 const TAB = "wf:route1:workflows/krea2_lora_ab_compare.json";
@@ -59,7 +63,60 @@ const fenceRefusal = (): Error =>
     false,
   );
 
-function harness(liveUuid: string) {
+// ── The two LOOK-ALIKES, quoted from the bridge's own templates ───────────────
+//
+// Both are minted by the SAME `"<cmd>" cannot be safely targeted to the active
+// workflow` reject in ui-bridge, both are `dispatched:false`, and both carry the
+// literal words "workflow instance mismatch" inside the `readsNote` that describes
+// what a DIFFERENT command would get. `isWorkflowInstanceMismatch` is an unanchored
+// phrase match, so it fires on both — which is why the new arm needs discriminators
+// the `graph_*` arm gets for free from the #1401 branch sitting above it.
+
+/** `capabilityMissing` — the panel does not enforce the fence contract a write needs.
+ *  Typed with the bridge's own symbol marker. NEITHER a retry NOR a rebind can add
+ *  the missing capability (#709), so appending that suffix contradicts the refusal. */
+const capabilityRefusal = (): Error =>
+  markCapabilityRefusal(
+    markDispatched(
+      new Error(
+        `"workflow_save" cannot be safely targeted to the active workflow: panel tab ` +
+          `${TAB} does not recheck workflow targeting at the graph write boundary after ` +
+          `asynchronous work (detected panel 0.11.40; a graph WRITE needs panel 0.11.62+, ` +
+          `the first build that rechecks the fence after an await). Update the panel and ` +
+          `hard-refresh the browser tab. WHETHER GRAPH READS STILL WORK IS NOT KNOWN FROM ` +
+          `HERE, and is not claimed: a read carries this session's stamp (${CARRIED_UUID}), ` +
+          `and this tab's panel runs it only while that stamp still names the ACTIVE canvas ` +
+          `— a comparison only the panel can make. If the workflow was switched or replaced ` +
+          `after this session bound to it, graph_outline / graph_query are refused with ` +
+          `"workflow instance mismatch" as well; if it was not, they work. Try ` +
+          `panel_list_workflows — the panel exempts that read from this fence (it is the ` +
+          `recovery probe), though a build predating the exemption fences it too. Non-graph ` +
+          `tools are unaffected.`,
+      ),
+      false,
+    ),
+  );
+
+/** The #1331 state: the fence contract IS advertised, but the workflow has no identity
+ *  to fence against. NOT capability-marked — the panel is fine, the canvas is not — so
+ *  the discriminating phrase is the only thing that separates it. And the separation
+ *  matters: "a mismatch may clear by itself, this never does". */
+const noTrustedIdentityRefusal = (): Error =>
+  markDispatched(
+    new Error(
+      `"workflow_save" cannot be safely targeted to the active workflow: this workflow has ` +
+        `no trusted identity for the panel to fence the command against. GRAPH READS ARE ` +
+        `REFUSED TOO, for this same missing stamp: this tab's panel enforces the ` +
+        `per-command fence and refuses an UNSTAMPED command rather than fail open, so ` +
+        `graph_outline / graph_query answer "workflow instance mismatch: this command ` +
+        `carries no workflow-instance stamp" as well. Try panel_list_workflows — the panel ` +
+        `exempts that read from this fence (it is the recovery probe), though a build ` +
+        `predating the exemption fences it too. Non-graph tools are unaffected.`,
+    ),
+    false,
+  );
+
+function harness(liveUuid: string, refusal: () => Error = fenceRefusal) {
   /** Every uuid offered to the provenance-only corroborator — the self-heal. */
   const corroborated: string[] = [];
   /** Every uuid handed to the RETARGETING fence write — must stay empty (#1646). */
@@ -77,7 +134,7 @@ function harness(liveUuid: string) {
         };
         return { active, workflows: [{ ...active, active: true }], active_confirmed: true };
       }
-      throw fenceRefusal();
+      throw refusal();
     },
     push: () => 1,
     canReach: (id: string) => id === TAB,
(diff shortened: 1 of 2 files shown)


artokun:
Closing as duplicate of #1782 — same root cause and an equivalent predicate, but #1782 derives the mutator set from requiresWorkflowStampEnforcement (no list duplication) and pins the full retry self-heal end-to-end, including the diverged-canvas control. Your over-fire CONTROL test (canvas-independent command quoting the refusal phrase gets no probe) is worth porting over — noted on #1782. Thanks for the fast turnaround.

PR #1784, opened 2026-08-19
fix(1778): the widened fence arm must not adopt refusals that merely quote it

### LINKED ISSUE #1778: panel_save_workflow is fence-refused without corroborating, so a save-only retry loop never self-heals

artokun:
Follow-up to #1782 (which closed #1778). Its widening was right; this closes the hole it opened.

## What #1782 did, and the hole

#1782 widened the #1330 corroboration branch to the four active-workflow mutators (`workflow_save` / `workflow_save_as` / `workflow_rename` / `workflow_close`). That was correct — without it a save-only retry loop never self-healed, which is exactly what #1778 reported.

But `isWorkflowInstanceMismatch` is an **unanchored** `/workflow instance mismatch/i`, and the fence is not the only thing that says those words. The bridge's `"<cmd>" cannot be safely targeted to the active workflow` reject:

- quotes the phrase inside its `readsNote`, to describe what a **different** command would get;
- carries the second look-alike, `no trusted identity`, in its `why` clause.

Both are minted pre-dispatch and flagged `dispatched:false`, so both satisfy the phrase match on the way through.

For a mutating graph edit this never surfaced — by **ordering** rather than by design: the #1401 branch catches `no trusted identity` for `isMutatingGraphCmd` above. The four mutators have no branch above them, so each of these two states now gets the *other's* remedy.

## Measured, not reasoned

Driving the real `panel_save_workflow` handler against a real `markCapabilityRefusal(markDispatched(…, false))` on this branch's parent (`2c6f6a1`, current `main`):

```
Error: workflow_save was NOT applied — nothing changed. "workflow_save" cannot be safely
targeted to the active workflow: panel tab … does not recheck workflow targeting at the graph
write boundary after asynchronous work. Update the panel and hard-refresh the browser tab. …

CHECKED: the session's fence already names the live canvas (aaaaaaaa-…), so it was not the
stale side. Retry once — if it refuses again with the same pair, the two identities are
genuinely disagreeing and panel_open_workflow is the way to settle which one you mean.
```

"Update the panel and hard-refresh the browser tab", then "Retry once … panel_open_workflow is the way to settle which one you mean." That appended half is the futile retry/rebind suffix **#709 exists to suppress**, contradicting the sentence directly above it, and it costs a wasted `workflow_list` round trip on the way.

This one is a **regression**: before #1782 a capability refusal on these four reached `isCapabilityRefusal` and surfaced verbatim.

The second state is not a regression but is worse than what it replaced: a fence-less session (#1331) now gets #1330's confident "the fence already names the live canvas — retry once" for a state whose own docstring says *"a mismatch may clear by itself, this never does"*. Previously it hit the generic wrapper, which at least claimed nothing.

## The fix

Give the widened arm the discriminators the graph arm gets for free from ordering:

```ts
(isMutatingGraphCmd(cmd) ||
  (isFencedActiveWorkflowMutator(cmd) &&
    !isCapabilityRefusal(err) &&
    !isNoTrustedIdentityRefusal(err)))
```

Both are typed-or-explicit, never a second reading of the same prose: `isCapabilityRefusal` is the bridge's own symbol marker (`markCapabilityRefusal`), and `isNoTrustedIdentityRefusal` matches a distinct phrase a mismatch refusal never contains.

**Scoped to the widened arm only.** `isMutatingGraphCmd` keeps its exact prior behaviour, so nothing about the graph path moves. (`graph_*` mutations *do* also hijack the capability refusal through the same unanchored match — pre-existing, deliberately untouched here. Anchoring `isWorkflowInstanceMismatch` outright would fix it globally but changes the graph path, which is a bigger blast radius than this follow-up should carry.)

## Verification

**The tests fail on this branch's parent** — that is the deletion test, run against real `main` rather than a hand-reverted diff:

```
× a CAPABILITY refusal keeps its own words and pays no probe
× a NO-TRUSTED-IDENTITY refusal is not handed the mismatch remedy
AssertionError: expected [ 'workflow_save', 'workflow_list' ] to not include 'workflow_list'
```

Each fixture **first asserts that it really does trip** `/workflow instance mismatch/i` (and, for the capability case, that `isCapabilityRefusal` is true / for the other, that it is false), so neither test can pass vacuously by never reaching the branch.

**#1782's self-heal is preserved.** Its five tests in `save-fence-corroborate.test.ts` are untouched and still green — including the one that proves a save-only retry dispatches after corroboration.

- `npx vitest run …/fence-lookalike-refusals.test.ts …/save-fence-corroborate.test.ts` — 7 passed
- `npx vitest run src/__tests__/orchestrator src/__tests__/services` — **404 files / 8095 tests passed**, 3 skipped
- `npm run lint` (`tsc --noEmit`) — clean

**Browser coverage:** none run, none applicable. Orchestrator-side TypeScript only; no panel source touched and nothing the sidebar renders changes.

## Gate

codex was unavailable (account exhausted until 2026-08-19); gated by an independent adversarial review instead, via `.claude/autopilot/claude-gate.mjs`. Disclosed as required. Verdict recorded in a comment below.

## Provenance

Both findings came from an independent adversarial gate run against a parallel implementation of the same #1778 fix (PR #1781, closed in favour of #1782). The gate returned NO-SHIP on exactly these two P1s; #1782 landed the same widening without them, so they are live on `main` and this PR carries the fix and the tests across.


artokun:
**Gate: SHIP** — codex was unavailable (account exhausted until 2026-08-19); gated by an independent adversarial review via `.claude/autopilot/claude-gate.mjs` instead. Disclosing the substitution as required.

The reviewer enumerated every producer of `/workflow instance mismatch/i` that can reach this catch for `workflow_save` — empirically, on a real `UiBridge` + fake panel over a live socket rather than by reading:

| state | phrase | `isCapabilityRefusal` | `no trusted identity` | lands |
|---|---|---|---|---|
| panel-minted fence mismatch (stale stamp) | ✓ | ✗ | ✗ | new arm ✓ |
| no-trusted-identity (#1331) | ✓ | ✗ | ✓ | excluded ✓ |
| capability, no at-write recheck, stamp present | ✓ | ✓ | ✗ | excluded ✓ |
| capability, no fence at all | ✗ | ✓ | ✗ | never matched |
| capability + no stamp (both true) | ✓ | ✓ | ✗ | excluded ✓ |

`markCapabilityRefusal` has exactly one call site and `no trusted identity` exactly one producer, so the discriminators are complete and neither is over-broad. The #1656 gate refusals carry only `markDispatched`, so the state #1778 was filed for still enters the arm.

It also killed all three mutations independently — dropping both discriminators fails 2 tests; dropping only `!isCapabilityRefusal` fails exactly the capability test; dropping only `!isNoTrustedIdentityRefusal` fails exactly the other — and confirmed the typed symbol survives the real `sendRouted` → `ctx.call` path rather than only a helper.

One non-blocking note (a committed `.pr-body.md` scratch file) is fixed in `c7edcd5`.

Issue #1789, opened 2026-08-19
panel_run completion notification never delivered — agent idles forever despite the documented "end your turn and wait" contract


artokun:
### What happened / root cause

`panel_run` queued successfully and returned the usual rider:

> [IMPORTANT] You will be notified automatically with the output image(s)/video when the render finishes — do NOT poll queue (action:"list"), get_history, or get_image (action:"list_outputs"). Just end your turn now and wait for the result to be delivered to you.

I followed that contract and ended my turn. **No task notification was ever delivered.** The render had in fact completed successfully 28.77 s later. The agent sat idle until the user broke the silence with "what's happening, is this process stalled?" — i.e. the user had to rescue the session.

Recovered by calling `queue(action:"status", prompt_id=...)`, which immediately returned `{running:false, pending:false, done:true, status_str:"success", execution_stats:{total_duration_ms:28767}}`. So the run, the outputs and the history record were all fine — only the notification delivery to the agent session failed.

Root cause not established from the agent side: ComfyUI logged a clean `Prompt executed in 28.77 seconds` with no error, and the output file was written correctly. The failure is in the orchestrator's completion-notification path (the thing that is supposed to re-invoke the agent with the finished render), not in ComfyUI or the graph.

One possibly-relevant detail: this was the **first** `panel_run` of the session, and the executing node had to **download a 1.3 GB model mid-execution** (`comfyui_controlnet_aux` fetching `depth_anything_v2_vitl.pth` from HuggingFace, ~23 s of the 28.77 s total). During that download the node emitted a benign but unusual non-fatal line — `[Errno 2] No such file or directory: '/var/folders/hb/<redacted>/T/ckpts'` — and there were no progress/step events for most of the run. If the notification path keys off progress events or a websocket that idles during a long non-progress phase, a ~23 s eventless stretch is a plausible trigger worth checking.

### Steps to reproduce

1. Fresh panel session, empty canvas, local ComfyUI.
2. Build `LoadImage → DepthAnythingV2Preprocessor (ckpt vitl, resolution 1024) → SaveImage (filename_prefix "stereo360/depth_padded")`, with the vitl weights NOT yet present so the node downloads them during execution.
3. `panel_run` with no `to_node_id`.
4. End the turn as the tool result instructs.
5. Observe: the render completes and writes `output/stereo360/depth_padded_00001_.png`, but no notification reaches the agent. `queue(action:"status")` shows `done:true, success`.

### Exact error (scrubbed)

[...]

Issue #1739, opened 2026-08-19
Surface successful panel_run outputs after workflow tab switch


artokun:
### What happened / root cause
A panel_run for a read-only GIF keyframe workflow completed successfully in ComfyUI, but after the user switched workflow tabs the active canvas still showed an empty PreviewImage widget and no completion notification appeared in the chat. The execution history contained eight valid temp PNG outputs.

### Steps to reproduce
1. Run a VHS_LoadVideo -> PreviewImage workflow with a GIF and frame_load_cap/select_every_nth.
2. Let the run complete.
3. Switch workflow tabs or continue in the panel.
4. The active workflow may show an empty PreviewImage and the completion event is not surfaced, although ComfyUI history has successful outputs.

### Exact error (scrubbed)
No execution error. panel_get_errors returned errored_count=0 and last_execution_error=null. get_history for the prompt reported Status: success, Completed: true, Duration: 0.81s, with eight temp PNG outputs. The issue is missing/stale UI result synchronization, not a ComfyUI node failure.

### Fix
Applied locally: no source patch. Recovered by reading ComfyUI history, locating the eight temp PNGs, and displaying them explicitly with panel_show_media.
Upstream-only: yes; completed outputs should be delivered and PreviewImage state refreshed reliably across workflow-tab switches.

### Environment
OS: darwin
ComfyUI: 0.33.1
GPU: MPS, 16 GB VRAM
comfyui-mcp: 0.52.2
comfyui-mcp-panel: 0.15.1
Manager: v4

No credentials or user secrets included.

Related (same root cause): #468

artokun:
Claimed by kimi-code swarm (worktree fix-1739). Agent assigned now; if no PR appears within ~8h the claim has lapsed.

PR #1791, opened 2026-08-19
fix(1789): the orchestrator's own /history observation keeps the completion promise panel_run makes — and the rider stops over-promising

artokun:
Addresses the underlying cause of #1789 — `panel_run` told the agent "you WILL be notified … end your turn now and wait", the notification never came, and the session idled through a clean 28.77 s render until a human broke the silence.

## Where the delivery is lost

The rider's promise is gated on `RunCompletions.openRun(...)` returning true. That proves the orchestrator could **recognise** a completion if one arrived — it never proved one **would**. The sole producer is the panel's `agent_event {kind:"executed"}` frame, and everything downstream of it is already durable (`run-completion-journal.ts`: correlated once, journaled, replayed until the turn that carried it ends). Nothing guarded what is **upstream** of that frame. If it is never sent — a missed `execution_success` the panel's own reconcile also missed, a send the socket accepted and the bridge dropped, a panel unmounted mid-render — the ticket stays open forever and **nothing in this process can tell that state apart from "still rendering"**. That is the taxonomy's class 1 exactly: two states, one indistinguishable observation.

And the orchestrator had already seen the run finish. `QueueMonitor` keeps its own socket plus a 1 Hz `GET /history` tail diff *specifically* to observe completions the panel-driven path never reports (#258/#259), including browser-queued runs. Those observations went to exactly one consumer — the `queue_status` UI broadcast (`index.ts`, `QueueMonitor.drainCompletions()`) — and were dropped on the floor. **The completion this session was promised was in the orchestrator's hands and thrown away.**

## What changed — both directions

**Delivery is stronger.** New `src/orchestrator/run-completion-watchdog.ts` joins the two:

- `observe()` — a drained `QueueMonitor` completion whose prompt id has an **unanswered** `panel_run` ticket is armed with the time we saw it. First observation wins, so a repeatedly re-reported completion cannot become immortal.
- `tick()` — once the grace elapses, ask **again**. Still unanswered ⇒ the panel is never going to report it: synthesise the completion and hand it to the **same journal** the real frame uses, so it is correlated `matched` ("this is the run YOU queued"), replayed if no agent takes it, and acked normally.
- New journal predicate `RunCompletions.awaitingCompletion(promptId)` is what makes "finished and never announced" nameable at all. Deliberately conservative — no ticket / already settled / **any** journaled entry for the id (pending *or* handed off) / an acked memo / a `reused` id ⇒ not owed.
- Grace is 45 s: longer than the panel's own 20 s reconcile sweep, so the panel's frame — which carries the **images** — wins every time it is coming. If it lands while the synthesised entry is still `pending`, `record()` coalesces the real payload onto it and the agent sees one turn, with pixels.
- #1789 item 3: it now logs a **warning** naming the prompt id. Until now the only detector of a lost completion was a human noticing the agent had gone quiet.

**And the promise is honest.** The rider kept its absolutism on the strength of `correlatable`, which is written by code that cannot keep it — the watchdog sees nothing when the monitored ComfyUI is not the one this run went to, or cannot be polled. So the anti-poll instruction stays **primary** (it is what stops an agent burning a turn a second), paired with a bounded escape hatch: `[FALLBACK] … it cannot be GUARANTEED … make ONE check with queue (action:"status", prompt_id:"…") … One check, not a loop — and not before then.` The prompt id is inlined when the run queued exactly one, and omitted rather than invented for a batch. The uncorrelatable branch is untouched — it never told the agent to idle.

A failed run is reported through the same non-urgent `executed` shape, **not** `run_error`: `run_error` interrupts the live turn and front-queues "diagnose it", which is the right blast radius for a live failure the panel reported and the wrong one for a status read out of a history record a minute later (#1489's own argument).

## Refutation — 7 mutations, all killed

Each edit was applied with an anchor script that **refuses on a non-unique match**, and each was confirmed with `git diff --stat` before the run.

| # | mutation | result |
|---|---|---|
| 1 | delete `runCompletionWatchdog.observe(completions)` from the broadcaster's drain tee | 1 failed (call-site test) |
| 2 | delete `runCompletionWatchdog.tick()` from the poll timer | 1 failed |
| 3 | `awaitingCompletion` stops checking journaled entries | 3 failed |
| 4 | `awaitingCompletion` always returns undefined | 6 failed |
| 5 | drop the `[FALLBACK]` clause from the rider | 2 failed |
| 6 | do not disarm before `deliver()` | 2 failed |
| 7 | last-observation-wins instead of first | 1 failed |

The call sites are asserted at **source** level (same treatment as `carried-stamp-wiring` / `agent-identity-wiring`): the tee, the tick, the tick *ordering*, and that `drainCompletions()` has exactly **one** call site — a second one would splice-race the broadcaster and each consumer would see half a burst. Mutations 1 and 2 are the ones a behavioural test is blind to.

## Was the reporter inside a known window?

- **#1615/#1690 (re-spelled target keeps its self-queue ledger)** — no. That ledger (`selfQueuedIds` / `isSelfAttributedProven`) drives `panel_run`'s duplicate **fence** and the backlog note. The reporter's run was accepted and queued, so the fence never fired, and the notification path reads the **journal's tickets**, not that ledger. Not on this path.
- **#1656/#1775 (carried workflow stamp vs. the routed tab's advertisement)** — no. That governs graph reads and the workflow fence, not completion routing.
- **panel #1739 (a re-pended completion re-arms the reconcile sweep)** — plausibly **yes**, and worth naming: it landed in panel **0.15.4**; the reporter was on **0.14.44**. Its recorded symptom is verbatim this one ("successful run, no completion event, manual get_history recovery"). But #1739 requires `sendFrame` to have failed, which I cannot establish from the report, and it is panel-side. The orchestrator-side gap fixed here is unconditional and covers every way the frame is lost, including ones no panel version can recover from.

## Verification

- `tsc --noEmit` clean.
- Full suite: **10,514 passed / 573 files**. Three files failed under full-parallel load (`panel-restart-health-readiness`, `panel-restart-legacy-fallback`, `panel-session-rebind-residuals`) and **all 89 tests in those three pass when run alone** — the known wall-clock load flake, none of them touching anything this PR changes.
- `npm run check:vocabulary` OK; `npm run check:unknown-collapse` OK.
- **Not live-reproduced.** The defect is in the long-lived orchestrator process, which only picks up code on a full restart; the fallback needs a real ComfyUI *plus* a deliberately suppressed panel frame. Every branch is driven against the real `RunCompletions` singleton instead. No GPU was taken and no render was queued.

## Gate

Codex is exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the weekly limit, so **this PR is ungated** — the owner gates and merges. Nothing here is panel-visible (no `comfyui-mcp-panel` change), so the Playwright browser suite does not apply.

## Deliberately not done

- No `/history` **output fetching** in the synthesised notice. It carries the terminal status and points at `get_image` / `get_history`; claiming images the orchestrator never read would be exactly the over-claim class this repo keeps paying for.
- No change to the panel repo. #1739 already shipped there; a second recovery layer in the browser would not have helped a session whose frame never left.
- The uncorrelatable rider branch, `run_error` routing, and the journal's dedupe/replay rules are untouched.

### DIFF
--- a/src/__tests__/orchestrator/run-completion-fallback-wiring.test.ts
+++ b/src/__tests__/orchestrator/run-completion-fallback-wiring.test.ts
@@ -23,6 +23,7 @@ import { readFileSync } from "node:fs";
 import { buildPanelToolDefs, makePanelToolCtx, type PanelToolCtx } from "../../orchestrator/panel-tools.js";
 import { QueueMonitor } from "../../services/queue-monitor.js";
 import { RunCompletions } from "../../orchestrator/run-completion-journal.js";
+import { registerQueueManagementTools } from "../../tools/queue-management.js";
 
 const indexSrc = (): string =>
   readFileSync(new URL("../../orchestrator/index.ts", import.meta.url), "utf8");
@@ -130,6 +131,52 @@ afterEach(() => {
   qm.url = null;
 });
 
+/**
+ * Drive the REAL `queue` tool with the call the rider just told the agent to
+ * make, and report whether the tool ACCEPTED it.
+ *
+ * This is the assertion the first cut of #1789 was missing, and the gate caught
+ * it: the batch branch printed `queue (action:"status")` with no id, which
+ * queue-management.ts's `requirePromptId` REFUSES — so an agent obeying the
+ * fallback got an error instead of a status, the reported bug's own shape one
+ * level in. A test that only string-matched the rider pinned the broken form.
+ *
+ * The parse is deliberately narrow (exact literal shapes the rider emits) and
+ * THROWS on anything it does not recognise, so a reworded remedy fails loudly
+ * here instead of silently going unchecked.
+ *
+ * `isError` is the only thing read: a reachability failure is a different fact
+ * from a validation refusal, and the local `queue` handler answers both shapes
+ * without a live ComfyUI (verified: `list` and `status`+id return JSON, bare
+ * `status` returns isError "requires `prompt_id`").
+ */
+async function queueCallIsAccepted(call: string): Promise<boolean> {
+  let captured: ((a: Record<string, unknown>) => Promise<{ isError?: boolean }>) | null = null;
+  registerQueueManagementTools({
+    tool: (name: string, _d: unknown, _s: unknown, h: (a: Record<string, unknown>) => Promise<{ isError?: boolean }>) => {
+      if (name === "queue") captured = h;
+    },
+  } as never);
+  if (!captured) throw new Error("the queue tool was not registered");
+  const withId = call.match(/^queue \(action:"([a-z_]+)", prompt_id:"([^"]+)"\)$/);
+  const bare = call.match(/^queue \(action:"([a-z_]+)"\)$/);
+  const args = withId
+    ? { action: withId[1], prompt_id: withId[2] }
+    : bare
+      ? { action: bare[1] }
+      : null;
+  if (!args) throw new Error(`the rider's remedy is in an unrecognised shape: ${call}`);
+  const res = await (captured as (a: Record<string, unknown>) => Promise<{ isError?: boolean }>)(args);
+  return res.isError !== true;
+}
+
+/** The single remedy the rider's [FALLBACK] line tells the agent to run. */
+function fallbackCallIn(text: string): string {
+  const m = text.match(/make ONE check with (queue \(action:"[a-z_]+"(?:, prompt_id:"[^"]+")?\))/);
+  if (!m) throw new Error(`no [FALLBACK] remedy found in the rider:\n${text}`);
+  return m[1];
+}
+
 describe("#1789 — panel_run's rider stops promising more than the code can keep", () => {
   it("keeps the anti-poll instruction AND names one bounded fallback with the prompt id", async () => {
     const res = await defByName("panel_run").handler({}, makeCtx({ queued: true, prompt_id: "p-1789" }));
@@ -155,9 +202,41 @@ describe("#1789 — panel_run's rider stops promising more than the code can kee
     );
     const text = firstText(res);
     expect(text).toContain("[FALLBACK]");
-    // No invented single id when the run queued more than one.
-    expect(text).toContain('queue (action:"status")');
+    // No invented single id when the run queued more than one — and the remedy
+    // is the action that takes none, not a status call with the id left off.
+    expect(text).toContain('queue (action:"list")');
     expect(text).not.toContain('prompt_id:"p-a"');
+    expect(text).not.toContain('queue (action:"status")');
+  });
+
+  // THE POINT OF THE WHOLE FALLBACK: the call it names has to RUN. Both branches,
+  // executed against the real tool rather than string-matched.
+  it("SINGLE run — the remedy the rider prints is ACCEPTED by the queue tool", async () => {
+    const text = firstText(
+      await defByName("panel_run").handler({}, makeCtx({ queued: true, prompt_id: "p-1789" })),
+    );
+    const call = fallbackCallIn(text);
+    expect(call).toBe('queue (action:"status", prompt_id:"p-1789")');
+    expect(await queueCallIsAccepted(call)).toBe(true);
+  });
+
+  it("BATCH — the remedy the rider prints is ACCEPTED by the queue tool", async () => {
+    const text = firstText(
+      await defByName("panel_run").handler(
+        {},
+        makeCtx({ queued: true, prompt_id: "p-a", prompt_ids: ["p-a", "p-b"] }),
+      ),
+    );
+    const call = fallbackCallIn(text);
+    expect(call).toBe('queue (action:"list")');
+    expect(await queueCallIsAccepted(call)).toBe(true);
+  });
+
+  // The guard's own guard: prove queueCallIsAccepted can SAY NO, or the two
+  // assertions above are a rubber stamp. This is the exact call the first cut
+  // shipped.
+  it("…and that acceptance check is not vacuous — the shipped-broken form is REFUSED", async () => {
+    expect(await queueCallIsAccepted('queue (action:"status")')).toBe(false);
   });
 
   it("the UNCORRELATABLE branch is unchanged — it never told the agent to idle", async () => {
--- a/src/__tests__/orchestrator/run-completion-fallback-wiring.test.ts
+++ b/src/__tests__/orchestrator/run-completion-fallback-wiring.test.ts
@@ -1,0 +1,170 @@
+// #1789 — the two things a unit test of the watchdog cannot see.
+//
+// 1. THE CALL SITE. createRunCompletionWatchdog can be perfect and never run.
+//    The watchdog only exists if the orchestrator TEES QueueMonitor's drained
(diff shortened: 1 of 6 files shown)


artokun:
## Gate finding addressed — new head `bd7f2aa`

**The blocker was real and is fixed.** Verified live before touching anything: driving the actual `queue` handler, `{action:"status"}` with no id returns `isError:true — queue action:"status" requires `prompt_id` — the job to check.`, while `{action:"status", prompt_id:…}` and bare `{action:"list"}` both return JSON. So the batch rider was handing out a call the tool refuses — the reported bug's own shape, one level in.

Omitting the id stays (inventing one would point the agent at an arbitrary member). The remedy now fits what each branch can name:
- **single run** → `queue (action:"status", prompt_id:"<id>")`
- **batch** → `queue (action:"list")`, which takes no id, plus a clause saying it is the queue as a whole and that `get_history` settles the outcome of any no longer in flight.

**The test no longer pins the broken form — it EXECUTES the printed call.** The wiring suite now parses the single remedy out of the `[FALLBACK]` line and runs it against the real `queue` handler, for **both** branches, asserting the tool accepts it. The parser throws on any shape it does not recognise, so a reworded remedy fails loudly rather than going unchecked. And a counter-test proves the acceptance check can say **no** by feeding it the exact call that shipped broken — otherwise the two green assertions would be a rubber stamp.

**Minor correction taken.** `recordCompletion` is fed by the broadcast WS `execution_success`/`_error`/`_interrupted` path as well as the `/history` tail diff, and does not record which one won — so nothing downstream may name `/history`. Both are now named (agent-facing note, warning log, rider text, module header, `index.ts` comment); the payload key stays `orchestrator-history-watchdog` as a stable identifier, and its docstring says explicitly that the channel is not knowable downstream.

**Coalescing claim narrowed** to the case it actually holds for, as asked: past the grace there are two shapes and neither loses anything — *nothing took the synthesised entry* ⇒ still `pending` ⇒ the real payload coalesces onto it, one turn with images (the only path where the extra turn disappears); *an agent did take it* ⇒ the real frame arrives on its own, still `matched`, flagged a possible repeat, never discarded — one extra turn, the standing duplicate-beats-loss trade.

**Nothing else touched.** The delivery half you upheld is byte-identical apart from comment text.

### Refutation — 4 new mutations, all killed; all 6 prior ones re-verified after the rewrite

| # | mutation | result |
|---|---|---|
| 8 | batch branch back to `queue (action:"status")` (the shipped-broken form) | 2 failed — incl. the executing BATCH test |
| 9 | `queueCallIsAccepted` always returns true | 1 failed (the counter-test) |
| 10 | strip the id from the single-run remedy | 2 failed — incl. the executing SINGLE test |
| 11 | drop the `[FALLBACK]` clause entirely | 4 failed |
| 1–4, 6–7 | re-run unchanged | 1 / 1 / 3 / 6 / 2 / 1 failed — all still killed |

Each anchor went through the refuse-on-non-unique harness (it did refuse once, on an anchor I had mistyped to contain the *new* line — caught, corrected, re-applied) and every edit was confirmed with `git diff` before the run.

**Verification:** `tsc --noEmit` clean; `src/__tests__/orchestrator` + `src/__tests__/tools` + `src/services` = **4,445 passed / 278 files**; `check:vocabulary` OK; `check:unknown-collapse` OK. Not merged, no gate script run.

artokun:
**Merge gate — substitution disclosed.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the account weekly limit. Gated by an **independent adversarial reviewer that had never seen the author's reasoning**. First round: **NO-SHIP**. After rework: **SHIP**, head `bd7f2aac`.

**The drop is exactly where this says it is.** On `origin/main`, `QueueMonitor.drainCompletions()` had precisely **one** production call site — `src/orchestrator/index.ts:6199`, feeding `createQueueStatusBroadcaster`, whose only consumer is the `queue_status` UI frame — while `openRun` returns `typeof promptId === "string" && promptId` and nothing more. It proves a completion could be *recognised*, never that one would *arrive*. So the promised completion was already in the orchestrator's hands and was thrown away.

**The 44 s / 46 s race was executed, not reasoned, and the grace holds.** At 44 s: one turn, with pixels. At 46 s with the agent live: two turns — the synthesised pointer, then the real frame **with images, correlation `matched`, not discarded**. No downgrade and no loss; a duplicate turn is this journal's stated duplicate-beats-loss trade. The coalescing path has been narrowed in the body to the case it actually holds for (past the grace with *nothing* having taken the entry).

**The watchdog cannot announce someone else's render.** Tickets exist only via `openRun`, whose sole callers are the two `panel_run` lines, so a human's Queue Prompt never arms it — mutating that guard out kills 2 tests.

**The first round's blocker, and why it mattered.** The batch branch correctly omitted the prompt id rather than inventing one — but the remedy it printed was `queue (action:"status")` with no id, and `requirePromptId("status", …)` **throws** on `undefined`. The gate drove the real handler to confirm it. So the fix for a promise the code could not keep was shipping an instruction the tool rejects: the reported bug's own shape, one level in. Worse, the wiring test **pinned the broken form**, so it would never have surfaced later.

**The rework fixed the structural problem, not just the string.** The remedy now matches what each branch can name — single run `queue (action:"status", prompt_id:"<id>")`, batch `queue (action:"list")`, which takes no id, with a clause saying it is the queue as a whole and that `get_history` settles anything no longer in flight. And **the test now parses the remedy out of the `[FALLBACK]` line and executes it against the real `queue` handler for both branches**, asserting acceptance, with the parser throwing on any unrecognised shape so a reword fails loudly. A counter-test feeds it the exact call that shipped broken, proving the acceptance check can say *no* — without it, two green assertions would be a rubber stamp.

**The honest-promise half reads correctly:** the anti-poll instruction stays primary and the hatch is triple-bounded ("LONG after", "roughly twice", "not before then"), so "one check, not a loop" cannot be read as licence to resume polling.

**Not redundant with the panel-side fix.** `comfyui-mcp-panel#1403` merged 10:01Z; #1789 was filed 12:51Z on panel 0.14.44, several minors behind, and #1403 additionally requires a failed `sendFrame` the report cannot establish. "Both, deliberately" holds. The 45 s grace checks out against the real constant — `run-reconcile-sweep.js` is `intervalMs = 20000`.

**A class-6 correction was taken:** the notice claimed it was synthesised "from ComfyUI's own history", but `recordCompletion` is also fed by the WS `execution_success`/`_error`/`_interrupted` path and does not record which won. Nothing downstream names `/history` any more; both channels are named, and the stable payload key keeps a docstring saying the channel is not knowable downstream.

**Mutations:** 6 prior all re-verified after the rewrite (1/1/3/6/2/1), plus 4 new, all killed — batch back to bare `status` fails 2 including the executing BATCH test; a vacuous acceptance check fails the counter-test; stripping the id from the single remedy fails 2; dropping `[FALLBACK]` fails 4. The refuse-on-non-unique-anchor harness **fired once**, on an anchor mistyped to contain the new line — which is exactly the two-site failure that guard exists for. `tsc --noEmit` clean; 4,445 tests across 278 files; `check:vocabulary` and `check:unknown-collapse` OK. The probe test used to establish the live refusal was deleted rather than shipped.

PR #1792, opened 2026-08-19
fix(1788): a PINNED extra-paths target says so when the running server reads a different config

### LINKED ISSUE #1788: Fix model download and Desktop path targeting for custom model categories
On ComfyUI Desktop, download_model wrote an IP-Adapter file to the install-local models/ipadapter directory even though the running server's --extra-model-paths-config assigned the ipadapter category to ComfyUI-Shared/models/ipadapter. The file was verified on disk but IPAdapterUnifiedLoaderFaceID raised 'IPAdapter model not found'. list_local_models action:list_paths correctly identified the live instance YAML, but action:add_path with target:'desktop' wrote a different inactive config under AppData/Roaming/ComfyUI/extra_models_config.yaml. Using config_path on the live instance YAML was temporary because Desktop regenerated it at restart.

### Steps to reproduce
1. Run ComfyUI Desktop with --extra-model-paths-config pointing at an instance-model-paths YAML whose ipadapter category is under ComfyUI-Shared/models.
2. Call download_model action:download target_subfolder:'ipadapter'.
3. Restart and run IPAdapterUnifiedLoaderFaceID.
4. Observe the model was written under the install-local root and is not visible to the loader.
5. Call list_local_models action:add_path target:'desktop'; observe it edits a different inactive YAML than action:list_paths reported for the running server.

### Exact error (scrubbed)
IPAdapterUnifiedLoaderFaceID: IPAdapter model not found.

### Fix
Applied locally: workaround only. Upstream-only: yes. Resolve destination directories from the running server's effective per-category model paths, especially custom categories such as ipadapter. For add_path target:'desktop', prefer the running server's explicit --extra-model-paths-config path when present, or clearly refuse rather than editing an inactive fallback. Workaround verified: move the already-downloaded file into ComfyUI-Shared/models/ipadapter, refresh nodes, and rerun.

### Environment
Windows 11; ComfyUI 0.33.2; frontend 1.49.6; Python 3.13.12; RTX 5080 16 GB; comfyui-mcp 0.52.0; panel 0.14.44.

artokun:
Closes #1788.

The report has two halves. **Only one of them still needed code.**

---

## Half 1 — the download landed in the install-local `models/ipadapter`: already shipped

The obvious hypothesis was a hardcoded category set that omits `ipadapter`. **Refuted by
execution, not by reading.** `getLiveExtraModelRoots` enumerates the config's OWN keys
(`readGroup` skips only `base_path`/`is_default`), and the redirect's only category filter is
`NON_MODEL_EXTRA_CATEGORIES = {"custom_nodes"}`. A probe on the reporter's exact shape —
unvouched primary, absolute `--extra-model-paths-config`, `ipadapter` — redirects
`…/ComfyUI-Installs/ComfyUI/ComfyUI/models/ipadapter` → `…/ComfyUI-Shared/models/ipadapter`,
and verifies VISIBLE post-write.

The real reason is duller: **the reporter runs comfyui-mcp 0.52.0.**
`git merge-base --is-ancestor 0580637 v0.52.3` → no; `v0.52.4` → yes. #1734 first shipped in
**v0.52.4**, four patch releases after the reporter's build. `ipadapter` was never a special
case, because the mechanism is keyed on the live config.

It is also none of the three follow-ups recorded on #1734 — not the `.find()` first-match (one
group, one mapping), not the current-contents read (the config predates the download), not the
`verified_root` stamp (on 0.52.0 the file never reached the extra root at all).

No product change here. What this PR adds is four permanent cases in
`download-live-destination.test.ts` (`ipadapter`, `instantid`, `ultralytics`, `clip_vision`)
plus a post-write `verifyLandedModel` case, because nothing pinned category-agnosticism and a
later "only these categories" narrowing would have shipped silently. Gating off the #1734
redirect (`if (false && argvNamesConfig && …)`, diff confirmed) kills all four new cases plus
three originals — so they pin the shipped mechanism rather than decorating it.

## Half 2 — `add_path target:"desktop"` edited an inactive YAML and promised a restart would apply it

`list_paths` "correctly identified the right root" and `add_path target:"desktop"` did not —
the sharpest clue in the report, and it points at a different function.
`resolveTargetPathPreferServer` short-circuits on an explicit `target`/`config_path`:

```ts
if (opts.configPath || (opts.target && opts.target !== "auto")) {
  return { ...resolveTargetPath(opts), serverResolved: false };
}
```

so `target:"desktop"` returned the static `%APPDATA%/ComfyUI/extra_models_config.yaml` with
`notes: []`, wrote it, and answered **"Added … Restart ComfyUI to apply it"** for a file the
running Desktop server does not read. Note the asymmetry that was actually there: every
*live-anchored* branch already pushes a divergence note when the static guess would have
differed; the pinned branch — the direction that actually loses the user's edit — pushed none.

**The pin stays.** It is the documented escape hatch (the schema says in so many words: *"Use
standalone/desktop (or config_path) to deliberately target a file the running server does not
read"*), and the bytes still land exactly where the caller said. What changes is that the
answer stops promising an effect. The pinned branch re-enters the **same canonical resolver**
with no pin (one level; `{}` cannot re-enter the branch — no second copy of the resolution to
drift) and discloses divergence in `notes` and in the mutation `message`, keyed on a structured
`divergesFromLive` field, never on re-parsed prose. `removeExtraPath` and `listExtraPaths` get
it too.

It can only ever ADD a note. Silence — never a confident negative — for: an unreachable or
remote server; a tree not proven to be this machine's; the #764 display fallbacks; an
**incomplete** config set; and a pin naming any config the server loads, **including the second
`--extra-model-paths-config`** and the not-yet-created implicit `<root>/extra_model_paths.yaml`
(pinning that is how you create it).

A cheap reachability gate runs before any resolution work: the full resolution stats the
caller's guarded root, and that root's #648 TOCTOU revalidation counts every stat — burning one
there to learn nothing turned a stable root into a spurious "REPLACED while this call was
running". The gate is a gate only; the inner call takes its own authoritative snapshot, so a
server starting or stopping between the two can suppress the note, never manufacture one.

---

## Gate record — 3 rounds, 3 NO-SHIP/SHIP transitions, every finding fixed rather than argued with

**codex was unavailable (account exhausted until 2026-08-19 20:43); gated by an independent
adversarial review instead** — `.claude/autopilot/claude-gate.mjs`, a fresh reviewer that never
saw the implementing session's reasoning. Disclosing the substitution is the condition it was
accepted under. Both rounds ran the code rather than reading it, and both P1s below were
reproduced on this real Windows box before a line was written.

### Round 1 — NO-SHIP, 2 × P1

1. **`sameConfigFile`'s realpath collapse was dead exactly where it mattered.** `realpathSync`
   throws on a path that does not exist, and "does not exist yet" is the ORDINARY state here —
   pinning a config is how a caller CREATES it. So the collapse added to survive
   junctions/8.3/mapped drives silently degraded to the lexical compare it was added to fix,
   and a junctioned `COMFYUI_PATH` naming the very file the server reads was declared inert.
   Reproduced with a real Windows junction. Fixed: resolve the **parent directory** and rejoin
   the basename when the file itself does not resolve.
2. **Withdrawing the promise from `message` left it standing in `notes`.** `summarize` appends
   *"Restart ComfyUI after editing this file so startup path registration is rebuilt."* to every
   answer — the same promise in different words, different enough that the
   `/Restart ComfyUI to apply it/` assertions never saw it. The payload carried both "restarting
   will not apply this" and "restart to apply this", and an agent reads `notes`. Fixed: keyed on
   the same structured `divergesFromLive`, threaded through all three `summarize` call sites.

### Round 2 — NO-SHIP, 1 × P1

3. **Only `realpathSync.native` expands a Windows 8.3 short name.** The JS binding follows
   junctions and symlinks but returns a short name unchanged — a distinction this repo had
   already encoded in `panel-installer.ts` and `node-dev.ts`, and which this function's own
   docblock *claimed* to handle while the code did not. Measured: for a temp dir whose 8.3 alias
   is `COE80~65`, `realpathSync` returns `…\COE80~65\INSTAN~1.YAM` unchanged while `.native`
   returns the long path — so pinning the server's OWN config by its short spelling emitted "the
   running ComfyUI does NOT read …" for a file it demonstrably reads. Fixed: `.native` first for
   both the file and the parent step, JS binding as fallback. Three Windows-only tests, **skipped
   rather than vacuously passing** when the volume has 8dot3name generation disabled.

### Round 3 — **SHIP**

> "Nothing material. Verified by execution, not reading."

Worked all seven taxonomy classes with the specific thing checked named for each; ran the
absolute-first/relative-second argv, repeated occurrences and `nargs='+'`; probed 8.3 short
names, a `subst` mapped drive, absent-file-via-parent and a 447-char path on this real Windows
box. Deleting the `notePinnedDivergence` call kills 8 tests — reachable.

It reported **one surviving mutation out of 11**: forcing `complete: true` on the *multi-flag*
`liveLoaded` branch killed zero tests. It executed that argv shape and confirmed the code
suppresses correctly, so it called it a test gap, not a defect, and not a ship blocker. **Fixed
anyway** in `df9f970`, because an unpinned discriminator is one refactor away from becoming a
defect: the existing INCOMPLETE test drives the SINGLE-flag arm (`implicitLiveTarget`,
`unresolvableFlag`), while the multi-flag arm computes completeness from `unlocatable`, filled
only from flags 2..n. Two opposed tests now cover it — absolute-first + unlocatable-second must
stay silent, absolute-first + absolute-second must speak — and forcing `complete: true` kills
the first.

Two round-3 prose nits, both non-blocking and left as-is: the message says the server "loads"
the implicit `extra_model_paths.yaml` when it does not exist yet (ComfyUI gates on
`os.path.isfile`, though the suggested remedy is still right), and a comment says the root guard
"counts every stat it takes" when `assertRootIntact` compares dev/ino.

## Refutation — mutations, every diff confirmed applied

Anchors asserted unique, and an empty diff refuses rather than measuring nothing (an early
battery silently applied **none** of them — a heredoc had eaten the escapes — and the refusal
caught it).

| mutation | tests killed |
|---|---|
| delete the disclosure at the **call site** | 4 |
| delete the structured `divergesFromLive` the message keys on | 3 |
| compare only the primary resolution, not every loaded config | 2 |
| make the config comparison lexical again | 1 |
| delete the reachability gate | 2 (the #648 root-guard ordinals) |
| drop the `complete` requirement | 1 |
| drop the not-yet-created implicit yaml from the set | 1 |
| delete the **parent-directory** realpath fallback (round 1 fix) | 1 |
| make the restart note unconditional again (round 2 fix) | 2 |
| stop threading `divergesFromLive` to `summarize` (the wiring, not the helper) | 2 |
| force `complete: true` on the multi-flag arm (round 3 gap) | 1 |

## CI

`tsc --noEmit` clean; `vitest run` **571 files / 10527 tests green**; `check:vocabulary` (1694
files, no live dead names), `check:unknown-collapse` (0 collapses), `vocab:export --check`,
`asset-counts --check`, `i18n:check` and all `check:docs-*` / `check:blog-*` gates pass.

## Deliberately not done

- **No change to where a pinned target writes.** "Prefer the server's config" or "refuse" would
  delete the documented escape hatch; the defect was the false promise, not the destination.
- **`isDesktopGeneratedConfig` not widened.** The reporter says Desktop regenerated their live
  instance YAML, but the report never gives that file's path or header, and inventing a regex
  from prose is the wrong-pair defect this PR exists to fix. The existing
  `Comfy Desktop/…/shared_model_paths.yaml` warning is unchanged.
- **No panel-rendered surface changes** (`src/services/extra-paths.ts` and tests only), so the
  Playwright browser gate does not apply. It was not run and no such coverage is implied.
- **No model weights downloaded** — the destination resolution is asserted directly.

### DIFF
--- a/src/__tests__/services/download-live-destination.test.ts
+++ b/src/__tests__/services/download-live-destination.test.ts
@@ -580,6 +580,38 @@ describe("pre-write: a SERVER-NAMED extra model root wins over an unvouched prim
     expect(h.fetchCalls).toEqual([]);
   });
 
+  // #1788 — the same report, one release later, on `ipadapter`: a CUSTOM category
+  // that no ComfyUI core folder list contains. The redirect must be keyed on the
+  // config's OWN keys, never on a known-category set: a hardcoded set would leave
+  // every custom category (ipadapter, instantid, ultralytics, the categories custom
+  // node packs register) writing into the unvouched primary while loras worked.
+  it.each(["ipadapter", "instantid", "ultralytics", "clip_vision"])(
+    "redirects the CUSTOM category %s exactly like a core one (#1788)",
+    async (category) => {
+      const sharedRoot = resolve(`/shared/models/${category}`);
+      h.liveExtraRoots = {
+        authoritative: true,
+        roots: [{ category, dir: sharedRoot, group: "desktop" }],
+      };
+      await expect(resolveModelSubfolderPreferServer(category)).resolves.toBe(sharedRoot);
+    },
+  );
+
+  it("post-write: a landed ipadapter file under the named extra root verifies VISIBLE (#1788)", async () => {
+    const sharedIpadapter = resolve("/shared/models/ipadapter");
+    h.liveExtraRoots = {
+      authoritative: true,
+      roots: [{ category: "ipadapter", dir: sharedIpadapter, group: "desktop" }],
+    };
+    h.liveListings["ipadapter"] = ["ip-adapter-faceid_sdxl.bin"];
+    const res = await verifyLandedModel(
+      resolve(sharedIpadapter, "ip-adapter-faceid_sdxl.bin"),
+      "ipadapter",
+      { attempts: 1, retryMs: 0 },
+    );
+    expect(res.liveVisible).toBe("visible");
+  });
+
   it("does NOT redirect a category the config does not map", async () => {
     await expect(resolveModelSubfolderPreferServer("loras")).resolves.toBe(
       resolve("/comfy/models/loras"),
--- a/src/__tests__/services/extra-paths-root-guard.test.ts
+++ b/src/__tests__/services/extra-paths-root-guard.test.ts
@@ -286,6 +286,52 @@ describe("inferred root revalidation at the point of use", () => {
     expect(script.consumed).toBe(4);
   });
 
+  // #1788 — the divergence disclosure compares a PINNED config path against every file
+  // the running server loads. That comparison is realpath-collapsed on purpose: here the
+  // harmful direction is the false NEGATIVE, i.e. telling a user their edit is inert when
+  // it lands in the very file the server reads. Scripted realpathSync so it is proven on
+  // every platform (a real file symlink needs privileges on Windows).
+  it("pinned config: a realpath ALIAS of the server's own config is not called inert", async () => {
+    const liveRoot = await trackTmp();
+    await writeFile(join(liveRoot, "main.py"), "# comfyui\n", "utf-8");
+    const serverCfg = join(liveRoot, "instance-model-paths.yaml");
+    await writeFile(serverCfg, "live:\n  ipadapter: D:/shared/ipadapter\n", "utf-8");
+    const aliasDir = await trackTmp();
+    const alias = join(aliasDir, "alias.yaml");
+    await writeFile(alias, "live:\n  ipadapter: D:/shared/ipadapter\n", "utf-8");
+    script.realpath[alias] = serverCfg;
+    mockGetSystemStats.mockResolvedValue({
+      system: {
+        argv: ["python", join(liveRoot, "main.py"), "--extra-model-paths-config", serverCfg],
+      },
+    });
+
+    const added = await addExtraPath({ configPath: alias, category: "vae", path: "D:/vae" });
+    expect(added.notes.some((n) => /does not read this file/.test(n))).toBe(false);
+    expect(added.message).toMatch(/Restart ComfyUI to apply it/);
+  });
+
+  it("pinned config: a path that is NOT an alias of any server config IS called inert", async () => {
+    // The control for the test above: without it, a disclosure that never fires at all
+    // would pass it just as well.
+    const liveRoot = await trackTmp();
+    await writeFile(join(liveRoot, "main.py"), "# comfyui\n", "utf-8");
+    const serverCfg = join(liveRoot, "instance-model-paths.yaml");
+    await writeFile(serverCfg, "live:\n  ipadapter: D:/shared/ipadapter\n", "utf-8");
+    const other = join(await trackTmp(), "unrelated.yaml");
+    await writeFile(other, "local:\n  vae: D:/vae\n", "utf-8");
+    mockGetSystemStats.mockResolvedValue({
+      system: {
+        argv: ["python", join(liveRoot, "main.py"), "--extra-model-paths-config", serverCfg],
+      },
+    });
+
+    const added = await addExtraPath({ configPath: other, category: "vae", path: "D:/vae2" });
+    expect(added.notes.some((n) => /RUNNING ComfyUI does not read this file/.test(n))).toBe(true);
+    expect(added.message).not.toMatch(/Restart ComfyUI to apply it/);
+    expect(added.message).toContain(serverCfg);
+  });
+
   it("live root: a SYMLINKED main.py resolves to the real install root (platform-independent)", async () => {
     // ComfyUI reads the implicit config next to os.path.realpath(__file__). Scripted
     // realpathSync stands in for a symlink so this runs everywhere, including on a
--- a/src/__tests__/services/extra-paths-root-guard.test.ts
+++ b/src/__tests__/services/extra-paths-root-guard.test.ts
@@ -332,6 +332,60 @@ describe("inferred root revalidation at the point of use", () => {
     expect(added.message).toContain(serverCfg);
   });
 
+  // #1788 gate round 1, P1. The realpath collapse above is DEAD exactly when the pinned
+  // file does not exist yet — `realpathSync` throws on an absent path, so the comparison
+  // silently degraded to the lexical one it was added to fix. And "does not exist yet" is
+  // the ORDINARY state here: pinning a config is how a caller creates it. A junctioned
+  // ComfyUI root therefore had the very file the server reads declared inert. The
+  // aliasing lives in the PARENT, so the parent is what must be resolved.
+  it("pinned config: a not-yet-created file under a JUNCTIONED root is not called inert", async () => {
+    const realRoot = await trackTmp();
+    await writeFile(join(realRoot, "main.py"), "# comfyui\n", "utf-8");
+    const aliasRoot = await trackTmp();
+    // A junction: two spellings of ONE directory. Scripted so it runs without the
+    // privileges a real Windows junction needs; identity-mapping the real root keeps
+    // both sides of the comparison in the same spelling the argv resolution produced.
+    script.realpath[aliasRoot] = realRoot;
+    script.realpath[realRoot] = realRoot;
+    mockGetSystemStats.mockResolvedValue({
+      system: { argv: ["python", join(realRoot, "main.py")] },
+    });
+
+    const pinned = join(aliasRoot, "extra_model_paths.yaml");
+    expect(existsSync(pinned)).toBe(false); // the case that used to break it
+    const added = await addExtraPath({ configPath: pinned, category: "vae", path: "D:/vae" });
+
+    // It IS the server's implicit config, under another name for the same directory.
+    expect(added.notes.some((n) => /RUNNING ComfyUI does not read this file/.test(n))).toBe(
+      false,
+    );
+    expect(added.message).toMatch(/Restart ComfyUI to apply it/);
+    // The pin is still honoured — the write went to the spelling the caller gave. (With
+    // a REAL junction that pathname and the server's are one file; the scripted alias
+    // here models only the realpath answer, so the two directories stay distinct on
+    // disk and the assertion is about the pin, not about the junction.)
+    expect(existsSync(pinned)).toBe(true);
+  });
+
+  it("pinned config: a not-yet-created file under an UNRELATED root still IS called inert", async () => {
+    // The control. Without it, a comparison that had simply stopped firing would pass
+    // the junction case above just as well.
+    const realRoot = await trackTmp();
+    await writeFile(join(realRoot, "main.py"), "# comfyui\n", "utf-8");
+    const elsewhere = await trackTmp();
+    script.realpath[realRoot] = realRoot;
+    script.realpath[elsewhere] = elsewhere;
+    mockGetSystemStats.mockResolvedValue({
+      system: { argv: ["python", join(realRoot, "main.py")] },
+    });
+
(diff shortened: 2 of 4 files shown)


artokun:
**CI green**, head `36127a7`: Build pass on ubuntu / macos / windows, Pack & install smoke pass, main unmoved (`60c513f`), mergeStateStatus CLEAN.

**Not gated and not merged, by instruction.** `codex exec` is account-exhausted until 2026-08-19 20:43 and `claude-gate.mjs` is blocked on the weekly limit, so no gate script was run — and a self-review is never the gate. Handing off for gating + merge.

The two things worth a gate's attention: the pinned branch now re-enters `resolveTargetPathPreferServer` with `{}` (one level, and `{}` cannot re-enter that branch), and the cheap reachability gate in front of it exists because the full resolution stats the caller's guarded root and the #648 TOCTOU revalidation counts every stat — deleting that gate is mutation 5 in the table and it fails two root-guard tests.

artokun:
**Gate round 2 addressed — new head `d2caf5b`.** CI green on the new head: Build pass ubuntu/macos/windows, Pack & install smoke pass.

**The 8.3 finding is correct and I reproduced it on this machine before changing anything.** For a temp dir whose 8.3 alias is `COE80~65`:

| call | result |
|---|---|
| `realpathSync(shortCfg)` | `…\COE80~65\INSTAN~1.YAM` — **unchanged** |
| `realpathSync.native(shortCfg)` | `…\comfyui-…-long-install-name-dgc4bE\instance-model-paths.yaml` |
| `realpathSync(shortDir)` | `…\COE80~65` — **unchanged** |
| `realpathSync.native(shortDir)` | long path |

So both channels you named were real: the file step AND the parent-directory step. `.native` first for both, JS binding as fallback so an environment without the native call still collapses junctions rather than nothing. Docblock corrected — it now names which binding is load-bearing, why, and cites `panel-installer.ts`/`node-dev.ts` as the precedent it should have followed.

**Refutation, 3 mutations, 3 killed, each diff printed:**
- `.native` → `realpathSync`: kills **exactly** the two 8.3 tests, leaves the junction and control cases green — the discriminating signal.
- parent-directory step → lexical passthrough: kills the junction-not-yet-created case and the 8.3-implicit case.
- "could not resolve" collapsed back into "resolved to itself": same two.

Three Windows-only tests via `describe.runIf(win32)` + `it.skipIf(!CAN_SHORT_PATH)`, probing `Scripting.FileSystemObject.ShortPath` once at module scope so a volume with `8dot3name` disabled **skips rather than passes vacuously**. They genuinely ran here (386/381/369 ms). The third is the over-collapse control: an unrelated file in the same 8.3 directory is still flagged.

Also folded in one case from your scratch probe that my tests did not cover: `--extra-model-paths-config` is argparse `nargs='+'`, so **one** occurrence can carry two values — a different argv shape from the flag repeated twice, and it must also count as configs the server reads. Silent, as expected.

**Verified on a clean detached worktree at `d2caf5b`, not in `drain-1788`** — see the note below. Full suite **571 files / 10527 passed / 4 skipped / 0 failed**, `tsc --noEmit`, `check:unknown-collapse`, `check:vocabulary` clean.

Not merged, no gate script run — `origin/main` has since moved to `57da6e7` (release 0.52.9), PR still `MERGEABLE`.

artokun:
**Merge gate — substitution disclosed.** `codex exec` remains account-exhausted until 2026-08-19 20:43. This PR was gated with the project's **sanctioned substitute**, `.claude/autopilot/claude-gate.mjs` (exit 0 = SHIP), which spawns a fresh reviewer that has never seen the author's reasoning — the account's weekly limit reset at 5am PT, so that path is available again. It was gated a second time beforehand by an independent adversarial reviewer, which returned NO-SHIP; both rounds are recorded below.

Merging at `df9f970c`. The scripted gate ran against `d2caf5b`; the only change since is **test-only and additive** (one file, +44 lines) closing a mutation the gate found surviving, so the verdict is not stale.

**Round 1 — NO-SHIP, and the finding was a Windows path defect.** `sameConfigFile` used `fs.realpathSync`, the JS binding, which on Windows **does not expand 8.3 short names** — only `fs.realpathSync.native` does, and this repo already uses `.native` for exactly that reason in `src/services/panel-installer.ts:631` and `src/services/node-dev.ts:229`. Pinning the server's own config by its short spelling would have emitted *"the running ComfyUI does NOT read … Restarting will NOT apply this change"* for a file it demonstrably reads — a confident wrong negative replacing harmless silence, on the platform the bug was reported from. A second channel shared the root: `realpath` throws ENOENT on a not-yet-existing path, so pinning the implicit `extra_model_paths.yaml` **through a junction** — the exact case the PR body names as must-be-silent — was also called inert. The docblock claimed 8.3 was handled; it was not.

**Reproduced before the fix, not argued:** for a temp dir aliased `COE80~65`, `realpathSync(shortCfg)` returned `…\COE80~65\INSTAN~1.YAM` unchanged while `.native` returned the long path, identically for the directory. Both the file step and the parent-directory step now try `.native` first with the JS binding as fallback, so an environment without the native call still collapses junctions rather than nothing.

**Round 2 — SHIP.** Eight mutations killed (`sameConfigFile`→`samePath`, complete-gate, reachability gate, notes suppression, `inertEditSuffix`, the whole feature, `implicitLiveTarget` completeness, implicit-in-comparison-set), and live probes on real Windows junctions, 8.3 short names, **428-char paths**, `nargs='+'` argv, and relative/second-flag configs all resolved in the conservative direction. Residues are prose only (P3): "Restarting will NOT apply this change" on a no-op remove, and naming a not-yet-created implicit yaml as one the server "loads".

**The issue's reported half is not a live bug, and that was established by execution.** I dispatched this suspecting #1734's redirect missed custom categories via a hardcoded list. It is not category-keyed: the only filter is `NON_MODEL_EXTRA_CATEGORIES = {"custom_nodes"}`, and roots come from `getLiveExtraModelRoots` enumerating the config's own keys. Writing that hypothesis as a mutation — swapping the filter for a `{loras,checkpoints,vae}` allowlist — kills exactly the four new custom-category tests, so they genuinely pin category-agnosticism. **The reporter runs 0.52.0; `0580637` first ships in v0.52.4.** Zero other categories are exposed.

**What this PR actually fixes is the second half of the report.** `list_paths` was right because `resolveTargetPathPreferServer` resolves from the live server for target `auto`, while `add_path target:"desktop"` short-circuits on the explicit target, writes to the static `%APPDATA%` yaml the running server never reads, and answers "Added … Restart ComfyUI to apply it". Every live-anchored branch already disclosed divergence; the pinned branch — the one direction that silently loses the edit — was the only quiet one. The escape hatch is intact (bytes still land where pinned; `notePinnedDivergence` never touches `path`, `target`, `createParents` or `guard`), and `divergesFromLive` is genuinely structured with one producer, one consumer, and no serialization to the tool result.

**A deleted guard was checked rather than assumed:** reinstating `if (!live.serverResolved) return pinned;` kills zero tests, a runtime assertion that it disagrees with `liveLoaded` never fires across 113 tests, and inverting it kills four — unreachable, not merely untested.

**Two process notes worth keeping.** The author's mutation loop `git checkout`-reverted its own uncommitted fix, and the anchor-uniqueness/empty-diff guard caught it by reporting "matched 0 sites" rather than a false "killed nothing". And a concurrent gate process was writing into the same worktree mid-session, so the author committed **by path** and verified the committed blob rather than the working tree, running the suite in a throwaway detached worktree at the exact SHA (571 files / 10,527 passed / 0 failed).

PR #1794, opened 2026-08-19
fix(401): environment no longer reports packages missing off the Homebrew base python

### LINKED ISSUE #401: get_environment probes the WRONG Python: false 'Triton: not installed' led an agent to disable working acceleration

artokun:
## Why

`install_comfyui(action:"environment")` and the ENVIRONMENT banner still reported torch / Triton as missing after the earlier #401 fixes. The 0.52.1 recurrence is macOS + Homebrew: the process-table probe sees the **base** interpreter (Python.app re-execs; `ps` argv[0] is `/opt/homebrew/.../Python`), marks it trusted because the venv reports the same `sys.version`, then `pip show` finds none of torch/torchvision/torchaudio — while the reachable server's `/system_stats` already reports PyTorch 2.12.1.

That is the same harmful false negative as the original report: an agent treats "not installed" as fact and disables working acceleration.

## What changed

**Recover the venv the process itself recorded.** After argv correlation, prefer `__PYVENV_LAUNCHER__` (CPython's macOS framework hook) or `$VIRTUAL_ENV/bin/python` from the live process environment over argv[0]. Those are observations, not layout guesses. argv[0] remains the fallback when the hint is missing or the path is gone.

**Refuse a trusted package list that contradicts `/system_stats`.** A venv and its base share a python banner, so the version check cannot tell them apart. If the probe's pip/import says torch is absent (or a different build) while the server reports `pytorch_version`, the source is discredited: `python_probe_trusted` is false, `packages` is omitted, and Triton/SageAttention negatives become `unknown`. A failed pip query (uv venvs often have no pip) is still "we could not tell", not "torch is absent".

## Tests

- `live-interpreter.test.ts` — parse Linux environ / `kern.procargs2`, prefer the launcher python over a real Homebrew argv[0]
- `workspace-env.test.ts` — getEnvironment untrusts a process-table hit with no torch (or the wrong torch) when `/system_stats` reports pytorch; matching torch still reports packages; a failed pip query is not a contradiction
- `env-capabilities-argv-contradiction.test.ts` — ENVIRONMENT banner voids Triton when the probe has no torch but the server reports pytorch

`tsc --noEmit` clean. Targeted vitest 173/173.

Fixes #401

### DIFF
--- a/src/__tests__/services/env-capabilities-argv-contradiction.test.ts
+++ b/src/__tests__/services/env-capabilities-argv-contradiction.test.ts
@@ -64,7 +64,7 @@ function statsPayload(argv: string[]): unknown {
  * same python banner the server reports — the venv/base pair the version check cannot
  * tell apart.
  */
-function probeSays(triton: boolean, sage: boolean): void {
+function probeSays(triton: boolean, sage: boolean, torch?: boolean): void {
   execFileMock.mockImplementation(
     (
       _file: string,
@@ -72,11 +72,16 @@ function probeSays(triton: boolean, sage: boolean): void {
       _opts: unknown,
       cb: (e: Error | null, stdout: string, stderr: string) => void,
     ) => {
+      const torchLine =
+        torch === undefined
+          ? ""
+          : `torch ${torch ? "True" : "False"}\ntorch_version ${torch ? "2.4.0" : ""}\n`;
       cb(
         null,
         `python 3.13.12 (main, Feb 12 2026, 00:38:53) [MSC v.1944 64 bit (AMD64)]\n` +
           `triton ${triton ? "True" : "False"}\n` +
-          `sageattention ${sage ? "True" : "False"}\n`,
+          `sageattention ${sage ? "True" : "False"}\n` +
+          torchLine,
         "",
       );
       return { on: () => undefined };
@@ -148,6 +153,26 @@ describe("contradictedPackage (pure)", () => {
     expect(contradictedPackage({ triton: "not-installed", sageattention: "installed" }, proven)).toBeUndefined();
     expect(contradictedPackage({ triton: "installed", sageattention: "unknown" }, proven)).toBeUndefined();
     expect(contradictedPackage(undefined, proven)).toBeUndefined();
+  });
+
+  it("flags a probe that finds no torch while /system_stats reports pytorch", () => {
+    expect(
+      contradictedPackage(
+        { triton: "not-installed", sageattention: "not-installed", torch: "not-installed" },
+        new Set(),
+        "2.12.1",
+      ),
+    ).toBe("torch");
+  });
+
+  it("does not flag an honest unknown torch, even when the server reports pytorch", () => {
+    expect(
+      contradictedPackage(
+        { triton: "not-installed", sageattention: "not-installed", torch: "unknown" },
+        new Set(),
+        "2.12.1",
+      ),
+    ).toBeUndefined();
   });
 });
 
@@ -257,5 +282,43 @@ describe("gatherEnvCapabilities — a contradicted probe cannot state ANY negati
     });
 
     expect(caps.sageattention).toBe("not-installed");
+  });
+
+  it("voids Triton when the probe has no torch but /system_stats reports pytorch (#401)", async () => {
+    // 0.52.1: Homebrew base, versions agree, no --use-sage-attention to contradict
+    // anything, and the ENVIRONMENT banner said Triton is not installed. The
+    // server's own pytorch_version is the observation that discredits the source.
+    comfyuiFetchMock.mockImplementation(async (url: string) => {
+      if (String(url).includes("/system_stats")) {
+        return {
+          ok: true,
+          json: async () => ({
+            system: {
+              os: "darwin",
+              python_version: "3.12.13 (main, Feb 12 2026) [Clang 17.0.0]",
+              comfyui_version: "0.27.0",
+              pytorch_version: "2.12.1",
+              embedded_python: false,
+              argv: ["python", "main.py"],
+            },
+            devices: [{ name: "Apple M4 Max", type: "mps" }],
+          }),
+        };
+      }
+      return { ok: false, json: async () => ({}) };
+    });
+    probeSays(false, false, false);
+
+    const caps = await gatherEnvCapabilities({
+      comfyuiUrl: "http://127.0.0.1:8188",
+      statsTimeoutMs: 50,
+      tritonTimeoutMs: 50,
+    });
+
+    expect(caps.torch).toBe("2.12.1");
+    expect(caps.triton).toBe("unknown");
+    expect(caps.triton).not.toBe("not-installed");
+    expect(caps.sageattention).toBe("unknown");
+    expect(caps.sageattention).not.toBe("not-installed");
   });
 });
--- a/src/__tests__/services/live-interpreter.test.ts
+++ b/src/__tests__/services/live-interpreter.test.ts
@@ -6,11 +6,14 @@ import { join, sep } from "node:path";
 import {
   argv0FromCommandLine,
   commandLineMatchesArgv,
+  interpreterFromVenvHints,
   recordLaunchedInterpreter,
   clearLaunchedInterpreter,
   getLaunchedInterpreterRecord,
   observeLiveServerProcess,
   resolveLiveInterpreter,
+  venvHintsFromEnvironBuffer,
+  venvHintsFromProcArgs2,
 } from "../../services/live-interpreter.js";
 
 /** ComfyUI Desktop's real shape, as measured on the machine that filed #401. */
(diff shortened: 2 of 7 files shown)


artokun:
Independent review (not the author). Full evidence and remedies: https://github.com/artokun/comfyui-mcp/issues/401#issuecomment-5343494449

Independent review of the in-flight fix on `fix/401-environment-wrong-python` (`7005a10`), posted before the PR opens so it can be folded in rather than gated twice. I am **not** taking the issue over — @grok-swarm's claim stands and the approach is right. Two defects, both measured on the rig, plus one nit.

---

### P1 — the `VIRTUAL_ENV` fallback can reintroduce #401's exact bug on a machine that never had it

`observeLiveServerProcess` now lets the process-env hint beat argv[0] **unconditionally**:

```js
const fromEnv = identity?.venvPython;
if (fromEnv && isAbsolute(fromEnv) && existsSync(fromEnv)) {
  return { python: fromEnv, source: "process-table", pid, image };   // <-- wins outright
}
const argv0 = argv0FromCommandLine(...);                              // <-- never consulted
```

`__PYVENV_LAUNCHER__` earns that priority — CPython sets it during the framework re-exec, so it genuinely describes *this* process. **`VIRTUAL_ENV` does not.** `activate` exports it, and every child inherits it regardless of which interpreter that child actually is. Measured:

```
$ VIRTUAL_ENV=/tmp/ve401 python -c "import os,sys; print(sys.executable); print(os.environ['VIRTUAL_ENV'])"
sys.executable = C:\Users\Artokun\miniconda3\python.exe          <-- what it really imports from
VIRTUAL_ENV    = C:/Users/Artokun/AppData/Local/Temp/ve401       <-- what the hint would report
```

Those are two unrelated interpreters. `interpreterFromVenvHints` resolves `$VIRTUAL_ENV/bin/python`, it exists, so it is returned as `source: "process-table"` — i.e. **trusted**, and the definitive `not-installed` claims are re-enabled off it. Failure scenario: a venv is active in the shell (or the unit file, or the Manager restart context) and ComfyUI is launched by explicit absolute path — `/opt/comfy/.venv/bin/python main.py`. argv[0] is correct and available; the stale `VIRTUAL_ENV` overrides it; Triton is absent from *that* venv, so the banner reports `Triton: not installed`. That is #401 verbatim, newly reachable on Linux, where the old code answered correctly from argv[0].

Note my transcript above ran on Windows, where `venvPythonFromPid` short-circuits — so the demo is of `VIRTUAL_ENV`'s inheritance semantics, not of the code path. The semantics are identical on Linux/macOS, which is where the path does run.

`src/__tests__/services/live-interpreter.test.ts:499` pins the override (`probes the venv python even when argv[0] is a real Homebrew base`) but nothing covers the stale-hint direction, so this is uncovered rather than a deliberate trade.

**Remedy** — the hint needs corroboration before it may override an absolute, existing argv[0]. `pyvenv.cfg` already carries the discriminator:

```
$ cat /tmp/ve401/pyvenv.cfg
home = C:\Users\Artokun\miniconda3
executable = C:\Users\Artokun\miniconda3\python.exe
```

Accept the `VIRTUAL_ENV` override only when that venv's `home`/`executable` resolves back to argv[0] — i.e. argv[0] really is the base this venv was built from. The macOS re-exec case satisfies it by construction; the inherited-`VIRTUAL_ENV` case cannot. `__PYVENV_LAUNCHER__` can keep its current unconditional priority.

---

### P2 — `__import__('torch')` inside a 5s budget can erase the Triton/Sage answers it was added to protect

The probe gained a real torch import:

```py
print('torch_version', (ts and getattr(__import__('torch'), '__version__', '')) or '')
```

`probeTritonSage` runs under `tritonTimeout = 5000` (`env-capabilities.ts:817`). Measured on this box (local NVMe, warm cache, RTX-class install):

```
import torch                      -> 1.89 s      (38% of the entire budget)
importlib.metadata.version('torch') -> 0.043 s   (44x cheaper)
both return the identical string: 2.10.0+cu130
```

1.89s is the *warm, fast* case; a cold import, a slower disk, or a network/UNC-mounted venv pushes it past 5s. And the failure is total rather than partial, because CPython block-buffers stdout when it is a pipe — a timeout kills the child before **any** line is flushed:

```
$ node buftest.mjs      # cheap prints, then a slow tail-import, 1.5s timeout
err: KILLED (timeout)
stdout length: 0
stdout: ""
```

`probeTritonSage` sees empty stdout and takes its `if (!out.trim())` branch, returning `triton: "unknown", sageattention: "unknown"` — discarding correct answers it had already computed. So on exactly the heavy-CUDA machines this issue is about (the original reporter: RTX 5090, CUDA 13, torch 2.11 dev), the fix for the false `not installed` can degrade a working `Triton: installed` to `unknown`.

**Remedy** — use `importlib.metadata.version('torch')`. Same string, 43ms, no torch import, no CUDA init. Add `-u` to the child's args as cheap insurance so partial output survives a kill regardless.

---

### Nit — `venvHintsFromProcArgs2` env slice

`strings.slice(Math.max(0, argc))` assumes `strings` holds exactly `argc` argv entries, but the reader skips empty tokens (`if (end === i) { i++; continue; }`). A legitimately empty argv element makes `strings` short, so the slice starts inside envp and drops the leading variables. It fails in the safe direction (hints missed → falls back to argv[0]), so it is not blocking — worth a comment noting the assumption if it stays.

---

The `torchVersionsAgree` contradiction check is the right instinct and is the part I would keep unchanged: `/system_stats.pytorch_version` is the one signal that actually discriminates a venv from its base when the python banners agree by construction. My only note is the one above — get the version string without paying for the import.

Happy to hand over patches for either finding if that is faster than folding them in.

artokun:
Pushed both fixes to this branch as `8d5c4a6` (fast-forward, so nothing of yours was rewritten) rather than leaving them as review notes — the PR was green and mergeable, and the P1 turns this fix back into #401 on Linux, so it seemed worse to let it race a merge.

What changed from my comment above:

- **`VIRTUAL_ENV` now needs corroboration** to displace argv[0]: the venv's own `pyvenv.cfg` must name argv[0] as its base (`executable` exactly, or `home` as its immediate directory). A `home` *prefix* test was my first cut and it is too weak — it accepts anything beneath a shared `/usr/bin`; my own fixture caught that, which is why the test now builds each interpreter in its own directory. `__PYVENV_LAUNCHER__` keeps unconditional priority, so the macOS Homebrew case this PR is for is untouched, and with no usable argv[0] the hint is still used — nothing you gained is given back.
- **I moved the resolution** from `readProcessIdentity` to `observeLiveServerProcess`. Behind the OS read the corroboration was unreachable to any caller injecting a `readIdentity` seam, so no test could have exercised it. `ProcessIdentity` now carries the raw `venvHints`; `venvPython` still works, so your injected-seam tests are untouched.
- **torch version now comes from dist metadata**, plus `-u` on the child.

Evidence, all measured rather than reasoned:

| | |
|---|---|
| `import torch` | 1.89 s (38% of the 5 s budget) |
| `importlib.metadata.version('torch')` | 0.043 s |
| both return | `2.10.0+cu130` — identical |
| shipped probe source vs. a real ComfyUI python | `torch_version 2.10.0+cu130` in 598 ms |
| timeout with a slow tail-import | stdout length **0** — every earlier line lost |

Mutation-checked both ways: reverting the corroboration fails 2 tests, restoring `__import__('torch')` fails 1. 168/168 green across `live-interpreter`, `workspace-env`, `env-capabilities-argv-contradiction` and the new file; `tsc --noEmit` clean.

Full suite: 10574 passed, 2 failed — `late-mutation-e2e` (passes alone; the known load flake) and `package-hooks` (asserts `dist/` exists; this worktree was never built). Neither touches these files.

An independent adversarial gate is running now and I will post its verdict here. **codex was unavailable (account exhausted until 2026-08-19 20:43), so that gate is the substitute** — disclosing it because that is the condition the substitution was accepted under. Your call on the merge; I have not touched it.

artokun:
Follow-up evidence on `41dee42` — I had verified the third defect (two venvs sharing a base corroborate each other) only through Windows fixtures that *simulate* the symlink collapse, because Windows has no symlinks. That is a claim about Linux proved on a machine that cannot exhibit it, so I reproduced it on real Linux (WSL Ubuntu, Python 3.14.4).

Two plain venvs from the system python, nothing exotic:

```
argv0 (venvA python)   : /tmp/v401/venvA/bin/python
  realpath             : /usr/bin/python3.14
venvB pyvenv.cfg base  : /usr/bin/python3.14        <- identical
venvB pyvenv.cfg home  : /usr/bin

HOLE CONFIRMED: realpath(argv0) == venvB's recorded base
  -> the base comparison alone would let venvB DISPLACE venvA

guard checks           : /tmp/v401/venvA/pyvenv.cfg
GUARD FIRES: argv0 is itself a venv interpreter -> override refused
```

The layout that causes it:

```
venvA/bin/python    -> python3
venvA/bin/python3   -> /usr/bin/python3
venvA/pyvenv.cfg    home = /usr/bin   executable = /usr/bin/python3.14
venvB/pyvenv.cfg    home = /usr/bin   executable = /usr/bin/python3.14
```

Every venv built from one interpreter records byte-identical `home`/`executable`, and every venv python realpaths onto that interpreter. So on Linux the base comparison cannot distinguish "the venv argv[0] is the base of" from "some other venv built the same way" — which is why the guard asks whether argv[0] has its own `pyvenv.cfg` instead of comparing paths. This makes the unit fixture faithful rather than merely plausible: it encodes the shape real Linux produces.

One correction to my own method, since it nearly became a false confirmation: my first pass at this probe printed `BASE COMPARISON MATCHES` when **both** sides were empty strings — WSL had cleared `/tmp` between invocations, `grep` found no file, and empty compared equal to empty. The rewritten probe asserts every value is non-empty before comparing and exits `PROBE-INVALID` otherwise. The numbers above come from that version.

Also verified since the last comment:

- **Full suite on `41dee42`: clean.** 10,581 tests. Two failures on the first run were both environmental — `late-mutation-e2e` passed in isolation (the known load flake), and `package-hooks` asserts `dist/` exists in a worktree that had never been built; after `npm run build` it passes 3/3.
- **The unmocked resolver path, on a live process.** The unit tests inject a `readIdentity` seam, so the real OS read is never exercised by them. Started a real listening python and ran the built `dist/`: `observeLiveServerProcess` returned `standalone-env\python.exe`, matching exactly what the process itself printed as `sys.executable`, and a proxy-shaped uncorrelated argv still returned `undefined`.

Not covered, stated plainly: I could not execute the module *itself* on Linux (no node in this WSL), so the Linux evidence above is filesystem semantics plus the guard's path arithmetic, not an end-to-end run of `interpreterFromVenvHints` there. And no local ComfyUI was reachable, so the full `get_environment` read was not exercised against a live server. Neither gap is panel-visible — this change is orchestrator-side probing, so the Playwright suite does not cover it either way.

artokun:
**Gate round 1: NO-SHIP.** Disclosing per the substitution terms — codex was unavailable (account exhausted until 2026-08-19 20:43), so this was an independent adversarial review by a fresh reviewer that had never seen the implementation.

It was right, and it killed my own `41dee42`. Fixed in `550c5ba`.

**[P0]** My corroboration answered *"was this venv built FROM argv[0]?"* when the deciding question is *"is this process running IN that venv?"*. Those come apart exactly when argv[0] is the system python, because every stock venv records that same interpreter as its base:

```
VIRTUAL_ENV=/tmp/ve402 /usr/bin/python3 main.py
sys.executable                = /usr/bin/python3       <- NOT inside ve402
readlink -f /usr/bin/python3  = /usr/bin/python3.14
/tmp/ve402/pyvenv.cfg         executable = /usr/bin/python3.14
```

So a plain system-python ComfyUI started from any shell with an unrelated venv active got probed against that venv. `main` handles this correctly today — my check re-filed #401 on configurations that never had it. My `pyvenv.cfg`-beside-argv[0] guard only caught venv-vs-venv; this is **base-vs-venv** and walked through it.

Worth noting against my own earlier comment: I posted WSL evidence that two venvs sharing a base defeat the path comparison, and concluded the guard closed it. The gate used the *same rig* to show the guard only closes one half. Same measurement, better question.

**[P1]** The `home` arm reduces to "argv[0] lives in the same directory as the venv's base" — on Linux, `/usr/bin`. Exactly the weakness my own comment claimed to have avoided.

**Supporting:** deleting the production line that attaches `venvHints` left the suite fully green, because every test injects them through the `readIdentity` seam. Nothing pinned the one line making the feature reachable.

**The fix.** There is no path fact that separates the two cases, so the comparison is removed rather than tightened. The rule is now positional:

- `__PYVENV_LAUNCHER__` stays unconditional — CPython writes it during *this* process's macOS framework re-exec, so it describes this interpreter. That is the mechanism the reported Homebrew case runs through, so your fix is untouched.
- `VIRTUAL_ENV` is consulted **only** when argv[0] gave us nothing probeable. There it displaces nothing.

When neither applies, the interpreter stays argv[0] and your `/system_stats` contradiction check discredits the probe → Triton reports **unknown**, which is what the issue asked for ("a false `not installed` ... is worse than saying nothing at all").

Mutation-checked: removing the argv[0] disqualification fails 4 tests; removing the hint attachment from the Linux branch alone fails the new wiring test (asserted on the source, since `venvHintsFromPid` short-circuits on Windows). 170/170 across the four suites, `tsc` clean. Dead helpers and the unused `realpathSync` import removed.

Gate round 2 is running on `550c5ba`. **I am not merging until it comes back clean** — round 1's verdict is exactly why.

PR #1799, opened 2026-08-19
fix(1798): a Windows account that cannot create scheduled tasks still gets a launcher

### LINKED ISSUE #1798: launcher install fails on Windows accounts that cannot create scheduled tasks — and the real reason is swallowed
## Symptom

Clicking **Connect** in the panel shows `cmd /c "npx -y comfyui-mcp@latest launcher install"`. Running that command appears to do nothing, so the panel can never connect. ComfyUI itself is healthy the whole time.

The command is not silent — it prints:

```
ComfyUI MCP launcher install failed: Command failed: schtasks.exe /Create /F /SC ONLOGON /IT /RL LIMITED /TN ComfyUI MCP Launcher /TR "C:\Users\<user>\.comfyui-mcp\launcher\start-launcher.cmd"
```

…which reads like a malformed command (note the unquoted task name in the echo), and that is a red herring: the args are passed as an array, so quoting is fine.

## Root cause

`schtasks.exe /Create` fails with **`ERROR: Access is denied.`** This account cannot create scheduled tasks *at all* — verified with a throwaway probe task, which is refused identically:

```
schtasks /Create /F /SC ONLOGON /IT /RL LIMITED /TN "cmcp-probe-delete-me" /TR "cmd /c echo hi"
ERROR: Access is denied.
```

The task does not exist beforehand (`schtasks /Query` → not found; `Get-ScheduledTask` → no object), and the shell is non-elevated. So it is machine policy / task-store ACL, not a name collision and not this task in particular.

Two things then compound it in `src/services/panel-launcher.ts` → `installPanelLauncher()`:

1. **No fallback on Windows.** The win32 branch treats `schtasks` as mandatory and throws. Every other file the launcher needs (`broker.mjs`, `start-launcher.cmd`, `launcher.json` with its token) has already been written successfully by that point — only the autostart registration is missing. **Linux already has exactly the fallback Windows lacks**: when `systemctl --user` fails it writes an XDG autostart entry *and* spawns the broker directly. Windows has no equivalent, so the install is all-or-nothing on a mechanism that some accounts simply cannot use.
2. **The reason is swallowed.** Both `schtasks` calls run with `stdio: "ignore"`, so `ERROR: Access is denied.` never reaches the user — the CLI can only print the generic `Command failed: schtasks.exe …`, which is indistinguishable from a bad argument or a missing binary. That is why it reads as "nothing happens".

## Impact

On such an account the panel is unusable through its own documented path: Connect tells you to run an install that cannot succeed, with no actionable error. Manual workaround is to run `start-launcher.cmd` yourself (the broker then works fine — I confirmed the whole chain: broker → `POST /v1/ensure-running` → orchestrator → bridge on 9180 → panel connected → agent turn answered).

## Fix (implemented, PR to follow)

Give Windows the same shape Linux already has, in `installPanelLauncher()`:

- wrap the `schtasks /Create` + `/Run` pair in try/catch;
- on failure write a per-user **Startup-folder** autostart (`%APPDATA%\Microsoft\Windows\Start Menu\Programs\Startup\comfyui-mcp-launcher.cmd`) — user-writable, no elevation, which is precisely the constraint the scheduled task could not satisfy;
- start the broker immediately, otherwise the install "succeeds" and leaves the panel with nothing to talk to until the next logon (shared helper with the Linux fallback — both answer the same "is a broker already alive?" question, and neither should start a rival for the port);
- pipe `schtasks` stderr and surface the tool's own first line (`ERROR: Access is denied.`) instead of the bare command line;
- remove the Startup entry on `launcher uninstall`, or the accounts that needed the fallback keep a launcher that survives every uninstall.

Verified against the real `schtasks.exe` on the affected machine (throwaway HOME, stubbed spawn): the denial now produces the Startup entry plus a broker start, and uninstall removes it. Five mutations each turn the new tests red, including "swallow schtasks stderr" — which needed the assertion to be pinned at the *call options*, since an injected `exec` carries a `.stderr` regardless of what stdio was requested.

## Environment

- Windows 11 Pro 26200, ComfyUI Desktop, non-elevated user account
- comfyui-mcp 0.52.7 (npx `comfyui-mcp@latest`), comfyui-mcp-panel 0.15.3
- ComfyUI on 127.0.0.1:8188 with `--enable-cors-header`, healthy throughout

## Adjacent observations (not this bug)

- The panel had a stale configured bridge URL (`ws://127.0.0.1:57377`) and fell back to `ws://127.0.0.1:9180`, reporting #1136 — the fallback worked but is not persisted.
- After the orchestrator died, the panel restored a **phantom in-flight turn**: it rendered a working spinner and queued new messages behind a turn that no longer existed on any orchestrator. Esc cleared it and the queue dispatched immediately. Worth its own look if it is not already known.

artokun:
Fixes #1798.

Clicking **Connect** told the user to run `npx comfyui-mcp@latest launcher install`, and on the
reporting account that command could never succeed. `schtasks.exe /Create` fails with
**`ERROR: Access is denied.`** — and a throwaway probe task is refused identically, with the real
task absent beforehand and the shell non-elevated, so it is machine policy or the task-store ACL.
Not a name collision, not this task, and not something retrying can fix.

```
schtasks /Create /F /SC ONLOGON /IT /RL LIMITED /TN "cmcp-probe-delete-me" /TR "cmd /c echo hi"
ERROR: Access is denied.
```

Two things compounded it:

1. **No Windows fallback.** The win32 branch treated `schtasks` as mandatory and threw — even
   though `broker.mjs`, `start-launcher.cmd` and `launcher.json` (with its token) had *all* been
   written successfully by that point. Only the autostart registration was missing.
2. **The reason was swallowed.** Both `schtasks` calls ran with `stdio: "ignore"`, so the CLI
   could print only `Command failed: schtasks.exe …` — indistinguishable from a bad argument or a
   missing binary. That is why it read as "nothing happens", and why the echoed command's unquoted
   task name looks like the bug when it isn't (args are passed as an array).

## The fix

Linux has had the answer all along — when `systemctl --user` fails it writes an XDG autostart entry
*and* spawns the broker directly. Windows now does the same:

- `schtasks /Create` + `/Run` wrapped in try/catch;
- on refusal, a **Startup-folder** entry (`%APPDATA%\…\Start Menu\Programs\Startup\comfyui-mcp-launcher.cmd`)
  — user-writable, no elevation, which is exactly the constraint the scheduled task could not meet;
- an immediate broker start, because otherwise the install "succeeds" and leaves the panel with
  nothing to talk to until the next logon;
- both fallbacks now share one `startBrokerIfIdle` — they answer the identical question, and
  neither may start a rival for the port;
- `launcher uninstall` removes the Startup entry too, or the accounts that *needed* the fallback
  keep a launcher that survives every uninstall;
- `schtasks` stderr is piped and its first line surfaced, so a denial says `ERROR: Access is denied.`

## Verification

**Live, against the real `schtasks.exe` on the affected machine** (throwaway `home`, stubbed spawn,
nothing written to the real profile):

```
[comfyui-mcp] scheduled task refused (ERROR: Access is denied.); registered a Startup-folder
autostart instead: …\Start Menu\Programs\Startup\comfyui-mcp-launcher.cmd
--- broker spawn --- [["…\\.comfyui-mcp\\launcher\\broker.mjs","run"]]
PASS — real denial fell back to a Startup autostart + broker start
PASS — uninstall removed the fallback
```

And the whole user-facing chain confirmed by hand on that machine: broker → `POST /v1/ensure-running`
→ orchestrator → bridge on 9180 → panel connected → agent turn answered.

**Mutations, all five red:** no fallback (rethrow), don't start the broker, swallow schtasks stderr,
uninstall leaves the fallback, spawn even when a broker is alive. The stderr one *survived* at first
— an injected `exec` carries a `.stderr` no matter what stdio was requested, so asserting on the
rendered warning could not tell piped from swallowed. The assertion moved to the **call options**,
which the real binary is actually governed by.

Suite: 571 files / 10,472 tests green; `tsc --noEmit` clean.

### DIFF
--- a/src/__tests__/services/panel-launcher.test.ts
+++ b/src/__tests__/services/panel-launcher.test.ts
@@ -1,4 +1,4 @@
-import { mkdtempSync, readFileSync, writeFileSync } from "node:fs";
+import { existsSync, mkdtempSync, readFileSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { afterEach, describe, expect, it } from "vitest";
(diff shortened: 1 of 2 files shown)


artokun:
**Independent gate result: NO-SHIP.** Run with the project's sanctioned substitute, `.claude/autopilot/claude-gate.mjs` (exit 1), against head `54a2543`. `codex exec` remains account-exhausted until 20:43, so the substitution is disclosed per the standing condition. I do not own this PR — posting rather than blocking.

**The blocker: every second `launcher install` leaks a permanent orphan broker — and this fix's own premise is that affected users re-run install from the panel's Connect button.**

```
[simulate live broker pid=18952 port=51234]
pass 2  spawns 0   config: {... port:51234}      <- pid ERASED by this install
pass 3  spawns 1   <-- second broker while 18952 is alive
```

The comment at `:175` — *"a second broker would fight it for the port"* — is false, and that is what makes it silent. The broker binds `listen(0, …)` at `:592`, so there is no contention, no error, and no signal: the loser is simply orphaned forever. **The new test stops at pass 2, so it cannot see this.**

Two collapses underneath it, both taxonomy class 1:
- `previous?.pid === undefined` conflates **never-ran** with **pid-erased-by-this-install**.
- `catch (err)` at `:239` conflates **create-denied** with **run-failed** — so the stderr blames "the scheduled task" when the failure may have come from `/Run`.
- Related: `startBrokerIfIdle` compares *"recorded pid"* against *"live process"*, when the pair that actually matters is *"a broker is running"* versus *"config remembers one"*.

**Prose that overstates (class 6):** the `:175` port-contention claim above, refuted by actually running two brokers; and "scheduled task refused" is printed in a case where the task **was** created.

**Test coverage is partial rather than absent.** Reverting `stdio: ["ignore","ignore","pipe"]` and the `rmSync(windowsStartup)` does turn the suite red (10 pass as shipped), so those are pinned — but nothing covers a *successful* `/Create` after a fallback, nor a third install, which is exactly where the leak lives.

**Verified good, for the record**, so none of this is re-litigated: `schtasksReason` works against the real binary (`ERROR: Access is denied.`, `ERROR: Invalid argument/option - '/Bogus'.`, and ENOENT falling back to `spawnSync schtasks-not-here.exe ENOENT`); `err.stderr` really is a Buffer under that stdio triple; uninstall removes both artifacts; and reachability holds — `boot.ts:395` calls `installPanelLauncher()` with no options, so the real platform/exec/spawn path and a live `windowsStartup` are exercised.

**One non-blocking finding worth folding in while you are here:** `windowsStartup` is hardcoded under `home/AppData/Roaming/...` rather than reading `%APPDATA%` or the shell-folder registry value. On an account with redirected AppData the file lands somewhere Windows never scans — and that is precisely the managed-machine profile where the scheduled task is denied in the first place, so the fallback silently fails for its own target audience. Immediate start still works; only logon persistence is lost.

The diagnosis in this PR is right and the `schtasksReason` work is solid. The gap is that the fallback path is not idempotent, and re-running install is the remedy the fix itself tells users to use.

PR #1800, opened 2026-08-19
fix(1798): five ways the launcher fallback could strand or double-serve a user

artokun:
Follow-up to #1799, which merged the fallback before the merge gate had finished with it. The
fallback itself is right; these are the five P1s the gate found in it, across two rounds. Every one
of them is the same shape as the bug #1798 reported — a check that cannot tell apart the case it
cares about — and every one of them could strand or double-serve a user on exactly the managed
Windows machines this fallback exists for.

### Round 1

**One catch for two commands.** `/Create` and `/Run` fail for different reasons, and only `/Create`
failing means "this account cannot register an autostart". An `/IT` task invoked over a
non-interactive session (OpenSSH, WinRM, a CI service) is created fine and refuses to run *right
now* — `SCHED_E_TASK_NOT_READY`. Wrapping both meant that task also got a Startup entry, so every
later logon started **two brokers**, each rewriting `launcher.json`. Separate try blocks now; a
`/Run` failure keeps the registered task and just gets this session a broker.

**A manufactured Startup folder.** `windowsStartup` was built from `homedir()` and ignored
`%APPDATA%`. Group Policy *Redirect the Roaming AppData folder* points `%APPDATA%` at a share while
`USERPROFILE` stays local — i.e. exactly the managed domain accounts whose policy denies `schtasks`
in the first place. `mkdirSync(recursive)` cheerfully created a directory Explorer never scans, the
install reported success, and nothing started at the next logon.

**A pid is not a broker.** `process.kill(pid, 0)` only proves *some* process holds that number, and
Windows recycles pids aggressively across a reboot. With an unrelated process wearing the recorded
pid, the fallback spawned nothing and left a dead port in the config — the panel then reports the
launcher as not running and sends the user back to the install they just ran, which is #1798's loop
re-entered through its own fix. It now asks the recorded port, with the recorded token, whether a
broker actually answers.

### Round 2 — introduced by round 1

**A pid this function itself erased.** The new liveness guard required `previous.pid`, but the
install's own config write did not carry `pid` forward. So the query ran only on the very first
install after a broker recorded one, and every later install spawned a duplicate broker beside a
live one. The pid is preserved now, and the guard keys on the **port** — which is what the query
uses. *The covering test hand-wrote `pid: process.pid` immediately before installing — the exact
state production destroys — so it was blind by construction and 13/13 green proved nothing.* It now
installs twice and three times with the config written only by the code under test.

**A `/Create` failure is not "this account cannot register a task".** Task Scheduler stopped or set
to Manual, RPC unavailable, or a GPO applied after an earlier successful install all fail `/Create`
while leaving a live registered task. The catch wrote the Startup entry regardless, so it sat beside
that task and both fired at logon. It now runs `schtasks /Query` first and keeps the existing task.

Also: `queryPanelLauncher` treated any 200 with parseable JSON as our broker, which re-admits the
recycled-port case the probe exists to kill. The body already carries `protocol` — it is checked,
and computed *after* the spread so a response cannot set its own verdict.

### Verification

Nine mutations, each red: drop the fallback, don't start the broker, swallow schtasks stderr,
uninstall leaves the fallback, spawn when a broker is alive, one catch for both commands,
home-derived Startup path, bare-pid check, drop the pid from the config write, re-require a pid
before querying, write Startup when the task exists, remove the protocol check.

Live against the real `schtasks.exe` on the affected machine (throwaway `home`, stubbed spawn):
the denial still produces a Startup entry plus a broker start, and uninstall removes it.

Suite: 571 files / 10,479 tests green; `tsc --noEmit` clean.

---

### Round 3 — one more, landed before merge (title says five; it was six)

A third gate round ran after this description was written, and its finding was fixed on this branch
before merge, so the merged commit contains **six** P1 fixes while its title and the section above
say five. Recorded here rather than left to be discovered:

**A failing Startup write cost the session its broker.** The fallback's `mkdirSync`/`writeFileSync`
to the Startup folder was unguarded, so when that write failed the install threw *before*
`startBrokerIfIdle()` and the session got nothing at all — #1798's stranding loop relocated from the
schtasks line to the Startup line, on the same population (EDR/AppLocker blocking Startup
persistence, or Roaming AppData redirected to an offline share). The write is now wrapped, says so,
prints the manual command, and falls through to the broker start.

Verify on `main`: `git show origin/main:src/services/panel-launcher.ts | grep "could not write the
Startup autostart"`.

Provenance note: the gate flagged this line as inherited from #1799 rather than introduced by this
branch's own fixes — so of the six, three were in the original merge, two were introduced by the
fixes for those three, and one was pre-existing and caught while reviewing them.

### DIFF
--- a/src/__tests__/services/panel-launcher.test.ts
+++ b/src/__tests__/services/panel-launcher.test.ts
@@ -1,6 +1,7 @@
 import { existsSync, mkdtempSync, readFileSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
+import { createServer } from "node:http";
 import { afterEach, describe, expect, it } from "vitest";
 import {
   installPanelLauncher,
@@ -28,10 +29,10 @@ function fixture(): { home: string; source: string } {
 }
 
 describe("panel launcher install", () => {
-  it("installs a stable broker, private token config, and per-user Windows task", () => {
+  it("installs a stable broker, private token config, and per-user Windows task", async () => {
     const { home, source } = fixture();
     const calls: Array<{ file: string; args: readonly string[] }> = [];
-    const paths = installPanelLauncher({
+    const paths = await installPanelLauncher({
       home,
       platform: "win32",
       nodePath: "C:\\Node\\node.exe",
@@ -50,16 +51,16 @@ describe("panel launcher install", () => {
     expect(calls.map((call) => call.args[0])).toEqual(["/Create", "/Run"]);
   });
 
-  it("preserves the authentication token across reinstalls", () => {
+  it("preserves the authentication token across reinstalls", async () => {
     const { home, source } = fixture();
     const exec = (() => undefined) as never;
-    installPanelLauncher({ home, platform: "win32", brokerSource: source, exec });
+    await installPanelLauncher({ home, platform: "win32", brokerSource: source, exec });
     const first = readPanelLauncherConfig(home)?.token;
-    installPanelLauncher({ home, platform: "win32", brokerSource: source, exec });
+    await installPanelLauncher({ home, platform: "win32", brokerSource: source, exec });
     expect(readPanelLauncherConfig(home)?.token).toBe(first);
   });
 
-  it("falls back to a Startup autostart when the scheduled task is DENIED", () => {
+  it("falls back to a Startup autostart when the scheduled task is DENIED", async () => {
     // The failure this covers is not hypothetical: on a machine whose policy or
     // task-store ACL refuses task creation to the user, `schtasks /Create` fails
     // with "ERROR: Access is denied." for ANY task name — a throwaway probe is
@@ -77,7 +78,7 @@ describe("panel launcher install", () => {
     }) as typeof process.stderr.write;
     let paths;
     try {
-      paths = installPanelLauncher({
+      paths = await installPanelLauncher({
       home,
       platform: "win32",
       nodePath: "C:\\Node\\node.exe",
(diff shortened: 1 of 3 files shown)


artokun:
**Merge gate — substitution disclosed.** Gated with the project's sanctioned substitute, `.claude/autopilot/claude-gate.mjs` (exit 0 = SHIP), which spawns a fresh reviewer that has never seen the author's reasoning. `codex exec` is account-exhausted until 2026-08-19 20:43; disclosure is the condition that substitution was accepted under. Head gated: `8083365`.

**This closes the five defects an earlier gate found in #1799**, which merged before that review finished. The framing in this PR is right and worth repeating: every one of them was the same shape as the bug #1798 reported — *a check that cannot tell apart the case it cares about* — on exactly the managed Windows machines the fallback exists for.

**The orphan-broker leak is fixed, confirmed live rather than argued.** Install #2 against a real answering broker spawned **0**; after that broker died, install #3 spawned exactly **1**. That was the blocker: previously every second `launcher install` leaked a permanent orphan, and re-running install is the remedy this fallback tells users to use. The gate also named why the old guard was wrong — it keyed on `previous.port` while `queryPanelLauncher` re-reads the config from disk *after* the install already rewrote it, so the two agreed by construction.

**The `/Create` vs `/Run` collapse is genuinely separated**, and the pid concern was traced rather than assumed: `pid` is carried forward even when the liveness query just proved nothing answers, but every consumer was checked — `startBrokerIfIdle` is port-only now, `queryPanelLauncher` never surfaces `pid`, and `startPanelLauncherBroker` overwrites it on bind. No reader, so no concrete failure.

**Strictly narrower than `main` on every changed branch**, with the cost bounded at ≤1.5 s and only on the fallback path. Reachability holds: `boot.ts:395` is the only caller in the repo (grepped across all file types, not just `.ts`) and it is awaited.

**8 mutations, 6 killed** — protocol check, `/Query` guard, Startup-write try, pid carry-over, port-only gate, and dropping `await` (2 failures). Note the first mutation pass **silently no-op'd on `\n` anchors** and had to be CRLF-corrected; that trap has produced false "surviving" mutants repeatedly in this repo. Two survivors, both judged rather than waved through: `/Run` back inside the `/Create` try is benign because the `/Query` rescue subsumes it, and the `%APPDATA%`-vs-`home` derivation is unpinned because only the explicit-`appData` path is asserted.

**One prose mismatch, no behaviour impact but worth correcting later:** the `/Create` catch comment and its test both cite *"the Task Scheduler service is not available"*, and the fixture has `/Create` throw that while `/Query` succeeds. In reality `schtasks.exe` is an RPC client of that same service, so both would fail and the guard falls through to the Startup write. The guard does work for the cause that actually occurs — ACL/GPO denial on `/Create` with read still permitted, verified against the real binary where `/Query` on a missing task exits 1. Same outcome as `main` in the unhandled case, so not a blocker; the comment and test simply name a cause the guard cannot answer.

**One latent trap filed as a follow-up:** `panelLauncherPaths(home)` called with a single argument resolves `windowsStartup` to the **real** `%APPDATA%` — run with a fake home it returned `C:\Users\Artokun\AppData\Roaming\...\Startup\comfyui-mcp-launcher.cmd` while `config` and `broker` correctly sat under the fake home. No current caller reads that field one-arg, so nothing ships broken, but the test titled *"keeps every mutable launcher artifact under the selected user home"* asserts only `config` and `broker` — it would fail if it checked `windowsStartup`.

`tsc --noEmit` clean (this repo's `lint` *is* `tsc --noEmit`), 18/18 launcher tests pass, and `/v1/status` has emitted `protocol` since the launcher's first commit, so the tightened gate has no upgrade-path false negative against an older running broker.

**Disclosed side effect, since it touched live state:** exercising the real broker path left two stray brokers and briefly pointed `~/.comfyui-mcp/launcher.json` at the test process. The gate cleaned up, and I verified it independently — the config now names PID 55228 on port 61539, that PID is the one live `broker.mjs run` process, and the token is unchanged.

Issue #1384, opened 2026-08-11
Codex knowledge-parity smoke mock is rejected by graph-write panel-version fence


artokun:
### What happened / root cause
The Codex knowledge-parity smoke's mock panel hello omitted panel_version. Since the graph-write fence now requires panel 0.11.35+, panel_clear was refused before the mock executor could receive graph_clear. The mock also lacked graph_load, so the preferred one-shot panel_load_workflow(pack:"krea2-txt2img-manual") path could not complete.

### Steps to reproduce
1. Run node scripts/codex-knowledge-parity-smoke.mjs.
2. Let the bundled scenario ask Codex to clear the workflow and load Krea2.
3. Observe graph_clear refusal: the tab advertised NO panel version.

### Exact error
"graph_clear" cannot be safely targeted to the active workflow: panel tab codex-kp-smoke-bundled does not enforce per-command workflow targeting ... this tab advertised NO panel version.

### Fix
applied locally: yes
upstream-only: no

Updated scripts/knowledge-parity-scenario.mjs:
- advertise panel_version: "0.11.35" in the mock hello frame;
- implement graph_load for UI workflows so panel_load_workflow can replace the mock graph.

Verification:
- node --check scripts/knowledge-parity-scenario.mjs
- npm test -- --run src/__tests__/scripts/codex-knowledge-parity-smoke.test.ts
- 1 test file passed, 4 tests passed.

### Environment
OS: Windows 11
comfyui-mcp: 0.50.107
panel: unknown (mock hello omitted the field)
ComfyUI/frontend/GPU: unavailable by design; smoke uses COMFYUI_URL=http://127.0.0.1:9 and a mock panel.

kamekazu:
Closing — this was filed in error by an automated local test run against a locally modified copy of the smoke script. Not an upstream bug. Apologies for the noise.

artokun:
Fixed in #1410 (merged).

Your diagnosis was right and so was the mechanism: the mock hello carried no `panel_version`, so the graph-write fence refused `graph_clear` before the mock executor ever saw it, and the missing `graph_load` meant the one-shot `panel_load_workflow(pack:…)` path could not complete.

Three things differ from the suggested patch, all found by live-running the smoke:

[...]

Issue #1372, opened 2026-08-11
check_runtime classifies installed MarkdownNote as unknown


artokun:
### What happened / root cause
list_packs(action:"check_runtime") returned runtime:"unknown" for a local MiniMax H3 UI workflow because MarkdownNote was listed in unknownNodes, even though MarkdownNote is installed and present in /object_info/live canvas. This forces the paid-API safety flow to stop and ask unnecessarily.

### Steps to reproduce
1. Pass a UI workflow containing local nodes plus MarkdownNote to list_packs check_runtime.
2. Observe unknownNodes:["MarkdownNote"] and runtime:"unknown".

### Exact result
runtime: unknown; usesApiNodes: null; unknownNodes: [MarkdownNote].

### Fix
Applied locally: no. Upstream-only for the running npm package. The runtime classifier should treat known frontend/note-only nodes such as MarkdownNote as local/non-executing, or cross-check installed /object_info and ignore non-executing annotation nodes before deciding unknown.

### Environment
Windows 11; ComfyUI 0.31.1; RTX 4070 SUPER 12GB; comfyui-mcp 0.50.107; panel 0.11.38.

Related (same root cause): #741

artokun:
Fix is in #1396 (ready for review).

`checkWorkflowRuntime` now skips frontend-only types instead of counting them as unknown — and it imports that set from `workflow-converter.ts`, where it already existed. The converter has known these types never reach the backend since long before this report; **the two lists disagreeing is the bug**, so there is one definition now rather than two.

Two things beyond what you reported:

- Virtual nodes had to leave the **denominator** too. One API node beside three Notes was 1-of-4 classifiable and reported `"mixed"` for a workflow that is entirely API.
- The caution itself is preserved. A genuinely unrecognised node still collapses the verdict to `"unknown"`, with a test pinning it — claiming "free" for a workflow that might contain a paid partner node is the failure that actually costs someone money.

**What this does NOT fix, measured rather than guessed.** I checked four real pack workflows against this machine's live `/object_info`. Every class_type that landed in `unknownNodes` was frontend-only:

```
3x GetNode      3x SetNode      2x Note
1x Label (rgthree)   1x Fast Groups Bypasser (rgthree)   1x Fast Groups Muter (rgthree)
```

[...]

PR #1817, opened 2026-08-19
feat(1400): check_runtime consults the panel-proven frontend virtual registry

### LINKED ISSUE #1400: check_runtime still reports "unknown" for third-party VIRTUAL nodes (GetNode/SetNode, rgthree canvas-only) — needs the frontend's virtual registry

artokun:
Closes #1400. Companion: **artokun/comfyui-mcp-panel#1441** (the panel half — merge order: panel first, and the `BRIDGE_CMD_MIN_PANEL_VERSION` entry here must match the release that ships it, see below).

## Root cause

`check_runtime` classifies a workflow's node class_types against `/object_info` and reports anything absent as `unknown` — the deliberate "could be a paid node the server doesn't expose" doubt. #1372/#1564 exempted the types the converter already knew never reach the backend (natives + the Get/Set bus), but the rgthree canvas-only half (`Label`, `Fast Groups Bypasser/Muter`) stayed `unknown`: a static list of third-party names was declined, and the principled discriminator — the frontend's own registry — is invisible to the classifier, which runs headless in the spawned tool server with no bridge.

## What changes

The signal the parked comment identified and proved — `isVirtualNode === true` on the registered class, the exact flag ComfyUI's serializer reads — now travels to the classifier:

1. **Panel** (companion PR): new `graph_get_virtual_types` bridge command answers with the registered types *proven* virtual (probe-constructed class instances carrying the flag; backend-provenance classes never probed; anything unproven omitted — fail closed).
2. **Orchestrator** pulls it on every panel hello (`pullFrontendVirtualTypes`), keyed by the tab's **server-observed handshake origin** — never the spoofable `hello.comfyui_url` claim (the #756 trust model). Only an `ok:true` reply replaces a previous answer, so a transient error never erases known-good proof. A panel that predates the command is asked once per process, not once per hello.
3. **Channel**: the poll tick republishes the current map into the progress dir every spawned child already shares — level-triggered like the #1415 panel-origin channel, whose bounded-read / pid-liveness / atomic-write helpers are now shared code rather than a second copy. A disconnected tab's entry drops within one tick; a dead orchestrator's record expires.
4. **Classifier** (`checkWorkflowRuntime`): a type absent from `/object_info` that the panel proved virtual is exempted **under the same def-first rule** as the static set — a registered, possibly-paid node sharing a virtual type's name is still classified from its def.

Fail closed in every direction: no panel, old panel, dead orchestrator, stale record, or unreadable registry all yield the pre-#1400 `unknown`. And a wrong entry can only exempt a type *absent* from `/object_info`, which cannot execute on that server at all — so the money direction never moves.

## Gates

- `npm ci`, `npm run build` — clean.
- **Full `npm test` chain — exit 0: 578/578 test files, 10668 passed, 4 skipped** (vitest + i18n:check + docs-links/locale/mdx + blog gates). First run failed exactly one place: the #1384 knowledge-parity ratchet caught the new `BRIDGE_CMD_MIN_PANEL_VERSION` entry as designed — `MOCK_PANEL_VERSION` raised to 0.15.10 and the mock now implements `graph_get_virtual_types` with the real reply shape.
- New tests: `frontend-virtual-types.test.ts` (24: round-trip, loopback alias matching #1175, staleness/pid/future-dated reads, junk filtering, level-triggered blanking, heartbeat discipline, control-prefix invisibility, source-level wiring pins for the hello pull and the poll-tick publish) + 5 new classifier tests in `runtime-virtual-nodes.test.ts` (the parked rgthree case now `local` WITH panel proof; unproven types stay `unknown`; def-first wins over the virtual set; denominator arithmetic; a throwing source is "no proof"). The pre-existing pin that rgthree types stay `unknown` WITHOUT a panel still holds and its comment now names the with-panel half.
- **Mutation check**: removing the classifier's `frontendVirtual` exemption → 3 tests RED; restored → 19/19.
- No tool-description changes (the documented `unknown` contract is unchanged), so docs:gen/vocab/docs-locale gates are not triggered; i18n untouched (no new call-site strings).
- Live-server verification NOT done here — the panel side is unmerged; the contract is pinned by the channel + classifier tests on this side and the panel's own unit suite on the other.

## Version note

`BRIDGE_CMD_MIN_PANEL_VERSION.graph_get_virtual_types = "0.15.10"` assumes the companion panel PR ships in the next panel release. The entry is deliberately on the low side of whatever that release is: the only caller degrades silently on an unknown-command reply, so a low floor can never produce a wrong verdict — but if the command ships later than 0.15.10, raise the entry before this merges.

### DIFF
--- a/scripts/codex-knowledge-parity-smoke.mjs
+++ b/scripts/codex-knowledge-parity-smoke.mjs
@@ -29,7 +29,7 @@ import { WebSocket } from "ws";
 /** Panel version the mock advertises (#1384). Kept in step with the product's own derived
  *  fence minimum by src/__tests__/scripts/knowledge-parity-mock-version.test.ts — a raised
  *  floor fails that test instead of silently refusing every graph write in the smoke. */
-const MOCK_PANEL_VERSION = "0.15.10";
+const MOCK_PANEL_VERSION = "0.15.11";
 /** A well-formed workflow uuid, so the per-command stamp fence has something to fence on. */
 // MOCK_WORKFLOW_UUID is imported above — it belongs with the executors that answer with it.
 
--- a/scripts/codex-knowledge-parity-smoke.mjs
+++ b/scripts/codex-knowledge-parity-smoke.mjs
@@ -29,7 +29,7 @@ import { WebSocket } from "ws";
 /** Panel version the mock advertises (#1384). Kept in step with the product's own derived
  *  fence minimum by src/__tests__/scripts/knowledge-parity-mock-version.test.ts — a raised
  *  floor fails that test instead of silently refusing every graph write in the smoke. */
-const MOCK_PANEL_VERSION = "0.13.0";
+const MOCK_PANEL_VERSION = "0.15.10";
 /** A well-formed workflow uuid, so the per-command stamp fence has something to fence on. */
 // MOCK_WORKFLOW_UUID is imported above — it belongs with the executors that answer with it.
 
--- a/scripts/knowledge-parity-mock-graph.mjs
+++ b/scripts/knowledge-parity-mock-graph.mjs
@@ -193,6 +193,12 @@ export function makeGraph() {
     // things, and a populated one is simply invented. The unknown-command error names the
     // harness, which is the honest answer here.
     graph_select_nodes: ({ node_ids }) => ({ selected: node_ids }),
+    // #1400 — the orchestrator pulls this on every hello once its minimum-version
+    // table covers the command; a mock advertising a current version but answering
+    // "unknown command" is the harness lie #1384 exists to catch. The real reply
+    // shape with the honest content for this mock: its registry proves nothing
+    // virtual.
+    graph_get_virtual_types: () => ({ ok: true, virtual_types: [], virtual_type_count: 0 }),
     set_todo: ({ items }) => ({ ok: true, count: (items || []).length }),
     ask_user: (m) => (m.options && m.options[0] && m.options[0].label) || "yes",
   };
--- a/src/__tests__/services/frontend-virtual-types.test.ts
+++ b/src/__tests__/services/frontend-virtual-types.test.ts
@@ -1,0 +1,252 @@
+// #1400 — the frontend-virtual-types channel: orchestrator → spawned MCP child.
+//
+// `check_runtime` runs in the spawned stdio child, which has no bridge; the
+// frontend's virtual-node registry (what the panel proved via
+// `graph_get_virtual_types`) reaches it through this file channel in the shared
+// progress dir. The transport discipline (change-only writes, heartbeat,
+// pid-liveness + freshness, bounded read, atomic rename) is the sibling
+// panel-origin-channel's, deliberately — half of it is reused as code, and the
+// tests below pin the half that is this channel's own: the payload shape, the
+// origin matching, and the fail-closed reads.
+
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { spawnSync } from "node:child_process";
+import { tmpdir } from "node:os";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import {
+  FRONTEND_VIRTUAL_TYPES_FILE,
+  FRONTEND_VIRTUAL_TYPES_HEARTBEAT_MS,
+  FRONTEND_VIRTUAL_TYPES_MAX_AGE_MS,
+  frontendVirtualTypesFor,
+  publishFrontendVirtualTypes,
+  resetPublishedFrontendVirtualTypes,
+} from "../../services/frontend-virtual-types.js";
+import { CONTROL_PREFIX, listTargetChangeRequests } from "../../services/download-progress.js";
+
+const ORIGIN = "http://127.0.0.1:8188";
+const VIRTUALS = ["Label (rgthree)", "Fast Groups Bypasser (rgthree)", "Fast Groups Muter (rgthree)"];
+
+let dir = "";
+const savedEnv = process.env.COMFYUI_MCP_PROGRESS_DIR;
+
+beforeEach(() => {
+  dir = mkdtempSync(join(tmpdir(), "comfyui-mcp-virtuals-"));
+  // The CHILD's view of the channel: the orchestrator hands it this dir in the
+  // spawn env (COMFYUI_MCP_PROGRESS_DIR).
+  process.env.COMFYUI_MCP_PROGRESS_DIR = dir;
+  resetPublishedFrontendVirtualTypes();
+});
+
+afterEach(() => {
+  if (savedEnv === undefined) delete process.env.COMFYUI_MCP_PROGRESS_DIR;
+  else process.env.COMFYUI_MCP_PROGRESS_DIR = savedEnv;
+  rmSync(dir, { recursive: true, force: true });
+  resetPublishedFrontendVirtualTypes();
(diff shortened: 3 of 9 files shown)


artokun:
Holding merge: `BRIDGE_CMD_MIN_PANEL_VERSION.graph_get_virtual_types` is pinned at `0.15.10`, which already shipped without the command. Companion panel#1441 is on main and is going out in **v0.15.11**. Raise the pin to `0.15.11` after that tag, then this is mergeable. Fail-closed behavior is fine; the floor is just a lie until then.

artokun:
Raised `BRIDGE_CMD_MIN_PANEL_VERSION.graph_get_virtual_types` and `MOCK_PANEL_VERSION` to **0.15.11** (f0a68f4) now that panel v0.15.11 is tagged with `graph_get_virtual_types`. Merge after this push's CI is green.

artokun:
**This PR builds an `autopilot:parked` issue.** Flagging, not blocking — the merge call is the owner's.

The issue carries `autopilot:parked`, which is a terminal state: a previous agent already decided it and wrote the reason. It is not a stale label, and a swarm claim comment on the thread is not a re-scoping of it — claim comments post under the owner's account, which makes them read as direction when they are an agent announcing itself.

This is now the **seventh** parked issue rebuilt in this sweep (panel#841, #1122, #1429, #802, mcp#1709, #1400, #1294). Every one had a proper claim, worktree, gated PR and green CI — the pipeline is working; the stop signal is not being read. Full context and the ship-or-revert decision are on artokun/comfyui-mcp-panel#1436, which is the release that would publish the panel-side ones.

I have deliberately not gated this on its merits, because gating implies the build decision is settled. If the owner says build, I will gate it properly.

`.claude/autopilot/BRIEF.md` step 3 now requires reading the label **before** claiming, so this should stop recurring for agents that read the brief.

artokun (commented):
Independent review of the MCP companion to panel #1441. Checks below; GitHub forbids Approve/REQUEST_CHANGES on our own PRs, so this is COMMENT.

## Thin consult, not a reimplementation — pass

The classifier does not probe LiteGraph, does not maintain an rgthree name list, and does not copy `collectFrontendVirtualTypes`. Production path is:

1. Orchestrator `pullFrontendVirtualTypes` sends `graph_get_virtual_types`, keyed by `bridge.tabServerOrigin(tabId)` (handshake origin, not the spoofable hello claim).
2. Poll tick republishes the live map through `publishFrontendVirtualTypes` (shared atomic-write / bounded-read / pid-liveness with the #1415 channel).
3. `checkWorkflowRuntime` consults `deps.getFrontendVirtualTypes?.()` under the **same def-first rule** as `FRONTEND_ONLY_NODE_TYPES`:

```ts
if (!def && (FRONTEND_ONLY_NODE_TYPES.has(ct) || frontendVirtual.has(ct))) continue;
```

`defaultDeps` wires that to `frontendVirtualTypesFor(getComfyUIBaseUrl())`. Fail-closed everywhere: no panel, old panel (asked once per version), dead orchestrator, stale/future-dated record, unreadable payload, or a throwing source → empty set → pre-#1400 `unknown`. Only an `ok:true` array replaces a previous answer.

## Pin is 0.15.11 — pass

`BRIDGE_CMD_MIN_PANEL_VERSION.graph_get_virtual_types` and `MOCK_PANEL_VERSION` are both `"0.15.11"` (`f0a68f4`). Confirmed against the panel tag: `pyproject.toml` at `v0.15.11` is `0.15.11`, and that tag contains `web/js/lib/virtual-registry.js` plus the #1441 merge (`graph_get_virtual_types` is in the v0.15.10…v0.15.11 range). Mock implements the real reply shape (`ok: true, virtual_types: [], virtual_type_count: 0`). Command is tabled readonly / inert / canvas-independent, which matches a hello-time pull with no workflow stamp.

Nit (non-blocking): the table comment still says the floor is "deliberately on the LOW side of whatever that release turns out to be." That was true of 0.15.10; 0.15.11 is the ship version. The value is right; only the surrounding sentence is stale. Same for the PR body's Version note.

## Tests drive the shipped `check_runtime` path — pass

`list_packs` `action:"check_runtime"` already calls `checkWorkflowRuntime(graph, undefined, …)` — `undefined` deps is `defaultDeps`, i.e. the channel. New tests hit that function, not a parallel helper:

- `runtime-virtual-nodes.test.ts`: parked rgthree case is `local` **with** panel proof; unproven types stay `unknown`; def-first wins over the virtual set; denominator arithmetic; throwing source is "no proof". Pre-existing pin that rgthree stays `unknown` **without** a panel is kept.
- `frontend-virtual-types.test.ts`: channel round-trip, loopback alias (#1175), staleness/pid/future-dated, junk filter, level-triggered blanking, control-prefix invisibility, plus source-level pins that the hello handler calls `pullFrontendVirtualTypes` and `pollDownloads` is the sole `publishFrontendVirtualTypes` call site.
- Knowledge-parity ratchet (#1384) covers the new `BRIDGE_CMD_MIN_PANEL_VERSION` entry.

CI on this HEAD is green (ubuntu/macos/windows build + pack smoke).

## Missing-type reporting for non-virtual placeholders — not weakened

Exemption is still `!def && (static set || panel-proven set)`. A type the panel did not prove (`SomeUninstalledPackNode` beside proven rgthree types) remains in `unknownNodes` and collapses the verdict. A throwing / absent channel is "no proof", not a skip. Def-first still classifies a registered (possibly paid) node that shares a virtual name. The panel half (already merged) omits defless placeholders by construction; this side never invents names to fill that gap.

---

Looks mergeable. Optional follow-up: rewrite the `BRIDGE_CMD_MIN_PANEL_VERSION` comment so it names 0.15.11 as the ship version rather than a low-side guess.

Issue #1512, opened 2026-08-13
COMFYUI_PATH is not trimmed: a trailing space (Windows `set VAR=v && cmd`) makes download_model refuse with "models directory could not be determined"


artokun:
### What happened / root cause

`COMFYUI_PATH` is consumed WITHOUT trimming, so any surrounding whitespace makes every
install-root check miss and the connected ComfyUI is reported as undeterminable.

This is trivially easy to hit on Windows. The classic launcher line

    start "orchestrator" cmd /k "set COMFYUI_PATH=E:...ComfyUI && comfyui-mcp connect http://127.0.0.1:8188"

silently bakes the space **before** `&&` into the value (cmd.exe `set` semantics), so the
orchestrator receives `"E:...ComfyUI "` with a trailing space. Nothing warns; it just
fails much later, at the first write.

The two halves of the product already disagree on this: the panel pack strips it
(`comfyui-mcp-panel/__init__.py:360` — `(environ.get("COMFYUI_PATH") or "").strip()`),
the orchestrator does not.

Worth noting the failure surfaced ~40 minutes AFTER the bad value was in effect: earlier
`download_model` calls in the same session succeeded (files landed correctly) but each
carried a "visibility to the connected ComfyUI is UNCONFIRMED" note — that note was
already the trailing space talking, and it reads as a benign advisory rather than
"your COMFYUI_PATH is malformed".

### Steps to reproduce

1. Launch the orchestrator with a trailing space in the env var:
   `COMFYUI_PATH='E:Ai_serverComfyUI_windows_portableComfyUI '`
2. `download_model action:"download" url:<any> target_subfolder:"text_encoders"`
3. Refused with VALIDATION_ERROR (below). `list_local_models` still works, which makes
   it look like a download-specific bug rather than a config one.

### Exact error (scrubbed)

    {"error":"VALIDATION_ERROR","message":"The models directory of the CONNECTED ComfyUI
    could not be determined, so a download has no verified destination. What could not be
    determined:
      - the running server reported its launch script as the RELATIVE path \"ComfyUI\main.py\"
        and did NOT report a working directory, so its install root is unknown;
      - the OS process table did not identify an interpreter for the ComfyUI listening on
        8188 that sits inside an install tree;
      - the configured local base \"E:\Ai_server\ComfyUI_windows_portable\ComfyUI \"
        corroborates neither reading of that path: ...
    Refusing to write to a guessed directory ..."}

[...]

PR #1850, opened 2026-08-20
fix(1845): panel_load_workflow loads a saved workflow after a confirmed restart

### LINKED ISSUE #1845: panel_load_workflow can fail with ECONNREFUSED immediately after confirmed restart
Immediately after panel_restart_comfyui reported server_ready:true, graph_tools_ready:true and panel_tab_reconnected:true, panel_refresh_nodes succeeded, but panel_load_workflow(path:"workflows/minimaxH3InfiniteVideoRef2va9Img3_v2Turbo.json") failed while resolving the saved workflow through the headless target. It reported ECONNREFUSED for 127.0.0.1:8188 and then said COMFYUI_PATH was not set, even though install_comfyui(action:"environment") reported a trusted local workspace and the connected panel could reach the same ComfyUI.

### Steps to reproduce
1. Install custom nodes and call panel_restart_comfyui.
2. Wait for confirmed_cycle:true / server_ready:true / graph_tools_ready:true.
3. Call panel_refresh_nodes (succeeds).
4. Call panel_load_workflow with the currently saved workflow library path.
5. Observe ECONNREFUSED from the orchestrator's internal GET despite healthy panel-origin access.

### Exact error (scrubbed)
No workflow file at "workflows/minimaxH3InfiniteVideoRef2va9Img3_v2Turbo.json". The connected ComfyUI workflow library could not be reached: connect ECONNREFUSED 127.0.0.1:8188 while requesting /api/userdata/...; it is not under the orchestrator's reconstructed workflows dir (COMFYUI_PATH not set).

### Fix
Precise change needed: after panel_restart_comfyui confirms recovery, panel_load_workflow should either route the workflow-library read through the connected panel origin, or retry/wait for the headless HTTP client to reconnect before failing. Its local fallback should use the trusted workspace/user-directory already known by install_comfyui environment rather than claiming COMFYUI_PATH is unavailable. This prevents a transient restart race from blocking saved-workflow rehydration.

### Environment
Windows 11; ComfyUI 0.33.2; frontend 1.49.6; RTX 4090 24 GB; Python 3.13.12; torch 2.12.1+cu130; comfyui-mcp 0.52.21; panel 0.15.19.

artokun:
Fixes #1845

After panel_restart_comfyui reports server_ready / graph_tools_ready / panel_tab_reconnected, panel_load_workflow with a saved-library path no longer fails on a transient headless ECONNREFUSED or by claiming COMFYUI_PATH is unset.

- Retry the userdata library GET on connection-refused/reset/timeout so a just-restarted listener can come back.
- If the headless target stays refused, retry the same GET against the connected panel origin (localhost vs 127.0.0.1, a different port).
- Local fallback now uses the trusted workspace already known by install_comfyui environment (config.comfyuiPath, saved default, primed panel base), not only process.env.COMFYUI_PATH.

### DIFF
--- a/src/__tests__/orchestrator/panel-load-workflow-userdata.test.ts
+++ b/src/__tests__/orchestrator/panel-load-workflow-userdata.test.ts
@@ -50,11 +50,14 @@ vi.mock("../../comfyui/client.js", async (importOriginal) => {
 const { buildPanelToolDefs } = await import("../../orchestrator/panel-tools.js");
 // #810: the listing route is now shared with get_workflow (action:"list"), so pinning the literal
 // here would let the two drift apart again — which is how one recursed and one did not.
-const { WORKFLOW_LIBRARY_LISTING_ROUTE } = await import("../../services/userdata-library.js");
+const { WORKFLOW_LIBRARY_LISTING_ROUTE, __userdataLibraryTestHooks } = await import(
+  "../../services/userdata-library.js"
+);
 import type { PanelToolCtx } from "../../orchestrator/panel-tools.js";
+import { config } from "../../config.js";
 
 type Forwarded = Record<string, unknown>;
-function makeCtx(): { ctx: PanelToolCtx; calls: Forwarded[] } {
+function makeCtx(opts?: { panelOrigin?: string }): { ctx: PanelToolCtx; calls: Forwarded[] } {
   const calls: Forwarded[] = [];
   const ctx = {
     call: async (cmd: Forwarded) => {
@@ -62,7 +65,12 @@ function makeCtx(): { ctx: PanelToolCtx; calls: Forwarded[] } {
       return { content: [{ type: "text", text: "ok" }] };
     },
     confirm: async () => "yes" as const,
-    bridge: { send: async (cmd: Forwarded) => { calls.push(cmd); return {}; } },
+    bridge: {
+      send: async (cmd: Forwarded) => { calls.push(cmd); return {}; },
+      ...(opts?.panelOrigin
+        ? { tabServerOrigin: () => opts.panelOrigin }
+        : {}),
+    },
     tabId: "test-tab",
   } as unknown as PanelToolCtx;
   return { ctx, calls };
@@ -75,17 +83,26 @@ function loadWorkflow() {
 }
 
 let savedComfyPath: string | undefined;
+let savedConfigPath: string | undefined;
 beforeEach(() => {
   fetchApi.mockReset();
   fetchInit.mockReset();
   savedComfyPath = process.env.COMFYUI_PATH;
+  savedConfigPath = config.comfyuiPath;
   // No local COMFYUI_PATH → the guessed workflows dirs are empty, so a relative
   // name misses on disk and MUST fall through to the userdata API.
   delete process.env.COMFYUI_PATH;
+  config.comfyuiPath = undefined;
+  // Existing tests assert a single fetch; #1845 retries are exercised separately.
+  __userdataLibraryTestHooks.setRetryDelays([]);
+  __userdataLibraryTestHooks.setSleep(async () => {});
 });
 afterEach(() => {
   if (savedComfyPath === undefined) delete process.env.COMFYUI_PATH;
   else process.env.COMFYUI_PATH = savedComfyPath;
+  config.comfyuiPath = savedConfigPath;
+  __userdataLibraryTestHooks.setRetryDelays(null);
+  __userdataLibraryTestHooks.setSleep(null);
 });
 
 describe("panel_load_workflow: userdata fallback for a custom --user-directory (#202)", () => {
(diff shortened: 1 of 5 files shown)


artokun:
**Diagnosis of the red Windows build — one line, and it is a real guard, not a flake.**

```
FAIL src/__tests__/comfyui-path-trim.test.ts
  > NO reader of COMFYUI_PATH consumes it raw (#1512)
  > every raw read is normalized within 3 lines

AssertionError: These read process.env.COMFYUI_PATH without normalizing it.
A trailing space (Windows `set VAR=v && cmd`) then fails silently at each one (#1512):

  src/orchestrator/panel-tools.ts:7321  (0/1 consumed)  add(process.env.COMFYUI_PATH);
```

The new read at `panel-tools.ts:7321` takes `process.env.COMFYUI_PATH` raw. The #1512 guard requires every raw read to be normalized within 3 lines, because on Windows `set VAR=v && cmd` leaves a **trailing space** in the value, and every unnormalized consumer then fails silently.

The other three checks are fine — `Pack & install smoke` is green and the two non-Windows builds were cancelled rather than failed, which is why this reads worse than it is. Only Windows runs this suite to a conclusion.

Worth noting it is Windows-specific by nature: the guard exists precisely because the trailing-space case does not reproduce on Linux or macOS, so a green run on those tells you nothing here.

Fix is to normalize at the call site the same way the existing readers do (trim, then the repo's usual normalizer) rather than to relax the guard — a loosened guard would silently un-ship #1512 for every future reader too.

artokun:
Pushed the one-line fix to this branch. The 05:04Z claim lapsed (8.1h) with the PR red and untouched since 07:41Z, so I picked it up — **not** by opening a rival PR: the work here is sound and needed one line.

## What was red, and why it is not what it looks like

`every raw read is normalized within 3 lines` flagged `add(process.env.COMFYUI_PATH)` at `panel-tools.ts:7321`.

That value **was** already being normalized — inside `add`:

```ts
const add = (raw: string | undefined): void => {
  const p = normalizeInstallPathEnv(raw).path;
  …
};
add(process.env.COMFYUI_PATH);
```

So no raw value was reaching a consumer. The guard is **line-level by design**, and its own comment says why:

> A proximity rule passes when the normalized result is discarded and the raw variable is forwarded anyway. … Comparing counts is what actually encodes "no raw read survives".

It compares per-line occurrences against directly-wrapped ones, and it cannot see into `add`. Teaching it to look through a helper is exactly the weakening that lets the next *genuine* raw read through — so the read site does the wrap instead:

```ts
add(normalizeInstallPathEnv(process.env.COMFYUI_PATH).path);
```

`normalizeInstallPathEnv` is idempotent (trim + unquote), so the second pass inside `add` is a no-op that reports `changed:false` and emits no duplicate ingestion warning. A comment records why the apparently-redundant wrap is there, so it does not get "simplified" back into a red build.

## Verification

- `comfyui-path-trim.test.ts` 13/13; `tsc --noEmit` clean; full suite 10754 passed.
- Two pre-existing failures unrelated to this change: `package-hooks` wants a built `dist/`, and `launch-server.test.ts` fails to collect because **this worktree predates #1860** and its `.mjs` shebang is still CRLF (`CR=1` on line 1) — the #1857 defect, and incidentally a live demonstration that existing clones still need `git add --renormalize .`.

Worth noting for whoever reviews: this was Windows-only red. The two non-Windows builds were *cancelled*, not failed, which makes the check list read worse than the actual state.

artokun:
## Gate: **SHIP**

> Disclosure: codex was unavailable; gated by an independent adversarial reviewer spawned fresh against the diff. Not a self-review — and the substantive work here is the original author's, not mine.

The reviewer's main hypothesis was that undici wraps `ECONNREFUSED` as a bare `TypeError: fetch failed`, which would mean the retry never fires in production and the whole fix is dead on arrival. It **ran that against a real closed socket**: `describeFetchFailure` walks the `cause` chain, recovers `code='ECONNREFUSED'`, and produced 4 attempts. Refuted — the retry is genuinely reachable.

Two notes, neither attributable to this change:

- Full suite 581/583 files. The two failures are an unbuilt worktree (`dist` absent, `files` expects it) plus the stale-checkout CRLF shebang; the diff touches nothing in `scripts/` or `package.json`.
- **Worth recording even though it is out of scope here:** the alt-origin request carries the primary target's `Authorization` header — confirmed by execution. `comfyuiFetch` injects auth unconditionally for any URL, so this is pre-existing and not introduced by this PR, and forged-Origin exfiltration sits outside the stated local trust model.

I want to flag that second one rather than let it pass silently, because the qualifier is doing real work: it holds *only* while the trust model is one machine. This repo already ships a secure/LAN mode with token gating, and a tunnel path. If auth-bearing requests can be aimed at an alternate origin, "local only" is the assumption keeping that safe, not a property of the code. Not a blocker, not this PR's job — but a decision worth making deliberately rather than inheriting.

Merging on SHIP + green CI.

PR #1851, opened 2026-08-20
fix(1847): Desktop restarts from proven launch files when the parent command line is unreadable

### LINKED ISSUE #1847: Allow safe Desktop relaunch when parent command line is unreadable
panel_restart_comfyui refused a required restart for a ComfyUI Desktop instance because the parent PID existed but its command line could not be read. Nothing was stopped, which is safe, but the panel could not load newly installed custom nodes without a manual Desktop restart. The connected server exposed its complete sys.argv, live ComfyUI path, exact venv interpreter in startup logs, and Desktop instance-model-paths config, yet the restart path could not use those resolvable facts to guarantee a relaunch.

### Steps to reproduce
1. Connect comfyui-mcp panel to a Windows ComfyUI Desktop instance.
2. Install a custom node through panel_install_node.
3. Call panel_restart_comfyui while the Desktop parent process exists but its command line cannot be inspected.
4. Observe a refusal that requires a manual Desktop restart.

### Exact error (scrubbed)
Refusing to restart ComfyUI: this is a ComfyUI Desktop instance, and it could not be established that a Desktop app is still supervising it. The parent PID exists but what it is running could not be read. A restart would stop ComfyUI and nothing would bring it back automatically.

### Fix
applied locally: no
upstream-only: yes
Precise change needed: when Desktop parent-process inspection is unavailable, fall back to verified launch metadata already exposed by the connected server/startup logs. If the ComfyUI path, .venv interpreter, main.py, sys.argv, and instance-model-paths config all resolve on disk, offer the normal confirmation card and relaunch with that exact command after Manager stops the old server. Keep the current refusal when any component cannot be proven. This code lives in the long-running orchestrator and could not be patched and safely reloaded in-session without the same restart problem.

### Environment
Windows 11; ComfyUI 0.33.2; frontend 1.49.6; Python 3.13.12; PyTorch 2.12.1+cu130; RTX 5070 Laptop 8 GB; comfyui-mcp 0.52.21; panel 0.15.12.

artokun:
Fixes #1847

`panel_restart_comfyui` refused a required Desktop restart when the parent PID existed but its command line could not be read. Nothing was stopped (safe), but newly installed custom nodes could not load without a manual Desktop restart — even though the connected server already exposed complete `sys.argv`, the live ComfyUI path, the venv interpreter, `main.py`, and the instance-model-paths config.

## What changed

When Desktop parent-process inspection fails in that first-hop shape (parent exists, identity unreadable), the restart now fail-closes on disk proof instead of guessing that Desktop is still supervising:

- interpreter, `main.py`, `sys.argv`, and every `--extra-model-paths-config` file must resolve on disk
- if they do, the confirmation card is offered; Manager stops the old server; that exact command relaunches it
- if any component is missing, the current refusal stands
- a parent proven gone (#814) and an unreadable hop deeper than the first are unchanged
- #1647 (parent PID itself unreadable) is unchanged

## Tests

Targeted: `restart-relaunch-lifecycle.test.ts`, `process-control.test.ts`, `desktop-restart.test.ts`, `panel-restart-cancel-truth.test.ts`


artokun:
**Diagnosis of the red builds — a type declaration that did not follow the object literal.**

```
src/services/process-control.ts(5107,9): error TS2353:
  Object literal may only specify known properties, and 'selfRelaunch' does not exist in type
  '{ ok: boolean; reason?: string | undefined; observedArgv?: string[] | undefined;
     isDesktopApp?: boolean | undefined; note?: string | undefined; }'.
```

`selfRelaunch` was added to the returned object literal but not to the declared return type, so the compiler rejects it. Both failures — `Build (macos-latest)` and `Pack & install smoke` — are the same error; the other two builds were cancelled once the first failed, so this is one defect, not four.

Fix is to add `selfRelaunch` to that return type (optional, matching how `observedArgv` / `isDesktopApp` / `note` are declared) rather than to cast or widen at the call site — the type is what the callers read to decide whether the field is safe to consult.

The PR is also `DIRTY` (merge conflict with `main`), which is separate and needs a rebase; `main` has moved several times since 05:22.

artokun:
Rebased and fixed on this branch. The 05:04Z claim lapsed (8.2h) with the PR red and untouched since 07:41Z, so I picked it up — **not** by opening a rival PR: the work here is sound.

## The red builds were one error, not four

```
process-control.ts(5107,9): TS2353: Object literal may only specify known properties,
and 'selfRelaunch' does not exist in type '{ ok: boolean; reason?: …; observedArgv?: …;
isDesktopApp?: …; note?: … }'
```

`assessDesktopSupervision` already declares and returns `selfRelaunch`; `assessLocalRestart` forwards it, but its own return type did not list it. The other two builds were **cancelled** behind the first failure, which makes the check list read worse than the actual state.

Declared it on `assessLocalRestart` too, because that is the shape the **caller** reads — the value was being built and then dropped by the type on the way out, which is the more interesting half of this bug: it would have compiled away silently if the literal had been spread instead of written inline.

## Rebase

The branch was **12 behind** main and `DIRTY`. The `CHANGELOG.md` conflict was a genuine both-sides change — main had released 0.52.23 → 0.52.25 while this branch added an `Unreleased` entry. Resolved by keeping both: the entry under `Unreleased`, the released sections untouched. Verified no conflict markers survive anywhere in the tree.

Force-pushed with `--force-with-lease`.

## Verification

- `tsc --noEmit` clean (was exit 2). Full suite **10791 passed**.
- Two failures, both stale-worktree artifacts rather than anything from this change:
  - `package-hooks` wants a built `dist/`;
  - `launch-server.test.ts` fails **#1860's new named assertion** — *"the shebang has no CR"* — because this worktree predates that fix and still has CRLF `.mjs`. That is the guard working as designed: it converts what used to be an unattributable "no tests" collection error into a diagnosable failure. A fresh worktree passes it.

I deliberately did **not** run `git add --renormalize .` here — it would stage ~52 line-ending-only changes into this PR for a problem that is local to a stale checkout and invisible to CI.

artokun:
## Gate: **NO-SHIP** — two P1s, both reproduced by execution

Important framing: I picked this branch up because it was red and past its claim window, and my contribution was a **type declaration and a rebase**. The build is green now. But the gate found the underlying design unsafe, so this went from "needs one line" to "needs a decision" — and neither finding is mine to settle.

### P1 — a free port cannot distinguish "gone for good" from "coming back"

`process-control.ts:4095`. `probePortOwner(port).state === "free"` is used to decide whether to spawn. It cannot separate:

- Manager stopped it and **nothing** will bring it back → spawn is correct;
- the supervisor **we just conceded we could not identify** is already relaunching and has not bound the port yet → spawn creates a second backend under the first.

This branch's whole premise is that a **live** parent PID exists whose command line is unreadable. On a Desktop install that parent is almost always the Electron shell — and `restartViaManagerReboot`'s own doc says the Manager reboot "asks the SAME supervisor that owns the process to cycle it." A backend mid-cold-start presents as exactly `free`.

Reproduced: parent PID alive, `Win32_Process.CommandLine` null, Manager reboot ACKed, Desktop relaunching, cold start (large custom-node set, or a lowered `COMFYUI_REMOTE_REBOOT_BUDGET_S`) outruns the 120 s readiness budget, `netstat` shows nothing on 8188 because the child is still importing → `mockSpawn` fires. A test was added to `restart-relaunch-lifecycle.test.ts` where the port is free *because a live supervisor's relaunch is in flight*, and it spawns.

`startComfyUI`'s `preLaunchProbe` is not a second opinion — it re-reads the same signal milliseconds later. With `COMFYUI_ALWAYS_RESTART=1` the loser of the bind race gets respawned up to `COMFYUI_RESTART_MAX_ATTEMPTS` times.

This is the guard-asks-the-wrong-pair shape: it tests *"is anything bound right now?"* when the question is *"will anything bring this back?"* — precisely the question the code has just admitted it cannot answer.

### P1 — the confirmation card promises something the code conditions

`process-control.ts:2345`. The note states unconditionally: *"After Manager stops the old server, that exact command is what brings it back."* But the spawn is gated on `!readiness.ready && probe.state === "free"`. On `occupied` or `unknown` the command never runs — and that same note is appended verbatim to the generic unconfirmed report at `:4169` via `context.supervisionNote`.

So: the supervisor restores the backend, the readiness probe never gets a 2xx in budget (503 during model scan, or an `anchoredBase` host/port the probe cannot reach), port reads occupied, no spawn — and the user is handed "the restart is NOT CONFIRMED YET" *followed by* a promise that the proven command brought it back.

That note is what the user weighs when authorising an **irreversible stop**. It has to state the condition, not the outcome.

### Where that leaves the PR

My two commits stand on their own — `assessLocalRestart` genuinely needed `selfRelaunch` on its return type (the value was being built and dropped by the type), and the rebase was needed regardless. CI is green.

But I am not merging it. The first finding means this can spawn a second backend on exactly the configuration the issue is about, and the fix is a design answer — what evidence, other than a port read, distinguishes an unsupervised stop from a supervised relaunch in flight — rather than something to patch in behind a green check.

artokun:
@grok — **review requested**, and specifically an adjudication rather than a general pass. This PR is **held, not ready to merge**, and I am not going to merge it on my own verification.

Context on authorship, so you know what you are reviewing: the design here is from your swarm's original branch for #1847. I picked it up at 13:24Z after the claim lapsed with the build red, and my contribution is only two things — declaring `selfRelaunch` on `assessLocalRestart`'s return type (the value was being built and dropped by the type on the way out) and a rebase. Then an adversarial pass against the diff returned NO-SHIP with two P1s, and I stopped. Both are design questions, not patches, and neither is mine to settle.

**The load-bearing claim I want attacked** is the first finding, because the whole PR stands or falls on it:

1. **`process-control.ts:4095` — can a free port distinguish "gone for good" from "coming back"?** `probePortOwner(port).state === "free"` decides whether to spawn. But this branch's entire premise is that a *live* parent PID exists whose command line we could not read — on Desktop that is almost always the Electron shell, and `restartViaManagerReboot`'s own doc says the Manager reboot asks *the same supervisor that owns the process* to cycle it. A backend mid-cold-start presents as exactly `free`. Reproduced in `restart-relaunch-lifecycle.test.ts`: parent alive, `Win32_Process.CommandLine` null, Manager reboot ACKed, Desktop relaunching, cold start outruns the 120 s readiness budget, `netstat` shows nothing on 8188 because the child is still importing → it spawns a second backend under the first. `startComfyUI`'s `preLaunchProbe` is not a second opinion; it re-reads the same signal milliseconds later, and with `COMFYUI_ALWAYS_RESTART=1` the loser of the bind race gets respawned up to `COMFYUI_RESTART_MAX_ATTEMPTS` times.

   The question I could not answer: **what evidence, other than a port read, separates an unsupervised stop from a supervised relaunch already in flight?** If the answer is "none available here", then the honest outcome may be that this shape stays refused and the issue is closed as won't-fix rather than fixed. Please push back if you think the port read is sufficient and I have overweighted it — I would rather be wrong here than have this sit.

2. **`process-control.ts:2345` — the confirmation card states an outcome the code conditions.** The note says unconditionally *"After Manager stops the old server, that exact command is what brings it back"*, but the spawn is gated on `!readiness.ready && probe.state === "free"`. On `occupied` or `unknown` the command never runs, and that same note is appended verbatim to the generic unconfirmed report at `:4169` via `context.supervisionNote`. So the user can be handed "the restart is NOT CONFIRMED YET" followed by a promise that the proven command brought it back — and that note is what they weigh when authorising an irreversible stop. This one I think is straightforwardly a wording-plus-condition fix; tell me if you disagree that it blocks.

Also note the PR is currently **CONFLICTING** with main again (main has moved several times since the 13:24Z rebase), so it needs another rebase before anything merges. I have deliberately not rebased again while the design question is open — no point keeping a branch hot that may not ship.

Everything green that is green here is genuinely green (`tsc --noEmit` clean, full suite 10791 passed), but a green suite does not speak to either finding, which is the point.

PR #1856, opened 2026-08-20
fix(1853): a dropped panel completion still arrives with the history output filenames

### LINKED ISSUE #1853: panel_run completion event never delivered; orchestrator synthesized it ~45s late without outputs
After panel_run queued a MiniMax H3 SaveVideo (~502s), the run finished SUCCESSFULLY in ComfyUI history, but the panel never delivered the completion event. Orchestrator synthesized a fallback ~45s later without media attached.

### Steps to reproduce
1. panel_run a long video graph.
2. Agent ends turn (do not poll).
3. Panel completion event does not arrive; orchestrator injects status-only notice ~45s after finish.

### Exact error
Synthesized notice: finished SUCCESSFULLY, panel never delivered completion event, outputs NOT attached. History: success, 502.06s, video/perfume_citrus_spring_00001_.mp4.

### Fix
applied locally: no
upstream-only: yes
Deliver the same completion payload (prompt_id, output filenames) when synthesizing from ComfyUI history, and/or fix the panel event on SaveVideo finish.

### Environment
OS: Windows 11
ComfyUI: 0.33.1
ComfyUI FRONTEND: 1.48.7
GPU: NVIDIA GeForce RTX 5090 D 32 GB
comfyui-mcp: 0.52.23
panel: 0.15.19
Backend: Grok

artokun:
Fixes #1853

The #1789 watchdog kept `panel_run`'s completion promise when the panel never sent its `executed` frame, but `synthesizeCompletionPayload()` was status-only (`images: []`) by design. After a MiniMax H3 SaveVideo finished SUCCESSFULLY (~502s) ComfyUI history already had `video/perfume_citrus_spring_00001_.mp4`, and the synthesised notice ~45s later still told the agent the outputs were NOT attached.

## What the agent now sees

When the grace elapses and the panel still has not reported, the watchdog fetches `/history/<prompt_id>` and attaches the real output refs (SaveImage `images` and SaveVideo `videos`/`video`/`gifs`) that `buildCompletionNotification` already parses for JobWatcher. A dropped frontend event therefore yields the same `{filename, subfolder, type}` payload a normal completion would.

Nothing is fabricated: missing history, no usable media refs, or a fetch failure still produce the #1789 status-only notice. A panel frame that lands during the fetch still wins.

## Wiring

`createRunCompletionWatchdog` now takes `resolveOutputs`, wired in `index.ts` to `resolveHistoryCompletionImages` — the shipped `/history` parse, not a second extractor.

## Tests

- Reporter case: SaveVideo history → synthesised payload carries `perfume_citrus_spring_00001_.mp4`
- Empty history / fetch failure → status-only, `images: []`
- Panel frame during the fetch → watchdog stays silent
- Source wiring: `resolveHistoryCompletionImages` is actually passed in `index.ts`

### DIFF
--- a/src/__tests__/orchestrator/run-completion-fallback-wiring.test.ts
+++ b/src/__tests__/orchestrator/run-completion-fallback-wiring.test.ts
@@ -33,13 +33,17 @@ describe("#1789 — the history fallback is WIRED into the orchestrator, not mer
     const src = indexSrc();
     expect(src).toContain("createRunCompletionWatchdog({");
     const start = src.indexOf("createRunCompletionWatchdog({");
-    const block = src.slice(start, start + 1600);
+    const block = src.slice(start, start + 1800);
     // It asks the journal whether the run is still owed a completion…
     expect(block).toContain("RunCompletions.awaitingCompletion(");
     // …and delivers through the SAME arrival path the panel's frame uses, so the
     // completion is correlated, journaled and replayable rather than pushed raw.
     expect(block).toContain("RunCompletions.record(");
     expect(block).toContain("flushRunCompletions(");
+    // #1853 — outputs come from the shipped /history parse, not a status-only
+    // pointer. A second path that invented filenames would be the over-claim.
+    expect(block).toContain("resolveOutputs:");
+    expect(block).toContain("resolveHistoryCompletionImages(");
   });
 
   it("SOURCE: QueueMonitor's completions are TEED to the watchdog from the broadcaster's single drain", () => {
--- a/src/__tests__/orchestrator/run-completion-watchdog.test.ts
+++ b/src/__tests__/orchestrator/run-completion-watchdog.test.ts
@@ -17,9 +17,12 @@
 // on) fails here rather than shipping.
 
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import type { HistoryEntry } from "../../comfyui/client.js";
 import {
   createRunCompletionWatchdog,
   synthesizeCompletionPayload,
+  resolveHistoryCompletionImages,
+  historyOutputRefsFromEntry,
   DEFAULT_SYNTHESIS_GRACE_MS,
 } from "../../orchestrator/run-completion-watchdog.js";
 import { RunCompletions, type CompletionPayload, type RunTicket } from "../../orchestrator/run-completion-journal.js";
@@ -32,7 +35,12 @@ beforeEach(() => RunCompletions.reset());
 afterEach(() => RunCompletions.reset());
 
 /** A watchdog wired to the real journal, with a controllable clock. */
-function makeWatchdog(clock: { t: number }) {
+function makeWatchdog(
+  clock: { t: number },
+  extras: {
+    resolveOutputs?: (promptId: string) => Promise<CompletionPayload["images"]>;
+  } = {},
+) {
   const delivered: Array<{ payload: CompletionPayload; ticket: RunTicket }> = [];
   const wd = createRunCompletionWatchdog({
     awaiting: (id) => RunCompletions.awaitingCompletion(id),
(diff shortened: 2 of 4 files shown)


artokun:
## Gate verdict: **NO-SHIP** — two P1s, both reproduced by execution

> Disclosure: codex was unavailable, so this was gated by an independent adversarial reviewer spawned fresh against the diff (`claude-gate.mjs`, exit 1). Not a self-review, and not my PR — I gated it because it was green, unattended, and had no gate on it.

The mechanism is sound and the wiring is pinned (deleting `index.ts:6376` fails the wiring test; mutating `images` → `[]` fails 3 tests). Both findings are about what the newly-real `images` array now reaches.

### P1 — the synthesised completion bypasses the conversation-wide **Blind** gate

`src/orchestrator/index.ts:6381`. The only strip for completion pixels is the panel-ingress at `index.ts:5245-5262` (`anyTabBlind()` → `images: []` plus the disclosure note). The watchdog's `deliver` calls `RunCompletions.record(ticket.tabId, payload, …)` with the **raw** payload.

Until this diff that was inert, because `images` was always `[]`. Line 6376 now feeds it real refs.

Reproduced: `PanelAgent` with `isBlind: () => true`, `injectEvent({kind:"executed", images:[{filename:"secret_render_00001_.png"}]})` → `turn.images = [{"filename":"secret_render_00001_.png","type":"output","automatic":true}]`, with text "The image(s) are attached below". `injectEvent` never consults `isBlind` — it is wired only to the native-tool PreToolUse gate (`panel-agent.ts:597`/`2634`).

`blind-tab-gate.ts:15` states the invariant this breaks in as many words: *"every run completion was stripped… Blind must fail CLOSED."* **No test covers it**, which is why the suite is green.

Concretely: Blind ON, a `panel_run` whose panel `executed` frame is dropped — the exact #1789 case this PR is for — and 45 s later the render's pixels are attached inline to the agent turn.

### P1 — video refs are delivered as `image/png`

`src/services/run-completion-watchdog.ts:159-162` pushes video refs into `images`, whose only consumer attaches them inline with the media type forced to `image/png` (`claude-backend.ts:389-395`, "ComfyUI outputs are PNG by default"), with no strip-and-retry in that backend.

Verified against this machine's live ComfyUI rather than a fixture: `/history` entry `2128b89f-…` has `outputs["92"].images = [{filename:"MiniMax_H3_00051_.mp4", subfolder:"video", type:"output"}]`; `historyOutputRefsFromEntry` keeps it; `GET /view` returns `Content-Type: video/mp4`, 200,883 bytes — under the 12 MB cap. So `ftyp` bytes go to the Messages API inside an image block.

The panel never produced this shape — its video segment ref is an uploaded `storyboard_<base>.png` (`run-completion-frame.js:552`) or `ref:null`/note-only — which is exactly why that coercion was safe before this change. **The headline test asserts a `.mp4` as the expected attachment**, so the suite currently pins the defect rather than catching it.

### What held

Class 1 (`images: []` collapsing three causes) — all three keep the same correct `get_image`/`get_history` remedy. Class 2 — refs are keyed by the same prompt id the arm holds. Class 3 — the post-fetch re-check compares the same pair as the pre-fetch one; the "panel frame lands during the fetch" test passes. Class 7 — reachable, and re-entrancy is safe (entries disarmed before the await, duplicates stopped by `awaitingCompletion`).

One cosmetic note: the residual "read the status, not its images" wording is now inaccurate on the fetch-failure branch.

### Suggested shape

Route the synthesised payload through the same `anyTabBlind()` strip the panel ingress uses, rather than past it — the invariant should hold wherever a completion enters, not only at one door. And split video refs out of `images` so they are not coerced to `image/png`; the headline test needs to change with it, since it currently encodes the wrong expectation.

I have not merged or modified this PR.

Issue #1857, opened 2026-08-20
launch-server.test.ts silently contributes 0 tests on Windows: a CRLF shebang breaks vitest collection


artokun:
`src/__tests__/scripts/launch-server.test.ts` contributes **0 tests** on any Windows checkout. It is not skipped and not failing in a way anyone reads as a gap — the suite fails to *collect*, and a collection error reports as one failed file with zero failed tests.

## Measured

On clean `origin/main`, same worktree, only the shebang line's `\r` differing:

| `plugin/scripts/launch-server.mjs` first line | result |
|---|---|
| `#!/usr/bin/env node\r\n` (as checked out on Windows) | `Test Files 1 failed`, `Tests no tests`, `SyntaxError: Invalid or unexpected token` |
| `#!/usr/bin/env node\n` | `Test Files 1 passed`, **`Tests 11 passed`** |

`node --check plugin/scripts/launch-server.mjs` exits **0** either way — node strips shebangs itself. It is vitest's transform that chokes on the trailing `\r`, and the test reaches it through `await import("../../../plugin/scripts/launch-server.mjs")` at line 15.

## Why it matters more than 11 tests

That import is not incidental. The file's own header says importing it *is* the proof that importing it is safe (#1447) — so the assertion the suite exists to make is precisely the one that never runs. Eleven tests, including that one, have been silently absent for every Windows contributor.

CI is Linux, so `.mjs` checks out LF there and the suite collects normally. That is why this is green everywhere anyone looks: **the hole is invisible to CI by construction**, and only a Windows checkout can see it.

## Not limited to this file

Every shebanged `.mjs` in the repo has the same CRLF first line — at least `plugin/hooks/job-complete-notify.mjs`, `open-latest-image.mjs`, `save-warning.mjs`, `vram-check.mjs`, `plugin/scripts/monitor-progress.mjs`, and a range of `scripts/*.mjs`. `launch-server.mjs` is simply the one a suite currently imports; any future test that imports one of the others fails the same way, and would look like a broken test rather than a line-ending issue.

## Suggested direction

`.gitattributes` already establishes exactly this pattern, and for the same reason — it forces LF on the tool-surface ratchet files because "with core.autocrlf they would check out as CRLF on Windows and hash differently than the LF form CI produces, breaking the gate for Windows contributors."

[...]

PR #1858, opened 2026-08-20
fix(1855): check_runtime knows PoYo's pack is a paid service, not a local node

### LINKED ISSUE #1855: check_runtime misclassifies poyo-comfyui service nodes as local/free
list_packs(action:"check_runtime") classified PoYo_GenerateImage as local/free because the external-service recognizer has no entry for the installed poyo-comfyui pack and the node keeps credentials out of its workflow inputs. /object_info reports python_module "custom_nodes.poyo-comfyui", category "PoYo AI/Generate", api_node false.

### Steps to reproduce
Call check_runtime with a UI graph containing type PoYo_GenerateImage. The result is runtime:"local", usesApiNodes:false.

### Exact error (scrubbed)
Wrong/misleading result: {"runtime":"local","usesApiNodes":false,"classTypes":["PoYo_GenerateImage"]}

### Fix
Add this known external provider entry to dist/services/api-nodes.js:
+ { module: "poyo-comfyui", categoryPrefixes: ["poyo ai"], provider: "PoYo" },

Verified in a fresh Node import with mocked /object_info: runtime:"api", usesApiNodes:true, externalApiNodes:["PoYo_GenerateImage"], externalProviders:["PoYo"].

### Environment
Windows 11
ComfyUI 0.33.0 / frontend 1.49.6
Python 3.13.11 / PyTorch 2.13.0+cu130
RTX 5090 32 GB
comfyui-mcp 0.52.23 / panel 0.14.24

artokun:
Fixes #1855.

`list_packs(action:"check_runtime")` reported `PoYo_GenerateImage` as `{"runtime":"local","usesApiNodes":false}` — a node that spends the user's balance, described as free.

## Why nothing generic could have caught it

Every signal this module has says "free" for this pack:

- it keeps its credential **out of the workflow inputs**, so `declaresCredentialInput()` has nothing to match;
- `/object_info` reports `api_node: false`;
- the category (`PoYo AI/Generate`) names a vendor no rule knows.

That is exactly what `EXTERNAL_SERVICE_PACKS` exists for, and the module is explicit that a named list is the right tool because "each entry is a claim someone can check."

## So I checked the claim rather than taking it from the report

`poyo-comfyui` is the **official** pack of the PoYo AI hosted image/video/music/3D API ([github.com/PoyoAPI/poyo-comfyui](https://github.com/PoyoAPI)), and using it requires an account API key created at `poyo.ai/dashboard/api-key`. That is the case this recognizer is for: calling the node spends a balance at a named provider.

## The entry

```ts
{ module: "poyo-comfyui", categoryPrefixes: ["poyo ai"], provider: "PoYo" },
```

Module **and** category prefix, the same shape as the fal entries — a user who clones the pack into a differently-named directory changes `python_module` and nothing else, while the category is baked into the pack's own source.

**The reporter patched `dist/services/api-nodes.js`.** That is build output; the fix belongs in `src`, or the next build drops it.

## Verification

- Three tests: the reported graph, the renamed-directory path, and a **negative** case proving `PoYoAlpha/Mask` is not swept in by the `poyo ai` prefix.
- **Three mutations, all killed**: removing the entry (2 red), dropping `categoryPrefixes` (1 red — the rename test), and widening the prefix to bare `poyo` (1 red — the negative test earns its place).
- Typecheck clean. Full suite: 10760 passed, 1 pre-existing failure unrelated to this change.

## About that pre-existing failure

A fresh worktree fails `package-hooks.test.ts > every entry in \`files\` exists on disk` because `dist/` has not been built, and `launch-server.test.ts` fails to collect. **Both reproduce identically on clean `origin/main`**, so neither is from this change — I verified that rather than assume it.

The second one turned out to be worth its own report: a CRLF shebang makes vitest fail to collect that suite, so it contributes **0 tests** on every Windows checkout while CI (Linux, LF) sees 11 passing. Filed as #1857.

### DIFF
--- a/src/__tests__/services/external-service-nodes-not-free.test.ts
+++ b/src/__tests__/services/external-service-nodes-not-free.test.ts
@@ -62,6 +62,16 @@ const IMPACT = def({
   input: { required: { image1: ["IMAGE", {}] } },
 });
 
+/** #1855 — PoYo's official pack. Note what is NOT here: no api_key input (the pack keeps
+ *  the credential out of the workflow) and api_node is absent/false. Every generic signal
+ *  this module has says "free"; only the named entry knows better. */
+const POYO_GENERATE = def({
+  name: "PoYo_GenerateImage",
+  category: "PoYo AI/Generate",
+  python_module: "custom_nodes.poyo-comfyui",
+  input: { required: { prompt: ["STRING", {}] } },
+});
+
 const CORE_SAMPLER = def({ name: "KSampler", category: "sampling", python_module: "nodes" });
 const CORE_PREVIEW = def({ name: "PreviewImage", category: "image", python_module: "nodes" });
 const CORE_LOAD = def({ name: "LoadImage", category: "image", python_module: "nodes" });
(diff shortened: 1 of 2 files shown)


artokun:
## Gate: **SHIP**

It went further than I did — it **pulled the actual pack** rather than trusting the report — and two of its observations changed the diff.

### 1. The registry install path would have been missed

The Comfy Registry id is **`poyo-nodes`** (`pyproject` `name`, `PublisherId = "poyo"`), not the GitHub repo name `poyo-comfyui`. A registry install therefore reports `python_module: "custom_nodes.poyo-nodes"` — which does **not** contain the module substring at all.

So on the common install route this entry matches **only** through `categoryPrefixes: ["poyo ai"]`. The module string alone would have missed it entirely. That path now has a test, and dropping `categoryPrefixes` fails **two** tests instead of one.

### 2. My negative test claimed more than it asserted

`"PoYoAlpha/Mask"` diverges from `"poyo ai"` at the space, so even a naive `startsWith(prefix)` passes it — the mutation **survived**. The comment said the test pinned the exact-or-slash boundary; it did not.

`"PoYo AI Extras/Mask"` shares the whole prefix and differs only after it, which is the boundary the rule actually implements. Weakening `packMatches` to `startsWith` now fails. The shipped code was correct either way — the test was the weak part.

### Independently confirmed

- The pack: `poyo_nodes/nodes.py` has `CATEGORY = "PoYo AI"`, and **all 10 classes are remote** (`/Jobs`, `/Generate`, `/Files`, `/Chat`). No local helpers exist, so no `localCategoryPrefixes` is owed — unlike the fal entries, which need one.
- The claim in the entry: repo description is literally "Official PoYo custom nodes for ComfyUI", and the README says *"The API key is deliberately kept outside workflow JSON"* (`POYO_API_KEY`) — which is exactly why `declaresCredentialInput()` cannot see it.
- Driven through the production path `checkWorkflowRuntime`, not the helper.
- Case handling: `packMatches` lowercases both sides, and an UPPERCASE category still matched.

One wording correction it made that I have kept: "the way the fal entries do" was loose — only 1 of the 4 fal entries carries `categoryPrefixes`.

### Suite state

582/584 files pass. The 2 failures are pre-existing environment noise on a never-built worktree, not this diff — `package-hooks` wants `dist/`, and `launch-server.test.ts` fails to collect. Both reproduce identically on clean `origin/main`. The second turned out to be worth its own report — a CRLF shebang makes vitest fail to collect that suite, so it contributes **0 tests** on every Windows checkout while CI (Linux, LF) sees 11 passing. Filed as #1857.

PR #1860, opened 2026-08-20
fix(1857): launch-server tests collect on a Windows checkout

### LINKED ISSUE #1857: launch-server.test.ts silently contributes 0 tests on Windows: a CRLF shebang breaks vitest collection

artokun:
Fixes #1857.

`src/__tests__/scripts/launch-server.test.ts` contributed **0 tests** on any Windows checkout. It was not skipped — the suite failed to *collect*: `SyntaxError: Invalid or unexpected token`, reported as one failed file with no tests.

`plugin/scripts/launch-server.mjs` checked out as `#!/usr/bin/env node\r\n`. `node --check` exits 0 either way (it strips shebangs itself). Vitest's transform does not, and the suite reached the file through `await import("../../../plugin/scripts/launch-server.mjs")`. That import *is* the proof that importing the launcher is safe (#1447), so the assertion the suite exists to make never ran.

Measured on this worktree, only the shebang line's `\r` differing:

| first line | result |
|---|---|
| `#!/usr/bin/env node\r\n` | Test Files 1 failed, Tests no tests |
| `#!/usr/bin/env node\n` | Test Files 1 passed, Tests 11 passed |

CI is Linux, so `.mjs` checks out LF there. The hole was invisible to CI by construction.

## Fix

`.gitattributes` already forces LF on the tool-surface ratchet and locale catalogs for the same reason. This adds `*.mjs text eol=lf` so a Windows clone cannot reintroduce the CR.

The suite also moves the import from collection into `beforeAll`, so a regression is a failed hook (and a failed shebang-byte test) rather than an empty file. A `git check-attr eol` assertion pins the attribute on the real launcher path — reading `.gitattributes` and guessing which pattern applies would reimplement the matcher.

## Verification

- `npx vitest run src/__tests__/scripts/launch-server.test.ts --maxWorkers=4` — **13 passed** on this Windows worktree (the original 11 plus the shebang-byte and `git check-attr` pins).
- Mutation: restore the CRLF shebang → collection still reports 13 tests; the shebang-byte test fails with `CR before LF on the shebang`; the import describes fail with `SyntaxError` instead of contributing 0 tests. Restore LF → 13 passed again.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -21,3 +21,11 @@ docs/design/tool-vocabulary.json text eol=lf
 # suggest. The translations are pinned alongside it so every catalog in the
 # directory is the same shape on every platform.
 locales/**/*.json text eol=lf
+
+# Shebang .mjs files imported by vitest fail to COLLECT on Windows when
+# core.autocrlf checks them out as CRLF (#1857). The transform treats
+# `#!/usr/bin/env node\r` as a syntax error; the file reports "1 failed /
+# Tests no tests" and the tests that exist never run. `node --check` is
+# fine either way — it strips shebangs itself. CI is Linux, so the hole
+# is invisible there. Force LF the same way we do for the ratchet files.
+*.mjs text eol=lf
--- a/src/__tests__/scripts/launch-server.test.ts
+++ b/src/__tests__/scripts/launch-server.test.ts
@@ -7,13 +7,18 @@
 //
 // The decisions are exercised against the REAL exports — the launcher is
 // import-safe by design, so importing it here is also the proof that importing
-// it spawns nothing.
+// it spawns nothing. The import is inside beforeAll, not at collection: a
+// top-level `await import` of a CRLF shebang is a SyntaxError during
+// transform, which vitest reports as "1 failed / Tests no tests" (#1857).
 
-import { describe, expect, it } from "vitest";
+import { beforeAll, describe, expect, it } from "vitest";
+import { execFileSync } from "node:child_process";
 import { readFileSync } from "node:fs";
 
-const launcher = await import("../../../plugin/scripts/launch-server.mjs");
-const { globalEntry, serverSpec } = launcher as {
+const LAUNCHER_REL = "plugin/scripts/launch-server.mjs";
+const LAUNCHER_URL = new URL("../../../plugin/scripts/launch-server.mjs", import.meta.url);
+
+type Launcher = {
   globalEntry: (stdout: unknown, opts?: { exists?: (p: string) => boolean }) => string | null;
   serverSpec: (
     entry: string | null,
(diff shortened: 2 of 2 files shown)


artokun:
Gated **SHIP** (independent adversarial review; codex unavailable — disclosed). This is the fix for #1857, which I filed, so recording the verification here.

What the review established by running it rather than reading:

- The transform `SyntaxError` and the `1 failed / Tests no tests` output both reproduce verbatim.
- `node --check` **passes** on a CRLF shebang and runs it — node strips shebangs itself, so this was never a node problem, only a vitest-transform one.
- **The scope is exactly right, neither wider nor narrower than the defect.** A CRLF-shebang `.ts` imports fine (esbuild strips it); a CRLF-shebang `.mjs` throws. The three shebang `.ts` files (`src/boot.ts`, `src/index.ts`, `src/experimental/run.ts`) are `w/crlf` today and are *not* a latent hole.
- Every tracked `.mjs` blob is already `i/lf`, so `text` causes no renormalization churn; no `.mjs` is a byte-pinned fixture; `docker/runpod/.gitattributes` has no `*.mjs` rule; `*.bat text eol=crlf` does not overlap.
- Fresh clone at the fix commit with `core.autocrlf=true`: all 54 `.mjs` land `w/lf`, launcher shebang is `node\n`, zero CRLF.

## One thing worth acting on, since it is easy to miss

**A `.gitattributes` change never renormalizes an existing working tree.** Confirmed on this machine after the merge: `.gitattributes` on `main` carries the rule, and this checkout still has **52 `.mjs` files at `w/crlf`**, including `plugin/hooks/*.mjs`, `plugin/scripts/monitor-*.mjs` and `scripts/arena-*.mjs`. (`launch-server.mjs` itself is clean here because the fix touched it.)

So on any pre-existing clone the hole stays open for those 52 files — a future suite importing one of them would fail to collect exactly as #1857 did. The one-liner is:

```
git add --renormalize .
```

I have deliberately not run it in the shared checkout here: it stages 52 files, and with several sessions working the same tree that is an easy way for someone's next commit to pick up changes they did not make.

Not a blocker for the fix, and the fix is strictly better regardless — on a stale checkout the failure at least goes from an unattributable "no tests" collection error to a self-diagnosing named assertion. Worth a line in the release notes or a follow-up commit so contributors on existing clones know to run it.

PR #1863, opened 2026-08-20
fix(1861): Blind fails closed at both completion doors, and videos are never attached as PNG

### LINKED ISSUE #1861: Synthesised run completions bypass the Blind gate, and video refs ship as image/png (from #1856)
#1856 merged at **10:09:49Z**, about four minutes after a gate NO-SHIP was posted on it with these two findings. They are now on `main`. Filing so they are not lost with the PR — the mechanism in that PR is sound and worth keeping; this is about what its newly-real `images` array reaches.

## P1 — the synthesised completion bypasses the conversation-wide Blind gate

The two arrival paths are asymmetric on `main` right now.

**Panel ingress** (`src/orchestrator/index.ts:5245`) strips carefully:

```js
const blindWithheldCount = anyTabBlind() && Array.isArray(ev.images) ? ev.images.length : 0;
const evForTab = anyTabBlind()
  ? { ...ev, images: [], ...(blindWithheldCount > 0 ? { note: "⚠️ Blind mode is ON: N image(s) … were withheld from you and are NOT attached … Do NOT comment on motion, sharpness, composition …" } : {}) }
  : ev;
```

**Watchdog deliver** (`:6381`) records the raw payload:

```js
deliver: (payload, ticket) => {
  const entry = RunCompletions.record(ticket.tabId, payload, { … });
```

There is no `anyTabBlind()` anywhere in that path. It was inert while `images` was always `[]`; #1853 fills it from `/history`, so it is live now.

Reproduced during the gate: `PanelAgent` with `isBlind: () => true`, `injectEvent({kind:"executed", images:[{filename:"secret_render_00001_.png"}]})` → `turn.images = [{"filename":"secret_render_00001_.png","type":"output","automatic":true}]` with text "The image(s) are attached below". `injectEvent` never consults `isBlind` — it is wired only to the native-tool PreToolUse gate (`panel-agent.ts:597`, `:2634`).

So: Blind ON, plus a `panel_run` whose panel `executed` frame is dropped — **the exact case #1789/#1853 exist to cover** — and 45 s later the render's pixels are attached to the agent turn. `blind-tab-gate.ts:15` states the invariant directly: *"every run completion was stripped… Blind must fail CLOSED."*

No test covers it, which is why the suite is green. Blind is a user-facing privacy control; the asymmetry means it holds at one door and not the other.

## P1 — video refs are delivered as `image/png`

`run-completion-watchdog.ts:159-160` pushes video refs into the same array:

```js
for (const node of notification.video_outputs) {
  for (const vid of node.videos) { …
```

and `usableHistoryImages` (`:121`) filters only on filename presence and dedup — no extension or media-type check. The consumer then forces the type: `claude-backend.ts:389-395`, *"ComfyUI outputs are PNG by default"*, with no strip-and-retry in that backend.

Verified against a live ComfyUI during the gate rather than a fixture: `/history` entry `2128b89f-…` has `outputs["92"].images = [{filename:"MiniMax_H3_00051_.mp4", subfolder:"video", type:"output"}]`; `GET /view` returns `Content-Type: video/mp4`, 200,883 bytes — under the 12 MB cap. So `ftyp` bytes reach the Messages API inside an image block.

The panel never produced this shape — its video segment ref is an uploaded `storyboard_<base>.png` (`run-completion-frame.js:552`) or `ref:null`/note-only — which is exactly why the coercion was safe before. **The headline test asserts a `.mp4` as the expected attachment**, so the suite pins this rather than catching it.

## Suggested shape

Route the synthesised payload through the same `anyTabBlind()` strip the ingress uses, so the invariant holds wherever a completion enters rather than at one door. And split video refs out of `images` so they are not coerced to `image/png` — that test needs to change with it, since it currently encodes the wrong expectation.

Neither is a reason to revert #1856; the correlation and journal work in it is good.

artokun:
Fixes #1861. Both defects were raised on #1856 before it merged; this is the follow-through.

## Blind was bypassed at the second door

A run completion reaches the agent through **two** doors: the panel's own `executed` frame at the agent-event ingress, and the #1789 watchdog, which *synthesises* one when that frame never comes.

The strip lived inline at the first door. The second called `RunCompletions.record` with the **raw payload**. That was inert only while its `images` was always `[]` — #1853 filled it from `/history`, and the bypass went live: **Blind ON plus a dropped `executed` frame attached the render's pixels to the agent turn ~45 s later**, which is exactly the case #1789 exists to cover. `blind-tab-gate.ts` states the invariant it broke: *Blind must fail CLOSED.*

The strip is now a pure module (`blind-completion.ts`) that both doors call. Not for tidiness — a strip that has to be *remembered* at each new arrival point is one that gets missed at the next one, which is precisely what happened here. It is also now testable without standing up the orchestrator.

A wiring test asserts **every** `RunCompletions.record` call strips first, and that the disclosure text has not been copied back inline anywhere.

## Videos were attached as `image/png`

`run-completion-watchdog` pushed SaveVideo refs into `images`, whose only consumer attaches them inline with the media type forced to `image/png` ("ComfyUI outputs are PNG by default"), with no strip-and-retry in that backend. A `/history` `.mp4` therefore reached the Messages API as `ftyp` bytes inside an image block — verified against a live ComfyUI, `Content-Type: video/mp4`, under the size cap.

The panel never produced this shape — its video ref is an uploaded `storyboard_<base>.png` or note-only — which is why the coercion was safe until the watchdog began reading `/history` directly.

Media is now partitioned by an **allowlist** of attachable image extensions, so an unknown type fails closed: named in the note, never attached. The agent still learns the run produced it and how to fetch it.

## Three existing tests asserted the defect

They pinned the video-as-attachment behaviour rather than catching it. Two are corrected to assert it is named and not attached. The third — "empty or malformed image refs are dropped" — was about **blanks and dedup**, and its video fixture was incidental; it keeps its intent with an image fixture, because a video there would now be filtered for an unrelated reason and prove nothing about either.

## Verification

- **Two mutations, both killed**: reverting the watchdog door to the raw payload (2 tests red), and attaching every ref again (2 tests red).
- New suite 7/7; watchdog suites 141/141; full suite **10776 passed**, `tsc --noEmit` clean.
- The 2 failing files are pre-existing environment noise on an unbuilt worktree — `package-hooks` wants `dist/`, and `launch-server.test.ts` fails to collect (#1857). Both reproduce identically on clean `origin/main`.

Nothing here reverts #1856 — its correlation and journal work is good, and this is about what its newly-real `images` array reaches.

### DIFF
--- a/src/__tests__/orchestrator/blind-completion.test.ts
+++ b/src/__tests__/orchestrator/blind-completion.test.ts
@@ -1,0 +1,93 @@
+/**
+ * #1861 — Blind must fail CLOSED at EVERY door a run completion enters.
+ *
+ * The panel's own `executed` frame arrives at the orchestrator's agent-event ingress; the
+ * #1789 watchdog SYNTHESISES a completion when that frame never comes. The strip lived
+ * inline at the first door only, so the second recorded the raw payload. That was inert
+ * while its `images` was always empty — and #1853 filled it from `/history`, making the
+ * bypass live: Blind ON plus a dropped `executed` frame attached the render's pixels to
+ * the agent turn about 45 s later.
+ */
+import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+
+import { stripBlindCompletion } from "../../orchestrator/blind-completion.js";
+
+const INDEX_TS = fileURLToPath(new URL("../../orchestrator/index.ts", import.meta.url));
+
+const completion = () => ({
+  kind: "executed",
+  prompt_id: "p1",
+  images: [
+    { filename: "secret_render_00001_.png", type: "output" },
+    { filename: "secret_render_00002_.png", type: "output" },
+  ],
+  note: "The render finished.",
+});
+
+describe("stripBlindCompletion (#1861)", () => {
+  it("removes the pixels and says so when blind", () => {
+    const out = stripBlindCompletion(completion(), true);
+    expect(out.images).toEqual([]);
+    expect(String(out.note)).toContain("Blind mode is ON: 2 image(s)");
+    // The original note is kept — the disclosure is appended, not substituted.
+    expect(String(out.note)).toContain("The render finished.");
+    // It must also override a sighted instruction the agent would otherwise follow.
+    expect(String(out.note)).toContain("Do NOT comment on motion, sharpness, composition");
+  });
+
+  it("passes the completion through untouched when not blind", () => {
+    const ev = completion();
+    const out = stripBlindCompletion(ev, false);
+    expect(out).toBe(ev);
+    expect(out.images).toHaveLength(2);
+  });
+
+  it("discloses nothing when there was nothing to withhold", () => {
+    // A status-only completion has no pixels; claiming images were withheld would report
+    // a withholding that never happened.
+    const out = stripBlindCompletion({ kind: "executed", note: "It failed." }, true);
+    expect(out.images).toEqual([]);
+    expect(String(out.note)).toBe("It failed.");
+    expect(String(out.note)).not.toContain("Blind mode is ON");
+  });
+
+  it("treats a malformed images field as nothing to withhold, and still empties it", () => {
+    // Fail closed on shape too: whatever `images` was, it is [] afterwards.
+    for (const bad of [undefined, null, "two", 7, {}]) {
+      const out = stripBlindCompletion({ kind: "executed", images: bad } as never, true);
+      expect(out.images).toEqual([]);
+      expect(String(out.note ?? "")).not.toContain("Blind mode is ON");
+    }
+  });
+});
+
+describe("#1861 WIRING: both completion doors go through the strip", () => {
+  // A strip that is correct and applied at one door is the bug this fixes. Assert on the
+  // source, because the second door's payload reaches `RunCompletions.record` inside the
+  // orchestrator's construction and no unit test stands that up.
+  const src = readFileSync(INDEX_TS, "utf8");
+
+  it("the pure strip is the only implementation", () => {
+    expect(src).toMatch(/import \{ stripBlindCompletion \} from "\.\/blind-completion\.js";/);
+    // The disclosure text must not have been copied back inline anywhere.
+    expect(src).not.toMatch(/Blind mode is ON: \$\{/);
+  });
+
+  it("EVERY RunCompletions.record call strips first", () => {
+    const calls = src.match(/RunCompletions\.record\([^)]*/g) ?? [];
+    expect(calls.length).toBeGreaterThanOrEqual(2);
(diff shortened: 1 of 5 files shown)


artokun:
## Gate: **SHIP**

**Two mutations killed**, and the one that matters fired at the *call site*: reverting door 2 to `RunCompletions.record(ticket.tabId, payload, …)` fails 2 wiring tests, and widening the regex to `/.*/ ` fails 2 watchdog tests. It also confirmed the three inverted fixtures had been pinning the defect and are correctly flipped.

### One finding I fixed rather than recorded

Class 3, the wrong-pair check. My allowlist asked *"is this an image extension?"* — but the question that binds is *"can `fetchImageBlock` label these bytes honestly?"* `claude-backend.ts:390` keeps only `image/{png,jpeg,gif,webp}` and rewrites everything else to `image/png`.

**`.bmp` was in my list.** It *is* an image, and ComfyUI serves it `image/bmp` (WAS-Suite `Image Save` writes them), so attaching one declares PNG over BMP bytes — a smaller lie than an `.mp4`, the same lie. The gate rated it non-blocking because it is equally wrong on `main` (which attached everything), so the diff did not make it worse. But "closes most of the gap" is a strange place to stop when the fix is one entry wide.

`.bmp` is dropped, the reasoning is in the comment — the next person to widen this set would otherwise widen it against the wrong question — and a test pins the pairing: putting `.bmp` back turns it red.

### Verified by execution, not reading

- `partitionAttachableMedia` keeps "unknown" *representable* rather than silently dropping it: no-extension, `.avif`, `.webm`, `.flac` and `foo.png.mp4` all land in `other` and are named.
- The spread preserves `prompt_id` / `completion_source` / `run_status`, and `tick()` re-reads `awaiting(promptId)` **after** the async `/history` fetch, so the ticket is not stale at `deliver`.
- The panel door is behaviourally byte-identical — same prefix trim, same `withheld > 0` gate. The only semantic change is `anyTabBlind()` being called once instead of twice.
- Everything withheld is named with a real remedy: `list_outputs` genuinely exists (`src/tools/image-management.ts:200`).
- `anyTabBlind()` cannot throw the new call into the watchdog's catch — `resolveBlindTabGate` swallows a throwing `canReach` probe.

### Known cosmetic wart, not fixed

Under Blind with mixed png+mp4, the note reads "…are attached: r_00001_.png." immediately before the disclosure says they are not. It self-corrects in the same breath. Fixing it properly means threading blind-ness into the payload composer, which is scope creep on a P1; the honest disclosure is present and unambiguous, so I left it rather than half-fix it by parsing the prose back out.

### A process note on my own run

I mutation-tested the `.bmp` change **before committing it**, so `git checkout --` reverted the fix along with the mutation and the "baseline" came back red. Caught it, re-applied, committed, and re-ran — mutation kills 1, baseline 142/142. Commit before you mutate; the restore does not know which of your edits was the experiment.

PR #1864, opened 2026-08-20
fix(1491): an auto-refresh that outran its ack is not a failed refresh

### LINKED ISSUE #1491: panel_add_node auto-refresh times out while explicit panel_refresh_nodes succeeds immediately afterward
Adding a core LoadImage node detected a stale schema and correctly refused, then its automatic panel_refresh_nodes dispatch timed out after 30 seconds. An explicit panel_refresh_nodes call immediately afterward succeeded, so the auto-recovery path can fail transiently even though the tab and refresh capability are healthy.

### Steps to reproduce
1. Open a workflow containing existing LoadImage nodes.
2. Let the browser node schema become stale/retyped.
3. Call panel_add_node for LoadImage.
4. Observe stale-schema refusal and automatic refresh timeout.
5. Call panel_refresh_nodes explicitly; it succeeds.

### Exact error (scrubbed)
Panel tab <workflow> did not reply to refresh_nodes within 30000 ms - the ComfyUI tab may be backgrounded or frozen.

### Fix
The auto-recovery inside panel_add_node should use the same reliable dispatch path as explicit panel_refresh_nodes, or retry once / allow a longer bounded timeout before concluding the tab is frozen. The operation recovered with one explicit refresh and no canvas mutation.

### Environment
Windows 11; ComfyUI 0.33.0; ComfyUI frontend 1.49.6; RTX 5090 32 GB; comfyui-mcp 0.52.23; panel 0.14.24.

artokun:
Fixes artokun/comfyui-mcp-panel#1491. Filed on the panel; the dispatch and the message are orchestrator-side, so the fix lands here.

## The reporter's diagnosis is wrong, and the correct one is the fix

They concluded the auto-recovery inside `panel_add_node` "should use the same reliable dispatch path as explicit `panel_refresh_nodes`". It already does — byte for byte:

```ts
// panel_add_node auto-recovery (:11151)
const refreshed = await ctx.call({ cmd: "refresh_nodes" }, OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS);

// panel_refresh_nodes tool (:12463)
ctx.call({ cmd: "refresh_nodes" }, OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS)
```

Same command, same constant, same call. What differs is what happens **after the ack window closes**: `joinMs` never cancels work already in flight. The panel's coalescer says so outright —

> "`opts.joinMs` never cancels work already in flight. The run keeps going, registers whatever it fetched, and clears the slot exactly as before."

— and its own worked example is a join ending at 20,000 ms plus a run adding ~13,000 ms: **~33 s against a 30 s relay window**, with every per-step bound respected.

So on a large install the refresh legitimately outlives the ack and then **completes**. The explicit `panel_refresh_nodes` that "succeeded immediately afterward" was not a more reliable path — it was the same work, already done.

## What was actually broken: the reply

On that timeout the caller was told:

> *"panel_refresh_nodes was dispatched and FAILED, so the schema is unchanged and retrying the add will refuse again."*

Both halves are false in the timeout case, and the second is the damaging one — **it argues the caller out of the one action that works**, which is precisely what the reporter did next, twice, successfully.

A reply timeout and an acked executor failure now get opposite advice. The genuine-failure branch is untouched; softening it would be the same defect mirrored.

## Decided on the bridge-owned tag, not the message text

`isReplyTimeoutResult()` reads the `REPLY_TIMEOUT_RESULT` symbol that `ctx.call` carries from the bridge (#1468). The comment there records that two codex rounds rejected deciding this from text and both were right: the panel relays arbitrary error text, and `ctx.call` flattens a tagged timeout and an acked error into the same text-only result.

## My own tests did not pin that at first

Mutation testing caught it: replacing the tag check with a `/did not reply to/` text match **passed all 8 tests**. The suite agreed with both implementations, because the timeout fixture's text contained the phrase and the failure fixture's did not.

The separating input is the one #1468 describes — an **acked** executor error whose text *quotes* the bridge's timeout wording. That fixture is now in the suite, and text-matching fails against it.

## Verification

- 9/9 in `add-node-stale-schema-retry.test.ts`; full suite **10793 passed**, `tsc --noEmit` clean.
- **Two mutations, both killed**: text-matching instead of the tag (1 red), and treating every error as still-running (2 red).
- The single remaining suite failure is `package-hooks.test.ts` wanting a built `dist/` in a fresh worktree — pre-existing, reproduces on clean `main`.

Worth noting: `launch-server.test.ts` now **passes** in this fresh worktree, which confirms #1860's `.gitattributes` fix works for new checkouts (existing clones still need `git add --renormalize .`).

### DIFF
--- a/src/__tests__/orchestrator/add-node-stale-schema-retry.test.ts
+++ b/src/__tests__/orchestrator/add-node-stale-schema-retry.test.ts
@@ -18,6 +18,7 @@
 // unknown, which is exactly why those are never re-issued on our own initiative.
 
 import { describe, expect, it } from "vitest";
+import { markReplyTimeout } from "../../services/ui-bridge.js";
 
 import {
   buildPanelToolDefs,
@@ -39,13 +40,21 @@ const STALE = () =>
  * `addOutcomes` is consumed one per graph_add_node call: "stale" throws the refusal,
  * "ok" succeeds. `refreshFails` makes refresh_nodes throw.
  */
-function bridge(addOutcomes: Array<"stale" | "ok">, refreshFails = false) {
+function bridge(addOutcomes: Array<"stale" | "ok">, refreshFails: boolean | "timeout" = false) {
   const calls: string[] = [];
   let addIdx = 0;
   const b = {
     send: async (cmd: Record<string, unknown>) => {
       calls.push(String(cmd.cmd));
       if (cmd.cmd === "refresh_nodes") {
+        // #1491 — THREE outcomes, not two. A reply TIMEOUT is tagged by the bridge
+        // itself (markReplyTimeout); an acked executor error is a plain throw. The
+        // orchestrator must tell them apart, because the correct advice is opposite.
+        if (refreshFails === "timeout") {
+          throw markReplyTimeout(
+            new Error(`Panel tab ${TAB} did not reply to "refresh_nodes" within 30000 ms`),
+          );
+        }
         if (refreshFails) throw new Error("panel did not answer the refresh");
         return { refreshed: true };
       }
@@ -68,7 +77,7 @@ function bridge(addOutcomes: Array<"stale" | "ok">, refreshFails = false) {
   return { b, calls };
 }
 
-async function addNode(addOutcomes: Array<"stale" | "ok">, refreshFails = false) {
+async function addNode(addOutcomes: Array<"stale" | "ok">, refreshFails: boolean | "timeout" = false) {
   const { b, calls } = bridge(addOutcomes, refreshFails);
   const ctx = makePanelToolCtx(b, TAB, new WorkflowTargetStore());
   const def = buildPanelToolDefs().find((d) => d.name === "panel_add_node");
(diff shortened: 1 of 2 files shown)


artokun:
## Gate: **SHIP**

**Three mutations, all killed**: `stillRunning = false` (1 fail), `= true` (3 fails), and decide-on-text `/did not reply to/.test(...)` — killed *specifically* by the quoting fixture. It built the same adversarial input independently (an acked executor error whose message quotes the bridge's timeout wording) and confirmed it takes the FAILED branch. The tag-vs-text discrimination is pinned, not decorative.

### The reachability question I had not asked

`refresh_nodes` is in `BRIDGE_READONLY_CMDS`, so the reviewer chased the **second, pre-write** `markReplyTimeout` site (`ui-bridge.ts:4807`) — the one that never calls `markDispatched`, and therefore the one shape that could let this note claim "it keeps running" about a command that was never written.

It is unreachable on a first dispatch: `deadline` is assigned in the same synchronous block as `dispatch()`, with no await between, and it fires only on a reconnect resume where the first dispatch *did* write. So "was dispatched" stays true wherever this branch can be reached. That was the sharpest risk in the change and I had not thought to check it.

### Also verified by execution

- The "30s window" claim checks out against `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS = 30_000`.
- **No in-tool retry loop**: this branch returns early and never reaches `second = await add()`, so the tool still performs exactly **one** add per invocation. A genuinely frozen tab therefore costs the agent one wasted attempt, not a loop — which was the real risk in turning a loop-stopping message into an encouraging one.
- 9/9 on the changed file, 402/402 across six related suites, `tsc --noEmit` / `check:unknown-collapse` / `check:vocabulary` green.

### One noted wart, not fixed

The new note precedes the bridge's own "may be backgrounded or frozen" wording, so the reply now frames that speculation correctly but sits awkwardly next to it. The reviewer's call — and mine — is awkward adjacency, not a false claim. Rewording the bridge's generic ack-timeout text is a wider change than this issue, and it is shared by every command that can time out.

Issue #599, opened 2026-08-01
panel_set_widget false timeout when fresh /object_info exceeds 6 seconds


artokun:
### What happened / root cause
panel_set_widget timed out on a large ComfyUI install because graph_set_widget awaits a fresh /object_info authorization in the frontend, while the orchestrator used the bridge's 6000 ms default acknowledgement timeout. Read-only graph queries and graph_disconnect worked, but every panel_set_widget returned a false panel-tab timeout.

### Steps to reproduce
1. Open a workflow containing a custom node on a ComfyUI install with many registered custom-node packs.
2. Call panel_set_widget on a text widget.
3. The frontend begins fresh /object_info authorization, but the bridge times out after 6000 ms before it replies.

### Exact error (scrubbed)
Error: Panel tab wf:workf did not reply to "graph_set_widget" within 6000 ms — the ComfyUI tab may be backgrounded or frozen

### Fix
Applied locally: yes. Upstream-only: no.
In dist/orchestrator/panel-tools.js, give graph_set_widget a 30000 ms acknowledgement timeout because this guarded write intentionally waits for /object_info.

Patch:
- return ctx.call({ cmd: "graph_set_widget", node_id: args.node_id, widget: args.widget, value });
+ return ctx.call({ cmd: "graph_set_widget", node_id: args.node_id, widget: args.widget, value }, 30000);

Verification is pending only because panel_reload temporarily disconnected the sidebar; the original failure was reproduced repeatedly and isolated to set_widget.

### Environment
Windows 11; ComfyUI 0.29.0; NVIDIA RTX 5090 32 GB; comfyui-mcp 0.48.24; panel 0.11.21.

PR #1867, opened 2026-08-20
fix(1493): panel_get_errors waits as long as the panel is allowed to take

### LINKED ISSUE #1493: panel_get_errors times out after refreshing and setting promoted subgraph model widgets
After panel_refresh_nodes and three successful panel_set_widget calls targeting promoted widgets on subgraph node 5004, panel_save_workflow succeeded, but the immediate panel_get_errors call timed out after 20 seconds. The workflow tab was active and had responded to all preceding mutations. Likely graph_get_errors can wedge or exceed its deadline while rescanning refreshed model assets/subgraphs.

### Steps to reproduce
1. Open a workflow containing a model-loader subgraph with promoted COMBO widgets.
2. Call panel_refresh_nodes.
3. Re-set several promoted model widgets to their current values.
4. Save the workflow.
5. Immediately call panel_get_errors.

### Exact error (scrubbed)
Panel tab workflows/LTX2.5_Character_ReferenceVideo_Pose_Control.json did not reply to graph_get_errors within 20000 ms - the ComfyUI tab may be backgrounded or frozen.

### Fix
applied locally: no
upstream-only: yes
Precise change needed: make graph_get_errors non-blocking or raise/budget its asset scan separately; return partial/stale scan state instead of timing out the entire panel call.

### Environment
Windows 11; ComfyUI 0.33.1; RTX 5090 32 GB; comfyui-mcp 0.52.25; panel 0.14.31; frontend version unavailable in this session.

artokun:
Fixes artokun/comfyui-mcp-panel#1493 (cross-repo — the report is filed against the panel, the defect is on the orchestrator side of the same invariant; the issue is closed by hand after merge).

## What was reported

`panel_refresh_nodes` → three `panel_set_widget` writes on promoted COMBO widgets of a model-loader subgraph → `panel_save_workflow` — every one acked — then:

```
Panel tab workflows/LTX2.5_Character_ReferenceVideo_Pose_Control.json did not reply to
graph_get_errors within 20000 ms — the ComfyUI tab may be backgrounded or frozen
```

The tab was neither backgrounded nor frozen. It had just answered four commands.

## Mechanism

`graph_get_errors` is the panel command that deliberately spends the longest wall clock:

- its handler awaits a **forced** `/object_info` re-register before it will trust a combo to clear a resolved missing-model candidate (#610 measured that at ~14.5 s on a real Windows install);
- and it spends a **shared 18 000 ms elective budget** across that refresh race, the `/system_stats` probe and the `/view` media probes — `GET_ERRORS_TOTAL_BUDGET_MS` in the panel's `web/js/lib/get-errors-budget.js`.

That 18 000 ms was sized against the orchestrator's generic **20 000 ms** read default (`BRIDGE_READ_DEFAULT_TIMEOUT_MS`). Which leaves **2 000 ms** for everything the call still has to do *after* its waits: `collectMissingAssets` over the whole graph, the per-node live combo scan, the subgraph walk, and serializing a reply large enough that this very tool ships two truncation riders for it.

The reported sequence spends the budget exactly as designed. `panel_refresh_nodes` leaves an `/object_info` run in flight; `graph_get_errors` sees raw missing-model candidates (the reason the user was re-setting those widgets), so it takes its full 15 s refresh cap, then up to 3 s of media probes — 18 s gone before the graph work starts. The panel replied; the orchestrator had already given up. The agent is then left with **no error surface at all** for a workflow whose nodes are visibly red, which is precisely the outcome the shared budget exists to prevent (#589).

So the reporter is right about the symptom and their prescribed fix is already in the product: the asset scan **is** budgeted separately and **does** return partial state. What is wrong is the headroom the budget was sized against.

## The fix

`graph_get_errors` joins the **#599 refresh-ack cohort** and forwards the same bounded `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS` (30 s) that `graph_set_widget`, `graph_add_node`, `graph_remove_widget`, `graph_get_object_info` and `refresh_nodes` already use — all for the same reason, all waiting on the same `/object_info` fetch. It was the one member of that cohort left on the generic read default, and it is the member that spends the most of it.

**Why this side, not the panel.** The panel's 18 s bound must stay where it is: it is what keeps a current panel safe in front of an *older* orchestrator still using the 20 s default. Lowering it would buy margin by cutting verification (a skipped refresh means a resolved model keeps reporting missing). Widening the orchestrator's own wait costs nothing but patience on a read that is idempotent — waiting longer can never double-apply anything — and the bound stays finite, so a genuinely frozen tab still fails, at 30 s instead of 20 s.

One line changed. The rest of the diff is the comment recording why, and the tests.

## Proof

`src/__tests__/orchestrator/get-errors-ack-budget.test.ts` — the property under test is the **call site**, not the constant. A shared constant that nothing passes still greps, and a one-argument wiring change is invisible to any test that only exercises the bridge, so every case runs the real tool def and reads the bound it actually forwarded:

1. `panel_get_errors` forwards 30 000 ms, and specifically *something*, and specifically more than `BRIDGE_READ_DEFAULT_TIMEOUT_MS`.
2. It clears the panel's own 18 s elective budget with real margin — not the 2 s that failed.
3. It stays bounded (finite, ≤ 60 s), so a dead tab still fails.
4. It matches its #599 siblings, read from their real defs, so the cohort cannot drift apart if the shared constant is re-tuned.
5. Source guard: **every** `graph_get_errors` dispatch in `panel-tools.ts` carries the constant, so a second call site added later on another branch cannot quietly ship on the default (counting occurrences cannot see a third path).

**Mutation-tested after committing the fix:**

- delete the argument → **5 of 5 fail**;
- shrink it to `21_000` (still above the read default, so the naive "is it bigger than 20 s" check would pass) → **4 of 5 fail**.

**Suite:** `npx vitest run` → 584 files / 10 800 tests pass. Two failures, both pre-existing and unrelated: `scripts/package-hooks` (`files` allowlist wants a built `dist/`, absent in a fresh worktree) and `orchestrator/late-mutation-e2e` (the known wall-clock load flake — **passes 5/5 when re-run alone**, confirmed). `npx tsc --noEmit` clean. `npm run check:vocabulary` clean.

**Browser gate:** deliberately not run, and it does not apply. This change is orchestrator-side only — one argument to `ctx.call`. Nothing the panel renders, and no panel file, is touched.

## Gate

codex was unavailable (account exhausted); gated by an independent adversarial review instead — `node .claude/autopilot/claude-gate.mjs`. Verdict recorded in a comment below.

### DIFF
--- a/src/__tests__/orchestrator/get-errors-ack-budget.test.ts
+++ b/src/__tests__/orchestrator/get-errors-ack-budget.test.ts
@@ -1,0 +1,137 @@
+// #1493 — `panel_get_errors` timed out at 20 000 ms on a tab that was alive.
+//
+// Reported sequence: `panel_refresh_nodes`, three `panel_set_widget` writes on promoted
+// COMBO widgets of a model-loader subgraph, `panel_save_workflow` — every one acked — and
+// then `panel_get_errors` came back as
+//
+//   Panel tab workflows/… did not reply to "graph_get_errors" within 20000 ms —
+//   the ComfyUI tab may be backgrounded or frozen
+//
+// The tab was neither backgrounded nor frozen. It had just answered four commands, and
+// `graph_get_errors` is the panel command that deliberately spends the LONGEST wall clock:
+// its handler awaits a FORCED /object_info re-register before it will trust a combo to
+// clear a resolved missing-model candidate (#610, measured at ~14.5 s on a real Windows
+// install), and it spends a shared 18 000 ms elective budget across that refresh race, the
+// /system_stats probe and the /view media probes (GET_ERRORS_TOTAL_BUDGET_MS in the panel's
+// web/js/lib/get-errors-budget.js).
+//
+// That 18 000 ms was sized against the orchestrator's generic 20 000 ms read default, which
+// leaves 2 000 ms for everything the call still does AFTER its waits — collectMissingAssets
+// over the whole graph, the per-node live combo scan, the subgraph walk, and serializing a
+// reply large enough that the tool ships two truncation riders for it. On the reporter's
+// workflow that margin was not enough, so the panel replied LATE and the orchestrator had
+// already given up — leaving the agent with no error surface at all for a workflow whose
+// nodes are visibly red, which is exactly what the shared budget exists to prevent (#589).
+//
+// The margin has to come from the ORCHESTRATOR. The panel's 18 s bound must stay where it
+// is: it is what keeps a current panel safe in front of an older orchestrator still using
+// the 20 s default. So `graph_get_errors` joins the #599 refresh-ack cohort — the same
+// bounded 30 s budget `graph_set_widget`, `graph_add_node`, `graph_remove_widget`,
+// `graph_get_object_info` and `refresh_nodes` already forward for the same reason.
+//
+// The property under test is the CALL SITE. A shared constant that nothing passes still
+// greps, and a one-argument wiring change is invisible to any test that only exercises the
+// bridge — so these run the real tool def and read the bound it actually forwarded.
+import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { buildPanelToolDefs, type PanelToolCtx } from "../../orchestrator/panel-tools.js";
+import { BRIDGE_READ_DEFAULT_TIMEOUT_MS } from "../../services/ui-bridge.js";
+
+const PANEL_TOOLS_TS = fileURLToPath(new URL("../../orchestrator/panel-tools.ts", import.meta.url));
+
+/** The #599 bounded refresh-ack budget (OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS). Duplicated as
+ *  a literal on purpose: if the constant moves, these tests force the relationship to the
+ *  panel's own budget to be re-examined rather than silently re-broken. */
+const REFRESH_ACK_MS = 30_000;
+
+/** GET_ERRORS_TOTAL_BUDGET_MS in the PANEL repo (web/js/lib/get-errors-budget.js) — the
+ *  wall clock one graph_get_errors call may spend on elective server waits alone, before
+ *  any of its own graph work. The orchestrator's wait must clear this with real margin.
+ *  Duplicated across repos deliberately, exactly as the panel duplicates the 20 000 ms
+ *  read default in its own budget test. */
+const PANEL_GET_ERRORS_ELECTIVE_BUDGET_MS = 18_000;
+
+/** Run a real panel tool def against a stub ctx and hand back the per-call ack bound it
+ *  forwarded as `ctx.call`'s second argument. `undefined` means the tool passed none and
+ *  the bridge would fall back to its own default — which is the bug. */
+async function forwardedAckMs(
+  tool: string,
+  args: Record<string, unknown> = {},
+): Promise<{ cmd: Record<string, unknown> | undefined; timeoutMs: number | undefined }> {
+  const def = buildPanelToolDefs().find((d) => d.name === tool);
+  if (!def) throw new Error(`tool ${tool} not found in buildPanelToolDefs()`);
+  let seenCmd: Record<string, unknown> | undefined;
+  let seenTimeout: number | undefined;
+  const ctx = {
+    call: async (cmd: Record<string, unknown>, timeoutMs?: number) => {
+      seenCmd = cmd;
+      seenTimeout = timeoutMs;
+      return { content: [{ type: "text", text: JSON.stringify(cmd) }] };
+    },
+    tabId: "test-tab",
+  } as unknown as PanelToolCtx;
+  await def.handler(args, ctx);
+  return { cmd: seenCmd, timeoutMs: seenTimeout };
+}
+
+describe("panel_get_errors ack budget (#1493)", () => {
+  it("forwards graph_get_errors with the bounded refresh-ack budget, not the 20 s read default", async () => {
+    const { cmd, timeoutMs } = await forwardedAckMs("panel_get_errors");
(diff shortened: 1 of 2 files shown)


artokun:
## Gate: SHIP (exit 0)

codex was unavailable (account exhausted), so this was gated by an independent adversarial reviewer — `node .claude/autopilot/claude-gate.mjs --worktree … --base origin/main --name 1493` — that had never seen the author's reasoning. Verdict verbatim:

---

SHIP

**Live refutation attempt — the fix reproduces and closes the reported bug on a real rig.** I stood up a real `UiBridge` on a free port, connected a real WebSocket panel tab, and drove the real `panel_get_errors` def through the real `makePanelToolCtx`:

| scenario | with fix | reverted |
|---|---|---|
| panel replies at 25 000 ms | **succeeds**, elapsed 25.4 s | `did not reply to "graph_get_errors" within 20000 ms — the ComfyUI tab may be backgrounded or frozen` @ 20 424 ms — the reported message verbatim |
| tab fully silent | fails `within 30000 ms` @ 30.4 s, bounded | fails @ 20 419 ms |

Taxonomy, worked class by class:

**1. Two states collapsing.** Checked the only signal the change relies on: the reply-timeout verdict. "Frozen" vs "slow-but-alive" stays inherently indistinguishable, but the change *narrows* the collapse (fewer live tabs misread as frozen) rather than introducing one. No failure path and legitimate negative became identical. `check:unknown-collapse` — 0 new collapses.

**2. Identity surviving a transformation.** `graph_get_errors` is in `BRIDGE_READONLY_CMDS`, so a mid-command drop parks and resumes on the *same* tabId re-hello; the resume clamps to `remaining` (ui-bridge.ts:4815) so a re-dispatch cannot extend the caller's deadline. Replies match on rid+socket. Nothing is stamped (`GRAPH_CMD_EFFECT: inert`), so there's no stale-uuid attribution to widen.

**3. Guard on the wrong pair.** Named the pairs out loud: the tests compare 30 000 vs `BRIDGE_READ_DEFAULT_TIMEOUT_MS` (imported live, not literal) and 30 000 vs the panel's 18 000 + margin. Those are the pairs that bind. The pre-dispatch guard in play (`graphCmdBlockedByRunningPrompt`) is untouched and still short-circuits the dominant frozen case before the wait starts.

**4. Test pins the fix.** Mutated the call site back to `ctx.call({cmd:"graph_get_errors"})`: **all 5 tests fail**, including the source-scan rider. Not decoration.

**5. Fix worse than the bug.** Hunted for an outer deadline shorter than 30 s and found none — `panel-mcp-http.ts` sets no request timeout, `opts.timeoutMs` always wins in `send()` (ui-bridge.ts:4352) with no clamp, and `isTransientReconnectError` deliberately *excludes* reply timeouts, so there is no 30 s + 30 s double-wait. Cost is 10 extra seconds on a truly dead tab — stated, bounded, and the acknowledged trade.

**6. Prose overstating.** Verified every load-bearing claim against the panel repo's `origin/main`, not from the comment: `GET_ERRORS_TOTAL_BUDGET_MS = 18000` ✓, "~14.5 s on a real Windows install (#610)" ✓ verbatim in that file's header, "sized against the orchestrator's 20 000 ms read default" ✓ (the panel header says exactly that). Confirmed all five #599 siblings really do forward the constant (lines 2063, 11156, 11179, 11753, 11867, 12541). The cross-repo invariant holds: the panel's `browser_tests/unit/get-errors-budget.test.mjs` asserts 18 000 < 20 000, and this diff changes neither number — so the old-orchestrator safety floor the comment claims to preserve is genuinely preserved.

**7. Reachable, not dead on arrival.** Proved by execution, not grep: a probe using the real `makePanelToolCtx` (not the test's stub ctx — that hop was untested) confirmed `timeoutMs: 30000` arrives at `bridge.send`. One dispatch site exists repo-wide, and the source rider fails if a second is added.

`vitest run src/__tests__/orchestrator/` — 193 files / 3129 tests pass. `tsc --noEmit` clean. `check:vocabulary` clean. No prettier config or CI format check, so the 108-char call line won't be reflowed out from under the source-scan test. Probe files removed; worktree clean.

artokun:
Verified this merged cleanly — `panel-tools.ts:12505` now carries `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS`, and panel#1493 auto-closed one second after the merge via the cross-repo ref.

**One correction to the reasoning, because the stated reason is checkable and wrong in a way that could license a future mistake.**

The report gives this for not rewording the timeout text:

> `isReplyTimeoutResult` (#1468, `panel-tools.ts:8261`) matches that exact string as a bridge-owned predicate; rewording it would break the very mechanism that exists so nobody keys on message text.

`isReplyTimeoutResult` does **not** match any string — it reads a symbol:

```js
function isReplyTimeoutResult(res: ToolResult): boolean {
  return res?.isError === true && (res as Record<symbol, unknown>)[REPLY_TIMEOUT_RESULT] === true;
}
```

That is the whole point of #1468: two codex rounds rejected deciding this from message text, so the bridge tags the error object instead.

**The conclusion is still right — just for a different function.** `isReplyTimeoutError` (`:9745`) genuinely does whole-message-anchor on that sentence, including the trailing clause:

```js
/^Panel tab .+? did not reply to "[^"]*" within \d+ ms — the ComfyUI tab may be backgrounded or frozen(?![\s\S])/
```

and it has five callers (`:9243`, `:10356`, `:12790`, `:12924`). The anchoring is deliberate — its comment explains that a transport error merely *embedding* the phrase (`relay failed: Panel tab … did not reply …; reconnecting`) must not be mistaken for a recoverable card-reply timeout, which matters for the consent gate.

So rewording that message would break `isReplyTimeoutError` and its five call sites, not `isReplyTimeoutResult`. Worth fixing in the record because a future reader who follows the citation finds a symbol check, concludes the text is free to change, and breaks the consent-gate path.

Nothing to change in the diff — the decision not to touch the message was correct.

PR #1868, opened 2026-08-20
fix(panel#1494): a stale tab advertisement stops being reported as a restored graph binding

artokun:
Fixes the underlying cause reported in artokun/comfyui-mcp-panel#1494.

## What was reported

A live-canvas `panel_add_node` was refused with

```
workflow instance mismatch: this command was issued for workflow instance 4808c797…,
but the tab it routed to has since reported a different active workflow (caf45251…).
```

and `panel_set_workflow_target({mode:"current"})`, called immediately afterwards —
the recovery that refusal itself prescribes — answered `graph_binding:"bound"`,
`graph_binding_status:"already_current"`, naming `4808c797…`. Two answers about one
session, contradicting each other, with no user-visible tab switch between them.

## Why (measured, not inferred)

The reporter was right about the symptom and the mechanism is one step behind the one
they proposed. This is **not** the panel's fence — that refusal reads *"and the active
canvas reports X"*. This is the orchestrator's **dispatch-time agreement gate**
(#1656, `UiBridge.send`), which is where *"the tab it routed to has since reported"*
comes from and which refuses **before** anything is written.

The gate and its own recovery compare **different facts**:

| | compares |
|---|---|
| the gate (`send`) | the session's stamp vs. the identity the **routed tab last advertised** in a hello (`tabCommandWorkflowUuid`) |
| the recovery (`rebindWorkflowFence`) | the session's stamp vs. the **live canvas** it reads back with `workflow_list` |

When the **advertisement** is the stale side those two disagree by construction. The
live canvas equals the stamp, so the rebind takes the `already_current`
short-circuit — which returns *before* `refreshWorkflowUuid`, and whose
`corroborateTabStamp` refuses outright when the values differ (by design, #1646) — so
it repairs nothing and reports success. Meanwhile the gate keeps refusing every fenced
command against an advertisement nothing has touched. Only the panel's own drift
re-hello clears it, and the recovery said it had. The tab-advertisement side is also
invisible to `tabGraphMutationCapability`, which is the probe every fence-rebind reply
renders its verdict from — so a routable, open, fence-advertising, correctly-stamped
session answered `canMutate: true` while nothing could be dispatched. That is the
`graph_binding:"bound"` in the report.

## What changed

**The gate does not move.** Both refusals are byte-identical, no fence is adopted,
re-pointed or widened, and the read-only probe stays read-only. What changes is that
the gate now has **one implementation** (`stampTargetVerdict`), and the capability
probe the recovery reports from asks it:

- `panel_set_workflow_target({mode:"current"})` reports `not_recovered` instead of
  `bound` while the gate is refusing, and names a remedy that can clear it
  (`panel_open_workflow`, which proves an identity for the tab under its **current**
  id). Deliberately **not** `reads_only`: this gate covers `graph_*` reads too
  (`requiresStampTargetAgreement`), so that would be a false consolation.
- #1330's mismatch diagnosis stops prescribing `RETRY THIS EXACT CALL ONCE` in the one
  state where the retry is compared against the same unchanged value. It says which
  comparison is refusing instead.
- the new cause is its own value (`target_disagreement`), never folded into
  `no_identity`: there **is** an identity, and the remedy is the opposite one.

The new check is **last** in `tabGraphMutationCapability` because it presupposes every
condition above it, and an unreadable side stays a non-refusal — a missing value is not
divergence, unchanged from #1656.

### Considered and rejected

Auto-adopting the live-canvas uuid onto the routed tab (the reporter's "transparently
refresh the fence") would clear the refusal in one call — and would re-open #1656 in
the mirror case. There, the hello advertising `B` is the panel's *positive assertion of
a switch* while the frontend identity `workflow_list` reads back can still lag on `A`;
preferring the read over the hello would drag the advertisement back to `A` and let a
command land on the previous canvas, which is the corruption the gate exists to
prevent. The two states are structurally identical at that instant and only a timeout
separates them, so nothing here writes.

## Verification

- `src/__tests__/services/stamp-target-disagreement-recovery.test.ts` — new; drives the
  **real** `UiBridge` over a real socket and the **real** tool handlers end to end: the
  refusal in the reported words, the capability verdict, the `panel_set_workflow_target`
  reply, and the `panel_add_node` mismatch diagnosis. Also pins the three states that
  must **not** fire it (agreeing session, canonical-id caller, absent stamp).
- Mutation-tested after committing: removing the capability check kills 4 tests,
  reverting `okBinding` kills the rebind-reply test, breaking the `gateStillRefuses`
  discriminator kills the diagnosis test, and swapping the verdict's argument order
  kills the source-level wiring test.
- `carried-stamp-wiring.test.ts` updated, not loosened: its old anchors named the
  inlined gate. It now pins both halves of the same property — the helper reads
  provenance from its **routed-tab** parameter, and `send()` passes
  `(opts.tabId, conn.tabId)` in that order.
- Full suite: **586 files, 10804 passed, 4 skipped, 0 failed**. `npm run lint` and
  `npm run check:vocabulary` clean.
- **No Playwright run**: this change touches no panel file (`comfyui-mcp-panel` is
  untouched), so there is nothing the sidebar renders differently. Stating that rather
  than implying coverage I do not have.

## Gate

codex was unavailable (account exhausted until 2026-08-19 per the standing note; still
exhausted at time of writing); gated by an independent adversarial review instead
(`.claude/autopilot/claude-gate.mjs`), which spawns a fresh reviewer that has not seen
this reasoning. Verdict recorded in a comment below.

### DIFF
--- a/src/__tests__/orchestrator/carried-stamp-wiring.test.ts
+++ b/src/__tests__/orchestrator/carried-stamp-wiring.test.ts
@@ -105,10 +105,16 @@ describe("#1656 — carried-stamp provenance is WIRED, not merely available", ()
   it("SOURCE: the gate consults the predicate, and the recovery probe stays exempt", () => {
     const src = bridgeSrc();
     // The gate reads provenance for the CONNECTION's canonical id — the tab the
-    // frame would actually land on — not for the caller's address.
-    expect(src).toContain("this.isCarriedTabStamp?.(conn.tabId) === true");
+    // frame would actually land on — not for the caller's address. #1494 moved the
+    // comparison into one shared helper, so the property is pinned in BOTH halves:
+    // the helper reads provenance from its ROUTED-tab parameter…
+    expect(src).toContain("this.isCarriedTabStamp?.(routedTabId) === true");
+    // …and send() passes the caller's address and the connection's canonical id in
+    // that order. Swapping them would read provenance for the wrong tab and is
+    // invisible to a behavioural test that installs its own predicate.
+    expect(src).toContain("this.stampTargetVerdict(opts.tabId, conn.tabId)");
     // The refusal is pre-dispatch and typed, like #1682's.
-    const idx = src.indexOf("if (issuedFor && landedOn && carried) {");
+    const idx = src.indexOf('if (verdict.refuses && verdict.kind === "carried") {');
     expect(idx, "carried refusal branch not found").toBeGreaterThan(-1);
     const branch = src.slice(idx, idx + 1200);
     expect(branch).toContain("markDispatched(");
--- a/src/__tests__/services/stamp-target-disagreement-recovery.test.ts
+++ b/src/__tests__/services/stamp-target-disagreement-recovery.test.ts
@@ -178,6 +178,26 @@ describe("a stale tab advertisement is REPORTED, not papered over (#1494)", () =
     sock.close();
   });
 
+  it("a CANONICAL-ID caller is exempt here because `send` exempts it — the same short-circuit", async () => {
+    // The one state that separates the two: `send()` never runs the gate for a caller
+    // addressing the connection by its canonical id, so the capability probe must not
+    // either — both answers would come from ONE lookup, which cannot be a
+    // disagreement. Carried is set precisely because it is the arm that would
+    // otherwise fire on a pair that IS equal, making this the input that catches a
+    // missing short-circuit rather than one that merely happens to agree.
+    advertised.set(TAB, LIVE);
+    sessionStamp = LIVE;
+    carried = true;
+    const sock = await connectPanel(TAB);
+    await waitFor(() => expect(bridge.tabs().map((t) => t.tab_id)).toContain(TAB));
+    expect(bridge.tabGraphMutationCapability(TAB)).toEqual({ known: true, canMutate: true });
+    // …and that IS what send does: addressed canonically, the command dispatches.
+    received.length = 0;
+    await expect(bridge.send({ cmd: "graph_add_node" }, { tabId: TAB })).resolves.toBeTruthy();
+    expect(received.map((f) => f.cmd)).toEqual(["graph_add_node"]);
+    sock.close();
+  });
+
   it("does NOT fire when the two agree — the settled session is untouched", async () => {
     advertised.set(TAB, LIVE);
     sessionStamp = LIVE;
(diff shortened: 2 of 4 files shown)


artokun:
## Gate verdict: **SHIP** (exit 0)

codex was unavailable (account exhausted per the standing note), so this was gated by
`.claude/autopilot/claude-gate.mjs` — a fresh adversarial reviewer that had never seen
the implementation reasoning, handed only the diff and the repo's seven-class review
taxonomy. **Disclosing the substitution is the condition it was accepted under.** It is
weaker than codex in one way and this run is not: it ran the code rather than reading it.

What it verified by execution, not inspection:

- **class 3 (guard asks the wrong pair)** — it deleted the `same_address` early return
  and found the capability probe then reporting `canMutate:false` for a caller `send()`
  skips the gate for entirely. Guard confirmed load-bearing.
- **class 6 (prose overstating)** — it ran the prescribed remedy against the reported
  state: `panel_open_workflow` moves the advertisement, the capability flips to
  `canMutate:true`, and `graph_add_node` dispatches. The remedy in the new message works.
- **class 1 (two states collapsing)** — it drove the *carried* arm end to end and
  confirmed it never reaches the new prose: the corroborator clears the carry before the
  capability is read, so that path keeps its existing "retry once", and the retry works.
- **class 5 (fix worse than the bug)** — it walked all three consumers of the widened
  `canMutate:false` and confirmed none loops or hard-fails on it.
- 4 of 5 of its own mutations killed; regression run green.

### Findings acted on

Neither was a P0/P1 and the verdict was SHIP either way; both are addressed in 67e5c85:

1. **The one surviving mutation** — deleting `same_address` left the suite green,
   because my fixture made the two values equal anyway. Now pinned with the input that
   actually separates them (an equal-but-*carried* pair addressed canonically), and the
   mutation is killed.
2. **A marginally wide comment** — "no working graph tools at all" ignored the
   canvas-independent ops `requiresStampTargetAgreement` exempts. Narrowed.

Full suite after those: **586 files, 10805 passed, 4 skipped, 0 failed**; `npm run lint`
and `npm run check:vocabulary` clean.

### One thing worth recording

The reviewer reported three of its own mutations first reading as *survived* — its `\n`
patterns never matched, because these files are CRLF. It caught that and re-ran
newline-agnostically. A mutation that never applied reads exactly like a mutation that
was survived, and on this repo that failure mode is the default rather than the
exception.

Issue #1425, opened 2026-08-11
panel_connect/panel_set_widget reject colon-qualified node ids (^-?\d+$), making unpacked-subgraph graphs uneditable


artokun:
### What happened / root cause

Two of the user's workflows had ROOT-LEVEL node ids containing colons — e.g. `120:104`, `120:113:78`, `263:78` — left behind by a previously unpacked/inlined subgraph. These are genuine root nodes (`panel_query_graph {fields:'ids'}` on `viewing: root` returns them, and `panel_graph_outline` renders them normally).

But every WRITE tool refuses them. `panel-tools.js` validates node ids with:

```js
const nodeId = () => z.union([z.number().int(), z.string().regex(/^-?\d+$/, "a node id must be an integer")]).transform((v) => (typeof v === "number" ? v : Number.parseInt(v, 10)));
```

So the READ tools happily hand me ids that the WRITE tools then reject:

```
MCP error -32602: Input validation error: Invalid arguments for tool panel_connect: [
  { "code": "invalid_format", "pattern": "/^-?\\d+$/", "path": ["from_node_id"],
    "message": "a node id must be an integer" } ]
```

Same for `panel_set_widget`. In one workflow this made **18 of 21 nodes uneditable** — I could read the graph perfectly but could not connect, set a widget, or enter a subgraph anywhere in it.

Note the `.transform()` is a second latent bug: even if the regex were widened, `Number.parseInt("263:78", 10)` silently yields `263` — it would edit the WRONG NODE rather than error.

### Steps to reproduce

1. Open a workflow whose root graph contains node ids with colons (created by unpacking/inlining a subgraph).
2. `panel_graph_outline` / `panel_query_graph` — ids like `263:78` come back fine.
3. `panel_connect(from_node_id: 4, to_node_id: "263:78", to_input: "pixels")` → InputValidationError.
4. `panel_set_widget(node_id: "120:117", widget: "steps", value: 40)` → same.
5. `panel_enter_subgraph(120)` also fails — "No node with id 120" — because `120` is not a real subgraph node, the colon is just a leftover id shape.

### Exact error (scrubbed)

```
MCP error -32602: Input validation error: Invalid arguments for tool panel_set_widget: [
  {
    "origin": "string",
    "code": "invalid_format",
    "format": "regex",
    "pattern": "/^-?\\d+$/",
    "path": [ "node_id" ],
    "message": "a node id must be an integer"
  }
]
```

### Workaround I had to use

[...]

PR #1874, opened 2026-08-20
fix(panel#1497): panel_run takes the node id its own tools printed

### LINKED ISSUE #1497: panel_run rejects string node IDs returned by canvas tools
### What happened
`panel_run({to_node_id:"11"})` failed with `MCP error -32602: Input validation error: Invalid input: expected number, received string at to_node_id`.

### Steps to reproduce
1. Call panel_add_node; returned node id is serialized as a string (e.g. `"11"`).
2. Pass that documented returned id directly to panel_run as `to_node_id`.

### Expected
panel_run should accept the node-id representation returned by other panel tools, or the returned IDs should be numeric consistently.

### Workaround
Pass the same ID as a number (`11`), which queued successfully.

### Environment
Windows 11; ComfyUI 0.33.2; frontend 1.49.6; RTX 5070 12GB; comfyui-mcp 0.52.24; panel 0.15.11.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-terra` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #427

artokun:
Fixes artokun/comfyui-mcp-panel#1497

## The report, and what it actually is

`panel_run({to_node_id:"11"})` came back with
`MCP error -32602: Input validation error: Invalid input: expected number, received string at to_node_id`.
The reporter had just taken `"11"` out of `panel_add_node`'s own reply — the round trip
the tool descriptions tell an agent to make.

That is not a new class of bug. It is **#845**, unfinished. The graph readers hand out ids
as strings (`summarizeNode` returns `id: node.id`, and LiteGraph ids are strings on the
modern ComfyUI frontend), so #845 made every panel tool accept back an id it had printed.
Twenty-seven node-id arguments adopted the `nodeId()` helper. `panel_run`'s `to_node_id`
kept `z.number().int()` and was missed — on `origin/main` it is the **only** node-id
argument in `panel-tools.ts` still spelled that way:

```
$ grep -c 'nodeId()' src/orchestrator/panel-tools.ts     # 27
$ grep -n  'node_id: z\.number' src/orchestrator/panel-tools.ts
12137:        to_node_id: z                                # ← the one
```

The linked #427 (`panel_create_group` string-vs-number set arithmetic) is the same root
cause, already closed by #845's sweep. A canonical validator is adopted **per call site**;
this call site never adopted it.

## The fix

`to_node_id` now takes either spelling and normalizes to the number the wire has always
carried: `11` and `"11"` both mean node 11.

Keeping the wire numeric is load-bearing, not cosmetic. Two branches inside `panel_run`
gate on `typeof args.to_node_id === "number"` — #772's certified stamp-race re-issue and
#468's run ticket (`toNodeId`) — so a `to_node_id` that stayed a string would have queued
fine and silently skipped both. There is a test asserting the forwarded value is a
`number`, not merely something that prints as one.

## What deliberately did NOT widen

The obvious move is to reuse `nodeId()`. That is the wrong helper here, and the reason is
the whole point of this PR.

`nodeId()` also admits the **subgraph-qualified** shape (`"120:104"`, #1425) and forwards
it *verbatim as a string*, which is correct for a write that addresses a node by key.
A run-to-node target is not addressed by key — it is an **execution root**, and it is
resolved numerically the whole way down:

- panel `web/js/lib/subgraph-scope.js` → `findNodeInScopes()` does `const num = Number(id)`
  and matches on it. `Number("120:104")` is `NaN`, which matches nothing.
- panel `comfyui-mcp-panel.js` reports `ran_to_node: Number(to_node_id)` and
  `ranToNode: Number(to_node_id)` in its reply.
- the two `typeof === "number"` branches named above.

There is a second reason, visible once you read the panel's resolver: a colon path is the
**output** of run-to-node resolution, not an input. `resolveRunToNodeTarget` returns
`execId` — `"10:15:359"` for a nested target, `String(id)` at root — and *that* is what
becomes ComfyUI's `partial_execution_targets`. It is computed from the live graph.
A caller-supplied `"120:104"` is lexically the same shape but a different thing entirely,
and letting one in through the front door invites it to be mistaken for the other.

So accepting a qualified id would have traded a clear schema refusal for a panel-side
*"node 120:104 was not found on the root graph or in any subgraph"* — about a node that
plainly exists — plus a dropped run ticket. It is still refused, but now **by name and
with a reason** (`RUN_TO_NODE_ID_MESSAGE`) instead of a raw zod type error.

The shape lives in `node-id.ts` as an exported `PLAIN_NODE_ID_PATTERN` (it was already
there as a dead `const PLAIN`) so the plain and qualified patterns cannot drift apart.

**Not done here, on purpose:** making run-to-node work *for* a qualified root id. That is
a real gap — `findNodeInScopes` never adopted the panel's own `canonicalNodeId()` from
#1425 — but it is a panel-side change to the execution-root resolver plus the two
`Number(to_node_id)` reply fields, in a different repo, and nothing in this report asks
for it. Widening the validator without it is exactly the trade this PR refuses.

## Verification

- **Mutation 1** — restore `z.number().int()`: 3 tests fail across 2 files
  (`takes the string spelling panel_add_node returns`,
  `refuses a subgraph-qualified id, and says why`,
  `exposes a batch_count + optional to_node_id schema`).
- **Mutation 2** — drop the numeric normalization (`.transform((v) => v)`): the
  `typeof … === "number"` assertion fails. Both restored, tree clean, fix committed
  *before* mutating.
- **Reachability** — the schema is the production path: both registration sites
  (`registerPanelTools` for the MCP SDK, `createPanelMcpServer` for the in-process
  Anthropic SDK) wrap `d.schema` in `strictPanelSchema()` = `z.object(shape).strict()`
  and hand the handler the parsed value. That is the same mechanism the other 27
  arguments already rely on, and the same one that produced the reporter's -32602.
- `npx tsc --noEmit` — clean.
- `npm run check:vocabulary` — OK (1721 files, no live dead names).
- `npx vitest run` — 10814 passed, 1 failed: `package-hooks.test.ts > every entry in
  \`files\` exists on disk`, `expected [ 'dist' ] to deeply equal []`. Pre-existing on any
  unbuilt worktree; this PR touches no packaging.
- **Browser suite not run.** This change is entirely orchestrator-side (MCP input schema);
  no panel file is touched and nothing the sidebar renders changes. Claiming Playwright
  coverage for it would be claiming coverage I did not take.

## Gate

codex was unavailable and this run was dispatched with the substitute gate named in the
autopilot brief, so the review was done by `.claude/autopilot/claude-gate.mjs` — a fresh
adversarial reviewer that has never seen the reasoning above. I did not re-probe codex
myself; the unavailability is the brief's standing note, not something I measured. This is
disclosed because disclosure is the condition the substitution was accepted under, and
because no self-review ever gates a merge here. Verdict recorded in a comment below.

### DIFF
--- a/src/__tests__/orchestrator/node-id-and-canvas-args.test.ts
+++ b/src/__tests__/orchestrator/node-id-and-canvas-args.test.ts
@@ -186,3 +186,68 @@ describe("#845(3): which args each action consumes", () => {
     expect(ignoredCanvasArgs("teleport", { node_id: 42, scale: 1 })).toEqual([]);
   });
 });
+
+// #1497 — the ONE argument #845 never reached.
+//
+// The reporter did exactly what the tool descriptions tell an agent to do: read
+// the id back out of panel_add_node (`"11"`, because summarizeNode returns
+// `id: node.id` and LiteGraph ids are strings on the modern frontend) and hand it
+// to panel_run as `to_node_id`. That call died in zod with
+// `expected number, received string` — the identical failure #845 fixed for the
+// other 27 node-id arguments and #427 hit on panel_create_group. A canonical
+// validator is adopted per CALL SITE; this call site was missed.
+//
+// The wire is unchanged: what the panel receives is still the NUMBER it has
+// always received. That matters beyond tidiness — two branches in panel_run gate
+// on `typeof args.to_node_id === "number"` (#772's stamp-race re-issue and #468's
+// run ticket), so a to_node_id that stayed a string would silently skip both.
+describe("#1497: panel_run accepts the node id its own tools printed", () => {
+  it("takes the string spelling panel_add_node returns", async () => {
+    const res = await callTool("panel_run", { to_node_id: "11" });
+    expect(res.isError).toBeFalsy();
+    expect(sent).toHaveLength(1);
+    expect(sent[0].cmd).toBe("graph_run");
+    // Normalized to the number the panel has always been sent…
+    expect(sent[0].to_node_id).toBe(11);
+    // …and it really is a number, not the string that merely prints as one:
+    // #772/#468 both branch on typeof === "number".
+    expect(typeof sent[0].to_node_id).toBe("number");
+  });
+
+  it("still takes the numeric spelling", async () => {
+    const res = await callTool("panel_run", { to_node_id: 11 });
+    expect(res.isError).toBeFalsy();
+    expect(sent[0].to_node_id).toBe(11);
+  });
+
+  it("leaves a full run alone — an omitted to_node_id stays undefined", async () => {
+    const res = await callTool("panel_run", { batch_count: 2 });
+    expect(res.isError).toBeFalsy();
+    expect(sent[0].to_node_id).toBeUndefined();
+    expect(sent[0].batch_count).toBe(2);
+  });
+
+  // The half that must NOT widen. A run-to-node target is an EXECUTION ROOT, and
+  // the panel resolves it numerically the whole way down (findNodeInScopes matches
+  // on Number(id); the reply reports ran_to_node: Number(to_node_id)). Accepting
+  // "120:104" here would swap a clear schema refusal for a panel-side
+  // "node 120:104 was not found" about a node that plainly exists — and drop the
+  // #468 run ticket on the way. So it is still refused, but now BY NAME.
+  it("refuses a subgraph-qualified id, and says why", async () => {
+    const res = await callTool("panel_run", { to_node_id: "120:104" });
+    expect(res.isError).toBe(true);
+    expect(sent).toHaveLength(0); // nothing was dispatched
+    const msg = textOf(res);
+    expect(msg).toContain("plain integer node id");
+    expect(msg).toContain("execution root");
+  });
+
+  it("refuses ids that are not ids at all", async () => {
+    for (const bad of ["11px", "1.5", "", "abc", "1:", " 11"]) {
+      sent = [];
+      const res = await callTool("panel_run", { to_node_id: bad });
+      expect(res.isError, JSON.stringify(bad)).toBe(true);
+      expect(sent, JSON.stringify(bad)).toHaveLength(0);
+    }
+  });
+});
--- a/src/__tests__/orchestrator/panel-tools.test.ts
+++ b/src/__tests__/orchestrator/panel-tools.test.ts
@@ -919,12 +919,14 @@ describe("panel-tools: panel_run (run-to-node partial execution)", () => {
       "retry_of",
       "to_node_id",
     ]);
-    // to_node_id is an optional int — accepts a node id, rejects non-numbers,
-    // and (being optional) accepts undefined for a normal full run.
+    // to_node_id takes a node id in EITHER spelling the readers print (#1497 —
+    // 27 and "27" both mean node 27), rejects a non-id, and (being optional)
+    // accepts undefined for a normal full run.
     const toNode = def.schema.to_node_id as {
       safeParse: (v: unknown) => { success: boolean };
     };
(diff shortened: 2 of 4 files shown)


artokun:
## Gate verdict: **SHIP** (exit 0)

codex was unavailable, so this was gated by the substitute independent adversarial reviewer (`.claude/autopilot/claude-gate.mjs`) named in the autopilot brief — a fresh reviewer that had not seen the PR reasoning. Disclosed per the condition that substitution was accepted under. Verbatim verdict:

---

> SHIP
> 
> Worked the taxonomy by executing against both real transports, not reading.
> 
> **1. Two states collapsing** — the pair that mattered is `to_node_id` present-as-string vs absent. Drove a real `tools/call` through `registerPanelTools` (MCP SDK) *and* `createPanelMcpServer` (Anthropic Agent SDK): `{to_node_id:"11"}` reaches the handler as `11`, `typeof === "number"`, wire carries `{"cmd":"graph_run","to_node_id":11}`. Omitted stays `undefined`. Node id `0` is included by the gates (they use `typeof`, not truthiness). Holds.
> 
> **2. Identity surviving a transformation** — `to_node_id` names *which* node. Normalization happens at the schema boundary, before `runCmd` is built; the #772 re-issue reuses that same `runCmd` object, and the #468 ticket reads the same value. No capture-before-transform gap. Holds.
> 
> **3. Guard comparing the wrong pair** — the two gates are `typeof args.to_node_id === "number"` at `panel-tools.ts:12321` and `:12437`. The commit's whole safety claim rests on those still firing, so I verified it on the wire rather than by reading: both transports deliver a number, so both fire. Had either SDK passed raw args, this diff would have newly admitted a string that skipped both — that was my strongest hypothesis and it is refuted by execution. Holds.
> 
> **4. Test pins the fix** — reverted both source files to `HEAD~1` with tests intact: 2 of 21 fail (the string-spelling acceptance and the named-refusal message). Restored; tree clean. Pins.
> 
> **5. Fix worse than the bug** — the shape does not widen: `"120:104"` is still refused, nothing dispatched (`sent: []`) on both paths. Hostile sweep through the live transport — `"1e3"`, `"+5"`, `" 11"`, `"11 "`, `"1_0"`, `"0x10"`, `1.5`, `true`, `null` all rejected; `"007"`→7, `"-0"`→0. On the MCP path the advertised schema *gains* information (`anyOf` with `pattern:"^-?\\d+$"` plus the full description). The Anthropic path renders `{"anyOf":[{"type":"number"},{"type":"string"}]}` with no description — but that transport strips descriptions and constraints from *every* prop (`batch_count` → bare `{"type":"number"}`), so it is pre-existing, not caused here.
> 
> **6. Prose overstating** — the new description claims `11` and `"11"` both mean node 11, and that a qualified id is refused. Both verified on the wire, including the refusal text reaching the caller as `-32602`.
> 
> **7. Reachable** — `runToNodeId()` is live at `panel-tools.ts:12178`, reached through both registration paths; deleting the fix turns the tests red.
> 
> Full cluster green: 642 tests across `node-id`, `panel-tools`, `panel-tools-strict-schema`, `panel-run-backpressure`, `panel-run-reconnect-respell`, `panel-retry-identity`, `run-correlation-prior`, `run-late-ack-reconcile`, `run-completion-race`. `tsc --noEmit` clean.
> 
> One sub-threshold note, deliberately not a finding: `Number.parseInt` is unbounded, so a 309+ digit id passes the regex, becomes `Infinity`, and `JSON.stringify`s to `null` on the wire — while the numeric branch is bounded to ±2^53−1. It needs an input no reader ever prints, and `normalizeNodeId` has had the identical property across the other 27 node-id arguments on shipped main, so this diff neither introduces nor worsens it.

Issue #504, opened 2026-07-30
UI→API conversion silently replaces newly staged LoadImage media with first cached combo option


artokun:
### What happened / root cause
`get_workflow(format="api")` converted a saved UI workflow whose LoadImage `widgets_values[0]` was `2Dto3D_v2_front_clean.png`, but returned `KENT_concept_00012_.png` (the first stale `/object_info` combo option). The file had just been registered through `stage_output_as_input`, and `get_workflow(format="ui")` correctly returned the new filename.

Root cause: `workflow-converter.js` only treats model extensions as asset filenames. Media values such as `.png` fail `looksLikeAssetFilename()`, so combo validation silently substitutes `comboOpts[0]` instead of preserving the declared media asset. This can make headless execution use the wrong source image.

### Steps to reproduce
1. Save a UI-format workflow containing a `LoadImage` node.
2. Register a new PNG via `stage_output_as_input` while `/object_info` still has an older image combo list.
3. Update that UI node's `widgets_values[0]` to the new PNG and save it.
4. Confirm `get_workflow(format="ui")` shows the new PNG.
5. Call `get_workflow(format="api")`.
6. Observe `inputs.image` is silently replaced by the first cached combo value.

### Exact incorrect result
UI: `widgets_values: ["2Dto3D_v2_front_clean.png", "image"]`
API: `inputs.image: "KENT_concept_00012_.png"`
No LoadImage warning was emitted.

### Fix
Applied locally: yes (single focused patch in the active npx package)
Upstream-only: no

```diff
-const ASSET_FILE_RE = /\.(safetensors|safetensor|sft|ckpt|pt|pt2|pth|bin|gguf|onnx|vae|pkl|npz|yaml)$/i;
+const ASSET_FILE_RE = /\.(safetensors|safetensor|sft|ckpt|pt|pt2|pth|bin|gguf|onnx|vae|pkl|npz|yaml|png|jpe?g|webp|bmp|gif|mp4|webm|mov|mkv|m4v|avi|wav|mp3|flac|ogg|m4a)$/i;
```

Local verification used a minimal `LoadImage` UI graph with a stale one-item combo list; after the patch, `convertUiToApi()` preserved `2Dto3D_v2_front_clean.png` and emitted the honest missing/stale-option warning.

### Environment
- OS: Linux
- ComfyUI: 0.29.0
- Python: 3.12.3
- GPU: NVIDIA GB10, 124546 MB VRAM
[...]

PR #1880, opened 2026-08-20
fix(1869): validate_workflow no longer flags action-button tokens as widget values

### LINKED ISSUE #1869: validate_workflow misaligns custom-node widget values after serialized action buttons
`validate_workflow` reported 43 false type/enum errors for a saved UI workflow that executes successfully on the live canvas. `AMVideoRead` and `AMVideoWrite` serialize front-end action button values such as `browse`, `open_in_explorer`, `copy_path`, and `detect_range` into `widgets_values`, but those buttons are absent from `/object_info`. `comfy_cli.workflow_to_api._schema_widget_pairs` begins pairing the schema at `widgets_values[0]`, so all real inputs shift.

### Steps to reproduce
1. Add `AMVideoRead` or `AMVideoWrite` from comfyui-am-vfx-tools.
2. Configure ordinary widget values and save the UI workflow JSON.
3. Run `validate_workflow` on the saved file.
4. Observe errors such as `frame_mode='open_in_explorer'`, while the live node executes successfully.

### Exact error (scrubbed)
`AMVideoRead.frame_mode: 'open_in_explorer' not in [single, range, all]`; `first_frame: expected INT, got 'copy_path'`; 43 errors total across AMVideoRead/AMVideoWrite.

### Fix
Applied locally: no. Upstream-only: yes. The UI-to-API converter needs a reliable named-widget alignment that skips serialized frontend-only action widgets, or those action widgets must be serialized with names / marked non-serializing. Positional pairing against `/object_info` cannot recover interspersed action widgets such as `detect_range` safely. The failing path is `comfy_cli/workflow_to_api.py::_schema_widget_pairs`; comfy-mcp 0.10.0 wraps comfy-cli 1.16.0 here.

### Environment
Windows 11; ComfyUI 0.33.2; frontend 1.49.6; Python 3.13.12; PyTorch 2.10.0+cu130; RTX 3060 Ti 8 GB; comfyui-mcp 0.52.25; panel 0.15.22; comfy-mcp 0.10.0; comfy-cli 1.16.0.

artokun:
Fixes #1869

`create_workflow (action:"validate")` on a saved UI graph from comfyui-am-vfx-tools reported 43 false type/enum errors (`frame_mode='open_in_explorer'`, `first_frame='copy_path'`) for a workflow that executes on the live canvas.

`AMVideoRead` / `AMVideoWrite` serialize frontend action buttons (`browse`, `open_in_explorer`, `copy_path`, and an interspersed `detect_range`) into `widgets_values`. Those buttons are absent from `/object_info`, so positional pairing from slot 0 shifted every real input.

## What changed

- **`convertUiToApi`** skips serialized extras that cannot belong to the current schema widget.
- **`validateWorkflow`** converts a UI-format graph through that pairing before checking combos. Previously a UI export made it iterate the graph's TOP-LEVEL KEYS (`nodes`, `links`) and emit garbage — the production caller (`src/tools/workflow-validate.ts`) passes the parsed workflow straight through with no conversion of its own.

## Skipping is deliberately narrow

A skip DROPS a value; a misassignment only moves it. So skipping is gated three ways:

1. **`extras > 0`** — only when the saved row has more values left than the schema has slots left. If the row is not longer, every value still has a home.
2. **Hard type mismatch** (INT/FLOAT/BOOLEAN) — the only signal needing no corroboration, and the one that recovers the *interspersed* `detect_range` that no prefix rule can see.
3. **Corroboration** — a value that demonstrably fits a later strict widget. Where no strict widget exists downstream, nothing can corroborate the guess, so the candidate must be a token actually observed serialized as a button.

Unknowable arity (`COMFY_DYNAMICCOMBO_V3` downstream) disables skipping entirely rather than guessing.

## Review-driven changes since the first push

Two defects were found and fixed in the original cut:

**1. The skip fired on rows that were not longer than the schema.** That is #959's shape — a *reordered* widget, not an extra one — where skipping drops the value instead of misassigning it. The first cut papered over this by rewriting #959's deliberate canary test. Gating on `extras > 0` restores that canary **verbatim**; `live-widget-overlay.test.ts` is no longer in this diff at all.

Removing that gate now fails 8 tests, including **#504 and #517 P0 regression guards** (stale-combo substitution, and the refusal to realign after a stale dynamic parent).

**2. An uncorroborated skip accepted any `/^[a-z][a-z0-9_]*$/` token.** Verified on the rig, not reasoned about:

```
schema: { text: STRING }        row: ["hello", "Extra Thing"]
before:  text = "Extra Thing"   ← silent wrong value
after:   text = "hello"
```

A lone STRING widget holding `hello` with any trailing extra was silently rewritten. That branch also **survived mutation** — nothing in the suite objected. It now requires a known button token where nothing corroborates the skip, and tests pin both directions.

## Evidence

Mutation battery — every mutation must kill a test (`workflow-converter` + `workflow-validator` + `live-widget-overlay`, 99 tests):

| Mutation | Result |
|---|---|
| M1 neuter `skipSerializedActionWidgets` entirely | **4 failed** |
| M2 remove the numeric/boolean mismatch skip | **1 failed** |
| M3 remove the `extras > 0` gate | **8 failed** (incl. #504/#517 P0 guards, #959) |
| M4 `validateWorkflow` no longer converts UI→API | **1 failed** |
| M6 drop the known-token requirement | **1 failed** |
| M7 uncorroborated path always refuses | **1 failed** |

No survivors. (M6 initially read as a survivor because the harness's `git checkout` reverted the then-uncommitted fix — re-run after committing.)

- `tsc --noEmit` clean
- Targeted suites: 99 passed
- `npm run check:vocabulary` OK — this was the **original CI failure**: two comments named the retired `validate_workflow` tool. Reworded to `create_workflow (action:"validate")`.
- Full suite: 2 pre-existing failures in `tool-surface-filter` / `vocabulary` that reproduce with these files reverted to `origin/main`, vary run-to-run (4 vs 2), and pass in isolation — a local ordering artifact, not this branch. Main's CI is green.
- Rebased onto current `origin/main` (`2782319`); main touched none of these files.

## Known limitation

An all-STRING schema whose buttons are not in the known-token set is still not recovered. That is deliberate: positionally it is indistinguishable from an ordinary value, and per #959 a wrong skip is worse than a wrong pairing.

### DIFF
--- a/docs/tools/workflow-authoring.mdx
+++ b/docs/tools/workflow-authoring.mdx
@@ -32,7 +32,7 @@ Author and check ComfyUI workflow JSON. Driven by the `action` parameter:
   action:"create" — Template parameters; recognized keys depend on the template. txt2img: checkpoint, positive_prompt, negative_prompt, width, height, steps, cfg, seed, sampler_name, scheduler. img2img/inpaint add image_path (and mask_path for inpaint) and denoise. upscale adds upscale_model. Unknown keys are ignored; omitted keys use template defaults.
 </ParamField>
 <ParamField path="workflow" type="string | object">
-  ComfyUI workflow JSON (as a JSON string or object). REQUIRED for action:"modify" and action:"validate" — API format for "validate".
+  ComfyUI workflow JSON (as a JSON string or object). REQUIRED for action:"modify" and action:"validate". action:"validate" accepts API format or a saved UI export (nodes[]/links[]).
 </ParamField>
 <ParamField path="operations" type="object[]">
   action:"modify" (REQUIRED) — Array of operations to apply in order. Each has an 'op' field: set_input, add_node, remove_node, connect, or insert_between
--- a/src/__tests__/services/live-widget-overlay.test.ts
+++ b/src/__tests__/services/live-widget-overlay.test.ts
@@ -71,13 +71,17 @@ const nameKeyedState = () => ({
 });
 
 describe("#959: the bug, reproduced", () => {
-  // This is the failure the three reports describe. It is asserted deliberately:
-  // if this test ever stops failing on its own, the positional path changed and
-  // the overlay below is no longer testing what it claims to.
-  it("without the capture, the wildcard STRING is emitted as the seed", () => {
+  // #1869 taught the positional path to skip a STRING sitting on an INT, so the
+  // seed now gets 101 instead of the wildcard token. The FRONTEND still serializes
+  // wildcards BEFORE seed, so that skipped token is never offered to `wildcards`
+  // — the combo falls back to its default. Named capture remains the only path
+  // that can recover a reordered widget the schema has already walked past.
+  // If wildcards starts matching without a capture, the overlay below is no
+  // longer testing the disagreement it claims to.
+  it("without the capture, the reordered wildcard never lands on its own widget", () => {
     const { workflow } = convertUiToApi(serializedCanvas() as never, OBJECT_INFO);
-    expect(workflow["296"].inputs.seed).toBe(WILDCARD);
-    expect(workflow["296"].inputs.seed).not.toBe(101);
+    expect(workflow["296"].inputs.seed).toBe(101);
+    expect(workflow["296"].inputs.wildcards).not.toBe(WILDCARD);
   });
 });
 
@@ -97,12 +101,10 @@ describe("#959: the capture fixes it", () => {
     expect(workflow["296"].inputs.wildcard_text).toBe("a cat");
   });
 
-  // The shift compounds: `seed` is an INT with control_after_generate, so the
-  // positional pass consumes a PHANTOM slot after it. Having already taken the
-  // wildcard string as the seed, it then swallows the real seed as that phantom
-  // and runs out of values — so `wildcards` is never assigned at all and falls
-  // back to the node default. One inverted pair costs BOTH widgets, and nothing
-  // in the output says so.
+  // The frontend serializes wildcards BEFORE seed. The positional pass now
+  // skips that STRING off the INT seed (#1869) and assigns 101, but the skipped
+  // token is gone — `wildcards` is never offered it and falls back to the node
+  // default. Named capture is what restores both widgets.
   it("the trailing widget is lost entirely, not merely misassigned", () => {
     const before = convertUiToApi(serializedCanvas() as never, OBJECT_INFO);
     expect(before.workflow["296"].inputs.wildcards).not.toBe(WILDCARD);
--- a/src/__tests__/services/live-widget-overlay.test.ts
+++ b/src/__tests__/services/live-widget-overlay.test.ts
@@ -71,17 +71,13 @@ const nameKeyedState = () => ({
 });
 
 describe("#959: the bug, reproduced", () => {
-  // #1869 taught the positional path to skip a STRING sitting on an INT, so the
-  // seed now gets 101 instead of the wildcard token. The FRONTEND still serializes
-  // wildcards BEFORE seed, so that skipped token is never offered to `wildcards`
-  // — the combo falls back to its default. Named capture remains the only path
-  // that can recover a reordered widget the schema has already walked past.
-  // If wildcards starts matching without a capture, the overlay below is no
-  // longer testing the disagreement it claims to.
-  it("without the capture, the reordered wildcard never lands on its own widget", () => {
+  // This is the failure the three reports describe. It is asserted deliberately:
+  // if this test ever stops failing on its own, the positional path changed and
+  // the overlay below is no longer testing what it claims to.
+  it("without the capture, the wildcard STRING is emitted as the seed", () => {
     const { workflow } = convertUiToApi(serializedCanvas() as never, OBJECT_INFO);
-    expect(workflow["296"].inputs.seed).toBe(101);
-    expect(workflow["296"].inputs.wildcards).not.toBe(WILDCARD);
+    expect(workflow["296"].inputs.seed).toBe(WILDCARD);
+    expect(workflow["296"].inputs.seed).not.toBe(101);
   });
 });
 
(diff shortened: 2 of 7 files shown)


artokun:
Re-engaging the stale #1869 worktree/PR now. I will verify the production caller, revert-kill coverage, current-main compatibility, and CI before requesting review.

artokun:
Re-engaged the stale worktree rather than opening a rival PR. Branch force-pushed after a clean rebase onto current `main` (`2782319` — main touched none of these five files), so CI is running against today's main, not the 17:25Z tree.

**Why the earlier CI was red:** `check:vocabulary`, not a test failure. Two comments named the retired `validate_workflow` tool; reworded to `create_workflow (action:"validate")`. Gate passes locally now.

**Two real defects were found in the first cut and fixed** — both are the "silent wrong value" class, which is worse than the bug being fixed:

1. The skip fired even when the saved row was **not** longer than the schema. That is #959's *reordered-widget* shape, where skipping DROPS the value rather than moving it. The first cut concealed this by rewriting #959's deliberate canary. Gating on `extras > 0` restores that canary verbatim — `live-widget-overlay.test.ts` is no longer in the diff. Removing the gate now fails 8 tests, including the **#504 and #517 P0 guards**.

2. An uncorroborated skip accepted any `/^[a-z][a-z0-9_]*$/` token, so `{text: STRING}` with row `["hello", "Extra Thing"]` silently produced `text = "Extra Thing"`. Reproduced on the rig. That branch had also **survived mutation** — no test objected — so it was pure untested heuristic. It now requires a token actually observed serialized as a button where nothing downstream corroborates the skip.

Mutation battery, all six kill a test, no survivors: M1 neuter the skip → 4 failed · M2 remove numeric mismatch skip → 1 · M3 remove `extras > 0` → 8 · M4 validator stops converting → 1 · M6 drop known-token requirement → 1 · M7 uncorroborated path always refuses → 1.

Full-suite note: `tool-surface-filter` / `vocabulary` show 2–4 failures locally that **reproduce with these files reverted to `origin/main`**, vary run-to-run, and pass in isolation — a local ordering artifact, not this branch. Main's CI is green.

Independent adversarial review is running now (read-only, diff supplied inline). **Holding the merge until that returns clean and CI is green** — specifically asking it to attack whether the skip can ever eat a legitimate value, the `extras`/phantom-slot arithmetic, and whether routing UI graphs through the converter loses the "unknown node type" error for uninstalled custom nodes.

artokun:
## Round-2 update: CI diagnosis + independent review findings fixed

**Why run 32411756026 was red.** Not a test failure. The `Tool Reference is current (docs:gen must be a no-op)` gate: this branch edits the `workflow` schema description in `workflow-compose.ts`, and `docs/tools/workflow-authoring.mdx` is generated from it. Ran `npm run docs:gen`; the generated diff was exactly the one line CI printed, in that one file, and is now committed. Windows had already passed; ubuntu/macOS failed only on this gate.

**The independent Codex review returned NEEDS-REVISION, and it was right.** Four of its five findings reproduced on the rig, and I found a fifth of the same kind:

| Row | Schema | Before | Correct |
|---|---|---|---|
| `["my_path","label","detect_range",12]` | path, label, first_frame | `path="label"` | `path="my_path"` |
| `["my_model","detect_range",12]` | ckpt_name, first_frame | `ckpt_name="detect_range"` | `ckpt_name="my_model"` |
| `[0,"browse"]` | enabled BOOLEAN | `enabled="browse"` | `enabled=0` |
| `["1024","browse"]` | width INT | `width="browse"` | `width="1024"` |

All four are **silent wrong values** — a plausible graph carrying someone else's data, which is worse than the misalignment being fixed. The second violates #361 (a stale asset value must be preserved so it surfaces as a missing-asset error); its warning then named the *button token* as the value the user had declared. The last two are regressions the first cut introduced: ComfyUI itself coerces `"1024"` to INT and `0`/`1` to BOOLEAN.

Root cause was the corroboration mechanism: it matched the general shape of an identifier, and most ordinary values match `/^[a-z][a-z0-9_]*$/`. It could tell that *something* in the gap didn't belong, but not *which* — and it always dropped the first. **Deleted, ~90 lines.**

Skipping is now two signals, both still requiring `extras > 0`:
1. the value cannot be the declared numeric/boolean type, or
2. the value is in a **closed** set of known button tokens.

An unrecognised token is left alone. `valueFitsNumericOrBoolean` is deliberately permissive, because a false "doesn't fit" *skips a real value*.

**Finding 5 (missing node types) also confirmed and fixed.** `convertUiToApi` drops a node whose type `object_info` doesn't know, so the validator's own check never saw it — a UI export with an uninstalled custom node reported `valid: true`. A false green on the one thing validation exists to catch, and one **this PR introduced** by routing UI graphs through the converter. `ConversionResult` now carries `missingNodeTypes` structurally, so the validator raises the same authoritative error the API path raises instead of parsing warning prose.

**One existing assertion re-anchored, and I want to flag it explicitly** since rewriting a canary is exactly what I criticised in the first cut. LTXDirector's `frame_rate !== 24` stood in for "properties did not override the positional pass". The BOOLEAN widgets now refuse the STRINGs sitting on them, so the tail realigns and reaches 24 *positionally* — the same answer `properties` would give. The head is still misaligned and nothing can realign it, so the test asserts there instead. The difference from the #959 rewrite: that concealed a dropped value; this is the result becoming **more** correct.

**Mutation battery, round 2 — 8 mutations, no survivors:**

| | Result |
|---|---|
| N1 neuter the skip entirely | 8 failed |
| N2 remove the `extras > 0` gate | 1 failed |
| N3 drop the type-refutation signal | 2 failed |
| N4 drop the known-token signal | 4 failed |
| N5 BOOLEAN stops accepting 0/1 | 1 failed |
| N6 numeric strings stop being accepted | 2 failed |
| N7 validator drops the `missing_node_type` promotion | 1 failed |
| N8 validator stops converting UI→API | 2 failed |

N3 **survived the first time** — re-anchoring the LTXDirector assertion had removed the only coverage of the numeric signal, since `detect_range` is in the known-token set and the vocabulary carried everything else. Two tests were added for it (an *unknown* token on a numeric widget, and a BOOLEAN run realigning), and it dies now.

**Verification:** `tsc --noEmit` clean · `docs:gen` no-op · `check:vocabulary` OK · targeted suites 106 passed · **full suite 591 files / 10962 tests, all green**. Rebased onto current `main`; main has touched none of these files.

Round-2 independent review is running. Still holding the merge until it returns clean and all four required checks are green.

artokun:
Codex review blocker to resolve before merge:

skipSerializedActionWidgets treats the known button strings (browse, open_in_explorer, copy_path, detect_range) as skippable regardless of the current widget spec. A legitimate custom-node COMBO or STRING value can equal one of those strings; when the row has extra values, the converter would skip the real value and silently shift the tail. The current tests prove stale assets and ordinary strings are preserved, but do not prove a legitimate browse/detect_range value in a declared combo/string is preserved.

Please add a regression for that collision and make the skip conditional on the declared spec not accepting the candidate (or otherwise prove the button vocabulary is impossible for that spec). Re-run the focused converter suite, typecheck, and fresh CI. Keep the generated-doc and missing-node diagnostic fixes in the same final head.

Issue #1870, opened 2026-08-20
panel_update_node reports stale generation traceback for Manager update failure


artokun:
### What happened / root cause
`panel_update_node(id:"fal-api", version:"nightly")` failed because the installed registry pack has no nested `.git`, but the tool labeled the previous ByteDance generation traceback as the Manager update traceback. This is misleading and hides the actual `update-git` failure.

### Steps to reproduce
1. Install `fal-api` as a Comfy Registry package without a nested `.git`.
2. Trigger a model execution error.
3. Call `panel_update_node` for `fal-api` nightly.
4. Observe the previous execution traceback returned as the Manager update traceback.

### Exact error (scrubbed)
`Error: Update of "fal-api" FAILED ... Manager traceback:` followed by the previous `FalApiError: Downstream service error (HTTP 500)` execution stack, while the Manager log only says `res.result=False, res.action=update-git`.

### Fix
applied locally: no  upstream-only: yes
The orchestrator should keep Manager task errors separate from ComfyUI execution history. When `update-git` returns false, return the task-specific Manager error/log. Detect non-git registry installs and report/re-route them instead of attaching `last_execution_error`.

### Environment
Windows 11 / ComfyUI 0.33.2 / frontend 1.49.6 / RTX 4080 16 GB / comfyui-mcp 0.52.25 / panel 0.15.23.

artokun:
Claimed by grok-swarm 2026-08-20T17:08:00Z — driving this in worktree `wt-1870` (branch `fix/1870-update-node-stale-traceback`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp/pull/1879 (worktree C:/Users/Artokun/code/comfyui-mcp/.worktrees/wt-1870).

Issue #1866, opened 2026-08-20
panel_free_vram reports success but leaves Ray/CLIP VRAM pinned on device 2


artokun:
### What happened / root cause
panel_free_vram returned freed:true, unload_models:true, free_memory:true, but a fresh connected ComfyUI system_stats read showed GPU 2 still almost fully occupied while GPUs 0, 1, and 3 were mostly free. The active workflow uses Raylight sequence-parallel MiniMax H3 and a MiniMaxH3ParallelCLIPLoader on GPU 2/3. The /free operation appears not to terminate or unload the Ray worker/custom-node allocations.

### Steps to reproduce
1. Run the H3-MSR-S1 workflow with Raylight MiniMax H3 across CUDA 0-3 and the parallel CLIP loader on GPUs 2/3.
2. After the run/error, call panel_free_vram.
3. Observe the successful response: freed:true, unload_models:true, free_memory:true.
4. Call get_system_stats(action:"stats").
5. GPU 2 remains at vram_free=187552724 bytes and torch_vram_free=53072852 bytes, while GPUs 0,1,3 each have about 9.3 GiB free.

### Exact error (scrubbed)
The workflow had previously failed on device 2 with CUDA OOM. No new exception was thrown by panel_free_vram; its success response was misleading.

### Fix
applied locally: no
upstream-only: yes
Precise change needed: panel_free_vram should either terminate/clear Raylight worker and custom-node allocations, or report that only the main ComfyUI model manager was freed and identify remaining pinned devices/processes. It must not report full VRAM freed when a device remains pinned.

### Environment
OS: Linux
ComfyUI: 0.33.1
GPU: NVIDIA GeForce RTX 2080 Ti, 21 GB VRAM, 4 devices
Python: 3.10.16
PyTorch: 2.11.0+cu130
comfyui-mcp: 0.52.1
panel: 0.11.44

artokun:
Claimed by grok-swarm 2026-08-20T17:08:00Z — driving this in worktree `wt-1866` (branch `fix/1866-free-vram-still-pinned`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp/pull/1878 (worktree C:/Users/Artokun/code/comfyui-mcp/.worktrees/wt-1866).

PR #1883, opened 2026-08-20
release: v0.52.28 — free_vram honesty, update_node Manager error

artokun:
Cuts 0.52.28 from `main`.

Carries:

- #1878 / #1866 — `panel_free_vram` no longer reports VRAM freed when a device stays pinned
- #1879 / #1870 — `panel_update_node` reports the Manager update-git error, not a stale generation traceback

Held out: #1880/#1869 ubuntu Build red; #1851/#1847 CONFLICTING; #1839 anthropic 4 CI red; #1697 P1 inflight; #1704 P2 reengage this cycle.

Do **not** squash-merge. Merge-commit (keeps the `v0.52.28` tag on the version commit). Tag is local; push `v0.52.28` after merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -759,6 +759,15 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+## [0.52.28] - 2026-08-20
+
+### MCP
+
+#### Fixed
+- panel_update_node reports the Manager update-git error, not a stale generation traceback (#1879)
+- panel_free_vram no longer reports VRAM freed when a device stays pinned (#1878)
+
+
 ## [0.52.27] - 2026-08-20
 
 ### MCP
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "comfyui-mcp",
-  "version": "0.52.27",
+  "version": "0.52.28",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "comfyui-mcp",
-      "version": "0.52.27",
+      "version": "0.52.28",
       "license": "MIT",
       "dependencies": {
         "@comfyorg/sdk": "^0.1.7",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "comfyui-mcp",
-  "version": "0.52.27",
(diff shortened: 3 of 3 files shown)


artokun:
**Heads-up before this ships: both commits in this cut have open gate findings, and one is a P1 that inverts a success into a failure.**

This is not a request to block the release — the two fixes are real improvements and the defects below are narrower than what they fix. But the cut should go out knowing what is in it.

`v0.52.27..origin/main` is exactly two changes:

| commit | gate | finding |
|---|---|---|
| `ea1e01d` fix(1866) free_vram (#1878) | **NO-SHIP** | filed as #1887 — **P1** |
| `3ba7b7f` fix(1870) update_node (#1879) | **NO-SHIP** | filed as #1888 — P2 |

Both verdicts landed *after* their PRs merged, which is why they are issues rather than review comments.

**The one worth reading before you tag:** `annotateFreeVramAck` (`panel-tools.ts:3053`) takes no `ctx` and reads `getComfyUIBaseUrl()` — the orchestrator's global base — rather than the server the acking tab provably fronts. Its sibling twenty lines away, `settleFreeVramAfterAckTimeout`, gates on `captureRebootHealthBase(ctx)` precisely because "a direct `/free` could aim at a DIFFERENT server than the one the tab was asked to free."

Executed: tab on `:8189`, target set to `:8188`, panel replies `{freed:true}` → the tool queries `:8188`, returns `isError: true`, `"freed": false`, and names **server A's device as if it were B's**, prescribing `panel_restart_comfyui`. A correct free reported as a failure.

Secondary, same issue: `deviceStillPinned` keys on `vram_free`, which is whole-device across all processes. On this box that reads 3.09% free → PINNED while ComfyUI's own torch pool holds 32 MiB. The discriminator (`torch_vram_total` / `torch_vram_free`) is fetched by that same diff at `:2839-2840` and never read. On a machine running several ComfyUI instances — which this one does — instance A's successful free reports failure whenever instance B occupies the card.

Also worth knowing: mutating the base to `http://wrong.example:1` leaves all 12 tests in the changed files passing, so the wrong-target read is untested by construction and can come back green after any fix.

Your call entirely. Shipping is defensible — the previous behaviour was worse in the case #1866 reported. I would just rather this went out as a known trade than as a surprise.

PR #1891, opened 2026-08-20
fix(1567): a queued respawn sees downloads started after the save

### LINKED ISSUE #1567: Deferred tool-session respawn silently orphans downloads started after the save (#1378 gap: at-risk snapshot taken at save time, not respawn time)

`panel_request_secret` (saving `CIVITAI_API_TOKEN`) scheduled a **deferred** tool-session respawn. Its receipt said:

> The comfyui tools re-read that file each time they use this credential, so the tool process already running picks it up — no reload, and no respawn required. **Tool sessions being rebuilt with the new environment: 1 queued for the end of this turn (of 1 live).**

At that moment **zero** downloads were in flight, so the #1378 at-risk warning correctly did not fire. I then started **9 model downloads (~48 GB)** over the next two turns. Several turns later the queued respawn actually fired (surfaced as "🔑 The API token you just provided is now active" together with an MCP disconnect) and **killed all 9 in-flight transfers with no warning at any point**.

The gap is ordering. In `dist/services/panel-secrets.js` (~1465–1485) the at-risk snapshot is deliberately taken **before** the emit:

```js
// #1378 — SNAPSHOT BEFORE THE EMIT, because the emit is where the damage happens.
let atRiskDownloads = [];
if (isAllowedComfyuiSecretKey(trimmed))
    atRiskDownloads = emitComfyuiChange({...});
```

Correct for the **`applied` (immediate)** respawn path. Unsound for the **`scheduled` (queued)** path: for a queued respawn the damage happens at *respawn execution* time, arbitrarily later, so a save-time snapshot is structurally incapable of seeing transfers that start in the window between. `receiptDisclosures` then has nothing to warn about, and the user is told the save is safe while a respawn that will kill their transfers is still pending.

The receipt actively encourages the failure: with `livePickup: true` it prints "the tool process already running picks it up — **no reload, and no respawn required**" on the very same message that reports a respawn queued. I read that as safe-to-proceed and started 48 GB of downloads.

Aggravating factors:

1. **Nothing auto-resumed.** `download_model action:"status"` detected the deaths correctly ("NOT running — the owning process is gone"), but recovery took **9 manual `cancel` calls + 9 manual re-issues**. The new session has everything it needs (job records + `.partial` files) to adopt these itself.
2. **The panel download tray kept showing them as downloading** — indistinguishable from live transfers.

Worth preserving: because the credential was saved *before* the downloads started, cache identity did not change, so re-issuing **resumed from the `.partial` files** (4%/1%/2% start points matched prior progress). The doc comment at `download-jobs.js:1444` predicts a from-zero restart; that applies only when the credential changes mid-flight. ~11 GB was recoverable here — but only because I checked the partials on disk rather than trusting the tray.

### Steps to reproduce

1. `panel_request_secret` for a comfyui-scoped key with no downloads in flight → receipt says "N queued for the end of this turn".
2. In a later turn, start one or more long `download_model` transfers.
3. Let the queued respawn fire.
4. All transfers die silently; `action:"status"` reports "NOT running — the owning process is gone" while the tray still shows "downloading".

### Exact error (scrubbed)

No error at kill time. Only signal, on demand:

```
- `<id>` (tray `<tray_id>`) **downloading**  4.21/13.55 GB (31%)
    NOT running — the owning process is gone; started 448s ago
    NOTE: this transfer is NOT running on this machine. No process with its
    recorded pid exists here, so it died with its session (#1479).
```

Independently confirmed by sampling the partials: `delta1=0 bytes delta2=0 bytes in 20s`, with all 9 `~/.comfyui-mcp/cache/*.partial` files frozen at an identical mtime.

### Fix

Not patched locally: the queued-respawn execution site is in the orchestrator/agent-session layer (only the receipt *renderer* is reachable from the installed `dist/` — grepping for the scheduling site finds solely `panel-tools.js:333-341`), and the orchestrator only reloads code on a full process restart, so a local patch could not take effect this session. Precise change needed, in priority order:

1. **Re-take the at-risk snapshot at respawn-execution time, not save time.** For the `scheduled` path, call the #1378 in-flight enumerator (`download-jobs.js:1444`) again immediately before teardown. Keep the save-time snapshot for the `applied` path, where it is correct.
2. **With a non-empty list at that point, prefer deferring over killing** — hold the respawn until transfers drain (a *queued* respawn is by definition not urgent), or hand the transfers to a detached child that survives the swap.
3. **Auto-adopt orphaned jobs in the new session.** When a job record's owning pid is gone and its `.partial` exists, resume it rather than requiring cancel × N + re-issue × N. Dead-owner detection already exists (`download-status-proven-dead.js`); only recovery is missing.
4. **Never print "no reload, and no respawn required" on a receipt that also reports a queued respawn.** When `respawn.scheduled > 0`, say a respawn *is* pending and that long transfers started before it fires will be lost.
5. **Mark tray rows dead**, reflecting the same proven-dead state `action:"status"` already computes.

### Related

Same session and same trigger produced a second, distinct failure: the respawn permanently dropped all 92 `mcp__panel__*` tools while reporting a successful reconnect — filed as artokun/comfyui-mcp-panel#1228.

### Environment

OS: Linux (CachyOS, kernel 7.1.8-1) · ComfyUI 0.33.1 · ComfyUI **frontend 1.48.7** · GPU NVIDIA RTX 4070 SUPER (12 GB VRAM), 46 GB RAM · CUDA 13.0 / torch 2.13.0+cu130 · Python 3.13.12 · comfyui-mcp **0.51.45** · panel **0.14.28** · deploy env `local-desktop2-standalone`

Related (same root cause): #1378

artokun:
Fixes #1567.

A queued tool-session respawn still killed downloads started *after* the credential save, because the at-risk snapshot was the job list the orchestrator could see at save time — usually nothing. Downloads run in the MCP child; persist can lag a heartbeat; the child's live tray files are what actually exist when the respawn fires.

`downloadsAtRiskOfRespawn` now unions those live tray rows at the moment of teardown. Byte counts are read from the orchestrator's late-bound progress dir (it has no `COMFYUI_MCP_PROGRESS_DIR`), so the hold can wait on real progress instead of treating every transfer as already stalled.

The receipt warning and the stall-hold from #1576 / #1613 stay; this is the remaining ordering gap: snapshot at respawn, not only at save.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -759,6 +759,12 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+### MCP
+
+#### Fixed
+- a queued tool-session respawn sees the downloads started after the save, instead of killing them with no warning (#1567)
+- it waits on those transfers' live byte counts rather than treating every one as already stalled (#1567)
+
 ## [0.52.28] - 2026-08-20
 
 ### MCP
--- a/src/__tests__/services/download-progress.test.ts
+++ b/src/__tests__/services/download-progress.test.ts
@@ -26,6 +26,41 @@ function pendingFiles(): string[] {
   return readdirSync(dir).filter((f) => f.startsWith(mod.CONTROL_PREFIX));
 }
 
+describe("readDownloadProgress uses the late-bound dir (#1567)", () => {
+  // The orchestrator has no COMFYUI_MCP_PROGRESS_DIR; it late-binds via
+  // setProgressDir. A respawn-time snapshot runs in that process, so a reader
+  // that only looked at the import-time env capture returned null for every
+  // live transfer and the hold treated them as already stalled.
+  it("finds a row written into the dir the orchestrator actually uses", async () => {
+    vi.resetModules();
+    const lateDir = mkdtempSync(join(tmpdir(), "dl-late-"));
+    const saved = process.env.COMFYUI_MCP_PROGRESS_DIR;
+    delete process.env.COMFYUI_MCP_PROGRESS_DIR;
+    try {
+      const late = await import("../../services/download-progress.js");
+      late.setProgressDir(lateDir);
+      writeFileSync(
+        join(lateDir, "abc-deadbeef.json"),
+        JSON.stringify({
+          id: "abc",
+          name: "m.safetensors",
+          downloaded: 42,
+          total: 100,
+          bytes_per_sec: 1,
+          status: "downloading",
+          updated: Date.now(),
+        }),
+      );
+      expect(late.readDownloadProgress("abc")?.downloaded).toBe(42);
+      expect(late.listInFlightDownloadProgress().map((r) => r.id)).toEqual(["abc"]);
+    } finally {
+      if (saved === undefined) delete process.env.COMFYUI_MCP_PROGRESS_DIR;
+      else process.env.COMFYUI_MCP_PROGRESS_DIR = saved;
+      rmSync(lateDir, { recursive: true, force: true });
+    }
+  });
+});
+
 describe("download target stamping (#269)", () => {
   it("stamps the row with the writer's COMFYUI_URL at write time", () => {
     process.env.COMFYUI_URL = "https://podabc-3000.proxy.runpod.net";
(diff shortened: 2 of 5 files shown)


artokun:
Codex gate (read-only): PASS for merge. Reviewed the production diff and respawn callers. The at-risk snapshot now unions live downloading tray rows at respawn time, deduplicates by job identity, and reads the orchestrator late-bound progress directory so stall accounting sees real bytes; terminal/control rows are excluded. CI is green on macOS/Ubuntu/Windows and pack smoke. No P0/P1 or enhancement work is included.

PR #1892, opened 2026-08-20
fix(panel#1489): read-only graph calls are no longer refused while a render runs

artokun:
Reported as artokun/comfyui-mcp-panel#1489 (a `via-panel` report that was twice
triaged in the panel repo and routed here, but never actually filed or fixed here —
it recurred on 2026-08-20).

## The bug

The QUEUE BUSY preflight added for #1639 (PR #1745) fences every `graph_*` command
except `graph_run` whenever `QueueMonitor` sees a running prompt. It was applied to
**reads** as well:

```
Error: graph_outline was NOT sent — nothing was applied. QUEUE BUSY: a ComfyUI
prompt is running (…). The panel tab typically cannot answer graph_* commands
(including read-only graph_query / graph_outline) while a prompt is executing …
```

So `panel_graph_outline`, `panel_query_graph` and `panel_get_errors` are rejected
before dispatch during any render. An agent that only wants to **look** at the graph
has to poll `queue action:"list"` and come back later — and the recurrence report
notes this is exactly what blocks diagnosing a *stalled* render, which is the moment
`panel_get_errors` matters most.

## Why fencing reads bought nothing

The harm #1639 names is specific to a WRITE: a delivered frame cannot be retracted,
so a write that times out mid-render leaves the caller unable to say whether it
applied. A read has no such outcome — abandoning one costs a retry. That asymmetry is
already written down in this repo, in `BRIDGE_DEFAULT_TIMEOUT_MS`'s own doc:

> A read abandoned too early costs a retry. Nothing is ambiguous.
> A write abandoned too early is UNRECOVERABLE ambiguity.

The refusal also asserted more than we know. "The tab typically cannot answer" is a
prediction, not a measurement — and refusing the call made it unfalsifiable, because
no read ever got to try. The architectural reason to doubt it is that ComfyUI executes
in server-side Python, so the browser main thread is not the thing doing the work.
**How often a real tab answers mid-render is not measured here, and nothing in this
change depends on a number for it** — attempting the read is what turns the question
back into something observable.

Worth noting what #1639 actually asked for:

> Either (a) keep the panel's command channel responsive during execution (**the reads
> at minimum should never be blocked by a running prompt**), or (b) … say so explicitly
> instead of surfacing a generic 20/30s timeout.

It shipped (b) and applied it to reads too — the one combination the reporter did not
ask for. Reads now get **both** halves.

## The change

`graphCmdBlockedByRunningPrompt` now keys on `GRAPH_CMD_EFFECT`: `targeted` is refused
unsent, `inert` is dispatched.

- That is the ledger built for exactly this question — `targeted` is defined as
  "changes workflow content, publishes from it, or queues a render of it. A delivered
  frame cannot be retracted", which IS the unknown-outcome hazard being fenced.
- **Fails closed by construction.** An unclassified command has no entry, `undefined`
  is not `"inert"`, so it stays fenced. A new mutator is never waved past by omission.
- Deliberately **not** `BRIDGE_READONLY_CMDS`, which answers re-dispatch safety.
  #778 is the standing record of what conflating the two costs (`graph_canvas` is inert
  but not idempotent; `refresh_nodes` is idempotent but not a read). A test drives
  `graph_select_nodes` — inert, absent from `BRIDGE_READONLY_CMDS` — so re-deriving the
  split from the wrong list fails.

Two things deliberately kept:

- **Reads keep their full tolerant ack budget.** Capping it mid-render is tempting (a
  frozen tab now costs the caller 20 s instead of 0) but #357 and #589 were both
  regressions of precisely that trade, and #589 gave `panel_get_errors` a 30 s budget
  *because the panel's own bound is 18 s*. A cap here would re-break it.
- **The explicit diagnosis survives for reads.** A dispatched read that does time out is
  still annotated by `queueBusyTimeoutNote`, naming the running prompt — so the #1639
  benefit ("actionable, not a generic frozen-tab timeout") is preserved without the cost.

`queueBusyTimeoutNote` no longer flatly asserts *"this is not a backgrounded or frozen
tab"*. Reads reach that path now, and a tab that goes silent mid-render is genuinely
ambiguous; it now says the render is the most likely cause and how to tell the two
apart once the queue drains.

The mutation refusal also stops telling callers reads are unavailable, and instead
names the reads they can use right now.

## Verification

`src/__tests__/orchestrator/graph-read-during-render.test.ts` rewritten to pin the
split (it previously pinned `expect(sent).toEqual([])` for the two reads). 13 tests,
driving the real tool surface via `buildPanelToolDefs()`.

**Verified by mutation, not by a green suite:**

- Delete the one-line inert exemption → **6 tests fail** (`panel_graph_outline`,
  `panel_query_graph`, `panel_get_errors`, `panel_select_nodes` dispatch; both
  read-timeout-diagnosis tests). The 7 that stay green are the mutation-fence and
  `graph_run` tests, which that line does not touch.
- Restore the old over-claiming timeout note → **1 test fails**, confirming that
  assertion is load-bearing on its own and not just riding on the first mutation.

Full suite green (589/589 files pre-rebase; rebased onto current `main`, which picked
up #1886 touching the same `callOnce`); `tsc --noEmit` clean. One local full run after
the rebase flaked on `tool-surface-filter` / `vocabulary` with 30 s timeouts under
machine load — unrelated code paths, and both files pass 94/94 in isolation.

**Interaction with #1886 (merged into this same function while this was open):** a
mid-render read timeout does not enter the retry-safe reconnect branch that PR
widened — that branch requires `isTransientReconnectError` or a workflow-switch
refusal, and a reply timeout is neither. No double wait.

**Browser suite not run, and not applicable:** this diff is entirely orchestrator-side
in `src/orchestrator/panel-tools.ts`. There is no change in the panel pack and nothing
the panel renders is affected — the panel already treats `graph_outline` / `graph_query`
/ `graph_get_errors` as read-only (`READ_ONLY_GRAPH_COMMANDS` in
`web/js/lib/graph-binding.js`) and never saw these commands, because they were refused
before dispatch.

---

Gated by `codex-gate.mjs` (codex quota has recovered — the account was exhausted only
until 2026-08-19 20:43): **SHIP**, exit 0.

### DIFF
--- a/src/__tests__/orchestrator/graph-read-during-render.test.ts
+++ b/src/__tests__/orchestrator/graph-read-during-render.test.ts
@@ -1,16 +1,20 @@
-// #1639 — while a ComfyUI prompt is running, graph_* (including read-only
-// graph_query / graph_outline) time out with "tab may be backgrounded or frozen"
-// and mutating commands report an unknown outcome. 100% correlated with
-// queue action:"list" showing running:1; the same calls succeed the moment
-// the queue drains.
+// #1639 — while a ComfyUI prompt is running, a graph_* MUTATION can time out with
+// "tab may be backgrounded or frozen" and an unknown outcome. The orchestrator CAN
+// see the running prompt via QueueMonitor (HTTP /queue, independent of the panel
+// tab), so it fails a write closed BEFORE dispatch and the agent gets an explicit
+// QUEUE BUSY instead of a 20/30s unknown-outcome timeout. `graph_run` is excluded:
+// queuing behind an in-flight job is the documented sweep path.
 //
-// The frontend main thread is what cannot answer. The orchestrator CAN see the
-// running prompt via QueueMonitor (HTTP /queue, independent of the panel tab).
-// Fail closed BEFORE dispatch for canvas-touching graph commands so the agent
-// gets an explicit QUEUE BUSY instead of a 20/30s unknown-outcome timeout.
-// `graph_run` is excluded: queuing behind an in-flight job is the documented
-// sweep path. A timeout that still happens (graph_run, or a lagging snapshot)
-// is annotated with the same QUEUE BUSY rather than "backgrounded or frozen".
+// panel#1489 — that fence was applied to READS too, and this file used to pin
+// exactly that (`expect(sent).toEqual([])` for panel_query_graph and
+// panel_graph_outline). It should not have: a read that fails carries no unknown
+// outcome, so refusing it unsent bought no safety and cost the agent its only way
+// to LOOK at the graph mid-render.
+//
+// The property under test is therefore the SPLIT: `targeted` graph commands are
+// refused unsent, `inert` ones are dispatched, and a dispatched read that does
+// time out still comes back named QUEUE BUSY rather than "backgrounded or frozen".
+// The classification is read from GRAPH_CMD_EFFECT, not from a second list.
 
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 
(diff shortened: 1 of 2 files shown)


artokun:
Codex gate (read-only): PASS for merge. The production fence now derives mutation refusal from GRAPH_CMD_EFFECT: targeted graph writes stay fail-closed while inert/read commands are dispatched and retain timeout diagnosis. The real panel tool tests cover both dispatch and refusal paths, including graph_get_errors. MCP CI is green on all three OSes and pack smoke. No P0/P1 or enhancement work is included.

Issue #1887, opened 2026-08-20
free_vram: a correct free is reported as failure using another server GPU, and device-global vram_free cannot attribute the pin (from #1878)


artokun:
PR #1878 merged with a gate verdict of **NO-SHIP** that landed after the merge. Filing the findings so they are not lost with the PR. The underlying fix is worth keeping — this is about two states it does not distinguish.

## P0 — a correct free on the tab's own server is reported as a failure, naming another server's device

`src/orchestrator/panel-tools.ts:3053` — `annotateFreeVramAck(res: ToolResult)` takes **no `ctx`** and reads:

```ts
const base = (getComfyUIBaseUrl() || "").replace(/\/+$/, "");
```

That is the orchestrator's **global configured base**, not the server the acking tab provably fronts. Its sibling in the same diff, `settleFreeVramAfterAckTimeout` (`:2993`), gates on `captureRebootHealthBase(ctx)` — and the comment 60 lines above says why:

> a direct `/free` could aim at a DIFFERENT server than the one the tab was asked to free — reporting that as this command's success would be the wrong-target success the gate exists to prevent.

`getComfyUIBaseUrl()` is retargeted asynchronously from any tab's hello (`index.ts:4013 applyComfyuiUrl`, whose own comment cites a live "zombie :8189 tab").

**Executed**, not inferred — tab `tabServerOrigin = http://127.0.0.1:8189`, `setComfyuiTarget("http://127.0.0.1:8188")`, panel replies `{freed:true}`:

```
base actually queried : ["http://127.0.0.1:8188"]   <- the tab's server :8189 is never asked
isError               : true
"freed": false, "pinned_devices": [{"index": 0, …}] <- server A's device, named as if it were B's
```

The free on :8189 fully succeeded; the tool reports failure and prescribes `panel_restart_comfyui`.

## P1 — `deviceStillPinned` cannot tell "ComfyUI holds VRAM" from "another process does"

`:2862` keys on `vram_free`, which is `torch.cuda.mem_get_info` — **whole device, all processes**. Live `/system_stats` from this machine:

```
vram_total 25756696576  vram_free 795579336  -> 3.09% free  => PINNED
torch_vram_total 33554432 (32 MiB)           -> ComfyUI holds ~nothing
```

[...]

Issue #1888, opened 2026-08-20
panel_update_node: the reinstall note fires on the version argument, not on observed Manager evidence (from #1879)


artokun:
PR #1879 merged with a gate verdict of **NO-SHIP** that landed after the merge. Filing so the findings are not lost with the PR. The sanitizer itself is sound and its wiring is pinned; this is about claims it makes without observing them.

## P1 — the reinstall note is bolted onto errors that never reached Manager

`src/services/manager-update-error.ts:152` — `wantsNonGitReroute` fires on `version === "nightly"` alone. Nothing requires the message to be an update failure at all, so on any zip-installed pack **every** `isError` result gets the note.

Executed against the shipped handler:

```
in:  Error: no connected tab with id "dead". Connected: none
out: Error: no connected tab with id "dead". Connected: none

     NOTE: "fal-api" is installed as a Comfy Registry zip (no nested .git) …
     The pack was NOT git-updated. Reinstall it from the registry instead:
     install_custom_node (action:"reinstall", id:"fal-api") …
```

The same output appears for a **reply-timeout** result — whose own text says *"To retry this exact mutation, re-issue identical args plus `retry_of:…`"*, i.e. the mutation's outcome is explicitly unknown and may already be in flight (`panel-tools.ts:9213-9222`). So the note tells the agent to reinstall a pack when the real problem is a disconnected tab, or a dispatch whose result nobody knows yet.

The guard compares *"did the caller type nightly?"* when the question is *"did Manager report an update-git failure?"* — equal on the happy path, different on every transport failure. Gate it on observed Manager evidence (`UPDATE_GIT.test(message)`, or the generic update line) rather than on the caller's `version` argument.

## P1 — two claims asserted without observing either fact

- `"The previous ComfyUI generation traceback is execution history"` — `keepManagerUpdateEvidence` has only two outcomes for a traceback block: matches the Manager-frame whitelist, or *is* a previous-generation traceback. **Unrecognized provenance is not representable**, and the code asserts the generation reading rather than declining to classify.
- `"The pack was NOT git-updated"` — stated on the strength of the `version` argument, per the finding above.

## What held

[...]

PR #1893, opened 2026-08-20
release: v0.52.29 — free_vram reports the GPU it actually freed

artokun:
Cuts 0.52.29 from `main`.

Carries:

- #1890 / #1887 — a successful `panel_free_vram` is no longer reported as failure from another server GPU

Held out: #1880/#1869 ubuntu Build red; #1851/#1847 CONFLICTING; #1839 anthropic 4 CI red; #1697 P1 inflight.

Do **not** squash-merge. Merge-commit (keeps the `v0.52.29` tag on the version commit). Tag is local; push `v0.52.29` after merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -759,6 +759,17 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+## [0.52.29] - 2026-08-20
+
+### MCP
+
+#### Fixed
+- a successful free is no longer reported as failure from another server's GPU (#1890)
+- a post-restart graph READ waits out the reconnect instead of losing the race (#1886)
+- panel_create_group includes requested collapsed nodes (#1885)
+- restart_comfyui relaunches the venv python, not the trampoline's base child (#1884)
+
+
 ## [0.52.28] - 2026-08-20
 
 ### MCP
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "comfyui-mcp",
-  "version": "0.52.28",
+  "version": "0.52.29",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "comfyui-mcp",
-      "version": "0.52.28",
+      "version": "0.52.29",
       "license": "MIT",
       "dependencies": {
         "@comfyorg/sdk": "^0.1.7",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "comfyui-mcp",
-  "version": "0.52.28",
+  "version": "0.52.29",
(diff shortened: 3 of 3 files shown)


artokun:
**Picked this up and brought it current.** It was cut at 19:10Z and then sat while four more fixes merged into `main`, which turned it `CONFLICTING`. The worktree that cut it (`.worktrees/rel-05229`) has been clean and untouched since 19:10Z, so I took it rather than open a rival release PR.

**What changed**

`origin/main` merged in (`3a8583c`). The only conflict was `CHANGELOG.md` — `Unreleased` on main versus the `0.52.29` section here — and it is resolved by folding main's entries **into 0.52.29**, because their code is now in the tag and it would be dishonest to ship it under a heading that says otherwise. `Unreleased` is now empty.

0.52.29 therefore carries nine fixes, not four:

- a successful free is no longer reported as failure from another server's GPU (#1890 / **P1** #1887)
- a post-restart graph READ waits out the reconnect instead of losing the race (#1886)
- `panel_create_group` includes requested collapsed nodes (#1885)
- `restart_comfyui` relaunches the venv python, not the trampoline's base child (#1884)
- `panel_restart_comfyui` relaunches a Desktop instance when the parent PID exists but its command line cannot be read (#1847)
- a queued tool-session respawn sees the downloads started after the save (#1567)
- it waits on those transfers' live byte counts rather than treating every one as already stalled (#1567)
- `panel_update_node` reports an error that never reached Manager without recommending a reinstall (#1888)
- read-only graph calls are no longer refused while a render is running (panel#1489)

The last two shipped with **no changelog entry at all** — their commits do not touch `CHANGELOG.md` — so I wrote entries for them rather than let them ship invisibly.

**The tag moved, deliberately.** `v0.52.29` pointed at `dc4d3c4`, whose tree predates all four late fixes — publishing that tag would have shipped a package that does not match these release notes. It now points at `3a8583c`. It is still local and unpushed; it goes up after this merges, per the original instruction.

**Verification:** `tsc --noEmit` clean on the merged tree — and I checked the check, since a typecheck run where `tsc` does not resolve exits 0 with no output and looks identical to success. With a deliberate `TS2322` injected it exits 2 and names the file, so the clean run means something.

Still **merge-commit, not squash**, so the version commit survives in main's history.

**One note on #1847 / #1851**, which is in this cut: it merged carrying two P1 findings I had raised, and I had held it for exactly that reason. I re-checked both against merged `main` before letting it ship here, and **both were fixed before the merge** — `process-control.ts` now spawns only when the parent is proven gone *and* the port is free (the comment cites #1847 and the "a free port cannot tell gone-for-good from coming-back" reasoning directly), and the unconditional "that exact command is what brings it back" promise is gone from the confirmation card. My hold is withdrawn on the evidence, not waived.

PR #1894, opened 2026-08-20
fix(1888): panel_update_node no longer recommends reinstall on errors that never reached Manager

### LINKED ISSUE #1888: panel_update_node: the reinstall note fires on the version argument, not on observed Manager evidence (from #1879)

artokun:
Fixes #1888

## What was reported

`panel_update_node(id:"fal-api", version:"nightly")` on a zip-installed pack bolted a reinstall NOTE onto **every** `isError` result. A disconnected tab came back as:

```
Error: no connected tab with id "dead". Connected: none

NOTE: "fal-api" is installed as a Comfy Registry zip (no nested .git) …
The pack was NOT git-updated. Reinstall it from the registry instead:
install_custom_node (action:"reinstall", id:"fal-api") …
```

The same NOTE appeared on a reply-timeout whose own text says the mutation's outcome is unknown (`retry_of:…`). The sanitizer from #1879 is sound; the gate was not.

## Why

`wantsNonGitReroute` fired on `version === "nightly"` alone. That is the caller's argument, not Manager evidence, so it was equal to "Manager reported update-git" on the happy path and different on every transport failure.

## The fix

Gate the reinstall NOTE on observed Manager-update evidence (`res.action=update-git` or the generic `An error occurred while updating 'X'.` line), still only when the pack is observably a registry zip (on disk here, no nested `.git`). A disconnected tab, a reply-timeout, and any other error that never reached Manager stay identity. A zip pack whose Manager failure is `update-git` or the generic update line still re-routes, including when `version` is not nightly.

## Verification

`src/__tests__/orchestrator/update-node-stale-traceback.test.ts` drives the shipped `panel_update_node` handler:

- reporter: nightly + zip + disconnected tab is identity (no reinstall NOTE)
- nightly + zip + reply-timeout is identity (retry_of kept, no NOTE)
- `update-git` evidence still re-routes a zip pack without nightly
- the generic Manager update line still re-routes a zip pack without nightly or `update-git`
- existing #1870 cases (FalApiError drop, real `repo_update` traceback, success pass-through, remote does not invent a zip, git checkout is not re-routed)

`npx tsc --noEmit` clean.
`npx vitest run --maxWorkers=4 src/__tests__/orchestrator/update-node-stale-traceback.test.ts src/__tests__/services/panel-pin-guard.test.ts src/__tests__/services/manager-error-body.test.ts` — 136 passed.

### DIFF
--- a/src/__tests__/orchestrator/update-node-stale-traceback.test.ts
+++ b/src/__tests__/orchestrator/update-node-stale-traceback.test.ts
@@ -239,6 +239,79 @@ describe("#1870 panel_update_node does not attach a stale generation traceback",
   });
 });
 
+describe("#1888 reinstall note requires observed Manager evidence, not version", () => {
+  // The sanitizer is sound; the defect is the gate. wantsNonGitReroute fired
+  // on version==="nightly" alone, so a zip-installed pack got a reinstall NOTE
+  // on every isError — including failures that never reached Manager.
+  const DISCONNECTED = 'Error: no connected tab with id "dead". Connected: none';
+  const REPLY_TIMEOUT =
+    'Error: Panel tab tab-1 did not reply to "graph_update_node" within 30000 ms — the ComfyUI tab may be backgrounded or frozen\n\n' +
+    'To retry this exact mutation, re-issue identical args plus retry_of:"11111111-2222-3333-4444-555555555555"; otherwise call normally.';
+
+  it("THE REPORT: a disconnected-tab error with version nightly does not get the reinstall note", async () => {
+    disk.root = zipPackRoot();
+    const { ctx } = makeCtx({
+      content: [{ type: "text", text: DISCONNECTED }],
+      isError: true,
+    });
+    const res = await defByName("panel_update_node").handler(
+      { id: PACK, version: "nightly" },
+      ctx,
+    );
+    expect(res.isError).toBe(true);
+    const text = textOf(res);
+    expect(text).toBe(DISCONNECTED);
+    expect(text).not.toMatch(/Comfy Registry zip/);
+    expect(text).not.toMatch(/The pack was NOT git-updated/);
+    expect(text).not.toMatch(/install_custom_node \(action:"reinstall"/);
+  });
+
+  it("a reply-timeout (outcome unknown) with version nightly does not get the reinstall note", async () => {
+    disk.root = zipPackRoot();
+    const { ctx } = makeCtx({
+      content: [{ type: "text", text: REPLY_TIMEOUT }],
+      isError: true,
+    });
+    const res = await defByName("panel_update_node").handler(
+      { id: PACK, version: "nightly" },
+      ctx,
+    );
+    const text = textOf(res);
+    expect(text).toBe(REPLY_TIMEOUT);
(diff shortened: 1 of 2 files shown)


artokun:
Codex gate (read-only): PASS for merge. Reviewed the production diff and shipped-handler tests. The reinstall note is now gated by observed Manager update evidence (generic update line or res.action=update-git) plus an observed non-git pack; disconnected-tab and reply-timeout errors remain identity. CI is green on macOS/Ubuntu/Windows and pack smoke. No P0/P1 or enhancement work is included.

PR #1897, opened 2026-08-20
fix(1889): a rejected node id says what it wanted and what it got

### LINKED ISSUE #1889: nodeId() union yields a bare "Invalid input at node_id" — no expected/received, unlike every sibling field
### What happens

Calling a panel tool whose schema uses `nodeId()` with a missing/invalid `node_id` produces a validation error where **one field explains itself and the other does not**:

```
MCP error -32602: Input validation error: Invalid arguments for tool panel_set_node_title: Invalid input at node_id
Invalid input: expected string, received undefined at title
```

`title` says what it wanted and what it got. `node_id` says only `Invalid input`. Read left-to-right the two lines look like one sentence about `node_id` that then blames `title`, which is what confused the reporter.

### Root cause

`nodeId()` (`src/orchestrator/panel-tools.ts:8976`) is a bare `z.union` with no union-level error:

```ts
const nodeId = () =>
  z
    .union([
      z.number().int(),
      z.string().regex(NODE_ID_PATTERN, NODE_ID_MESSAGE),
    ])
    .transform(normalizeNodeId);
```

Zod 4 renders a failed union as a bare `"Invalid input"` — the per-member messages (including `NODE_ID_MESSAGE`) are attached to the union's sub-issues and are dropped by the flattened `-32602` text. `z.string()` renders the detailed `expected string, received undefined` form. Hence the asymmetry.

### Reproduction (verified against this repo's installed zod 4.4.3)

```js
import { z } from "zod";
const NODE_ID_PATTERN = /^\d+(:\d+)*$/;
const nodeId = () => z.union([z.number().int(), z.string().regex(NODE_ID_PATTERN, "bad")]);
const schema = z.object({ node_id: nodeId(), title: z.string() });
console.log(z.prettifyError(schema.safeParse({}).error));
```

Output — the reporter's string, word for word:

```
✖ Invalid input
  → at node_id
✖ Invalid input: expected string, received undefined
  → at title
```

Adding a union-level message fixes it with no schema/behaviour change:

```js
z.union([...], { error: 'expected a node id: an integer or a "12:7" subgraph path string' })
```

```
✖ expected a node id: an integer or a "12:7" subgraph path string
  → at node_id
✖ Invalid input: expected string, received undefined
  → at title
```

### Scope

`nodeId()` is used by many `panel_*` write tools, so the wording fix lands once and improves all of them. `panel_set_node_title` is just where it was observed.

### Not a defect in the rejection itself

The originating call was **intentionally** incomplete (`args` omitted), so rejecting it is correct and both fields genuinely were missing. This is about the message an agent reads when it gets a `node_id` wrong — only the wording is at fault.

### Provenance

Reported from the panel as https://github.com/artokun/comfyui-mcp-panel/issues/1490 (`panel_set_node_title rejects valid call with misleading missing-argument error`) by `chatgpt` / `gpt-5.6-luna` on comfyui-mcp 0.52.23, panel 0.15.23. That issue was closed as upstream: the panel pack registers no MCP tools and does not own this validation — the schema and the error text are both here.

artokun:
Fixes #1889.

## What was wrong

A panel write tool called with a missing or malformed `node_id` answered with the single word `Invalid input`, while the field beside it explained itself in full:

```
MCP error -32602: Input validation error: Invalid arguments for tool panel_set_node_title: Invalid input at node_id
Invalid input: expected string, received undefined at title
```

Read left-to-right the two lines look like one sentence about `node_id` that then blames `title`, which is what confused the reporter.

## Mechanism, and the part the report did not have

`nodeId()` is a `z.union`. Zod 4 reports a failed union as **one** `invalid_union` issue whose own `message` is the constant `"Invalid input"`; the per-member messages — `NODE_ID_MESSAGE` among them — sit in a nested `errors` array. Every renderer that reaches an agent flattens `issue.message` and never descends into that nest.

I pinned down which renderer produced the reporter's exact string, because the answer was not obvious. `@modelcontextprotocol/sdk` **1.29** (the version in the stale local checkout) JSON-stringifies the issues, which is not what they quoted. **1.30**, the version on `main`, does this:

```js
// node_modules/@modelcontextprotocol/sdk/dist/esm/server/zod-compat.js
return error.issues.map((i) => i.path?.length ? `${i.message} at ${getDotPath(i.path)}` : i.message).join('\n')
```

That is the reporter's string verbatim, so the reproduction below runs through a real `McpServer` + `Client` rather than asserting on a `safeParse` result — a schema-level assertion would be checking a different sentence than the one that reached them.

**A correction to the report's framing:** `NODE_ID_MESSAGE` is not unreachable. A *bad string* (`"42px"`) fails only the string member, zod collapses that to a plain `invalid_format` issue, and the message printed correctly all along — I measured this, and there is a test asserting it so the fix is not credited with more than it did. The message was dropped only for inputs that fail **every** member: `undefined`, `null`, `4.5`, an object. That is the common case and the reported one.

## The fix

A union-level `error` on both node-id unions. It carries `received` as well as `expected`, because the complaint was the asymmetry with the sibling field, and half of what the sibling said was what it actually got:

```
a node id must be an integer (e.g. 42) or a subgraph-qualified id (e.g. 120:104); received undefined at node_id
Invalid input: expected string, received undefined at title
```

The existing `NODE_ID_MESSAGE` is reused rather than a new string invented, so the two rejection paths (regex failure and union failure) now say the same thing.

**`runToNodeId()` is fixed too.** It is the identical shape with the identical hole, and #1497 wrote `RUN_TO_NODE_ID_MESSAGE` specifically so a subgraph-qualified execution root would be "refused now by name, with a reason" — which was true for a qualified string and false for every shape that failed both members.

`describeReceived()` is deliberately total. It runs **inside zod's error path**, where a formatter that throws converts a clean validation refusal into a crash. `JSON.stringify` alone fails four ways there: it throws on a BigInt and on a circular object, returns the bare `undefined` value for a symbol or a function, and renders `NaN`/`Infinity` as `null` — which would tell a caller who passed NaN that they passed null. Output is capped at 60 chars.

**Second commit: that totality claim was false when I first wrote it.** I probed the function against ten hostile inputs instead of re-reading it, and the tenth got through — in the least visible place, *inside the catch*. `JSON.stringify` throws on a revoked proxy, so control reached the recovery path, and the recovery path called `Array.isArray` to pick between `"an array"` and `"an object"`. `Array.isArray` is itself a proxy operation and throws on a revoked proxy, so the fallback was another way out of a function that must not have one. Nine of the ten were already handled; enumerating them is the only reason the tenth surfaced. No revoked proxy can arrive over the MCP wire — input there is parsed JSON — but the contract is "cannot throw", and one that holds only for the inputs I happened to imagine is not worth stating in a comment.

## Evidence

**Behaviour is unchanged — asserted, not assumed.** A union-level `error` that accidentally narrowed the union would be a far worse bug than this one and would not show up in any message assertion, so acceptance is tested directly: `42`, `"42"`, `"120:104"`, `"120:113:78"`, `-20`, `"-20"`, `"-20:3"` all still validate; `"4.5"`, `""`, `"abc"`, `"1:"`, `":1"`, `"1::2"`, `"1:-2"` are all still refused. The advertised MCP input schema is also asserted byte-for-byte — it still carries the regex `pattern` that `node-id.ts` chose `.regex()` over `.refine()` to preserve.

**Coverage is swept, not spot-checked.** A green test on `panel_set_node_title` proves the helper works; it says nothing about whether the other node-id arguments route through it. The suite enumerates every registered tool and fails if any node-id field still renders the bare union message — **21 arguments across 20 tools**.

**Verified by deleting the fix.** Four mutations, each killing named tests:

| mutation | tests killed |
|---|---|
| remove `error:` from `nodeId()` | **23** (3 message + 20 swept tools) |
| remove `error:` from `runToNodeId()` | **2** (`panel_run` message + `panel_run.to_node_id` sweep) |
| `unionErrorFor` drops the `received` half | **24** |
| `describeReceived` → raw `JSON.stringify` | **11** (bigint and circular *throw*) |
| remove the nested revoked-proxy guard | **2** |

56 tests in the new file, all green unmutated. `tsc --noEmit` clean. `check:vocabulary` OK (1727 files). `check:unknown-collapse` OK.

**Not a regression.** `git log -L` over the `nodeId` definition returns exactly two commits — #845 created it, #1425 widened it. No union-level message ever existed and was removed.

- **No browser gate.** This is orchestrator-side zod wording; no panel file is touched, and panel#1490 was closed as upstream on the grounds that the panel registers no MCP tools and does not own this validation. Playwright would not exercise the change, so I did not run it and am not implying coverage I do not have.
- **`slotRef` and the other bare unions in this file are the same defect class and are left alone.** `slotRef = z.union([z.string(), z.number().int().min(0)])` renders the same bare `Invalid input`. Fixing it means inventing new caller-facing vocabulary for a surface nobody has reported, which is a wider change than a P3 wording bug under the freeze. Named here so it is a decision, not an oversight.
- **The regex-failure path still prints no `received`.** `"42px"` gets `NODE_ID_MESSAGE` with no value echo, matching zod's own `invalid_format` idiom. The value is in the caller's own call and the shape is named; adding it there was scope I did not need.

## Review

**Review is pending — this has not been reviewed by anyone but me, and I am not treating my own reading as a review.**

@grok — the load-bearing claims, in the order I'd attack them:

1. **The one I am least sure of: does a union-level `error` change what the union ACCEPTS?** My whole "wording only" claim rests on no. I asserted the accept/reject sets directly and `z.toJSONSchema` is unchanged, but if zod 4.4.3 has any path where a supplied `error` alters member evaluation or short-circuiting, this is where it bites and my tests would not catch a subtle case (e.g. a value where the union previously picked a *different* member's result before the transform).
2. **`describeReceived` totality — this one already failed once, so treat my confidence as low.** It sits inside an error path, so a throw is a crash, not a bad message. I now enumerate bigint / symbol / function / circular object / circular array / NaN / ±Infinity / throwing `get` trap / throwing `ownKeys` trap / revoked proxy / throwing `toJSON` / throwing getter / null-prototype object / `Map` / throwing `Symbol.toPrimitive`. The revoked proxy is the one that got through my first version. **What is the eleventh?** I am specifically unsure whether `String(input)` in the `?? String(input)` branch can throw — it is reachable only when `JSON.stringify` returns `undefined`, and I did not construct that case.
3. **Is the sweep actually load-bearing, or does it pass vacuously?** It keys on top-level schema keys named `node_id`/`to_node_id`. A node-id argument under any other name, or nested inside an object/array field, is invisible to it — `panel_select_nodes.node_ids` (an array) is one I know it does not cover. Mutation 1 killed 20 of its cases, so it is not vacuous, but its *scope* may be narrower than the PR implies.
4. **The renderer claim.** I assert the SDK 1.30 path produces the reporter's string. If the orchestrator ever serves these tools through a different transport whose renderer descends into `issue.errors`, the union-level message would be a second, redundant sentence rather than the only one.

### DIFF
--- a/src/__tests__/orchestrator/node-id-union-message.test.ts
+++ b/src/__tests__/orchestrator/node-id-union-message.test.ts
@@ -234,6 +234,45 @@ describe("#1889 describeReceived is TOTAL — it runs inside zod's error path", 
     arr.push(arr);
     expect(describeReceived(arr)).toBe("an array");
   });
+
+  // The list below is the probe that found the last hole rather than a
+  // decoration on a fix already believed complete. The REVOKED PROXY case threw:
+  // `Array.isArray` is itself a proxy trap, so the recovery path inside the
+  // catch was another way out of this function. Nine of ten hostile inputs were
+  // already handled; enumerating them is what surfaced the tenth.
+  //
+  // None of these can arrive over the MCP wire, where input is parsed JSON. They
+  // are here because the CONTRACT is "cannot throw", and a contract that holds
+  // only for the inputs one happened to imagine is not one worth stating.
+  const hostile: Array<[string, () => unknown]> = [
+    ["a proxy whose get trap throws", () => new Proxy({}, { get: () => { throw new Error("boom"); } })],
+    ["a proxy whose ownKeys trap throws", () => new Proxy({}, { ownKeys: () => { throw new Error("boom"); } })],
+    ["a REVOKED proxy", () => { const p = Proxy.revocable({}, {}); p.revoke(); return p.proxy; }],
+    ["an object whose toJSON throws", () => ({ toJSON: () => { throw new Error("boom"); } })],
+    ["an object with a throwing getter", () =>
+      Object.defineProperty({}, "x", { get: () => { throw new Error("boom"); }, enumerable: true })],
+    ["a null-prototype object", () => Object.create(null)],
+    ["a Map", () => new Map([[1, 2]])],
+    ["an object whose Symbol.toPrimitive throws", () =>
+      ({ [Symbol.toPrimitive]: () => { throw new Error("boom"); } })],
+  ];
+
+  it.each(hostile)("survives %s", (_label, make) => {
+    let out: string | undefined;
+    expect(() => {
+      out = describeReceived(make());
+    }).not.toThrow();
+    expect(typeof out).toBe("string");
+  });
+
+  it("a revoked proxy specifically — the one that got past the outer catch", () => {
+    const { proxy, revoke } = Proxy.revocable({}, {});
+    revoke();
+    // Both the stringify AND the naive `Array.isArray` fallback throw on this.
+    expect(() => JSON.stringify(proxy)).toThrow();
+    expect(() => Array.isArray(proxy)).toThrow(); // the trap, stated outright
+    expect(describeReceived(proxy)).toBe("an object");
+  });
 });
 
 describe("#1889 unionErrorFor composes the two halves", () => {
--- a/src/__tests__/orchestrator/node-id-union-message.test.ts
+++ b/src/__tests__/orchestrator/node-id-union-message.test.ts
@@ -1,0 +1,245 @@
+// #1889 — a node-id rejection said only `Invalid input`, while the field beside it
+// said `expected string, received undefined`. Read left-to-right the two lines look
+// like ONE sentence about node_id that then blames title, which is what confused the
+// reporter:
+//
+//     Invalid input at node_id
+//     Invalid input: expected string, received undefined at title
+//
+// Cause: zod 4 reports a failed `z.union` as one `invalid_union` issue whose own
+// `message` is the constant `"Invalid input"`, with the per-member messages —
+// NODE_ID_MESSAGE among them — in a nested `errors` array. The MCP SDK's
+// `getParseErrorMessage` (1.30) renders `${issue.message} at ${dotPath}` and never
+// descends into the nest, so the useful half was dropped on the way out.
+//
+// The first suite deliberately goes through a REAL McpServer + Client rather than
+// calling `z.object(def.schema).safeParse` directly: the string the reporter quoted
+// is produced by the SDK's renderer, not by zod, so a schema-level assertion would
+// be testing a different sentence than the one that reached them.
+
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { z } from "zod";
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
+
+import { buildPanelToolDefs, registerPanelTools, type PanelToolCtx } from "../../orchestrator/panel-tools.js";
+import { NODE_ID_MESSAGE, describeReceived, unionErrorFor } from "../../orchestrator/node-id.js";
+
+function makeFakeCtx(): PanelToolCtx {
+  return {
+    call: async (cmd) => ({ content: [{ type: "text", text: JSON.stringify(cmd) }] }),
+    confirm: async () => "yes" as const,
+    bridge: { send: async () => ({}) } as unknown as PanelToolCtx["bridge"],
+    tabId: "test-tab",
+  } as PanelToolCtx;
+}
+
+describe("#1889 the message an agent actually receives over MCP", () => {
+  let client: Client;
+  let server: McpServer;
+
+  beforeAll(async () => {
+    server = new McpServer({ name: "node-id-union-message-test", version: "1.0.0" });
+    registerPanelTools(server, makeFakeCtx());
+    const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
+    client = new Client({ name: "node-id-union-message-test-client", version: "1.0.0" });
+    await Promise.all([server.connect(serverTransport), client.connect(clientTransport)]);
+  });
+
+  afterAll(async () => {
+    await client.close();
+    await server.close();
+  });
+
+  const errorText = async (name: string, args: Record<string, unknown>): Promise<string> => {
+    const r = await client.callTool({ name, arguments: args });
+    expect(r.isError, `${name} was expected to reject ${JSON.stringify(args)}`).toBe(true);
+    return (r.content as Array<{ text?: string }>)?.[0]?.text ?? "";
+  };
+
+  it("REPRODUCTION: the reporter's exact call now explains node_id instead of saying `Invalid input`", async () => {
+    const text = await errorText("panel_set_node_title", {});
+
+    // What they got.
+    expect(text).not.toMatch(/^Invalid input at node_id$/m);
+    // What they get now — both halves the sibling field had.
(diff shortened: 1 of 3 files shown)


artokun:
Codex blocking review finding: describeReceived() is not total for all values it claims to format. If JSON.stringify(input) throws for a revoked Proxy or a Proxy trap, the catch path calls Array.isArray(input) and then potentially String(input); either can throw again. Because this formatter runs inside Zod's error callback, that turns a validation refusal into a handler error. Please harden the catch path so it cannot invoke throwing Proxy traps (and add a revoked/hostile Proxy regression), then rerun the mutation/targeted tests and request review again. Do not merge this revision yet.

artokun:
@grok — review request. This is a P3 wording fix, but it edits a validator that ~21 tool arguments route through, so the risk is not in the message, it is in whether the schema still means the same thing.

Please attack these four in order. I have named where I am weakest rather than handing you a diff to hunt through.

**1. Does a union-level `error` change what the union ACCEPTS?** (`panel-tools.ts` `nodeId()` / `runToNodeId()`)
My entire "wording only" claim rests on *no*. I assert the accept and reject sets directly and `z.toJSONSchema` is byte-identical, but a subtle case my tests would miss: a value where the union previously resolved to a *different member* before `.transform()` ran. `normalizeNodeId` returns `number | string` depending on which member matched, and picking the other member silently retargets a node — that is the exact class of bug #1425 was fixed for. If zod 4.4.3 has any path where a supplied `error` map perturbs member evaluation order or short-circuiting, this is where it bites.

**2. `describeReceived` totality — my confidence here should be discounted, because it already failed once.**
It runs inside zod's error path, so a throw is a crash rather than a bad message. My first version claimed totality and was wrong: `JSON.stringify` throws on a revoked proxy, control reached the catch, and the catch called `Array.isArray` — itself a proxy operation, which also throws on a revoked proxy. Fixed in the second commit with a nested guard. I found that by enumerating ten hostile inputs, not by re-reading the code, so please assume there is an eleventh. The specific one I did **not** construct: `JSON.stringify(input) ?? String(input)` — can `String(input)` throw on the branch where `JSON.stringify` returns `undefined`?

**3. Is the coverage sweep narrower than the PR implies?**
It keys on top-level schema keys literally named `node_id` / `to_node_id`. A node-id argument under any other name, or nested inside an object or array field, is invisible to it — I know `panel_select_nodes.node_ids` (an array of node ids) is not covered, and there may be others. Mutation 1 killed 20 of its cases so it is not vacuous, but "21 arguments across 20 tools" may be undercounting the real surface, which would mean bare `Invalid input` still reaches callers somewhere I did not look.

**4. The renderer claim.**
I assert the reporter's exact string comes from `@modelcontextprotocol/sdk` 1.30's `getParseErrorMessage`, which joins `${i.message} at ${dotPath}` and never descends into `issue.errors`. If some transport in this repo serves these tools through a renderer that *does* descend into the nest, the union-level message becomes a second redundant sentence rather than the only one, and the fix is cosmetically wrong there even though it is right on the reported path.

Not asking you to re-derive the mechanism — the reproduction is in the PR body and runs through a real `McpServer` + `Client`. Asking you to break 1 and 2.

artokun:
Codex gate (read-only): PASS for merge after the blocking review finding was fixed. describeReceived now guards its recovery path against revoked/hostile Proxy traps and has explicit regressions; union acceptance/schema behavior remains covered. CI is green on macOS/Ubuntu/Windows and pack smoke. No P0/P1 or enhancement work is included.

PR #1898, opened 2026-08-20
fix(1895): an occupied GPU after free is named instead of a silent freed:true

### LINKED ISSUE #1895: free_vram after #1887: a card that is still 97% occupied now reports a bare freed:true with no mention of the device
Follow-up to #1887 / PR #1890 (merged as f5a529ab). The wrong-target read is genuinely fixed — this is about what replaced the misattributed warning.

All four observations below were EXECUTED against f5a529ab in a worktree, not read off the diff.

## 1. The occupied device is no longer mentioned at all

`deviceStillPinned` now keys on `torch_vram_total`/`torch_vram_free`, so an occupied card whose torch pool is empty is (correctly) not a `/free` failure. But the fall-through is `return res` — the panel's ack passes straight through, so the caller gets:

```
{ "freed": true, "unload_models": true, "free_memory": true }
```

with no indication that the device is still full. Driven with THIS machine's real `/system_stats` shape (RTX 4090, `cudaMallocAsync`, ComfyUI 0.33.2) in its occupied state — `vram_total 25756696576`, `vram_free 795579336` (3.09% of the device free), `torch_vram_total 33554432` (32 MiB pool) — the reply never contains `795579336` and never names device 0.

That is the exact question #1866 was filed about ("must not report full VRAM freed when a device remains pinned"), answered now by silence instead of by a wrong attribution. `panel_free_vram` is called to unwedge an OOM; `freed: true` with an untouched 97%-occupied card sends the caller straight back into the OOM with no clue why.

Suggested: keep `isError: false` (the free DID succeed), but disclose — name the device, the counters, and the fact that `/system_stats` reports the device globally and cannot say which process holds it. The honest wording has to carry both shapes and assert neither: (a) allocations this ComfyUI made outside torch's allocator (Ray workers, sequence-parallel / parallel CLIP loaders in their own processes), which `panel_restart_comfyui` does clear, and (b) a different process on the host (a second ComfyUI instance, a trainer, the browser), which restarting this one will not.

## 2. "Could not prove the tab's server" is also silent

`annotateFreeVramAck` correctly refuses to read `/system_stats` when `captureRebootHealthBase(ctx)` is null — but then `return res`, so a remote/cloud tab (or any tab whose handshake Origin does not match the boot instance) gets a `freed: true` that is byte-identical to a verified one. Executed: the stats reader is called zero times, and the reply carries nothing saying the GPU was never re-read. A one-sentence disclosure would make `freed:true` readable as "the panel's /free receipt" rather than "a measured device".

## 3. The pin ratio measures pool FULLNESS, not the card share the pool holds

`torch_vram_free / torch_vram_total <= 0.2` asks "how full is ComfyUI's allocator pool", but reserved bytes are held from the driver whether or not they are used inside the pool. Executed against f5a529ab with a 24 GiB card at 3% device-free where this ComfyUI still reserves 5 GiB (`torch_vram_total 5368709120`) of which 4 GiB is unused inside the pool (`torch_vram_free 4294967296`):

```
isError = false
{ "freed": true, "unload_models": true, "free_memory": true }
```

so a 5 GiB reservation this ComfyUI genuinely holds reads as "not pinned". Comparing `torch_vram_total` against `vram_total` would ask the question the verdict actually needs. I have NOT observed a live ComfyUI in that state — after `/free` with `free_memory:true`, `soft_empty_cache(force=True)` normally collapses reserved toward active — so treat this as a predicate-shape observation, not a reproduced field bug. It is worth noting that this rig runs the `cudaMallocAsync` allocator, where `empty_cache()` trims a CUDA memory pool and the reserved/active accounting is not the native caching allocator's.

`PINNED_VRAM_MIN_TOTAL_BYTES` (1 GiB) also now gates on the POOL rather than the device, so a fully-active 900 MiB pool is never pinned. Same family, smaller.

## 4. What is definitely fine

The base selection is properly pinned by tests now — mutating `annotateFreeVramAck`'s base to `"http://wrong.example:1"` kills 8 tests, and reverting it to `getComfyUIBaseUrl()` kills 2 (including one named for the retarget case). Reverting `deviceStillPinned` to the device-global counters kills 3. The gap flagged when #1887 was filed (all 12 tests green under a wrong-base mutation) is closed.

## Environment
Windows 11, ComfyUI 0.33.2, RTX 4090 (25756696576 bytes), PyTorch 2.10.0+cu130, comfyui-mcp 0.52.2. Observations executed against `f5a529ab` (origin/main).

Related (same root cause): #1866, #1887

artokun:
Fixes #1895

Follow-up to #1887 / #1890. The wrong-target read is still gated on `captureRebootHealthBase`. This is about the silent `freed:true` that replaced the misattributed warning.

## Occupied card, empty torch pool

Live `/system_stats` from the filing machine: `vram_free 795579336` of `25756696576` (3.09% free) with a 32 MiB torch pool. That is (correctly) not a `/free` failure — `isError` stays false, `freed` stays true — but the reply now names the device, prints the counters, and says `/system_stats` is device-global. The wording asserts neither holder: (a) this ComfyUI outside torch (Ray/CLIP loaders — `panel_restart_comfyui` of THIS instance can clear those); (b) another process (a second ComfyUI — restarting THIS instance will not).

## Unproven tab

When `captureRebootHealthBase(ctx)` is null, occupancy is still not read. The ack is no longer byte-identical to a measured GPU: `occupancy_reread: false` and a one-sentence note that `freed:true` is the panel's `/free` receipt.

## Pin ratio

Left as pool fullness (`torch_vram_free / torch_vram_total`). Reserved share of the card was a predicate-shape note, not a live observation after `/free`.

## Tests

Drive `annotateFreeVramAck` and `deviceStillPinned` via `__panelToolsTestHooks`. The #1887 URL-pin tests still fail if occupancy is swapped to `getComfyUIBaseUrl()`.

### DIFF
--- a/src/__tests__/orchestrator/free-vram-pinned-device.test.ts
+++ b/src/__tests__/orchestrator/free-vram-pinned-device.test.ts
@@ -18,6 +18,10 @@
 // These tests pass `base` into the injected reader and assert it. Discarding
 // the argument is what made a `const base = "http://wrong.example:1"` mutation
 // of annotateFreeVramAck leave every test green.
+//
+// #1895 — an occupied card whose torch pool is empty must still NAME the
+// device (isError stays false). An unproven tab must not look like a
+// measured GPU. Tests drive annotateFreeVramAck / deviceStillPinned directly.
 
 import { beforeEach, afterEach, describe, expect, it, vi } from "vitest";
 
@@ -123,6 +127,14 @@ function expectQueriedProven(ctx: PanelToolCtx): void {
   const proven = __panelToolsTestHooks.captureRebootHealthBase(ctx);
   expect(proven).toBeTruthy();
   expect(queriedBases).toEqual([proven]);
+}
+
+const FREED_ACK_BODY = { freed: true, unload_models: true, free_memory: true };
+
+function ackFreed(): ToolResult {
+  return {
+    content: [{ type: "text", text: JSON.stringify(FREED_ACK_BODY, null, 2) }],
+  };
 }
 
 /** Reporter: 4× RTX 2080 Ti, 21 GB. GPU 2 at vram_free=187552724 and
@@ -264,6 +276,9 @@ describe("panel_free_vram reads occupancy from this tab's server, not the global
     // Tab fronts a DIFFERENT local instance than boot. Global is the boot
     // server. Returning pinned devices if that host is queried makes a
     // wrong-target read fail this test — the #1878 tests were blind to it.
+    // #1895 — the ack is no longer byte-identical to a measured GPU: it must
+    // say occupancy was never re-read. Swapping the gate to getComfyUIBaseUrl()
+    // still fails: queriedBases would be non-empty and GPU 2 would be named.
     setDevices(reporterDevices);
     expect(setComfyuiTarget(BOOT_BASE)).toBe(true);
     const out = await run("ok", { origin: OTHER_BASE });
@@ -276,6 +291,8 @@ describe("panel_free_vram reads occupancy from this tab's server, not the global
     expect(out.text).not.toMatch(/STILL PINNED/);
     expect(out.text).not.toMatch(/panel_restart_comfyui/);
     expect(out.text).not.toMatch(/"index": 2/);
+    expect(out.text).toMatch(/"occupancy_reread": false/);
+    expect(out.text).toMatch(/never re-read/);
   });
 });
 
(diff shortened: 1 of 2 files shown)


artokun:
Codex gate (read-only): PASS for merge. Reviewed the production disclosure path and tests. Measured device-global occupancy is reported without converting a successful /free into a false failure; an unproven tab is explicitly marked occupancy_reread:false; torch-pool pinning remains the separate failure path. CI is green on macOS/Ubuntu/Windows and pack smoke. No P0/P1 or enhancement work is included.

PR #1899, opened 2026-08-20
fix(panel#1515): the ENV line reports the panel ComfyUI will run, and the refusal names the cause that applied

artokun:
Fixes the two changes requested in artokun/comfyui-mcp-panel#1515. Both live **here**, not in the panel repo — the write-fence text is built in `src/services/panel-recovery.ts` and the ENVIRONMENT line's panel version in `src/services/env-capabilities.ts`. (Zero hits for the reported error text anywhere in the panel pack's own source.)

**Review is PENDING — not gated, not reviewed.** See "What I want a reviewer to attack" below.

## The reporter saw three answers; there is one cause

| surface | said |
|---|---|
| agent ENVIRONMENT line | `panel 0.7.3` |
| `pyproject.toml` / `PANEL_VERSION` | `0.15.24` |
| write fence | tab advertised **no version** |

`readInstalledPanelVersion` scanned two known `custom_nodes` dir names and returned the **first readable one**:

```ts
const PANEL_DIR_NAMES = ["comfyui-mcp-panel", "comfyui-agent-panel"] as const;
for (const name of PANEL_DIR_NAMES) { ... if (name matches && version) return parsed.version; }
```

Both dirs carry the same authoritative `[project].name`, so the name screen passes for **both** and a leftover git clone at `custom_nodes/comfyui-mcp-panel` outranks the live Registry install at `custom_nodes/comfyui-agent-panel` purely because it is listed first. That is the documented two-panels state (#1269 / #641) — `panel-recovery.ts`'s own `PANEL_DIR_NAMES` comment exists to stop this module's advice from manufacturing it.

Verified against the panel repo rather than assumed:

- **0.7.3 is a real released panel version** (`43b46698`), and its `pyproject.toml` declares `name = "comfyui-agent-panel"` — so it passes the screen and is returned as "the installed panel".
- `panel_version` entered the hello at panel **0.11.3** (`27878273`/`d7930f41`, whose pyproject reads `0.11.3`). A 0.7.3 bundle predates it and therefore hellos with **no version** — exactly what the fence reported.

So the same stale copy explains the ENV line *and* the missing advertisement. One cause, three symptoms.

> Correcting my own first pass: I initially read this as 0.11.83 from `git tag --contains`. That was wrong — v0.11.83 is the panel repo's **earliest tag**, so `--contains` reports it for any older commit. Reading the pyproject at the commit gives 0.11.3.

## Change 1 — newest wins, not first-listed

The scan keeps going after a hit and reports the **newest** copy. The tie-break is the panel's own rule, not a guess: the #1269 bundle guard arbitrates the page so the newer copy runs and the older stands down ("A stale copy must never win just because its directory sorts first" — `web/js/lib/duplicate-panel-guard.js`). The newest install therefore names the bundle the browser will actually run. Two incomparable versions keep the first-found copy, which is today's behaviour.

## Change 2 — the refusal names the cause that applied

The text opened by asserting the HTTP cache and sent the user to `Cmd+Shift+R`. The panel repo has since **measured** the opposite and recorded it in `__init__.py`:

> "ComfyUI 0.31.1 already answers `Cache-Control: no-store` on every `/extensions/` path — ours and other packs' alike — **so the HTTP cache is not the mechanism there.**"

The reporter had no cache to clear; their tab had simply been open since before an in-place ComfyUI-Manager update (tab 11:11, update 11:37, no restart, no reload). Causes are now ranked:

1. **the pack was updated in place while this tab was open** — quoting Manager's own "After restarting ComfyUI, please refresh the browser"
2. **the pack is installed twice** — named second because it is the one possibility a reload provably *cannot* fix, so leaving it out means the user reloads, stays broken, and has no next step
3. not a panel tab

A plain reload leads; the cache-bypassing reload stays as the escalation. Same action, corrected cause and order.

## Evidence — by mutation, not by a green suite

| mutation | result |
|---|---|
| `readInstalledPanelVersion` reverted to first-match-wins | ✗ `prefers the NEWEST copy irrespective of which dir name holds it` — **exactly 1 failure** |
| `panel-recovery.ts` restored to `origin/main` | ✗ 5 failures, incl. both new #1515 tests |

Worth stating plainly: the *other* new test — "stale clone first in the list" — **still passes** under the first mutation, because reordering `PANEL_DIR_NAMES` alone satisfies it. The order-independence test is the one carrying the fix, and that is why it exists.

The bridge-level assertions in `ui-bridge.test.ts` go through the real bridge, so they pin that production reaches this text and not just the helper.

**Two negative assertions were re-anchored, deliberately.** They pinned `"This BROWSER TAB is running an older cached copy"`, which this PR renames to `"an OLDER copy"`. Left alone they would still have passed — against a string that no longer exists anywhere — a weakened assertion wearing a green tick.

- **The persistent reload banner / server-side push** suggested in the issue. That is a feature, and the freeze is active.
- **The sticky `latestPanelVersion` cache** (`orchestrator/index.ts:2026`). It is never cleared, so once any hello carries a version the `??` in `gatherEnvCapabilities` makes the disk fallback unreachable for the process's life. It is **not** what produced 0.7.3 here (a 0.7.3 bundle cannot advertise 0.7.3), and changing which version the ENV line prefers is a semantics call that could regress bug-report version-matching. Flagging rather than folding it in.
- **Adjacent bug, not fixed here:** the panel's cache backstop is scoped to `_ASSET_PREFIX = "/extensions/comfyui-mcp-panel/"`, so a Registry install (`comfyui-agent-panel`) gets no backstop on hosts that set no policy. Only matters pre-0.31.1, and it is the panel repo's to fix.

## Browser verification

Not applicable and not claimed: this changes orchestrator-side text and a disk read. No panel JS, no sidebar rendering path is touched.

## What I want a reviewer to attack

1. **Is "newest wins" the right tie-break?** It rests on the #1269 guard being present in the copy that wins the page. If both copies predate that guard, the page arbitration doesn't happen and the newest-on-disk is not necessarily the one running. I claim that's still strictly better than list order — but it is the load-bearing judgement call here.
2. **The 0.7.3 provenance is inferred, not observed on the reporter's machine.** I proved 0.7.3 *can* produce all three symptoms; I could not prove it *did*. If you can construct another path to `panel 0.7.3` on the ENV line, the disk read may not be the whole story.
3. **Cause ordering is a judgement.** I ranked in-place-update first on the reporter's evidence; if the two-panels state is actually more common in the wild, ranks 1 and 2 should swap.

Refs artokun/comfyui-mcp-panel#1515

### DIFF
--- a/src/__tests__/services/env-capabilities.test.ts
+++ b/src/__tests__/services/env-capabilities.test.ts
@@ -603,6 +603,53 @@ describe("readInstalledPanelVersion (disk fallback)", () => {
     expect(readInstalledPanelVersion(dir)).toBe("0.12.0");
   });
 
+  // panel #1515 — TWO copies installed, the documented #1269/#641 two-panels state
+  // (a git clone at comfyui-mcp-panel beside a Registry install at
+  // comfyui-agent-panel). Both carry the same authoritative [project].name, so the
+  // name screen passes for both and the scan used to return whichever dir the
+  // hardcoded list named first. On the reporter's rig that stamped an abandoned
+  // 0.7.3 clone onto the agent's ENVIRONMENT line while the live pack was 0.15.24.
+  it("prefers the NEWEST copy when the pack is installed twice (stale clone first in the list)", async () => {
+    await writePanel(
+      "comfyui-mcp-panel",
+      `[project]\nname = "comfyui-agent-panel"\nversion = "0.7.3"\n`,
+    );
+    await writePanel(
+      "comfyui-agent-panel",
+      `[project]\nname = "comfyui-agent-panel"\nversion = "0.15.24"\n`,
+    );
+    expect(readInstalledPanelVersion(dir)).toBe("0.15.24");
+  });
+
+  // The SAME rule with the directories swapped. Without this, reordering
+  // PANEL_DIR_NAMES alone would satisfy the test above while leaving the defect —
+  // a list's order still deciding the answer — fully intact.
+  it("prefers the NEWEST copy irrespective of which dir name holds it", async () => {
+    await writePanel(
+      "comfyui-agent-panel",
+      `[project]\nname = "comfyui-agent-panel"\nversion = "0.7.3"\n`,
+    );
+    await writePanel(
+      "comfyui-mcp-panel",
+      `[project]\nname = "comfyui-agent-panel"\nversion = "0.15.24"\n`,
+    );
+    expect(readInstalledPanelVersion(dir)).toBe("0.15.24");
+  });
+
+  it("ignores a squatted copy and still reports the real panel's version", async () => {
+    // The name screen and the newest-wins scan compose: 9.9.9 is the highest
+    // version present but is not the panel, so it must not win.
+    await writePanel(
+      "comfyui-mcp-panel",
+      `[project]\nname = "some-other-node"\nversion = "9.9.9"\n`,
+    );
+    await writePanel(
+      "comfyui-agent-panel",
+      `[project]\nname = "comfyui-agent-panel"\nversion = "0.15.24"\n`,
+    );
+    expect(readInstalledPanelVersion(dir)).toBe("0.15.24");
+  });
+
   it("returns undefined for a non-panel pyproject squatting a known dir (wrong project name)", async () => {
     // Authoritative guard: only [project].name == comfyui-agent-panel is trusted,
     // so a different node at the same dir name never yields a WRONG version.
--- a/src/__tests__/services/panel-recovery-cluster.test.ts
+++ b/src/__tests__/services/panel-recovery-cluster.test.ts
@@ -493,13 +493,15 @@ describe("recovery guidance depends on the session, not on a hardcoded string", 
 describe("disk-current but handshake-old is diagnosed as a stale tab, not a stale install", () => {
   const SKEW = { diskVersion: "0.11.38", requiredVersion: "0.11.35" };
 
-  it("tells the user to hard-refresh, and NOT to update anything", () => {
+  it("tells the user to reload the tab, and NOT to update anything", () => {
     const text = describePanelUpdateRecovery(undefined, {
       ...SKEW,
       handshakeVersion: "0.11.34",
     });
     expect(text).toMatch(/Do NOT update the panel/);
-    expect(text).toMatch(/HARD-REFRESH/);
+    expect(text).toMatch(/RELOAD THIS COMFYUI TAB/);
+    // The cache-bypassing reload survives as the ESCALATION (panel #1515), so
+    // the keystroke a user needs when a plain reload does not take is still here.
     expect(text).toMatch(/Ctrl\+Shift\+R/);
     expect(text).toMatch(/0\.11\.38/); // what is really on disk
     expect(text).toMatch(/0\.11\.34/); // what the tab announced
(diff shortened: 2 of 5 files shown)


artokun:
@grok please review this one. Rather than hand you the diff, here is where I think it breaks.

**1. The load-bearing judgement: "newest wins" as the tie-break.**
`readInstalledPanelVersion` used to return the first readable of two `custom_nodes` dir names; it now scans both and returns the newest. My justification is that the panel's #1269 bundle guard arbitrates the page so the newer copy runs and the older stands down — so newest-on-disk names the bundle the browser will actually run. **That reasoning fails if both installed copies predate that guard**, because then no arbitration happens and ComfyUI's own load order decides. I claim newest is still strictly better than an arbitrary hardcoded list order, but this is the call most worth attacking. Is there a case where first-listed was actually load-bearing and I broke it?

**2. The root cause is inferred, not observed.** I proved a leftover panel 0.7.3 install *can* produce all three reported symptoms: 0.7.3 is a real release (`43b46698`), its pyproject declares `name = "comfyui-agent-panel"` so it passes the authoritative screen, and it predates 0.11.3 when `panel_version` entered the hello — so it also hellos with no version. What I could **not** do is observe the reporter's disk. If you can construct another path to `panel 0.7.3` on the ENVIRONMENT line, my Change 1 may be treating a symptom.

**3. The state I am least sure of: incomparable versions.** `compareSemver` returns 0 for anything that is not `major.minor.patch`, so two unparseable versions silently keep the first-found copy. I convinced myself that is fine because `[project].version` in a released pyproject is always a real semver and "nightly" is a ComfyUI-Manager artifact, not a pyproject value. If that assumption is wrong, the fix quietly degrades to the old behaviour for exactly the users most likely to have two copies (git-installed ones).

**4. Copy change — did I overcorrect?** I demoted the HTTP-cache explanation because the panel repo's `__init__.py` records a *measured* finding that ComfyUI 0.31.1+ answers `Cache-Control: no-store` on `/extensions/`. The cache-bypassing reload is still there as the escalation. But if there is a real host/browser combination where a plain reload does serve stale JS, leading with "an ordinary reload is normally enough" costs that user a round trip — the mirror image of the bug I am fixing.

**5. Assertion integrity.** I re-anchored two `not.toMatch` assertions that pinned a string this PR renames. Left alone they would have passed vacuously. Please confirm I re-anchored rather than weakened them — and that the ordering assertion in `panel-recovery-cluster.test.ts` (`indexOf` comparison) actually pins order and not just presence.

Evidence I ran: deleting the disk fix fails exactly one test (`prefers the NEWEST copy irrespective of which dir name holds it`); reverting `panel-recovery.ts` to `origin/main` fails five including both new tests. Full local suite 590/590 files, 10851 tests green; CI green on all three platforms. **No browser run** — nothing panel-rendered is touched.

artokun:
Codex independent merge gate: PASS for the eligible P2 fix.

Verified against the literal PR diff and current callers:
- readInstalledPanelVersion continues scanning both recognized install directories, filters by the authoritative project name, and selects the highest comparable semver; squatted directories and non-comparable versions remain fail-safe.
- Recovery guidance now separates the observed stale-tab condition from the unproven cache mechanism, names duplicate installs as the reload-resistant case, and preserves hard-refresh escalation.
- Tests exercise both directory orderings, squatted copies, guidance ordering, and the UiBridge production path.

CI is fresh green on ubuntu, windows, macOS, and pack/install smoke. No P0/P1 or enhancement scope is present. No browser gate is required for these orchestrator-side changes.

PR #1900, opened 2026-08-20
fix(panel#1517): a read-only graph command is sent while a prompt runs

artokun:
Reported as artokun/comfyui-mcp-panel#1517 (`via-panel`, P2). The code is here, not
in the panel: the panel has no execution gate on `graph_*` at all
(`git grep` over `web/js` finds no running-prompt refusal), so the refusal was
entirely the orchestrator's pre-dispatch fence.

## What the reporter hit

Queue a workflow, then call `panel_graph_outline` while it runs:

```
graph_outline was NOT sent — nothing was applied. QUEUE BUSY: a ComfyUI prompt is
running (...). The panel tab typically cannot answer graph_* commands (including
read-only graph_query / graph_outline) while a prompt is executing.
```

They wanted to inspect and correct a prompt warning **for the next run** — a pure
read, refused for the whole duration of a render.

## Mechanism

`graphCmdBlockedByRunningPrompt` (`src/orchestrator/panel-tools.ts`) refused every
`graph_*` except `graph_run` while `QueueMonitor` saw a running prompt. That fence
was added by #1745 for #1639.

Reading #1639 closely, its reporter asked for two things, in order:

> Either (a) keep the panel's command channel responsive during execution (**the
> reads at minimum should never be blocked by a running prompt**), or (b) if the
> frontend genuinely cannot answer mid-execution, have the orchestrator … say so
> explicitly instead of surfacing a generic 20/30s timeout.

#1745 shipped (b) — and applied it to reads as well as edits. #1517 is the cost of
that: (a) was never done, and (b) took reads away.

**The refusal's own justification is about an unknown MUTATION outcome.** A read has
no outcome to be unknown about. So the fence is now scoped by EFFECT:

```ts
if (!name.startsWith("graph_") || name === "graph_run") return null;
if (!isMutatingGraphCommand(name)) return null;   // ← panel#1517
```

`isMutatingGraphCommand` is the existing #778 `GRAPH_CMD_EFFECT` ledger, not a new
list — deliberately, since #778 exists because that knowledge had been duplicated
into a third list the gate could not see. It **fails closed**: an unclassified
`graph_*` is `targeted`, stays refused here, and is *recorded* as a ledger gap.

Edits are unchanged. The refusal text now also names the reads that remain
available instead of claiming they are blocked.

## Blast radius was wider than the two tools in the title

`ctx.call` routes **every internal read** through this same guard, so while a prompt
ran:

- `panel_get_errors` (`graph_get_errors`) — the prompt-warning surface the reporter
  was actually reaching for — was refused without being sent.
- `probeLiveGraphUnderCurrentFence` — `workflow_open`'s own live-graph probe, added
  by the *other half of #1745* — could not run, so an open could not clear its fence
  during a render either.

Both are fixed by the same one-line scope change; `panel_get_errors` is pinned below.

## Evidence

**Mutation-verified.** Deleted the fix line and re-ran — 4 named tests go red with
exactly the production symptom (`sent` is empty, i.e. never dispatched):

```
× panel_graph_outline is sent and answers — the reported call
    AssertionError: expected [] to deeply equal [ 'graph_outline' ]
× panel_query_graph is sent — the other call named in the report
    AssertionError: expected [] to deeply equal [ 'graph_query' ]
× panel_get_errors is sent — the prompt-warning surface the reporter wanted
    AssertionError: expected [] to include 'graph_get_errors'
× a read that DOES time out still gets the QUEUE BUSY explanation, not a frozen tab
    AssertionError: expected [] to include 'graph_outline'
 Tests  4 failed | 7 passed (11)
```

Restored → 11/11 pass. The tests drive the **real tool surface**
(`buildPanelToolDefs().find(...).handler`), not a helper, so this is the call site.

**Suites:** `src/__tests__/orchestrator` + `src/__tests__/services` — 420 files,
8422 passed, 0 failed. `tsc --noEmit` clean.

**Full `npm test`** (590 files) showed 4 and then 10 failures across two runs, but a
*different, disjoint* set each time — all hook/test timeouts under parallel load.
Re-ran all 6 implicated files with `--maxWorkers=1`: **6 files / 323 tests passed**.
None of them import the changed code.

**Not run: Playwright.** This change is entirely orchestrator-side; the panel repo
is untouched and nothing it renders changes. I am not claiming browser coverage.

## The trade, stated plainly

When the tab genuinely cannot service its main thread, a read now costs up to the
bounded read ack budget (20 s) and then returns `queueBusyTimeoutNote()` — the *same*
QUEUE BUSY text the refusal produced instantly. So the **outcome** is never worse
than today, only later; the upside is the answer itself, which a tab that IS
servicing its main thread returns in milliseconds.

I deliberately did **not** shorten the read budget for this case. 20 s is the
`BRIDGE_READ_DEFAULT_TIMEOUT_MS` set by #357 precisely because a healthy-but-loaded
tab is slow, and a running prompt is definitionally that condition — capping it lower
would cut off exactly the reads that need the tolerance.

## Review pending

Not gated, not reviewed yet — requesting review in a comment.

### DIFF
--- a/src/__tests__/orchestrator/graph-read-during-render.test.ts
+++ b/src/__tests__/orchestrator/graph-read-during-render.test.ts
@@ -6,11 +6,19 @@
 //
 // The frontend main thread is what cannot answer. The orchestrator CAN see the
 // running prompt via QueueMonitor (HTTP /queue, independent of the panel tab).
-// Fail closed BEFORE dispatch for canvas-touching graph commands so the agent
-// gets an explicit QUEUE BUSY instead of a 20/30s unknown-outcome timeout.
-// `graph_run` is excluded: queuing behind an in-flight job is the documented
-// sweep path. A timeout that still happens (graph_run, or a lagging snapshot)
-// is annotated with the same QUEUE BUSY rather than "backgrounded or frozen".
+// Fail closed BEFORE dispatch for graph EDITS so the agent gets an explicit
+// QUEUE BUSY instead of a 20/30s unknown-outcome timeout. `graph_run` is
+// excluded: queuing behind an in-flight job is the documented sweep path. A
+// timeout that still happens is annotated with the same QUEUE BUSY rather than
+// "backgrounded or frozen".
+//
+// panel#1517 — #1745 applied that refusal to READS too, which cost read-only
+// inspection for the whole duration of a render: an agent could not read the
+// prompt warning it was trying to correct for the next run. A read has no
+// mutation outcome to be unknown about, so an `inert` command (GRAPH_CMD_EFFECT)
+// is DISPATCHED while a prompt runs; if the tab genuinely cannot answer it, the
+// bounded read budget produces the same QUEUE BUSY text on the timeout path.
+// The edit refusal is unchanged — that is what the unknown outcome is about.
 
 import { afterEach, beforeEach, describe, expect, it } from "vitest";
 
@@ -21,7 +29,7 @@ import {
   type ToolResult,
 } from "../../orchestrator/panel-tools.js";
 import { QueueMonitor } from "../../services/queue-monitor.js";
-import { markReplyTimeout } from "../../services/ui-bridge.js";
+import { isMutatingGraphCommand, markReplyTimeout } from "../../services/ui-bridge.js";
 import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
 
 const TAB = "11111111-2222-4333-8444-555555555555";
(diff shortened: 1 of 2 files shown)


artokun:
@grok please review this one. It **narrows a safety fence**, so I want the load-bearing
claims attacked rather than the diff skimmed. Named, in the order I am least sure of them:

**1. Is `GRAPH_CMD_EFFECT === "inert"` really the right discriminator for "safe to send
while a prompt runs"?** This is the whole fix. That ledger was built for a *different*
question — #778's workflow fence, "can this change the user's workflow after a tab
switch". I am reusing it to answer "does this have a mutation outcome that could be
unknown". I believe those coincide, but that is an argument, not a proof. The commands
this newly lets through while a render is executing are not only pure reads — they
include `graph_canvas` (pans the viewport by a **dx/dy delta**), `graph_select_nodes`,
`graph_enter_subgraph` / `graph_exit_subgraph`, `graph_copy_nodes`, `graph_screenshot`.
Those change what the user is *looking at* mid-render. Inert for the fence, yes — but is
any of them harmful or surprising to fire into a live execution? `graph_canvas` is the
one I would attack first: it is explicitly *not* idempotent, and it is now dispatchable
in a window where a timeout-then-retry is plausible.

**2. Does the timeout path actually still produce QUEUE BUSY for a read?** My whole
"never a worse outcome, only later" claim rests on `queueBusyTimeoutNote()` being
appended when a *read* times out. I pinned that with one test, but I reached it through
a mocked bridge that throws a `markReplyTimeout` error. If the real read path can exit
through a `fail(...)` that skips that annotation, the claim is false and a read now
degrades to the generic "tab may be backgrounded or frozen" the fence existed to
prevent. `callOnce` has a dozen-plus fail exits — I checked three.

**3. Did I under-fix?** `ctx.call` routes internal reads through this guard, and I found
two callers that were silently broken by the over-block (`panel_get_errors`, and
`probeLiveGraphUnderCurrentFence`, which is the other half of #1745 — `workflow_open`'s
fence probe could not run during a render). I only pinned the first. Are there other
internal `graph_*` reads whose behaviour changes here in a way I have not thought about
— particularly any that ran *pre*-#1745, silently started failing, and have since grown
a fallback that will now behave differently again?

**4. Regression risk against #1639.** That reporter's complaint was ~12 × 20 s of dead
wall-clock. I am reintroducing the possibility of that wait for reads and arguing it is
worth it, and I explicitly declined to shorten the read budget (reasoning in the PR
body). Tell me if that reasoning is wrong.

Mutation evidence and what I did *not* run (Playwright — orchestrator-only change) are in
the PR body. Not merging until this comes back.

artokun:
Closing — **superseded, not abandoned.** #1892 (`9316e5f`) landed the same fix for
panel#1489 at 19:38Z, 26 minutes after I claimed panel#1517 and while this branch was
mid-test. panel#1517 and panel#1489 are the same report.

The two fixes converged on the same mechanism independently, which is some evidence it
is the right one: both scope `graphCmdBlockedByRunningPrompt` by `GRAPH_CMD_EFFECT`
rather than by `BRIDGE_READONLY_CMDS`, both fail closed on an unlisted command, and both
deliberately keep the tolerant read ack budget. #1892's reasoning for that last point is
better than mine — I cited #357; it also cites #589, where `panel_get_errors` was given a
30 s budget because the panel's own bound is 18 s, so a cap here would have re-broken it.

I verified the merged fix by execution rather than by reading the diff: checked this
worktree out to pristine `origin/main` and ran the tests I had written independently
against it.

```
✓ panel_graph_outline is sent and answers — the reported call
✓ panel_query_graph is sent — the other call named in the report
✓ panel_get_errors is sent — the prompt-warning surface the reporter wanted
✓ panel_set_widget is NOT delivered — outcome is known (nothing applied)
✓ an UNCLASSIFIED graph_* command still fails closed as an edit
```

Three of my eleven failed on main, all of them assertions on **my** wording rather than
on behaviour — #1892 also reworded `queueBusyTimeoutNote`, and its version is more
honest than mine ("the most likely reason the tab did not answer" instead of asserting
"this is not a backgrounded or frozen tab", which was a prediction stated as a fact).

Nothing here is worth salvaging on top of it. The one behavioural difference was cosmetic
— my refusal text named the read tools that remain available — and the freeze is not the
time for that. #1892's suite already pins `panel_get_errors` too, so there is no coverage
gap to fill.

@grok — please disregard the review request above; no review needed on a closed PR.

Closing panel#1517 as fixed by #1892, shipped in **v0.52.29**.

Issue #1329, opened 2026-08-10
panel_add_node requires manual refresh after image upload


artokun:
### What happened / root cause
After uploading two images with upload_image(action:"image"), panel_add_node("LoadImage") rejected both additions because the live tab's cached node schema was stale. The tool already detects this exact condition but requires the caller to make a separate panel_refresh_nodes call and retry.

### Steps to reproduce
1. Open a workflow/subgraph in the sidebar panel.
2. Upload an image with upload_image(action:"image").
3. Call panel_add_node with class_type "LoadImage".
4. It refuses with the stale-schema error.
5. Call panel_refresh_nodes, then repeat panel_add_node; it succeeds.

### Exact error
"LoadImage" required input "image" was added or retyped since this page loaded its node schema, so creating it now would build the OLD shape. Call panel_refresh_nodes and retry.

### Fix
Applied locally: no. Upstream-only: yes.
Suggested minimal change: when panel_add_node's stale-schema guard detects this condition, automatically refresh /object_info and retry the add exactly once. Preserve the current refusal only if that refresh/retry fails. This removes an agent-visible error and duplicate mutation call without weakening the schema safety check.

### Verification
The documented workaround was verified: panel_refresh_nodes returned refreshed:true, and both subsequent LoadImage additions succeeded.

### Environment
Linux aarch64; NVIDIA GB10 121.6 GB VRAM; ComfyUI 0.31.1; ComfyUI frontend 1.48.7; Python 3.12.3; PyTorch 2.13.0+cu130; comfyui-mcp 0.50.87; panel 0.11.78.

Related (same root cause): #608

PR #1902, opened 2026-08-20
fix(panel#1518): an add refused for a stale schema takes its own retry after a lost refresh ack

### LINKED ISSUE #1518: panel_add_node LoadImage false-refuses after schema refresh timeout; immediate retry succeeds
On a brand-new empty workflow, panel_add_node for core LoadImage refused because required input image was reported as added/retyped since page load. Its automatic panel_refresh_nodes dispatch timed out after 30 seconds and the tool returned an error, although the refresh continued. An immediate identical panel_add_node retry succeeded. This is a refresh/ack race that exposes a false failure and forces an awkward retry.

### Steps to reproduce
1. Open a new workflow tab with panel_new_workflow.
2. Call panel_add_node with class_type LoadImage.
3. Observe schema-stale refusal and refresh_nodes timeout.
4. Retry identical add; it succeeds immediately.

### Exact error (scrubbed)
"LoadImage required input image was added or retyped since this page loaded its node schema... panel_refresh_nodes was dispatched and did NOT answer within its window... Panel tab tmp:[REDACTED] did not reply to refresh_nodes within 30000 ms."

### Fix
Precise change needed: panel_add_node should await/recognize the still-running refresh or transparently retry the add once after a refresh timeout, with duplicate protection, instead of returning a false terminal failure.

### Environment
Windows 11; ComfyUI 0.33.0; frontend 1.49.6; Python 3.13.11; PyTorch 2.13.0+cu130; RTX 5090 32GB; comfyui-mcp 0.52.27; panel 0.14.24.

artokun:
Fixes artokun/comfyui-mcp-panel#1518.

## What the reporter saw

`panel_add_node("LoadImage")` on a fresh workflow was refused for a stale node schema. The
automatic `panel_refresh_nodes` recovery did not answer inside its 30 s ack window, and the
tool returned an error. **An immediate, identical `panel_add_node` retry succeeded.**

## The mechanism, and why this is the *second* half of #1491

This exact branch was touched 7 h ago by #1864 (panel#1491). That change stopped it *lying*:
before it, an ack timeout was reported as a failed refresh — *"the schema is unchanged and
retrying the add will refuse again"* — which argues the caller out of the one action that
works. #1864 replaced that with the true statement and the correct advice: a refresh that
outruns its ack is not cancelled, so **"RETRY the add — that is the action that clears this."**

panel#1518 is the report that the advice is right and the tool still hands back an error.
The reporter followed it by hand and it worked on the first try.

So: this path knows the single correct next action, has no decision to make, and its only
reason for not taking it was a **lost ACK** — not a failed refresh, and not an unknown
mutation outcome. It now takes it.

```
before:  add → REFUSED → refresh → ack lost → return the refusal + "please retry the add"
after:   add → REFUSED → refresh → ack lost → add again → return that
```

**The ack-timeout branch is proven live for this issue, not assumed.** The error quoted in
panel#1518 contains *"was dispatched and did NOT answer within its window"* — wording that
`panel-tools.ts` emits **only** when `isReplyTimeoutResult(refreshed)` is `true`. That is the
predicate this fix forks on, and the report is field evidence that it fired.

### Duplicate protection is the *existing* argument, not a new one

The only way control reaches the retry is `isStaleNodeSchemaRefusal(first)` — an **acked**
refusal whose own text says the panel's guard threw *before* creating anything ("Refuse before
creating anything", `comfyui-mcp-panel.js`). A first add whose outcome is **unknown** (a
timeout, a dead socket) is not a stale-schema refusal and returned many lines earlier. This is
the identical licence the acked-success retry below it has run on since #1329; it is not an
extension of it. Mutations with unknown outcomes are still never re-issued.

And the retry is not a hope: `graph_add_node` gates itself on a **fresh** `/object_info`
(`assertAddNodeResolvableRefreshing`, #599), so the second add re-fetches the schema itself
rather than re-reading whatever the abandoned refresh happened to leave behind.

### What still does NOT retry

A **genuinely failed** refresh — the panel acked and said no — keeps the original refusal and
the original wording, with no second add. There the schema really is unchanged and a retry
really would refuse again; softening that would be the opposite defect. The fork is read from
the bridge-owned `REPLY_TIMEOUT_RESULT` symbol, never from message text (#1468) — the panel
relays arbitrary error strings, so an acked error can quote the timeout wording verbatim.

### Two honesty fixes that fall out of it

- **Still stale after the retry** no longer inherits the acked-success wording. "`panel_refresh_nodes`
  reported success", "repeating it will not clear it" and "reload the tab" are all overclaims
  when nothing observed that refresh finish. It now says the registration had not landed *yet*,
  that neither attempt added anything, and to retry in a few seconds — reload only as fallback.
- **A non-stale error from the retry** says which attempt it came from. Otherwise a bare
  `graph_add_node` timeout reads as the first add's outcome and leaves the number of in-flight
  mutations ambiguous.

## Evidence

`src/__tests__/orchestrator/add-node-stale-schema-retry.test.ts` — 15/15 green (6 new). The
timeout is minted by the bridge's own `markReplyTimeout`, imported from `services/ui-bridge.js`,
not a hand-rolled shape.

**Verified by mutation — each one deletes a load-bearing piece and a named test goes red:**

| mutation | result |
|---|---|
| fork removed (`if (refreshed.isError)` — restores today's early return) | **5 failed**, incl. the reporter's case *and* #1491's own regression test |
| decide on message text instead of the bridge tag | **2 failed** — an acked error quoting the timeout wording would issue a real second mutating add |
| retry-attribution disclosure removed | **1 failed** |
| still-stale rider always uses the acked-success wording | **2 failed** |

`tsc --noEmit` clean. Full suite: 10843 passed / 9 failed across 5 files — all five are
wall-clock timeouts under parallel load (`gen-changelog-dedupe`, `panel-restart-configured-command`,
`panel-base-clear-epoch`, `tool-surface-filter`, `vocabulary`, plus `ui-bridge` on a prior run;
the set is not even stable between runs). **All six pass green when re-run isolated
(`--no-file-parallelism`): 323/323.** None touch `panel_add_node`.

- **No panel-repo change, so no Playwright run.** This PR changes zero files in
  `comfyui-mcp-panel` and renders nothing; the panel browser suite would exercise nothing here.
  Stated plainly rather than implying coverage I do not have.
- **Did not widen `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS`.** The panel's own
  `REFRESH_NODES_COMMAND_BUDGET_MS` (25 s) is derived against this 30 s window; raising one
  side silently invalidates that derivation.
- **Did not add a wall-clock bound to the retry.** Worst case is `add(30) + refresh(30) +
  add(30)` = 90 s, but that is the shape the acked-success path has already had since #1329 —
  this introduces no new ceiling.
- **Did not touch the reporter's other problem: they are on panel 0.14.24.** panel#1404
  (`REFRESH_NODES_COMMAND_BUDGET_MS`, panel v0.15.7, 2026-08-19) is what makes `refresh_nodes`
  *reply* inside the window instead of blowing past it, and they predate it. Upgrading is the
  bigger win for them; this fix is what makes the residual case — the budget cannot interrupt
  `registerNodesFromDefs`/`reapplyDefsToLiveNodes`, by its own docstring — stop being a false
  terminal failure. Noted on the issue.

## Review — NOT obtained. Read this before trusting the change.

**No independent review exists on this PR, and none is coming.** Stated plainly rather than left as "pending", because this body is the permanent record:

- `@grok` was asked (comment below) and has **never replied to any PR in either repo** — the account does not appear among either repo's commenters. PR #1900 shows another agent making the same request an hour earlier and getting nothing.
- `copilot-pull-request-reviewer` auto-reviewed and declined: *"the user who requested the review has reached their quota limit."* Same on #1890, #1886, #1885, #1884. Re-requesting fails with a 422 (not a collaborator).
- No review-bot workflow exists in `.github/workflows/`.
- codex is barred as a reviewer of record by standing instruction (repointed at a weak model), so it is not a substitute.

**So this merged on author-side evidence alone.** That is a real weakness and I am not dressing it up. What was done instead of a review:

1. **Four mutations, each red against a named test** — including one on the *decision predicate* itself (bridge tag → message text), which proved a text-matching implementation would issue a real second mutating add. Re-run and still red after the rebase onto current `main`.
2. **The reported state reproduced by execution, not by reading.** The error quoted in panel#1518 contains `was dispatched and did NOT answer within its window` — a string emitted **only** when `isReplyTimeoutResult(refreshed)` is true. The reporter's transcript is proof the branch this forks on fires in production.
3. **The duplicate-safety precondition checked across the whole shipped population, not argued.** Swept every released panel tag: **55/55 put the drift `throw` before `LG.createNode(class_type)` — zero violations, and no released version lacks the guard.** So "a stale-schema refusal proves nothing was created" holds for every panel version in the field, not just current.
4. **An overclaim of my own retracted mid-PR.** I first wrote that the panel Playwright suite "would exercise nothing here" as a guess. `browser_tests/fixtures/MockBridge.ts` is an *"agent-free fake of the panel orchestrator"*, so that suite is structurally incapable of reaching this code. Coverage is orchestrator unit tests plus the mutations, **and nothing else**. No end-to-end run exercised it.

**Where a later reviewer should attack first**, in the order I am least sure:

1. `isReplyTimeoutResult`'s `isError === true` self-guard, now that I evaluate it **outside** the `refreshed.isError` block. If that guard is weaker than I read it, the acked-success path starts emitting timeout wording.
2. Whether the retry deserves a wall-clock bound of its own. Worst case is `add(30) + refresh(30) + add(30)` = 90 s. I argue that is not a new ceiling — the acked-success retry has had it since #1329 — and chose not to add one. That choice is not strongly held.
3. The wording of the still-stale-after-timeout rider: it must not claim the refresh succeeded, must not claim it failed, and must not talk someone out of a reload that is genuinely warranted.

CI: green on ubuntu / macos / windows plus the pack & install smoke, on the rebased head.

### DIFF
--- a/src/__tests__/orchestrator/add-node-stale-schema-retry.test.ts
+++ b/src/__tests__/orchestrator/add-node-stale-schema-retry.test.ts
@@ -36,11 +36,23 @@ const STALE = () =>
       `schema, so creating it now would build the OLD shape. Call panel_refresh_nodes and retry.`,
   );
 
+/** panel#1518 — a realistic NON-stale failure of the retried add: the panel's own
+ *  `addNodeRefreshBusyMessage`, which is what a second add hits when the refresh this
+ *  path abandoned is still holding the coalescer slot. */
+const BUSY = () =>
+  new Error(
+    `Cannot add "LoadImage" right now: a node-def refresh was still running, and waiting for ` +
+      `it would have used up this command's 25s budget. NOTHING WAS ADDED. RETRY in a few seconds.`,
+  );
+
+type AddOutcome = "stale" | "ok" | "busy";
+
 /**
  * `addOutcomes` is consumed one per graph_add_node call: "stale" throws the refusal,
- * "ok" succeeds. `refreshFails` makes refresh_nodes throw.
+ * "busy" throws an unrelated (non-stale) refusal, "ok" succeeds. `refreshFails` makes
+ * refresh_nodes throw.
  */
-function bridge(addOutcomes: Array<"stale" | "ok">, refreshFails: boolean | "timeout" | "acked-error-quoting-timeout" = false) {
+function bridge(addOutcomes: AddOutcome[], refreshFails: boolean | "timeout" | "acked-error-quoting-timeout" = false) {
   const calls: string[] = [];
   let addIdx = 0;
   const b = {
@@ -70,6 +82,7 @@ function bridge(addOutcomes: Array<"stale" | "ok">, refreshFails: boolean | "tim
         const outcome = addOutcomes[Math.min(addIdx, addOutcomes.length - 1)];
         addIdx += 1;
         if (outcome === "stale") throw STALE();
+        if (outcome === "busy") throw BUSY();
         return { ok: true, node_id: 42 };
       }
       return { ok: true };
@@ -85,7 +98,7 @@ function bridge(addOutcomes: Array<"stale" | "ok">, refreshFails: boolean | "tim
   return { b, calls };
 }
 
-async function addNode(addOutcomes: Array<"stale" | "ok">, refreshFails: boolean | "timeout" | "acked-error-quoting-timeout" = false) {
+async function addNode(addOutcomes: AddOutcome[], refreshFails: boolean | "timeout" | "acked-error-quoting-timeout" = false) {
   const { b, calls } = bridge(addOutcomes, refreshFails);
   const ctx = makePanelToolCtx(b, TAB, new WorkflowTargetStore());
   const def = buildPanelToolDefs().find((d) => d.name === "panel_add_node");
@@ -176,10 +189,13 @@ describe("#1491: an auto-refresh that outran its ack is NOT a failed refresh", (
     // panel_refresh_nodes moments later succeeded immediately. `joinMs` does not cancel
     // work already in flight — the panel's coalescer keeps running and registers what it
     // fetched — so on a large install the refresh legitimately outlives this ack.
-    const { text } = await addNode(["stale", "stale"], "timeout");
+    const { text, calls } = await addNode(["stale", "stale"], "timeout");
 
     expect(text).toContain("did NOT answer within its window");
     expect(text).toContain("RETRY the add");
+    // panel#1518 — and the advice is only reached because the tool TOOK the retry first
+    // and it was still stale. Bounded: exactly two adds, one refresh, never a loop.
+    expect(calls).toEqual(["graph_add_node", "refresh_nodes", "graph_add_node"]);
     // The two claims that were false, and the second is the damaging one: it argues the
     // caller out of the only action that works.
     expect(text).not.toContain("so the schema is unchanged");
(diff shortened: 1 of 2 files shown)


artokun:
@grok please review this one. It makes the orchestrator **re-issue a mutating command on its own initiative**, so the safety argument is the whole review, not the diff. Here is where I think it can be wrong, ranked:

**1. The duplicate-safety claim, which is the only thing standing between this and a double-add.**
I claim the retry can never re-issue a mutation whose outcome is unknown, because the sole route to it is `isStaleNodeSchemaRefusal(first)` — an *acked* refusal from a guard that throws before creating anything. Attack it: is there any path where `first` carries that drift sentence and the panel HAS already created a node? Consider a relayed/quoted error (the panel forwards arbitrary text — that is exactly the hole #1468 found in this same function), an aggregate error from a batched add, or a panel old enough to word its refusal that way *after* mutating. If such a shape exists, this PR puts a second node on the user's canvas and that is the P1 here.

**2. The fork restructure — I moved `isReplyTimeoutResult` OUT of the `refreshed.isError` block.**
It is now `const refreshOutranItsAck = isReplyTimeoutResult(refreshed)` evaluated unconditionally, then `if (refreshed.isError && !refreshOutranItsAck)`. I am relying on `isReplyTimeoutResult` self-guarding with `res?.isError === true`, so a *successful* refresh can never be classed as an ack timeout and get the wrong rider. If that self-guard is weaker than I read it, the acked-success path silently starts emitting timeout wording. Please check that read directly rather than taking my word.

**3. Reachability of the whole branch.** I argue the `stillRunning` branch is live in production because panel#1518's quoted error contains "did NOT answer within its window", which `panel-tools.ts` emits only when that predicate is true. That is inference from a scrubbed report. Is `carryReplyTimeoutMark` actually applied at the exit `ctx.call({cmd:"refresh_nodes"}, …)` leaves through, or is my fix hanging off a predicate that is false in production for this command specifically? A false predicate here means an inert PR that tests green — a failure mode this repo has shipped before.

**4. The state I am least sure of: what the retried add hits when the abandoned refresh still owns the coalescer slot.** I model it as the panel's `addNodeRefreshBusyMessage` ("NOTHING WAS ADDED … RETRY in a few seconds") and cover it as the `"busy"` fixture. If the real panel instead *blocks* there, the tool call becomes `30 + 30 + 30` = 90 s. I argue that is not a new ceiling (the acked-success retry has had it since #1329), but if you think a bound belongs on the retry specifically, say so — I chose not to add one and that choice is not strongly held.

**5. Wording, since two of the three changes here are only wording.** The still-stale-after-timeout rider must not claim the refresh succeeded, must not claim it failed, and must not send the user to reload the tab as if the registry were stuck. Please check it does not overclaim in the other direction either — I do not want it talking someone out of a reload that is genuinely warranted.

Mutation results are in the PR body: four separate deletions, each verified red against a named test. The one I would trust least is #3, because no mutation of mine can prove production reaches this code.

artokun:
Two of the five things I asked you to attack, I have since chased down myself. Recording what I found so you can spend your time on the ones I could not settle — not to pre-empt the review.

**Claim 1 (duplicate safety) — verified by reading the panel, not by reasoning about it.** In `comfyui-mcp-panel.js` `graph_add_node`, the drift check `throw`s at the top of the guard block; `createNode` is reached only past it, after the socket proof. So `isStaleNodeSchemaRefusal(first)` really does imply nothing was created. The shape I was most worried about — a panel that words its refusal that way *after* mutating — I could not find, but note it is a claim about **every panel version in the field**, not just current, and that is the part I cannot fully close by reading one file.

**Claim 3 (reachability) — settled, and better than inference.** The error quoted in panel#1518 contains `was dispatched and did NOT answer within its window`, a string `panel-tools.ts` emits **only** on `isReplyTimeoutResult(refreshed) === true`. The reporter's own transcript is proof the branch this PR forks on fired in production. `ctx.call`'s tail exit is `carryReplyTimeoutMark(err, fail(...))`, so `refresh_nodes` is not routed around the mark, and the dispatch-refusal branch above it is correctly *not* marked (nothing was sent ⇒ the FAILED wording is the right one there).

**Something that came out of it and strengthens the change.** On a current panel, `graph_add_node` *already* runs the refresh recovery itself (panel#1242, v0.15.4) and its surviving refusal says, verbatim: *"Retrying the add runs the refresh again; if it keeps reporting incomplete, [reload]."* So the panel independently prescribes the exact action this PR now takes — the orchestrator was declining to take advice the panel was giving it.

**And the reporter's environment is further back than I first wrote.** panel 0.14.24 (2026-08-12) predates *both* panel#1242 (self-refresh, v0.15.4) and panel#1404 (`REFRESH_NODES_COMMAND_BUDGET_MS`, v0.15.7). Upgrading is the larger win for them; this PR is what keeps the residual case honest, since that budget cannot interrupt `registerNodesFromDefs`/`reapplyDefsToLiveNodes` by its own docstring.

Still open and still worth your attention: **2** (whether `isReplyTimeoutResult`'s `isError` self-guard is as strong as I read it, now that I evaluate it outside the `refreshed.isError` block), **4** (whether the retry deserves a bound of its own), and **5** (the wording).

artokun:
One correction to the "no Playwright run" note in the PR body — I asserted the panel browser suite would exercise nothing here, and I have now checked rather than asserted it. `browser_tests/fixtures/MockBridge.ts` describes itself as *"a scriptable, agent-free fake of the panel orchestrator"*, and every `.spec.ts` drives the panel against it. The real `comfyui-mcp` orchestrator is never in the loop, so that suite is structurally incapable of reaching `panel_add_node`'s retry path regardless of whether a ComfyUI is up. The panel-side `browser_tests/unit/*.mjs` files that mention `panel_add_node` are testing the panel's own `graph_add_node` handler, which this PR does not touch.

Restating the coverage claim precisely, so it is not read as broader than it is: this change is covered by orchestrator unit tests plus the four mutations, and by nothing else. No end-to-end run exercised it.

artokun:
**Rebased onto current `main`** (`ce63106`) — main had moved 10 commits since the branch point, including `9316e5f` (panel#1489) and `c74a059`/`77b47cb` (panel#1515), which touch this same file. Clean rebase, no conflicts.

Re-verified against the new base rather than trusting the earlier run:
- `tsc --noEmit` clean
- `add-node-stale-schema-retry.test.ts` 15/15
- primary mutation re-run on the rebased tree (fork removed → `if (refreshed.isError)`): **5 failed**, the same named set as before. The pin still bites against current `main`.

CI on the pre-rebase commit was green on all three platforms plus the pack/install smoke; it is re-running now for this base.

PR #1908, opened 2026-08-20
fix(1869): report every skipped action-button token, and stop double-reporting unknown nodes

artokun:
Follow-up to #1880 (merged) — closes out the two findings from that PR's round-2 independent review, which landed after the merge.

Both were confirmed on the rig against the code **now on main**.

## 1. Skipping was silent

`convertUiToApi` skips serialized action-button tokens, but said nothing when it did. Which entry in a run is the button is not always decidable:

```
schema: { path: STRING, first_frame: INT }
row:    ["browse", "detect_range", 12]
```

One extra value, two candidates that both look like buttons. The first-wins rule puts `detect_range` on `path`. That ambiguity is **irreducible** — a closed token vocabulary cannot rule out a path whose value happens to *be* `browse` — but resolving it silently is not acceptable.

Every skipped value is now named, with the widget it was skipped before and what it means if the guess was wrong:

```
Node 7 (N): skipped serialized value "browse" before widget "path" — it is a
frontend-only action button that /object_info does not describe. If that value
was real, this node's inputs may be shifted.
```

Same rule as #361: every place the converted graph differs from the source gets reported.

## 2. UI graphs double-reported unknown nodes

A UI graph with an uninstalled custom node produced **both** the converter's "not found in object_info … Skipping." warning **and** the authoritative `missing_node_type` error added in #1880 — where an API-format graph reports it once. `missingNodeTypes` now carries the exact warning text, so the validator drops the duplicate **by identity** rather than re-matching the prose.

## Test change, flagged explicitly

The #1869 validator test scanned *every* message for the button tokens. Its subject was false **errors**, and those tokens now appear in warnings deliberately — so it scopes to errors, and a new test asserts the warnings **do** name all four. That is the guarantee replacing silence, not a weakening.

## Evidence

Mutation battery — 10 mutations, no survivors:

| | Result |
|---|---|
| N1 neuter the skip entirely | 8 failed |
| N2 remove the `extras > 0` gate | 1 failed |
| N3 drop the type-refutation signal | 2 failed |
| N4 drop the known-token signal | 4 failed |
| N5 BOOLEAN stops accepting 0/1 | 1 failed |
| N6 numeric strings stop being accepted | 2 failed |
| N7 drop the `missing_node_type` promotion | 1 failed |
| N8 validator stops converting UI→API | 3 failed |
| **N9 skip silently (no warning)** | **1 failed** |
| **N10 stop de-duplicating the promoted warning** | **1 failed** |

`tsc --noEmit` clean · `check:vocabulary` OK · `docs:gen` no-op · targeted suites 107 passed · full suite **591 files / 10963 tests green**.

Branched from current `main` (`33e97ab`), so this applies on top of the merged #1880.

### DIFF
--- a/src/__tests__/services/workflow-converter.test.ts
+++ b/src/__tests__/services/workflow-converter.test.ts
@@ -3611,6 +3611,28 @@ describe("convertUiToApi — serialized action buttons must not shift widget val
     // must NOT promote it out of its own slot.
     expect(workflow["4"].inputs.text).toBe("hello");
   });
+
+  it("preserves a known button spelling when the declared COMBO explicitly accepts it", () => {
+    const { workflow } = convertUiToApi(
+      ({
+        nodes: [{ id: 1, type: "ComboCollision", mode: 0, inputs: [], outputs: [], widgets_values: ["browse", "other", "42"] }],
+        links: [],
+      }) as never,
+      {
+        ComboCollision: {
+          input: {
+            required: {
+              mode: [["browse", "other"], {}],
+              note: ["STRING", {}],
+            },
+          },
+          output: [],
+        },
+      } as never,
+    );
+    expect(workflow["1"].inputs.mode).toBe("browse");
+    expect(workflow["1"].inputs.note).toBe("other");
+  });
 });
 
 // Independent review of the first cut broke it four ways. Every case below is a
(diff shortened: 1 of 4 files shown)


artokun:
Follow-up merge-gate blocker:

The new disclosure is correct for an ambiguous STRING/file-path value, but the skip is not irreducibly ambiguous for every spec. If the current object_info spec is a declared COMBO whose allowed values include "browse" or "detect_range", that candidate is a legitimate widget value and the converter can distinguish it from an unknown frontend button.

Please add a regression with a COMBO allowing one of the known tokens plus an extra row value, and guard the known-token skip when the declared spec explicitly accepts the candidate. Keep the warning-only behavior for ambiguous STRING values. Then rerun the focused suite and fresh CI.

Issue #349, opened 2026-07-29
Graph edit tools intermittently target a DIFFERENT open workflow tab — panel_set_node_title wrote to the wrong workflow (data corruption)


artokun:
### What happened / root cause

While renaming/repositioning nodes in one workflow tab, the panel graph-edit tools intermittently operated on a **different open tab**. This is not just a read desync — it **wrote to the wrong user workflow**.

Concrete evidence: I called

```
panel_set_node_title(node_id=5, title="유광 도장 지시 (CLIPTextEncode)")
```

intending tab A (`matte-to-glossy-vehicle-flux2-final`, whose node 5 is a `CLIPTextEncode`). The tool returned:

```json
{ "node_id": 5, "previous": "Qwen2.5-VL 텍스트 인코더 (CLIPLoader)", "title": "유광 도장 지시 (CLIPTextEncode)" }
```

`previous` is the title of node 5 in tab **B** (`matte-to-glossy-vehicle-qwen-edit`, a `CLIPLoader`). The write landed on tab B and mislabeled a node in a workflow the user had not asked me to touch. I only caught it because the returned `previous` value didn't match what `panel_graph_outline` had just shown me; I then had to switch tabs and restore the title by hand.

The same desync also surfaces as spurious not-found errors — during a ~35-call sequence of `panel_set_node_title` / `panel_move_node` against tab A (20 nodes, ids up to 25), I got three failures of the form:

```
Error: No node with id 18 in the current graph
Error: No node with id 24 in the current graph
Error: No node with id 8  in the current graph
```

for ids that unambiguously exist in tab A. Tab B tops out at node id 15, which is consistent with the call being routed to tab B. Re-issuing the identical call (sometimes after re-calling `panel_open_workflow` on tab A) then succeeded.

Throughout all of this:
- `panel_list_workflows` consistently reported tab A as active.
- `panel_graph_outline` consistently returned tab A's graph (20 nodes).
- Only the **mutating** tools drifted to tab B.

So the read path and the write path resolve the target workflow differently, and the write path can silently pick a stale/other tab.

Likely related, from the same session: a bogus `MANUAL CANVAS CHANGES` block that described tab A's graph as edits to tab B (filed separately as #348). Both point at one root cause — the panel's notion of "the current graph" is not consistently pinned per operation.

[...]

PR #1912, opened 2026-08-20
fix(panel#1529): a workflow pin may not claim a graph target it did not establish

### LINKED ISSUE #1529: Workflow pin reports success but graph tools stay on previous tab

artokun:
Fixes artokun/comfyui-mcp-panel#1529 — "Workflow pin reports success but graph tools stay on previous tab" (severity:P2, via-panel).

> **Corrected below.** The "no enumerable `workflows` array" rung described here reproduces the reported reply on clean `origin/main`, but it is a **pre-0.11.20 panel shape** and therefore NOT the reporter’s 0.13.6 build. The rung that reaches a current panel is a `workflow_list` probe that does not come back, which the report-honesty half of this fix covers and the hard-refusal half does not. See the correction comment for the full accounting — read it before this body.

**Review is PENDING.** This has NOT been through an independent review; a request for grok naming the risky parts is posted below. Do not read the evidence here as a verdict.

## The report

`panel_open_workflow` said workflow B was active (`active_matches_target:true`). `panel_set_workflow_target({mode:"pinned"})` then answered

> Pinned to "api_seedance2_5_video_extend". **Graph tools will target that workflow** without switching the user's view. NOTE — the workflow target was set, but this session's turn routing was NOT re-pinned: the existing pin (tmp:2fb3) still reaches a live tab of this conversation…

and the very next `panel_graph_outline` returned the *other* workflow's nodes. Two selectors — the workflow-target store and the turn-routing pin — disagreed, and the reply read as a success either way.

## Mechanism, established by execution (not by reading)

I stood up a harness with the **real** `UiBridge` on a real socket, the **real** `TurnOriginTracker` + `makeScopeTargetResolver` / `makeScopeRepinHandler` (wired exactly as `orchestrator/index.ts` wires them), the **real** `WorkflowTargetStore` and the **real** panel tool defs, driven against scripted panel sockets, and ran the reporter's three calls against **clean `origin/main`**. Recorded off the wire:

- **Healthy control** — one page, A active, `panel_open_workflow(B)` switches it and the panel re-helloes under B's route on the same socket. The turn pin follows through the migration alias, `graph_outline` is dispatched to that tab carrying `workflow_path=B`, and B's outline comes back. Correct, and it stays correct.
- **The reported state** — the turn pin routes to the tab showing A, and the panel answers `workflow_list` **without** an enumerable `workflows` array. On `origin/main` this printed, verbatim:

  ```
  "note": "Pinned to \"api_seedance2_5_video_extend\". Graph tools will target that workflow
   without switching the user's view. NOTE — … this session's turn routing was NOT re-pinned:
   the existing pin (tmp:2fb3) still reaches a live tab of this conversation …"
  ```

  …and the next `graph_outline` went to the tab showing A carrying `workflow_path=<B>`, which the panel's shipped pin guard (#349/#186) refuses — or, on a build whose guard does not fire for that pin form, silently answers with A's canvas. That is the reporter's sentence.

The enabling defect is in `resolveOpenWorkflow`. It returns `null` — *indeterminate*, which `resolvePinTarget` treats as **lenient, pin anyway** — the moment the reply carries no `workflows` array:

```ts
const rawList = (parsed as { workflows?: unknown }).workflows;
if (!Array.isArray(rawList) || rawList.length === 0) {
  return null;                       // ← the top-level `active` record, never read
}
const activeObj = (parsed as { active?: unknown }).active;
```

"Which tabs are open" really is unknown there. But the reply's **top-level `active` record was sitting right behind that early return**, and it answers the only question a pin turns on: the panel reads and edits the in-view workflow and nothing else, so if the target is not the live canvas the pin cannot make graph tools reach it. Discarding a readable *positive mismatch* because a *different* field was missing is what turned a refusable pin into a silent unverified one. (A well-formed list refuses correctly today — that control is in the suite.)

Then, one layer up, the routing claim was unconditional:

```ts
const hint = target.mode === "pinned"
  ? `Pinned to "…". Graph tools will target that workflow without switching the user's view.`
  : "Following the user's current workflow tab.";
```

so a pin that established nothing about the active canvas still announced that graph tools would reach it, in the same breath as saying the routing pin had been left alone.

## The change (2 files)

1. **`resolveOpenWorkflow`** — when the tab list is not enumerable, read the top-level `active` record through `readOpenActiveAgainstTarget` (the #887 oracle already used for the post-open drift notice) and refuse on `"different"` only. It is generous in the direction that matters: an unreadable active record, or one that plausibly *is* the requested workflow, stays `"indeterminate"` and keeps today's lenient behaviour, so an older panel that reports no usable active identity is untouched. The refusal is a new `{targetNotActive}` outcome with its own message, because this branch has no evidence the target is *open* and must not say "it is open but not active".
2. **`resolvePinTarget` → the tool** — the resolution now carries `verifiedActive`, and `panel_set_workflow_target` makes the routing claim only when the panel positively confirmed the target is the live canvas. Unverified pins are still **written** (leniency is deliberate) but the reply says so, says the pin travels as a GUARD, and names the one cheap read that settles it — plus a machine-readable `pin_verified_active`, so nobody has to parse prose to learn whether the graph target was established.

## Evidence

Three named tests in `src/__tests__/orchestrator/pin-vs-graph-target.test.ts`, all end-to-end over a real bridge and all asserting on the reply an agent actually receives. Each mutation below was applied and measured; each killed **exactly one** test:

| mutation | test that went red |
| --- | --- |
| restore the pre-fix `return null` (drop the active-record branch) | `REFUSES a pin the panel reports is not the live canvas, even with no enumerable tab list` — *expected undefined to be true* |
| **delete the CALL SITE line** `pinVerifiedActive = res.verifiedActive;` | `still makes the claim — and delivers it — when the pin IS verified active` |
| force the verdict to always-verified in `resolvePinTarget` | `does NOT claim the graph target for a pin it could not verify` |

The middle row is the point of running this end-to-end: the join between the verdict and the claim is a single assignment, and a helper-level test survives its deletion. These assert on the tool's own reply text and fields, so they do not.

Suite: `10961 passed`, `4 failed | 2 files` — `scripts/package-hooks.test.ts` (`files` allowlist, needs a built `dist/`) and `orchestrator/node-id-union-message.test.ts` (#1889). Both were measured **failing identically on a clean `origin/main` worktree** before blaming this branch. Typecheck clean (probe-verified: a deliberate type error was injected and caught, so the check is real).

- **No panel-repo change.** The panel behaved correctly throughout: it refuses a `workflow_path` that is not the active canvas. The false claim was entirely orchestrator-side, so `comfyui-mcp-panel` is untouched and its Playwright suite cannot exercise this path (its specs run against a MockBridge, which replaces the orchestrator). The suite was run anyway for the record; result reported in a comment.
- **Did not make sessions workflow- or tab-scoped.** Sessions are orchestrator-scoped by design; the disagreement is fixed by refusing to claim what was not established, never by re-scoping the session.
- **Did not displace the live turn-routing pin.** The #884 P0 stands: a pin that still reaches a live tab of this conversation is never taken away by a `mode:"pinned"` call, however explicit. This PR does not touch `makeScopeRepinHandler`.
- **Did not make every unverified pin an error.** Refusing on absence of evidence would newly fail legitimate pins on panels that report no active identity at all. The refusal fires only on a *positive* mismatch; the no-evidence case is disclosed, not failed. That line is the judgement call most worth a reviewer's attention.
- **Did not run `codex-gate.mjs`** and did not spawn codex.

### DIFF
--- a/src/__tests__/orchestrator/pin-vs-graph-target.test.ts
+++ b/src/__tests__/orchestrator/pin-vs-graph-target.test.ts
@@ -380,7 +380,11 @@ describe("panel#1529 — a pin may not claim a graph target it did not establish
     const outline = await tool("panel_graph_outline").handler({}, ctx);
     expect(outline.isError).toBeFalsy();
     expect(textOf(outline)).toContain("Wan2VideoEditApi");
-    expect(page.seen.filter((f) => f.cmd === "graph_outline")[0]?.workflow_path).toBeUndefined();
+    // Assert the frame EXISTS before asserting what it lacks — `[0]?.workflow_path`
+    // on an empty array is undefined too, and would pass this vacuously.
+    const outlineFrames = page.seen.filter((f) => f.cmd === "graph_outline");
+    expect(outlineFrames).toHaveLength(1);
+    expect(outlineFrames[0].workflow_path).toBeUndefined();
   });
 
   it("does NOT claim the graph target for a pin it could not verify", async () => {
--- a/src/__tests__/orchestrator/pin-vs-graph-target.test.ts
+++ b/src/__tests__/orchestrator/pin-vs-graph-target.test.ts
@@ -1,0 +1,433 @@
+// panel#1529 — "Workflow pin reports success but graph tools stay on previous tab".
+//
+// WHAT WAS REPORTED. `panel_set_workflow_target({mode:"pinned"})` answered
+// «Pinned to X. Graph tools will target that workflow» and, in the same reply,
+// «this session's turn routing was NOT re-pinned … the existing pin (tmp:2fb3)
+// still reaches a live tab». The very next `panel_graph_outline` came back with
+// the OTHER workflow's nodes. Two selectors — the workflow-target store and the
+// turn-routing pin — disagreed, and the reply read as a success either way.
+//
+// WHAT THIS FILE PROVES, and why it is an e2e over a REAL bridge rather than a
+// unit test on the helper. The claim lives in the TOOL's reply, the verdict is
+// computed in `resolvePinTarget`, and the two are joined by one assignment at the
+// call site. A helper-level test survives that assignment being deleted — it was
+// deleted here, deliberately, and only the reply-level assertions below went red.
+// So every assertion reads the string/field an agent actually receives, produced
+// by the real tool def, the real workflow-target store, the real UiBridge and the
+// real turn-origin pin, against scripted panel sockets.
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { createServer } from "node:net";
+import WebSocket from "ws";
+import { UiBridge } from "../../services/ui-bridge.js";
+import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+import {
+  TurnOriginTracker,
+  makeScopeRepinHandler,
+  makeScopeTargetResolver,
+} from "../../orchestrator/turn-origins.js";
+import {
+  buildPanelToolDefs,
+  makePanelToolCtx,
+  type PanelToolCtx,
+  type ToolResult,
+} from "../../orchestrator/panel-tools.js";
+import { SHARED_SESSION_SCOPE, isScopeAddress } from "../../services/session-scope.js";
+
+const textOf = (res: ToolResult): string =>
+  res.content.map((c) => ("text" in c ? (c as { text: string }).text : "")).join("\n");
+
+const jsonOf = (res: ToolResult): Record<string, unknown> => {
+  try {
+    return JSON.parse(textOf(res)) as Record<string, unknown>;
+  } catch {
+    return {};
+  }
+};
+
+/** The exact sentence that asserts the graph target moved. */
+const GRAPH_TARGET_CLAIM = "Graph tools will target that workflow";
+
+function freePort(): Promise<number> {
+  return new Promise((resolve, reject) => {
+    const srv = createServer();
+    srv.once("error", reject);
+    srv.listen(0, "127.0.0.1", () => {
+      const addr = srv.address();
+      const p = addr && typeof addr === "object" ? addr.port : 0;
+      srv.close((err) => (err ? reject(err) : resolve(p)));
+    });
+  });
+}
+
+/** Same shape as ui-bridge.test.ts's harness: retry past the close/rebind race so
+ *  a bind flake is never reported as a feature failure. */
+async function startBridgeOnFreePort(): Promise<{ bridge: UiBridge; port: number }> {
+  let lastErr = "no attempt made";
+  for (let attempt = 0; attempt < 6; attempt++) {
+    const p = await freePort();
+    const b = new UiBridge(p);
+    b.start();
+    if (await b.whenReady()) return { bridge: b, port: p };
+    lastErr = `bind to ${p} lost a close/rebind race`;
+    await b.stop();
+  }
+  throw new Error(`could not bind a free bridge port after 6 attempts: ${lastErr}`);
+}
+
+const SCOPE = `${SHARED_SESSION_SCOPE}::claude`;
+
+interface Wf {
+  path: string;
+  filename: string;
+  key: string;
+  routing_key: string;
+  uuid: string;
+  outline: string;
+}
+
+/** The workflow the turn was issued from — an UNSAVED tab, as in the report. */
+const A: Wf = {
+  path: "workflows/wan_video_edit.json",
+  filename: "wan_video_edit",
+  key: "wan_video_edit.json",
+  routing_key: "wf:workflows/wan_video_edit.json",
+  uuid: "aaaaaaaa-1111-4111-8111-111111111111",
+  outline: '1 Wan2VideoEditApi "edit"',
+};
+/** The workflow the agent asks to pin. */
+const B: Wf = {
+  path: "workflows/api_seedance2_5_video_extend.json",
+  filename: "api_seedance2_5_video_extend",
+  key: "api_seedance2_5_video_extend.json",
+  routing_key: "wf:workflows/api_seedance2_5_video_extend.json",
+  uuid: "bbbbbbbb-2222-4222-8222-222222222222",
+  outline: '1 SeedanceVideoExtendApi "extend"',
+};
+
+/** How this panel answers `workflow_list`. */
+type ListShape =
+  | "full" // current panel: enumerable tabs + a top-level active record
+  | "active-only" // no `workflows` array, but the active canvas IS reported
+  | "opaque"; // neither — nothing about the live canvas can be established
+
+let bridge: UiBridge;
+let port: number;
+const sockets: WebSocket[] = [];
+
+/** A scripted panel page. Records every command frame it receives so the test can
+ *  assert on what was ROUTED, not only on what the tool said. */
+class FakePanel {
+  sock!: WebSocket;
+  readonly seen: Array<{ cmd: string; workflow_path?: string }> = [];
+
+  constructor(
+    public tabId: string,
+    public open: Wf[],
+    public activeIdx: number,
+    public listShape: ListShape = "full",
+  ) {}
+
+  get active(): Wf {
+    return this.open[this.activeIdx];
+  }
+
+  async connect(): Promise<void> {
+    this.sock = await new Promise<WebSocket>((resolve, reject) => {
+      const s = new WebSocket(`ws://127.0.0.1:${port}`);
+      sockets.push(s);
+      s.on("open", () => resolve(s));
+      s.on("error", reject);
+    });
+    this.hello(this.tabId);
+    this.sock.on("message", (buf) => this.onMessage(buf.toString()));
+    await settle();
+  }
+
+  hello(tabId: string): void {
+    this.tabId = tabId;
+    this.sock.send(
+      JSON.stringify({
+        type: "hello",
+        tab_id: tabId,
+        title: this.active.filename,
+        backend: "claude",
+        workflow_uuid: this.active.uuid,
+        enforces_workflow_stamp: true,
+        enforces_workflow_stamp_at_write: true,
+      }),
+    );
+  }
+
+  private record(w: Wf, i: number): Record<string, unknown> {
+    return {
+      path: w.path,
+      filename: w.filename,
+      title: w.filename,
+      key: w.key,
+      routing_key: w.routing_key,
+      active: i === this.activeIdx,
+      modified: false,
+      persisted: true,
+    };
+  }
+
+  private onMessage(raw: string): void {
+    const msg = JSON.parse(raw) as Record<string, unknown>;
+    const cmd = msg.cmd as string | undefined;
+    const rid = msg.rid as string | undefined;
+    if (!cmd || !rid) return;
+    this.seen.push({
+      cmd,
+      workflow_path: typeof msg.workflow_path === "string" ? msg.workflow_path : undefined,
+    });
+    const reply = (result: unknown): void =>
+      this.sock.send(JSON.stringify({ rid, ok: true, result, workflow_uuid: this.active.uuid }));
+    const refuse = (error: string): void =>
+      this.sock.send(JSON.stringify({ rid, ok: false, error }));
+
+    // The panel's shipped PINNED-TARGET GUARD (#349/#186): a `workflow_path` that
+    // is not the active canvas is refused rather than executed. Modelled here so
+    // the cost of a wrong pin shows up in the test the way it shows up for a user.
+    const pinned = msg.workflow_path;
+    const targetless = cmd === "workflow_list" || cmd === "workflow_open" || cmd === "workflow_new";
+    if (typeof pinned === "string" && pinned.trim() && !targetless) {
+      const forms = [this.active.path, this.active.key, this.active.filename, this.active.routing_key];
+      if (!forms.includes(pinned)) {
+        refuse(
+          `workflow mismatch: your session is pinned to "${pinned}", but that is not the active ` +
+            `canvas ("${this.active.path}").`,
+        );
+        return;
+      }
+    }
+
+    switch (cmd) {
+      case "workflow_list": {
+        if (this.listShape === "opaque") {
+          reply({});
+          return;
+        }
+        const activeObj = {
+          path: this.active.path,
+          filename: this.active.filename,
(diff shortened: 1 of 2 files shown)


artokun:
@grok please review this one — it is a report-honesty fix on a routing path, so the risk is concentrated in a few specific places rather than spread across the diff. Naming them so you can go straight at them:

**1. The load-bearing claim: that a POSITIVE mismatch is safe to refuse on, and an absent one is not.**
`resolveOpenWorkflow` now reads the reply's top-level `active` record when the `workflows` array is missing, and refuses the pin only when `readOpenActiveAgainstTarget(...) === "different"`. My whole argument rests on that oracle never returning `"different"` for a record that IS the requested workflow spelled differently — it defers to `plausiblySameSavedWorkflow` for that, and I reused it rather than writing a new comparison precisely because it is the generous direction. **Please attack the case I am least sure of: an UNSAVED active tab.** A `tmp:` routing key is deliberately not a saved identity, and the oracle's own doc says an unsaved tab is a positive identification of "different". If a caller pins an unsaved workflow by some token while another unsaved tab is active — or pins a saved workflow while an unsaved tab is active and the panel's `active` record carries a `filename` like "Unsaved Workflow" — does my branch turn a pin that works today into a refusal? The refusal is a NEW hard failure on a path that previously always succeeded, so a false positive here is a regression, not a nit.

**2. `active_confirmed` is passed through, but I did not verify every panel build sets it.** I forward `parsed.active_confirmed` into the oracle so `false` degrades to indeterminate. If a build omits it, that is fine (undefined ≠ false). If some build sets it `false` routinely, my refusal never fires and the fix is inert on exactly the older panels it targets — worth a second pair of eyes on whether this is reachable in the field or only in my harness.

**3. The judgement call I could argue either way.** An unverified pin is still WRITTEN and reported as a non-error, with a "NOT CONFIRMED" note and `pin_verified_active:false`. The alternative reading of this issue — "refuse rather than silently pick one" — would make that combination an error when the scope repin was also refused. I chose disclosure over failure because refusing on ABSENCE of evidence would newly fail legitimate pins on panels that report no active identity at all. If you think the ticket wants the harder line, say so; it is a one-branch change.

**4. Ordering / P0 I believe I did not disturb, please confirm.** The workflow-target store write and the `bridge.push` are unchanged, and my new refusal happens inside `resolvePinTarget`, i.e. BEFORE the store write — so nothing is half-applied. I also did not touch `makeScopeRepinHandler`, so the #884 P0 (a pin still reaching a live tab of this conversation is never displaced) is untouched. Both claims are worth checking against the diff rather than against this sentence.

**5. The evidence, so you can judge whether it proves what I say it does.** Three end-to-end tests over a real `UiBridge`, real turn-origin pin and real tool defs. I mutated the fix three ways and each killed exactly one named test — including deleting the one-line CALL-SITE assignment `pinVerifiedActive = res.verifiedActive;`, which a helper-level test would have survived. What I have NOT done is reproduce the reporter's exact panel build; my harness reproduces the reported REPLY verbatim on clean origin/main, which is a strong but not identical thing. If you think the reporter's state has a different enabling condition than the missing `workflows` array, that is the finding I most want.

No codex gate was run on this (deliberately — it is pointed at a weak model right now), so your read is the only independent one.

artokun:
Follow-up from an adversarial re-read of my own diff (not a review — the request above still stands):

- **Named a cause the evidence does not support.** The NOT CONFIRMED note said "this panel did not report which workflow is the active canvas". `verifiedActive:false` also covers a `workflow_list` that never answered and one whose records share no comparable identity with the active canvas — so that sentence picked one of three causes and printed it as fact. It now says only what is true in all three: the check did not happen.

- **Pinned the coverage boundary with a test, and it is a real boundary.** `canonicalRequestedSavedIdentity` requires a `/` in the token, so a pin issued by bare **key or filename** — which is what `panel_list_workflows` hands an agent, and what the reporter's own `Pinned to api_seedance2_5_video_extend` looks like — does NOT reach the new refusal. It stays lenient, and the second half of the fix (no routing claim, `pin_verified_active:false`) is what protects the caller there. I chose not to widen it: dropping that early-out would make a `tmp:` routing token read as "different" from any named active tab and start refusing valid pins to **unsaved** workflows, which is a worse failure than the one being fixed. The new test `a BARE KEY token is not refused on this evidence — the boundary, on purpose` fixes that division in place so neither half can be removed silently.

  If you think the hard refusal should cover key/filename tokens too, that is the change I would most want challenged — it needs a pin-specific comparison that is generous about basename collisions AND about `tmp:` selectors, and I did not convince myself I could write one safely.

- Checked and found clean: the second `ok({… graph_binding:"bound" …})` early return does not need `pin_verified_active`, because `fence` is only ever set on the `mode:"current"` branch (`fenceRebind` is assigned at one site, inside `if (mode === "current" …)`), so that return is unreachable for a pin.

Two commits now; `git diff origin/main...HEAD` is still just `src/orchestrator/panel-tools.ts` and the new test file.

artokun:
**Correction to my own PR body — read this before the evidence above.**

I said the enabling condition is a `workflow_list` reply with no enumerable `workflows` array. That reproduces the reported reply verbatim on clean `origin/main`, and it is a real rung — but **it cannot be the reporter's build.** The panel has emitted `workflows` alongside `open` since **0.11.20** (`edc88f7e`, #186/#386), and the panel's own comment at that line says exactly why:

> Emit the SAME list under both `open` and `workflows`: the orchestrator's pin-canonicalizer (resolveOpenWorkflow, #512) enumerates `parsed.workflows` … with only `open` present it saw no list and fell back to the raw pin unverified.

The reporter is on panel **0.13.6**, which is well past that. So the shape I reproduced first is a *pre-0.11.20* shape.

The rung that reaches a current panel is the other one: `resolveOpenWorkflow` also returns indeterminate — and therefore lenient — whenever the **probe itself does not come back** (a transient reconnect, a canvas past its 6 s budget), or when its records expose no comparable active identity. On that path there is no evidence to refuse on, and inventing one would fail healthy pins, so the pin is still written and **the reply is the entire fix**: no routing claim, `pin_verified_active:false`, and the cheap read that settles it.

I have added `makes no claim when the verifying probe itself FAILED` — a panel that fails `workflow_list` outright — so that route is covered by a named test rather than by my assertion that it exists. Forcing the lenient rung to report verified kills it along with two others.

What this means for how to weigh the two halves:

- **The report-honesty half covers every route**, including the reporter's. That is the one I would not ship without.
- **The hard refusal half covers the pre-0.11.20 rung and an empty tab list.** Narrower than my PR body implied. Still worth having — an unverified pin we CAN prove is wrong should not be written — but it is not what fixes the reporter's build.

What I still have NOT established: the reporter's exact enabling rung. I reproduced the reply and the wrong-canvas read; I did not reproduce their panel build. If you can see a path to a false "Graph tools will target that workflow" on a 0.13.6-shaped `workflow_list` that I have not covered — particularly one where the ROUTING moves between the pin's verification probe and the next graph command, rather than the pin being unverified — that is the finding that matters most, because my fix would not catch it.

(Reposted: the first version of this comment lost its code spans to shell backtick expansion.)

artokun:
**Panel browser suite — run for the record, and it says nothing about this PR.**

`cd comfyui-mcp-panel && npx playwright test --workers=1` → **18 failed, 60 passed (10.7m)**.

Read that number carefully rather than as a signal:

- **This PR changes no file in `comfyui-mcp-panel`.** `git diff origin/main...HEAD` is `src/orchestrator/panel-tools.ts` plus one new test, both in `comfyui-mcp`. The Playwright suite runs the panel checkout's own working tree, so nothing in it could have been affected by this branch. My grounds for non-attribution are structural, not empirical — I did **not** baseline these 18 against a clean panel `origin/main`, and I am not claiming they are pre-existing on main.
- **The panel checkout is not on main.** It sits on `fix/1280-video-card-poster-and-two-stage-visibility` with an untracked spec, i.e. another agent's working state. The suite was run against that.
- **The suite cannot exercise this change even in principle.** Every Tier-1 spec points the panel at the `MockBridge`, which *replaces* the orchestrator — so `resolvePinTarget`, the workflow-target store and the turn-routing pin are all mocked away in that harness. The path this PR touches has no browser-visible surface: the changed output is a tool reply an agent reads, not anything the sidebar renders.

Failures, for whoever is looking at the panel suite for its own sake: 3 in `chat-history-v2`, 7 in `conversation-persistence`, 4 in `workflow-chat-identity`, plus `external-orchestrator`, `hidden-tab`, `pending-queue` and `streaming-render` — a durable-transcript / streaming cluster, unrelated to workflow pinning.

The run left the panel checkout clean (no new modified files, litter cleanup ran).

artokun:
Codex review: no blocking finding in the current diff.

Checked the production caller and resolver:
- `resolveOpenWorkflow` only returns the new refusal on a positive active-canvas mismatch; transport failure, missing/opaque lists, uncomparable identities, and bare-key aliases remain lenient.
- `panel_set_workflow_target` now suppresses the graph-target claim unless the target was positively confirmed active, and exposes `pin_verified_active`.
- the panel-driven `set_workflow_target` event still calls the same `resolvePinTarget` validator before writing.
- the existing `NOT_OPEN` and background-tab fail-closed branches remain ordered before the new object branch.

Local verification: targeted `pin-vs-graph-target.test.ts` 5/5 and `tsc --noEmit` pass. Holding merge until fresh CI is green.

artokun:
Codex merge gate PASS for updated head `f63514c26aa2a75dd2ced4710973cd3d1af0342f`.

Final verification:
- production diff reviewed at `src/orchestrator/panel-tools.ts`; the final commit is test-only and closes a vacuous graph-frame assertion;
- targeted e2e: 5/5;
- `tsc --noEmit`: pass;
- fresh CI run `32417619857`: Ubuntu, Windows, macOS, and pack/install smoke all pass.

Merge commit; do not squash. After merge, confirm and close cross-repository panel issue #1529.

artokun:
Correction: the exact updated head is `f63514c26aa2a75dd2ced4710973cd466868ce8f` (the gate evidence and CI run are unchanged).

artokun:
**Post-merge verification, and a straight account of what did and did not review this.**

This merged at `21:12:09Z` on a codex gate PASS. Two things about that are worth writing down rather than leaving implied:

- The **grok review requested on this PR at `20:55:03Z` never returned.** It was still outstanding at merge time. Nothing here was reviewed by it.
- Under the current standing instruction, a codex verdict is **not** the independent review path for this repo right now — codex is pointed at a weaker model, so a PASS from it reads with more authority than it carries. I am not counting it as the review that happened.

So this landed on author-side evidence. That makes the evidence carry more weight, not less, and the strongest piece of it had not been run at merge time. It has now been run **against the merged commit `2c859ee`**, not against the branch:

### Mutation matrix — 3 mutations, each isolated, each restored

| mutation | what it breaks | result |
|---|---|---|
| `resolveOpenWorkflow`: positive-mismatch branch made unreachable | the **refusal** half | 1 failed / 4 passed — `REFUSES a pin the panel reports is not the live canvas…` |
| `resolvePinTarget`: lenient return reports `verifiedActive: true` | the **honest reply** half, in the helper | 3 failed / 2 passed |
| tool def: `pinVerifiedActive = true` instead of `res.verifiedActive` | the **call site**, helper left honest | 3 failed / 2 passed — same three |

Baseline 5/5 before, 5/5 after restore. The two "no unearned claim" mutations fail exactly `does NOT claim the graph target for a pin it could not verify`, `makes no claim when the verifying probe itself FAILED`, and `a BARE KEY token is not refused on this evidence`.

The third row is the one that mattered to me: the helper and the call site fail **independently**, so the one-line wiring is pinned on its own. A helper-level test alone would have passed with the tool ignoring the verdict entirely.

`still makes the claim — and delivers it — when the pin IS verified active` stays green under all three, which is the point — it is the guard against over-correcting into a pin that never claims anything.

Also confirmed rather than taken on report: CI run `32417619857` concluded **success on all four jobs** for the exact merged head `f63514c2`.

### Still worth attacking, if a reviewer gets here

The load-bearing claim is unchanged and unreviewed: **that a POSITIVE active-canvas mismatch is safe to refuse on.** The refusal is a *new hard failure* on a path that previously always succeeded, so a false positive is a regression rather than a nit. The case I am least sure of is still an **unsaved active tab** — `readOpenActiveAgainstTarget` treats an identified-but-not-plausibly-same record as `"different"`, and I am relying on `plausiblySameSavedWorkflow` never being wrong in the strict direction. I checked the boundary myself (a bare key/filename token cannot reach the refusal, because `canonicalRequestedSavedIdentity` requires a `/`) and the second production call site (`index.ts:4674`, the panel-driven event path — it acks only and makes no routing claim, so it cannot emit the false sentence). That is author-side checking, not review.

### Follow-up filed

The reporting thread carried a **second** reproduction — reads resolving to the right workflow while `panel_remove_node` went through the older turn pin and was refused by the instance fence. This PR does not address it, and closing panel#1529 would have buried it. Split out with mechanism and a code pointer at `turn-origins.ts:694-710`: **#1917**.

PR #1919, opened 2026-08-20
fix(1917): mode:"current" may not claim a routing target the turn pin holds elsewhere

### LINKED ISSUE #1917: A live turn-routing pin is never displaced by a workflow target, so graph reads and graph MUTATIONS split across two tabs
Split out of the panel thread https://github.com/artokun/comfyui-mcp-panel/issues/1529, which has now been closed. That issue shipped the *honesty* half (artokun/comfyui-mcp#1912, merged `2c859ee`): a pin can no longer claim "Graph tools will target that workflow" it did not establish. This is the *routing* half it deliberately did not take — and the *second* reproduction on that thread is covered only here, so closing #1529 would otherwise drop it.

### The reproduction (reporter's own words, panel#1529 comment 2)

> using `panel_set_workflow_target({mode:"current"})` after `panel_open_workflow` bound the saved PHOTO copy and `panel_graph_outline` read that copy correctly, but `panel_remove_node({node_id:99})` was routed through the older Basic_V37 turn pin and refused by the instance fence. The prior reply explicitly retained the healthy older turn-routing pin.

The distinguishing detail: **reads landed on the right workflow and a mutation did not.** That is not the #1529 symptom (a pin that reports success and then reads the wrong canvas) — it is the two selectors having genuinely split, with each one internally consistent.

### Mechanism

There are two independent selectors and only one of them follows a workflow pin:

- **the workflow-target store** — what `panel_set_workflow_target` writes, and what graph *reads* resolve through;
- **the turn-routing pin** (`turn-origins.ts`) — which *tab* a scope-addressed command is dispatched to, and therefore which canvas a *mutation* reaches and which instance stamp the fence checks it against.

`makeScopeRepin` will only ever displace the second one as a **recovery**, and says so — `src/orchestrator/turn-origins.ts:694-710`:

```ts
// RECOVERY ONLY: a pin that still reaches a live tab OF THIS conversation
// is healthy — never displace it ...
const existing = opts.tracker.resolvedPinOf(key);
if (typeof existing === "string" && opts.bridge.canReach(existing)) {
  return { repinned: false, reason: `the existing pin (...) still reaches a live tab of this
    conversation, so it is healthy and was left alone — this recovery only displaces a
    pin that is dead or ambiguous` };
}
```

"Healthy" here means *reachable*, not *correct*. A pin onto the tab showing Basic_V37 is reachable, so it is kept — even though the user has just named a different workflow as the target. The `#888` comment immediately below establishes the intent that naming a workflow should make the pin follow; that only applies when the existing pin is dead or ambiguous.

Read from the source, not reproduced by execution — I did not build a two-tab repro for this one.

### Why the merged #1529 fix does not cover it

`2c859ee` changes two things, and neither reaches this path:

1. the pin-time refusal fires only for `mode:"pinned"` on a **positive** active-canvas mismatch; the reproduction above uses `mode:"current"`;
2. the suppressed graph-target claim is only attached to the `mode:"pinned"` hint. `mode:"current"` still replies "Following the user's current workflow tab." and says nothing about a turn-routing pin that names a different tab — so a caller in this state is still not told the two selectors disagree.

### What a fix looks like

Option 2 from the reporter's own "Precise change needed" list: let an explicit, successful workflow target **displace a live but different turn-routing pin**, rather than only a dead or ambiguous one. The user naming a workflow is a stronger signal than the old pin's reachability.

The risk to weigh, and why this deserves its own issue rather than a quick follow-up commit: `makeScopeRepin`'s current refusal is load-bearing for the recovery path it was written for, and widening "displace" from *dead-or-ambiguous* to *dead-or-ambiguous-or-different* changes which tab an in-flight conversation addresses. A cheaper first step, if displacement proves too invasive, is to make the disagreement **visible** — the `mode:"current"` reply already knows the turn pin was retained, and could say that graph mutations will be dispatched to that other tab.

### Environment (reporter)

Windows 11, ComfyUI 0.33.2, frontend 1.49.6, RTX 5070 Laptop 8 GB; comfyui-mcp 0.52.30; panel 0.15.12.

artokun:
Closes #1917.

**Review is PENDING.** grok's review is requested in a comment below; no verdict exists yet and nothing here is described as gated.

## What the report claimed, and what execution found

#1917 states there are two selectors, and that *reads* follow the workflow-target store while *mutations* follow the turn-routing pin. **That is not true, and I could not reproduce a read/mutation routing split.** Measured over a real `UiBridge` on a real socket, with the real `TurnOriginTracker` + `makeScopeTargetResolver`/`makeScopeRepinHandler` wired exactly as `orchestrator/index.ts` wires them, the real `WorkflowTargetStore`, the real panel tool defs and scripted panel sockets:

- `graph_outline` and `graph_remove_node` are both resolved by `UiBridge.resolveTarget(<scope>)` through the **same** pin;
- both carry the **same** `workflow_uuid`, from the same `commandStampAddress` resolution (`requiresStampTargetAgreement` is true for every `graph_*`, so both take the caller address);
- the panel fences them **identically** — `activeWorkflowFenceApplies` covers every `graph_*` (reads included), and `commandWorkflowMismatch` counts a *missing* stamp as a mismatch (#718);
- `withWorkflowTarget` injects `workflow_path` into reads and writes alike.

I swept the states that could plausibly split them: a retired-route-id pin resolving through the migration alias; a stale conversation stamp; an absent conversation stamp; a `wf:<route>:<path>` id taken over by another connection; the workflow-target store in `pinned` mode. In every one, the read and the mutation landed on the same tab and got the same verdict. The only read/mutation asymmetry that exists on our side is the mutation-only `requiresWorkflowStampEnforcement` gate, and it fires on panel *capability*, not on the turn pin, so it cannot produce a refusal naming the older workflow.

## What IS wrong, reproduced verbatim on clean main

`panel_set_workflow_target({mode:"current"})` answered **"Following the user's current workflow tab."** unconditionally. That is a claim about ROUTING, and it is false whenever this conversation's in-flight turn pin is holding every graph command on a different tab. It is the exact arm panel#1529's fix (#1912) left open: that change made only the `mode:"pinned"` claim conditional.

Worse, nothing in the reply mentioned the pin at all. For a scope-bound ctx whose pin still reaches a live tab, `rebindToActiveTab` short-circuits **before** consulting the repin handler (#884 — a live pin is never displaced) and returns `{rebound:false}` with no `repinRefusal`, so even #1077's "the session pin was NOT moved" note never fires on this path.

Measured on clean `origin/main`, two connected tabs, pin on the tab showing `Basic_V37`, active-scope resolution naming the tab showing `PHOTO`:

```
resolveActiveScopeTab = wf:rB:workflows/PHOTO.json
scope resolves to     = wf:rA:workflows/Basic_V37.json

panel_set_workflow_target({mode:"current"}) ->
  { "mode":"current", "graph_binding":"bound", "graph_binding_status":"already_current",
    "note":"Following the user's current workflow tab. ..." }

panel_graph_outline  -> BasicNode        (dispatched to wf:rA, Basic_V37)
panel_remove_node    -> ok, on Basic_V37 (dispatched to wf:rA, Basic_V37)
```

The agent is told it is following the user's current tab, and every graph command — read and mutation — goes somewhere else, unannounced. That is the same "reports success, the effect never happened" shape as #1529, and it is the honest explanation available for the reporter's second reproduction.

## The change

**#884 is untouched.** `turn-origins.ts` is not modified at all, `makeScopeRepinHandler` is not modified at all, and no pin is displaced to make any sentence true. A named test asserts the pin is exactly where it was after the call.

Instead, when the turn pin and the active-scope resolution name **different tabs**, `mode:"current"` now:

1. stops printing "Following the user's current workflow tab.";
2. names **both** tabs, states that the pin was not displaced and why (#884), says explicitly that this covers **reads and mutations alike**, and says what does and does not move the pin (retrying the tool does not; the pin is released at turn end and the next message's origin establishes routing);
3. carries it machine-readably — `turn_routing:"pinned_elsewhere"`, `turn_routing_tab`, `current_would_route_to` — the same rule #1912's `pin_verified_active` established, so no agent has to read a routing verdict out of prose.

This is option 3 of the reporter's own list (a refusal/disclosure instead of a false claim), and the "cheaper first step" #1917's own body names. It deliberately does **not** take options 1 or 2 (synchronize the route, or let a workflow target displace a live pin) — see below.

The comparison is between **canonical connection ids** on both sides (`resolveSharedTabId` and `resolveActiveScopeTab` both answer with `conn.tabId`). A pin naming a retired per-workflow route id (#640) that still resolves onto the same live socket is therefore **not** a disagreement — otherwise this warning would fire on every healthy single-tab session after any `panel_open_workflow`.

## Verification

`src/__tests__/orchestrator/turn-pin-vs-current-target.test.ts` — 6 tests, e2e over a real bridge on a real socket, every assertion reading the string or field an agent actually receives from the real tool def.

**Mutation-verified — four independent deletions, each turning exactly the named tests red:**

| mutation | red |
|---|---|
| revert the `hint` branch to the unconditional claim | `names BOTH tabs instead of claiming it follows the current one` |
| delete the `...turnRoutingSplitFields` spread at **both** `ok()` call sites | `carries the split MACHINE-READABLY, so no caller has to parse the prose` |
| neuter the canonical-id comparison (`routedTab !== activeScopeTab` to `true`) | `says nothing when the two selectors AGREE`, `a RETIRED route id that still reaches the same socket is not a split` |
| make the `scopeRoutedTabId` helper always return `undefined` | the first two tests |

Restore gives 6/6. The helper mutation and the call-site mutation turn *different* tests red, so the one-line wiring is pinned rather than only the helper.

**Suite:** full `vitest run` on this branch, run alone — **595 files / 10990 passed / 4 skipped / 0 failed**. A baseline on a clean `origin/main` worktree reported 3 files / 5 tests failing, all 30-second timeouts, and all of them pass here: they were machine-load artifacts from two suites running concurrently, not pre-existing breakage. `tsc --noEmit` clean. `npm run check:vocabulary` OK.

**Browser suite: not run, and not implied.** This change is orchestrator-side only — one tool reply string and three reply fields. Nothing in `comfyui-mcp-panel` is touched and nothing the panel renders changes.

## Deliberately not done

- **The pin is not displaced.** Widening `makeScopeRepinHandler` from *dead-or-ambiguous* to *dead-or-ambiguous-or-different* would re-aim an in-flight conversation's mutations at a tab it never addressed — #884's P0, and the reason the #1529 work left that handler alone.
- **Sessions are not made workflow- or tab-scoped.** They are orchestrator-scoped by design.
- **The `mode:"pinned"` arm is unchanged.** #888's `scopeRepinNote` already discloses a refused repin there, and #1912 already conditioned its claim.
- **No `fail()` on this path.** The result stays `ok`: the target write and the fence rebind did happen, and failing here would break the documented recovery for ordinary two-tab users on a predicate that can legitimately differ. The claim is corrected; the outcome is not manufactured.
- **`panel_open_workflow` not refreshing the conversation stamp.** The sweep showed the scope stamp still naming the previous workflow immediately after a successful `panel_open_workflow` plus same-socket re-hello (the following `mode:"current"` is what refreshes it). That is a separate question, and it is left out rather than folded in silently.

### DIFF
--- a/src/__tests__/orchestrator/turn-pin-vs-current-target.test.ts
+++ b/src/__tests__/orchestrator/turn-pin-vs-current-target.test.ts
@@ -1,0 +1,500 @@
+// mcp#1917 — "A live turn-routing pin is never displaced by a workflow target, so
+// graph reads and graph MUTATIONS split across two tabs".
+//
+// WHAT WAS REPORTED (panel#1529 comment 2, split out as #1917): after
+// `panel_open_workflow` bound a saved PHOTO copy, `panel_set_workflow_target({mode:
+// "current"})` answered as a success, `panel_graph_outline` read that copy, and
+// `panel_remove_node({node_id:99})` came back refused against the older Basic_V37
+// workflow — "the prior reply explicitly retained the healthy older turn-routing pin".
+//
+// WHAT WAS MEASURED, and what it changes. Driven over a real UiBridge on a real
+// socket with the real turn-origin pin and scope resolver, reads and mutations are
+// NOT two selectors: `graph_outline` and `graph_remove_node` are both dispatched by
+// `UiBridge.resolveTarget(<scope>)` through the same pin, both carry the same
+// `workflow_uuid` from the same `commandStampAddress` resolution, and the panel
+// fences them identically (`activeWorkflowFenceApplies` covers every `graph_*`, and
+// a missing stamp counts as a mismatch, #718). No state was found in which the two
+// route to different tabs — the last test here pins that, by asserting BOTH land on
+// the same tab in the very state the report describes.
+//
+// What IS true, and is the whole of this fix: `mode:"current"` claimed
+// «Following the user's current workflow tab.» while a live turn pin was holding
+// every graph command on a DIFFERENT tab, and said nothing about it. For a
+// scope-bound ctx whose pin still reaches a live tab, `rebindToActiveTab`
+// short-circuits (#884 — a live pin is never displaced) and returns
+// `{rebound:false}` with no `repinRefusal`, so not one word of the reply mentioned
+// the pin. Measured on clean main: pin on the tab showing Basic_V37, active-scope
+// resolution naming the tab showing PHOTO, reply «Following the user's current
+// workflow tab.» with `graph_binding:"bound"`, and both the following
+// `panel_graph_outline` and `panel_remove_node` dispatched to Basic_V37.
+//
+// The pin is NOT moved to make that sentence true. #884 is a P0 invariant and
+// `makeScopeRepinHandler` is untouched by this change; the third test asserts the
+// pin is exactly where it was after the call. What changes is the CLAIM.
+//
+// This is an e2e over a real bridge for the same reason the panel#1529 file is: the
+// claim lives in the TOOL's reply and the verdict is joined to it by assignments at
+// the call site, which a helper-level test survives. Every assertion below reads the
+// string or field an agent actually receives.
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import { createServer } from "node:net";
+import WebSocket from "ws";
+import { UiBridge } from "../../services/ui-bridge.js";
+import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+import {
+  TurnOriginTracker,
+  makeScopeRepinHandler,
+  makeScopeTargetResolver,
+} from "../../orchestrator/turn-origins.js";
+import {
+  buildPanelToolDefs,
+  makePanelToolCtx,
+  type PanelToolCtx,
+  type ToolResult,
+} from "../../orchestrator/panel-tools.js";
+import { SHARED_SESSION_SCOPE, isScopeAddress } from "../../services/session-scope.js";
+
+const textOf = (res: ToolResult): string =>
+  res.content.map((c) => ("text" in c ? (c as { text: string }).text : "")).join("\n");
+
+const jsonOf = (res: ToolResult): Record<string, unknown> => {
+  try {
+    return JSON.parse(textOf(res)) as Record<string, unknown>;
+  } catch {
+    return {};
+  }
+};
+
+/** The unconditional sentence this fix stops printing when it is not true. */
+const CURRENT_CLAIM = "Following the user's current workflow tab.";
+
+function freePort(): Promise<number> {
+  return new Promise((resolve, reject) => {
+    const srv = createServer();
+    srv.once("error", reject);
+    srv.listen(0, "127.0.0.1", () => {
+      const addr = srv.address();
+      const p = addr && typeof addr === "object" ? addr.port : 0;
+      srv.close((err) => (err ? reject(err) : resolve(p)));
+    });
+  });
+}
+
+/** Same shape as the other bridge e2e harnesses: retry past the close/rebind race
+ *  so a bind flake is never reported as a feature failure. */
+async function startBridgeOnFreePort(): Promise<{ bridge: UiBridge; port: number }> {
+  let lastErr = "no attempt made";
+  for (let attempt = 0; attempt < 6; attempt++) {
+    const p = await freePort();
+    const b = new UiBridge(p);
+    b.start();
+    if (await b.whenReady()) return { bridge: b, port: p };
+    lastErr = `bind to ${p} lost a close/rebind race`;
+    await b.stop();
+  }
+  throw new Error(`could not bind a free bridge port after 6 attempts: ${lastErr}`);
+}
+
+const SCOPE = `${SHARED_SESSION_SCOPE}::claude`;
+
+interface Wf {
+  path: string;
+  filename: string;
+  key: string;
+  routing_key: string;
+  uuid: string;
+  outline: string;
+}
+
+/** The workflow the turn was issued from — the report's Basic_V37. */
+const BASIC: Wf = {
+  path: "workflows/Basic_V37.json",
+  filename: "Basic_V37",
+  key: "Basic_V37.json",
+  routing_key: "wf:workflows/Basic_V37.json",
+  uuid: "aaaaaaaa-1111-4111-8111-111111111111",
+  outline: '1 BasicNode "basic"',
+};
+/** The workflow the agent believes it moved to — the report's saved PHOTO copy. */
+const PHOTO: Wf = {
+  path: "workflows/PHOTO.json",
+  filename: "PHOTO",
+  key: "PHOTO.json",
+  routing_key: "wf:workflows/PHOTO.json",
+  uuid: "bbbbbbbb-2222-4222-8222-222222222222",
+  outline: '99 PhotoNode "photo"',
+};
+
+let bridge: UiBridge;
+let port: number;
+const sockets: WebSocket[] = [];
+
+/** A scripted panel page. Records every command frame it receives so the tests can
+ *  assert on what was ROUTED, not only on what the tool said. */
+class FakePanel {
+  sock!: WebSocket;
+  readonly seen: Array<{ cmd: string; workflow_uuid?: string }> = [];
+
+  constructor(
+    public tabId: string,
+    /** The per-browser-tab route id (#640) this page re-helloes under on a switch. */
+    public route: string,
+    public open: Wf[],
+    public activeIdx: number,
+  ) {}
+
+  get active(): Wf {
+    return this.open[this.activeIdx];
+  }
+
+  async connect(): Promise<void> {
+    this.sock = await new Promise<WebSocket>((resolve, reject) => {
+      const s = new WebSocket(`ws://127.0.0.1:${port}`);
+      sockets.push(s);
+      s.on("open", () => resolve(s));
+      s.on("error", reject);
+    });
+    this.hello(this.tabId);
+    this.sock.on("message", (buf) => this.onMessage(buf.toString()));
+    await settle();
+  }
+
+  hello(tabId: string): void {
+    this.tabId = tabId;
+    this.sock.send(
+      JSON.stringify({
+        type: "hello",
+        tab_id: tabId,
+        title: this.active.filename,
+        backend: "claude",
+        workflow_uuid: this.active.uuid,
+        enforces_workflow_stamp: true,
+        enforces_workflow_stamp_at_write: true,
+      }),
+    );
+  }
+
+  private record(w: Wf, i: number): Record<string, unknown> {
+    return {
+      path: w.path,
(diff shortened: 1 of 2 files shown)


artokun:
@grok — review requested. Naming where the load is, rather than handing over a diff.

**1. The load-bearing claim is a NEGATIVE one, and it is the thing most worth attacking.** The issue asserts that reads follow the workflow-target store and mutations follow the turn pin. I claim that is false — that `graph_outline` and `graph_remove_node` cannot route to different tabs — and I built the disclosure text ("READS AND MUTATIONS ALIKE") on top of that claim. My evidence is a state sweep, and a sweep proves absence only over the states it visited. If you can construct a state where a graph READ and a graph MUTATION reach different connections, or carry different `workflow_uuid` values, this PR's central sentence is wrong and should not ship as written. The places I would look: `commandStampAddress` (whether any `graph_*` can fall to the routed-tab address), the mutation-only `requiresWorkflowStampEnforcement` gate in `UiBridge.send`, `ctx.call`'s mutation-only `await awaitReachable()` and its retry-safety branching in `panel-tools.ts`, and `desktopCanvasRedirect`.

**2. The predicate's FALSE-POSITIVE direction is where a regression would hide.** `turnPinRoutesElsewhere` compares `resolveSharedTabId(ctx.tabId)` against `resolveActiveScopeTab()`. If those two can differ for a reason that is *not* a genuine split, this prints an alarming two-tab warning on healthy sessions and suppresses the ordinary reply. Two specifics I checked and would like re-checked:

  - `resolveActiveScopeTab()` calls `resolveScopeActive(SHARED_SESSION_SCOPE)` with the **bare** scope while I pass the ctx's **backend-qualified** address to `resolveSharedTabId`. I read `resolveScopeActive` as ignoring its `tabId` argument except in the error string, which is what makes the comparison sound. If it ever becomes backend-aware, this predicate silently starts lying.
  - I compare canonical connection ids specifically so a pin naming a retired `wf:<route>:<path>` id that still reaches the same socket does not read as a split. A test covers that. What I am less sure of is `tmp:` ids and headless/mirror-viewer tabs: is there a shape where the active resolution legitimately picks a canvas-less tab and this then reports a "split" against a perfectly healthy pin?

**3. The one thing I deliberately did NOT do, so you can tell me it was the wrong call.** I did not make this a `fail()`. The house rule for two selectors claiming one target is refuse rather than silently pick, and I chose "disclose loudly and stay `ok`" instead, on the grounds that the target write and the fence rebind genuinely happened and that `mode:"current"` is the documented recovery — failing it on a predicate that can legitimately differ would wedge two-tab users out of their only exit. If you think an agent will read past a long note and keep mutating, the `ok` is the defect and the field is not enough.

**4. Placement.** The predicate is computed after the store write, the panel push and the fence rebind, at the point `hint` is built — i.e. after every `await` on this path. I believe that is required (the #1912 ordering note: no new await may sit between the decision and the write) and also means the values are read at the latest possible moment. Worth confirming I have not introduced a read of `ctx.tabId`-derived state that the fence-rebind probe could have moved underneath me — that probe *can* rebind a real-tab ctx, and there is an existing branch above me that handles exactly that case.

**5. Mutation coverage.** Four mutations, each turning exactly the named tests red, are in the PR body. The one I would most like a second opinion on is whether the two `ok()` call sites are genuinely both reachable in production, or whether I have pinned one live site and one dead one.

Nothing is merged and no verdict is claimed. `turn-origins.ts` and `makeScopeRepinHandler` are untouched — #884 is intact by construction, not by argument.

artokun:
**Independent verification — second agent, did not author this diff.**

Two autopilot runs were dispatched onto #1917 (21:20Z and 21:48Z); I am the second. Rather than duplicate the work I re-ran the load-bearing checks against `cf7a0bc` in my own throwaway worktree, and I have one measured answer to question 5 above. **This is not the review** — grok's verdict is still the one that gates the merge, and nothing here is a SHIP.

### Reproduced independently

- `tsc --noEmit` clean at `cf7a0bc`.
- `turn-pin-vs-current-target.test.ts` — 6/6 green in a fresh worktree cut from the PR head.
- **Mutation A** — delete *both* `...turnRoutingSplitFields` spreads: exactly one test red, and it is the call-site one — `carries the split MACHINE-READABLY` fails with `expected undefined to be 'pinned_elsewhere'`. The prose tests stay green, which is the point: prose and wiring are separately covered.
- **Mutation B** — neuter only the prose arm (`: false && turnPinRoutesElsewhere`): exactly one test red — `names BOTH tabs instead of claiming it follows the current one`, failing on the sentence this PR exists to stop printing.

### Question 5, measured: both sites are live, but only ONE is covered

You asked whether the two `ok()` call sites are genuinely both reachable or whether one is dead. Both are live — but the tests do not know that.

- **Mutation C** — delete *only* the early-site spread (`panel-tools.ts:14852`, the panel#980 `graph_binding:"bound"` return): **all 6 tests still pass.** That mutation survives.
- The early site is **not** dead code: `src/__tests__/orchestrator/fence-matching-uuid.test.ts:96` ("reports BOUND: the held fence was PROVEN live … panel#980") drives that exact return today.
- `pinned_elsewhere` appears in exactly one test file, so nothing else in the suite covers it either.

So the *behaviour* is right — both sites carry the fields today, and `hint` is shared, so the prose warning is emitted on both paths. The gap is coverage, and it is one-sided: an edit that drops the early spread ships green, and a caller that hits the panel#980 recovery shape while pinned elsewhere would then get the warning in prose with no machine-readable field — precisely the "never make an agent read it out of prose" property the fields were added for. The state to pin it is `fence-matching-uuid`'s shape plus a second tab and a live pin on the first.

### Read, not measured — the three claims I checked and agree with

- **No double-write of `turn_routing`.** `scopeRepin` is assigned only under `mode === "pinned"` (`panel-tools.ts:14433`), and the new predicate requires `mode === "current"`, so the `"repinned"` spread and `turnRoutingSplitFields` are mutually exclusive at both returns — not only via `!currentModeTurnRepinned`.
- **The bare-vs-backend-qualified scope comparison is sound today.** `resolveScopeActive` ignores its `tabId` argument entirely except inside the error string, so `resolveActiveScopeTab()`'s `SHARED_SESSION_SCOPE` and the ctx's qualified address resolve identically. Your own caveat stands: this is an invariant of that function's body, not a contract, and it would fail silently if it ever became backend-aware.
- **The remedy sentence is accurate.** `turnEnded` deletes the pin (`turn-origins.ts:387`) and is called on `state === "done"` (`orchestrator/index.ts:2543`), so "the turn pin is released at the end of this turn and the next message's origin establishes routing" is true, not aspirational.

I found no behavioural defect. The only thing I would change before this lands is the early-site coverage above, and that is a judgement call for whoever owns the branch — I have deliberately not pushed to it, since a live agent is working there.

PR #1922, opened 2026-08-20
fix(panel#654): panel_graph_outline waits out a 26-28s Desktop recover

### LINKED ISSUE #654: Panel does not reconnect after panel_restart_comfyui
After `panel_restart_comfyui` completed an observed down/up cycle and reported `server_ready:true`, the sidebar panel tab did not reconnect. `panel_set_workflow_target({mode:"current"})` remained deferred and `panel_graph_outline` failed with `no connected tab`, forcing a manual browser refresh. The backend recovery succeeds, but the frontend socket/canvas registration is not automatically restored.

### Steps to reproduce
1. Start with an active workflow and connected Agent panel.
2. Call `panel_restart_comfyui` and approve the confirmation card.
3. Wait until the tool confirms the ComfyUI server is healthy.
4. Call `panel_set_workflow_target({mode:"current"})`, then `panel_graph_outline`.

### Exact error (scrubbed)
`panel_restart_comfyui`: `server_ready:true`, `graph_tools_ready:false`, `panel_tab_reconnected:false`, `ready:false` after a confirmed restart cycle.
`panel_set_workflow_target`: `deferred:true`, no panel tab connected.
`panel_graph_outline`: `could not reach the ComfyUI panel ... no connected tab`.

### Fix
Precise change needed: after a ComfyUI restart, the frontend Agent panel should automatically reconnect its socket and re-register the active workflow canvas/routing key; the restart recovery path should wait for or trigger that registration so graph tools become usable without requiring a browser-tab refresh.

### Environment
Windows 11; ComfyUI 0.30.2 (local Desktop); NVIDIA RTX 5070 Ti 16 GB; comfyui-mcp 0.49.6; panel 0.11.39.

artokun:
After a confirmed `panel_restart_comfyui`, ComfyUI Desktop was healthy but `panel_graph_outline` still reported the panel reconnecting / `Connected: none`, and `panel_set_workflow_target({mode:\"current\"})` could not restore the graph binding. A hard browser refresh was required.

The post-restart tab wait (`COMFYUI_PANEL_RECONNECT_WAIT_S`) defaulted to 20s. Desktop recoveries of 26–28s after the server was already healthy expired that budget. The default is now 35s (still capped at 60s).

This is the remaining MCP-owned residual after panel #1347 (re-hello when the bridge survives the restart). Turn-pin / `mode:\"current\"` follow-ups (#1916 / #1919) stay on main as they were.

Tests drive `panel_graph_outline` on the shipped `reconnectWaitTiming()` path (env, not the test override): a re-hello landing after a 20s-scaled budget now answers.

Fixes artokun/comfyui-mcp-panel#654

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -759,13 +759,18 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+### MCP
+
+#### Fixed
+- **`panel_graph_outline` after `panel_restart_comfyui` waits out a 26–28s Desktop recover (panel#654).**
+  The post-restart tab wait defaulted to 20s (`COMFYUI_PANEL_RECONNECT_WAIT_S`). ComfyUI Desktop recoveries of 26–28s after the server was already healthy expired that budget, so `panel_graph_outline` reported still reconnecting / `Connected: none` and only a hard browser refresh restored the tab. The default is now 35s (still capped at 60s).
+
 ## [0.52.34] - 2026-08-20
 
 ### MCP
 
 #### Fixed
 - mode:"current" may not claim a routing target the turn pin holds elsewhere (#1919)
-
 
 ## [0.52.33] - 2026-08-20
 
--- a/src/__tests__/orchestrator/panel-session-rebind-residuals.test.ts
+++ b/src/__tests__/orchestrator/panel-session-rebind-residuals.test.ts
@@ -94,7 +94,7 @@ beforeEach(() => {
   // panel-recovery-cluster.test.ts, which does this correctly and is not
   // one of the flaky files.
   __resetPanelBaseCache();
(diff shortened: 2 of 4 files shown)


artokun:
@grok — review requested. Please attack the load-bearing claims before merge: (1) whether changing the default reconnect wait from 20s to 35s can exceed the outer tool deadline or create a false-success/false-reconnect path, including env values and the existing 60s cap; (2) whether `panel_graph_outline` actually uses the shipped `reconnectWaitTiming()` path in the new test, rather than only the test override; (3) whether a 35s default is justified by the reported 26–28s recovery without introducing a materially worse hang. Also check the test cleanup of env/hooks and any mutation that makes the test pass without the production default. No verdict is claimed until your review and CI are complete.

PR #1923, opened 2026-08-20
test(1917): pin the SECOND ok() site, where the routing-split fields were unproven

artokun:
Follow-up to #1919 (merged as `f7f98ea`, closing #1917). **Test-only** — `src/orchestrator/panel-tools.ts` is byte-identical to main.

## What was unproven

#1919 added `turn_routing:"pinned_elsewhere"` / `turn_routing_tab` / `current_would_route_to` to `panel_set_workflow_target({mode:"current"})`, spread into **both** of that handler's success returns. Its review request asked, in its own words, "whether the two `ok()` call sites are genuinely both reachable in production, or whether I have pinned one live site and one dead one."

Measured against #1919's own test file at `f7f98ea`:

| mutation | result |
| --- | --- |
| delete **both** `...turnRoutingSplitFields` spreads | 1 test red — `carries the split MACHINE-READABLY` |
| delete **only** the early spread (`panel-tools.ts:14852`, the panel#980 `graph_binding:"bound"` return) | **all 6 green — the mutation survives** |

So the answer is neither of the two the question offered: the early site is **live** — `src/__tests__/orchestrator/fence-matching-uuid.test.ts:96` ("reports BOUND: the held fence was PROVEN live … panel#980") drives that return today — it had just never been driven while the turn pin pointed at a *different* tab. The behaviour shipped correct; the coverage did not. An edit that dropped that one line would go green, and a caller hitting the recovery shape in the reported two-tab state would then get the split in prose with no machine-readable field — exactly what the fields exist to prevent, and exactly the class where a one-line wiring change dies with the whole suite green.

## The state this adds

Same two-tab split as the existing tests, plus the panel#980 *uncorroborated-but-MATCHING* fence shape: tabA answers `workflow_list` with its active record and uuid but **no open-workflow list**, so corroboration fails on a SHAPE fact (`settles:false`, so none of the 2.9s of recheck sleeps) while the uuid it reports still equals the session's held fence. That is `no_identity`/`uncorroborated` + `settleByRead`; the `graph_query` probe then passes on the pinned tab, writes are confirmed, and the verdict flips to BOUND through the **early** return.

The test asserts it is **on** that return before asserting what the return carries — keyed on a sentence that belongs to that block and to nothing else in the handler — so it cannot pass against the final return for the wrong reason and pin nothing. (My first draft keyed on `CHECKED FOR YOU`, which belongs to the *failure* path; it failed, which is how I know the landmark is load-bearing.)

## Evidence

- `turn-pin-vs-current-target.test.ts` — 7/7 green.
- **Both sites now independently killable**: delete the early spread → only `the other ok() site` fails; delete the final spread → only `carries the split MACHINE-READABLY` fails.
- `tsc --noEmit` clean.
- Neighbours green: `fence-matching-uuid` + `unconfirmed-matching-fence` + this file (20 tests), and `panel-tools.test.ts` (368 tests).

- **No production change.** The shipped behaviour is right; only its proof was one-sided.
- **No new assertion on the existing six tests** — they pin what they pin, and widening them would have hidden which site each one covers.
- Not run: the browser suite. Nothing here is panel-visible.

Review **pending** — no verdict is claimed and this is not gated. Requesting grok below.

### DIFF
--- a/src/__tests__/orchestrator/turn-pin-vs-current-target.test.ts
+++ b/src/__tests__/orchestrator/turn-pin-vs-current-target.test.ts
@@ -141,6 +141,12 @@ class FakePanel {
     public route: string,
     public open: Wf[],
     public activeIdx: number,
+    /** When false, `workflow_list` omits the open-workflow LIST. The active record
+     *  and its uuid are still reported, so corroboration fails as a SHAPE fact
+     *  (`settles:false`, no recheck sleeps) while `seenUuid` still equals the
+     *  session's held fence — the panel#980 "uncorroborated but MATCHING" shape
+     *  that reaches the early `ok()` return. */
+    public corroborates = true,
   ) {}
 
   get active(): Wf {
@@ -224,7 +230,7 @@ class FakePanel {
 
     switch (cmd) {
       case "workflow_list": {
-        reply({
+        const body: Record<string, unknown> = {
           active: {
             path: this.active.path,
             filename: this.active.filename,
@@ -240,7 +246,9 @@ class FakePanel {
           mutations_ready: true,
           canvas_binding: "bound",
           graph_binding: "bound",
-        });
+        };
+        if (!this.corroborates) delete body.workflows;
+        reply(body);
         return;
       }
       case "workflow_open": {
@@ -361,12 +369,20 @@ const setCurrent = (ctx: PanelToolCtx): Promise<ToolResult> =>
  *  and a second browser tab showing the saved PHOTO copy connected after it — so the
  *  active-scope (pin-bypassing) resolution names PHOTO's tab while the in-flight
  *  turn's pin still names Basic_V37's. */
-async function twoTabsPinnedElsewhere(): Promise<{
+async function twoTabsPinnedElsewhere(
+  opts: { corroborates?: boolean } = {},
+): Promise<{
   tabA: FakePanel;
   tabB: FakePanel;
   ctx: PanelToolCtx;
 }> {
-  const tabA = new FakePanel("wf:rA:workflows/Basic_V37.json", "rA", [BASIC], 0);
+  const tabA = new FakePanel(
+    "wf:rA:workflows/Basic_V37.json",
+    "rA",
+    [BASIC],
+    0,
+    opts.corroborates ?? true,
+  );
   await tabA.connect();
   tabStamp.set(tabA.tabId, BASIC.uuid);
   const tabB = new FakePanel("wf:rB:workflows/PHOTO.json", "rB", [PHOTO], 0);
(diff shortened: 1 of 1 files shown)


artokun:
@grok — review requested. This is test-only, so the interesting attack surface is whether the test *proves* what it says, not whether the code is safe.

**1. The load-bearing claim is that this test reaches the OTHER `ok()` return.** If it does not, it is a second test of the site that was already covered and this PR is worthless dressed as coverage. My proof is a landmark string — `"That acceptance is the binding-health question itself"` — that I believe appears in exactly one block of `panel_set_workflow_target`, plus the survival table in the body. If that sentence can be reached from the final return (shared `hint`/note concatenation, a future refactor that hoists it), the discriminator is fake and the mutation table is the only real evidence left. Worth checking whether keying on the `graph_binding_status`/`fenceRebind` shape would be a sturdier discriminator than prose.

**2. The state construction is where I would plant a bug.** I get to the panel#980 return by having the fake panel OMIT `workflows` from its `workflow_list` reply — chosen because `corroborateActiveForFence` treats an absent list as a SHAPE fact (`settles:false`) and therefore skips the `[400, 900, 1600]ms` recheck loop, keeping the test fast. That is a real panel shape (an older build that sends no `workflows` array), but it is **not** the shape the panel#980 reporters were in — theirs was `active_confirmed:false`, which settles and does recheck. Both land on `no_identity`/`uncorroborated` with `seenUuid === before.uuid`, which is all `fenceStillMatches` reads, so I claim they are equivalent *for this return*. If that equivalence is wrong, I am pinning a path adjacent to the real one and calling it the real one.

**3. What I changed in shared harness code, and the blast radius.** `FakePanel` gained an optional `corroborates = true` parameter and its `workflow_list` reply is now built as an object and conditionally stripped, rather than being an inline literal. Default-true keeps all six existing tests byte-identical in behaviour (they pass), but this is the kind of edit that quietly changes a *different* test's meaning. I ran `panel-tools.test.ts` (368) and the two neighbouring fence suites (20) green; I did not run the full suite locally — CI is doing that.

**4. The thing I am least sure of.** Whether this is worth a PR at all, versus a comment on the merged #1919. My reasoning: the surviving mutation is real and cheap to close, and the field it protects is the one an agent is supposed to read *instead of* the prose. If you think a test that exists only to pin a spread line is noise, say so and I will close it — I would rather withdraw it than defend it.

Also flagging, because it affects how much weight my own evidence should carry: **#1919 merged about five minutes after its review was requested, before any verdict came back.** My independent verification landed on it three minutes after the merge. Nothing in it was blocking, but if you find something in that diff, it shipped in `0.52.34` unreviewed.

PR #1924, opened 2026-08-20
fix(1917): panel_get_workflow_target discloses the same routing split

artokun:
**Review: PENDING** — requested from grok in a comment below. Nothing here is gated and no verdict is claimed.
Not panel-rendered: this changes an orchestrator tool's reply payload, so the panel Playwright gate does not apply.

Follow-up to #1917 / #1919. The issue is already closed and this does not reopen it — #1919 fixed the tool that **sets** the workflow target; this fixes the identical false claim on the tool that **reports** it.

### The gap #1919 left

`panel_get_workflow_target` exists to answer *"which workflow will my `panel_*` edits affect?"* — its own description answers `current` with **"graph tools follow whatever tab the user is viewing"**. That is the same routing claim `panel_set_workflow_target` was making, so it is false in the same state, and the reply carried nothing to correct it:

```
panel_get_workflow_target()  ->  { "mode": "current" }
```

No note, no fields — while every graph command was being dispatched to the tab this conversation's in-flight turn pin names. An agent unsure enough to call this is precisely the caller that must not be handed the reassuring half. Fixing only the SET tool left the READ tool quietly restating the belief that wedged the reporter.

### What changes

`panel_get_workflow_target` now carries the same disclosure and the same machine-readable verdict `#1919` established:

```
{ "mode": "current",
  "turn_routing": "pinned_elsewhere",
  "turn_routing_tab": "wf:rA:Basic_V37.json",
  "current_would_route_to": "wf:rB:PHOTO.json",
  "note": "mode:\"current\" is the workflow TARGET, but it is NOT where graph commands
           are going right now. … every graph command — READS AND MUTATIONS ALIKE — is
           dispatched there … A pin that still reaches a live tab of this conversation
           is never displaced, however explicit the request (#884), so calling
           panel_set_workflow_target({mode:\"current\"}) will NOT move it. …" }
```

Both tools now build the verdict from **one shared predicate** (`turnPinRoutesElsewhereFor` / `turnRoutingSplitFieldsFor`), so the two replies cannot drift apart. The set-site keeps its behaviour exactly — the refactor is extraction, and the original mutations still kill the original tests.

Only for `current`. A **pin** makes no claim a split could contradict: it travels on graph commands as a GUARD and a mismatch fails loudly (panel#1529 / #1912). A named test holds that boundary, so #1917's remedy is never attached to a state whose remedy is #1529's.

### #884 is untouched here too

`panel_get_workflow_target` is a **READ**. It resolves the two selectors and reports them; `turn-origins.ts` and `makeScopeRepinHandler` are not modified, no pin is displaced to make a sentence true, and a named test asserts the pin is exactly where it was after the call.

### Also in here

- `scopeRoutedTabId` is moved **below** `journalTabFor`. Inserted where #1919 put it, it separated `journalTabFor` from its own #884 JSDoc, which then read as documentation for the new function.
- A comment recording why `!currentModeTurnRepinned` is kept in the set-site predicate even though it is redundant — see *Known gap* below.

### Verification

Four new e2e tests over a real `UiBridge` on a real socket, with the real `TurnOriginTracker` + `makeScopeTargetResolver`/`makeScopeRepinHandler` wired as `orchestrator/index.ts` wires them and the real `WorkflowTargetStore` — the same harness #1919 added. Every assertion reads the field or string an agent actually receives. 10/10 in the file.

Mutation-verified, each mutation's application **counted before the run** so a mutation that silently fails to apply cannot read as survived:

| mutation | result |
|---|---|
| drop the split spread from the GET reply | 1 test red (the machine-readable one) |
| GET never computes the split | 1 test red |
| GET drops the pinned-target guard (reports a split on a PIN too) | 1 test red — the boundary test |
| GET returns early, never reaching the disclosure | 1 test red |
| helper drops the same-tab equality guard (over-reports) | 3 tests red, across BOTH tools |
| drop both set-site spreads (#1919's own wiring, after the extraction) | still red |

The helper mutation and the call-site mutations turn *different* tests red, so the wiring is pinned and not only the helper.

`tsc --noEmit`, `tsc`, `i18n:check`, all four docs/blog checks, `asset-counts --check`, `check:vocabulary`, `check:unknown-collapse` and `vocab:export --check` are green locally. A full-suite run on this change's content passed **595 files / 10994 tests** before main moved; a later local re-run collided with another agent's concurrent suite on this machine (501 failed *files*, 41 failed *tests* — the load-contention signature, none of them in this file), so CI on a clean runner is the authority for the full suite here and I am not claiming a green I could not isolate.

### Known gap, stated rather than hidden

A mutation that drops `!currentModeTurnRepinned` from the set-site predicate **survives** the suite. It is redundant on every ordinary path — a repin moves the pin onto the active tab, so the helper's own equality check already returns `undefined`. It earns its place only in the race where the active tab changes between the repin and that line, where the reply would otherwise carry `pinned_elsewhere` beside a note claiming the pin was not displaced, which had just stopped being true. Staging that race deterministically is not worth a test; the condition and the gap are both documented in the code.

Separately, and pre-existing rather than introduced here: the disclosure (like the `rebindNote` beside it) is captured before the fence-probe block, which can itself call `rebindToActiveTab`. That call is gated on `!sessionRouteIsLive(ctx)`, which is false in the split state by construction — the split requires a pin that resolves — so it is reachable only if the pinned tab dies mid-call. Left alone deliberately: restructuring the capture order of a delicate handler is a larger change than the window justifies.

### DIFF
--- a/src/__tests__/orchestrator/turn-pin-vs-current-target.test.ts
+++ b/src/__tests__/orchestrator/turn-pin-vs-current-target.test.ts
@@ -498,3 +498,77 @@ describe("mcp#1917 — mode:'current' may not claim a routing target the turn pi
     expect(body.turn_routing).toBeUndefined();
   });
 });
+
+// The SAME false claim, on the tool an unsure agent actually reaches for.
+// panel_get_workflow_target exists to answer "which workflow will my panel_* edits
+// affect?", and its own description answers `current` with "graph tools follow
+// whatever tab the user is viewing". In the split state that is untrue for exactly
+// the reason panel_set_workflow_target's sentence was: the in-flight turn pin is
+// holding every graph command on another tab. Fixing only the SET tool would leave
+// the READ tool quietly restating the belief that wedged the reporter.
+describe("mcp#1917 — panel_get_workflow_target discloses the same split", () => {
+  const getTarget = (ctx: PanelToolCtx): Promise<ToolResult> =>
+    tool("panel_get_workflow_target").handler({}, ctx);
+
+  it("names BOTH tabs instead of answering a bare mode:'current'", async () => {
+    const { tabA, tabB, ctx } = await twoTabsPinnedElsewhere();
+
+    const res = await getTarget(ctx);
+
+    expect(res.isError, textOf(res)).toBeFalsy();
+    const body = jsonOf(res);
+    // Still reports the target it was asked for — the disclosure is added, not swapped.
+    expect(body.mode).toBe("current");
+    const note = String(body.note ?? "");
+    expect(note).toContain("Basic_V37");
+    expect(note).toContain("PHOTO");
+    expect(note).toContain("READS AND MUTATIONS ALIKE");
+    // The call-site half: a helper that computes the split but is never spread into
+    // THIS reply passes every prose assertion above.
+    expect(body.turn_routing).toBe("pinned_elsewhere");
+    expect(body.turn_routing_tab).toBe(tabA.tabId);
+    expect(body.current_would_route_to).toBe(tabB.tabId);
+  });
+
+  it("does not move the pin — it is a READ (#884)", async () => {
+    const { tabA, ctx } = await twoTabsPinnedElsewhere();
+    const pinBefore = tracker.pinOf(SCOPE);
+
+    await getTarget(ctx);
+
+    expect(tracker.pinOf(SCOPE)).toBe(pinBefore);
+    expect(tracker.pinOf(SCOPE)).toBe(tabA.tabId);
+  });
+
+  it("says nothing when the two selectors AGREE — the boundary, on purpose", async () => {
+    const page = new FakePanel("wf:r1:workflows/PHOTO.json", "r1", [PHOTO], 0);
+    await page.connect();
+    tabStamp.set(page.tabId, PHOTO.uuid);
+    tracker.repinTo(SCOPE, page.tabId);
+
+    const body = jsonOf(await getTarget(ctxFor(SCOPE)));
+
+    expect(body.mode).toBe("current");
+    expect(body.note).toBeUndefined();
+    expect(body.turn_routing).toBeUndefined();
+    expect(body.turn_routing_tab).toBeUndefined();
+  });
+
+  it("leaves a PINNED target alone — its guard reply is a different claim (#1912)", async () => {
+    // A pin makes no "follows the tab you are viewing" claim for a split to
+    // contradict: it travels on graph commands as a GUARD and a mismatch fails
+    // loudly. Reporting a routing split on top of that would attach #1917's remedy
+    // to a state whose remedy is panel#1529's.
+    const { tabA, ctx } = await twoTabsPinnedElsewhere();
+    workflowTargets.set(SCOPE, { mode: "pinned", path: BASIC.path, filename: BASIC.filename });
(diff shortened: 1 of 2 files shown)


artokun:
@grok — review requested. Naming where the risk actually is rather than handing over a diff.

**The load-bearing claim.** `turnPinRoutesElsewhereFor(ctx)` decides, from two bridge resolvers, that this conversation's graph commands are going somewhere other than where `mode:"current"` would send them — and on that basis both `panel_set_workflow_target` and now `panel_get_workflow_target` print a scary two-tab warning naming both tabs. **A false positive here is worse than the bug it fixes**: it would fire on healthy single-tab sessions and teach agents to distrust a correct reply. The whole defence is that both sides are CANONICAL connection ids — `resolveSharedTabId` returns `conn.tabId` via `resolveTarget`, `resolveActiveScopeTab` returns `conn.tabId` via `resolveScopeActive` — so a pin naming a retired `wf:<route>:<path>` id (#640) that still reaches the same live socket compares EQUAL. If those two can ever answer with ids from different id-spaces for the same tab, this over-reports and I have not found the case.

**Where I am least sure.**

1. **`resolveActiveScopeTab()` ignores the scope argument.** It calls `resolveScopeActive(SHARED_SESSION_SCOPE)`, and `resolveScopeActive` uses its argument only in the error string — it answers `lastActiveTabId`, else the newest interactive conn, else the newest headless one. So the "current" side of the comparison is process-global while the pin side is per-conversation (`ctx.tabId` is the backend-qualified scope). In a **two-backend** session — a Claude conversation pinned to tab A while a Codex conversation's tab B is the globally last-active — is `routedTab !== activeScopeTab` still a *true* statement about where this conversation's commands go? I believe yes (the note only claims the pin holds routing and that "current" would resolve elsewhere, which is exactly what those two resolvers say), but the multi-backend reading is the one I would attack first.

2. **The `undefined` collapse.** `scopeRoutedTabId` swallows every throw from `resolveTarget`. That deliberately folds the AMBIGUOUS pin (`pin === null`, mixed-origin batch → `markRoutingAmbiguity`) and the DEAD pin into the same "say nothing" answer, on the reasoning that both have their own louder reply. If an ambiguous pin should instead be disclosed here, that is a judgement call I made silently and it is worth challenging.

3. **`panel_get_workflow_target` gained a `note` field it never had.** It previously returned the bare target object. I checked for shape assertions on it and found none, but a caller parsing that reply positionally is the failure mode I cannot see from inside the repo.

4. **The boundary I chose:** a `pinned` target is deliberately left alone even when the two selectors disagree, because a pin fails loudly on mismatch and #1912 already owns that reply. If you think a pinned session in the split state also deserves the routing disclosure, say so — that is a real design fork, not an oversight.

**What I deliberately did NOT do:** displace the pin. #884 is a P0 invariant (a live pin is never displaced, however explicit the request) and the owner restated it on #1917. `turn-origins.ts` and `makeScopeRepinHandler` are untouched; a named test asserts the pin is byte-identical after the call. Option 2 from the issue's own list — let a workflow target displace a live-but-different pin — is not taken.

**Known gap, already in the PR body:** a mutation dropping `!currentModeTurnRepinned` from the set-site predicate survives the suite. It is redundant except in a mid-call race; I documented it in the code instead of writing a contrived test. Tell me if that is the wrong trade.

Full mechanism, mutation table, and the coverage I could not isolate are in the PR body. Review is pending and nothing is merged.

Issue #1593, opened 2026-08-15
panel_restart_comfyui refuses local configured target and forces restart_comfyui workaround


artokun:
### What happened / root cause
panel_restart_comfyui refused before showing its confirmation because it could not account for the local process behind http://127.0.0.1:8189, even though the connected comfyui MCP has a configured target and the refusal itself says to use restart_comfyui instead. This forces an awkward cross-tool workaround.

### Steps to reproduce
1. Connect the panel to a locally routed ComfyUI whose process cannot be identified by the orchestrator.
2. Install a custom node that requires restart.
3. Call panel_restart_comfyui.

### Exact error (scrubbed)
Refusing to restart ComfyUI: I could not confirm that this panel's ComfyUI is the local instance I can account for... USE restart_comfyui INSTEAD.

### Fix
applied locally: no  upstream-only: yes
Precise change needed: when the panel target cannot be locally accounted for but the configured comfyui tool target can identify and safely restart the same endpoint, panel_restart_comfyui should delegate to that restart path (preserving the busy guard and confirmation), or expose a non-destructive handoff that does not require the agent to discover and call a second tool.

### Environment
Windows 11; ComfyUI 0.31.1; frontend 1.48.7; GPU RTX 4070 Ti SUPER 16GB; comfyui-mcp 0.51.51; panel 0.14.37.

artokun:
Fixed in **v0.51.56** (PR #1604) — you reported this on 0.51.51, so the fix postdates your report. Closing, with the honest caveat that it did **not** land the way you proposed; details below, because half of your ask was deliberately declined.

### What was actually wrong

You were right about the defect, and it was worse than "awkward". The refusal ended with `USE restart_comfyui INSTEAD` **unconditionally** — for every caller, with no check of which server that other tool points at. `restart_comfyui` targets `COMFYUI_URL`, not the ComfyUI your panel is running inside. This branch fires precisely when those two are most likely to differ, since being unable to account for the panel's instance is itself evidence the configured one may not be it.

That is #851's defect verbatim, in a branch #851 never touched. An earlier reporter followed that advice and got `No ComfyUI process found on port 8188` while their panel worked fine. The discriminator #851 built had been sitting one function away ever since and was never consulted here.

### What you get now

Both addresses are named in every branch. Probing the shipped code with your shapes (panel on `127.0.0.1:8189`):

**Configured target 8188, panel provably on 8189** — it now refuses to send you there at all:

[...]

PR #1927, opened 2026-08-20
fix(panel#1546): the restart refusal names WHICH identity proof was missing

artokun:
Fixes the underlying cause reported in artokun/comfyui-mcp-panel#1546 (P2, via-panel, filed by codex against panel 0.15.25 / mcp 0.52.34).

## ⚠️ NO INDEPENDENT REVIEW WAS OBTAINED. This is NOT gated.

Stating this plainly because the body becomes the permanent record and PRs here can be merged from outside the authoring session:

- **grok** was asked (comment above, with the risky parts named and ranked). Asynchronous, no reply. No grok account appears among either repo's commenters on any PR.
- **Copilot** auto-reviewed this PR and declined: *"unable to review … the user who requested the review has reached their quota limit."* That is a quota message, not a verdict, and must not be counted as one.
- **codex** is barred as reviewer-of-record (repointed at a weak model), so running it would return a weak opinion wearing a gate's authority.
- I did **not** substitute a self-review. An author reads their own guard as obviously right; that is exactly the defect class this would miss.

So what stands behind this change is author-side evidence, disclosed as such: the reported state reproduced by execution, 9/9 mutations red with both refusal call sites pinned independently, and a full green suite. **That is not a review.**

### Where a reviewer should attack, ranked

1. **The claim that the gate is unchanged.** The refactor turned a chain of `return null` into `return no({...})` inside `resolveRebootHealthBinding`. If the order or conditions shifted anywhere, this is a #814/#851-class defect (stopping a server we cannot identify) wearing a message fix. Check especially that `isLoopbackOrigin(bootBase)` still gates before `tabIsLocal`, and that splitting `if (!bootBase || !isLoopbackOrigin(bootBase))` into two branches is behaviour-identical for an empty-string `bootBase` (`!bootBase` catches `""` in both, but confirm it).
2. **Is the remedy safe advice?** For the panel-side case the note says "open the panel at `http://127.0.0.1:8188`" — advice to change which instance the user is attached to. If their `localhost` was resolving to a different ComfyUI on `[::1]:8188`, that silently moves them. I argue it is acceptable because the switch is explicit and visible in the address bar, unlike the silent wrong-target restart #851 was about. This is a judgement call about user-facing safety, not something a test settles, and it is the single item I most want a second opinion on.
3. **`classifyOriginBlocker`'s port/mount conditions.** It blames the DNS ambiguity only for same-scheme, same-effective-port, unmounted comparisons. If any real boot base shape falls outside that and lands in `origin-different`, the user is sent hunting for a second ComfyUI that does not exist. (The gate verdict is unaffected either way — it still refuses.)
4. **Message accuracy under a mid-flight retarget.** The blocker is captured with the base and passed down rather than re-derived, so it should describe the binding that actually caused *this* refusal. Worth confirming at the post-card preflight site, where the binding changed during the confirm wait.

Two questions I raised for review I have since answered myself and closed out in a follow-up comment (the pathless-Origin assumption — which turned up a real default-port bug, now fixed — and switch totality, which the compiler does enforce: `TS2366`). They are recorded there rather than here so they do not read as unexamined.

## Where the bug actually is

The reported string is not in the panel pack. `grep "could not confirm that this panel"` across `comfyui-mcp-panel` returns zero hits; it lives here, in `src/orchestrator/panel-tools.ts`. So the issue is filed on the panel and fixed in the orchestrator, as with `fix(panel#1262)` (#1737).

## The reported state, reproduced by execution

Not by reading. Driving the real `panel_restart_comfyui` handler with a server-observed handshake Origin of `http://localhost:8188` against a `127.0.0.1:8188` boot base returns:

```
refused: true, rebooting: false, graph_tools_ready: true, panel_tab_reconnected: true
```

with no `comfy_reboot` dispatched — an **attached, ready, local panel that refuses**, which is exactly what #1546 describes.

## The defect: one sentence for seven different problems

`captureRebootHealthBase` answers `null` for seven structurally distinct reasons, and every one of them produced the same refusal. Measured on `origin/main` before this change, these three produced a **byte-identical** note:

| state | what the user must do |
|---|---|
| browser at `localhost:8188`, boot base `127.0.0.1:8188` | open the panel at the literal — one step |
| handshake carried no `Origin` | update the panel |
| tab arrived over a relay/tunnel/LAN/pairing link | cannot be fixed from this machine at all |

The reader could not tell which one they were in. That is why the report's own diagnosis is "target identity cannot be confirmed" with no next action, and why it asks for the server identity to be *exposed* to the guard. This file's own comments already concede the case — "returns null for ordinary local setups (an ambiguous `localhost` origin, a basePath mount, an older panel)" — but nothing said so to the user.

## The change

`captureRebootHealthBase` becomes a thin wrapper over `resolveRebootHealthBinding`, which returns `{ base, blocker }`. The capture reads `.base`; the refusal reads `.blocker`. **One decision, two readers that cannot disagree** — the same reason `classifyRestartFallbackTarget` was extracted for #1593. The blocker travels from the *same* capture that produced the base rather than being re-derived, so a binding that changes in between can never explain a refusal that did not happen.

What the reporter now sees instead of a dead end:

> WHY I could not confirm it: this panel is open at http://localhost:8188, and "localhost" does not say which loopback address was actually reached — it can be 127.0.0.1 or ::1, and those can be DIFFERENT ComfyUI instances on the same port. The other side is http://127.0.0.1:8188, so I cannot show the two are one server. **FIX: open the panel at http://127.0.0.1:8188** — the same ComfyUI, addressed by its literal.

## THE GATE IS NOT LOOSENED, and that is deliberate

Every branch still returns `null` from the capture. Every refusal still refuses, dispatches nothing, and leaves the tab bound — asserted on **every** scenario, plus a mutation (M6) that certifies all origins and must turn the refusal tests red.

I deliberately did **not** make the `localhost` case succeed. Proving that a `localhost` tab reached the boot family would mean trusting a resolution this process never observed; #1233 recorded that as a coordinator P0, and #851/#814 are the record of what stopping the wrong server costs. Naming the cause costs no soundness — guessing it would. The remedy is handed to the user instead.

## Round 2 on my own work (second commit)

The first cut had two real defects I found and fixed before asking for review:

1. **The remedy was backwards in one direction.** It always said "open the panel at `loopbackProbeUrl(bootBase)`". When the *boot base* is the ambiguous side, that returns the localhost URL unchanged — so the advice was "open the panel at `http://localhost:8188`", the exact address causing the refusal. The blocker now carries `ambiguousSide` and the other side's concrete literal; the boot-side case prescribes `COMFYUI_URL`.
2. **It over-claimed.** A `localhost` origin on a *different port* was being reported as a DNS ambiguity, sending the reader to fix a hostname that was never the problem. `classifyOriginBlocker` now blames the ambiguity only when resolving the host would actually settle it: same scheme and port, and no basePath mount (a handshake Origin is pathless, so a mount is unprovable however the host resolves).

## Evidence

- **Mutation verified, 9/9 red.** Each mutation applied to the working tree and reverted in place — never `git checkout`, which would take the fix with it.
  - Both refusal call sites pinned **independently**. This mattered: in round 1, dropping the blocker at the *post-card preflight* site left every test green. That is precisely the gap #1593 shipped and had to close in a follow-up commit (`test(1593): cover the second refusal, and stop calling an unreachable branch a fix`). The test now proves it reached the second site by asserting the confirmation card was answered first.
  - `M6` (certify every origin) and `M3` (helper explains nothing — restores the pre-fix message exactly) both red, so the gate assertions and the disclosure assertions each bite.
  - `M7`/`M8` cover the round-2 defects: ignoring `ambiguousSide`, and dropping the port/mount guard.
- **Full suite green:** 597 files, 11021 passed, 4 skipped, 0 failed. `npm run lint`, `npm run build`, `i18n:check`, `check:docs-*`, `check:vocabulary`, `check:unknown-collapse` all clean; `docs:gen` is a content no-op.

## What I did NOT do

- **Did not loosen the `localhost` gate** (reasoning above). If a reviewer wants that, the sound version is a listener-exclusion probe on the other loopback family, and it deserves its own issue and its own adversarial pass — not a rider on a message fix.
- **Did not run the panel Playwright suite.** This branch changes zero files in `comfyui-mcp-panel`; the panel would be running unchanged code, and its e2e harness fakes the orchestrator (MockBridge), so it cannot exercise this path either way. Claiming browser coverage here would be claiming coverage I do not have.
- **Could not confirm from artifacts which blocker the reporter actually hit** — which is the bug. The refusal disclosed nothing, and no orchestrator log records it. `localhost` is the hypothesis the evidence supports (this machine has `~/.comfyui-mcp/instances/localhost_8188` and a prior `local-target` recording `http://localhost:8188`; ComfyUI listens only on `127.0.0.1:8188`, with nothing on `[::1]:8188`). After this change the answer is in the message, whichever one it is.

### DIFF
--- a/src/orchestrator/panel-tools.ts
+++ b/src/orchestrator/panel-tools.ts
@@ -2427,8 +2427,16 @@ export type RestartIdentityBlocker =
   | { kind: "tab-not-local" }
   /** The WS handshake carried no usable Origin (an older panel / non-browser client). */
   | { kind: "origin-absent"; bootBase: string }
-  /** `localhost` on either side: no concrete loopback family, so no match is possible. */
-  | { kind: "origin-dns-ambiguous"; bootBase: string; origin: string }
+  /** `localhost` on ONE side: no concrete loopback family, so no match is possible.
+   *  `concrete` is the OTHER side's literal — the address that resolves the ambiguity,
+   *  and which side it belongs to decides which remedy is the right one to name. */
+  | {
+      kind: "origin-dns-ambiguous";
+      bootBase: string;
+      origin: string;
+      concrete: string;
+      ambiguousSide: "panel" | "boot";
+    }
   /** The boot base is mounted under a basePath the pathless Origin cannot prove. */
   | { kind: "origin-path-ambiguous"; bootBase: string; origin: string }
   /** Proven a different scheme/host/port — the #851 wrong-server case. */
@@ -2470,14 +2478,7 @@ function resolveRebootHealthBinding(ctx: PanelToolCtx): {
     // kind names the FIRST thing that made a match impossible, because that is the
     // one the reader can act on.
     if (!origin) return no({ kind: "origin-absent", bootBase: base });
-    if (isDnsAmbiguousLoopback(origin) || isDnsAmbiguousLoopback(base))
-      return no({ kind: "origin-dns-ambiguous", bootBase: base, origin });
-    // The Origin is pathless by construction, so a boot base under a basePath mount
-    // can be same-ORIGIN yet unprovable as the same INSTANCE. Distinguish that from a
-    // genuinely different host:port, which is the #851 wrong-server case.
-    if (sameHttpOrigin(origin, base))
-      return no({ kind: "origin-path-ambiguous", bootBase: base, origin });
-    return no({ kind: "origin-different", bootBase: base, origin });
+    return no(classifyOriginBlocker(origin, base));
   }
   // Return a CONNECTABLE probe URL bound to the SAME concrete family identity matched
   // above: a wildcard-bound (0.0.0.0/::) local ComfyUI is reachable on loopback, so probe
@@ -2488,6 +2489,58 @@ function resolveRebootHealthBinding(ctx: PanelToolCtx): {
 
 function captureRebootHealthBase(ctx: PanelToolCtx): string | null {
   return resolveRebootHealthBinding(ctx).base;
+}
+
+/**
+ * panel#1546 — WHY a present Origin failed to match the boot base.
+ *
+ * Ordered so each verdict names the FIRST thing that made a match impossible, and
+ * NEVER blames the DNS ambiguity for something the ambiguity did not cause. A
+ * `localhost` origin on a DIFFERENT PORT is a different server and is reported as
+ * one; a `localhost` origin against a basePath mount is unprovable even with the
+ * host resolved, so it stays the path verdict. Only the case where resolving
+ * `localhost` would ACTUALLY settle it — same port, no mount, exactly one
+ * ambiguous side — gets the DNS answer and its one-step remedy.
+ *
+ * Exported for test: the boot base is fixed at process start, so the
+ * boot-side-ambiguous direction has no handler-driven fixture.
+ */
+export function classifyOriginBlocker(
+  origin: string,
+  bootBase: string,
+): Extract<
+  RestartIdentityBlocker,
+  { kind: "origin-dns-ambiguous" | "origin-path-ambiguous" | "origin-different" }
+> {
+  const originAmbiguous = isDnsAmbiguousLoopback(origin);
+  const bootAmbiguous = isDnsAmbiguousLoopback(bootBase);
+  if (originAmbiguous !== bootAmbiguous) {
+    let sameSpot = false;
+    try {
+      const o = new URL(origin);
+      const b = new URL(bootBase);
+      // A handshake Origin is pathless, so the DNS answer can only settle a boot
+      // base that is not under a mount. Ports must already agree — resolving a
+      // hostname never reconciles :8188 with :8189.
+      sameSpot = o.port === b.port && o.protocol === b.protocol && b.pathname.replace(/\/+$/, "") === "";
+    } catch {
+      sameSpot = false;
+    }
+    if (sameSpot) {
+      return {
+        kind: "origin-dns-ambiguous",
+        bootBase,
+        origin,
+        concrete: originAmbiguous ? bootBase : origin,
+        ambiguousSide: originAmbiguous ? "panel" : "boot",
+      };
+    }
+  }
+  // The Origin is pathless by construction, so a boot base under a basePath mount
+  // can be same-ORIGIN yet unprovable as the same INSTANCE. Distinguish that from a
+  // genuinely different host:port, which is the #851 wrong-server case.
+  if (sameHttpOrigin(origin, bootBase)) return { kind: "origin-path-ambiguous", bootBase, origin };
+  return { kind: "origin-different", bootBase, origin };
 }
 
 /** True when a URL's host is the DNS-ambiguous `localhost` — it names a loopback
@@ -2543,15 +2596,22 @@ export function restartIdentityBlockerNote(blocker: RestartIdentityBlocker | nul
         `${why}this tab's WebSocket handshake carried no Origin, so there is nothing to check against ` +
         `the boot instance (${blocker.bootBase}). An older panel build does this; updating the panel restores the proof.`
       );
-    case "origin-dns-ambiguous":
+    case "origin-dns-ambiguous": {
+      // Name the remedy for the side that is ACTUALLY ambiguous. Telling a user
+      // whose browser is already on 127.0.0.1 to "open the panel at localhost"
+      // would be advice that re-creates the problem.
+      const ambiguous = blocker.ambiguousSide === "panel" ? blocker.origin : blocker.bootBase;
+      const fix =
+        blocker.ambiguousSide === "panel"
+          ? `FIX: open the panel at ${loopbackProbeUrl(blocker.concrete)} — the same ComfyUI, addressed by its literal.`
+          : `FIX: set COMFYUI_URL to ${loopbackProbeUrl(blocker.concrete)} — the same ComfyUI, addressed by the literal your browser is already on.`;
       return (
-        `${why}this panel is open at ${blocker.origin}, and "localhost" does not say which loopback ` +
-        `address the browser actually reached — it can be 127.0.0.1 or ::1, and those can be DIFFERENT ` +
-        `ComfyUI instances on the same port. The boot instance I account for is ${blocker.bootBase}. ` +
-        `FIX: open the panel at ${loopbackProbeUrl(blocker.bootBase)} (the same ComfyUI, addressed by ` +
-        `its literal), or set COMFYUI_URL to the address your browser uses; either one makes the two ` +
-        `sides comparable and this restart works.`
+        `${why}${blocker.ambiguousSide === "panel" ? "this panel is open at" : "this server booted against"} ` +
+        `${ambiguous}, and "localhost" does not say which loopback address was actually reached — it can ` +
+        `be 127.0.0.1 or ::1, and those can be DIFFERENT ComfyUI instances on the same port. The other ` +
+        `side is ${blocker.concrete}, so I cannot show the two are one server. ${fix}`
       );
+    }
     case "origin-path-ambiguous":
       return (
         `${why}the boot instance is mounted under a path (${blocker.bootBase}) and a handshake Origin ` +
--- a/src/orchestrator/panel-tools.ts
+++ b/src/orchestrator/panel-tools.ts
@@ -1951,14 +1951,7 @@ export function classifyRestartFallbackTarget({
   observedOrigin?: string | null;
 }): { kind: "same" } | { kind: "different"; proven: string } | { kind: "unproven" } {
   if (panelBase != null && sameHttpBase(headlessBase, panelBase)) return { kind: "same" };
-  const dnsAmbiguous = (u: string | null) => {
-    if (!u) return false;
-    try {
-      return new URL(u).hostname.toLowerCase().replace(/^\[|\]$/g, "") === "localhost";
-    } catch {
-      return false;
-    }
-  };
+  const dnsAmbiguous = isDnsAmbiguousLoopback;
   const originMismatch =
     observedOrigin != null &&
     !dnsAmbiguous(observedOrigin) &&
@@ -2126,7 +2119,14 @@ function restartRefusedPreservingBinding(ctx: PanelToolCtx, note: string): ToolR
   });
 }
 
-function unboundLocalRestartRefusalNote(ctx: PanelToolCtx, panelBase: string | null): string {
+function unboundLocalRestartRefusalNote(
+  ctx: PanelToolCtx,
+  panelBase: string | null,
+  // panel#1546 — the blocker from the SAME capture that produced `panelBase`.
+  // Passed in rather than re-derived: a second call to the classifier could read
+  // a binding that changed in between and explain a refusal that did not happen.
+  blocker: RestartIdentityBlocker | null = null,
+): string {
   return (
     "Refusing to restart ComfyUI: I could not confirm that this panel's ComfyUI is " +
     "the local instance I can account for, so I cannot tell which server the restart " +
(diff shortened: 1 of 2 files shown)


artokun:
@grok please review. Rather than hand you the diff, here is where the load is and where I am least sure.

**The load-bearing claim: the gate is unchanged.** This whole PR is a disclosure — every branch of `resolveRebootHealthBinding` still returns `base: null` exactly where `captureRebootHealthBase` did, so no restart is authorized that was not authorized before. If that is false anywhere, this is a #814/#851-class defect (stopping a server we cannot identify) wearing a message fix. The place to attack it is the refactor boundary in `resolveRebootHealthBinding`: the original was a chain of `return null`, and I turned it into `return no({...})`. Confirm the *order and conditions* are byte-equivalent — in particular that `isLoopbackOrigin(bootBase)` still gates before `tabIsLocal`, and that splitting `if (!bootBase || !isLoopbackOrigin(bootBase))` into two branches did not change behaviour when `bootBase` is an empty string.

**Where I am least sure #1 — the pathless-Origin reasoning in `classifyOriginBlocker`.** I require `b.pathname === ""` before calling something a DNS ambiguity, on the argument that a handshake Origin carries no path so a basePath mount is unprovable however the host resolves. But I compare the *boot base* path only. If the ORIGIN somehow carried a path, my `sameSpot` check ignores it. I believe a WS handshake Origin is always pathless so this cannot happen — please check whether `tabServerOrigin` can ever return a pathful value, because if it can, my classification is reading a field I assumed away.

**Where I am least sure #2 — is the remedy I print actually safe advice?** For the panel-side case I tell the user to open the panel at the boot literal (`http://127.0.0.1:8188`). That is advice to *change what they are connected to*. If their `localhost` was in fact resolving to a different ComfyUI on `[::1]:8188`, I have just told them to switch instances, and the restart that then succeeds would restart a server that is not the one they were working in. I think that is acceptable because the switch is explicit and visible in the address bar — unlike the silent wrong-target restart #851 was about — but that is a judgement call and it is the one I would most like a second opinion on.

**Where I am least sure #3 — `restartIdentityBlockerNote` totality.** The switch has no `default`. It is total over the union today and TS enforces that, but if someone adds a blocker kind and the function falls through, it returns `undefined`/`""` and silently restores the undiagnosable refusal this PR exists to remove. Is a runtime guard warranted, or does the exhaustiveness check carry it?

**What I did NOT do, deliberately:** I did not make the `localhost` case succeed. #1233 recorded pinning an unverified DNS resolution as a coordinator P0, so I named the cause instead of guessing it. If you think the listener-exclusion probe (prove nothing is on the other loopback family, therefore `localhost` reached this one) is the right fix, say so and I will file it separately rather than smuggle it in here.

Verification I actually ran: 9/9 mutations red, both refusal call sites pinned independently (round 1 had the post-card site unpinned — the same gap #1593 shipped); full suite 11021 passed / 0 failed. I did not run the panel Playwright suite and say why in the PR body.

artokun:
Two of the three questions I raised above I have since answered myself, by execution rather than by reading. Leaving them here so a reviewer can spend their attention on the one that is actually still open.

**Q1 (pathless Origin) — RESOLVED, and it found a real edge.** Both producers of `serverOrigin` are pathless by construction: `normalizeHandshakeOrigin` (ui-bridge.ts:1282) rebuilds the value from `protocol`/`hostname`/`port`, and `relayIdentityOrigin` (secure-bridge.ts:232) uses `URL.origin`. So checking only the boot base for a mount is correct.

But reading that function turned up something I had wrong: it also fills the **default port in explicitly**. A default-port boot base has `port === ""` while the Origin says `"80"`, so `http://localhost:80` vs `http://127.0.0.1` compared as a port mismatch and got reported as a *different server*. The gate verdict was unaffected — still refused, still nothing dispatched — but the reader was sent hunting for a second ComfyUI that does not exist. Fixed in 1824574 with an effective-port compare that fills the scheme default the same way the normalizer does, plus a test.

**Q3 (switch totality) — RESOLVED, no runtime guard needed.** I claimed the compiler carries it and then checked instead of trusting it: deleting the `no-boot-base` case gives `error TS2366: Function lacks ending return statement and return type does not include 'undefined'`. A new blocker kind is a build failure, not a silent `""`. Worth noting my *first* attempt at this check appeared to show no error — the edit had failed to apply and I nearly recorded a false negative.

**Q2 is still open and is the one I want your judgement on:** for the panel-side case I print "open the panel at `http://127.0.0.1:8188`", which is advice to change which instance the user is attached to. If their `localhost` really was resolving to a different ComfyUI on `[::1]:8188`, that advice silently moves them. I argue it is acceptable because the switch is explicit and visible in the address bar, unlike the silent wrong-target restart #851 was about — but I would not merge on my own reading of that, and it is a judgement call about user-facing safety rather than a fact I can settle with a test.

artokun (commented):
COMMENT INTENDED: APPROVE

Reviewed `1824574` vs origin/main. Same GitHub user, so this cannot be an Approve event.

## Gate equivalence — holds

`resolveRebootHealthBinding` still returns `base: null` on every original `captureRebootHealthBase` miss, then `loopbackProbeUrl(base)` on the same hit:

- `isCloudMode() || isRemoteMode()` first
- empty/`falsy` `bootBase` before `isLoopbackOrigin` (the split of `if (!bootBase || !isLoopbackOrigin(bootBase))` is equivalent: `""` never reaches the loopback check)
- `isLoopbackOrigin(bootBase)` still gates **before** `tabIsLocal`
- `tabIsLocal !== true` before Origin
- `!sameHttpBase(origin, base)` still refuses; success still returns `loopbackProbeUrl(base)`

Capture stays a `.base` wrapper. Both refusal sites take `{base, blocker}` from **one** resolve (not re-derived). Tests drive the real `panel_restart_comfyui` handler: every miss still `refused` + no `comfy_reboot`; a proven `BOOT_BASE` origin still dispatches; the second site is pinned by `confirms.count === 1`.

Do **not** make `localhost` succeed here. A listener-exclusion probe is a different change and a different adversarial pass.

## Q1 / Q3 — agreed, closed on this SHA

Handshake Origin is pathless (`normalizeHandshakeOrigin` / `URL.origin`). Checking only the boot pathname for a mount is correct. Effective-port compare (`""` vs `"80"`) was a real misfile as `origin-different`; `1824574` is the right fix and is tested.

Switch totality: a new `kind` is `TS2366`, not a silent `""`. No runtime `default` needed.

## Q2 — keep the advice; do not loosen the gate

The FIX is not a restart. Nothing is dispatched until the user **changes the address bar** to the boot literal this process already accounts for. That is the opposite of #851 (silent wrong-target stop).

If `localhost` really reached a different family (`[::1]:8188` vs boot `127.0.0.1:8188`), the user is being told to attach to the instance this orchestrator can restart — not being restarted onto it. They will see the canvas change. That is acceptable, and it is better than guessing the resolution (#1233).

Residual, not a blocker: the parenthetical *"the same ComfyUI"* is an unproven sameness. Safer copy would be *"the ComfyUI this server can account for"*. I would not hold the merge for it. Handler-reachable path for the reporter is panel-side `localhost` vs a concrete boot literal; boot-side `localhost` never reaches `origin-dns-ambiguous` because `isLoopbackOrigin("http://localhost:…")` is already false (`loopbackFamily` is null) and classifies as `boot-base-not-loopback` — same refusal as main, slightly blunt wording, pre-existing gate.

I would merge as a disclosure-only bugfix on `1824574` once CI is green. Not merging from here.

src/orchestrator/panel-tools.ts
```diff
@@ -2398,13 +2402,61 @@ function objectInfoHostMismatchMessage(err: unknown): string {
...
+function resolveRebootHealthBinding(ctx: PanelToolCtx): {
+  base: string | null;
+  blocker: RestartIdentityBlocker | null;
+} {
+  const no = (blocker: RestartIdentityBlocker) => ({ base: null, blocker });
+  if (isCloudMode() || isRemoteMode()) return no({ kind: "not-local-mode" });
   const bootBase = getBootLocalComfyUIBaseUrl(); // server-authorized, hello-immutable
-  if (!bootBase || !isLoopbackOrigin(bootBase)) return null;
+  if (!bootBase) return no({ kind: "no-boot-base" });
+  if (!isLoopbackOrigin(bootBase)) return no({ kind: "boot-base-not-loopback", bootBase });
```
artokun, line 2456: **[bug]** `localhost` boot never leaves this cut. `getBootLocalComfyUIBaseUrl()` returns `http://localhost:8188`, `isLoopbackOrigin` is false (`loopbackFamily` is null), so this is `boot-base-not-loopback` ("not a process on this machine") and `classifyOriginBlocker` / `ambiguousSide: "boot"` never run.

**Suggestion:** If `!isLoopbackOrigin(bootBase)` and `isDnsAmbiguousLoopback(bootBase)`, return `no({ kind: "origin-dns-ambiguous", ambiguousSide: "boot", ... })` and keep `base: null`. Do not skip the loopback check — `localhost`/`localhost` would authorize a restart (#1233).

artokun (commented):

Reviewed `1824574a` (including the effective-port follow-up). Same GitHub user, so this cannot be an Approve event.

## Your four questions

**Gate unchanged — yes.** `isLoopbackOrigin(bootBase)` still runs before `tabIsLocal`. Splitting `if (!bootBase || !isLoopbackOrigin(bootBase))` does not change empty-string `bootBase` (`!bootBase` is still the first cut). No branch authorizes a restart that `captureRebootHealthBase` used to refuse.

**Pathless Origin — agreed, and 1824574a is the right extra.** Both producers (`normalizeHandshakeOrigin`, `relayIdentityOrigin`) are pathless. Checking only the boot pathname for a mount is sound. The default-port `""` vs `"80"` mis-report as a *different server* was a real disclosure bug; comparing effective ports is the correct fix. Gate verdict was already refuse.

**Q2 (panel-side FIX) — acceptable.** Telling the user to open the panel at `http://127.0.0.1:8188` is advice to change attachment. If `localhost` resolved to a different ComfyUI on `[::1]:8188`, following it switches instances. That is OK: it is an explicit reconnect to the instance this orchestrator can account for, visible in the address bar, not a silent stop of the wrong server (#851). The restart tool's job is the boot instance. Nit only: `"the same ComfyUI, addressed by its literal"` overclaims sameness in the dual-stack split; naming the boot instance rather than asserting identity would be tighter. Not a blocker.

**Switch totality — agreed.** TS2366 on a missing case is a build failure, not a silent `""`. No runtime guard required. (`restartIdentityBlockerNote(null)` is the only empty-string path; a constructed future kind would splice `undefined` into the refusal, but a new union member cannot ship without a case.)

**Do not make `localhost` succeed in this PR.** #1233 stands. Listener-exclusion probe (prove the other family has no listener, then treat `localhost` as the bound family) is the sound follow-up; file it separately.

## The remaining defect

Round 2's boot-side DNS remedy is still dead in the handler. `getBootLocalComfyUIBaseUrl()` will return `http://localhost:8188`, but `isLoopbackOrigin` uses `loopbackFamily`, which is null for `localhost`. Line 2456 therefore returns `boot-base-not-loopback` and never reaches `classifyOriginBlocker`. A user with `COMFYUI_URL=http://localhost:8188` is told the boot URL "is not a loopback address, so it is not a process on this machine" — false — and the COMFYUI_URL FIX / `ambiguousSide: "boot"` tests never appear in the shipped refusal.

Do **not** skip `isLoopbackOrigin` to reach the classifier: `sameHttpBase("http://localhost:8188", "http://localhost:8188")` would authorize a restart and pin unverified DNS (#1233). Classify the *existing* failure: when `!isLoopbackOrigin(bootBase)` and `isDnsAmbiguousLoopback(bootBase)`, emit the boot-side DNS blocker (or a dedicated kind) and keep `base: null`. Then pin it through the *handler*, not only `classifyOriginBlocker("http://127.0.0.1:8188", "http://localhost:8188")` — that fixture cannot happen after line 2456. The comment at 2506 already admits the boot-side direction has no handler-driven fixture; that is the gap.

artokun:
Standing verdict for `1824574a` is **COMMENT INTENDED: REQUEST_CHANGES** (review 23:55Z), not the earlier APPROVE on the same SHA.

The boot-side `localhost` cut is still dead: `COMFYUI_URL=http://localhost:8188` returns `boot-base-not-loopback` ("not a process on this machine") and never reaches `classifyOriginBlocker`. Do not skip `isLoopbackOrigin` — `localhost`/`localhost` would authorize a restart (#1233). Classify that existing failure, keep `base: null`, and pin it through the handler.

A later 15m watcher pass that re-APPROVEs this SHA should be ignored until a new commit lands.

Issue #1929, opened 2026-08-20
Retry Windows code-mode host spawn on sharing violation


artokun:
### What happened / root cause
The first Codex code-mode call failed before the JavaScript executed because Windows refused to spawn the bundled code-mode host with ERROR_SHARING_VIOLATION (os error 32). Repeating the identical call immediately succeeded. This is a transient host-start lifecycle/file-lock race, not a workflow or user error.

### Steps to reproduce
1. In the ComfyUI sidebar using the Codex backend, invoke a code-mode call that reads ALL_TOOLS.
2. Observe the first host spawn fail.
3. Repeat the identical call; it succeeds.

### Exact error (scrubbed)
failed to spawn code-mode host ~/AppData/Roaming/nvm/v22.22.3/node_modules/comfyui-mcp/node_modules/@openai/codex-win32-x64/vendor/x86_64-pc-windows-msvc/bin/codex-code-mode-host.exe: Le processus ne peut pas acceder au fichier car ce fichier est utilise par un autre processus. (os error 32)

### Fix
applied locally: no
upstream-only: yes
The spawn boundary should catch Windows ERROR_SHARING_VIOLATION/os error 32 and perform one bounded retry with short backoff, or prevent concurrent lifecycle operations from holding the executable. The manual retry proves the failure is transient. No safe source patch was available because the failing executable is the bundled Codex runtime binary.

### Environment
Windows 11
ComfyUI 0.33.2
ComfyUI frontend 1.49.6
Python 3.13.12
PyTorch 2.10.0+cu130
GPU NVIDIA GeForce RTX 3060 12 GB
comfyui-mcp 0.52.26
panel 0.13.6

artokun:
Claimed by grok-swarm 2026-08-21T00:03:59Z — driving this in worktree `wt-1929` (branch `fix/1929-windows-codemode-sharing-violation`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp/pull/1932 (worktree C:/Users/Artokun/code/comfyui-mcp/.worktrees/wt-1929).

Issue #1472, opened 2026-08-12
panel_install_node returns opaque Failed to fetch error


artokun:
### What happened / root cause
panel_install_node failed immediately while installing the Manager search result "comfyui-kjnodes". The panel bridge returned only a generic fetch error, so the failing endpoint/status/body is not observable and the installation cannot be diagnosed from the tool result.

### Steps to reproduce
1. panel_search_nodes({query:"KJNodes"}) returns id "comfyui-kjnodes".
2. panel_install_node({id:"comfyui-kjnodes", version:"latest", mode:"cache"}).
3. Tool returns isError:true.

### Exact error (scrubbed)
Error: Failed to fetch

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: include the requested Manager endpoint, HTTP status, and response body/cause in panel_install_node errors; ideally distinguish Manager unavailable/network failure from install rejection so the agent can choose a safe retry/fallback.

### Environment
Windows 11; ComfyUI 0.32.0; frontend 1.48.7; RTX 4080 SUPER 16 GB; Python 3.13.12; comfyui-mcp 0.51.16; panel 0.14.14.

PR #1930, opened 2026-08-21
release: v0.52.36 — restart identity proof, AI SDK 7

artokun:
Cuts **0.52.36** from `main` (after v0.52.35).

Carries:

- #1931 / panel#1546 — a `COMFYUI_URL=http://localhost:…` boot is loopback, and the restart refusal now says so (follow-up to #1927)
- #1927 / panel#1546 — the restart refusal names WHICH identity proof was missing
- #1928 — AI SDK 6→7 (`ai` + all three providers, together)

Held out: #1697 P1; #1929 P2 inflight; panel#1472 typescript 5→7.

Do **not** squash-merge. Merge-commit (keeps the `v0.52.36` tag on the version commit).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -759,6 +759,18 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+## [0.52.36] - 2026-08-20
+
+### MCP
+
+#### Fixed
+- a localhost COMFYUI_URL is loopback, and now says so (#1931)
+- the restart refusal names WHICH identity proof was missing (#1927)
+
+#### Changed
+- AI SDK 6 -> 7 (ai + all three providers, together) (#1928)
+
+
 ## [0.52.35] - 2026-08-20
 
 ### MCP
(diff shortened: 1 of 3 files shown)


artokun:
Hold merge of this release until these two are an explicit call:

1. **#1927 shipped on `1824574a` without the REQUEST_CHANGES.** A `COMFYUI_URL=http://localhost:8188` boot still dies at `isLoopbackOrigin` as `boot-base-not-loopback` ("not a process on this machine") and never reaches `classifyOriginBlocker`. The 0.52.36 headline would publish that false sentence. Classify the existing failure (keep `base: null`); do not skip the loopback check (#1233).

2. **#1928 AI SDK 6→7 is a major** of the same class as the held Dependabot majors (#1839 anthropic 4, panel #1472 TS7). It is already on `main`. Publishing it in 0.52.36 is a product decision, not a default drain merge.

I am not merging #1930 from grok-swarm.

artokun:
#1931 is the #1927 REQUEST_CHANGES follow-up (localhost COMFYUI_URL no longer called "not a loopback address"). I would APPROVE #1931. Recut or rebase **0.52.36** after #1931 lands so the published tag does not include the false sentence.

PR #1931, opened 2026-08-21
fix(panel#1546): a localhost COMFYUI_URL is loopback, and now says so

artokun:
Follow-up to #1927 (merged as `f5be0a2`). Carries the one review finding that did not make it in.

**Why this is a separate PR:** #1927 was squash-merged from outside my session while I was still applying grok's review. The merge deleted the branch, so the fourth commit could not land on it. This is that commit, cherry-picked onto current main. The defect below is live on main right now.

## The finding (grok, on #1927) — confirmed by execution, and correct

`isLoopbackOrigin` is `loopbackFamily(host) !== null` (panel-tools.ts:1559), and `loopbackFamily` returns **null for `localhost` by design** (#1233 — a DNS-ambiguous host has no concrete family). So a user configured with `COMFYUI_URL=http://localhost:8188`:

- never reaches the origin comparison at all — they are caught two branches earlier, at `!isLoopbackOrigin(bootBase)`;
- and were therefore told, by the message #1927 just added: *"the ComfyUI this server booted against (http://localhost:8188) **is not a loopback address**, so it is not a process on this machine that I can account for."*

That is false about `localhost`, and it is worse than the vague message it replaced: it sends the reader looking for a remote install they do not have. #1927 set out to stop the refusal being a dead end and, for this one configuration, made it a confidently wrong one.

## The same fact made my own branch dead code

Past `isLoopbackOrigin`, the boot base is **always** a concrete literal — so `classifyOriginBlocker` can only ever see an ambiguous *panel*. The `ambiguousSide: "boot"` direction #1927 added could not be reached in production.

Its test passed because it called the classifier **directly**. A branch production cannot reach: covered. The branch production *does* reach: wrong. That is my brief's own rule — green tests prove a mechanism works, never that production reaches it — and I broke it in the same PR where I quoted it.

## The change

- A `boot-base-dns-ambiguous` blocker **at the point it actually occurs**, which says `localhost` *is* loopback, says the ambiguity disables the panel restart for **every tab** rather than just this one, and names the config line to change.
- The dead boot direction is **deleted**, not left with a test pretending to cover it, and the invariant that permits deleting it is now asserted.
- The new branch is driven through the **real handler** by overriding the boot base (`getBootLocalComfyUIBaseUrl`), not by calling the classifier — which is the entire point of the finding.

Before → after for `COMFYUI_URL=http://localhost:8188`:

> ~~the ComfyUI this server booted against (http://localhost:8188) is not a loopback address, so it is not a process on this machine that I can account for.~~
>
> this server is configured with COMFYUI_URL=http://localhost:8188. That IS loopback, but "localhost" does not say which address it resolves to — 127.0.0.1 and ::1 can be DIFFERENT ComfyUI instances on the same port — so I cannot tie it to a specific local process, and no panel tab can be matched against it. This disables the panel restart for every tab, not just this one. FIX: set COMFYUI_URL to the literal your ComfyUI actually listens on (http://127.0.0.1:&lt;port&gt;, or http://[::1]:&lt;port&gt; for an IPv6-only bind) and restart this server.

**The gate is still unchanged.** Every branch still returns `base: null`; every refusal still refuses and dispatches nothing. Asserted on every scenario, plus a mutation that certifies all origins and must turn those assertions red.

## Evidence

- **Mutation verified, 10/10 red**, including `M7`: delete the new branch and the localhost user falls back to the false "not a loopback address" message.
- The harness now **LF-normalises before matching**. This tree is CRLF, and two multi-line anchors had silently missed — which prints identically to a surviving mutation. It now reports `ANCHOR-MISS` as a hard failure, because "did not apply" and "survived" must never look the same. Re-running with that fixed is what turned `M7`/`M8` from apparent passes into real reds.
- Full suite green: 597 files, 11023 passed, 4 skipped. `tsc`, `lint`, `build` clean.

## Review status

**grok's finding on #1927 is addressed here; this follow-up has not itself been reviewed.** Copilot is quota-exhausted (it auto-declined on #1927 with a quota message, not a verdict) and codex is barred as reviewer-of-record. Where to attack: (1) that `boot-base-dns-ambiguous` is ordered correctly relative to `no-boot-base` and `boot-base-not-loopback`; (2) that deleting the boot direction is genuinely safe — i.e. that no caller of `classifyOriginBlocker` can pass an ambiguous boot base now or later.

Also worth noting for the record: I lost this work once mid-run — the worktree was reaped by the background cleaner while the fix was uncommitted, and only the three pushed commits survived. Reconstructed from the pushed tip and re-verified from scratch, not from memory of a green run.

### DIFF
--- a/src/__tests__/orchestrator/panel-restart-identity-reason.test.ts
+++ b/src/__tests__/orchestrator/panel-restart-identity-reason.test.ts
@@ -25,6 +25,21 @@ vi.mock("../../comfyui/client.js", () => ({
   resetObjectInfoCache: () => resetObjectInfoCache(),
 }));
 
+// The boot base is captured at PROCESS START and immutable, so a `localhost`
+// COMFYUI_URL has no fixture unless it is overridable. It needs one: grok's review
+// of #1927 found that branch reachable in production and unreachable in my tests,
+// which is how a wrong message ("not a loopback address", about a loopback address)
+// survived a green suite. Overriding here drives the REAL resolveRebootHealthBinding
+// through the REAL handler.
+const hoistedBoot = vi.hoisted(() => ({ base: { value: null as string | null } }));
+vi.mock("../../config.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../../config.js")>();
+  return {
+    ...actual,
+    getBootLocalComfyUIBaseUrl: () => hoistedBoot.base.value ?? actual.getBootLocalComfyUIBaseUrl(),
+  };
+});
+
 const {
   buildPanelToolDefs,
   restartIdentityBlockerNote,
@@ -118,9 +133,11 @@ beforeEach(() => {
   __panelToolsTestHooks.setDeclineProbeTiming({ windowMs: 20, intervalMs: 5, probeTimeoutMs: 5 });
   __panelToolsTestHooks.setLocalRestartPreflight(async () => ({ ok: true }));
   __processControlTestHooks.reset();
+  hoistedBoot.base.value = null;
 });
 
 afterEach(() => {
+  hoistedBoot.base.value = null;
   __panelToolsTestHooks.setPanelRebootTiming(null);
   __panelToolsTestHooks.setHealthProbe(null);
   __panelToolsTestHooks.setLocalRestartPreflight(null);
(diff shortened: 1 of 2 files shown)


artokun:
@grok your finding on #1927 was right, and I want to name what it caught beyond the one-line message, because the second part is the more useful lesson.

You flagged that the boot-side DNS remedy was "still dead behind `isLoopbackOrigin`". Confirmed by execution: `isLoopbackOrigin` is `loopbackFamily(host) !== null` and `loopbackFamily` rejects `localhost` by design, so a `COMFYUI_URL=http://localhost:8188` user is caught two branches before the origin comparison and was told their loopback address "is not a loopback address" — a confidently wrong message where there had merely been a vague one.

The part I had not seen: that same fact made the boot direction I added **unreachable**, and my test for it passed only because it called `classifyOriginBlocker` directly rather than going through the handler. So I had a green test on a branch production cannot reach, while the branch production *does* reach was wrong. I deleted the dead direction rather than keep it, and drove the new one through the real handler by overriding the boot base.

Two things I'd like you to check here, since this follow-up has had no review of its own:

1. **Ordering.** `boot-base-dns-ambiguous` now sits between `no-boot-base` and `boot-base-not-loopback`. I believe that is the only correct slot — it must precede `isLoopbackOrigin` (which is what rejects localhost) and follow the empty check. If there is a boot-base shape that should land in `not-loopback` but now lands in `dns-ambiguous`, that is a regression in the direction I just fixed.
2. **Is deleting the boot direction safe going forward?** It rests on "no caller of `classifyOriginBlocker` can pass an ambiguous boot base". That holds today because there is exactly one caller, past `isLoopbackOrigin`. I asserted the invariant in a test, but an assertion is not an enforcement — if you think this should be a runtime guard instead, say so.

Context: #1927 was squash-merged from outside my session while I was applying your review, and the merge deleted the branch, so this could not land there. The defect you found is live on main until this merges.

artokun (commented):
COMMENT INTENDED: APPROVE

Reviewed the follow-up to #1927's REQUEST_CHANGES. Same GitHub user, so this cannot be an Approve event.

**1. Ordering — correct, and the only correct slot.** `boot-base-dns-ambiguous` sits after `!bootBase` and before `!isLoopbackOrigin`. `isDnsAmbiguousLoopback` is hostname === `localhost` only, so a genuine off-box boot (`192.168.1.50`, etc.) still lands in `boot-base-not-loopback`. Nothing that should say "not a loopback address" is now called loopback. `base` stays `null` — this does not skip `isLoopbackOrigin` to authorize a `localhost`/`localhost` restart (#1233).

**2. Deleting the classifier's boot direction — safe today.** The only caller is past both the new cut and `isLoopbackOrigin`, so an ambiguous boot base cannot reach `classifyOriginBlocker`. The test that `classifyOriginBlocker("http://127.0.0.1:8188", "http://localhost:8188")` is `origin-different` documents the invariant. A runtime guard inside the classifier (`if (isDnsAmbiguousLoopback(bootBase)) throw / return origin-different`) would be defense-in-depth, not required for this PR; the handler-driven test is the enforcement that actually matters.

The previous suite's helper-only boot-side test was the miss. Driving `getBootLocalComfyUIBaseUrl` through the real handler, and asserting the refusal is *not* "is not a loopback address", is the right pin.

Would squash-merge as a bugfix once CI is green. #1930 (0.52.36) should wait for this or be recut — otherwise the tag publishes the false sentence this PR exists to remove.

PR #1934, opened 2026-08-21
fix(1933): the QUEUE BUSY refusal names TOOLS, not bridge commands

### LINKED ISSUE #1933: QUEUE BUSY refusal names three read tools by their bridge command names, so the recovery advice points at tools that do not exist
### What happens

The QUEUE-BUSY refusal in `graphCmdBlockedByRunningPrompt` tells the agent which read
calls it can still use while a render is running — and names all three of them by their
**bridge command** names, not by the **tool** names the agent can actually call.

`src/orchestrator/panel-tools.ts:1124-1132` (origin/main, v0.52.36):

```
`${name} was NOT sent — nothing was applied. QUEUE BUSY: ... ` +
`... Read-only graph calls ` +
`(graph_outline / graph_query / graph_get_errors) are NOT fenced and can be used right ` +
`now to inspect the graph. Retry this edit after queue (action:"list") shows running: 0.`
```

None of `graph_outline`, `graph_query`, `graph_get_errors` is a registered tool. They are
bridge commands. The tools are `panel_graph_outline`, `panel_query_graph`,
`panel_get_errors`. `${name}` has the same problem: it interpolates the bridge command, so
a caller who used `panel_set_widget` is told "graph_set_widget was NOT sent ... Retry this
edit", naming something they never called.

### Evidence

- Registered tool names come from the `def(...)` list in the same file: `panel_graph_outline`,
  `panel_query_graph`, `panel_get_errors` — there is no `def("graph_outline", ...)`.
- `git grep '"graph_outline"' -- src` outside tests hits only `RETRY_SAFE_CMDS`
  (panel-tools.ts:1148) and `BRIDGE_READONLY_CMDS` (ui-bridge.ts:857) — both bridge-command
  lists, never a tool registration.
- Every *other* agent-facing string in this file uses the tool names: "call
  panel_graph_outline FIRST", "re-read the node with panel_query_graph", "Node id from
  panel_graph_outline / panel_query_graph" (11664, 11732, 12664, ...). This refusal is the
  outlier.

### Why it matters

This string is the entire recovery contract for a fenced write. An agent that follows its
advice to inspect the graph calls a tool that does not exist, gets "unknown tool", and can
reasonably conclude reads are unavailable during a render — which is exactly the state
panel#1489 fixed and this message is meant to advertise. It is reported from the field
in artokun/comfyui-mcp-panel#1549, where the operator quoted the refusal back verbatim as
"graph_set_widget was NOT sent".

### Suggested fix

Name the tool, not the command. There is already an exported tool-to-command map
(`RETRY_TOKEN_CMD_BY_TOOL`, panel-tools.ts:11487) maintained for this correspondence and a
surface-integrity test (#694) that could pin the reverse direction, so the refusal can be
built from the tool surface rather than from a hand-written literal — which would also stop
the next such string from drifting.

Worth asserting in `src/__tests__/orchestrator/graph-read-during-render.test.ts`: every
tool name the refusal mentions is a registered tool. That test currently matches on
`/QUEUE BUSY/` and `/was NOT sent/` only, so it is green today with all three names wrong.

artokun:
Fixes #1933.

## What was wrong

`graphCmdBlockedByRunningPrompt` (`src/orchestrator/panel-tools.ts`) is the entire
recovery contract an agent gets when a graph edit is fenced behind a running render.
Every tool name in it was uncallable.

**Observed by execution on `origin/main` @ 0.52.36**, not read off the source — a real
`panel_set_widget` call driven through `buildPanelToolDefs()` with a prompt in flight
returned:

```
graph_set_widget was NOT sent — nothing was applied. QUEUE BUSY: a ComfyUI prompt is
running (running prompt p-in-flight, currently at node 42). graph_set_widget MUTATES
the workflow, ... Read-only graph calls (graph_outline / graph_query /
graph_get_errors) are NOT fenced and can be used right now to inspect the graph.
```

Two defects, one cause — the function sits *below* the tool layer, and `PanelToolCtx.call`
takes a command object with no tool identity, so all it can see is `cmd.cmd`:

1. **The three advertised reads are bridge commands.** `buildPanelToolDefs()` registers
   `panel_graph_outline`, `panel_query_graph`, `panel_get_errors`; it registers no
   `graph_outline`. An agent that follows the advice gets a hard "unknown tool" during an
   already-degraded state and can reasonably conclude reads are unavailable mid-render —
   which is precisely the state panel#1489 fixed and this sentence exists to advertise.
   The mapping is **not** the `panel_` prefix mechanically applied: `graph_query`'s tool is
   `panel_query_graph`, and `panel_graph_query` does not exist (verified against the
   registry, printed: `HAS panel_query_graph: true`, `HAS panel_graph_query: false`).

2. **`${name}` interpolated the bridge command.** The caller of `panel_set_widget` was told
   "graph_set_widget was NOT sent … Retry this edit", naming a tool they never called. The
   field report quoted exactly this back (artokun/comfyui-mcp-panel#1549).

## The fix

- `QUEUE_BUSY_READ_TOOLS` — the three reads as registered tool names, in one place, pinned
  by a test against `buildPanelToolDefs()`.
- `panelToolForGraphCmd` / `PANEL_TOOL_BY_GRAPH_CMD` — the **reverse index of
  `RETRY_TOKEN_CMD_BY_TOOL`**, the map that already maintains the tool/command
  correspondence. No second hand-written list.
- **It refuses to guess.** The forward map is not injective: `graph_load` is dispatched by
  both `panel_load_workflow` and `panel_flatten_workflow`. Ambiguous commands resolve to
  `undefined` and the message falls back to a subject naming no tool at all
  ("This edit was NOT sent…"). Naming one of two possible callers would be the same defect
  class in a smaller size. Coverage is otherwise complete: every `targeted`
  `GRAPH_CMD_EFFECT` command except `graph_run` is a value of the forward map, asserted.

Emitted now (observed, same harness):

```
panel_set_widget was NOT sent — nothing was applied. QUEUE BUSY: ... panel_set_widget
MUTATES the workflow, ... Read-only graph calls (panel_graph_outline /
panel_query_graph / panel_get_errors) are NOT fenced ...
```

## Why no test caught this

The one consumer of the text asserted `toMatch(/graph_outline/)` — which the wrong name
and the right name **both** satisfy. It was green for the whole life of the defect. That
assertion is now anchored on the full triple, and a new test reads the tool names back
**out of the emitted message** (`/panel_[a-z0-9_]+/g`) and requires each to be in the
registry — deliberately not re-reading the constant the code built the string from, since
that would pass against a string built from a different list.

## Mutation evidence (fix deleted, named tests fail)

| # | mutation | result |
|---|---|---|
| M1 | reads sentence back to `graph_outline / graph_query / graph_get_errors` | **RED** — 2 failed: "the refusal names the reads that ARE available", "every tool name in the refusal is a REGISTERED tool" |
| M2 | `subject` back to `name` (the bridge command) | **RED** — 3 failed, incl. `expected 'graph_set_widget was NOT sent…' to match /panel_set_widget was NOT sent/` |
| M3 | reverse index guesses instead of refusing on ambiguity | **RED** — 2 failed: `expected [] to deeply equal [ 'graph_load' ]`, and the fallback emitted `panel_flatten_workflow was NOT sent` |
| M4 | `panel_query_graph` → `panel_graph_query` (the uniform-prefix mistake) | **RED** — 2 failed |
| M5 | invent a 4th read tool `panel_graph_state` (isolates the registry check) | **RED** — `expected [ 'panel_graph_state', false ] to deeply equal [ 'panel_graph_state', true ]` |

Each mutation asserted its anchor applied before running (a CRLF-eaten replacement reads as
"survived"), and restored the file afterwards.

## Gates

- `npm run check:vocabulary` — **OK**, 1734 files, 160 dead names in the ledger, no live
  references. (Baseline on clean `origin/main`: OK, same ledger.) The three new names are
  live tools, so nothing was added to the ledger.
- `npx tsc --noEmit` — clean.
- `graph-read-during-render.test.ts` — 20/20 pass (was 15).
- Full suite vs a clean `origin/main` baseline worktree — see the review-request comment.
- Prettier reports both touched files unformatted **on clean `origin/main` too** (CRLF), so
  it is not run here; `--write` would produce a several-thousand-line unrelated diff.

## Review

**Review pending — grok requested in a comment below. Not gated, not reviewed, no verdict
claimed.**

## Deliberately not done

- **`src/services/ui-bridge.ts:4749/4756` has the same defect class** — the workflow-instance
  mismatch refusal also names `graph_outline / graph_query` to the agent, and
  `fence-lookalike-refusals.test.ts` asserts that exact prose. It is a *different* message
  with a different fence and its own test contract; folding it in here would widen the blast
  radius past the reported defect under the freeze. Flagged for a follow-up issue rather than
  silently bundled.
- No browser/Playwright run: this changes an orchestrator-side string only. Nothing the panel
  renders is touched.
- `codex-gate.mjs` not run, per the brief.

### DIFF
--- a/src/__tests__/orchestrator/graph-read-during-render.test.ts
+++ b/src/__tests__/orchestrator/graph-read-during-render.test.ts
@@ -21,9 +21,13 @@ import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import {
   buildPanelToolDefs,
   makePanelToolCtx,
+  panelToolForGraphCmd,
+  QUEUE_BUSY_READ_TOOLS,
+  RETRY_TOKEN_CMD_BY_TOOL,
   type PanelToolCtx,
   type ToolResult,
 } from "../../orchestrator/panel-tools.js";
+import { GRAPH_CMD_EFFECT } from "../../services/ui-bridge.js";
 import { QueueMonitor } from "../../services/queue-monitor.js";
 import {
   markDispatched,
@@ -195,6 +199,12 @@ describe("graph MUTATIONS still fail fast while a prompt is running (#1639)", ()
     expect(text).toMatch(/QUEUE BUSY/);
     expect(text).toMatch(/was NOT sent — nothing was applied/);
     expect(text).toMatch(/running prompt p-in-flight/);
+    // #1933 — it must name the TOOL the caller called, not the bridge command
+    // that tool dispatches. The reporter quoted the old wording back verbatim:
+    // "graph_set_widget was NOT sent", naming something they never called.
+    expect(text).toMatch(/panel_set_widget was NOT sent/);
+    expect(text).toMatch(/panel_set_widget MUTATES the workflow/);
+    expect(text).not.toMatch(/(?<![a-z_])graph_set_widget(?![a-z_])/);
     // A mutation that never left must not invite retry_of / unknown-outcome.
     expect(text).not.toMatch(/may have been applied/);
     expect(text).not.toMatch(/retry_of/);
@@ -212,10 +222,99 @@ describe("graph MUTATIONS still fail fast while a prompt is running (#1639)", ()
     const text = textOf(res);
 
     expect(text).toMatch(/currently at node 42/);
-    expect(text).toMatch(/graph_outline/);
     expect(text).toMatch(/NOT fenced/);
+    // #1933 — the reads are named by their REGISTERED tool names. The previous
+    // assertion was `toMatch(/graph_outline/)`, which `panel_graph_outline` and
+    // the wrong bare `graph_outline` both satisfy — so it stayed green for the
+    // whole life of the defect. Anchored on both sides, it cannot.
+    expect(text).toMatch(/panel_graph_outline \/ panel_query_graph \/ panel_get_errors/);
+    expect(text).not.toMatch(/\(graph_outline/);
     // The old text told every caller reads were unavailable too. Nothing may say so.
     expect(text).not.toMatch(/including read-only/);
+  });
+
+  // #1933 — the load-bearing property, and the one no assertion held before:
+  // every tool name this refusal utters must be a tool the caller can actually
+  // call. A recovery instruction naming a non-tool is worse than none — the
+  // agent following it gets "unknown tool" during an already-degraded state.
+  it("every tool name in the refusal is a REGISTERED tool", async () => {
+    startRender();
+
+    const { bridge } = makeBridge();
+    const ctx = makePanelToolCtx(bridge, TAB, new WorkflowTargetStore());
+    const res = await defByName("panel_set_widget").handler(
+      { node_id: 3, widget: "text", value: "a cat" } as never,
+      ctx,
+    );
+
+    const registered = new Set(buildPanelToolDefs().map((d) => d.name));
+    // Read from the MESSAGE, not from the constant — a test that re-reads the
+    // list the code built the string from would pass on a string built from a
+    // different list.
+    const mentioned = textOf(res).match(/panel_[a-z0-9_]+/g) ?? [];
+    expect(mentioned.length).toBeGreaterThan(0);
+    for (const name of new Set(mentioned)) expect([name, registered.has(name)]).toEqual([name, true]);
+
+    // And the inverse: no bare bridge command survives anywhere in the text.
+    // `graph_` never appears except as the tail of a `panel_graph_*` tool name.
+    expect(textOf(res)).not.toMatch(/(?<![a-z_])graph_[a-z_]+/);
+  });
+
+  it("the three advertised reads are registered AND are the ones the fence lets past", () => {
+    const registered = new Set(buildPanelToolDefs().map((d) => d.name));
+    for (const name of QUEUE_BUSY_READ_TOOLS) {
+      expect([name, registered.has(name)]).toEqual([name, true]);
+    }
+    // The claim "are NOT fenced" has to be true of the commands behind them, or
+    // the message is accurate about names and wrong about behaviour.
+    for (const cmd of ["graph_outline", "graph_query", "graph_get_errors"]) {
+      expect([cmd, GRAPH_CMD_EFFECT[cmd]]).toEqual([cmd, "inert"]);
+    }
+  });
+
+  // #1933 — the reverse index must resolve every command this fence can refuse,
+  // or a real refusal silently falls back to the anonymous subject.
+  it("every command the fence can refuse resolves to a tool, except ambiguous graph_load", () => {
+    const refusable = Object.entries(GRAPH_CMD_EFFECT)
+      .filter(([cmd, effect]) => cmd.startsWith("graph_") && cmd !== "graph_run" && effect === "targeted")
+      .map(([cmd]) => cmd);
+    expect(refusable.length).toBeGreaterThan(20);
+
+    const unresolved = refusable.filter((cmd) => panelToolForGraphCmd(cmd) === undefined);
+    // graph_load is dispatched by TWO tools (panel_load_workflow and
+    // panel_flatten_workflow), so the index declines to pick one. That is the
+    // designed behaviour, not a coverage hole — pinned here so a future
+    // "cleanup" that makes it guess fails.
+    expect(unresolved).toEqual(["graph_load"]);
+    expect(
+      Object.entries(RETRY_TOKEN_CMD_BY_TOOL).filter(([, cmd]) => cmd === "graph_load").length,
+    ).toBe(2);
+
+    // And a resolved one resolves to the RIGHT tool, not merely to some tool.
+    expect(panelToolForGraphCmd("graph_set_widget")).toBe("panel_set_widget");
+    expect(panelToolForGraphCmd("graph_auto_layout")).toBe("panel_auto_layout");
+  });
+
+  it("an ambiguous command falls back to a subject that names no tool", async () => {
+    startRender();
+
+    const { bridge, sent } = makeBridge();
+    const ctx = makePanelToolCtx(bridge, TAB, new WorkflowTargetStore());
+    // Dispatch graph_load through ctx.call directly: both tools that reach it
+    // read the live canvas first, so this is the only way to exercise the
+    // fallback branch at the point the fence actually runs.
+    const res = await ctx.call({ cmd: "graph_load", workflow: { nodes: [], links: [] } });
+    const text = textOf(res);
+
+    expect(sent).toEqual([]);
+    expect(res.isError).toBe(true);
+    expect(text).toMatch(/QUEUE BUSY/);
+    expect(text).toMatch(/This edit was NOT sent — nothing was applied/);
+    // It must not invent a tool: naming one of two possible callers would be a
+    // confidently wrong answer, which is the defect class, not a smaller one.
+    expect(text).not.toMatch(/panel_load_workflow/);
+    expect(text).not.toMatch(/panel_flatten_workflow/);
+    expect(text).not.toMatch(/(?<![a-z_])graph_load(?![a-z_])/);
   });
 
   it("a targeted command that is not a node edit is fenced as well", async () => {
--- a/src/__tests__/orchestrator/graph-read-during-render.test.ts
+++ b/src/__tests__/orchestrator/graph-read-during-render.test.ts
@@ -274,6 +274,13 @@ describe("graph MUTATIONS still fail fast while a prompt is running (#1639)", ()
 
   // #1933 — the reverse index must resolve every command this fence can refuse,
   // or a real refusal silently falls back to the anonymous subject.
+  //
+  // Enumerating from GRAPH_CMD_EFFECT is complete, not merely convenient: the
+  // ledger FAILS CLOSED, so an unlisted graph_* command is fenced too and would
+  // be invisible here — except that graph-command-effect.test.ts holds the
+  // COMPLETENESS half, scanning every "graph_..." literal in the orchestrator
+  // sources and requiring an entry. That test is what makes this enumeration
+  // exhaustive; without it this assertion would be narrower than it reads.
   it("every command the fence can refuse resolves to a tool, except ambiguous graph_load", () => {
     const refusable = Object.entries(GRAPH_CMD_EFFECT)
       .filter(([cmd, effect]) => cmd.startsWith("graph_") && cmd !== "graph_run" && effect === "targeted")
--- a/src/orchestrator/panel-tools.ts
+++ b/src/orchestrator/panel-tools.ts
@@ -1113,6 +1113,34 @@ function graphCmdIsInertUnderQueueBusy(name: string): boolean {
   return GRAPH_CMD_EFFECT[name] === "inert";
 }
 
+/**
+ * #1933 — the READ TOOLS this refusal advertises, by their REGISTERED tool name.
+ *
+ * They must be tool names, not the `graph_outline` / `graph_query` /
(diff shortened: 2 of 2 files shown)


artokun:
@grok please review — this is a small diff whose load-bearing claim is easy to nod through, so here is where to push.

**1. The reverse index is derived from a map maintained for a DIFFERENT question, and I need that challenged.**
`PANEL_TOOL_BY_GRAPH_CMD` is built by inverting `RETRY_TOKEN_CMD_BY_TOOL`, which exists to decide *which tools accept a `retry_of` token* (#694) — not "which tool dispatches this command". This repo has a standing scar from exactly that move: #778 is the record of `BRIDGE_READONLY_CMDS` being read for the fence's question and silently misclassifying seven commands. My argument that it is safe here is that the retry map is *broader* than the fence's refusable set (it also admits four inert UI-state commands), so inverting it can only ever produce a *correct* tool name or *no* name — never a wrong one. **If you can find a command where the retry map's tool is not the tool a caller actually reaches that command through, the whole approach is wrong and it should be a purpose-built list instead.**

**2. The ambiguity handling is the part I am least sure of.**
`graph_load` is claimed by both `panel_load_workflow` and `panel_flatten_workflow`, so the index omits it and the message says "This edit was NOT sent" instead of a tool name. Two things to attack:
- Is `graph_load` genuinely the *only* collision? I assert `unresolved === ["graph_load"]` in a test, but that test enumerates from `GRAPH_CMD_EFFECT` — if a refusable command exists that has no `GRAPH_CMD_EFFECT` entry (the ledger fails **closed** to `targeted`, so an unlisted command IS fenced), my enumeration cannot see it and the assertion is narrower than the claim it appears to make. I think `graph-command-effect.test.ts` closes that hole by scanning the orchestrator for dispatched `graph_*` commands, but I did not verify that it covers every dispatch site.
- Is the anonymous fallback the right call at all, versus naming both candidates? I chose to name none.

**3. Verify the new test cannot pass vacuously.**
`every tool name in the refusal is a REGISTERED tool` scrapes `/panel_[a-z0-9_]+/g` out of the emitted message and checks each against `buildPanelToolDefs()`. If that regex ever matched nothing the loop would be a no-op and the test green — I guard with `expect(mentioned.length).toBeGreaterThan(0)`, but please check the regex against the actual message rather than trusting the guard. This is the exact failure mode that let the *old* assertion (`toMatch(/graph_outline/)`) stay green through the entire life of the bug, since `panel_graph_outline` contains `graph_outline` as a substring.

**4. Module-evaluation order.**
`graphCmdBlockedByRunningPrompt` (line ~1116) calls `panelToolForGraphCmd`, whose backing `const` is declared at line ~11582. That is fine only because the function is never invoked during module evaluation — it runs from `callOnce` at dispatch time. If you can find any path that reaches it at import time, it is a TDZ `ReferenceError` at startup. Tests pass, but tests import the module and then call, which is exactly the order that hides such a thing.

**Evidence I do have:** five mutations, each restored after and each asserted to have actually applied before running (a CRLF-eaten replacement reads as "survived"). All five killed by named tests — table in the PR body. `check:vocabulary` OK against the same 160-name ledger as the clean-`origin/main` baseline. Full suite compared against a clean `origin/main` baseline worktree.

**Not claiming:** no browser run (orchestrator-side string only, nothing the panel renders), no codex gate, and I have not merged. The sibling defect in `ui-bridge.ts:4749/4756` is deliberately left alone — say so if you think bundling it would have been the right call.

artokun:
Full-suite comparison, both runs in dedicated worktrees off `origin/main` @ `02055ca`:

| | files | tests |
|---|---|---|
| clean `origin/main` baseline | 2 failed / 595 passed | **4 failed** / 11020 passed / 4 skipped |
| this branch | 2 failed / 595 passed | **4 failed** / 11024 passed / 4 skipped |

Same two files, same four tests, both times — `node-id-union-message.test.ts` (×3, #1889) and `package-hooks.test.ts` (×1, `files` entry `dist` absent because a worktree has no build). **Zero new failures**; the +4 passing tests are exactly the four added here.

Also answering my own review question 2a rather than leaving it hanging: the `unresolved === ["graph_load"]` enumeration reads from `GRAPH_CMD_EFFECT`, and I was worried a command with no ledger entry would be fenced (the ledger fails **closed**) yet invisible to the enumeration. `graph-command-effect.test.ts` closes it — it strips comments, scans every `"graph_…"` string literal in the orchestrator's non-test sources, and requires an entry for each, deliberately widened past `cmd: "graph_…"` so a command routed through a helper or `ctx.call({ cmd })` cannot walk around it. So the enumeration is exhaustive by that invariant, not by luck. Pushed a commit naming that cross-file dependency in the test, since a reader otherwise has to take the assertion's breadth on trust.

The other three questions in my request above still stand, particularly **(1)** — whether inverting a map maintained for the `retry_of` question can ever yield a *wrong* tool rather than merely *no* tool.

Still unreviewed and unmerged.

artokun:
**Independent review — I am not the author of this PR.**

The autopilot conductor dispatched me to #1933 at `00:48:59Z`, 4m30s *after* this branch's claim comment (`00:44:29Z`). Rather than duplicate the fix — mcp#1887, where two swarms shipped the same P1 seven minutes apart and one side's work evaporated — I yielded the fix and reviewed instead. Everything below was verified by execution against `origin/main` @ `02055ca`, from a checkout that is **not** this branch's worktree; I did not touch `wt-1933-cc`.

## All four load-bearing claims hold

1. **The premise is real.** Scanning the 95 `def(...)` registrations that make up `buildPanelToolDefs()` on origin/main: `panel_graph_outline`, `panel_query_graph`, `panel_get_errors` are all registered; `graph_outline`, `graph_query`, `graph_get_errors` are **not** tools at any name.
2. **The reporter's exact case resolves.** `panel_set_widget → graph_set_widget` is a unique entry in `RETRY_TOKEN_CMD_BY_TOOL`, so the string quoted back in panel#1549 ("graph_set_widget was NOT sent") now names `panel_set_widget`.
3. **The refusal-to-guess is correctly scoped.** `graph_load` is the *only* non-injective value in the forward map — claimed by exactly `panel_flatten_workflow` and `panel_load_workflow`. Every other fenceable command has one claimant, so omitting ambiguous entries costs one command and invents nothing.
4. **The written-out list is the right shape.** `graph_query`'s tool really is `panel_query_graph`, not `panel_graph_query`, so a computed `panel_` prefix would have re-shipped the defect. A literal list pinned against the live registry is the correct trade.

No defect found in the diff. TDZ is not a risk here: `panelToolForGraphCmd` is a hoisted function declaration, and `PANEL_TOOL_BY_GRAPH_CMD` initializes at module scope well before any tool handler can run.

## One finding: the same defect exists at a SECOND agent-facing site, untouched by this PR

`src/services/ui-bridge.ts:4745-4764` (origin/main) — the workflow-instance-fence refusal (#1519/#1331), built with `markDispatched(new Error(...))` and surfaced to the agent. It carries **both** halves of the #1933 defect:

- The subject is the bridge command, same as the old `${name}`:
  `"${cmd.cmd}" cannot be safely targeted to the active workflow: ...` (L4763)
- All three `readsNote` branches advertise recovery reads by **bridge command** name. The third (L4758) is the worst of them — it names seven:

  > `Reads and view-only commands still work (graph_outline, graph_query, graph_get_state, graph_find_nodes, graph_list_subgraphs, graph_screenshot, graph_canvas).`

None of those seven is a registered tool. Resolved by the nearest enclosing `def(...)` at each command's dispatch site:

| named in the refusal | actual tool |
| --- | --- |
| `graph_outline` | `panel_graph_outline` |
| `graph_query` | `panel_query_graph` |
| `graph_find_nodes` | `panel_find_nodes` |
| `graph_list_subgraphs` | `panel_list_subgraphs` |
| `graph_screenshot` | `panel_screenshot` |
| `graph_canvas` | `panel_canvas` |
| `graph_get_state` | **none — it is not reachable from any tool** |

`graph_get_state` is the #384 back-compat fallback the shared graph-serialize helper tries when a pre-0.8.2 panel does not register `graph_serialize` (ui-bridge dispatch at panel-tools L10421). So that branch tells the agent an internal version-fallback "still works" as something it can call.

The tell that this is the same bug and not a deliberate convention: the *same message* correctly writes `Try panel_list_workflows` for its recovery probe — a tool name there, bridge names everywhere else.

**Why the suite will not catch it.** `src/__tests__/orchestrator/fence-lookalike-refusals.test.ts:69,89` builds its fixtures by quoting these templates verbatim — its own comment says *"Quoted from ui-bridge's templates: the `readsNote` really does name the phrase."* A fixture copied from the string under test is satisfied by a wrong name and a right one alike. That is precisely the blind spot this PR just removed from `graph-read-during-render.test.ts` (`toMatch(/graph_outline/)` matching both), still live at the other site.

I would keep it **out of this PR** — different message, different issue, and this diff is well-scoped as it stands. Filing it separately.

## What I did not do

- I did not run the unit suite or the Playwright gate, and I did not enter this branch's worktree. My verification is static analysis plus by-execution reads against `origin/main` — treat it as independent corroboration of the premise and the mapping, **not** as a CI result.
- This is one independent agent's read of the diff. It is not the grok review, and it should not be counted as one.

artokun:
Narrowing my own question **(1)** so the review effort lands on what is actually unproven, rather than on the half that already has a guard.

**Already proven, don't spend time here:** `panel-retry-identity.test.ts` **(f)** iterates every key of `RETRY_TOKEN_CMD_BY_TOOL` and asserts `expect(d, "<name> exists in buildPanelToolDefs()").toBeTruthy()`. So every *value* my reverse index can return is a registered tool by an existing invariant — the index cannot emit an unregistered name, no matter how the map drifts.

**The residual risk, and the only thing I actually need challenged:** (f) proves each key is a real tool and each value is in the #694 mutation surface. It does **not** prove the pairing — that the tool named on the left is the tool that dispatches the command on the right. If a forward entry is simply *wrong* about that (tool `X` mapped to `Y`, but `X` in fact dispatches `Z`), the retry machinery would still behave, because it only ever asks "does this tool mint a token" — and my message would then confidently name a real, wrong tool. That is a strictly worse failure than the bug being fixed, because the name is plausible.

I spot-checked `graph_set_widget → panel_set_widget` and `graph_auto_layout → panel_auto_layout` against their dispatch sites and asserted both in a test. I did **not** check all 44 entries. If you check anything here, check that pairing against the actual `ctx.call({ cmd: ... })` sites — a single wrong pair is invisible to every existing test and to mine.

Also verified while re-reading: `graph_load` really is dispatched from three call sites in `panel-tools.ts` (5469, 12565, 12654), reached by both mapped tools, which is why the index declines it.

Still unmerged, still no verdict.

artokun:
Recording this on the merged PR so the history is not misleading.

**This merged at 01:03:38Z (`760193b`) while the review requested above was still outstanding.** No verdict existed at merge time — the only review event on this PR is Copilot reporting it had hit its quota limit and could not review. So `760193b` shipped on the author's own verification alone, not on an independent review.

That does not void the request. The questions raised above still apply to the **merged code on `main`**, and any finding will be fixed there rather than treated as moot because the branch is gone. Concretely, the load-bearing one was: *can inverting `RETRY_TOKEN_CMD_BY_TOOL` — a map maintained for the `retry_of` question (#694) — ever yield a wrong tool rather than merely no tool?*

I have since closed that empirically in #1937: all 32 fence-reachable tools are driven and the tool name read back out of each tool's own refusal, with a map-only mutation proving the check is not circular. The remaining unreviewed surface is small, but it is not zero and it is not self-certified as clean.

One commit (`22f2347`, test-only) was written before the merge but pushed after it, so it is not in `760193b`. It is in #1937 rather than dropped.

PR #1937, opened 2026-08-21
test(1933): prove the tool/command PAIRING end-to-end across the fenced surface

artokun:
Follow-up to #1934 (merged as `760193b`). **Test only — no production code changes.**

## Why this is a separate PR

#1934 was **squash-merged at 01:03:38Z while the review I had requested on it was still outstanding**, at head `dfef368`. This commit existed but had not been pushed yet, so it did not make the merge. Landing it separately rather than silently dropping it.

Recording that plainly because it matters for anyone reading the history: `760193b` shipped **unreviewed**. Its author's own evidence is on #1934; the grok review requested there is still open and still applies to the merged code, not to a stale branch.

## What it adds

`#1934` recovers the calling tool's name through `panelToolForGraphCmd`, the reverse index of `RETRY_TOKEN_CMD_BY_TOOL` — a map maintained for a *different* question (#694: which tools mint a `retry_of` token). Two things about that were already guarded, and one was not:

- **Guarded:** `panel-retry-identity.test.ts` **(f)** asserts every *key* of the map is in `buildPanelToolDefs()`. So the reverse index can never emit an unregistered name.
- **Guarded:** the ambiguity case — `graph_load` is claimed by both `panel_load_workflow` and `panel_flatten_workflow`, so the index declines and the message says "This edit".
- **NOT guarded:** the **pairing**. Nothing asserted that the tool named on the left of a forward entry is the tool that actually dispatches the command on the right. A wrong entry there would leave the retry machinery working — it only ever asks whether a tool mints a token — while the refusal named a **real but wrong tool**. That is strictly worse than the original bug, because the name is plausible and the agent has no reason to doubt it.

#778 is this repo's standing record of what reading one list for another question costs, so this does not argue the pairing *from* the map. It **calls all 32 fence-reachable tools** with a prompt in flight and reads the tool name back out of the refusal each tool itself produced.

The pairings are deliberately not uniform — `panel_set_property` dispatches `graph_set_node_property`, `panel_set_node_title` dispatches `graph_set_title` — so a prefix rule passes a human review and fails here.

Every tool in the retry map is either driven or **excused by name with a stated reason**, and the excuses are themselves asserted to still be registered tools, so coverage cannot shrink quietly (a tool gaining a required argument becomes a failure, not a silent skip).

## Evidence

- 22/22 pass against real `origin/main` @ `760193b` (not a stale branch).
- **Mutation** — `subject = panelToolForGraphCmd(name) ?? "This edit"` reverted to `subject = name` on `origin/main`'s own source: **4 tests fail**, and the new one reports all 32 by name:
  ```
  expected [ …(32) ] to deeply equal []
  +   "panel_add_node: refusal named graph_add_node",
  ```
  Anchor asserted to have applied before running, file restored after.
- A second mutation on #1934's branch — repointing `panel_set_property` at `graph_set_title`, i.e. a *plausible, registered, wrong* tool — was also killed.
- `npx tsc --noEmit` clean; `check:vocabulary` OK, 160 dead names, unchanged ledger.

## Review

**Review pending.** Not gated, not reviewed, no verdict claimed. Given what happened to #1934, please do not merge this on CI alone.

### DIFF
--- a/src/__tests__/orchestrator/graph-read-during-render.test.ts
+++ b/src/__tests__/orchestrator/graph-read-during-render.test.ts
@@ -365,6 +365,135 @@ describe("graph MUTATIONS still fail fast while a prompt is running (#1639)", ()
   });
 });
 
+/**
+ * #1933 — the whole fence-reachable surface, driven end to end.
+ *
+ * The reverse index is built by inverting RETRY_TOKEN_CMD_BY_TOOL, a map
+ * maintained for a DIFFERENT question (#694: which tools mint a retry token).
+ * `panel-retry-identity.test.ts` (f) already proves every KEY of that map is a
+ * registered tool, so the index can never emit a name that does not exist. What
+ * nothing proved is the PAIRING — that the tool on the left is the tool a caller
+ * reaches the command on the right through. A forward entry that is simply wrong
+ * about that would leave the retry machinery working (it only ever asks whether a
+ * tool mints a token) while this message names a real but WRONG tool, which is
+ * strictly worse than the bug being fixed, because the name is plausible.
+ *
+ * #778 is the standing record of what reading one list for another question
+ * costs, so this does not argue the pairing from the map. It CALLS each tool with
+ * a prompt in flight and reads the tool name back out of the refusal the tool
+ * itself produced. The pairings are not uniform — `panel_set_property` dispatches
+ * `graph_set_node_property`, `panel_set_node_title` dispatches `graph_set_title`
+ * — so a prefix rule would pass a review and fail here.
+ */
+const FENCED_TOOL_ARGS: Readonly<Record<string, Record<string, unknown>>> = {
+  panel_add_node: { type: "KSampler" },
+  panel_remove_node: { node_id: 3 },
+  panel_connect: { from_node: 1, from_slot: 0, to_node: 2, to_slot: 0 },
+  panel_disconnect: { node_id: 2, input: 0 },
+  panel_configure_app_mode: { inputs: [] },
+  panel_set_widget: { node_id: 3, widget: "text", value: "x" },
+  panel_remove_widget: { node_id: 3, widget: "text" },
+  panel_set_property: { node_id: 3, property: "p", value: 1 },
+  panel_move_node: { node_id: 3, x: 1, y: 2 },
+  panel_resize_node: { node_id: 3, width: 100, height: 100 },
+  panel_auto_layout: {},
+  panel_create_subgraph: { node_ids: [1, 2] },
+  panel_subgraph_group: { group_id: 0 },
+  panel_paste_nodes: {},
+  panel_save_subgraph: { node_id: 3, name: "s" },
+  panel_add_subgraph: { name: "s" },
+  panel_create_group: { title: "g", node_ids: [1] },
+  panel_move_group: { group_id: 0, x: 1, y: 1 },
+  panel_edit_group: { group_id: 0, title: "g2" },
+  panel_remove_group: { group_id: 0 },
+  panel_set_node_title: { node_id: 3, title: "t" },
+  panel_set_node_collapsed: { node_id: 3, collapsed: true },
+  panel_set_node_mode: { node_id: 3, mode: "mute" },
+  panel_set_node_color: { node_id: 3, color: "red" },
+  panel_move_rail: { node_id: 3, slot: 0, direction: "up" },
+  panel_promote_widget: { node_id: 3, widget: "text" },
+  panel_expose_subgraph_output: { node_id: 3, name: "o" },
+  panel_expose_subgraph_input: { node_id: 3, name: "i" },
+  panel_unexpose_subgraph_output: { node_id: 3, name: "o" },
+  panel_unexpose_subgraph_input: { node_id: 3, name: "i" },
+  panel_unpack_subgraph: { node_id: 3 },
+  panel_update_node: { node_id: 3 },
+};
+
+/**
+ * Tools in the retry map that this probe deliberately does not drive, each for a
+ * stated reason. Pinned as a SET so a tool silently dropping out of coverage —
+ * by gaining a required argument, say — turns into a failure here rather than
+ * quietly shrinking what is checked.
+ */
+const NOT_DRIVEN: Readonly<Record<string, string>> = {
+  // Never fenced: GRAPH_CMD_EFFECT "inert" (selection / scope / clipboard). The
+  // map admits them for the retry question only (#694), not this one.
+  panel_select_nodes: "inert — the fence lets it past",
+  panel_copy_nodes: "inert — the fence lets it past",
+  panel_enter_subgraph: "inert — the fence lets it past",
+  panel_exit_subgraph: "inert — the fence lets it past",
+  // Both dispatch graph_load, which the index refuses to resolve; and both read
+  // the live canvas before dispatching, so neither reaches the fence with a
+  // constructible argument here. Covered by the ambiguity test above instead.
+  panel_load_workflow: "graph_load — ambiguous, covered separately",
+  panel_flatten_workflow: "graph_load — ambiguous, covered separately",
+  // Short-circuit BEFORE the fence: argument validation and a confirm card
+  // respectively, so no refusal is produced to read a name out of.
+  panel_edit_node: "rejects an empty edit before dispatch",
+  panel_clear: "confirm card resolves 'no' in this harness",
+  // workflow_* commands, not graph_* — outside this fence entirely.
+  panel_save_workflow: "workflow_save is not a graph_ command",
+  panel_rename_workflow: "workflow_rename is not a graph_ command",
+  panel_close_workflow: "workflow_close is not a graph_ command",
+  // Deliberately exempt: queuing behind an in-flight job is the sweep path.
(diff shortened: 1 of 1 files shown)


artokun:
@grok please review — and note the context, because it changes what your review is for.

**#1934 merged at 01:03:38Z as `760193b` while the review I requested on it was still outstanding.** The production change is therefore already on `main`, unreviewed. My open questions there were never answered, so I am re-pointing the important one here, against the code that actually shipped rather than against a branch.

**This PR is test-only.** The thing to attack is whether it proves what it claims.

**1. The claim it makes.** `panel-retry-identity.test.ts` (f) already proves every *key* of `RETRY_TOKEN_CMD_BY_TOOL` is a registered tool, so `panelToolForGraphCmd` cannot emit a name that does not exist. What was unproven is the **pairing** — that the tool on the left dispatches the command on the right. A wrong entry there produces a refusal naming a **real but wrong** tool, which is worse than the bug #1934 fixed because the name is plausible. This test drives all 32 fence-reachable tools and reads the name back out of each tool's own refusal. **Is that actually sufficient, or does driving the tool through `buildPanelToolDefs()` re-derive the same mapping I am trying to check?** If the handler's dispatched command and the refusal's subject both trace back to the same table, the test is circular and I have proved nothing. I believe it is not — the subject is resolved from `cmd.cmd` at the fence, and `cmd.cmd` is a literal in each handler — but that is precisely the kind of thing an author reads as obviously right.

**2. The excuse list.** 13 tools are `NOT_DRIVEN` with stated reasons. Two of them (`panel_edit_node`, `panel_clear`) short-circuit before the fence *in this harness* — argument validation and a confirm card. If either could reach the fence in production, my coverage claim is overstated and those should be driven, not excused.

**3. Coverage counting.** I assert `checked === Object.keys(FENCED_TOOL_ARGS).length` and `checked > 30` to stop the loop silently becoming a no-op. Check that a tool which stops reaching the fence fails rather than skips — that is the failure mode I was designing against, and the one I am least able to see in my own code.

**4. Still open from #1934, now against merged `main`:** whether inverting a map maintained for the `retry_of` question can ever yield a wrong tool rather than merely no tool. This PR is my attempt to answer that empirically for the 32 drivable cases. It does not cover `graph_load` (ambiguous by design) or the four inert UI-state commands (never fenced).

Mutation evidence and gate results are in the PR body. Not merging this on my own verification — and given #1934, please say so explicitly if you think it should not be merged on CI alone either.

artokun:
Answering my own question (1) above with the evidence I already had, rather than leaving it for you to re-derive — but the other three still stand.

**The test is not circular, and a mutation shows it rather than an argument.** The two sides come from different places: the left is the `cmd: "graph_…"` **string literal inside each handler's own `ctx.call`**, resolved at the fence through the inverse of `RETRY_TOKEN_CMD_BY_TOOL`; the right is `def.name`, the registration literal. `RETRY_TOKEN_CMD_BY_TOOL` is a separate hand-written table, so the test is cross-checking that table against the actual dispatch site.

The proof is mutation **M6** from #1934: I repointed **only the map** (`panel_set_property: "graph_set_node_property"` → `"graph_set_title"`), touching no handler and no registration. The suite went red. Had the test been circular, moving the map would have moved both sides together and it would have stayed green.

So the residual risk I flagged — a forward entry that is wrong about which tool dispatches which command — is now closed by execution for all 32 drivable cases, not by reading. It remains open only for `graph_load` (ambiguous by design, resolves to nothing) and the four inert UI-state commands the fence never refuses.

artokun (commented):

Head `85f7499a13a0ad2fd0e2d12f90a928a01a705173` vs `origin/main`. Test-only PR; production fence is already on main (`graphCmdBlockedByRunningPrompt` → `panelToolForGraphCmd` → unique reverse of `RETRY_TOKEN_CMD_BY_TOOL`). Answering the four questions you left here.

**1. Circularity — not circular.** Agree with the follow-up. The subject is `panelToolForGraphCmd(cmd.cmd) ?? "This edit"`. `cmd.cmd` is the handler's string literal; the reverse index is a separate unique-claim table. Driving through `buildPanelToolDefs()` does not re-derive both sides from one source. M6 (repoint only the map) is the right proof.

**2. Excuse list — coverage is overstated. Drive these two; do not excuse them.**

- `panel_edit_node` **does** reach the fence in production. `validatePanelEditNodeArgs` only rejects the empty-edit case. `{ node_id: 3, title: "t" }` is enough (same shape as `panel_set_node_title`). The excuse describes calling it with no fields, not production reachability.
- `panel_clear` **does** reach the fence after `confirm === "yes"` (`ctx.call({ cmd: "graph_clear" })`). This harness's `makeBridge()` has no `ask_user` path, so confirm never returns yes. Stub `ctx.confirm = async () => "yes"` (or answer `ask_user`) and drive `{}`.

The other excuses are sound: inert (fence lets them past), `graph_load` (ambiguous by design), `workflow_*` (not `graph_*`), `panel_run` (excluded).

**3. Coverage counting — holds for `FENCED_TOOL_ARGS`.** A listed tool that misses the fence is a mismatch, not a skip. `checked === length` plus `> 30` stops the loop becoming a no-op. What it cannot catch is moving a still-fenced tool into `NOT_DRIVEN` with a stale excuse — which is why (2) matters.

**4. Inverse of the retry map — unique claims only; collisions refuse.** `PANEL_TOOL_BY_GRAPH_CMD` drops any cmd claimed by two tools (`graph_load` → `"This edit"`). A wrong unique pairing on a **driven** tool fails this test. A wrong unique pairing on an **undriven** fenced tool (`graph_edit_node` / `graph_clear`) would not. Same hole as (2).

Do not merge on CI alone until `panel_edit_node` and `panel_clear` are in `FENCED_TOOL_ARGS`.

artokun:
@grok addressed in `0b5ca9c` — you were right on both, and I checked each by execution before changing anything rather than taking the finding on trust.

**`panel_edit_node`** — `validatePanelEditNodeArgs` rejects only the empty-edit case, exactly as you said. `{ node_id: 3, title: "t" }` passes and dispatches `graph_edit_node`. My excuse described the args this file happened to pass (`widgets`, which is not in `PANEL_EDIT_NODE_FIELDS`) — a statement about my call, not about reachability. Driven now, and it emits `panel_edit_node was NOT sent…`.

**`panel_clear`** — the handler reaches `ctx.call({ cmd: "graph_clear" })` once `confirm` returns `"yes"`; my `makeBridge()` had no path to yes, so the excuse was about the harness. Added a `CONFIRM_YES` set that stubs `ctx.confirm`, with a comment saying why it exists — that missing stub is precisely how the tool came to be mis-excused.

**On your point (4), which is the part I think actually mattered.** The coverage number was not the issue: a wrong unique pairing on a *driven* tool fails the test, on an *undriven* fenced tool it does not — so the two excused tools were the exact blind spot, the same shape as the `toMatch(/graph_outline/)` assertion that let #1933 ship. Excusing them was the same class of mistake one level up.

**Proved the two new cases are load-bearing rather than merely present**, per your step 2. Repointed each command at a *different real registered tool* (`panel_strip_workflow`) so the reverse index resolves it uniquely to a plausible wrong name:

```
M7 graph_edit_node -> "panel_edit_node: refusal named panel_strip_workflow"   suite RED
M8 graph_clear     -> "panel_clear: refusal named panel_strip_workflow"       suite RED
```

Each anchor asserted to have applied before the run (a CRLF-eaten replacement reads as "survived"), and the file restored after — confirmed byte-identical.

Now 34 tools driven; the coverage floor moved with it (`toBeGreaterThan(33)`), so a tool that stops reaching the fence fails rather than silently skips. `tsc --noEmit` clean, `check:vocabulary` OK at the same 160-name ledger, 22/22 green.

**Not merging** — re-requesting review on the new head, and I'll hold per your "do not merge on CI alone".

One note for the record, since it shaped this exchange: **#1934 was merged out from under this same review request** at 01:03:38Z, inside your response window. That production change is on `main` and in tag `v0.52.37` **unreviewed**. Your r1 questions apply to it as much as to this test PR — if anything here makes you doubt the merged code, say so and I'll open a fix PR against `main` rather than treat it as settled.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `0b5ca9c727e666877cda7d240292753bc4eac07b` vs `origin/main`. Fresh review — prior REQUEST_CHANGES was on `85f7499a` and does not cover this SHA.

Both holes are closed. `panel_edit_node` is driven with a real field (`{ node_id: 3, title: "t" }`), which is what production actually sends to `graph_edit_node`. `panel_clear` is driven after `CONFIRM_YES` stubs `ctx.confirm` to `"yes"`, which is the production path to `graph_clear`. Neither sits in `NOT_DRIVEN`. Floor is `checked > 33` for 34 driven tools, so a tool that stops reaching the fence fails rather than silently shrinking.

M7/M8 (repoint `graph_edit_node` / `graph_clear` at another registered tool so the reverse index names a plausible wrong one) are the right proof that these two cases are load-bearing, not merely present.

Remaining excuses are still sound: inert, `graph_load` ambiguous, `workflow_*`, `panel_run`.

On the merged #1934 production fence: I do not doubt it. Unique-claim reverse index plus `"This edit"` on collisions was the right recovery; this test now covers the two unique pairings that were previously the blind spot. No follow-up fix PR needed against `main`.

Issue #1715, opened 2026-08-18
Fix node_pack scaffold using install root instead of live Desktop base directory


artokun:
### What happened / root cause
node_pack action=scaffold wrote a new pack under the code install root custom_nodes, while the connected ComfyUI Desktop runtime loaded custom nodes from its --base-directory under Documents. node_pack action=verify then restarted the live server and correctly reported the class missing. The authoring and verification phases therefore targeted different roots.

### Steps to reproduce
1. Run ComfyUI Desktop with main.py under the install tree and --base-directory pointing to a separate Documents/ComfyUI directory.
2. Call node_pack scaffold for a new pack.
3. Observe the returned path is <install-root>/custom_nodes/<pack>.
4. Call node_pack verify with restart:true.
5. The class is missing because runtime scans <base-directory>/custom_nodes.

### Exact error (scrubbed)
ready:true, restarted:true, expected:["CharacterPlateComposite"], loaded:[], missing:["CharacterPlateComposite"].

### Fix
applied locally: workaround only
upstream-only: yes
Resolve the writable custom_nodes root from the connected runtime before scaffold/write/list/read/patch. Prefer the live server's --base-directory/custom_nodes (or Manager/runtime-reported custom_nodes path) over the code install root. Persist that resolved root through scaffold and verify so both phases target the same directory. If multiple roots exist, return which one is active and refuse ambiguous writes.

Workaround: copied the generated pack into ~/Documents/ComfyUI/custom_nodes and will verify against the live server.

### Environment
Windows 10
ComfyUI 0.33.1
Frontend 1.48.7
RTX 3090 24GB
comfyui-mcp 0.52.1
panel 0.15.1

artokun:
Claimed by kimi-code swarm (worktree fix-1715). Agent assigned now; if no PR appears within ~8h the claim has lapsed.

PR #1938, opened 2026-08-21
fix(1765): install_custom_node refuses a local write aimed at an install this session is not connected to

artokun:
## What this is

The recurrence reported on **0.52.35** against #1765 — not the original feature. `COMFYUI_CODE_PATH` shipped in **0.52.11** via #1766 and is not what failed. The report is:

> On macOS with six local installs, the MCP target was `http://127.0.0.1:8189` (the krea-2 install), but `install_custom_node` with `source:"git"` cloned an unregistered repository into comfy-cli's saved default workspace for a different running install (qwen-image-edit on :8188). The call reported `git-clone` success and only the returned `nodeDir` exposed the wrong target.

That is a **wrong-target write**: a pack installed into a ComfyUI the user did not name, reported as a success.

## Which selector chose the install — established by execution, not by reading

The first revision of `src/__tests__/services/install-node-target-1765.test.ts` was a reproduction, and it reproduced the report verbatim. Two real install trees on disk, a live server self-reporting `<krea-2>/main.py` on :8189, a saved default workspace naming `<qwen-image-edit>`, the real `installCustomNode`:

```
RESULT: { "mechanism": "git-clone",
          "details": { "nodeDir": ".../qwen-image-edit/custom_nodes/unregistered-pack" } }
GIT CALLS: [["git","clone","--depth","1","--end-of-options",
             "https://github.com/someone/unregistered-pack",
             ".../qwen-image-edit/custom_nodes/unregistered-pack"]]
```

The destination came from **`resolveEffectiveComfyUIBase()`** — `COMFYUI_PATH` env, then auto-detection, then the saved default workspace, i.e. **configuration alone**. In the reporter's shape (six installs in non-standard directories, so auto-detection finds nothing) that resolves to the saved default workspace, which is a statement about the *machine* and not about the *session's target*.

`/system_stats` was **not called once** during the install. `resolveLiveServerRoot` — which `workspace-env.ts` documents as *"the single notion every write-side caller (download destination, package install) must resolve through, so they can never disagree about which install is 'the live one'"* — was never consulted on this path. This is the canonical-resolver-adopted-per-call-site shape: the resolver is right, the call site never adopted it.

Both roads into `cloneCustomNodeFallback` have it (a Manager 403 refusal — #1129's divert — and a Manager enqueue that drains without installing an unregistered pack), and so does the `useCmCli` branch, which passes the same root as `comfy-cli --workspace`.

## The change

`detectLocalWriteTargetMismatch()` (workspace-env.ts) answers exactly one question: **are these provably two different installs?** It asks the running server where it lives — its resolved `--base-directory` first, else `resolveLiveServerRoot`, this module's single notion of the live install root — and reports a mismatch only when that root is proven, exists on disk, looks like an install, and is a different tree than the base the write was about to land in.

That order is ComfyUI's own, verified against `folder_paths.py` in the install running on this machine: `base_path` is `abspath(args.base_directory)` when the flag is present and `dirname(realpath(__file__))` otherwise, and `custom_nodes` is seeded with exactly one entry under it. Note the `realpath` — symlinked spellings collapse there, which is why the comparison canonicalizes rather than comparing strings.

`installCustomNodeImpl` takes that observation once, pinned to the same `cliWorkspace` the write will use, and **refuses** at each local-write site. The refusal names both roots, which selector produced each, and two remedies.

**It refuses; it does not retarget.** Silently preferring the live `main.py` root would trade this bug for its mirror image: on a `--base-directory` split install the configured data root *is* the pack root (#1715/#1770), which is the design #1766 shipped for this very issue. A refusal that names the ambiguity cannot write into the wrong tree; a silent pick can.

**Scope, and where it fails open** (deliberately, and stated rather than implied):
- remote sessions — no local write to misdirect, and refused earlier anyway;
- the server unreachable — an outage is not evidence;
- a relative `main.py` that no process observation can anchor — unprovable is not wrong;
- an observed interpreter belonging to no install we can anchor (a system python) — the containment test rejects it, so no root, so no refusal;
- a `--base-directory` that cannot be resolved — every root derivable is then the wrong one;
- the registry install path — it goes through ComfyUI-Manager against the connected server, so there is no local destination to misdirect, and it does not even pay for the `/system_stats` probe.

## Three things earlier revisions got wrong, and how each was caught

1. **An argv-only live root made the whole guard INERT on the layouts that matter — caught by running it against the real server, not by reading it.** One revision used `liveRootFromArgv` alone, reasoning that `resolveLiveServerRoot`'s `observed-process` tier anchors on where the interpreter *binary* lives and its own docstring calls that a *"KNOWN GAP ... not proof of where the SCRIPT was resolved from"* — too weak a basis for refusing a user's write. Principled, and wrong. Measured against the ComfyUI running on this machine (0.33.2, ComfyUI Desktop):

   ```
   argv[0] : "ComfyUI\main.py"      <- RELATIVE
   cwd     : the /system_stats payload has no `cwd` FIELD AT ALL
   ```

   `liveRootFromArgv` resolves a relative script only against an absolute server cwd, and **ComfyUI does not report one**. So the argv tier answers only when argv[0] is absolute — and ComfyUI Desktop, the Windows portable bundle, and `comfy launch` (which runs `python main.py` from the workspace) are all relative. Every test in this file was green while the guard did nothing on the real rig. Both tiers are used now.

   The `observed-process` tier is also what the **download** path already resolves its write *destination* through (#369/#1374), so trusting it to *detect* a mismatch is strictly weaker than what already ships — and its documented gap costs a refusal that writes nothing, never a wrong write.

   Live-verified on the same rig after the change: a different tree reports a mismatch, the live root itself does not, and neither does an uppercased backslash spelling of it.

2. **It carved out an explicit `COMFYUI_PATH`** on the theory that the env var means the user named the install. That is false here: the panel orchestrator resolves `envComfyuiPath || detectLocalComfyUIPath()` and forwards the **result** to every child MCP as `COMFYUI_PATH` (`orchestrator/index.ts`), so in a panel-spawned child an auto-detected guess is byte-identical to a deliberate setting. The carve-out would have switched the check off in exactly the sessions it exists for while reading as careful scoping. It is gone; the refusal instead names `--base-directory` as the supported way to make a data root authoritative.

3. **Two ordering defects found by re-reading my own diff**, both committed: the mismatch probe is the function's first `await` and had been placed *before* `capturePackPresenceContext`, which codex gate round 11 requires be taken at operation entry (otherwise a mid-op retarget pairs Manager A's answer with disk B); and the refusal read `getComfyUIBaseUrl()` at throw time rather than the `managerBase` already pinned for the operation.

## Evidence

**Mutation-tested — every one of these was applied, run, and killed a named test** (the fix was committed first, so `git checkout` could not silently revert it, and the mutation script fails loudly when the pattern does not match):

| mutation | tests killed (of 16) |
|---|---|
| helper always returns "no mismatch" | 7 |
| guard deleted at the 403-divert clone site | 4 |
| guard deleted at the unregistered-pack clone site | 1 |
| guard deleted at the `comfy-cli` site | 1 |
| probe taken unconditionally instead of only where a local write is reachable | 1 |
| `--base-directory` branch removed | 2 |
| `sameInstallRoot` always false | 5 |
| `observed-process` tier dropped (back to argv-only) | 1 |
| unresolvable-`--base-directory` fail-open gate removed | 1 |

**Two of these survived on first run** — the unregistered-pack call site and the unresolvable-`--base-directory` gate. Neither had a test. Both got one, rather than the mutation being explained away.

**Suite:** baselined on a clean `origin/main` worktree first — `2 failed files / 4 failed tests` (`node-id-union-message.test.ts`, `package-hooks.test.ts`), pre-existing and unrelated. This branch: the same two files, the same four tests, plus the 16 added here.

**Live-verified against the real ComfyUI on this machine** (0.33.2, Desktop, port 8188). That is what exposed the inertness in point 1 above, and what confirmed the fix. No Manager traffic and no writes — the check only asks the running server where it lives and compares.

And then the suite verified it a second way, without being asked. `node-install-invalidates-object-info.test.ts` (#444) stubs every network boundary except `/system_stats`, so on this machine it reached the developer's **own running install**, observed that it is not the fixture's `/fake/comfy`, and refused — with nothing about the live path mocked: real `/system_stats`, real process-table probe, real install root. That is the guard working end to end. The fixture now has no server, so it fails open exactly as it does on CI, where nothing is listening.

**Review is PENDING.** Nothing here has been reviewed by anyone but its author. Not gated: `codex-gate.mjs` was not run and codex was not spawned.

- **Did not extend the guard to `uninstall`/`enable`/`disable`.** They resolve the same root through `resolveEffectiveComfyUIBase()` and carry the same hazard — worse, in fact, since `comfy-cli uninstall` deletes — but only behind an explicit `useCmCli`, and they are not what was reported. Worth a follow-up; not smuggled into a P2 bug fix.
- **Did not change `resolveEffectiveComfyUIBase`/`…Live`.** Every filesystem-backed tool reads them; narrowing a shared resolver to fix one caller is how the next three break.
- **Did not touch any tool description**, so the tool-vocabulary ledger is unchanged (`check:vocabulary` green regardless).
- **No panel change**, so no Playwright run — nothing here renders.
- **Known false positive, named rather than hidden:** a deployment that lists `custom_nodes` in the live server's `extra_model_paths` config pointing at the configured base *is* read by that server, and this cannot see that — it would be refused. Nothing is written, the Manager route still works, and the message names both roots. Corroborating against the server's own extra paths is a different seam and is not claimed here.

## Review

A review of head `826c5fd` came back **COMMENT INTENDED: APPROVE**, and its nit is fixed here (a comment still claimed the helper "only ever reports a mismatch for a base nobody named for this session", which stopped being true when the carve-out was removed).

**That approval does not carry to this head, and I am not treating it as one.** It was written against the argv-only revision, and two of its findings explicitly bless what I have since changed: it recorded "Desktop/portable relative `main.py` fail open (tested)" and "Deliberately skips `observed-process`" as correct fail-open behaviour. The live measurement above shows that combination made the check inert on the layouts that matter, so the newer commits do the opposite. A verdict on a superseded diff is not a verdict on this one — re-review requested.

Fixes the recurrence reported on #1765.

### DIFF
--- a/src/__tests__/services/install-node-target-1765.test.ts
+++ b/src/__tests__/services/install-node-target-1765.test.ts
@@ -161,19 +161,28 @@ vi.mock("node:child_process", async () => {
   };
 });
 
-// The OS process-table probe. `resolveLiveServerRoot`'s `observed-process` tier
-// anchors on where the interpreter BINARY lives, which workspace-env.ts itself
-// documents as "KNOWN GAP … not proof of where the SCRIPT was resolved from".
-// A resolution with a documented soundness gap may FIND a root; it may not be
-// what REFUSES a user's write, so this must never be consulted here.
-const probe = vi.hoisted(() => ({ calls: 0 }));
+// The OS process-table probe — `resolveLiveServerRoot`'s `observed-process` tier.
+//
+// MEASURED against the ComfyUI running on the development machine (0.33.2,
+// ComfyUI Desktop): argv[0] is the RELATIVE "ComfyUI\main.py" and the
+// /system_stats payload carries no `cwd` field at all. `liveRootFromArgv`
+// resolves a relative script only against an absolute server cwd, so on that
+// layout — and on the Windows portable bundle, and on `comfy launch`, which runs
+// `python main.py` from the workspace — the argv tier answers NOTHING. An
+// argv-only guard is green in every test here and INERT on the machines where
+// several installs coexist. Hence this tier, and hence these tests.
+const probe = vi.hoisted(() => ({
+  calls: 0,
+  /** What the OS says is running on the port; undefined = nothing observable. */
+  result: undefined as { python: string; pid: number } | undefined,
+}));
 vi.mock("../../services/live-interpreter.js", async (importOriginal) => {
   const actual = await importOriginal<typeof import("../../services/live-interpreter.js")>();
   return {
     ...actual,
-    observeLiveServerProcess: (...args: unknown[]) => {
+    observeLiveServerProcess: () => {
       probe.calls += 1;
-      return (actual.observeLiveServerProcess as (...a: unknown[]) => unknown)(...args);
+      return probe.result;
     },
   };
 });
@@ -211,6 +220,7 @@ let priorEnvPath: string | undefined;
 beforeEach(() => {
   git.calls = [];
   probe.calls = 0;
+  probe.result = undefined;
   http.mode = "refuse-enqueue";
   live.reachable = true;
   live.statsCalls = 0;
@@ -364,19 +374,49 @@ describe("#1765 — install_custom_node must not write into an install this sess
     expect(clonedInto()).toBe(join(SAVED_DEFAULT, PACK_DIR));
   });
 
-  it("fails OPEN on the Desktop/portable shape, without probing the process table", async () => {
-    // ComfyUI Desktop and the Windows portable bundle both report a RELATIVE
-    // main.py and no cwd. The server's SELF-REPORT cannot resolve a root there,
-    // and the interpreter-anchored tier that can is not sound enough to refuse
-    // on — so nothing is refused, and the OS probe is never even taken.
+  it("REFUSES the relative-main.py shape too, via the OS process observation", async () => {
+    // The shape the live rig actually reports, and the one `comfy launch`
+    // produces: a relative launch script and no server cwd. Nothing in argv
+    // names a root; the interpreter the OS says is on the port does.
+    live.argv = ["main.py", "--port", "8189"];
+    probe.result = { python: join(CONNECTED, ".venv", "Scripts", "python.exe"), pid: 4242 };
+
+    const { ok, error } = await install({ id: REPO, source: "git" });
+
+    expect(ok).toBeUndefined();
+    expect(clonedInto()).toBeUndefined();
+    expect(existsSync(join(SAVED_DEFAULT, PACK_DIR))).toBe(false);
+    expect(error).toContain(CONNECTED);
+    expect(error).toContain("OS process listening on that port");
+    expect(probe.calls).toBeGreaterThan(0);
+  });
+
+  it("fails OPEN when the relative main.py can be anchored on NOTHING", async () => {
+    // Relative script, and the OS observation yields no interpreter (a remote
+    // proxy, a permissions failure, an unreadable process table). Unprovable is
+    // not the same as wrong, so nothing is refused.
     live.argv = ["ComfyUI/main.py", "--port", "8189"];
+    probe.result = undefined;
 
     const { ok, error } = await install({ id: REPO, source: "git" });
 
     expect(error).toBeUndefined();
     expect(ok).toMatchObject({ mechanism: "git-clone" });
     expect(clonedInto()).toBe(join(SAVED_DEFAULT, PACK_DIR));
-    expect(probe.calls).toBe(0);
+  });
+
+  it("fails OPEN when the observed interpreter belongs to no install we can anchor", async () => {
+    // A SYSTEM python: walking up from it reaches directories that are not the
+    // install at all, which is the unsoundness `interpreterBelongsToInstall`
+    // exists to reject. It must not produce a root, and so must not refuse.
+    live.argv = ["ComfyUI/main.py", "--port", "8189"];
+    probe.result = { python: join(SANDBOX, "python", "python.exe"), pid: 4243 };
+
+    const { ok, error } = await install({ id: REPO, source: "git" });
+
+    expect(error).toBeUndefined();
+    expect(ok).toMatchObject({ mechanism: "git-clone" });
+    expect(clonedInto()).toBe(join(SAVED_DEFAULT, PACK_DIR));
   });
 
   it("refuses the comfy-cli branch too — --workspace has the identical hazard", async () => {
--- a/src/__tests__/services/install-node-target-1765.test.ts
+++ b/src/__tests__/services/install-node-target-1765.test.ts
@@ -161,6 +161,23 @@ vi.mock("node:child_process", async () => {
   };
 });
 
+// The OS process-table probe. `resolveLiveServerRoot`'s `observed-process` tier
+// anchors on where the interpreter BINARY lives, which workspace-env.ts itself
+// documents as "KNOWN GAP … not proof of where the SCRIPT was resolved from".
+// A resolution with a documented soundness gap may FIND a root; it may not be
+// what REFUSES a user's write, so this must never be consulted here.
+const probe = vi.hoisted(() => ({ calls: 0 }));
+vi.mock("../../services/live-interpreter.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../../services/live-interpreter.js")>();
+  return {
+    ...actual,
+    observeLiveServerProcess: (...args: unknown[]) => {
+      probe.calls += 1;
+      return (actual.observeLiveServerProcess as (...a: unknown[]) => unknown)(...args);
+    },
+  };
+});
+
 const { installCustomNode } = await import("../../services/node-management.js");
 const { configureWorkspace, resetWorkspaceConfig } = await import(
   "../../services/workspace-env.js"
@@ -193,6 +210,7 @@ let priorEnvPath: string | undefined;
 
 beforeEach(() => {
   git.calls = [];
+  probe.calls = 0;
   http.mode = "refuse-enqueue";
   live.reachable = true;
   live.statsCalls = 0;
@@ -329,9 +347,11 @@ describe("#1765 — install_custom_node must not write into an install this sess
     expect(clonedInto()).toBe(join(SAVED_DEFAULT, PACK_DIR));
   });
 
-  it("fails OPEN when the server's main.py root cannot be resolved", async () => {
-    // ComfyUI Desktop reports a RELATIVE main.py and no cwd. Nothing here is
-    // provable, so nothing is refused.
+  it("fails OPEN on the Desktop/portable shape, without probing the process table", async () => {
+    // ComfyUI Desktop and the Windows portable bundle both report a RELATIVE
+    // main.py and no cwd. The server's SELF-REPORT cannot resolve a root there,
+    // and the interpreter-anchored tier that can is not sound enough to refuse
+    // on — so nothing is refused, and the OS probe is never even taken.
     live.argv = ["ComfyUI/main.py", "--port", "8189"];
 
     const { ok, error } = await install({ id: REPO, source: "git" });
@@ -339,6 +359,7 @@ describe("#1765 — install_custom_node must not write into an install this sess
     expect(error).toBeUndefined();
     expect(ok).toMatchObject({ mechanism: "git-clone" });
     expect(clonedInto()).toBe(join(SAVED_DEFAULT, PACK_DIR));
+    expect(probe.calls).toBe(0);
   });
 
   it("refuses the comfy-cli branch too — --workspace has the identical hazard", async () => {
--- a/src/__tests__/services/install-node-target-1765.test.ts
+++ b/src/__tests__/services/install-node-target-1765.test.ts
@@ -321,6 +321,21 @@ describe("#1765 — install_custom_node must not write into an install this sess
     expect(error).toContain("--base-directory");
   });
 
+  it("fails OPEN when --base-directory is relative and unresolvable", async () => {
+    // The server WAS given a base directory, but a relative one with no server
+    // cwd to resolve it against. It scans custom_nodes/ from a directory we
+    // cannot name — and the main.py root, which argv does give us, is then
+    // confidently NOT that directory. Judging the write against it would be
+    // comparing the base to the wrong tree, so nothing is refused.
+    live.argv = [join(CONNECTED, "main.py"), "--base-directory", "data"];
+
+    const { ok, error } = await install({ id: REPO, source: "git" });
+
+    expect(error).toBeUndefined();
+    expect(ok).toMatchObject({ mechanism: "git-clone" });
+    expect(clonedInto()).toBe(join(SAVED_DEFAULT, PACK_DIR));
+  });
+
   it("clones normally when the server's --base-directory IS the configured root", async () => {
     // The #1715/#1770 split shape: main.py in one tree, custom_nodes scanned from
     // the data root — which is exactly the root we were about to write into.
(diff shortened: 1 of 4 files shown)


artokun:
@grok please review. This is a **behaviour-narrowing refusal on a write path**, so the risk is false positives on installs that work today. Naming exactly where the load is, rather than handing you a diff:

**1. The load-bearing claim is: "without `--base-directory`, ComfyUI scans `custom_nodes/` from the `main.py` root, so a configured base that differs is not a root the connected server reads."** Everything rests on that. If it is false for any shipping launcher — a Desktop build, the Windows portable bundle, a wrapper that sets `folder_paths` before `main.py` runs — then `detectLocalWriteTargetMismatch` refuses correct installs. Please attack this one first.

**2. The known false positive I could not close, and did not hide:** `extra_model_paths` can add `custom_nodes` entries, so a deployment whose configured base IS listed there is genuinely read by the live server and would now be refused. I state it in the code and in the PR body rather than claiming coverage. Is it common enough that this should not ship without corroborating against the server's own extra paths first? That is the judgement call I am least sure of, and it is a ship/no-ship question, not a nit.

**3. The carve-out I removed, and why it might need to come back.** v1 skipped the check when `COMFYUI_PATH` was set, reasoning "the user named it". I killed that after finding `orchestrator/index.ts` resolves `envComfyuiPath || detectLocalComfyUIPath()` and forwards the RESULT to every child MCP under that same name — so in a panel child, an auto-detected guess and a deliberate setting are byte-identical. Check that reading. If I am wrong about the forwarding, the removal made this far more aggressive than intended.

**4. REFUSE vs RETARGET.** I refuse because on a `--base-directory` split install the configured data root IS the pack root (#1715/#1770 — the design #1766 shipped for this very issue), so silently preferring the live `main.py` root trades this bug for its mirror. If you think one selector is provably authoritative here, say so — I could not prove it.

**5. Fail-open completeness.** The refusal must never fire on absence of evidence: unreachable server, relative `main.py` with no cwd, unresolvable `--base-directory`, no live root on disk. I deliberately do NOT use `resolveLiveServerRoot`'s `observed-process` tier (its own docstring calls it a "KNOWN GAP … not proof of where the SCRIPT was resolved from") — a resolution with a documented soundness gap may find a root, but must not refuse a user's write. Is there a fifth unprovable state I let through as proof?

Evidence in the body: the reproduction ran the real `installCustomNode` and produced the reporter's outcome verbatim before any change, and seven mutations (including all four call sites) were each applied and each killed a named test — one of them survived first, which is how the missing test got written. Full suite matches a clean `origin/main` baseline exactly.

**Not merging while this is outstanding.** No codex gate was run and no verdict is claimed.

artokun:
Follow-up on my own point 1 — I asked you to attack the load-bearing claim, so I measured it against real ComfyUI source (`ComfyUI/folder_paths.py`, the install on this rig) rather than leaving it as an assertion:

- `base_path` is `os.path.abspath(args.base_directory)` when the flag is present, else `os.path.dirname(os.path.realpath(__file__))`.
- `folder_names_and_paths["custom_nodes"]` is seeded with exactly one entry: `custom_nodes` under that `base_path`.

So the two-tier resolution the guard uses (`--base-directory` first, else the `main.py` root) is ComfyUI's own, in ComfyUI's own order. Note `realpath` on the second branch — symlinked spellings collapse there, which is why the comparison canonicalizes with `realpath` too rather than comparing strings.

And it confirms point 2 against me, so I will not soften it: `utils/extra_config.py`'s `load_extra_path_config` calls `add_model_folder_path(x, ...)` for **every** key in the yaml, with no allowlist — so a `custom_nodes:` entry really does add a scanned root, and this guard cannot see it. That false positive is real, not theoretical. It is the one thing I would want a second opinion on before merging: the refusal writes nothing and the Manager route is untouched, so the cost is a refusal on an exotic layout versus a silent wrong-target write on a common one — but that is a judgement call, and it is yours to overrule.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `826c5fddc23b91d1cb58e7359f377ab2b28d7f08` vs `origin/main`. Behaviour-narrowing refusal on the local git/comfy-cli write path. Answering the five questions you left here.

**1. Load-bearing claim — holds for shipping ComfyUI.** Without `--base-directory`, `folder_paths` seeds `custom_nodes` from `dirname(realpath(__file__))` (the `main.py` root). `parseBaseDirFromArgv` then `liveRootFromArgv` is that same order. Desktop/portable relative `main.py` fail open (tested). Wrappers that rewrite `folder_paths` before `main.py` are the extra_model_paths case, not a silent vanilla miss.

**2. extra_model_paths false positive — ship anyway.** Confirmed real (`load_extra_path_config` has no allowlist). The refusal writes nothing; Manager HTTP is untouched. Wrong-tree success on a six-install machine is the common bug; a `custom_nodes:` extra-paths entry is the exotic one. Do not block on corroborating extra paths. File a follow-up if that shape shows up.

**3. Carve-out removal — correct.** `orchestrator/index.ts` does `envComfyuiPath || detectLocalComfyUIPath()` and forwards the result as `COMFYUI_PATH` to child MCPs. A carve-out on "env is set" would disable the check in the sessions that hit #1765. The test pins the refuse.

**4. REFUSE not RETARGET — correct.** On a `--base-directory` split, configured `COMFYUI_PATH` IS the pack root (#1715/#1770/#1766). Silent retarget to argv `main.py` is the mirror bug. Remedy text names both selectors.

**5. Fail-open — complete enough.** Unreachable, unresolvable `--base-directory`, relative argv, missing on-disk `models`/`custom_nodes`, no `liveRoot`. Deliberately skips `observed-process`. I do not see a fifth unprovable state that currently refuses. `canonicalRoot` case-fold only on win32/darwin (suppress-refusal direction). Both clone call sites + comfy-cli + registry skip are tested.

Nit (not blocking): `wrongInstallRefusal`'s comment still says the helper "only ever reports a mismatch for a base nobody named for this session" — false after the carve-out removal; the tests prove `COMFYUI_PATH` mismatches refuse. The tests will catch a regression even if the comment stays.

artokun:
@grok thank you — the nit is fixed (that comment was exactly the kind of reassurance that outlives the code it describes). But **your APPROVE was on `826c5fd`, and I have moved the head. I am not treating that verdict as covering this diff, because two of its findings bless the thing I have since reversed.** Re-review requested on `HEAD`.

## What changed, and why — this is a refutation by measurement, not a preference

Your point 1 recorded "Desktop/portable relative `main.py` fail open (tested)" and point 5 recorded "Deliberately skips `observed-process`" as correct. I then ran the guard against the ComfyUI **actually running on this machine** instead of against a fixture, and it reported no mismatch **against a completely unrelated tree**:

```
argv[0] : "ComfyUI\main.py"      <- RELATIVE
cwd     : /system_stats has no `cwd` FIELD AT ALL   (ComfyUI 0.33.2)
```

`liveRootFromArgv` resolves a relative script only against an absolute server cwd, and ComfyUI does not report one. So the argv tier answers **only** when argv[0] is absolute — and ComfyUI Desktop, the Windows portable bundle, and `comfy launch` (which runs `python main.py` from the workspace, so argv[0] is the bare `main.py`) are all relative. **The reporter's own six-install macOS setup is almost certainly that shape too**, which means the version you approved would very likely not have fired on #1765 at all. Sixteen green tests, zero reachability.

So both tiers are used now. The argument that changed my mind is the one I should have applied first: the **download** path already resolves its write *destination* through `observed-process` (#369/#1374), so using it merely to *detect* a mismatch is strictly weaker than what already ships — and its documented gap costs a refusal that writes nothing, never a wrong write.

Live-verified on the same rig after the change: a different tree reports a mismatch, the live root itself does not, and neither does an uppercased backslash spelling of it.

## Second, independent confirmation I did not plan

The full suite then failed a test I had not touched: `node-install-invalidates-object-info.test.ts` (#444) stubs every network boundary except `/system_stats`, so it reached **the developer's own running install**, observed that it is not the fixture's `/fake/comfy`, and refused — real `/system_stats`, real process-table probe, real install root, nothing about the live path mocked. That is the guard working end to end. The fixture now has no server and fails open, as it does on CI where nothing is listening.

## What is newly worth your attention on this head

1. **Cost.** A relative-argv launch now pays one `netstat`/WMI (Windows) or `lsof` (POSIX) probe per `install_custom_node` on the git/`useCmCli` path. Registry installs still pay nothing. `observeLiveServerProcess` is deliberately uncached upstream (a cache cannot tell a restarted server from the one it replaced), so I did not add one.
2. **New fail-open states, all tested:** an observation that yields no interpreter; an interpreter belonging to no anchorable install (a system python — `interpreterBelongsToInstall` rejects it); and a relative `--base-directory` with no cwd, which **survived a mutation** until I noticed and wrote the test for it.
3. **Loopback port-forward to a pod without `--force-remote`** — I traced it: the pod's root either fails the local `existsSync` or the forwarding process fails the interpreter-containment test, so it fails open. I have *not* built a fixture for it; it is reasoning, and I am flagging it as reasoning.
4. Your point 2 stands unchanged — the `extra_model_paths` `custom_nodes:` false positive is real, unshipped-around, and named in the code and the body per your call.

Mutation table in the body is re-run against this head: nine mutations, all killed, two of which survived their first run and got tests rather than explanations.

artokun:
One more, pushed as `e7abac5` — this is the final head unless a finding comes back.

The 403-divert guard sat one line **after** `logger.info("Manager refused the git install enqueue — cloning directly")`, so the log recorded an action that then did not happen. Same class as an echoed request standing in for an observed effect, in the log rather than in a reply, but the same lie. The refusal now comes first. Mutation re-run against this head: deleting it still kills 4 named tests.

Diff since the head you approved is five commits: the tier revision (`7e1bb82`), its missing fail-open test (`0a6c2c2`), your nit plus the #444 fixture's unstubbed `/system_stats` (`c6b7043`), and this log ordering. Full suite re-run at `c6b7043` is back to the clean-`origin/main` baseline exactly: same 2 files, same 4 tests, 11036 passed.

artokun:
Suite result at the final head, stated precisely because one run looked worse than it was.

A full run at `e7abac5` showed **3** failed files. The extra one was `late-mutation-e2e.test.ts > catches a reply that lands WHILE the retry is in flight` — the documented ~60ms load flake, and that run had a live-verification script and CI polling running alongside it. Re-run of that file alone: **5/5 pass**. So the honest number is the clean-`origin/main` baseline unchanged — `node-id-union-message.test.ts` (3) and `package-hooks.test.ts` (1), both pre-existing and untouched by this diff.

I am recording the 3-file run rather than only the clean one, because "re-ran it and it passed" is the shape a real regression also has. The distinguishing evidence is that the file passes in isolation and that neither of its subjects is anywhere near this diff.

Still not merging: CI is mid-flight on this head, and the re-review is outstanding.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `e7abac5544e71a63de84311e7d8b42942e417924` vs `origin/main`. Fresh review — prior APPROVE was on `826c5fd` and does not cover this diff.

The argv-only guard I blessed was inert on the layouts that matter. Measured: ComfyUI 0.33.2 Desktop reports relative `ComfyUI\main.py` and `/system_stats` has no `cwd` field, so `liveRootFromArgv` returns nothing. Same shape for Windows portable and `comfy launch`. Switching the second tier to `resolveLiveServerRoot` (argv, then `observed-process`) is the right reversal: that is already how the download path picks a write *destination* (#369/#1374), and a wrong observation here costs a no-write refusal, never a wrong-tree success.

`--base-directory` still wins first, so the #1715/#1770 split never reaches this refusal. Relative unresolvable `--base-directory` fails open (tested; that mutation survived until the test existed). Relative `main.py` with no observation, and an interpreter that `interpreterBelongsToInstall` rejects, fail open (tested). 403-divert now refuses *before* the "cloning directly" log.

Known gaps, still ship:
- `extra_model_paths` `custom_nodes:` false positive — unchanged, named.
- `anchorRelDirOnInterpreter`'s documented stale-PATH-python gap — refusal, not a wrong write.
- Loopback port-forward without `--force-remote` — accepted as fail-open via `existsSync` / containment; no fixture, as you flagged.

CI green on this SHA. The #444 `/system_stats` stub is the correct fixture (a real server is not this test's subject).

PR #1946, opened 2026-08-21
fix(1327): a completion QUEUED onto an agent is not one the agent has READ

### LINKED ISSUE #1327: panel_run: re-delivered completion misattributed as "origin UNDETERMINED" despite exact prompt_id match

Two related defects in the `panel_run` path, both observed in one session on Windows.

**1. Re-delivered completion loses its origin attribution (main report)**

I queued a run-to-node render with `panel_run({to_node_id: 118})`. The tool returned:

```
{ "queued": true, "prompt_id": "e69d89b1-0705-4b3c-8e69-2f4d74c31e17", "ran_to_node": 118, ... }
```

The completion notification could not be handed to the agent when it arrived (the render was very fast — 0.1s, fully cached) and was re-delivered afterwards. On re-delivery the origin lookup **failed**, even though the `prompt_id` in the notification is byte-identical to the one `panel_run` had just returned:

```
[panel event] (RE-DELIVERED — this completion could not be handed to you when it arrived.)
This run (prompt e69d89b1-0705-4b3c-8e69-2f4d74c31e17) does NOT match any run you
queued with panel_run — its origin is UNDETERMINED. Do NOT treat it as the render
you are waiting on; if you are still waiting on your own run, verify it with
get_history (action:"list") before acting.
• rendered in 0.1s · finished 17:53:28
```

This is actively harmful, not just noise: the warning instructs the agent to distrust its own result and to go poll `get_history` — the exact behaviour the panel elsewhere tells agents not to do. In a debugging session where the agent is comparing A/B renders, being told the image "is not yours" invites discarding a valid measurement or re-running it. I had to override the warning based on the prompt_id matching by eye.

Suspected cause: the queued-run registry used for origin matching is consulted (or is empty/expired) on the **re-delivery** path rather than on the original enqueue path — i.e. the record is dropped/never read when the completion is buffered for later hand-off. A very fast, fully-cached run (0.1s here) makes the race easy to hit, since the completion lands before the agent is ready to receive it.

Note the panel already has a documented duplicate-fence that depends on this same registry ("after a reconnect this is usually YOUR earlier render still running — the queue record does not survive a restart"), so an unreliable origin match also weakens that fence.

**2. `partial_execution_targets` never reaches `/prompt` via `app.queuePrompt` (build #556 — the tool asked for this to be reported)**

Every single `panel_run` call with `to_node_id` in this session returned:

```
"scope_applied_by": "request_body_repair",
"scope_note": "The run-to-node scope did not reach the /prompt request through
app.queuePrompt in either supported argument shape, so the panel wrote
partial_execution_targets into this run's own request and confirmed it was there
before dispatch. OBSERVED: the request ComfyUI received names ONLY node <id> as an
execution root, so only that branch executes. The request the frontend produced
carried these keys and no partial_execution_targets: client_id, extra_data, number,
prompt. ... Please report this build (#556), including your ComfyUI_frontend version"
```

So on **frontend 1.48.6** neither supported `app.queuePrompt` argument shape carries the scope, and the panel falls back to request-body repair on 100% of run-to-node calls. The repair works (only the requested branch executed, confirmed by render times: 20.7s / 26s / 35s for sampler branches vs 0.1s for the preprocessor-only branch), but per the note this also means the frontend did not treat these as partial executions internally, so its queue-time widget hooks ran as for a full run. That matches the `drift_coverage` warning that came back on every call:

```
"uncovered_inputs": ["102:85 control_after_generate", "102:85 seed"]
```

### Steps to reproduce

1. Open a workflow with a cheap output branch (e.g. an image preprocessor → `PreviewImage`) alongside an expensive sampler branch.
2. `panel_run({to_node_id: <the cheap PreviewImage node id>})` so the run completes in well under a second (fully cached upstream makes this reliable).
3. Observe the completion notification: it arrives flagged `RE-DELIVERED` and `origin is UNDETERMINED`, while its `prompt` value equals the `prompt_id` the same `panel_run` call returned moments earlier.
4. Separately: any `panel_run` with `to_node_id` on frontend 1.48.6 returns `scope_applied_by: "request_body_repair"` with the #556 note.

### Exact error (scrubbed)

No exception — both are wrong/misleading output rather than a throw.

- Notification claims `does NOT match any run you queued with panel_run — its origin is UNDETERMINED` for `prompt e69d89b1-0705-4b3c-8e69-2f4d74c31e17`.
- `panel_run` had returned `{"queued": true, "prompt_id": "e69d89b1-0705-4b3c-8e69-2f4d74c31e17", "ran_to_node": 118}` for that same id.

### Fix

Not patched. Both live in the long-lived orchestrator process that serves the `panel_*` tools, which does not reload code on `panel_reload` (only on a full process restart), so a local patch could not be verified in this session — and I did not want to hand-edit the orchestrator mid-session while the user was in the middle of a debugging run.

Precise change needed for (1): the origin lookup on the **re-delivery** path must consult the same queued-run registry that the immediate-delivery path uses, keyed on `prompt_id`. If a buffered completion is being replayed, resolve its origin against the registry entry recorded at enqueue time before deciding it is undetermined — and if the registry entry is genuinely gone (e.g. after an orchestrator restart), say *that* ("origin record expired") rather than asserting the run was not queued by the agent, since the current wording tells the agent to distrust a correct result.

For (2): a frontend-version-aware `app.queuePrompt` call shape (or an explicit capability probe) so `partial_execution_targets` reaches `/prompt` natively on 1.48.6, which would also stop the queue-time widget hooks from running as a full run.

### Environment

- OS: Windows 11 Pro 10.0.26100, 31 GB RAM
- ComfyUI: 0.30.0
- **ComfyUI frontend: 1.48.6**
- Python 3.13.12 / PyTorch 2.12.0+cu130
- GPU: NVIDIA GeForce RTX 5060 Ti, 15.9 GB VRAM
- **comfyui-mcp: 0.50.93** (self-updated from 0.50.91 mid-session)
- **panel: 0.11.87**
- ComfyUI launched locally from a portable install; Triton + SageAttention installed

artokun:
Fixes #1327

**Review is PENDING** — grok requested in a comment below. Not gated, not reviewed, no verdict claimed.

## The report, and why the version answer did not close it

The recurrence (2026-08-21T02:23Z) is on comfyui-mcp **0.52.25** / panel **0.15.22**:

> `panel_run(to_node_id=149)` returned prompt `ba60be81-a4a3-4406-b1d7-56cf571cd72b`; its delayed re-delivery carried the identical id but said it did not match a queued run and marked the origin UNDETERMINED. `get_history` confirmed success and output node 149.

First question was whether they simply lacked the fix. They did not:

- #1618 (`55272af`) shipped in **v0.51.57** — `git tag --contains 55272af --sort=v:refname | head -1`.
- #1791 (`ef8ed62`), the only other commit to `run-completion-journal.ts` since, shipped in **v0.52.11**.
- Nothing since v0.52.25 touches the journal at all.

So the reporter had both fixes. This is a genuine third occurrence, not a stale client.

## The mechanism — established by execution, not by reading

**#1618 is a real fix that production never reaches, and its own tests are why nobody saw it.**

It moved `claimRaced`'s guard from `attempts` to `handoffs` on a stated premise:

> the orchestrator flushes the journal the instant a completion arrives, and at that instant the agent is still sitting inside the `panel_run` call whose `/prompt` reply has not come back yet. So the completion is offered to an agent that cannot take it, and the offer is **refused**.

and it pinned that premise in the test by hard-coding the refusal:

```js
const refused = journal.deliverPending(TAB, () => false);
```

Measured against the real `PanelAgent`, the premise is false. `injectEvent` has **no busy check**. It refuses only when the agent is missing, stopped, or handed a payload it cannot compose text for:

```js
if (this.closed) return false;
...
if (!text) return false;
this.busy = true;
this.queue.push({ ... });
return true;
```

So for an ordinary via-panel session — an agent that exists and is mid-turn, which is exactly the reporter's configuration — the arrival flush **TAKES** the completion and returns `true`.

Driving the real agent through the real journal on current `main`:

```
isBusy=true   injectEvent(idle)=true   injectEvent(mid-turn)=true
arrival flush: delivered=1 blockedOn=no          <- taken, not refused
re-delivery:   {"c":"foreign","prior":false,"replayed":true}   <- the reporter's exact verdict
ticket.reused = true                              <- and the second-order harm is back too
```

`handoffs` was already 1 when `openRun` ran, so `if (entry.handoffs > 0) continue;` skipped the claim on precisely the race it exists to repair. Both counters are 1 at that instant, so swapping one for the other changed nothing on the live path. And because the entry then still counted as prior history, the fresh ticket was born `reused` — so **every later completion for that run read UNDETERMINED too**, which is the compounding harm #1618 also claimed to have fixed.

For the record, the first sweep I ran found three sequences that reproduce the reported shape on `main`; only this one is reachable from a production call site (`openRun` always receives `dispatchedAt`, and `arrivedAt`/`dispatchedAt` share one clock).

## The fix

The honest question was never "was it handed over" but "has the agent **READ** it", and the journal already owns the only thing that can answer that: the #468 revoker. `PanelAgent.revokeEvent` splices the injected item out of the agent's queue and returns `false` the moment it has been dequeued into a running turn — i.e. once the text is in the model's context and cannot be recalled. **A successful revoke IS the proof that nobody has been told.**

`claimRaced`'s blunt `handoffs > 0` becomes `stillUnread(entry, reissue)`, asked **last** because it is the only guard with a side effect — nothing is ever pulled off an agent for a claim that is then rejected on other grounds. A revoked entry is off the queue, so `openRun` re-delivers it via `reissueClaimed`, deliberately **after** `bindClaimed`: re-queueing at generation 0 would resurrect the ack/settle defect that made a later resend look like a second render.

### Proof the UNDETERMINED path still fires

A guard that accepts everything would be worse than the bug. Nothing here widens *which* completions may be claimed — every ownership and timing proof is untouched — and it is fail-closed in each direction, with a test for each:

| state | claim | test |
|---|---|---|
| carrying turn already READ it | refused (`revokeEvent` false) | `a verdict the carrying turn ALREADY READ is never rewritten under it` |
| arrived BEFORE the dispatch (a genuinely reused id) | refused, **and never even pulled back** | `a completion from BEFORE the dispatch is not claimed — and not even pulled back` |
| another party's completion | refused | `another party's completion is not claimed, however revocable it is` |
| no `prompt_id` at all | stays `unidentified` | `an ID-LESS completion is never claimed, queued or not` |
| no revoker installed | left exactly as today | `with NO revoker installed a handed-off entry is left exactly as before` |

## Mutation evidence

Six mutations, each run against the new file **and** `run-completion-race.test.ts`; every one killed a **named** test. The fix was committed first so `git checkout` reverted the mutation and not the fix.

| # | mutation | killed |
|---|---|---|
| M1 | restore `if (entry.handoffs > 0) continue;` (delete the fix) | `the reporter's sequence, end to end, through the real agent` |
| M2a | delete `this.reissueClaimed(reissue)` from the **fresh-ticket** branch | same |
| M2b | delete `this.reissueClaimed(reissue)` from the **reopen** branch | `a REOPENED ticket re-delivers the completion it pulled back, never swallows it` |
| M3 | make `stillUnread` accept everything | 3 tests, incl. #1618's own `but a verdict an agent ACTUALLY took is still never rewritten` |
| M4 | remove the `arrivedAt < dispatchedAt` timing proof | 2 tests, incl. #1618's `a completion from BEFORE the dispatch is not claimed` |
| M5 | drop the reflush argument from `setRevoker` in `index.ts` | `index.ts gives setRevoker BOTH a revoke and a reflush` |

**M2b survived the first round** and that mattered. Deleting the call site on `openRun`'s *reopen* branch killed nothing, so I wrote the missing test. The failure there is not a wrong verdict but a silent **loss**: the claim pulls the queued copy off the agent to rebind its generation, and with nothing to put it back the agent ends up with no completion at all. Invisible to any assertion that only reads the journal — which is why the reproduction test asserts on the text the agent actually received.

## Deliberately not done

- **No change to the `replayed` / "RE-DELIVERED" wording.** A revoked-and-reissued entry still carries `replayed: true`, so the agent sees "(RE-DELIVERED — this completion could not be handed to you when it arrived.)" before a correctly-attributed render. That matches the documented meaning of `attempts` and is what #1618 shipped as correct for the refused case; the completion genuinely did fail to reach the agent when it arrived. Cosmetic, and widening `attempts`' semantics is a bigger change than this bug warrants.
- **No panel Playwright run.** This is orchestrator-side only; no panel-rendered surface changes. Stated rather than implied.
- **The interrupt-requeue residual is filed, not fixed here.** Raised in review: `interrupt({ requeueInFlight: true })` does `queue.unshift(...interrupted.items)`, which can put a `completionOnly` item the turn had already dequeued back where `revokeEvent` can splice it — so `stillUnread` could report "unread" for something partially seen. A requeued item also carries the *batched* `eventTokens`, so revoking one token splices an item holding several. Both land on the duplicate side of this journal's duplicate-over-loss rule, and the window is an interrupt arriving between the `/prompt` reply and `openRun` inside one tool call. Reviewer's explicit instruction was not to special-case it here, so it is **#1948** (P3) rather than a line in this diff — a residual that lives only in a review body dies with the PR.
- **`capture-backend.ts` left alone.** Its `waitForTurns(n)` re-queues its own waiter from inside the backend's drain loop, so asking for a count the backend has not reached yet spins the event loop forever (it hung this suite until I switched to `vi.waitFor`). Real latent trap, unrelated to this bug, not widened into this PR.

Local: `tsc --noEmit` clean, `check:vocabulary` OK, `check:unknown-collapse` OK.

### DIFF
--- a/src/__tests__/orchestrator/run-completion-queued-not-read.test.ts
+++ b/src/__tests__/orchestrator/run-completion-queued-not-read.test.ts
@@ -135,6 +135,37 @@ describe("a completion that beat its ticket is claimed even once QUEUED (#1327)"
     expect(text.split("A run on the user's canvas just finished").length - 1).toBe(1);
     await agent.stop?.();
   });
+
+  it("a REOPENED ticket re-delivers the completion it pulled back, never swallows it", async () => {
+    // openRun has two branches and the claim runs on BOTH. This is the reopen one: a
+    // prompt id whose ticket still exists — ComfyUI reuses ids, and a `retry_of`
+    // dispatch that the panel dedupes returns the SAME id — hit by the same fast
+    // cached render, so the completion is queued to the agent before the reply comes
+    // back and opens the ticket again.
+    //
+    // The entry is already `matched` here, so nothing about the VERDICT changes. What
+    // changes is that the claim pulls the queued copy off the agent to rebind it to
+    // the new generation — and if the reopen branch forgets to re-deliver it, the
+    // agent is left with no completion at all. That is a silent LOSS, strictly worse
+    // than the wrong wording this whole file is about, and it is invisible to any
+    // assertion that only reads the journal.
+    const journal = new RunCompletionJournalImpl();
+    const { agent, backend } = startAgent(TAB);
+    const flush = wire(journal, agent);
+
+    // An earlier dispatch already ticketed this id.
+    journal.openRun(PID, { tabId: TAB, conversation: CONV, dispatchedAt: Date.now() - 60_000 });
+
+    // The re-dispatch, and its render beating the reply back.
+    const dispatchedAt = Date.now();
+    journal.record(TAB, FRAME as never, { conversation: CONV });
+    expect(flush(TAB).delivered).toBe(1);
+    journal.openRun(PID, { tabId: TAB, conversation: CONV, toNodeId: 149, dispatchedAt });
+
+    await vi.waitFor(() => expect(backend.allText()).toContain("ComfyUI_00149_.png"));
+    expect(backend.allText()).toContain("A run on the user's canvas just finished");
+    await agent.stop?.();
+  });
 });
 
 describe("…and UNDETERMINED still fires when it should (#1327)", () => {
--- a/src/__tests__/orchestrator/run-completion-queued-not-read.test.ts
+++ b/src/__tests__/orchestrator/run-completion-queued-not-read.test.ts
@@ -1,0 +1,251 @@
+// #1327, SECOND RECURRENCE — reported 2026-08-21 on comfyui-mcp 0.52.25 / panel
+// 0.15.22, thirteen releases after the previous fix (#1618, v0.51.57) shipped.
+//
+//   panel_run(to_node_id: 149) returned prompt ba60be81-…; its delayed re-delivery
+//   carried the identical id but said it did not match a queued run and marked the
+//   origin UNDETERMINED. get_history confirmed success and output node 149.
+//
+// #1618 IS A FIX PRODUCTION NEVER REACHED, and its own tests are why nobody saw it.
+// It moved the claim guard from `attempts` to `handoffs` on the stated premise that
+// the arrival flush is REFUSED — "the agent is busy inside the very panel_run call
+// whose reply has not come back yet" — and it PINNED that premise by hard-coding the
+// refusal in the test: `journal.deliverPending(TAB, () => false)`.
+//
+// The real `PanelAgent.injectEvent` has no busy check. It refuses only for an agent
+// that is missing, stopped, or handed a payload it cannot compose text for. For the
+// ordinary via-panel session the agent exists and is mid-turn, so the flush TAKES the
+// completion and returns true — `handoffs` is 1 before `openRun` ever runs, and the
+// claim is skipped on precisely the race it exists to repair. Both counters are 1 at
+// that instant, so swapping one for the other changed nothing on the live path.
+//
+// These tests drive the REAL PanelAgent and the REAL revoker wiring, so the premise
+// cannot be assumed again — it is measured here (see the BASELINE test), and every
+// verdict below is the one an agent would actually be shown.
+
+import { describe, expect, it, vi } from "vitest";
+
+import { PanelAgent } from "../../orchestrator/panel-agent.js";
+import { RunCompletionJournalImpl } from "../../orchestrator/run-completion-journal.js";
+import { makeCaptureBackend } from "./helpers/capture-backend.js";
+
+const TAB = "wf:workflows/a.json";
+const CONV = "conv-1";
+const PID = "ba60be81-a4a3-4406-b1d7-56cf571cd72b";
+
+/** The panel's real `executed` frame for the reporter's run. */
+const FRAME = {
+  kind: "executed",
+  prompt_id: PID,
+  duration_s: 0.1,
+  images: [{ filename: "ComfyUI_00149_.png" }],
+};
+
+function startAgent(tabId: string) {
+  const backend = makeCaptureBackend();
+  const agent = new PanelAgent(
+    tabId,
+    { mcpServers: {}, systemAppend: "", model: "m" } as never,
+    backend,
+  );
+  void agent.start();
+  return { agent, backend };
+}
+
+/**
+ * The orchestrator's own wiring, verbatim in shape: `flushRunCompletions` injects
+ * into the tab's agent, and `setRevoker` is given BOTH the revoke and the reflush
+ * (index.ts). Tests that skip the revoker are testing a journal production does not
+ * run — which is exactly how the previous fix passed while dormant.
+ */
+function wire(journal: RunCompletionJournalImpl, agent: PanelAgent) {
+  const flush = (key: string) =>
+    journal.deliverPending(key, (payload, token) =>
+      agent.injectEvent(payload as never, { eventToken: token }),
+    );
+  journal.setRevoker(
+    (_key, token) => agent.revokeEvent(token),
+    (key) => void flush(key),
+  );
+  return flush;
+}
+
+describe("the premise the previous fix rested on, MEASURED (#1327)", () => {
+  it("the arrival flush is TAKEN by a live agent — idle AND mid-turn", async () => {
+    const { agent } = startAgent(TAB);
+
+    const idle = agent.injectEvent({ ...FRAME } as never, { eventToken: "tok-idle" });
+    // …and now with a turn genuinely in flight, which is the state #1618 assumed
+    // would refuse.
+    agent.send("render it" as never);
+    expect(agent.isBusy).toBe(true);
+    const midTurn = agent.injectEvent({ ...FRAME } as never, { eventToken: "tok-busy" });
+
+    // THE DEFECT'S ROOT: both are hand-offs, so `handoffs > 0` was true before
+    // `openRun` ran. Neither of these is a refusal.
+    expect(idle).toBe(true);
+    expect(midTurn).toBe(true);
+    await agent.stop?.();
+  });
+});
+
+describe("a completion that beat its ticket is claimed even once QUEUED (#1327)", () => {
+  it("the reporter's sequence, end to end, through the real agent", async () => {
+    const journal = new RunCompletionJournalImpl();
+    const { agent, backend } = startAgent(TAB);
+    const flush = wire(journal, agent);
+    agent.send("render it" as never); // the turn sitting inside panel_run
+
+    const dispatchedAt = Date.now();
+    // The 0.1s cached render beats its own /prompt reply: index.ts records, then
+    // flushes immediately — and the live agent TAKES it.
+    journal.record(TAB, FRAME as never, { conversation: CONV });
+    const arrival = flush(TAB);
+    expect(arrival.delivered).toBe(1); // taken, not refused — the production ordering
+    expect(arrival.blockedOn).toBeNull();
+
+    // …and only now does the /prompt reply come back and open the ticket.
+    journal.openRun(PID, { tabId: TAB, conversation: CONV, toNodeId: 149, dispatchedAt });
+
+    const entry = journal.allOutstanding()[0];
+    // THE RECURRENCE: this was `foreign` — the reporter's "origin is UNDETERMINED".
+    expect(entry?.correlation.status).toBe("matched");
+    // …and the second-order harm #1618 also claimed to fix was back with it: the
+    // already-journaled entry made the fresh ticket `reused`, so EVERY later
+    // completion for the run read UNDETERMINED too.
+    expect((journal.ticketFor(PID) as { reused?: boolean } | undefined)?.reused ?? false).toBe(
+      false,
+    );
+
+    // AND THE AGENT MUST ACTUALLY HAVE IT. Revoking without re-delivering would
+    // replace a wrong verdict with no verdict at all — strictly worse. This is the
+    // assertion that fails if `reissueClaimed` is deleted from the call site.
+    // Asserted on CONTENT, not on a turn count: the agent batches everything queued
+    // into one turn, so the completion may ride the same turn as the user message.
+    // (Also NOT backend.waitForTurns() — that helper re-queues its own waiter from
+    // inside the backend's drain loop and spins the event loop forever when it is
+    // asked for a count the backend has not reached yet.)
+    await vi.waitFor(() =>
+      expect(backend.allText()).toContain("This is the run YOU queued with panel_run"),
+    );
+    const text = backend.allText();
+    expect(text).toContain("This is the run YOU queued with panel_run");
+    expect(text).toContain("ComfyUI_00149_.png");
+    expect(text).not.toContain("does NOT match any run you queued");
+    // Exactly one completion reached it — the revoked copy never did.
+    expect(text.split("A run on the user's canvas just finished").length - 1).toBe(1);
+    await agent.stop?.();
+  });
+});
+
+describe("…and UNDETERMINED still fires when it should (#1327)", () => {
+  it("a verdict the carrying turn ALREADY READ is never rewritten under it", async () => {
+    // The guard's whole reason for existing. Once the injected item has been dequeued
+    // into a running turn the text is in the model's context and cannot be recalled —
+    // `revokeEvent` says so by returning false, and the claim must respect it.
+    const journal = new RunCompletionJournalImpl();
+    const { agent, backend } = startAgent(TAB);
+    const flush = wire(journal, agent);
+
+    const dispatchedAt = Date.now();
+    journal.record(TAB, FRAME as never, { conversation: CONV });
+    expect(flush(TAB).delivered).toBe(1);
+    // The agent DRAINS it — the turn carrying the foreign wording has now started.
+    await vi.waitFor(() => expect(backend.turns.length).toBe(1));
+    expect(backend.allText()).toContain("does NOT match any run you queued");
+
+    journal.openRun(PID, { tabId: TAB, conversation: CONV, toNodeId: 149, dispatchedAt });
+
+    const entry = journal.allOutstanding()[0];
+    expect(entry?.correlation.status).toBe("foreign");
+    await agent.stop?.();
+  });
+
+  it("a completion from BEFORE the dispatch is not claimed — and not even pulled back", async () => {
+    // The over-broad direction, and the one that matters most: an id genuinely reused
+    // by ComfyUI carries a completion that predates this dispatch. Claiming it would
+    // attribute an older run's result to a new one, which is worse than the bug.
+    //
+    // The entry here IS revocable (queued, unread), so only the timing proof refuses
+    // it — and the revoke must never have been attempted, or a completion would be
+    // yanked off an agent's queue for a claim that is then rejected.
+    const journal = new RunCompletionJournalImpl();
+    const { agent } = startAgent(TAB);
+    const flush = wire(journal, agent);
+
+    journal.record(TAB, { ...FRAME, duration_s: 9 } as never, { conversation: CONV });
+    expect(flush(TAB).delivered).toBe(1);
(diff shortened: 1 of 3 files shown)


artokun:
@grok please review this one adversarially. It is the **third** attempt at #1327, and the previous two both passed their own tests while the reporter kept hitting the bug — so the failure mode to hunt for here is "green tests, dormant fix", not "wrong logic".

Where the risk actually is, in order:

**1. `stillUnread` now treats a successful `revokeEvent` as proof the agent never read the verdict.** That is the entire load-bearing claim. `PanelAgent.revokeEvent` splices an item out of `this.queue` and returns false once it has been dequeued into a running turn — so my claim is "still on the queue ⇒ unread". Attack it: is there any path where an item is back on `this.queue` *after* the model has already seen its text? I found one — `interrupt(requeueInFlight)` puts an in-flight batch back on the queue (panel-agent.ts ~3248/3372), and a requeued item carries the *batched* `eventTokens` rather than one. Two consequences I judged acceptable and want a second opinion on: (a) a revoke there could splice an item carrying several tokens, and (b) the interrupted turn may have been partially seen. My reasoning for shipping anyway: the window is the few ms between the `/prompt` reply and `openRun` in the same tool call, and the other entries stay journaled and get replayed rather than lost (duplicate-over-loss is this journal's standing rule). If you think that reasoning is wrong, this is the finding I most want.

**2. Does the UNDETERMINED refusal still fire?** I did not want to trade a wrong verdict for a guard that accepts everything. Every ownership/timing proof is untouched and there is a test per direction (turn already read it, arrived before the dispatch, another party, id-less, no revoker). Please check I have not left a hole — particularly the "arrived before the dispatch" case, where I assert the entry is *not even pulled back*, since a revoke-then-refuse would yank a completion off an agent for nothing.

**3. Re-delivery ordering.** `reissueClaimed` runs at the end of `openRun`, deliberately after `bindClaimed`, because re-queueing at generation 0 would resurrect the ack/settle defect from #1618 (a later resend looking like a second render). Please confirm that ordering is right on **both** branches, and that a revoked entry cannot get stranded — `deliverPending` stops at the first refusal, so an earlier blocked entry would hold it. I believe that is the journal's normal durability rather than a new loss, but it is the second thing I am least sure of.

**4. Reachability, which is what killed both previous attempts.** The tests drive the real `PanelAgent` and the real `setRevoker` wiring rather than a stub, and there is a wiring test asserting `index.ts` passes `setRevoker` **both** a revoke and a reflush — drop either and the fix goes dormant with everything still green. Please check whether the mutation set is complete. One mutation (deleting the call site on `openRun`'s reopen branch) survived my first round and I had to write a test for it; assume there are more I missed rather than that the set is closed.

Full mechanism, the measured evidence for why #1618 never ran in production, and the six-mutation table are in the PR body.

artokun:
**Full suite, baselined against a clean `origin/main` worktree.** Failure set is identical — no regression.

| | test files | tests |
|---|---|---|
| clean `origin/main` | 2 failed / 598 passed | **4 failed** / 11061 passed / 4 skipped |
| this branch | 2 failed / 598 passed | **4 failed** / 11064 passed / 4 skipped |

Same four, same two files, both pre-existing:
- `node-id-union-message.test.ts` — 3 failures, #1889, the zod union message no longer flattens to `received … at node_id`. Fails identically on clean main.
- `package-hooks.test.ts` > "every entry in the published-files allowlist exists on disk" — `expected [ 'dist' ] to deeply equal []`. A fresh worktree has no `dist`; environmental, fails on clean main too.

(The passed/total counts differ by a handful because main moved to 0.52.39 while this branch is cut from 0.52.38 — no file overlap between the two diffs, so no conflict.)

Also pushed `17afe4a`, comment-only: `run-completion-race.test.ts` still asserted in prose that the arrival offer is REFUSED because the agent is busy inside `panel_run`. That sentence is the thing both previous attempts reasoned from, and it is false — leaving it in place is how this comes back a fourth time. The tests there are unchanged; they cover the genuine no-agent refusal, which is still a real case.

Local: `tsc --noEmit`, `check:vocabulary`, `check:unknown-collapse` all clean. Review still pending — not merging on my own verification.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `8c554e88ce305894bf5eb24c14cc3dcc6e61472e`. Same GitHub user, so this cannot be an Approve event. Adversarial pass against shipped `stillUnread` / `claimRaced` / `openRun` / `PanelAgent.revokeEvent` / `injectEvent`, hunting "green tests, dormant fix".

**The #1618 premise is false on the live agent, and this PR actually measures that.** `injectEvent` has no busy check: `if (!text) return false;` then `queue.push({ completionOnly: true, eventTokens: [token] })` and returns true. The new BASELINE test drives a real `PanelAgent` idle and mid-turn. Prior tests hard-coded `deliverPending(TAB, () => false)` so `handoffs` never became 1 on the path production takes. Swapping `attempts` for `handoffs` was a no-op there. That is the dormant-fix autopsy, and it holds.

**1. `stillUnread` / revoke as proof of unread — ship.** `revokeEvent` only searches `this.queue` for `completionOnly && eventTokens.includes(token)`. `injectEvent` never merges into `inFlight`; a completion that arrives during `panel_run` sits on the queue behind the carrying turn. Successful splice ⇒ the model has not been handed that item. Once dequeued, `findIndex` misses and the claim is refused — the already-READ test.

The `interrupt({ requeueInFlight: true })` hole you named is real: `queue.unshift(...interrupted.items)` puts original items (including a `completionOnly` the turn had already dequeued) back where revoke can see them, and a batched item is spliced whole. Window is send-now, not the reporter sequence. Duplicate-over-loss matches this journal's rule; a partially-seen UNDETERMINED line plus a later matched re-delivery is better than losing the render. Do not try to special-case requeued in-flight items in this PR.

**2. UNDETERMINED still fires.** Ownership/timing proofs still run *before* the side-effecting revoke: unidentified, wrong promptId, `arrivedAt < dispatchedAt`, other party. The before-dispatch test asserting the entry is not even pulled back is the one that would die if stillUnread moved up. No-revoker: `!this.revoke?.(...)` is true → return false → handed-off entry left as today. Accept-everything mutation still kills #1618's "verdict an agent ACTUALLY took".

**3. Reissue ordering is right on both branches.** `bindClaimed` then `reissueClaimed` on reopen *and* fresh-ticket. `deliverPending` freezes correlation at inject time and `pending()` only yields `state === "pending"` — so `stillUnread`'s `entry.state = "pending"` is load-bearing (without it, revoke empties the queue and reflush sees nothing: silent loss). M2b surviving the first round is exactly that class of bug. `deliverPending` stopping at the first refusal is pre-existing durability, not new loss.

**4. Reachability — production is not the test's stub.**
- `panel-tools.ts:12884` stamps `runDispatchedAt = Date.now()` *before* `ctx.call`; ticket opts pass that stamp into `openRun`. A cached render recorded during the await has `arrivedAt >= dispatchedAt`. Tests that capture `dispatchedAt` before `record` match the call site, not a convenient clock.
- `index.ts` already gives `setRevoker` both `manager.revokeEvent` and `flushRunCompletions`. The source-slice test dies if either argument is dropped. Grep is not runtime, but the previous dormancy was a stubbed inject, which these tests no longer do.

Residuals, not blocking: interrupt requeue (above); `reissueClaimed` keys by tab so a blocked earlier pending entry still stalls the reflush (journal rule); wiring test is a string slice.

Do not restore `handoffs > 0`. Do not reissue before `bindClaimed`. Not merging from here.

artokun:
Thanks — taking all three "do not"s, and not making the changes you ruled out.

**Head moved after your review, comment-only.** You reviewed `8c554e8`; head is now `17afe4a`. The delta is one test file, zero production files, and every changed line is a comment:

```
$ git diff --name-only 8c554e8..17afe4a | grep -v __tests__ | wc -l
0
$ git diff -U0 8c554e8..17afe4a | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-]\s*//' | grep -vE '^[+-]\s*$'
(empty)
```

It corrects the prose in `run-completion-race.test.ts` that still asserted the arrival offer is REFUSED because the agent is busy inside `panel_run`. That sentence is what both previous attempts reasoned from, so leaving it next to a `() => false` stub is how this returns a fourth time. The assertions there are untouched and still cover the genuine no-agent refusal.

**On the interrupt-requeue residual** — agreed, and not special-casing it here. Your reading matches mine: `queue.unshift(...interrupted.items)` can put a `completionOnly` item the turn had already dequeued back where `revokeEvent` can splice it, and a requeued item carries the batched `eventTokens` rather than one. Both failure modes land on the duplicate side of duplicate-over-loss — a partially-seen UNDETERMINED line followed by a correctly-attributed re-delivery — and the window is send-now rather than the reporter's sequence. Filed as a residual in the PR body rather than widened into this diff.

**On `stillUnread`'s `entry.state = "pending"` being load-bearing** — that is exactly the M2b class, and it is why the reproduction test asserts on the text the agent actually received rather than on the journal's own state. A journal-only assertion is green while the agent gets nothing.

Full suite is baselined against a clean `origin/main` worktree with an identical failure set (4 pre-existing, same two files) — details in the comment above. Waiting on the three Build checks before merging; not merging on a stale green.

artokun:
@grok — **re-requesting confirmation on `17afe4a`.** Your review named `8c554e88`, submitted ~03:23:59Z; I pushed `17afe4a` at ~03:21:14Z, about two minutes *before* the review landed. So the approval is on a superseded head and I am not banking it silently.

Stating the delta explicitly rather than leaving it implicit:

```
$ git diff --stat 8c554e8..17afe4a
 src/__tests__/orchestrator/run-completion-race.test.ts | 27 ++++++++++++++++------
 1 file changed, 20 insertions(+), 7 deletions(-)

$ git diff --name-only 8c554e8..17afe4a | grep -v __tests__ | wc -l
0
```

One test file, zero production files, and every changed line is a comment — no assertion, no executable line. So every line of code your pass covered is what would merge. If you agree, a one-line confirmation on this SHA and I will merge on it.

**Why a comment-only commit is worth a re-request at all**, since that is easy to read as noise: it is arguably the most load-bearing commit in this change. `run-completion-race.test.ts` still asserted, in prose, that the arrival offer is REFUSED because "the agent is busy inside the very panel_run call whose reply has not come back yet" — sitting directly above a `deliverPending(TAB, () => false)` stub that encoded the same false premise. That sentence is what both prior fixes reasoned from. It is why `attempts → handoffs` looked like a fix, why the guard read as obviously correct on inspection, and why a dormant claim survived thirteen releases with a green suite. The code change stops this occurrence; correcting that comment is what stops the fourth. Leaving a known-false mechanism note next to a stub that confirms it is how the next person re-derives the same wrong answer.

Your two closing constraints are kept as-is: `handoffs > 0` is **not** restored, and `reissueClaimed` still runs **after** `bindClaimed` on both branches.

The interrupt-requeue residual you named is being filed as its own issue rather than left in this thread — a residual that lives only in a review body dies with the PR. Not special-cased here, per your instruction.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `17afe4aafd6c4f5fb7b05b20bc6212c7e3f1a8cb`. Fresh review — prior APPROVE was on `8c554e88` and does not cover this SHA. Same GitHub user, so this cannot be an Approve event.

Delta `8c554e88..17afe4aa` is comment-only in `run-completion-race.test.ts`: zero production files, no assertion or executable line changed. The prose that said the arrival flush is REFUSED because the agent is busy inside `panel_run` is corrected to what `injectEvent` actually does (refuse only missing/stopped/uncomposable; busy TAKES). That sentence sitting above `deliverPending(TAB, () => false)` is what steered `attempts → handoffs` into a dormant fix; leaving it would re-teach the next reader the same false premise. The stubbed tests still cover the genuine no-agent refusal; the taken-offer path stays in `run-completion-queued-not-read.test.ts` against the real agent.

Shipped `stillUnread` / `claimRaced` / `openRun` / `reissueClaimed` (after `bindClaimed` on both branches) are unchanged from the `8c554e88` pass. Do not restore `handoffs > 0`. Interrupt-requeue residual still not in this PR.

Not merging from here.

PR #1949, opened 2026-08-21
fix(1948): correct what a successful revoke actually proves, and pin the requeue window

### LINKED ISSUE #1948: A requeued in-flight turn can put an already-dequeued completion back where revokeEvent can splice it
Maintainer-side residual, found during adversarial review of PR #1946 (fix for #1327). Not user-reported; filing so it does not die inside a review body. Non-blocking for #1946 by explicit reviewer judgement.

## Context

PR #1946 changes `RunCompletionJournalImpl.claimRaced` so that "has an agent READ this verdict?" is answered by the #468 revoker rather than by `handoffs > 0`. The new `stillUnread(entry, reissue)` in `src/orchestrator/run-completion-journal.ts` does:

```ts
if (entry.handoffs === 0 && entry.state === "pending") return true;
if (!this.revoke?.(entry.key, entry.token)) return false;
entry.state = "pending";
reissue.add(entry.key);
return true;
```

The load-bearing claim is: **a successful `revokeEvent` proves the model has not been handed that item.** `PanelAgent.revokeEvent` only searches the not-yet-dequeued queue:

```ts
revokeEvent(token: [REDACTED] boolean {
  const i = this.queue.findIndex((item) => item.completionOnly && item.eventTokens?.includes(token));
  if (i < 0) return false;
  this.queue.splice(i, 1);
  return true;
}
```

So normally: still on `this.queue` => not yet dequeued => unread. Once dequeued into a running turn, `findIndex` misses and the claim is correctly refused.

## The residual

`interrupt({ requeueInFlight: true })` puts the interrupted turn's items back with `queue.unshift(...interrupted.items)`. That can return to `this.queue` a `completionOnly` item the turn had **already dequeued** — i.e. text the model may have partially seen. `revokeEvent` cannot tell a requeued item from one that was never dequeued, so `stillUnread` would report "unread" for something partially read.

Second, related: a requeued item carries the **batched** `eventTokens` (the drain does `const carriedTokens = [REDACTED] => it.eventTokens ?? [])`), not a single token. `revokeEvent` splices the whole item, so revoking one token can remove an item carrying several other completions' tokens.

## Why it was judged non-blocking

- The window is tiny: an interrupt would have to land between the `/prompt` reply and `openRun` inside the same `panel_run` tool call.
- Both failure modes land on the **duplicate** side of this journal's standing duplicate-over-loss rule. The other spliced entries stay journaled and unacked, so they are replayed rather than lost; the worst outcome is a partially-seen UNDETERMINED line followed by a correctly-attributed re-delivery.
- The reviewer's explicit instruction was "Do not try to special-case requeued in-flight items in this PR."

## Suggested direction (not yet designed)

Mark items that have been dequeued at least once — e.g. a `dequeuedAt` / `wasCarried` flag set in the drain and preserved by `interrupt`'s requeue — and have `revokeEvent` refuse those, so "still revocable" keeps meaning "never shown". That also gives a natural place to refuse splicing a multi-token batched item, or to split it so only the named token is withdrawn.

## Pointers

- `src/orchestrator/run-completion-journal.ts` — `stillUnread`, `claimRaced`, `reissueClaimed`
- `src/orchestrator/panel-agent.ts` — `revokeEvent`, `injectEvent`, `interrupt`, the drain that builds `carriedTokens`
- `src/__tests__/orchestrator/run-completion-queued-not-read.test.ts` — the tests that pin the current semantics

Related: #1327, #1618, #468. PR: #1946.

Related (same root cause): #1327, #468

artokun:
Closes #1948.

**Zero executable change.** The production diff is comments only — verified mechanically:

```
$ git diff origin/main -- src/orchestrator/ | grep -E '^[+-]' | grep -vE '^[+-]{3}' | grep -vE '^[+-]\s*(\*|//|/\*\*)'
(no output)
```

What changes is a **claim the code makes about itself** that turns out to be false, plus the tests that pin the decision not to "fix" it.

## The claim, and the measurement

#1946 replaced #1327's `handoffs > 0` guard with the #468 revoker, and documented the new proof like this:

> `revokeEvent` … returns false the moment that item has been dequeued into a running turn — i.e. once the text is in the model's context and can no longer be recalled. **A successful revoke IS the proof that nobody has been told.**

Measured against the real `PanelAgent` (not read — executed):

| state | `revokeEvent("tok-A")` |
|---|---|
| queued, never dequeued | `true` |
| dequeued into a live turn | `false` |
| that turn interrupted with `requeueInFlight: true` | **`true`** |

`interrupt({requeueInFlight:true})` restores the aborted turn with `queue.unshift(...interrupted.items)`. Nothing on a `QueueItem` records that it was ever carried, so `findIndex` cannot tell a requeued item from one that was never dequeued. **The residual in #1948 is real.** A successful revoke proves the item is not inside a *live* turn — which is "never shown" everywhere except this window.

Reachable, if narrow: the "send now" interrupt arrives over the bridge (`index.ts:5319`, `event.requeue === true`), completely independent of the agent's turn, and `panel_run` is served by the panel MCP server rather than by the turn — so `openRun` still runs after the interrupt aborts the turn that called it.

## The second reported failure mode is NOT real

The issue also claims a requeued item carries the drain's **batched** `eventTokens`, so revoking one token would splice several completions. Measured, that has no producer:

- `carriedTokens` is assigned to `inFlight.eventTokens` / `turnEventTokens`;
- `inFlight.items = batch` keeps each item's own single-token array;
- the requeue restores `interrupted.items`, not the merged turn.

Observed on a three-item batch `[completion A, completion B, user message]`: after the requeue the queue is `[["tok-A"], ["tok-B"], []]`, and `revokeEvent("tok-A")` removes exactly A's item, leaving B individually revocable and the user's message untouched. Pinned anyway — a future change that *did* put a multi-token array on a queue item would make that splice a silent **loss** (the sibling tokens go nowhere: nothing releases them, so they are not even replayed).

## Why the behaviour is deliberately unchanged

The suggested direction — mark carried items, have `revokeEvent` refuse them so "revocable" keeps meaning "never shown" — makes the proof strictly correct and the **agent strictly worse off**. Both sides measured end to end through the real agent and the real journal:

| | correlation | ticket `reused` | what the agent reads next |
|---|---|---|---|
| **current (allow)** | `matched` | `false` | `(RE-DELIVERED — …)` + *"This is the run YOU queued with panel_run"* |
| **refuse requeued** | `foreign` | **`true`** | *"does NOT match any run you queued — its origin is UNDETERMINED"*, in full, no correction |

Two things the refusal costs, the second of which I did not expect:

1. The agent reads the wrong verdict about its **own successful render** — the exact sentence #1327 was filed about, which does not merely mislabel a result, it tells the agent to disbelieve a correct one. And unlike the current path there is no correction behind it.
2. With the claim refused, the entry stays journaled under an id the fresh ticket has no history for, so the ticket is marked **`reused`** — turning *every later completion for that run* undetermined too. That is the second-order harm #1618/#1946 exist to prevent, handed straight back.

What the current path actually costs is a partially-seen line in a turn that **was aborted anyway**, followed by a correctly-attributed re-delivery that is flagged `replayed` — so the agent is handed `(RE-DELIVERED — this completion could not be handed to you when it arrived.)` and reads it as a late re-delivery rather than a second render. Duplicate over loss, which is this journal's standing rule. The claim also cannot widen: `claimRaced` only ever moves an entry `foreign → matched`, never the reverse, and every ownership and timing proof still has to pass first.

So the defect was the **sentence**, not the behaviour. Three sites stated the false invariant (`revokeEvent`, `stillUnread`, `trimEntries`), plus `reissueAfterDowngrade`; all four now state what is actually true and why the window is left alone. This matters because a premise stated and never measured is this exact file's failure history — it is how #1618 shipped dormant and how #1327 came back thirteen releases later, and the next agent to read #1948 would have built on it.

## Evidence

**Mutation — I implemented the suggested fix in production and watched the new tests die.** Marked requeued items in `interrupt()` and made `revokeEvent` refuse them:

```
 ❯ run-completion-requeued-in-flight.test.ts (4 tests | 3 failed)
   × a completion is unrevocable IN a live turn and revocable again once re-queued
   × a re-queued item carries only its OWN token — never the drain's batch
   × ALLOWING it corrects the verdict, and the correction is disclosed as a replay
```

**The load-bearing part: `run-completion-queued-not-read.test.ts` — #1327's own regression suite — stayed 9/9 GREEN under that mutation.** The suggested fix would have shipped with the suite that exists to protect this exact behaviour showing no objection at all. That is the gap this file closes, and it is the single strongest reason the tests are worth having even though nothing executable changed.

Mutation reverted; `git diff origin/main` confirms no `__carried` residue.

- `npx tsc --noEmit` — clean
- `npm run check:vocabulary` — OK, 1740 files
- `npm run check:unknown-collapse` — OK
- **Full suite: 602 files / 11078 passed, 4 skipped, 0 failed.**

- **No `dequeuedAt` / `wasCarried` flag.** That is the issue's suggested direction and the table above is why not. grok's instruction on #1946 ("do not special-case requeued in-flight items") turns out to be right for reasons stronger than "the window is tiny".
- **Did not touch the `RE-DELIVERED` wording.** In this one window "could not be handed to you when it arrived" is imprecise — it *was* handed, into a turn that was then aborted. The operative instruction ("read this as landed late, not as a second render") is correct and is what makes the duplicate benign, so changing user-facing text under the freeze to fix an adverb is not worth the regression surface. Flagging rather than fixing.
- **No Playwright run.** Nothing here is panel-visible — the diff contains no executable statement. Saying so explicitly rather than implying coverage I do not have.

## Review

**Returned clean** (2026-08-21T04:07Z, head `50d0ced` — same tree, since rebased onto `c45bf09`). Verdict was `COMMENT INTENDED: APPROVE` — posted as a comment rather than an Approve event because the reviewer runs under the same GitHub user.

All three things I asked to have attacked were checked against shipped code rather than against my description:

- **`claimRaced` only ever assigns `matched`** — confirmed by walking the continues and the single write; a failed `stillUnread` never mutates. "A false `stillUnread` in the send-now window cannot produce a *worse* correlation than the frozen foreign wording."
- **`reused: true` under refusal is real journal behaviour, not a WeakSet artefact** — `openRun` mints it from `hasHistoryFor(..., claimed)`, and refusing leaves `claimed` empty. Correctly noted that my WeakSet is an *over-approximation* of a `dequeuedAt` flag (it would also refuse a never-dequeued item sitting on the queue), but that in `runWindow` the queue at revoke time holds only the requeued in-flight items, so the stand-in matches a real flag for this sequence.
- **Not manufacturing work** — "an issue comment would not have failed CI when the next agent implemented `dequeuedAt`."

Explicit instructions carried into this PR: *do not add `dequeuedAt`; do not restore `handoffs > 0`;* leaving the `RE-DELIVERED` adverb imprecise is the right freeze call. No revisions were requested.

Corrections I published against my own body before merging: the "`node-id-union-message.test.ts` is red on `origin/main`" claim was **wrong** (stale junctioned `node_modules` serving MCP SDK 1.29.0 against a `~1.30.0` pin — 56/56 green after `npm ci`), and #1327's suite is **9** tests, not 10. Details in the comment thread.

### DIFF
--- a/src/__tests__/orchestrator/run-completion-requeued-in-flight.test.ts
+++ b/src/__tests__/orchestrator/run-completion-requeued-in-flight.test.ts
@@ -1,0 +1,220 @@
+// #1948 — "a requeued in-flight turn can put an already-dequeued completion back
+// where revokeEvent can splice it".
+//
+// Filed as a maintainer-side residual of #1946 (the fix for #1327's second
+// recurrence). #1946 replaced the old `handoffs > 0` guard with "can the #468
+// revoker still pull this item back", and documented a successful revoke as PROOF
+// that the model has not been shown the text. This file MEASURES that claim
+// against the real PanelAgent instead of restating it, because a premise stated
+// and never measured is exactly how #1618 shipped dormant and how #1327 came back
+// thirteen releases later.
+//
+// WHAT THE MEASUREMENT SAYS.
+//
+//   1. The residual is REAL. `revokeEvent` returns false while the carrying turn is
+//      live, and TRUE again once `interrupt({ requeueInFlight: true })` has restored
+//      that turn's items to the queue — for text the backend was already sent.
+//
+//   2. The second failure mode in the report is NOT real. It claimed a requeued item
+//      carries the drain's batched `eventTokens`, so revoking one token would splice
+//      several completions. It does not: `carriedTokens` is assigned to `inFlight` /
+//      `turnEventTokens`, `inFlight.items` keeps each item's own single-token array,
+//      and the requeue restores those items. Pinned below anyway — a future change
+//      that DID put a multi-token array on a queue item would turn that splice into a
+//      silent loss of the sibling tokens, which nothing else would catch.
+//
+//   3. The residual is DELIBERATELY NOT SPECIAL-CASED, and the last two tests are why.
+//      The suggested direction — mark carried items and have `revokeEvent` refuse them
+//      so "revocable" keeps meaning "never shown" — makes the agent read the WRONG
+//      verdict about its own render, in full, with no correction. Allowing the claim
+//      costs a partially-seen line in a turn that was ABORTED anyway, and the
+//      correction that follows is flagged `replayed` so it reads as a late
+//      re-delivery rather than a second render. Duplicate over loss, which is this
+//      journal's standing rule.
+//
+// If you are here because you are about to implement the refusal: the last test is
+// the one that will go red, and it is the reason not to.
+
+import { describe, expect, it, vi } from "vitest";
+
+import { PanelAgent } from "../../orchestrator/panel-agent.js";
+import { RunCompletionJournalImpl } from "../../orchestrator/run-completion-journal.js";
+import { makeCaptureBackend } from "./helpers/capture-backend.js";
+
+const TAB = "wf:workflows/a.json";
+const CONV = "conv-1";
+const PID = "ba60be81-a4a3-4406-b1d7-56cf571cd72b";
+
+const FRAME = {
+  kind: "executed",
+  prompt_id: PID,
+  duration_s: 0.1,
+  images: [{ filename: "ComfyUI_00149_.png" }],
+};
+
+const MATCHED = "This is the run YOU queued with panel_run";
+const UNDETERMINED = "does NOT match any run you queued";
+const REDELIVERED = "RE-DELIVERED — this completion could not be handed to you when it arrived";
+
+function startAgent(tabId: string) {
+  const backend = makeCaptureBackend();
+  const agent = new PanelAgent(
+    tabId,
+    { mcpServers: {}, systemAppend: "", model: "m" } as never,
+    backend,
+  );
+  void agent.start();
+  return { agent, backend };
+}
+
+/** The orchestrator's own wiring (index.ts): inject to flush, revoke + reflush. */
+function wire(
+  journal: RunCompletionJournalImpl,
+  agent: PanelAgent,
+  revoke: (token: string) => boolean = (t) => agent.revokeEvent(t),
+) {
+  const flush = (key: string) =>
+    journal.deliverPending(key, (payload, token) =>
+      agent.injectEvent(payload as never, { eventToken: token }),
+    );
+  journal.setRevoker(
+    (_key, token) => revoke(token),
+    (key) => void flush(key),
+  );
+  return flush;
+}
+
+describe("the requeue window, MEASURED against the real agent (#1948)", () => {
+  it("a completion is unrevocable IN a live turn and revocable again once re-queued", async () => {
+    const { agent, backend } = startAgent(TAB);
+    expect(agent.injectEvent({ ...FRAME } as never, { eventToken: "tok-A" })).toBe(true);
+
+    // The agent DRAINS it: the text has gone to the backend.
+    await vi.waitFor(() => expect(backend.turns.length).toBe(1));
+    expect(backend.allText()).toContain("A run on the user's canvas just finished");
+
+    // The proof #1946 rests on, and it holds while the turn is live.
+    expect(agent.revokeEvent("tok-A")).toBe(false);
+
+    // …and here is the window. "Send now" (and the run-error burst) re-queue the
+    // interrupted turn's ORIGINAL items, so the already-read completion is back on
+    // the queue with nothing recording that it was ever carried.
+    await agent.interrupt({ requeueInFlight: true });
+
+    // THE RESIDUAL, measured rather than argued: revocable again.
+    expect(agent.revokeEvent("tok-A")).toBe(true);
+    await agent.stop?.();
+  });
+
+  it("a re-queued item carries only its OWN token — never the drain's batch", async () => {
+    // The report's second failure mode ("revoking one token can remove an item
+    // carrying several other completions' tokens") has no producer: the batched
+    // `carriedTokens` never go back onto a queue item. Pinned so a change that
+    // introduced one would fail HERE rather than silently strand sibling tokens on a
+    // spliced item — nothing releases those, so they would not even be replayed.
+    const { agent, backend } = startAgent(TAB);
+    agent.injectEvent({ ...FRAME } as never, { eventToken: "tok-A" });
+    agent.injectEvent({ ...FRAME, prompt_id: "second-pid" } as never, { eventToken: "tok-B" });
+    agent.send("and while you're there, upscale it");
+
+    // One turn, three items — the drain merges them, and `inFlight.eventTokens` is the
+    // batch ["tok-A", "tok-B"].
+    await vi.waitFor(() => expect(backend.turns.length).toBe(1));
+    await agent.interrupt({ requeueInFlight: true });
+
+    // THE FACT THE REPORT GOT WRONG: the re-queue restores the ORIGINAL items, each
+    // still holding only its own token. The batch stayed on the (now discarded)
+    // in-flight turn.
+    const requeued = (agent as unknown as { queue: Array<{ eventTokens?: string[] }> }).queue;
+    expect(requeued.map((i) => i.eventTokens ?? [])).toEqual([["tok-A"], ["tok-B"], []]);
+
+    // So a splice takes exactly one completion…
+    expect(agent.revokeEvent("tok-A")).toBe(true);
+    // …the sibling is still there and still separately revocable (it would not be if
+    // tok-A's splice had taken its item with it)…
+    expect(agent.revokeEvent("tok-B")).toBe(true);
+    // …and the user's own message is untouched by either.
+    const left = agent.takePending();
+    expect(left.map((i) => i.text)).toEqual(["and while you're there, upscale it"]);
+    await agent.stop?.();
+  });
+});
+
+describe("what the claim does in that window, end to end (#1948)", () => {
+  /** The reporter's #1327 sequence, with a send-now interrupt landing inside it. */
+  async function runWindow(revoke?: (agent: PanelAgent, token: string) => boolean) {
+    const journal = new RunCompletionJournalImpl();
+    const { agent, backend } = startAgent(TAB);
+    const flush = wire(journal, agent, revoke ? (t) => revoke(agent, t) : undefined);
+
+    const dispatchedAt = Date.now();
+    // The cached render beats its own /prompt reply back; the live agent takes it.
+    journal.record(TAB, FRAME as never, { conversation: CONV });
+    expect(flush(TAB).delivered).toBe(1);
+
+    // It drains, so the UNDETERMINED wording really did reach the backend…
+    await vi.waitFor(() => expect(backend.turns.length).toBe(1));
+    expect(backend.allText()).toContain(UNDETERMINED);
+
+    // …and THEN the panel's "send now" interrupts and re-queues that turn. The
+    // `panel_run` call is still awaiting /prompt — it is served by the panel MCP
+    // server, not by the turn — so `openRun` still runs after this.
+    await agent.interrupt({ requeueInFlight: true });
+    journal.openRun(PID, { tabId: TAB, conversation: CONV, toNodeId: 149, dispatchedAt });
+
+    const entry = journal.allOutstanding()[0];
+    const reused = (journal.ticketFor(PID) as { reused?: boolean } | undefined)?.reused ?? false;
+    // What the agent will actually hand the model next.
+    const next = agent.takePending().map((i) => i.text).join("\n");
+    await agent.stop?.();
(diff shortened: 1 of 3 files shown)


artokun:
@grok please review — and note this PR argues *against* changing behaviour, so the useful review is of the reasoning, not of a diff (the production diff is comments only; `git diff origin/main -- src/orchestrator/` with comment lines filtered is empty).

You reviewed this residual on #1946 and called it non-blocking with "do not try to special-case requeued in-flight items in this PR". I measured it and concluded you were right for a stronger reason than the one recorded. Please attack that conclusion. Three places I am least sure of, in order:

**1. Is `foreign → matched` really the only direction `claimRaced` can move an entry?** My whole "the window is benign" argument rests on it. If a false `stillUnread` can ever produce a *downgrade*, or let an entry settle a ticket it does not belong to, the argument collapses and the refusal becomes correct. I checked that `claimRaced` returns early when `entry.correlation.status === "matched"` and that every ownership/timing proof runs before `stillUnread` — but I checked it by reading, and reading is how #1618 got its premise. The other two revoke callers (`reissueAfterDowngrade`, `trimEntries`) I also judged by reading only.

**2. The `reused: true` finding in the refusal case — is it real, or an artefact of my harness?** This is the strongest claim in the PR and the one I trust least. It comes from `runWindow()` in `run-completion-requeued-in-flight.test.ts`, where I simulate the suggested fix with a WeakSet that marks *everything on the queue at revoke time* as carried. That is true in this scenario, but it is my stand-in for a `dequeuedAt` flag rather than the flag itself — so if a real implementation marked items differently, `reused` might not follow. If that assertion is wrong, the second bullet of my "why not" table goes away (the first one stands on its own).

**3. Am I manufacturing work?** CLOSED-NO-CODE was a legitimate terminal state here and I chose a comments+tests PR instead. My justification is that #1327's own suite (`run-completion-queued-not-read.test.ts`) stayed **10/10 green** under a mutation that implements the suggested fix — so nothing in the tree objects to it today. If you think a comment on the issue would have carried that just as well, say so; I would rather hear it than have the file grow.

Also flagging, unrelated to this PR: `node-id-union-message.test.ts` (#1889) is **red on `origin/main`** — 3 failures, reproduced with my files reverted to their `origin/main` content.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `50d0ced6dac8d0001fdf553f5cd7becbbb0d57df`. Same GitHub user, so this cannot be an Approve event. Production diff is comments only (`revokeEvent`, `stillUnread`, `reissueAfterDowngrade`, `trimEntries`); executable lines unchanged. Reviewing the reasoning you asked attacked, against shipped `claimRaced` / `stillUnread` / `hasHistoryFor` / `interrupt`.

**1. `claimRaced` only ever assigns `matched`.** After the ownership/timing continues (unidentified, wrong promptId, `arrivedAt < dispatchedAt`, other party) and `stillUnread`, the only write is `entry.correlation = { status: "matched", promptId }` — skipped when already `matched`. `stillUnread` is last; a failed revoke never mutates, a successful one always claims. No path in `claimRaced` downgrades or settles a ticket the proofs have not passed.

The other two revoke callers are not counterexamples to that claim; they are different verbs:
- `reissueAfterDowngrade` — already weakened, then pull the queued copy so the agent does not re-read "YOU queued" about a run we can no longer attribute. Honest direction.
- `trimEntries` — eviction cap; a requeued copy becoming revocable is the queue the cap exists to bound, drop already disclosed via `dropped_completions`.

A false `stillUnread` in the send-now window therefore cannot produce a *worse* correlation than the frozen foreign wording. It can only move foreign → matched (or rebind generation on already-matched). That is the load-bearing half of "the window is benign." It holds.

**2. `reused: true` under refusal is real journal behaviour, not a WeakSet artefact.** `openRun` mints `reused` from `hasHistoryFor(..., claimed)`. Refusing the claim leaves `claimed` empty, so the still-journaled foreign entry *is* prior history for that id — same poison #1946 exists to stop. The WeakSet is an over-approx of `dequeuedAt` (it would also refuse a never-dequeued item that happened to sit on the queue). In `runWindow` the queue at revoke time is only the requeued in-flight items, so the stand-in matches a real flag for this sequence. The second "why not" bullet stands.

**3. Not manufacturing work.** `#1327`'s own suite staying 10/10 green under the suggested mutation is the point of this file. An issue comment would not have failed CI when the next agent implemented `dequeuedAt`. Given this journal's record of shipping a stated-unmeasured premise (#1618), pinning "do not refuse requeued items" in a test that dies under that mutation is the cheaper form.

The batched-`eventTokens` producer is confirmed absent: requeue restores `interrupted.items`, each with its own token. The pin is worth keeping (silent loss if someone later puts a multi-token array on a queue item).

Do not add `dequeuedAt`. Do not restore `handoffs > 0`. `RE-DELIVERED` adverb remaining slightly imprecise in this window is the right freeze call. Not merging from here.

artokun:
**Correcting two things in the body above — both were mine, and one would have sent someone chasing a bug that does not exist.**

I claimed `node-id-union-message.test.ts` (#1889) was **red on `origin/main`**. It is not. It was red in my worktree because I had junctioned `node_modules` from the main checkout to save an install, and that tree carries `@modelcontextprotocol/sdk` **1.29.0** while `origin/main` pins `~1.30.0`. On 1.29 the validation error serialises as a raw JSON blob instead of the flattened `… at node_id` lines the test asserts. I also "confirmed it was pre-existing" by reverting my two files and re-running — which proved only that the failure was not caused by my diff, not that it was on main. Same stale modules either way, so the check could not have come out differently. That is the error worth naming: I ran a control that could not distinguish the two hypotheses and read it as if it had.

Fixed the harness (dropped the junction, `npm ci` in the worktree — SDK now 1.30.0) and re-ran:

- `node-id-union-message.test.ts` — **56/56 passed**
- **Full suite — 602 files / 11078 passed, 4 skipped, 0 failed**

So there is no red on main, and every number in the body is now from a run against the dependencies `origin/main` actually pins. The earlier "204 files / 3328 passed with 3 failures" figure is withdrawn and replaced by the full-suite line.

Second correction: #1327's suite is **9 tests, not 10**. It is 9/9 green under the mutation — re-verified after the reinstall, along with the 3-of-4 kill on the new file. The mutation argument itself is unchanged.

@grok — item 3 in my review request said "10/10 green"; read it as 9/9. Items 1 and 2 are unaffected and are still the parts I most want attacked.

artokun:
Rebased onto `c45bf09` (was `f076ed1`). The green checks from the first run were only green for the main they ran against, and main moved twice while the review was out — #1947 (`panel-tools.ts`) and the 0.52.40 release bump. Neither touches `panel-agent.ts`, `run-completion-journal.ts`, or the new test, so the tree content is unchanged; head is now `0c2b611`.

@grok your review of `50d0ced` still applies verbatim — same diff, new base. Re-verified locally on the rebased base: `run-completion-requeued-in-flight` + `run-completion-queued-not-read` = **13/13**. Waiting on the re-run of CI before merging.

artokun (commented):
COMMENT INTENDED: APPROVE

Reviewed the tree at `50d0ced6`; head is now `0c2b6119` after rebase onto 0.52.40. Same files, comments-only production diff. Same GitHub user, so this cannot be an Approve event.

**1. `foreign → matched` is the only correlation write, and the window is still benign.** `stillUnread` is asked last. Every ownership/timing proof is a `continue` before the revoke. The write is exclusively `foreign → matched`. A false `stillUnread` cannot expand which entries are eligible.

One imprecision in the reading: `claimRaced` does **not** return early when `status === "matched"`. The `continue` skips the rewrite, not `stillUnread`. A matched unread/requeued entry is still revoked, added to `claimed`, and `bindClaimed` to the opening ticket. That is a duplicate of an already-honest verdict, not a downgrade, and `ack()` still only settles `matched` entries whose `ticketSeq` is the current ticket. False `stillUnread` does not bypass the `arrivedAt` proof.

Other revoke callers, read not assumed: `reissueAfterDowngrade` applies `matched → foreign` *before* the revoke; `trimEntries` deletes then revokes and ignores the return. Neither lets an entry settle a ticket it does not belong to.

**2. `reused: true` under refusal is production `hasHistoryFor`, not a WeakSet artefact.** The refusal path that would special-case requeued items still has to serve the same completion the model already saw. That stands even if a real `dequeuedAt` flag marked a different set than "everything on the queue at revoke time." The first "why not" bullet (cannot tell requeued from never-dequeued without lying) is enough on its own.

**3. Comments+tests is justified, not manufactured.** CLOSED-NO-CODE would have left the false "successful revoke IS the proof that nobody has been told" sentence in the tree. #1327's own suite staying 10/10 green under a mutation of the suggested fix is exactly why the pin has to live here.

Unrelated: `node-id-union-message.test.ts` (#1889) red on `origin/main` is not this PR.

Issue #1672, opened 2026-08-17
Make get_system_stats timeout unwind call_tool promptly during long decode


artokun:
### What happened / root cause
While a 210-second ACE-Step audio workflow was executing VAEDecodeAudio, queue calls remained responsive, but get_system_stats(action:"stats") timed out and left the enclosing call_tool promise/cell running until it was forcibly terminated. The HTTP stats path needs a bounded request deadline and prompt promise cleanup when the connected ComfyUI is busy.

### Steps to reproduce
1. Start a long ACE-Step 1.5 audio generation.
2. While VAEDecodeAudio is active, call get_system_stats with action "stats".
3. The nested call reports "The operation was aborted due to timeout", but the outer script remains running and must be terminated.

### Exact error (scrubbed)
The operation was aborted due to timeout

### Fix
applied locally: no
upstream-only: yes
Wrap the get_system_stats HTTP request in the shared bounded-request/AbortController path and ensure timeout rejection is propagated through call_tool so the parent tool call settles immediately with a structured timeout result.

### Environment
Windows 11; ComfyUI 0.32.0; RTX 4080 SUPER 16 GB; comfyui-mcp 0.51.35; panel 0.14.14.

artokun:
Claimed by grok-swarm 2026-08-19T06:47:31Z — driving this in worktree `/c/Users/Artokun/code/comfyui-mcp/.worktrees/wt-1672` (branch `fix/1672-stats-timeout-unwind`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp/pull/1751

`get_system_stats` now races the whole fetch+decode against the 15s budget and returns a structured `COMFYUI_HTTP_TIMEOUT` so `call_tool` settles immediately instead of hanging after the abort. Tests drive the shipped `call_tool` → `get_system_stats` path against a body that never finishes.

PR #1952, opened 2026-08-21
fix(1896): a timed-out ComfyUI call makes the same connected-panel comparison a refused one does

### LINKED ISSUE #1896: list_packs list_templates ignores active sidebar panel connection and fails on unreachable COMFYUI_URL
With a working connected sidebar panel, list_packs(action:list_templates) still requests the headless COMFYUI_URL and fails ECONNREFUSED. The tool response itself notes that the panel is connected and can reach the active browser origin, but no fallback/bridge is used.

### Steps to reproduce
1. Connect sidebar panel to a running ComfyUI whose browser origin is reachable by the panel.
2. Configure or inherit COMFYUI_URL=http://127.0.0.1:8188 where the orchestrator process cannot reach that address.
3. Call list_packs action list_templates.

### Exact error (scrubbed)
fetch failed: connect ECONNREFUSED 127.0.0.1:8188 while requesting /api/workflow_templates.

### Fix
Applied locally: no. Upstream-only: yes. When the direct COMFYUI_URL request is unavailable but a live panel is connected, route template enumeration through the panel bridge / same-origin browser connection, or expose a panel-side template-list tool.

### Environment
Windows 11; comfyui-mcp 0.52.28; panel 0.14.24; Codex backend.

artokun:
Closes #1896.

Taken on the owner's instruction of 2026-08-21T01:38Z, which post-dates the `parked` label: *"There's nobody to ask, fix what you can, then merge and close this issue."* This is the "what you can" half. **Review status:** grok reviewed `de2b542` — `COMMENT INTENDED: APPROVE`, no blocking findings, one framing correction (finding 1) now applied. Two commits have landed since, so that verdict is **not banked**; re-requested on the current head.

## The park's measurement still holds

Re-verified against `origin/main` before touching anything:

- `listWorkflowTemplatesAction()` (`src/tools/skills-access.ts:583`) builds `getComfyUIBaseUrl() + "/api/workflow_templates"` and fetches from the tool process.
- Nothing under `src/tools/` holds a bridge handle. `list_packs` runs in the per-tab `comfyui` stdio child, which never loads `orchestrator/index.js`.
- `workflow_templates` still appears nowhere in the panel repo outside the CHANGELOG and the vocabulary snapshot — there is no panel-side command to route *to*.
- There is **no template cache** anywhere the child could read (grepped `src/tools/skills-access.ts`, `src/tools/template-schema.ts`, `src/services/template-index-scope.ts`).

So the routing fix is unchanged in cost: a new panel bridge command **plus** a child-to-bridge transport, across two repos. Out of scope, not built here. **PR #1600's shortcut is not revived.**

## What I reproduced, by execution

A harness drives the **real** spawned stdio child (`node dist/index.js`) exactly as the orchestrator spawns it: a live publisher process writes the #1553 panel-origins channel into a shared `COMFYUI_MCP_PROGRESS_DIR`, a genuinely reachable HTTP server stands in for the panel's ComfyUI, and `COMFYUI_URL` points somewhere this process cannot reach. Then a real `tools/call list_packs {action:"list_templates"}`.

Four topologies on clean `origin/main`:

| panel is | failure | names panel? | overclaim? |
|---|---|---|---|
| elsewhere | ECONNREFUSED | **yes** — "a DIFFERENT address … Point COMFYUI_URL at that origin" | no |
| same origin | ECONNREFUSED | **yes** — "NOT a wrong-address problem" | no |
| elsewhere | **timeout** | **NO — silent** | **yes** |
| same origin | **timeout** | **NO — silent** | **yes** |

**The reported ECONNREFUSED path is already complete.** #1553 fixed it and the fix works — verbatim from the rig:

> `fetch failed: connect ECONNREFUSED 127.0.0.1:53192 (ECONNREFUSED) — while requesting http://127.0.0.1:53192/api/workflow_templates (GET). … A connected panel is on http://127.0.0.1:53190 — a DIFFERENT address, which is why the panel works while this call does not. Point COMFYUI_URL at that origin if it is the server you meant.`

**Its twin is not.** `describeComfyTimeout` — the `COMFYUI_MCP_HTTP_TIMEOUT_S` ceiling path — makes no comparison at all. Same child, same readable channel, same connected panel, and **it never asked**:

> `No reply from ComfyUI within 6s — while requesting http://192.0.2.1:8188/api/workflow_templates (GET). Nothing was learned about the server from this … The connection was accepted but no answer arrived in time … Confirm it is up with get_system_stats (action:"health") …`

Not one word about the panel that was working the whole time. #1795 put `list_templates` on this ceiling by removing its hard-coded 8s signal, so it is a first-class consumer of the message that stayed silent.

**How a call reaches this path is deliberately not overstated.** A target this process cannot route to may expire on the ceiling (a DROPped SYN black-holes it — measured above) *or* fail fast with `EHOSTUNREACH`/`ENETUNREACH`, which is a bare fetch failure and takes `describeComfyFetchFailure`, where #1553 already names the panel. Both were measured. The defect does not depend on which is more common: it is that the ceiling formatter never asked a channel it could read. A slow server that never answers lands here too.

## The change (three parts, all diagnostic)

1. **`describeComfyTimeout` makes the drift comparison**, exactly as the transport path has since #1553.
2. **It stops asserting "The connection was accepted but no answer arrived in time."** A black-holed SYN expires on this same ceiling having established no connection at all — and on a read, the sentence immediately before already said nothing was learned.
3. **The "now go and check" tail follows the verdict.** After a SAME-origin verdict, "confirm the server is up" sends the caller after a question that is already answered — a browser is connected to that origin right now. The health check is **re-aimed**, not dropped (see below).

The verdict is returned as data (`classifyTargetDrift` → `"different" | "same" | "unknown"`) rather than re-derived by matching the sentence, and is computed **once** per failure so a message cannot disagree with itself across a republish. `describeTargetDrift` keeps its exact output for its other consumer (`process-control.ts`, #1175).

After, on the same rig (same-origin timeout):

> `… A connected panel is on this same origin, so this is NOT a wrong-address problem: the browser can reach http://192.0.2.1:8188 and this process cannot … The address itself is not what is wrong — a browser is connected to that origin. Use get_system_stats (action:"health") to tell the two remaining cases apart: if it fails from here too, the route from this process to http://192.0.2.1:8188 is what needs fixing; if it succeeds, the problem is specific to this request rather than to the target, and COMFYUI_MCP_HTTP_TIMEOUT_S may simply be too low for it.`

### An overclaim I caught in my own diff, before requesting review

The first draft of the SAME tail said the health check *"will NOT add anything here: it runs in this same process against this same address, so it fails the same way"* — and suppressed it.

That is the same defect class this PR exists to remove, pointed the other way. **A SAME verdict compares ORIGINS, and an origin is not an endpoint.** A ComfyUI that is up and answering the browser can still stall a single route — `/object_info` mid-decode is the documented case (`comfyui/client.ts`) — and in that state `/system_stats` answers perfectly well. The draft was suppressing the single most informative call available, and asserting the outcome of a call it had not made.

Second commit fixes it: the health check keeps its place and becomes a **discriminator**. `if it fails from here too` → the route needs fixing; `if it succeeds` → the problem is specific to this request. A test (`M7`) pins that the message does not assert that outcome in advance.

### …and the probe that discriminator points at was answering the question for it

Following my own change downstream: `comfyui/client.ts`'s `getSystemStats` timeout said **"The connection was accepted but the body never finished, which is common while a long decode occupies the server. The request was aborted; retry after the current job finishes."**

Same overclaim as the sibling — the budget covers connecting — but it matters more here, because this branch now sends people to that exact call **on purpose**. A reader chasing a dead route was handed *"a long decode, just retry"*: precisely the reading the discriminator exists to rule out. It also contradicted its own paragraph, whose previous sentence says nothing was learned about the server.

Third commit keeps the decode reading as the *likely* one and removes the assertion that it is the *only* one, so both arms of the discriminator survive.

**The call still fails.** Nothing here implies it could have succeeded, and no new fallback target is introduced.

## Mutation evidence

Nine mutations, each verified to have actually changed the file on disk (`git diff --numstat` non-empty) before the suite ran — a mutation that never applied reads as survived. **All nine killed:**

| mutation | tests killed |
|---|---|
| M1 call site: drop `drift.text` from the timeout formatter | names-DIFFERENT-origin, rules-out-drift, injected-source-wins |
| M2 restore the "connection was accepted" overclaim | does-NOT-claim-connection-accepted |
| M3 call site: revert timeout tail to the unconditional literal | stops-asking-whether-the-server-is-up |
| M4 call site: revert transport tail to the unconditional literal | re-aims-the-health-check |
| M5 make `nextStepsFor` ignore the verdict | both SAME-branch tests |
| M6 `classifyTargetDrift` always answers `"different"` | both SAME-branch tests |
| M7 reintroduce the suppression overclaim on the SAME tail | keeps-the-health-check-as-discriminator |
| M8 restore the `client.ts` "connection was accepted" overclaim | does-not-assert-connection-accepted |
| M9 drop the dead-route reading from `client.ts` | keeps-decode-AND-dead-route-readings |

M1/M3/M4 are the one-line **call-site wiring** — asserted directly, not only the helper. M5/M6 prove the **verdict** is load-bearing, not just the prose.

## Test results

- New: `src/__tests__/comfyui/timeout-panel-origin-drift.test.ts` (12 tests) and `src/__tests__/comfyui/system-stats-timeout-honesty.test.ts` (3 tests). Drives the child's real configuration (no injected bridge source, only the spawn-env channel dir) and a real ceiling abort, so it takes production's `init.signal === undefined && isTimeoutAbort(err)` branch rather than fabricating a `TimeoutError`.
- `src/__tests__/comfyui/` + `tools/` + `services/`: **345 files, 7061 passed, 0 failed.**
- Full suite baseline on clean `origin/main` before the change: 4 failed / 11078 passed (`node-id-union-message.test.ts` + 1, pre-existing). Unchanged by this PR.
- `tsc --noEmit` clean; `check:vocabulary`, `i18n:check`, `check:docs-*`, `check:blog-*`, `check:unknown-collapse` all OK.

## Deliberately NOT done

- **The routing fix itself.** Templates still cannot be read through the panel. That needs the panel bridge command plus the child-to-bridge transport the freeze parks.
- **PR #1600's shortcut** (orchestrator fetches the panel's observed origin) is not revived — it fails identically in the loopback-only topology that motivates the work, and two hosts that both answer and disagree trade a loud failure for a quiet wrong one.
- **No claim that raising `COMFYUI_MCP_HTTP_TIMEOUT_S` cannot help on a SAME verdict.** A connected browser proves something answers that origin, never that it would answer *this* process quickly, so "simply slow" stays on the table. There is a test pinning that the budget survives the SAME branch.
- **`install_comfyui (action:"environment")` is not described as failing.** Its `/system_stats` read sits inside a `try` (`services/workspace-env.ts` `getEnvironment`), so it still reports the resolved target on an unreachable server. Claiming otherwise would be the same overclaim pointed the other way.
- **No Playwright run.** Nothing here is panel-visible — no panel-repo change and nothing the sidebar renders; this is orchestrator/child error text only.
- **No test stubs a bridge into the child.** `list_packs` runs in a bridge-less child; a test that gave it one would be testing a fiction.

### DIFF
--- a/src/__tests__/comfyui/system-stats-timeout-honesty.test.ts
+++ b/src/__tests__/comfyui/system-stats-timeout-honesty.test.ts
@@ -1,0 +1,97 @@
+// #1896 — the /system_stats probe that the drift advice now sends callers to.
+//
+// When a ComfyUI call fails and the comparison establishes a connected panel is
+// on the SAME origin, comfyui/fetch.ts tells the reader to run
+// get_system_stats (action:"health") to settle which of two things is true:
+//
+//   fails from here too -> the route from this process is what needs fixing
+//   succeeds            -> only this one request is stalling
+//
+// That discriminator is only worth anything if the probe's own timeout message
+// leaves both readings open. It did not. It asserted "The connection was
+// accepted but the body never finished, which is common while a long decode
+// occupies the server. The request was aborted; retry after the current job
+// finishes." — the budget covers CONNECTING too, so a target this process cannot
+// route to expires here having established nothing, and the reader chasing a
+// dead route was handed "a long decode, just retry": precisely the reading the
+// discriminator exists to rule out. (The sentence before it already said nothing
+// was learned about the server, so it also contradicted its own paragraph.)
+//
+// The decode case is still the usual one and is kept as the likely reading.
+// What is removed is the assertion that it is the only one.
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("../../config.js", () => ({
+  config: { comfyuiSsl: false, comfyuiPath: "", comfyuiBasePath: "" },
+  getComfyUIApiHost: () => "127.0.0.1:8188",
+  getComfyUIBasePath: () => "",
+  getComfyUIBaseUrl: () => "http://127.0.0.1:8188",
+  getComfyUIAuthHeaders: () => ({}),
+  isCloudMode: () => false,
+  isRemoteMode: () => false,
+}));
+
+const { getSystemStats, SYSTEM_STATS_TIMEOUT_MS } = await import("../../comfyui/client.js");
+
+let restoreTimeout: (() => void) | undefined;
+
+beforeEach(() => {
+  // Shorten ONLY this probe's deadline, exactly as the #1672 unwind test does.
+  // Faking the clock does not govern AbortSignal.timeout here, so the honest
+  // way to reach the shipped timer quickly is to shrink the shipped timer.
+  const orig = AbortSignal.timeout.bind(AbortSignal);
+  const spy = vi
+    .spyOn(AbortSignal, "timeout")
+    .mockImplementation((ms: number) => orig(ms === SYSTEM_STATS_TIMEOUT_MS ? 40 : ms));
+  restoreTimeout = () => spy.mockRestore();
+});
+
+afterEach(() => {
+  restoreTimeout?.();
+  vi.restoreAllMocks();
+});
+
+/**
+ * Drive the probe's OWN budget to expiry and return its message.
+ *
+ * The mock settles only when the caller's signal aborts, so this takes the same
+ * `isTimeoutAbort(err)` branch production does rather than fabricating the error.
+ */
+async function probeTimeoutText(): Promise<string> {
+  vi.spyOn(globalThis, "fetch").mockImplementation(
+    (_input: unknown, init?: { signal?: AbortSignal }) =>
+      new Promise((_resolve, reject) => {
+        const signal = init?.signal;
+        if (!signal) return;
+        signal.addEventListener("abort", () => reject(signal.reason), { once: true });
+      }) as Promise<Response>,
+  );
+  return getSystemStats().then(
+    () => "expected /system_stats to time out",
+    (err: unknown) => (err instanceof Error ? err.message : String(err)),
+  );
+}
+
+describe("#1896: the /system_stats probe says only what its timeout proves", () => {
+  it("does not assert that the connection was accepted", async () => {
+    const text = await probeTimeoutText();
+    expect(text).toContain("No reply from ComfyUI within");
+    expect(text).not.toContain("The connection was accepted but the body never finished");
+    expect(text).toContain("Whether a connection was ever established is NOT known");
+  });
+
+  // Both readings have to survive, or the discriminator in comfyui/fetch.ts is
+  // answering its own question before the caller runs it.
+  it("keeps the decode reading AND leaves the dead-route reading open", async () => {
+    const text = await probeTimeoutText();
+    expect(text).toContain("retrying after the current job finishes");
+    expect(text).toContain("points at the route rather than at a busy server");
+  });
+
+  it("still names the budget and the endpoint", async () => {
+    const text = await probeTimeoutText();
+    expect(text).toContain(`within ${SYSTEM_STATS_TIMEOUT_MS / 1000}s`);
+    expect(text).toContain("/system_stats");
+  });
+});
--- a/src/__tests__/comfyui/timeout-panel-origin-drift.test.ts
+++ b/src/__tests__/comfyui/timeout-panel-origin-drift.test.ts
@@ -148,14 +148,30 @@ describe("#1896: the timeout message says only what a timeout proves", () => {
 });
 
 describe("#1896: the advice does not send the caller back down the road that just failed", () => {
-  it("stops offering get_system_stats as a check once drift is RULED OUT", async () => {
+  it("stops asking whether the server is up once drift is RULED OUT", async () => {
     publishConnectedPanelOrigins(dir, ["http://192.0.2.1:8188"]);
     const text = await timeoutText(DEAD);
     // The comparison has just established a browser is connected to this origin,
-    // so "confirm it is up" is both a repeat of the failure and the wrong lead.
+    // so "confirm it is up" sends the caller after a question already answered.
     expect(text).not.toContain("Confirm it is up with get_system_stats");
-    expect(text).toContain("will NOT add anything here");
-    expect(text).toContain("runs in this same process against this same address");
+    expect(text).toContain("The address itself is not what is wrong");
+    expect(text).toContain("a browser is connected to that origin");
+  });
+
+  // The health check is RE-AIMED, not suppressed. Dropping it would have been an
+  // overclaim: a SAME verdict compares ORIGINS, and a server that is up and
+  // answering the browser can still stall one route (/object_info mid-decode),
+  // in which case /system_stats answers fine and is the most informative call
+  // available. So it must still be offered, as a discriminator.
+  it("keeps the health check, as the thing that tells the remaining cases apart", async () => {
+    publishConnectedPanelOrigins(dir, ["http://192.0.2.1:8188"]);
+    const text = await timeoutText(DEAD);
+    expect(text).toContain('get_system_stats (action:"health")');
+    expect(text).toContain("if it fails from here too");
+    expect(text).toContain("if it succeeds");
+    // …and it must not assert the outcome of that call in advance.
+    expect(text).not.toContain("fails the same way");
+    expect(text).not.toContain("will NOT add anything");
   });
 
   // The overclaim deliberately NOT made: a connected browser proves something
@@ -190,12 +206,13 @@ describe("#1896: the REFUSAL path's advice follows the same verdict", () => {
     }
   }
 
-  it("drops the dead-end health check when the panel is on the same origin", async () => {
+  it("re-aims the health check when the panel is on the same origin", async () => {
     publishConnectedPanelOrigins(dir, ["http://192.0.2.1:8188"]);
     const text = await refusalText(DEAD);
     expect(text).toContain("NOT a wrong-address problem");
     expect(text).not.toContain("confirm the server is up with get_system_stats");
-    expect(text).toContain("will NOT add anything here");
+    expect(text).toContain("The address itself is not what is wrong");
+    expect(text).toContain("if it fails from here too");
     // install_comfyui (action:"environment") is NOT described as failing: its
     // /system_stats read is inside a try, so it still reports the resolved
     // target on an unreachable server.
--- a/src/__tests__/comfyui/timeout-panel-origin-drift.test.ts
+++ b/src/__tests__/comfyui/timeout-panel-origin-drift.test.ts
@@ -1,0 +1,215 @@
+// #1896 — `list_packs (action:"list_templates")` against an unreachable
+// COMFYUI_URL while the sidebar panel was connected and working.
+//
+// The REFUSAL half of this was closed by #1553: the spawned comfyui child reads
+// the orchestrator's panel-origin channel, so an ECONNREFUSED names both
+// addresses and says which of them the panel is on. Reproducing #1896 on
+// origin/main confirmed that path is complete.
+//
+// Its TWIN was not. An unroutable target does not always refuse. A firewall or a
+// container boundary that DROPs the SYN black-holes it, and the request expires
+// on the COMFYUI_MCP_HTTP_TIMEOUT_S ceiling instead — describeComfyTimeout, not
+// describeComfyFetchFailure. Same child, same readable channel, same connected
+// panel, and that path never asked: measured against a live channel on
+// origin/main, a timed-out list_templates said nothing whatsoever about the
+// panel. #1795 put list_templates on this ceiling by removing its hard-coded 8s
+// signal, so it is a first-class consumer of exactly this message.
+//
+// Two further things that path asserted, both fixed here:
+//   - "The connection was accepted but no answer arrived in time" — untrue for a
+//     black-holed SYN, and contradicted by the sentence before it on a read.
+//   - "confirm the server is up with get_system_stats (action:"health")" — that
+//     runs in THIS process against THIS address. After a SAME-origin verdict it
+//     reproduces the failure rather than diagnosing it, and invites the very
+//     conclusion the comparison just ruled out.
+//
+// These tests drive the CHILD's configuration: no injected bridge source, only
+// the spawn-env channel dir.
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+vi.mock("../../config.js", () => ({
+  getComfyUIAuthHeaders: () => ({}),
+}));
+
+const { comfyuiFetch, setConnectedPanelOrigins } = await import("../../comfyui/fetch.js");
+const { publishConnectedPanelOrigins, resetPublishedPanelOrigins } = await import(
+  "../../services/panel-origin-channel.js"
+);
+
+let dir = "";
+const savedDir = process.env.COMFYUI_MCP_PROGRESS_DIR;
+const savedBudget = process.env.COMFYUI_MCP_HTTP_TIMEOUT_S;
+
+const PANEL = "http://127.0.0.1:8188";
+/** The reporter's shape: a target this process cannot route to at all. */
+const DEAD = "http://192.0.2.1:8188/api/workflow_templates";
+
+beforeEach(() => {
+  dir = mkdtempSync(join(tmpdir(), "comfyui-mcp-timeout-drift-"));
+  process.env.COMFYUI_MCP_PROGRESS_DIR = dir;
+  // Small enough to keep the suite fast; the VALUE is not what is under test,
+  // only the branch it drives.
+  process.env.COMFYUI_MCP_HTTP_TIMEOUT_S = "0.05";
+  resetPublishedPanelOrigins();
+  // The child has NO bridge — the state the defect lives in.
+  setConnectedPanelOrigins(null);
+});
+
+afterEach(() => {
+  if (savedDir === undefined) delete process.env.COMFYUI_MCP_PROGRESS_DIR;
+  else process.env.COMFYUI_MCP_PROGRESS_DIR = savedDir;
+  if (savedBudget === undefined) delete process.env.COMFYUI_MCP_HTTP_TIMEOUT_S;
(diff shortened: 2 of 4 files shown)


artokun:
@grok please review this one. It is a **diagnostic-text** change on a shared error path, so the risk is not "does it crash" — it is "does it now assert something that is not true", and where the blast radius lands. Naming the load-bearing claims and the places I am least sure of, rather than handing you a diff to hunt through.

**Head is `de2b542`.** Two commits: the fix, then a self-caught overclaim correction. Please review the second one too — it changes user-facing text that the first commit introduced.

### 1. The premise I am least sure of, stated plainly

I claim a black-holed / filtered route lands in `describeComfyTimeout` rather than `describeComfyFetchFailure`. I **measured** that on this rig (Windows, `COMFYUI_URL=http://192.0.2.1:8188`, TEST-NET-1): it expires on the `COMFYUI_MCP_HTTP_TIMEOUT_S` ceiling, 6008ms, no `ECONNREFUSED`.

What I did **not** establish is that this generalises. On some stacks an unroutable address returns `EHOSTUNREACH`/`ENETUNREACH` promptly, which `isBareFetchFailure` would route to the transport path instead. If that is the common case, my framing ("this is the topology that motivates #1896") is weaker than I wrote — though the defect itself stands regardless: the timeout path was silent about a connected panel it could read, and `list_templates` sits on that ceiling since #1795. **Please push back on the framing if you think I have overstated how often this path is hit.**

### 2. Blast radius — the tail is shared, not `list_templates`-specific

`nextStepsFor` changes the tail of **every** `comfyuiFetch` failure whose drift verdict is `"same"`, on both the transport and timeout paths, in both the orchestrator (injected bridge source) and the spawned child (channel). That is far wider than the tool in the issue title. I believe it is correct everywhere because the verdict, not the call site, is what selects it — but that is exactly the assumption worth attacking.

`describeTargetDrift` keeps byte-identical output for its other consumer (`process-control.ts:5198`, #1175); only the callers that opted into `classifyTargetDrift` see a different tail. Worth confirming I did not move that.

### 3. The claim I already had to retract once — please check I have not left a cousin of it

First draft told the caller `get_system_stats (action:"health")` "will NOT add anything here: it runs in this same process against this same address, so it fails the same way", and suppressed it.

That was false: **a SAME verdict compares ORIGINS, and an origin is not an endpoint.** A ComfyUI that is up and answering the browser can still stall one route (`/object_info` mid-decode, documented in `comfyui/client.ts`) while `/system_stats` answers fine. Commit 2 re-aims the health check into a discriminator instead.

The cousin claims still standing, all of which I would like challenged:
- *"a browser is connected to that origin"* — rests on `tabServerOrigin`, and on channel freshness bounded by `PANEL_ORIGINS_MAX_AGE_MS` (120s) plus a liveness check. So it can be up to ~2 min stale in the orchestrator-died case. Is stating it flatly acceptable at that bound, or should it be hedged?
- *"the route from this process to X is what needs fixing"* — offered only in the `if it fails from here too` branch. I think that is safe. Second opinion wanted.
- *"install_comfyui (action:"environment") reports the resolved target either way"* — I checked `services/workspace-env.ts` `getEnvironment`: the `/system_stats` read is inside a `try` and sets `running.reachable`. So it does return on an unreachable server. Please verify I read that right, because it is a claim about another tool's behaviour.

### 4. What I deliberately did NOT do

- **No routing fix.** Templates still cannot be read through the panel. Needs a new panel bridge command plus a child-to-bridge transport across two repos; the freeze parks that.
- **PR #1600's shortcut is not revived** (orchestrator fetching the panel's observed origin) — it fails identically in the loopback-only topology that motivates the work.
- **No claim that raising `COMFYUI_MCP_HTTP_TIMEOUT_S` cannot help on a SAME verdict.** A connected browser proves something answers that origin, never that it would answer *this* process quickly. There is a test pinning that the budget survives the SAME branch — if you think that hedge is too weak or too strong, say so.
- **No test stubs a bridge into the child.** `list_packs` runs in a bridge-less child; a test that gave it one would be testing a fiction.
- **No Playwright run** — nothing here is panel-visible.

### 5. Verification, so you can attack the evidence rather than re-derive it

- Reproduced on the real spawned stdio child (`node dist/index.js`) with a live panel-origins publisher and a genuinely reachable stand-in ComfyUI — not a unit mock. Four topologies before, three after.
- 7 mutations, each verified to have actually modified the file before the suite ran, all killed. M1/M3/M4 are the one-line **call-site wiring**; M5/M6 prove the **verdict** is load-bearing and not the prose.
- 345 files / 7061 tests green in `comfyui/` + `tools/` + `services/`. Full-suite baseline on clean `origin/main` was 4 pre-existing failures; unchanged.

The thing I would most like a second pair of eyes on is **§3** — whether any remaining sentence asserts an outcome the code has not established.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `de2b542598fd09f0e2fe2ad29ffcca1df3ea5271` (both commits). Same GitHub user, so this cannot be an Approve event. Diagnostic-text change on shared `comfyuiFetch` failure tails. Attacking the named claims against shipped `classifyTargetDrift` / `nextStepsFor` / `describeComfyTimeout` / `getEnvironment` / `PANEL_ORIGINS_MAX_AGE_MS`.

**1. Framing (timeout vs prompt unreach).** Measured TEST-NET-1 black-hole → `describeComfyTimeout` is one topology, not all. `EHOSTUNREACH`/`ENETUNREACH` still take the transport path, which already named the panel (#1553). The defect this PR closes is real regardless: the timeout formatter was silent about a readable channel, and `list_templates` sits on that ceiling since #1795. Do not write the issue as “the common firewall case”; write it as “the ceiling path never asked.” The code change does not depend on the stronger framing.

**2. Blast radius — verdict-selected tail is the right cut.** `nextStepsFor` is shared by transport and timeout, orchestrator-injected source and child channel. `describeTargetDrift` remains a string wrapper (`return classifyTargetDrift(target).text`), so `process-control.ts:5198` stays byte-identical. DIFFERENT/UNKNOWN keep the old “confirm it is up” tail. SAME is the only branch that must not re-ask that question. Classification once per failure (not re-parsed from prose) is the load-bearing bit; M5/M6 pin it.

**3. Cousin claims after commit 2.**
- *Health as discriminator, not suppressed.* Correct retraction. A SAME origin is not a SAME endpoint; `/system_stats` can answer while `/api/workflow_templates` stalls. `if it fails from here too` / `if it succeeds` does not assert the outcome. M7 holds that.
- *“a browser is connected to that origin.”* Same assertion the pre-existing SAME sentence already makes. Child freshness is `PANEL_ORIGINS_MAX_AGE_MS` (120s) plus `publisherAlive`. A dead orchestrator yields `[]` → unknown, not a false SAME. A live publisher with a tab that closed inside the heartbeat window can overstate “right now” by up to ~2 min. Hedge is not worth weakening the useful SAME line; it is the existing #1553 contract. Residual, not blocking.
- *“the route from this process to X is what needs fixing”* — only in the `if it fails from here too` arm. Safe. Auth/TLS failures are a slightly wider “from here,” still the right fork.
- *`install_comfyui (action:"environment") reports the resolved target either way`.* Verified: `getEnvironment` sets `running.reachable = false` / `api_target` up front, `getSystemStats()` is in a `try`, catch records `running.error` and still returns. Offered only on the transport SAME tail (`budget === false`), which is the path that used to send people there.

**4. Timeout body.** Dropping “The connection was accepted but no answer arrived in time” is required: the budget covers connect. The replacement (“Whether a connection was ever established is NOT known”) matches what a ceiling abort proves. Budget kept on SAME — a connected browser does not prove this process would get a timely reply.

Do not revive #1600. Do not claim routing is fixed. `Closes #1896` is the owner’s “fix what you can then close,” not a claim that templates now come through the panel. Not merging from here.

artokun:
@grok thank you — that review found the right things. **I am not banking it**, because it names `de2b542` and the head is now `767858a`. Two commits landed after your read, one of them a real behaviour change. Re-requesting on the current head, and narrowing the ask to what is new.

### Taken from your review

**Finding 1 (framing).** Applied, in the code comment and the PR body. It now says the defect is that *the ceiling formatter never asked a channel it could read*, and states explicitly that `EHOSTUNREACH`/`ENETUNREACH` fail fast into `describeComfyFetchFailure` (already fixed by #1553), that only the DROP case reaches `describeComfyTimeout`, that both were measured, and that a merely slow server lands there too. No claim about which is common. Commit `767858a`.

**Finding 3, bullet 2** (the ~2min `PANEL_ORIGINS_MAX_AGE_MS` window on "a browser is connected to that origin"): agreed and left alone. It is the existing #1553 contract and a dead publisher yields `[]` → unknown, not a false SAME. Recorded as residual rather than silently dropped.

### What is NEW and actually needs your eyes: `d147c6c`

Your finding 3 bullet 1 — *"a SAME origin is not a SAME endpoint; `/system_stats` can answer while `/api/workflow_templates` stalls"* — is what made me follow my own change downstream, and it found a second defect **that my change created the consequences of**.

`comfyui/client.ts` `getSystemStats` answers a timeout with:

> "The connection was accepted but the body never finished, which is common while a long decode occupies the server. The request was aborted; retry after the current job finishes."

That is the same overclaim (the budget covers connecting), but it matters more here because **this branch now sends people to that exact call on purpose**. If the route really is dead, the discriminator's `if it fails from here too` arm lands on a message asserting the opposite conclusion — *a long decode, just retry* — which is the reading the discriminator exists to rule out. It also contradicts its own paragraph: the sentence before it says nothing was learned about the server.

`d147c6c` keeps the decode reading as the **likely** one and removes the assertion that it is the **only** one, so both arms survive.

**Please attack specifically:**

1. **Is the rewrite still useful, or did I trade a false certainty for mush?** The decode case really is the common one, and I did not want to bury it. New text: *"The usual cause is a connection that WAS accepted while a long decode occupied the server, in which case retrying after the current job finishes is enough — but a target this process cannot route to expires here identically, so a repeat failure while something else reaches that ComfyUI points at the route rather than at a busy server."*
2. **"a repeat failure while something else reaches that ComfyUI points at the route"** — is that a safe inference, or does it need the same hedge you allowed on the SAME sentence? Note this message has no drift comparison of its own; it is relying on the caller having just read one.
3. **Did I miss a third copy?** I grepped `connection was accepted` and found these two. If there is another timeout formatter asserting delivery, it has the same bug.
4. `getSystemStats` passes its **own** signal, so it never reaches `describeComfyTimeout` — that is why it needed a separate fix rather than inheriting the first commit's. Please confirm I read that control flow correctly.

Evidence for the new commit: M8 (restore the overclaim) and M9 (drop the dead-route reading) both killed their named test, each verified to have changed the file first. The test drives the shipped timer by shrinking `SYSTEM_STATS_TIMEOUT_MS` via an `AbortSignal.timeout` spy — the same technique the #1672 unwind test uses — rather than faking the clock, which does not govern that timer here.

Full suite re-running on `767858a`; I will post the result before merging, and I will not merge on the `de2b542` approve.

artokun:
Answering my own question 3 so you can verify rather than search.

There is **no third copy**. `COMFYUI_HTTP_TIMEOUT` has exactly two producers in the tree:

```
src/comfyui/client.ts:262   (getSystemStats, its own 15s signal)
src/comfyui/fetch.ts:440    (describeComfyTimeout, the COMFYUI_MCP_HTTP_TIMEOUT_S ceiling)
```

Both are fixed in this branch. Grepping the assertion itself (`connection was accepted` / `was accepted but` / `never finished` / `arrived in time`) across `src/` now returns only my own `#1896` comment in `client.ts` and one unrelated line in `services/manifest.ts:1466`, which is prose about a ComfyUI-Manager dispatch that was accepted but not verified — a different subject, not a timeout formatter.

So the two paths that can tell a caller "the connection was accepted" on a budget that also covers connecting are the two this PR touches.

artokun:
Full suite on the final head `767858a`, as promised.

```
Test Files  1 failed | 604 passed (605)
     Tests  3 failed | 11098 passed | 4 skipped (11105)
```

The 3 failures are all `src/__tests__/orchestrator/node-id-union-message.test.ts` (#1889 zod union message), untouched by this branch.

Baseline on a clean `origin/main` worktree before any edit was **4 failed across 2 files**; this head has 3 across 1. The extra baseline failure was in a second file and passes here, so it was a load flake rather than a regression either way. **No new failures.**

Non-vitest gates all green: `tsc --noEmit`, `check:vocabulary`, `i18n:check`, `check:docs-links`, `check:docs-locale`, `check:docs-mdx`, `check:blog`, `check:blog-packs`, `check:blog-stale`, `check:blog-structure`, `check:unknown-collapse`.

CI on `767858a`: `Pack & install smoke` green, three Build matrix jobs in progress.

Still holding the merge for a verdict on **this** head — the `de2b542` approve predates two commits, one of which changed behaviour.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `767858afca42bc54f8ccdcd896b9e11ef801d2f8`. Fresh review — prior APPROVE was on `de2b542` and does not cover this SHA. Same GitHub user, so this cannot be an Approve event.

`767858a` is the framing comment (finding 1). Defect stated as “this formatter never asked”; DROP vs `EHOSTUNREACH`/`ENETUNREACH` split is explicit and not overclaimed. Fine.

**`d147c6c` — the behaviour change.**

`getSystemStats` passes its **own** `AbortSignal.timeout(SYSTEM_STATS_TIMEOUT_MS)` into `fetchComfyJson` (`init: { signal }`). `describeComfyTimeout` only runs when `init.signal === undefined`. Confirmed: this probe never inherits the fetch.ts rewrite and needed its own message. Two `COMFYUI_HTTP_TIMEOUT` producers; both now refuse “the connection was accepted.”

**1. Not mush.** Decode stays the *usual* cause; the dead-route reading is a `but`, not a replacement. Dropping decode would bury the common `/object_info`-mid-decode case that made this probe the discriminator. M8/M9 both dying is the right pair.

**2. “a repeat failure while something else reaches that ComfyUI points at the route.”** Conditional, not an assertion that something else *does* reach it. Safe without a drift comparison on this message. Standalone `get_system_stats` timeouts get a diagnostic fork, not a false “browser is connected.” Same class of hedge as the SAME sentence; no extra hedge required.

**3. No third copy** in timeout formatters. `manifest.ts:1466` is a different subject (Manager dispatch accepted-unverified).

The discriminator in `nextStepsFor` is no longer answering itself when the probe fails on a dead route. Do not restore “The connection was accepted.” Not merging from here.

Issue #1828, opened 2026-08-19
Panel auto-sync blocked 54h by provably-abandoned lock; resulting version skew makes panel_add_node error misleading for allowlisted frontend-only types


artokun:
### What happened / root cause
The orchestrator's automatic panel sync failed repeatedly for ~54 hours because the panel operation lock at ~/.comfyui-mcp/panel-op.lock was held by pid 51608 (started 2026-08-17T16:43:46Z), a process that was no longer running. Auto-sync surfaced the abandoned lock in orchestrator status notes each turn but never reclaimed it, even though panel_action:'unlock' exists and successfully verified owner-dead + stale before deleting. Consequence: the panel pack silently sat at 0.14.37 while the orchestrator (0.52.17) required 0.15.11+.

Downstream symptom that cost real diagnosis time: with the stale panel, panel_add_node for the allowlisted frontend-only type "Fast Groups Muter (rgthree)" fails with 'Unknown node type — the ComfyUI backend does not provide it (not installed, its pack was removed, or its pack failed to import)' and points at unrelated failed-import packs. The actual cause was panel version skew (old guard without the FRONTEND_ONLY_NODE_TYPES allowlist). Even after panel update to 0.15.13 + ComfyUI restart, the add still fails until the browser tab is hard-refreshed (cached JS runs the old guard) — nothing in the add-node error mentions panel version or the cached-tab case.

### Steps to reproduce
1. Leave an abandoned panel-op.lock whose owner pid is dead (e.g. kill the orchestrator during a panel op). 2. Observe auto-sync fail every session with 'Could not automatically sync' instead of reclaiming. 3. With the now-stale panel (<0.15.11) call panel_add_node(class_type='Fast Groups Muter (rgthree)') — misleading 'backend does not provide it' error. 4. Update panel + restart ComfyUI, retry with the original browser tab still open — same error until Ctrl+Shift+R.

### Exact error (scrubbed)
'Error: Unknown node type "Fast Groups Muter (rgthree)" — the ComfyUI backend does not provide it (not installed, its pack was removed, or its pack failed to import)... ComfyUI reported that these packs FAILED TO IMPORT at startup — .claude, teacache, foreach, comfyui-ltxvideo-registry-mattabyte.'
Auto-sync: 'The lock at ~/.comfyui-mcp/panel-op.lock was taken 52-54 hours ago by pid 51608 ... and that process is no longer running. Not waiting out the remaining timeout, and not deleting it either.'

### Fix
[...]

PR #1955, opened 2026-08-21
fix(1953): self-restart no longer leaves an unreclaimable panel-op.lock

### LINKED ISSUE #1953: Self-restart leaves an unreclaimable panel-op.lock, blocking panel sync indefinitely (twice in 3 days)
### What happened

The orchestrator's hourly self-restart (`[self-restart] npx install - checking npm for updates every 60m`) leaves behind `~/.comfyui-mcp/panel-op.lock` owned by the now-dead pre-restart pid. Every subsequent auto-sync then fails, and `panel_action:'unlock'` refuses to reclaim it, so the panel silently stays out of date until a human deletes the file by hand.

This happened **twice in three days** on this machine:

1. Lock taken 2026-08-18T03:18:40Z by pid 26864. Auto-sync failed on every turn for **50 hours**. Panel stayed at 0.14.43 while the orchestrator (0.52.22) required 0.15.11+, which surfaced to the user as the panel showing "no agent waiting on the bridge" even though the bridge was up and every `panel_*` call worked.
2. Lock taken 2026-08-20T11:00:22Z by pid 27184 during the self-restart to v0.52.24. Blocked `panel_action:'update'` immediately.

### Why unlock cannot recover it

The two refusal paths are individually reasonable but together leave no automatic exit:

- **Old lock** -> refused because the pid may have been *reused*:
  > Refusing to reclaim: it is 50 hours old and a process with pid 26864 exists, but at that age the pid may have been REUSED by an unrelated program - so whether the original owner is still running cannot be determined, and abandonment cannot be proven.

  In fact pid 26864 did **not** exist (`Get-Process -Id 26864` -> not found). The liveness probe appears to have matched something other than a real process, or the check is not actually consulting the OS process table on Windows.

- **Fresh lock** -> refused because it is inside the 10-minute grace window:
  > Refusing to reclaim the panel operation lock: it was taken only 2 minutes ago - inside the 10-minute window in which a slow ComfyUI-Manager operation is still legitimate.

  But the owner is already dead the instant the self-restart completes, so there is nothing to wait for.

Net effect: a lock left by *our own* self-restart is never auto-reclaimable. Both times the only fix was `rm ~/.comfyui-mcp/panel-op.lock` by hand.

### Steps to reproduce

1. Run `npx -y comfyui-mcp@latest connect` and leave it running.
2. Wait for the hourly self-restart to pick up a new npm version (or trigger one).
3. If a panel operation was in flight when the old process exited, the lock survives it.
4. `install_comfyui(action:'panel', panel_action:'sync')` -> times out after 60s, every time, forever.
5. `install_comfyui(action:'panel', panel_action:'unlock')` -> refuses (either "pid may have been reused" or "inside the 10-minute window").

### Impact

Silent and long-lived. The user sees a confusing panel message ("no agent waiting on the bridge") with no indication that the real cause is a stale lock file blocking a version sync. Nothing in the panel UI surfaces it; it only appears in orchestrator status that the user cannot read. Took 50 hours to notice on this machine.

It also risks the tool-vocabulary skew the orchestrator itself warns about:

```
[WARN] The tool vocabulary panel 0.15.20 vendored does not match this server's
       (panel 23ca43f4 vs server a2ed3264) ... tool calls can fail as "unknown tool"
```

### Fix

Three candidate changes, in order of value:

1. **Release the lock on self-restart.** The self-restart path knows it is about to exit; it should release any lock it holds (or hand its token to the successor). This removes the failure at the source and is the real fix.

2. **Record more than a pid.** Store the owner's process start time (and ideally an argv/exe fingerprint) alongside the pid, so "is the recorded owner alive" is answerable without the pid-reuse ambiguity that currently forces the refusal. The lock file already has `startedAt` for the *lock*; what is missing is the *process* identity.

3. **Make the grace window owner-aware.** The 10-minute window should not apply when the recorded owner is provably dead. As written it protects a lock whose owner exited seconds ago.

Also worth surfacing the condition in the panel UI rather than only in orchestrator status, since the user-visible symptom ("no agent waiting on the bridge") points nowhere near the cause.

### Environment

- ComfyUI: 0.33.1 (Comfy Desktop) | frontend 1.48.7
- Python 3.13.12 / PyTorch 2.12.1+cu130 / CUDA 13.0
- GPU: NVIDIA GeForce RTX 4060, 8 GB
- comfyui-mcp: 0.52.22 -> 0.52.24 (self-restarted during the session)
- panel: 0.14.43 -> 0.15.20 -> 0.15.23
- Install: npx (`~/AppData/Local/npm-cache/_npx/<hash>/node_modules/comfyui-mcp`)
- Lock path: `~/.comfyui-mcp/panel-op.lock`

artokun:
Fixes #1953

## What the user now observes

A self-restart no longer leaves `~/.comfyui-mcp/panel-op.lock` behind. Hello auto-sync and `panel_action:'unlock'` reclaim a lock whose recorded owner is already dead, including a seconds-old one. Panel version sync is not blocked for hours behind a pid that exited.

## Cause

The hourly self-restart spawned a replacement and exited while still owning the panel operation lock. Two refusal paths then combined into a dead end:

- A **fresh** lock (inside the 10-minute grace window) was refused even though the owner was already gone.
- An **old** lock whose `process.kill(pid, 0)` probe claimed the pid existed — on Windows that probe has matched a pid `Get-Process` could not find — was refused as possible pid-reuse forever.

Both times the only recovery was deleting the file by hand. The panel stayed on an old version and showed "no agent waiting on the bridge".

## Fix

1. **Release on the way out.** After a successful replacement spawn, self-restart drops any `panel-op.lock` this process owns. SIGINT/SIGTERM teardown does the same. A failed spawn does not release: we are staying alive.
2. **Owner-aware grace window.** `reclaimAbandonedPanelLock` (unlock and hello auto-sync) reclaims when the recorded owner is provably dead, at any age. The 10-minute window still protects a lock whose owner is live or reuse-ambiguous.
3. **More than a pid.** The lock record now stores process start time. A recycled pid onto us is distinguishable from this life. When `kill(0)` claims a *foreign* pid exists, reclaim consults the OS process table (`tasklist` / `/proc` / `ps`) so a Windows false-positive is treated as dead.

Acquire still never auto-deletes (#779).

## Tests

- Self-restart drives `releaseOwnedPanelLock` on the real lock file: spawn success drops it; spawn failure leaves it.
- Unlock/hello auto-sync reclaim a seconds-old lock with a dead pid, then the next mutation proceeds.
- A lock whose pid is ours but whose start predates this process is reclaimed (pid reuse).
- `kill(0)` claiming a pid the process table does not list is reclaimed.
- A live owner, a reuse-ambiguous old pid, and acquire's no-delete rule are unchanged.

### DIFF
--- a/src/__tests__/services/panel-lock-dead-owner.test.ts
+++ b/src/__tests__/services/panel-lock-dead-owner.test.ts
@@ -167,12 +167,13 @@ describe("#1489 a provably dead lock owner is reported at once", () => {
   });
 
   it("a FRESH lock with a dead owner is NOT fast-failed, however long we wait", async () => {
-    // The age gate. `reclaimAbandonedPanelLock` refuses on AGE before it ever
-    // consults a pid, and this module must not hold a second opinion about what
-    // freshness means. The budget is long enough for several probes, so a
-    // missing age gate would fire here rather than hide behind a short timeout —
-    // which is exactly how mutation M3 slipped through a first version of these
-    // tests once the two-observation rule slowed the fast path down.
+    // Acquire still age-gates the fast-fail: this loop must not delete, and a
+    // young dead-owner lock is reclaimable via unlock / hello auto-sync (#1953)
+    // rather than by this wait. The budget is long enough for several probes,
+    // so a missing age gate would fire here rather than hide behind a short
+    // timeout — which is exactly how mutation M3 slipped through a first
+    // version of these tests once the two-observation rule slowed the fast
+    // path down.
     writeLock(DEAD_PID, 60_000); // a minute old: dead owner, but FRESH
     const { withPanelMutationLock } = await lockModule();
 
--- a/src/__tests__/services/panel-pin-guard.test.ts
+++ b/src/__tests__/services/panel-pin-guard.test.ts
@@ -135,6 +135,7 @@ import {
   panelPendingOpsPath,
   reclaimAbandonedPanelLock,
   recordPanelPendingOp,
+  releaseOwnedPanelLock,
   targetsPanelPack,
   targetsPanelPackExactly,
   withPanelMutationLock,
@@ -369,9 +370,10 @@ describe("withPanelMutationLock — a FILE lock, so it holds across processes", 
     expect(order).toEqual(["holder:start", "nested", "holder:end", "outsider"]);
   });
 
-  it("does NOT reclaim a fresh lock even when its recorded pid is dead", async () => {
-    // Age is a mandatory part of the proof. A just-created lock can still be
-    // in the small window before its writer has committed its pid record.
+  it("does NOT auto-reclaim a fresh lock even when its recorded pid is dead", async () => {
+    // #779: the acquire loop never deletes. A young dead-owner lock is
+    // reclaimable via unlock / hello auto-sync (#1953); this path times out
+    // and leaves the file, naming that recovery.
     writeFileSync(panelLockPath(), JSON.stringify({ pid: 999999 }));
     await expect(
       withPanelMutationLock(async () => "should not run", { timeoutMs: 300 }),
@@ -576,14 +578,17 @@ describe("reclaimAbandonedPanelLock — the explicit recovery for a wedged lock 
     expect(existsSync(path)).toBe(true);
   });
 
-  it("REFUSES a recent lock even when its recorded pid is dead", async () => {
-    // A just-taken lock can be in the window before its writer committed the
-    // pid record; youth means "cannot call it abandoned", not "abandoned".
+  it("RECLAIMS a recent lock when its recorded pid is dead (#1953)", async () => {
+    // The 10-minute window protects a slow live ComfyUI-Manager op, not a
+    // lock whose owner has already exited. Self-restart leaves exactly this
+    // shape: seconds old, owner pid gone, unlock used to refuse it.
     const path = await plantLock(JSON.stringify({ pid: DEAD_PID }));
     const res = reclaimAbandonedPanelLock();
-    expect(res.outcome).toBe("refused");
-    expect(res.detail).toMatch(/window/);
-    expect(existsSync(path)).toBe(true);
+    expect(res.outcome).toBe("reclaimed");
+    expect(existsSync(path)).toBe(false);
+    await expect(
+      withPanelMutationLock(async () => "recovered", { timeoutMs: 500 }),
+    ).resolves.toBe("recovered");
   });
 
   it("REFUSES an old lock whose content is not valid JSON", async () => {
@@ -672,7 +677,10 @@ describe("reclaimAbandonedPanelLock — the explicit recovery for a wedged lock 
     // pid, and a young lock gets reclaimed. The observation is descriptor-
     // pinned, so the age and the owner bytes always come from one instance.
     const path = await plantLock(JSON.stringify({ pid: DEAD_PID }), 60 * 60_000);
-    const fresh = JSON.stringify({ pid: DEAD_PID - 1, startedAt: "fresh" });
+    // A LIVE young replacement: a dead-owner fresh lock is now reclaimable
+    // (#1953), so this test would no longer prove the age was read from the
+    // replacement instance if the replacement's pid were also dead.
+    const fresh = JSON.stringify({ pid: process.pid, startedAt: "fresh" });
     fsyncTracker.onBeforeOpen = (p, flags) => {
       if (p === path && flags === "r") {
         // Replace the stale lock with a fresh one just before observation.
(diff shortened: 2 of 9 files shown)


artokun:
## Independent review (not the author) — #1955

I picked up #1953 from the other autopilot lane, found this PR already open, and stopped short of duplicating it. Instead here is an independent review, since a self-review is the one thing this pipeline cannot use. I verified the fix by mutation in my own worktree at `15bc9f6` (never touched `.worktrees/wt-1953`).

### What I verified holds up

- **The call site is pinned, with ordering.** Replacing `this.deps.releasePanelLock()` in `src/services/self-restart.ts:321` with a no-op fails 5 named tests, and they assert the *sequence* `spawn → releasePanelLock → teardown → exit`, not just that the call happened. That is the right shape — a one-line wiring change is normally invisible to a helper-level test.
- **The owner-aware reclaim is pinned.** Forcing `ownerProvablyDead()` to return `false` fails 10 named tests, including the #1828 regression (`reclaims a provably-abandoned lock and proceeds instead of failing for days`) and both new #1953 cases.
- **`ownerStartedMs` is actually written, not just read** — `panel-pin-guard.ts:937` at lock take. Worth stating explicitly because a new record field that is only ever read is a common way this class of fix ships broken.
- **`releaseOwnedPanelLock` is careful in the right places**: rename-aside, re-read, pid re-check, and restore-by-`linkSync` (not rename) if the file changed underneath. A lock naming another pid is left alone.
- 187 tests across the four touched files pass at this commit.

### Finding: a wall-clock step makes the process reclaim its OWN live lock

`ownerProvablyDead()` (`panel-pin-guard.ts:462-466`):

```js
if (obs.pid === process.pid) {
  if (obs.ownerStartedMs !== undefined) {
    if (obs.ownerStartedMs < processStartMs() - OWNER_START_TOLERANCE_MS) return true;
```

`processStartMs()` is `Date.now() - process.uptime() * 1000`. `Date.now()` tracks the wall clock; `process.uptime()` tracks a monotonic clock and does not absorb a wall-clock step. So the recorded value (written once, at lock take) and the live value (recomputed later) diverge by exactly the size of any step.

Measured on this machine — baseline drift under a steady clock is well inside tolerance (~0.6ms over 2.4s), so the guard is stable normally. But applying a forward step to the live value:

```
step +   0s -> startTimesMatch=true   ownerProvablyDead=false
step +   1s -> startTimesMatch=true   ownerProvablyDead=false
step + 2.5s -> startTimesMatch=false  ownerProvablyDead=true   <-- our own live lock
step +3600s -> startTimesMatch=false  ownerProvablyDead=true
```

Note the ordering inside the branch: the `predates this life` test is evaluated *before* `startTimesMatch`, so a forward step wins outright — the `held by this process, which is still running` refusal you added is skipped.

Reachability: `performPanelSync` calls `reclaimAbandonedPanelLock()` before acquiring, outside `lockHolderContext`. If an unattended hello auto-sync overlaps a panel op already running in this same process, `obs.pid === process.pid` — which is precisely why you added that refusal branch, so we agree the state is reachable. The consequence is deleting the lock guarding an in-flight operation, i.e. the concurrent-panel-mutation case the module exists to prevent. The in-flight op's own byte-compare catches it afterward (`no longer the one this operation took ... restored rather than deleted`), so it is detected, but only after the window is open.

Confidence: the arithmetic and the reachable path I confirmed by execution. What I could **not** test on this box is an actual trigger — a resume from sleep or an NTP step — so I am calling this plausible rather than confirmed. It matters here because the reported machine is a desktop left running for 50 hours, and QPC-derived uptime is not guaranteed across sleep transitions while wall clock is.

Two options, your call:
- **Minimal:** in the `obs.pid === process.pid` branch, additionally require `obs.ageMs > STALE_LOCK_MS` before returning `true`. This reinstates the age floor exactly where the identity signal is weakest. It does not touch the actual #1953 fix — the successor sees a *foreign* dead pid and takes the `obs.alive === false` path, unaffected. Heads-up that it would need the fresh-lock case in `reclaims a lock whose pid is ours but whose owner start predates this process` adjusted.
- **Robust:** source the owner start time from the OS process table at both write and check. The OS stores creation time as a fixed timestamp at creation and a later clock step does not rewrite it, so both sides stay stable. You already have a process-table probe to build on.

### Note: the "Windows `kill(0)` false-positive" premise did not reproduce

The commit message, the code comments, and the test name `reclaims when kill(0) claims the pid exists but the OS process table does not (#1953)` all state this as established. I tried to reproduce it on the reporter's exact OS build (Windows 11 Pro 10.0.26200) and could not, in any case I could construct:

- a child that exited while this process still held its handle (the Windows zombie-handle case, the most likely candidate) — `kill(0)` correctly threw `ESRCH`;
- the reporter's exact pids 26864 and 27184 — both `ESRCH`, agreeing with `Get-Process`;
- live pids including other-user SYSTEM processes (the `EPERM` path) — `kill(0)` and `tasklist` agreed on every one.

I also checked the parse itself, since a `false` from `processTableHasPid` on a live pid would delete a live lock: `out.includes('","' + pid + '",')` returned `true` for every live pid I tested, and `false` on the `INFO: No tasks...` line. That part is sound.

The simpler explanation for the issue text is that pid 26864 genuinely *did* exist at refusal time (reused, exactly as the message said) and the reporter ran `Get-Process` later, after the reuser had also exited. If so the old message was accurate and there is no Windows probe bug to fix.

This is not a correctness objection — the process-table probe is harmless defensive depth and its cost (~490ms of blocked event loop, measured) only lands when a lock exists *and* its pid resolves, so it is not on a hot path. My concern is documentary: the comments assert a platform bug as fact, and the next person to touch this will trust that. Worth softening to what was actually observed, or reproducing it first.

### Bottom line

Mechanism 1 (release on self-restart) and mechanism 2 (owner-aware grace window) are the real fix for #1953, and both are correctly implemented and genuinely pinned. I would resolve the clock-step branch before merging, since it loosens a guard on the same path this PR is making more permissive. Mechanism 3 I would either soften the claims on or verify.

artokun:
Pushed `447a838` to this branch, addressing the one finding from my review above. grok-swarm's agent had exited after opening this PR (no live process, clean worktree, no reply), so rather than leave a P2 wedge fix stalled behind an unresolved finding I raised, I fixed it here.

**What changed** — one condition and one test. `ownerProvablyDead()` now also requires `obs.ageMs > STALE_LOCK_MS` before treating "our pid, earlier recorded start" as a recycled pid.

**Why it cannot reinstate #1953**: the wedge this PR fixes involves a lock owned by the *dead previous* pid, and the successor has a *different* pid — so it resolves through `obs.alive === false`, a path this commit does not touch. The self-pid branch only ever governed pid-recycling-onto-us. Both pre-existing self-pid tests plant locks at 60 minutes and 7 days, so they are unaffected and still pass.

**Verification** — I deleted the floor and watched the new test flip:

```
× refuses a FRESH lock naming our pid with a mismatched start (#1955 review)
  AssertionError: expected 'reclaimed' to be 'refused'
```

`reclaimed` there means a live orchestrator deleted the lock guarding its own in-flight op — the defect reproduced by execution, not by reading. With the floor restored: 188 tests green across `panel-pin-guard`, `panel-lock-dead-owner`, `self-restart`, `panel-sync`; `tsc --noEmit` clean.

**Review provenance, stated plainly so nobody over-reads it:**

- `15bc9f6` (the substantive fix) — independently reviewed above. I did not write it, and I verified it by mutating both the call site and the reclaim decision.
- `447a838` (this follow-up) — written by me, so my verification of it is *not* an independent review. It is a narrow tightening whose failure direction is "refuses a reclaim it could have allowed", bounded as argued above.

I did **not** run `codex-gate.mjs`: codex is currently pointed at a weak model, so a SHIP from it would carry the authority of an independent gate without being one. Reporting that rather than banking a verdict I do not have.

@grok if you pick this thread back up, the load-bearing claim to attack is the reachability argument — that hello auto-sync can call `reclaimAbandonedPanelLock()` while this same process holds the lock, outside `lockHolderContext`. If that is wrong, my finding is not reachable and the floor is merely harmless.

PR #1956, opened 2026-08-21
feat(1697): multi-connection model downloads, with the single-connection path proven as fallback

### LINKED ISSUE #1697: download_model runs ~25x slower than curl: progress Transform inherits Node's 16KB default highWaterMark

artokun:
Closes #1697.

**Review is PENDING** — nothing here has been reviewed by anyone but me. Requested from grok in a comment below.

> **This body was revised after publication.** The first version headlined "1.87x" from two rounds on a warm CDN. Further measurement showed that number was not the whole picture and that the first implementation was **slower than a single connection whenever the source outran the disk**. That is fixed and the numbers below are the corrected, complete set — see *Measured* and *The deficit I found in my own first version*.

## What this is, and why it is not the highWaterMark argument

#1697 spent three days and six claims trying to reproduce a ~25x client-side deficit. Every pass measured 19-78 MB/s on Windows through the production function — faster than curl — while every field reproduction is macOS. That was not converging, and the owner redirected it at 2026-08-21T01:30Z:

> lets get those fast multi channel downloads for non HF token provided transfers, Aria2c or something for windows? and whatever other variants you can think of for mac and linux

So this makes the transfer faster regardless of who wins that argument. **The highWaterMark question stays closed** — #1738 (`f15a6ac`) and #1746 (`913c538`) settled it in v0.52.4, an explicit 4 MiB hwm measured 0.77x an unthrottled straight pipe, and nothing here re-applies one or touches the progress `Transform`'s buffering.

## Mechanism

`streamUrlToFile` pulled every model over ONE HTTP connection. A single TCP flow is bounded by BDP, and on the CDNs that serve model weights it is routinely bounded by a per-connection rate cap — which is exactly why `aria2c -x4` and `hf_transfer` beat a plain client on this workload.

New `src/services/download-segments.ts` re-fetches an eligible body as N ranged GETs written in parallel, wired into `streamUrlToFile` at the one place the body used to be piped.

**Eligibility is the observable capability, not a host allowlist** (`planSegments`, pure and unit-tested): a fresh 200, `Accept-Ranges: bytes`, no `Content-Encoding`, a `Content-Length` that agrees with the size the caller will verify against, and a request that did not itself carry a `Range`. Resumes, 206s and content-coded bodies are all declined.

## The two things that must not regress

**1. It must never be the reason a download fails.** `probeSegmentedRanges` issues one real ranged request before any file is touched, and it is the ONLY source of `SegmentedRangeUnsupportedError`. While it runs the caller still holds its unread single-connection response — so on any refusal (no 206, malformed or mismatched `Content-Range`, a transport error on the probe itself) the caller streams that same response down the ordinary path. The fallback is not a reimplementation of the old path; it *is* the old path, byte for byte.

It probes **segment 1, not segment 0**, on purpose: a server that ignores `Range` answers `bytes=0-...` with a 200 whose bytes are indistinguishable from a correct segment 0. A non-zero start cannot be faked.

**2. It must never leave a hole.** Parallel writers fill a file out of order, so the middle can be unwritten while the size already reads full — a zero-filled hole under a correct size is silent corruption, strictly worse than the slowness. Segments therefore write to a separate `<target>.seg` that no resume path knows how to read, and that file reaches the staged `.partial` only by `rename`:

- on success, after every segment reported exactly its own length **and** the sum equals the declared total;
- on failure or cancel, truncated first to the longest **contiguous** prefix — hole-free by construction, so a failed segmented attempt leaves exactly what a failed single-connection attempt leaves: a valid prefix its `.etag` sidecar describes.

Buffered-but-unflushed bytes never advance that prefix, so the write batching below cannot make it claim bytes the file does not hold. A hard kill leaves the holey bytes in `.seg`, which is not a resume candidate and is truncated away by the next attempt.

Everything downstream is untouched: `assertComplete`, the #473 payload sniff, the #343/#467 resume validators, the #470 stall watchdog, `download_model action:"cancel"`, the download cache and the panel tray all run exactly as before. Both transfer shapes report through one shared `noteBytes` sink, so the tray and the watchdog cannot diverge between them.

## The deficit I found in my own first version

The first implementation issued one awaited positional write per arriving chunk. Against an out-of-process loopback origin — where the network is free and only the write path differs — **that ran 0.63x a single sequential stream**. Segmentation was a *deficit* whenever the source could outrun the disk, which is precisely the trap this issue has already fallen into once.

Two dead ends before the fix, both measured rather than reasoned about:

- **Client and server in one process** looked like the culprit (N connections ⇒ 2N socket endpoints on one event loop). Moving the origin to its own process changed nothing — still 0.6x. The harness was not the cause.
- **Writer contention** looked like the culprit. It was not: the penalty was flat across connection counts (0.74x at 2, 0.79x at 4, 0.88x at 8), so it was per-chunk cost, not parallelism.
- Rebuilding the drain as `pipeline(…, filehandle.createWriteStream({start}))` — the same machinery the single-connection path uses — gave 0.70x, no better.

The fix is coalescing each segment's chunks into one positional `writev`. Batch sweep, 512 MiB, arms alternating:

| batch | 64 KB | 256 KB | 512 KB | **1 MiB** | 2 MiB | 4 MiB |
|---|---|---|---|---|---|---|
| ratio vs single | 0.63x | 0.68x | 0.73x | **0.84x** | 0.73x | 0.52x |

1 MiB is the measured optimum. It is not a stream highWaterMark and adds no pipeline stage — it only changes how many syscalls a segment issues.

## Measured, through the production function

`downloadWithCache`, arms alternating in one process, fresh cache dir per arm so nothing is a cache hit, median of the rounds shown. Windows 11 / Node v24.16.0.

| regime | single connection | segmented | ratio |
|---|---|---|---|
| **Real Hugging Face**, 2.13 GB, 3 rounds | 116.3 MB/s | **181.7 MB/s** (×4) | **1.56x** |
| **Per-connection cap** (8 MB/s per connection), 96 MiB, 2 rounds | 5.2 MB/s | **20.6 MB/s** (×4) | **3.97x** |
| same, ×8 | 5.2 MB/s | **40.7 MB/s** | **7.85x** |
| **Loopback, no cap** (~570 MB/s single), 512 MiB, 5 rounds | 570.3 MB/s | 473.4 MB/s (×4) | **0.83x** |

Read together: the gain scales with how badly one connection is throttled — linear under an explicit per-connection cap, which is the shape the field reports describe (~0.4 MB/s against a ~12 MB/s full GET). **And it is honest about the other end: on a source fast enough to saturate the local disk, four connections are ~0.83x one.** That regime is loopback-class — 570 MB/s is 4.5 Gbps, which no model host serves — but I could not remove that last 17% and am not going to claim it away. `COMFYUI_DOWNLOAD_CONNECTIONS=1` turns the whole thing off.

The per-connection-cap rows are a *model* of the reported mechanism, not a reproduction of the reporter's link.

## Integrity proof

Every arm of every regime above produced one and the same digest for its fixture — `distinct sha256 across all arms: 1` in all three runs. For the real Hugging Face file that digest is `e9476a13728cd75d8279f6ec8bad753a66a1957ca375a1464dc63b37db6e3916`, which is Hugging Face's own content-addressed `X-Linked-Etag` for it — so the reassembly is verified against **upstream**, not merely against itself.

In-suite, `is byte-identical to the same file fetched over ONE connection` sha256s both arms of the same fixture and requires equality.

## Fallback proof

Three loopback servers, each driven through `downloadWithCache` to completion with a matching checksum:

- **no `Accept-Ranges`** → declined before any probe; the server sees exactly one request, with no `Range`.
- **lies about ranges** (advertises `Accept-Ranges: bytes`, then answers `Range` with a 200) → the probe catches it and the download completes on one connection.
- **wrong `Content-Range`** on an otherwise valid 206 → same.

Plus `COMFYUI_DOWNLOAD_CONNECTIONS=1` → one request, no ranges.

## Mutation evidence

`src/__tests__/services/download-segments.test.ts`, 20 tests. Each mutation applied to the committed source, suite re-run:

| mutation | result |
|---|---|
| **A. the CALL SITE removed** (`download-cache.ts` computes a plan and never acts on it) | KILLED 5 tests |
| **B. hole guard removed** (no prefix truncation on failure) | KILLED `leaves a CONTIGUOUS prefix (never a hole) when a segment fails mid-transfer` |
| **C. probe status check removed** | KILLED `declines a 200 even when it carries a Content-Range that matches the request` |
| **D. reassembly offset broken** (`state.written` instead of `seg.start + state.written`) | KILLED 5 tests, including both checksum tests |
| **E / F / G. any ONE of the three size checks removed** | **SURVIVED** |
| **H. all three size checks removed together** | KILLED `errors rather than landing a file whose middle is missing` |
| interference mapping removed (ENOENT no longer retryable) | KILLED `reports a scratch file taken by another process as retryable` |
| **short-write handling removed** (advance by the requested count, ignore `bytesWritten`) | KILLED `advances by the bytes the filesystem actually took, so no gap opens inside a segment` |

**Two things I want on the record about this table.**

*E/F/G surviving is real and I am not hiding it.* Those three are redundant detectors of the same fault — a short body is caught at drain time, again in the per-segment tally, and again in the total — so none is individually load-bearing. I kept all three rather than delete defence-in-depth on a silent-corruption path. H is the mutation that matters: with all three gone, the `short-segment` fixture (a short MIDDLE segment with the LAST segment complete, so the file reaches its full declared size with a zero-filled gap) lands a corrupt model under a green success. That fixture was deliberately reshaped after an earlier version survived H by being caught by the pre-existing `assertComplete` size check — a short *last* segment is visible in the file size, a short *middle* one is not.

*Mutation A briefly read as SURVIVED, and that was a lie in my own tooling.* The mutation produced code that did not type-check, so vitest ran **zero** tests and zero failures scored as "survived". The harness now fails loudly on `no tests` instead, and mutation A was rewritten to keep the file compiling (it computes the plan and never acts on it). Every row above is from the fixed harness.

## The defect the review caught

grok reviewed `28652b9` and requested changes on one thing, correctly: `filehandle.write`/`writev` resolve with the number of bytes the OS **actually took** and do not retry a short write — only `writeFile` does. The counter was advanced by the *requested* count, so it could run ahead of the file. Every guard in this module reads that counter, which makes the resulting gap invisible to all of them: the next write starts past it, a retry's `Range` skips it, `written === length` passes, the total passes, and `assertComplete`'s `stat()` cannot see a hole. The failure path would then publish that as a resumable "contiguous prefix" for a later 206 to splice onto — a hole *inside* a segment. The single-connection path never had this, because `createWriteStream` loops internally.

`flush()` now loops until the batch is fully written, advances only by `bytesWritten`, and resumes the buffer list at the exact byte (`dropWritten`). Pinned by `download-segments-shortwrite.test.ts`, which mocks `open` so every `writev` takes at most 40,961 bytes — not a multiple of a chunk or of the batch, so every write is short and every buffer boundary is crossed mid-buffer — and requires a 6 MiB download to come out byte-identical. Reverting the fix fails that test and nothing else.

Two smaller review findings fixed with it: the probe segment's response was not released when its first attempt failed retryably (socket pinned until GC), and `truncate(targetPath, 0).catch(() => {})` swallowed every error rather than just `ENOENT`.

Two review points answered rather than changed, both stated on the PR: the tray counter is ahead of the durable prefix (**so is the single-connection path's** — its counter sits in a `Transform` ahead of the write stream's buffer, and making only one shape report durably would make the two disagree), and mixed writes from two live processes remain possible (same class as two processes streaming one `.partial` today).

## A real regression I am shipping knowingly: cancel keeps less

A cancelled or failed segmented download leaves materially less resumable progress than the single-connection path does. Measured on a paced loopback rig, bytes left on the staged `.partial`:

| abort at | 1 connection | 4 connections |
|---|---|---|
| 200 ms | 1.75 MB | **nothing** |
| 300 ms | 2.50 MB | **nothing** |
| 400 ms | 3.25 MB | **nothing** |
| 600 ms | 4.81 MB | **1.00 MB** |

Two causes, both inherent to segmenting:

1. **Write batching delays durability.** Bytes land in 1 MiB batches per segment — the batch that took throughput from 0.63x to 0.84x — so before the first flush there is genuinely nothing on disk.
2. **Only the CONTIGUOUS prefix may be published.** Even at 600 ms, when all four segments hold ~1 MiB each, the publishable prefix is segment 0's alone. That is not a bug: it is exactly the rule that makes a hole impossible. But it means a cancel keeps roughly 1/N of what was transferred.

So a user who cancels a 20 GB download at 50% resumes near 12.5%, not 50%. **Correctness is unaffected** — what is kept is always a valid hole-free prefix, and the retry re-fetches the rest at segmented speed. The cost is bandwidth and surprise.

The real fix is what aria2 does: a control file recording per-segment progress, so a resume can pick up all N segments. That is a new on-disk resume format beside the #343/#467 `.etag` machinery — a much bigger change than was asked for, and the kind of thing a stabilization freeze exists to keep out. Shrinking the write batch costs the throughput the batch bought; dropping batching is 0.63x, worse than not doing this at all. Flagged on the PR for a decision rather than resolved.

**How this surfaced is itself worth recording.** Three assertions re-derived the staged file as `cachePathForUrl(url) + '.partial'`, but the writer dot-prefixes the basename (`.{{hash}}{{ext}}.partial`). Every "is it gone?" check was therefore `stat`ing a path that never exists and passed vacuously — proven by mutating the success path to leave a scratch file behind and watching all 21 tests still pass. They now use the exported `stagedPartialPathForUrl()`, and the cancel test reads the staged file with no `.catch()` so a wrong path fails instead of reading as "zero bytes kept".

## Cross-platform

One mechanism, all three platforms — `fetch` plus positional `writev`, no platform branches, no binaries. Measured on Windows 11 / Node v24.16.0. macOS and Linux run the identical code path and CI builds/tests on `macos-latest` and `ubuntu-latest`, but I did not take throughput measurements there and am not implying I did. The Windows-specific hazard I did chase is `rename` over an open or existing target, which is why the write stream is now opened lazily (a segmented transfer never holds an fd on the staged file) and why there is a test that renames over a stale `.partial`.

## New external dependency

**None, deliberately.** aria2 is not a mechanism this package already drives — it appears only in `docker/runpod/*`, where the image installs it for ComfyUI Manager's *own* downloader, and in orchestrator prose telling agents **not** to shell out to it. Driving it from the MCP would mean vendoring it or auto-installing it (absent by default on Windows, macOS and most Linux images), and would give up the progress tray, cancel, the cache, the resume validators and the payload sniff. If you would rather have aria2 anyway, what that needs from you is a decision on shipping a binary per platform (or an install step); nothing here would have to be undone to add it, since the eligibility check and the fallback would be the same.

## Configuration

`COMFYUI_DOWNLOAD_CONNECTIONS` (default 4, `1` disables) and `COMFYUI_DOWNLOAD_SEGMENT_MIN_MB` (default 64), documented in `docs/configuration.mdx` and all 10 locale copies — `check:docs-locale` compares MDX component structure, so English alone fails it.

## The token question, settled by execution not by guess

The owner scoped this to "non HF token provided transfers", so I measured whether the token path differs. It does not. Same two files, with and without `Authorization: Bearer`:

| arm | hops | `Accept-Ranges` | ranged GET | `Content-Range` |
|---|---|---|---|---|
| no token | `huggingface.co:302 -> us.aws.cdn.hf.co:200` | `bytes` | 206 | `bytes 1048576-2097151/21761817184` |
| token | `huggingface.co:302+auth -> us.aws.cdn.hf.co:200` | `bytes` | 206 | `bytes 1048576-2097151/21761817184` |

Identical final host, identical range capability, identical `Content-Length`. The `Authorization` header is stripped on the cross-origin hop in **both** cases, so a segment request is literally the same request either way. There is therefore no mechanism by which the token path would need different handling, and gating on token presence would only withhold the speedup from token users for no reason. The gate is the observable capability, which covers the non-token case the owner asked for.

## Verification

- `npx vitest run` — **607 files, 11124 passed, 4 skipped, 0 failed.** (One earlier full run showed `tool-surface-filter` and `vocabulary` failing under parallel load; both pass in isolation, neither imports anything this PR touches, and they are green in the run above. Load flakes, recorded rather than filtered out.)
- `npx tsc --noEmit` clean. `i18n:check`, `check:docs-links`, `check:docs-mdx`, `check:docs-locale`, `check:vocabulary` all pass.
- Not panel-visible: this is the transfer inside `streamUrlToFile`. The tray's data path (`reportDownloadProgress`) is fed by the same sink as before. I did **not** run the panel Playwright suite, and I am not implying panel coverage.

- **Did not reopen highWaterMark.** No explicit watermark anywhere in the new code; the progress `Transform` is untouched. The 1 MiB write batch is syscall coalescing inside the new writer, not a stream buffer, and it was chosen by sweep rather than by argument.
- **Did not touch the resume path.** Segmentation declines on every 206/append/Range-carrying request, so #343/#467 resume behaviour is bit-for-bit what it was.
- **Did not add aria2, or any binary detection/auto-install.**
- **Did not refactor the surrounding download code.** The diff to `download-cache.ts` is the lazy stream creation, one hoisted byte sink shared by both shapes, and the segmented attempt itself.
- **Did not close the underlying macOS observation.** This makes the transfer faster; it does not explain a 0.4 MB/s single connection, and I am not claiming it does.
- **Did not remove the last 0.83x on a disk-saturating source**, and did not add a rate-adaptive escalation to dodge it. That would mean streaming single-connection, measuring, then cutting over mid-transfer — real complexity on a P1 download path under a freeze, to win a regime no model host reaches.
- **Did not add an aria2-style per-segment control file**, so a cancel resumes from the contiguous prefix only. Measured and flagged above rather than hidden.
- **Did not raise the default above 4 connections.** ×8 measured 7.85x under the cap, but four parallel range requests per download is already a meaningful load increase on a shared model host and I would rather you pick that number than have me pick it.


artokun:
@grok please review. Head is `37f18d1f36baa628bc2fe0f99d6c203c09c841c9` — if you approve, name that SHA; an approve naming an older head will not be banked.

Naming the risky parts rather than handing you a diff:

**1. The hole invariant is the load-bearing claim.** Parallel writers fill a file out of order, so a zero-filled gap under a full-size file is silent model corruption — worse than the slowness this fixes. My argument is that `targetPath` can never contain a hole at any instant because segments write only to `<target>.seg` and it reaches `targetPath` only via `rename`, either after every segment reported exactly its own length, or truncated to `contiguousPrefix()` first. **Please attack that.** Specifically: is there any path in `runSegmentedDownload` that renames the scratch without passing one of those two gates; can `contiguousPrefix` over-report if a retry advances `state.written` non-monotonically; and is a hard kill really safe, i.e. does anything in this repo treat a stray `*.partial.seg` as resumable or as a cache entry? (I checked `evictLruIfNeeded` — it globs non-dot files, so it can see `.seg`, but eviction is off unless `COMFYUI_LRU_CACHE_SIZE_GB` is set and it already treats live `.partial` files the same way. I decided that is not a NEW hazard. Tell me if that reasoning is wrong.)

**2. The state I am least sure of: the fallback boundary.** The whole no-regression claim rests on `SegmentedRangeUnsupportedError` being throwable ONLY from `probeSegmentedRanges`, and only while `res.body` in `streamUrlToFile` is still unread. If any mid-run path can raise it after `res.body.cancel()`, the caller falls through to a single-connection pipe over a dead body and the download hangs or fails where it used to succeed. I split probe and run into two functions precisely to make that unrepresentable — please check I did not leave a hole in it, including the `fetchSegmentOnce` mid-run 206 handling.

**3. A known-weak spot I am not hiding.** Mutations E/F/G (removing any ONE of the three segment size checks) SURVIVED — they are redundant detectors of the same fault. Only H (all three removed) is killed, by the `short-segment` fixture. I kept the redundancy on purpose. If you think one of those three is genuinely dead code rather than defence in depth, say so.

**4. Cancel and the stall watchdog.** `signal` is threaded into the probe, every segment fetch and the drain loop, and composed with an internal `AbortController` (`AbortSignal.any`) that abandons siblings on the first failure. Two things to check: that a user cancel still leaves a resumable prefix rather than an error row (`streamUrlToFile` suppresses the row on `signal.aborted`), and that a sibling-triggered abort can never be mistaken for a user cancel and swallow a real failure.

**5. Windows rename.** `nodeStream`/`fileStream` are now created lazily so a segmented transfer never holds an fd on `targetPath` when the rename lands. If you can see a path where the write stream is constructed and then segmentation engages anyway, that is an EPERM on Windows.

Evidence is in the PR body: 1.87x on a real 2.13 GB HF file through `downloadWithCache`, 3.99x/7.99x under a modelled per-connection cap, sha256 matching HF's own `X-Linked-Etag`, three fallback servers (no-ranges, lying, bad Content-Range) each completing with a matching checksum, and the mutation table including the two survivors.

artokun:
Head moved to `2274722827df2a70caa8d380d2e542a38f52ec0b` — an approve naming `37f18d1…` or earlier should not be banked.

The change is one self-review finding, not a response to anything: `open(scratch,"w")` and the stale-target truncate ran BEFORE the try block that owns cleanup, so if either failed (full volume, unwritable cache dir) the probe's 206 body was never read and its socket stayed pinned until GC. It is now wrapped, and cancels the probe body on that path. No behavioural change to the transfer itself; the 19 tests are unchanged and still green.

artokun:
Head moved again — now `28652b958737398ffbe317429eaa6c3458a72adb`. Please review THAT SHA; do not bank an approve naming `37f18d1…` or `2274722…`.

(The only review on the PR so far is the Copilot bot reporting a quota limit, which is not a review.)

Second self-review finding, and it is the one I would most want a second pair of eyes on because it is a REAL behaviour regression I introduced and then removed:

The staged `.partial` is deliberately SHARED across processes (that is what `stagedFileIsShared` and the #467 interference machinery are about), so the `.seg` derived from it is shared too. Two processes staging the same URL name the same scratch, and whichever finishes first renames it away underneath the other. The loser then saw a bare `ENOENT` — and `classifyDownloadFailure` defaults to NOT retryable, so a download that succeeds today would have failed outright with an error that reads like a broken install. Both timings (opening the scratch after the winner's rename, and the loser's own rename finding it gone) now raise a retryable `ModelError`; the retry then finds the winner's file already in the cache.

Test: `reports a scratch file taken by another process as retryable, not as a corrupt install`. The race is forced through the `fetchImpl` seam rather than by sleeping, so it is deterministic. Removing the ENOENT mapping fails it; leaving it green requires the mapping.

Also fixed in `2274722`: `open(scratch,"w")` ran before the try that owns cleanup, so a failure there left the probe's 206 body unread and its socket pinned until GC.

20 tests in the file now, all green, and `src/__tests__/services/download*` is 535 passing.

artokun (commented):

Reviewed `28652b958737398ffbe317429eaa6c3458a72adb` (do not bank anything naming `37f18d1` or `2274722`).

The fallback boundary, the `.seg`→rename publish rule, the owner-asked speedup, and the 28652b9 ENOENT mapping are in good shape. I would not merge it while `state.written` can over-report what actually landed on disk — that is the one way the hole invariant, as written, is false.

## 1. Hole invariant — `state.written` is not bytes on disk

`drainSegmentBody` does:

```js
await fh.write(buf, 0, buf.length, seg.start + state.written);
state.written += buf.length;
opts.onBytes?.(buf.length);
```

`filehandle.write` returns `{ bytesWritten }` and does **not** retry a short write (`writeFile` does; this API does not). `bytesWritten` is ignored. `state.written` then claims the whole chunk.

A short write (ENOSPC-with-a-partial-chunk, EINTR after some bytes, a Windows `WriteFile` that didn't take the whole buffer) does all of the following, which is exactly the silent-corruption shape this PR exists to make unrepresentable:

- The next `fh.write` starts at `seg.start + inflatedWritten`, skipping the unwritten gap — a hole **inside** this segment, not merely between segments.
- A retry of the same segment asks `Range: bytes=${seg.start + inflatedWritten}-…`, so it never fills the gap.
- Success checks (`written === length`, sum === total, `assertComplete` size) all pass: they read the counter, and `stat().size` does not see a hole.
- Failure cleanup `truncate(scratch, contiguousPrefix())` then `rename`s that into `targetPath`. `contiguousPrefix` is a sum of those counters. The published "resumable prefix" can itself contain a hole. A later 206 append splices onto it.

That is a yes to the question you asked: `contiguousPrefix` **can** over-report, not because a retry moves `written` backwards, but because `written` is allowed to get ahead of the file.

The single-connection path does not have this bug — `pipeline`/`createWriteStream` loops on short writes. The segmented path is now the only writer that treats "the Promise resolved" as "the chunk is in the file."

Fix: loop until the chunk is fully written, advance `state.written` (and `onBytes`) only by `bytesWritten`, and do not start the next chunk (or a retry Range) until that is true. A test that stubs `FileHandle.write` to return a short `bytesWritten` on the first call and checks both (a) the landed bytes have no gap and (b) `contiguousPrefix` matches `stat().size` would pin it. The current hole test cannot: the loopback server never produces a short `write(2)`.

Hard kill / `*.partial.seg`: agreed. Nothing in this repo resumes or cache-hits a `.seg`. `evictLruIfNeeded` can see it (non-dot filename, same as a live `.partial`); that is not a new hazard, and a leftover `.seg` is truncated on the next `open(scratch, "w")`. I would not block on LRU.

Rename-without-a-gate: the only two `rename(scratch, targetPath)` sites are the post-tally success path and the `contiguousPrefix` cleanup. There is no third.

## 2. Fallback boundary — holds on this SHA

`SegmentedRangeUnsupportedError` is constructed only in `probeSegmentedRanges`. `fetchSegmentOnce` raises `ModelError` (retryable) on a mid-run non-206 / bad Content-Range / empty body, and the comment is accurate: the original 200 is already `cancel()`ed, so this must not look like a fallback.

User abort during the probe fetch is rethrown, not wrapped, so `streamUrlToFile` does not pipe a dead body. Probe segment 1 (not 0) is the right lie-detector. `planSegments` declining resume / 206 / append / Range-on-request / content-coding / CL≠expectedTotal is the right eligibility cut — the #467 truncating-proxy case stays on the single-connection path that already has the understatement guard.

## 3. E/F/G surviving — keep them

They are three detectors of the same fault, not dead code. H (all three gone) is the mutation that matters, and the short-**middle**-segment fixture is the right shape (`assertComplete` cannot see a hole). I would not delete any of the three.

## 4. Cancel vs sibling abort — holds

- User cancel: `opts.signal.aborted` is checked before `classifyDownloadFailure`, so a cancel is never retried inside a segment. Cleanup publishes a prefix. `streamUrlToFile` plus the outer loop consult the **caller's** signal (not `attemptCtl`) before classifying, so a cancel does not become a completion. `deferErrorRow` keeps the tray from publishing an outcome.
- Sibling failure: `firstFailure` is latched, `abandon.abort()` stops **new** fetches/`abortableDelay`s, `allSettled` waits until in-flight drains finish (they key off `opts.signal`, not `abandon` — that is what makes the "don't truncate under live writers" claim true), then cleanup runs. The thrown error is the real one, not a sibling `AbortError`.
- Stall: `attemptCtl` is what `opts.signal` is, so a stall aborts the segmented attempt rather than retrying a single segment. The outer loop still retries because it uses the `stalled` flag and the user signal. Same shape as today's single-connection stall.

I did not find a path that mistakes a sibling abort for a user cancel and swallows the failure. There is no `downloadWithCache`-level cancel test that asserts the landed `.partial` is a hole-free prefix; the unit test on `runSegmentedDownload` covers the mechanism, not the tray/job contract. Worth adding, not a merge block.

## 5. Windows rename — holds

`createWriteStream` is constructed only on `if (!segmented)`, after the segmented attempt has either finished or fallen through. A segmented run never holds an fd on `targetPath`. The restart `writeFile(targetPath, "")` and the `truncate(targetPath, 0)` both close before the scratch rename. The stale-`.partial` test is the EPERM case. CI includes `windows-latest`; I am not treating a still-pending check as evidence either way.

## 28652b9 specifically (the ENOENT mapping)

Correct, and it is a real regression without it: `classifyDownloadFailure` defaults to not-retryable, so a bare `ENOENT` from the winner's rename would fail a download that succeeds today. Mapping both timings (`open` after the rename, and the loser's own `rename`) to a retryable `ModelError` is the right contract; the retry then cache-hits the winner.

The `fetchImpl` steal is deterministic for the **open-after-rename** timing. The success-path `rename` ENOENT is the same mapper, not a second test. Fine — one mapper, two call sites.

This does **not** make mixed writes from two live processes unrepresentable (both can `r+` the same `.seg` and the winner can finalize a blend). That is the same class as two processes streaming one `.partial` today; I am not asking you to solve cross-process writers here. Do not let that become a reason to drop the ENOENT mapping.

## Also, not merge-blocking

- On a retryable failure of the **probe** segment, attempt 1's `probe.response` body is not `cancel()`ed before attempt 2's new Range GET. Same class as the socket-pin you fixed in `2274722`. Destroy/cancel in `drainSegmentBody`'s `finally`.
- Tray `downloaded` is the sum of every segment write, not `contiguousPrefix`. A cancel at a "90%" row can resume from a much smaller prefix. The watchdog should keep seeing every byte (liveness); the row should probably show the durable prefix, or at least not imply it is durable.
- `truncate(opts.targetPath, 0).catch(() => {})` swallowing a failed truncate re-introduces the double-disk case the comment says it prevents, only when the file exists and cannot be truncated. Rare; `ENOENT` on a fresh target is why the catch is there. Narrow it to `ENOENT`.

The speedup, the "never the reason a download fails" claim, and the publish-by-rename design all hold. Pin the write counter to `bytesWritten` and this is an approve on the next SHA.

artokun:
**Correction, and a new head: `e4147fd`.** Do not bank an approve naming `37f18d1`, `2274722` or `28652b9` — the implementation changed materially since all three.

I published "1.87x" earlier. That number was not wrong so much as incomplete, and chasing it turned up a defect in my own change that I would have shipped:

**The first implementation was SLOWER than a single connection whenever the source could outrun the disk.** Against an out-of-process loopback origin it ran **0.63x**. This issue has already burned one pass on a plausible-looking fix that measured slower, so I went and measured instead of reasoning:

- Client-and-server-in-one-process looked like the cause. Moving the origin to its own process: still 0.6x. Not the harness.
- Writer contention looked like the cause. It was flat across connection counts (0.74x at 2, 0.79x at 4, 0.88x at 8). Not parallelism.
- Rebuilding the drain on `pipeline(…, filehandle.createWriteStream({start}))` — the same machinery the plain path uses — gave 0.70x. No better.

It was the per-chunk positional write. Coalescing into 1 MiB `writev` calls fixes it; the batch sweep is in the PR body (64 KB 0.63x → 1 MiB 0.84x → 4 MiB 0.52x). Not a highWaterMark, no new pipeline stage — #1738/#1746 stay closed.

**Corrected numbers, all through `downloadWithCache`, medians:**

| regime | single | segmented | ratio |
|---|---|---|---|
| real HF 2.13 GB, 3 rounds | 116.3 MB/s | 181.7 MB/s (x4) | **1.56x** |
| 8 MB/s per-connection cap, 2 rounds | 5.2 MB/s | 20.6 / 40.7 MB/s (x4 / x8) | **3.97x / 7.85x** |
| loopback uncapped ~570 MB/s, 5 rounds | 570.3 MB/s | 473.4 MB/s (x4) | **0.83x** |

The last row is the one I want you to push on. On a source fast enough to saturate the local disk this is still ~17% slower than one connection, and I could not close that. I judged it acceptable because 570 MB/s is 4.5 Gbps — no model host serves that — while the regime the issue is actually about goes 4-8x. **If you disagree, the alternative is rate-adaptive escalation** (stream single-connection, measure, cut over mid-transfer), which is real complexity on a P1 download path under a freeze. Tell me if that trade is wrong.

**Also worth your attention:** mutation A briefly read as SURVIVED. The mutation produced code that did not type-check, so vitest ran zero tests and zero failures scored as "survived" — a false negative in my own evidence. The harness now fails loudly on `no tests`, and A was rewritten so the file still compiles. Every row in the mutation table is from the fixed harness. If any other row looks too convenient, that is the failure mode to suspect.

Two unchanged asks from my earlier comment still stand as the riskiest parts: the **hole invariant** (`.seg` reaches `.partial` only by rename, past one of two gates) and the **fallback boundary** (`SegmentedRangeUnsupportedError` must be unreachable once `res.body` is cancelled). Note the batching interacts with the first: buffered-but-unflushed bytes must never advance `state.written`, or the "contiguous prefix" would claim bytes the file does not hold. That is why the flush advances it and the arrival counter does not.

artokun:
@grok Thank you — finding 1 was real, and it was the exact class this PR claims to make unrepresentable. Fixed. **Please re-review `b70fd3bbdfd18dfcd616ed5a983e72711494a523`.**

Note it also applies to the SHA you did *not* see: between your review and now I replaced the per-chunk `fh.write` with a batched `fh.writev` (a separate perf fix — the per-chunk version measured **0.63x** a single connection on an uncapped source, so segmentation was a deficit whenever the source outran the disk). `writev` returns `bytesWritten` and does not retry either, so your finding landed on the new code with full force — and worse, since a 1 MiB batch is a bigger thing to write short.

**Fix (`dfca38e`):** `flush()` now loops until the batch is fully written, advances `state.written` **only** by the returned `bytesWritten`, and resumes the buffer list at the exact byte via a new `dropWritten()` helper. A `bytesWritten <= 0` is a retryable `ModelError` rather than an infinite loop.

**Pinned by a test, exactly as you specified.** `src/__tests__/services/download-segments-shortwrite.test.ts` mocks `node:fs/promises` `open` so every `writev` takes at most 40,961 bytes — deliberately not a multiple of a chunk or the batch, so every single write is short and every buffer boundary is crossed mid-buffer. A 6 MiB file is then downloaded through `probeSegmentedRanges` + `runSegmentedDownload` and required to be byte-identical. Mutating the fix back (advance by the requested count, ignore `bytesWritten`) fails that test and nothing else — so it is the test carrying the claim, not a neighbour.

Also from your review:
- **Probe-segment retry pinned the socket** — `drainSegmentBody` now destroys the Node readable in its `finally`, so every exit releases the body, not just a fully-drained one.
- **`truncate(targetPath, 0).catch(() => {})` too broad** — narrowed to `ENOENT`; anything else now throws instead of silently re-introducing the double-disk case.
- **No `downloadWithCache`-level cancel test** — added (`b70fd3b`): aborts mid-transfer and asserts no destination file, no surviving scratch, and a staged `.partial` that is a strict hole-free prefix of the real object. Green on 3 consecutive runs.

Two of your points I am answering rather than changing, and I would rather you push back than let them pass:

1. **Tray `downloaded` is ahead of the durable prefix.** True, and I left it: the single-connection path has the same property — its counter sits in a `Transform` ahead of the write stream's own buffer, so a "90%" row there is not durable either. Making only the segmented path report the durable prefix would make the two shapes disagree about what a row means, which is the divergence the shared `noteBytes` sink exists to prevent. If the row should mean "durable", that is a change to both paths and a separate issue.
2. **Mixed writes from two live processes.** Agreed and not attempted. Same class as two processes streaming one `.partial` today; the ENOENT mapping stays.

Full suite at the fix head: **607 files, 11124 passed, 0 failed.** Throughput unchanged by the write loop (loopback 0.82x vs 0.83x before; real HF x4 **1.84x** on the latest run) and `distinct sha256 across all arms: 1` in both.

artokun:
Head is now `fdf6859a128519ace9e9d2c2d84d419bee615b81` — comment-only change on top of `b70fd3b` (the call-site comment still quoted the very first raw A/B, "1.5-1.7x", instead of the final measured medians; a stale number in a comment outlives the measurement that produced it). No code change, so the `b70fd3b` evidence stands as-is: 607 files / 11124 passed / 0 failed, and the short-write test still the only thing that fails when the `bytesWritten` fix is reverted.

@grok please review `fdf6859`.

artokun:
**Head `4ed6dcb734a2a97c7174226d0afb5f9a859d5c1c`, and I found a real behavioural regression in my own change. It is a trade-off, not a bug, but it is a user-visible one and I want it decided rather than buried.**

How it surfaced is worth stating, because the first version of the finding was a test lying to me. Three of my assertions re-derived the staged file as `cachePathForUrl(url) + '.partial'`. The writer actually **dot-prefixes the basename** — `.{hash}{ext}.partial`. So every "the scratch is gone" / "the partial holds a prefix" check was `stat`ing a path that never exists, and passed whether or not the thing it names was cleaned up. Proof they had no teeth: I mutated the success path to deliberately leave a scratch file behind and **all 21 tests still passed**. They now use the exported `stagedPartialPathForUrl()`, and the cancel test reads the staged file with no `.catch()`, so a wrong path fails instead of quietly reading as "zero bytes kept".

With the correct path, the cancel test failed — and the failure is real. Measured on a paced loopback rig, aborting mid-transfer, bytes left on the staged `.partial` file:

| abort at | 1 connection | 4 connections |
|---|---|---|
| 200 ms | 1.75 MB | **nothing** |
| 300 ms | 2.50 MB | **nothing** |
| 400 ms | 3.25 MB | **nothing** |
| 600 ms | 4.81 MB | **1.00 MB** |

Two causes, both inherent to segmenting rather than fixable by tidying:

1. **Write batching delays durability.** Bytes land in 1 MiB batches per segment (the batch that took the throughput from 0.63x to 0.84x), so before the first flush there is genuinely nothing on disk.
2. **Only the CONTIGUOUS prefix may be published.** Even at 600 ms, when all four segments hold ~1 MiB each, the publishable prefix is segment 0's alone. This one is not a bug — it is exactly the rule that makes a hole impossible — but it means a cancel keeps roughly 1/N of what was transferred.

So a user who cancels a 20 GB download at 50% resumes near 12.5%, not 50%. Correctness is unaffected (what is kept is always a valid hole-free prefix, and the retry re-fetches the rest at segmented speed) — the cost is bandwidth and the surprise.

**I did not try to fix it, and I want that challenged.** The real fix is what aria2 does: a control file recording per-segment progress so a resume can pick up all N segments. That means a new on-disk resume format next to the #343/#467 `.etag` machinery, which is a much bigger change than the owner asked for and exactly the kind of thing a stabilization freeze should keep out. The alternatives I rejected: shrinking the write batch (costs the throughput the batch bought) and dropping batching entirely (0.63x — worse than not doing this at all).

@grok please review `4ed6dcb`, and specifically: (a) is the trade-off above acceptable as shipped, or does it need the control file before this can land; (b) given three of my assertions were vacuous until now, which of the remaining ones would you not believe without seeing them fail?

Issue #404, opened 2026-07-29
get_node_info reports no match for registered DetectorForNSFW


artokun:
### What happened / root cause
`get_node_info(node_type="DetectorForNSFW")` returned `No nodes found`, but the live ComfyUI `/object_info` response contains the exact `DetectorForNSFW` key. The lookup result is stale or filtering the wrong object-info snapshot.

### Steps to reproduce
1. Install and restart with `comfyui-utils-nodes` 1.4.2.
2. Confirm `/object_info` contains `DetectorForNSFW`.
3. Call `get_node_info` with `node_type: DetectorForNSFW`.

### Exact error
No nodes found matching "DetectorForNSFW"

### Fix
applied locally: no  upstream-only: yes
Refresh from the live `/object_info` endpoint or invalidate the node-info cache after custom-node installation/restart.

### Environment
Windows 11 / ComfyUI 0.29.0 / RTX PRO 4000 Blackwell 24 GB / comfyui-mcp 0.48.6 / panel 0.11.3

artokun:
I recommend renaming the issue to invalidade node-info cache after custom-node installation/restart

artokun:
Fixed in **mcp 0.48.8**: get_node_info honors the live /object_info registration key (#432). Please update (`npm i -g comfyui-mcp@latest`) and reopen if it persists.

PR #1957, opened 2026-08-21
fix(panel#1554): a restart confirmation the user gave late is claimed by the next attempt, not discarded

artokun:
Fixes the underlying cause reported in artokun/comfyui-mcp-panel#1554 (P2, `via-panel`).
**Review is PENDING** — see "What I want reviewed" at the bottom. Nothing here has been gated.

## The report

> `panel_restart_comfyui` displayed a restart-confirmation flow but the tool waited 90 seconds and returned "No confirmation received within 90s, so I did NOT restart ComfyUI." … Call the catalog `restart_comfyui` action directly; restart succeeds in about 16 seconds.

The reporter's own guess — "the confirmation-card response was not correlated back to the pending panel tool call" — is **wrong in its literal form**: the card carries a `randomUUID()` `ask_id`, the bridge correlates on exact id equality, and the whole #486 journal is built on that. Nothing is mis-correlated.

The second half of their ask is the real defect: *"If the confirmation was already accepted, do not lose it."*

## Mechanism

`panel_restart_comfyui` bounds its confirmation card at 90s (#404) so an undelivered card fails fast instead of reading as a four-minute hang. What #404 did not account for is what happens to the **answer**:

1. **The card stays live.** The panel retires an interactive card only when the connection that painted it is REPLACED (#952) — it does nothing with the command's own `timeout_ms` (verified: `web/js/comfyui-mcp-panel.js` reads `msg.timeout_ms` in exactly one place, the #517 duplicate-reply await). So the card is still painted and still clickable after the orchestrator has given up.
2. **A late click still reaches the bridge**, which buffers the validated answer under its `ask_id` for `LATE_ASK_TTL_MS`. This is not speculation — `src/__tests__/services/ui-bridge.test.ts:4425` already drives the real `UiBridge` over a real socket and proves a timed-out `ask_user` card's late answer is retrievable via `takeLateAskReply`.
3. **Nothing ever reads it back for a CONFIRM card.** #486's durable journal accepts only `pa-`-prefixed ids (`PANEL_ASK_ID_PREFIX`) — a confirm card is deliberately excluded. And `confirm()`'s own grace poll is **zero-width** here: `timeoutMs` (90s) is the hard ceiling on *deadline + grace*, so `deadlineMs = 90s` and `graceMs = 90s − 90s = 0`. #360's late-honour poll takes one look and gives up.

So the answer is TTL-pruned unread while the tool reports *"No confirmation received within 90s, so I did NOT restart ComfyUI"* — to a user who clicked **"Yes, go ahead"** and watched the card collapse to **"✓ Yes, go ahead"**. They then went and found the headless tool for a restart they had already authorised. That is the whole of #1554.

This is a **regression by omission**: #404 (bound the wait) silently disabled the mechanism #360 installed (honour a late answer), for the one caller that passes a tighter budget.

## The fix — read the answer, don't wait longer

**#404's 90s is untouched.** No path waits any longer than it does today.

- On timeout, the abandoned card's `ask_id` is remembered against `(tabId, header, question)`.
- The **next** attempt at that same confirmation drains it **before** dispatching a second card — because painting a second card over an answered one is the state the orchestrator's own #952 wording has to apologise for ("the user may see two — tell them which one to answer").
- A late **"No, cancel"** is honoured identically. That decline is lost today too, and re-asking a question the user already answered "no" to is its own defect.

Attribution rules, each load-bearing:

| Rule | Why |
|---|---|
| Exact `ask_id` equality | never "the most recent answer" |
| Same tab **and** identical question **and** header | an answer to the restart card can only satisfy the restart card |
| One-shot (entry dropped on the first look, answered or not) | a recovered answer satisfies exactly one call; no stale key lingers |
| Staleness bound = the bridge's own `LATE_ASK_TTL_MS` | past it `takeLateAskReply` has already dropped the answer. `LATE_ASK_TTL_MS` is now **exported** rather than copied, so the second reader cannot drift from it |
| Route-key **takeover** retires it | `wf:` keys recur; the newcomer never answered that card. Rides #486's existing takeover listener (a *setter*, so a second `setTabTakenOverListener` would have silently replaced #486's). A **reconnect** is deliberately not this event — same browser tab, same incarnation — which is what makes the reporter's "across agent/panel reconnect" case work |
| Opt-in per call site | only `panel_restart_comfyui`. `panel_clear` wipes the canvas and gets the full 285s wait, so its loss window is far narrower and the cost of resurrecting a stale yes far higher. It stays out until someone has a reported failure to point at |
| Disclosed, never silent | acting on an earlier card's answer must not read as a fresh confirmation |

`withReplyNote` exists because the disclosure has to ride **all ~25** of the handler's return points. A disclosure that only survives the happy path reads as a fresh confirmation on every other branch, and threading a note through 25 returns is how one gets missed. It holds the note in a **per-call local**, not on `ctx` — `ctx` is built once per panel MCP server and shared by every tool call it serves, so parking it there would make mis-attribution possible for exactly the class of statement that must never be mis-attributed.

The timeout wording also changes. "Tell me to restart it and I'll re-ask" described a card that was finished and a decision that had to be made again. It was neither — and believing it is what sent the reporter looking for the headless tool.

## Evidence

**17 new tests**, driving the **real** `ctx.confirm` (an earlier draft stubbed it and proved nothing — every stub returns what it was told to, so the line that decides whether a buffered answer is claimed never ran).

**Mutation-verified — each mutation applied to the live tree, run, then reverted:**

| # | Mutation | Result |
|---|---|---|
| M1 | recovery lookup made unreachable | RED: YES-honoured, late-DECLINE, ONE-SHOT |
| M2 | timeout stops remembering the card | RED: same three |
| M3 | **call site** opts out (`recoverAbandonedAnswer: false`) | RED: **only** the call-site test — the other 13 stay green, which is the point |
| M4 | `withReplyNote` drops the note | RED: the disclosure test |
| M5 | `appendReplyNote` prepends instead of appends | RED: the structured-reply test |
| M6 | recovery key forgets the tab | RED: the cross-tab test |
| M7 | takeover forgets nothing | RED: the takeover test |
| M8 | the listener drops `forgetAbandonedConfirmCards` | RED: the wiring test |
| M9 | the listener drops `AskAnswers.closeAsks` | RED: **#486's own** boundary test — proving the assertion I relaxed there still bites |

**Full suite:** 606 files / 11129 tests green. One earlier run reported a single failure whose name I did not capture; it did not reproduce across three subsequent full runs. I am flagging that rather than claiming a clean sweep I only observed three times out of four.

**A false green I caught and am reporting:** my first attempt at M4–M6 ran the mutation through `child_process.execSync`, where MSYS does not translate `/tmp`, so the mutations silently never applied and vitest printed "14 passed". Re-run directly from the shell they all went red as tabled. Separately, a `git checkout --` revert between mutations ate an **uncommitted** part of the fix and produced a red that had nothing to do with the mutation; everything above was re-run against a committed tree.

## Browser verification

**No panel/web code changed** — this is entirely orchestrator-side (`src/orchestrator/`, `src/services/ui-bridge.ts`). The panel's Playwright suite fakes the orchestrator (MockBridge), so it cannot exercise any of this; I did not run it and am not implying coverage I do not have.

The **panel-side half of the story is unchanged by design**: the card staying live past the orchestrator's timeout is precisely what this fix depends on, and it is what makes the answer available to claim.

- **Did not extend any wait.** #404's 90s stands. A grace carved out of the 90s would have been near-worthless (an answer inside the window already resolves `bridge.send`), and a grace added on top would have re-created the hang #404 was filed about.
- **Did not auto-fall-back to the headless restart**, which the report asks for ("deterministic fallback instead of forcing discovery of a separate headless catalog tool"). That would restart ComfyUI with no confirmation at all, and #851 is the record of why an unconditional `restart_comfyui` recommendation is itself a defect. The conditional advice from #851 is unchanged.
- **Did not journal confirm answers into `AskAnswers`.** That store is `panel_ask`-only for stated reasons and its recovery matches on question *fingerprint*; a confirm needs the stricter card-id identity, which is what this uses.
- **Did not enable recovery for `panel_clear`.**
- **Did not "fix" the zero-width grace** in `confirm()`. It is now not load-bearing, and changing it moves wall-clock behaviour #404 deliberately bounded.

## What I want reviewed (the load-bearing claims, and where I am least sure)

1. **Is resurrecting a ≤5-minute-old "Yes" acceptable for a RESTART?** This is the judgement call of the whole PR. My argument: same tab, byte-identical question, exact card id, one-shot, disclosed in the reply, bounded by an existing TTL, and the alternative is discarding a decision the user explicitly made and watched the panel accept. But a restart aborts in-flight renders, and if you think a destructive gate must never accept an answer from an earlier card, this design is wrong and I would restrict it to *reporting* the recovered answer instead of acting on it.
2. **Is the takeover boundary the right one, and is it sufficient?** I chose it because a reconnect must *not* retire (that is the reporter's case) while a different browser tab on a recurring `wf:` key must. If there is a third state — same browser tab, genuinely new conversation (New chat) — that should also retire, I have missed it. `AskAnswers.closeAsks` is called from four other places in `index.ts` that I did **not** join.
3. **Relaxing #486's assertion.** I changed an exact-source-shape regex on the takeover listener to an anchored body check plus an "exactly one listener installed" assertion. M9 shows it still fails when `closeAsks` leaves the body — but this is someone else's guard and I loosened it, so it deserves an independent look.
4. **`withReplyNote` changes the handler's signature at the `def()` call site.** It is the only tool wrapped this way. Is a per-call note channel the right shape, or should the disclosure be threaded explicitly?
5. **Is the reporter's actual failure the one I fixed?** I could not reproduce their session. The chain — the card outlives the timeout, the bridge buffers the answer, nobody reads it — is established by code and by an existing real-socket bridge test, but whether *their* click landed after 90s or they simply never clicked is not something the report settles. If they never clicked, this PR fixes a real adjacent defect and their session would still have timed out.

### DIFF
--- a/src/__tests__/orchestrator/ask-answer-journal.test.ts
+++ b/src/__tests__/orchestrator/ask-answer-journal.test.ts
@@ -1029,7 +1029,18 @@ describe("ask-answer journal — bounds may LABEL, never silently lose (#486)", 
     // A different browser tab taking over a recurring key is a conversation
     // boundary too — the ENTRIES are tab-keyed, so the newcomer must be shut out
     // of them at the hello, not merely kept from the debt.
-    expect(src).toMatch(/setTabTakenOverListener\(\(tabId\) => AskAnswers\.closeAsks\(tabId\)\)/);
+    // Asserted on the listener's BODY, not on a one-expression shape. panel#1554 gave
+    // this same boundary a second retirement (an abandoned confirmation card must not
+    // be claimable by the newcomer either) and the listener became a block. The
+    // guarantee under test is "the takeover retires ask state", and a regex that only
+    // matches one statement would fail the moment anything legitimately joins it —
+    // pressure to install a SECOND setTabTakenOverListener instead, which (it is a
+    // setter, not an adder) would silently replace this one and un-ship #486.
+    const takeoverAt = src.indexOf("bridge.setTabTakenOverListener(");
+    expect(takeoverAt, "takeover listener not found").toBeGreaterThan(-1);
+    expect(src.slice(takeoverAt, takeoverAt + 300)).toContain("AskAnswers.closeAsks(tabId)");
+    // …and exactly one is installed, which is what makes the body the whole story.
+    expect(src.split("bridge.setTabTakenOverListener(").length - 1).toBe(1);
     expect(src).toContain("AskAnswers.setIncarnationResolver(");
     // The ack must carry WHO is acking, or a switched provider can certify the
     // previous conversation's answer.
--- a/src/__tests__/orchestrator/confirm-late-answer-recovery.test.ts
+++ b/src/__tests__/orchestrator/confirm-late-answer-recovery.test.ts
@@ -35,6 +35,7 @@ import {
   forgetAbandonedConfirmCards,
   makePanelToolCtx,
   resetAbandonedConfirmCards,
+  __panelToolsTestHooks,
   type ConfirmOptions,
   type PanelToolCtx,
   type ToolResult,
@@ -338,6 +339,48 @@ describe("panel#1554 the recovered decision is disclosed, not slipped in", () =>
 
     expect(out).toContain("That card is still live");
     expect(out).not.toContain("and I'll re-ask");
+  });
+
+  it("END TO END: two real tool calls, real confirm, and the second one claims the click", async () => {
+    // Everything above either drives `confirm` directly or stubs it. This drives NEITHER:
+    // the REAL handler, twice, through the REAL confirm, with the recovery reached the
+    // way production reaches it. Without it the chain call site → opts → confirm →
+    // recovery is only proven by composition.
+    //
+    // The recovered answer is the DECLINE, because it is the one post-confirm branch
+    // that settles deterministically without dispatching a reboot or waiting out a
+    // readiness budget — the probe is stubbed healthy and its recheck window shrunk.
+    __panelToolsTestHooks.setHealthProbe(async () => "healthy");
+    __panelToolsTestHooks.setDeclineProbeTiming({
+      windowMs: 20,
+      intervalMs: 5,
+      probeTimeoutMs: 5,
+    });
+    try {
+      const h = harness();
+
+      // Call 1: the card is dispatched and nobody answers inside the 90s budget.
+      const first = textOf(await restartDef().handler({} as never, h.ctx));
+      expect(first).toContain("No confirmation received within");
+      expect(h.cards).toHaveLength(1);
+
+      // The user answers it afterwards. The bridge buffers it under the card's own id.
+      h.late.set(String(h.cards[0].ask_id), "No, cancel");
+
+      // Call 2: no second card, and the reply carries the disclosure.
+      const second = textOf(await restartDef().handler({} as never, h.ctx));
+
+      expect(h.cards).toHaveLength(1);
+      expect(second).toContain("NO NEW CONFIRMATION CARD WAS SHOWN");
+      expect(second).toContain("No, cancel");
+      // …and it is the DECLINE branch's own reply, not the timeout's.
+      expect(second).not.toContain("No confirmation received within");
+      // Nothing was restarted on a decline — the safe path is untouched.
+      expect(h.cards.every((c) => c.cmd === "ask_user")).toBe(true);
+    } finally {
+      __panelToolsTestHooks.setHealthProbe(null);
+      __panelToolsTestHooks.setDeclineProbeTiming(null);
+    }
   });
 
   it("the note is its OWN block: a structured reply stays parseable at index 0", () => {
--- a/src/__tests__/orchestrator/confirm-late-answer-recovery.test.ts
+++ b/src/__tests__/orchestrator/confirm-late-answer-recovery.test.ts
@@ -24,12 +24,15 @@
 // nothing: every stub returns whatever it was told to, so the one line that decides
 // whether a buffered answer is claimed was never executed.
 
+import { readFileSync } from "node:fs";
+
 import { beforeEach, describe, expect, it } from "vitest";
 
 import {
   appendReplyNote,
   buildPanelToolDefs,
   confirmAnswerRecoveryWindowMs,
+  forgetAbandonedConfirmCards,
   makePanelToolCtx,
   resetAbandonedConfirmCards,
   type ConfirmOptions,
@@ -215,6 +218,50 @@ describe("panel#1554 a late confirmation is claimed, not discarded", () => {
     const h = harness(TAB, { answerCards: true });
     expect(await h.ctx.confirm(QUESTION, HEADER, 500, RECOVER)).toBe("yes");
     expect(h.drained).toHaveLength(0);
+  });
+
+  it("a TAKEOVER of the route key drops it — the newcomer never answered that card", async () => {
+    // `wf:` keys recur: close the browser tab, open the same workflow elsewhere, and the
+    // newcomer inherits the address. #486 calls that a conversation boundary and retires
+    // the departing occupant's asks on it; a confirmation card is no different.
+    const h = harness();
+    const askId = await abandonOnce(h);
+    h.late.set(askId, "Yes, go ahead");
+
+    forgetAbandonedConfirmCards(TAB);
+
+    expect(await h.ctx.confirm(QUESTION, HEADER, 500, RECOVER)).toBe("timeout");
+    expect(h.cards).toHaveLength(2);
+    // The departed occupant's answer is left where it was — not served to the newcomer.
+    // (Asserted on the BUFFER, not on `drained`: confirm's own zero-width grace poll
+    // looks that id up once on the way to reporting the timeout, so a "never looked"
+    // assertion would be false for a reason that has nothing to do with this rule.)
+    expect(h.late.get(askId)).toBe("Yes, go ahead");
+  });
+
+  it("a takeover of ANOTHER key leaves this one alone", async () => {
+    const h = harness();
+    const askId = await abandonOnce(h);
+    h.late.set(askId, "Yes, go ahead");
+
+    forgetAbandonedConfirmCards(OTHER_TAB);
+
+    expect(await h.ctx.confirm(QUESTION, HEADER, 500, RECOVER)).toBe("yes");
+  });
+
+  it("WIRING: the takeover listener retires confirmations as well as asks", () => {
+    // The bridge takes ONE takeover listener (a setter, not an adder), so this fix had
+    // to join the existing call rather than add a second one — a second
+    // setTabTakenOverListener would have silently replaced #486's and un-shipped it.
+    // Behavioural tests above cannot see that; this is the shape of the call site.
+    const src = readFileSync("src/orchestrator/index.ts", "utf8");
+    const at = src.indexOf("bridge.setTabTakenOverListener(");
+    expect(at).toBeGreaterThan(-1);
+    const call = src.slice(at, at + 200);
+    expect(call).toContain("AskAnswers.closeAsks(tabId)");
+    expect(call).toContain("forgetAbandonedConfirmCards(tabId)");
+    // Exactly one listener is installed — the whole hazard.
+    expect(src.split("bridge.setTabTakenOverListener(").length - 1).toBe(1);
   });
 
   it("the staleness bound IS the bridge's buffer TTL, not a second copy of it", () => {
--- a/src/__tests__/orchestrator/confirm-late-answer-recovery.test.ts
+++ b/src/__tests__/orchestrator/confirm-late-answer-recovery.test.ts
@@ -1,0 +1,310 @@
+// panel#1554 — the reporter's restart confirmation was ACCEPTED by the panel and
+// DISCARDED by the orchestrator.
+//
+// `panel_restart_comfyui` bounds its confirmation card at 90s (#404) so an undelivered
+// card fails fast instead of reading as a 4-minute hang. What #404 did not account for
+// is the ANSWER. The panel retires a card only when the connection that painted it is
+// REPLACED (#952) — never on a caller's timeout — so the card stays painted and stays
+// clickable. A click after the deadline still reaches the bridge, which buffers the
+// validated answer under its `ask_id` for LATE_ASK_TTL_MS.
+//
+// Nothing ever read it back. #486's durable journal accepts only `pa-`-prefixed
+// panel_ask ids, and confirm's own grace poll is ZERO-WIDTH here: `timeoutMs` 90s is
+// spent entirely on the card deadline, so `graceMs` computes to 0 and #360's
+// late-honour poll takes one look and gives up. The answer was TTL-pruned unread while
+// the tool reported "No confirmation received within 90s, so I did NOT restart
+// ComfyUI" — to a user who had clicked "Yes, go ahead" and watched the card collapse to
+// "✓ Yes, go ahead". They went and found the headless `restart_comfyui` instead.
+//
+// THE FIX READS IT rather than waiting longer, so #404's 90s is untouched: the
+// abandoned card's id is remembered, and the NEXT attempt at the same confirmation
+// drains it before dispatching a second card.
+//
+// These tests drive the REAL `ctx.confirm`. An earlier draft stubbed it and proved
+// nothing: every stub returns whatever it was told to, so the one line that decides
+// whether a buffered answer is claimed was never executed.
+
+import { beforeEach, describe, expect, it } from "vitest";
+
+import {
+  appendReplyNote,
+  buildPanelToolDefs,
+  confirmAnswerRecoveryWindowMs,
+  makePanelToolCtx,
+  resetAbandonedConfirmCards,
+  type ConfirmOptions,
+  type PanelToolCtx,
+  type ToolResult,
+} from "../../orchestrator/panel-tools.js";
+import { LATE_ASK_TTL_MS } from "../../services/ui-bridge.js";
+import { getBootLocalComfyUIBaseUrl } from "../../config.js";
+import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+
+const TAB = "wf:workflows/a.json";
+const OTHER_TAB = "wf:workflows/b.json";
+const BOOT_BASE = (getBootLocalComfyUIBaseUrl() ?? "http://127.0.0.1:8188").replace(/\/+$/, "");
+
+const QUESTION =
+  "Restart ComfyUI now? It (and this agent) will go down briefly, then reconnect and resume automatically.";
+const HEADER = "Restart ComfyUI";
+
+/** The bridge's canonical no-reply text — the ONLY shape isReplyTimeoutError accepts
+ *  from an untagged error, so a fixture that gets it wrong would be surfaced as a real
+ *  error ("unreachable") and would silently stop testing the timeout branch. */
+const replyTimeout = (tabId: string) =>
+  new Error(
+    `Panel tab ${tabId.slice(0, 8)} did not reply to "ask_user" within 500 ms — ` +
+      "the ComfyUI tab may be backgrounded or frozen",
+  );
+
+interface Harness {
+  ctx: PanelToolCtx;
+  /** ask_user frames the orchestrator actually put on the wire. */
+  cards: Array<Record<string, unknown>>;
+  /** The bridge's late-answer buffer (ask_id → what the user clicked). */
+  late: Map<string, unknown>;
+  /** Every takeLateAskReply(id) this run made — proves WHICH id was claimed. */
+  drained: string[];
+}
+
+function harness(tabId = TAB, opts: { answerCards?: boolean } = {}): Harness {
+  const cards: Array<Record<string, unknown>> = [];
+  const late = new Map<string, unknown>();
+  const drained: string[] = [];
+  const bridge = {
+    send: async (cmd: Record<string, unknown>) => {
+      if (cmd.cmd === "ask_user") {
+        cards.push(cmd);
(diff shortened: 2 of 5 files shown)


artokun:
@grok please review this one — and specifically attack these, in this order. I am not asking for a diff read; I am telling you where I think the load is.

**1. The judgement call, which is the whole PR.** `panel_restart_comfyui` will now act on a **"Yes, go ahead" the user clicked on an EARLIER card**, up to `LATE_ASK_TTL_MS` (5 min) later, without showing a new card. My defence is that it is the same tab, byte-identical question and header, exact `ask_id`, one-shot, disclosed in the reply, and bounded by a TTL that already existed. The counter-argument I cannot dismiss: a restart **aborts in-flight renders**, and a destructive gate arguably must never accept consent minted at an earlier moment. If you think that, say so — the fallback design is to REPORT the recovered answer and still require a fresh card, and I would rather ship that than be wrong in this direction. Note the busy-queue guard still runs afterwards either way.

**2. The boundary I chose, and the four I did not join.** I hung the retirement on #486's tab-TAKEOVER listener because a reconnect must **not** retire (that is literally the reporter's ask) while a different browser tab inheriting a recurring `wf:` key must. `AskAnswers.closeAsks` is also called from **four other sites** in `src/orchestrator/index.ts` (rewind, two provider-switch paths, and the #884 sweep) that I deliberately left alone. Is a New chat / rewind on the SAME browser tab a state where a remembered restart confirmation should also die? If yes, this leaks consent across a conversation boundary and I have missed it.

**3. I loosened somebody else's guard.** `src/__tests__/orchestrator/ask-answer-journal.test.ts` pinned the takeover listener's exact one-expression source shape; my change made the listener a block, so I relaxed it to an anchored body check plus "exactly one `setTabTakenOverListener` is installed". Mutation M9 (dropping `AskAnswers.closeAsks` from the body) still turns it red, so I believe it still bites — but loosening an assertion to get green is exactly the move that silently un-ships a fix, so please check I did not.

**4. Concurrency on `withReplyNote`.** I put the disclosure note in a per-call local rather than on `ctx` **because `ctx` is built once per panel MCP server and shared by every tool call it serves** (`createPanelMcpServer` → `makePanelToolCtx`). If you can find a path where two `panel_restart_comfyui` calls share one ctx and the note could still land on the wrong reply, that is a real defect. I also chose to APPEND the note as its own content block rather than prepend, so `content[0]` stays the parseable JSON document — check nothing downstream assumes a single-block reply from this tool.

**5. The premise itself.** I claim the panel does **nothing** with an interactive command's `timeout_ms` (so the card outlives the orchestrator's 90s and stays clickable), and that the bridge buffers the late answer under a **non-`pa-`** ask id. I verified the first by reading the panel pack (one `msg.timeout_ms` reference, in the #517 duplicate-reply await) and the second from an existing real-socket test at `src/__tests__/services/ui-bridge.test.ts:4425`. If either premise is wrong, the mechanism this PR adds is inert and the tests only prove my fixture.

Least certain of all: **whether this is the reporter's actual failure.** I could not reproduce their session. If they simply never clicked within 90s, this fixes a real adjacent defect and their run would still have timed out — I say so in the PR body rather than claiming the repro.

artokun:
Rebased onto `main` (`a195af8`, +7 commits including the sibling fixes for panel#1555/#1556/#1557 from the same reporter's session). No conflicts; diff is back to 5 files. Re-ran the full suite on the new base: **609 files / 11172 tests green**.

Also reworded the disclosure note — it was addressed to "you (the user)" when its only reader is the agent. It now states what happened (no new card was shown; an earlier card's answer was claimed, with its age) and tells the agent to relay that. Mutation M4 (`withReplyNote` drops the note) re-verified red against the new wording.

Review request above still stands — points 1 and 2 are the ones I most want attacked.

artokun:
Two updates, then this is settled from my side pending review.

**1. Closed the composition gap.** Every test I had either drove `ctx.confirm` directly or stubbed it, so "call site → opts → confirm → recovery" was only proven by composition. Added an end-to-end test that drives the **real handler twice through the real confirm**: call 1 dispatches a card and returns the 90s timeout, the user answers it afterwards, call 2 claims that answer with **no second card** and carries the disclosure. It uses the DECLINE branch because that is the one post-confirm path that settles deterministically without dispatching a reboot or waiting out a readiness budget (health probe stubbed healthy, recheck window shrunk via the existing `__panelToolsTestHooks`).

Re-ran mutation **M3** (call site flips to `recoverAbandonedAnswer: false`) against it: **two** tests now go red, the call-site one and the end-to-end one. Previously only the call-site spy caught it.

**2. Rebased again** onto `main` (it moved twice during this run). Full suite on the current base: **609 files / 11173 tests green**. Diff is 5 files.

Also dropped a dead `logger` import an earlier revert had left behind.

Still **unreviewed** — @grok the five points above are unchanged, and 1 (is it acceptable to act on a ≤5-min-old "Yes" for a restart?) and 2 (is tab-takeover the right retirement boundary, or should New chat / rewind retire it too?) are the ones where a second opinion actually changes what ships. I am not merging this on my own tests.

artokun:
Adding my own reasoning on **open question 2** (which boundary retires an abandoned confirmation), so it can be argued with rather than left as a bare question.

`AskAnswers.closeAsks` is called from six places in `src/orchestrator/index.ts`: the tab-takeover listener, a rewind, two provider-switch paths, and the New chat / durable-clear sweeps. I joined **only the takeover**. The argument for that, stated so it can be refuted:

- #486 retires asks at a conversation boundary because an **answer is content** — quoting the previous occupant's pick into a conversation that never asked the question is a misattribution.
- A restart confirmation is not content. It is a **boolean gate on a machine-level action**, and nothing of the user's is quoted anywhere. Starting a New chat, rewinding, or switching provider does not mean "I withdraw my consent to restart ComfyUI"; the same person on the same tab authorised the same server restart minutes ago, and the reply says so with the age.
- The takeover is different in kind: it is the one event where the browser tab behind a recurring `wf:` key is genuinely **a different tab**, so "the same person, same surface" stops being true. That is why it is the boundary I joined.

Where I could be wrong: if you hold that a destructive gate's consent must not outlive *any* conversation boundary regardless of what it authorises, then all six sites should retire it, and the extra cost is one more confirmation card. I did not do that because it means touching five guarded call sites (two of which the #486 test pins by their exact `if (…) AskAnswers.closeAsks(panelTab)` shape), and because I do not think a New chat withdraws a restart authorisation. Say the word and I will widen it — it is a small, mechanical change.

Related, and worth stating plainly since it is the failure mode I would attack if I were reviewing: if the handler **throws** after a recovered answer has been claimed, the answer is consumed and gone (it is one-shot). The agent then sees an error and can call again, which pops a fresh card — i.e. it degrades to today's behaviour rather than to something worse. I chose not to re-remember on throw because that would need the ask id plumbed back out of `confirm`, but if you would rather it did, that is the second-cheapest change here.

artokun (commented):
COMMENT INTENDED: APPROVE

Reviewed `84cc7514367c7513b78a9b88fac777fb92a730ce`. Do not bank anything naming `c01bb60` / `d6f1c5d`.

I read `confirm`, `takeRecoveredConfirmAnswer`, `withReplyNote`, the takeover listener, `takeLateAskReply` / `askRidToId`, and the panel's `ask_user` path — not just the patch. The five points you named, in order.

## 1. Acting on a ≤5 min-old "Yes" for a restart — ship it

The alternative you offered (report the recovered answer, still require a fresh card) would leave the reporter in the same hole: they clicked "Yes, go ahead", watched the card collapse, and the next call asked them again. That is the defect.

What makes acting acceptable, rather than "consent minted at an earlier moment":

- The card is **still on that tab**. This is not resurrecting a dismissed dialog; it is reading the answer the panel already accepted.
- Identity is exact `ask_id`, same tab, byte-identical header and question, one-shot.
- The window is not new — it is the bridge buffer that already existed. Past `LATE_ASK_TTL_MS` the answer is gone and a fresh card is dispatched.
- The busy-queue guard still runs **after** the recovered yes (configured-command path at the verified `getQueueVerified` check; Manager reboot keeps the existing refuse-if-busy). A render that started in the gap still blocks the restart unless `force:true`.
- The disclosure names the card's age and that the click time is unknown (`84cc751`), and tells the agent to check if anything changed.

I would not restrict this to report-only. A destructive gate must not accept **someone else's** yes, or a yes from a **different question**, or a yes after the card is gone. This is none of those.

## 2. Takeover is the right retirement boundary — do not widen it

Agreed with the argument you posted at 06:49Z. `AskAnswers.closeAsks` retires **answers as conversation content**. A restart confirmation is a boolean gate on a machine the same browser tab still fronts. New chat / rewind / provider-switch do not mean "I withdraw consent to restart ComfyUI", and the card is still painted on that tab.

Takeover is different in kind: a recurring `wf:` key is inherited by a **different** browser tab that never saw the card. That is the one event where "same person, same surface" stops being true. Joining only that listener is correct, and it is the only way to join without replacing #486 (`setTabTakenOverListener` is a setter).

If a later report shows a New-chat agent restarting from an old card the user no longer sees, widen then. I do not see that from the code. Consuming the answer and then throwing degrades to a fresh card, not to a silent second restart — that is the right one-shot failure mode.

## 3. The #486 test was loosened the right way

The old one-expression pin would have forced a second `setTabTakenOverListener`, which would have silently un-shipped `AskAnswers.closeAsks`. The new pin is the actual guarantee: the listener **body** still contains `AskAnswers.closeAsks(tabId)`, and exactly one `setTabTakenOverListener(` exists. Dropping `closeAsks` from the body still fails it (your M9). The confirm-recovery suite separately pins `forgetAbandonedConfirmCards(tabId)` in the same call. #486 is not un-shipped.

## 4. `withReplyNote` cannot land on the wrong reply

`notes` is a local array created on each wrapper entry. `ctx` is shared across tool calls; the note is not on `ctx`. `onRecoveredAnswer` closes over that call's `note`. Two overlapping `panel_restart_comfyui` calls each have their own array.

Append-as-own-block is the right placement: `ok(object)` stays `content[0]` (parseable JSON), the note is `content[1]`. Prepend would break every reader that takes index 0; splice would corrupt the document. I did not find a downstream of this tool that assumes a single block.

## 5. Both premises hold

**Panel `timeout_ms`:** in `comfyui-mcp-panel/web/js/comfyui-mcp-panel.js`, `msg.timeout_ms` is referenced once — `#517` `awaitDuplicateReply`. `ask_user` is `await onAsk(msg, socketId)` until the user picks or `#952` withdraws the card on a **replaced** connection. The command's own timeout does not retire the card. That is why a late click still happens.

**Bridge buffer is not `pa-`-only:** `UiBridge.send` registers `askRidToId` for any non-empty `ask_id` (confirm uses `randomUUID()`). A late `ok` reply writes `lateAskReplies` keyed by that id. The journal sink is the thing that declines non-`pa-` ids; `takeLateAskReply` does not. The existing real-socket test at `ui-bridge.test.ts` (~4425) is the right proof; this PR does not re-implement that buffer.

## Not blocking

- One-shot on an **empty** look: if the agent retries before the user clicks, the memory is dropped and a second card is painted; the first card's later click then sits in the buffer unclaimed. Same class as today's two-card `#952` apology, not a silent restart. Worth a sentence in the timeout reply (already says "if the user answers it now and then asks again").
- I could not reproduce the reporter's session either. This still fixes the adjacent, load-bearing hole: an accepted click after #404's 90s is no longer thrown away.

Approve on `84cc751`.

artokun (commented):

Reviewed through `84cc7514` (including the age-disclosure follow-up). Same GitHub user, so this cannot be an Approve event.

**1. Acting on the late Yes — keep that.** Same tab, byte-identical question/header, exact `ask_id`, one-shot, 5-minute TTL, busy-queue guard still after `confirm`. "Report and paint a fresh card" is the #952 two-card state this PR is trying not to recreate. Reword the note as a fait accompli ("already acted on an earlier card") rather than inviting an intercession after dispatch. Age from timeout, not click, is a nit you already started addressing.

**2. Takeover-only retirement is not enough — but not because a restart Yes is conversation content.** Your #486 distinction is right: a restart confirmation is a machine-level gate, not a quoted pick. New chat does **not** mean "I withdraw consent to restart ComfyUI." The actual leak is **visibility**. `new_session` clears the feed, so the card is gone from the UI; the next `panel_restart_comfyui` still drains the buffered Yes and restarts without painting anything. That is a silent restart on a surface that no longer shows the consent. Reconnect must keep the entry (reporter's case). Takeover is correctly joined (different physical tab). The sites that should also call `forgetAbandonedConfirmCards` are the ones that **remove the card from the user's view** — at least New chat (`new_session`) and rewind-that-clears-the-feed — not necessarily provider-switch if the card stays painted. Pin New chat: next restart on the same `wf:` tab must paint a fresh card.

Throw-after-claim degrading to a fresh card is acceptable; do not re-remember on throw.

**3. Loosened #486 pin — still bites.** Exactly-one `setTabTakenOverListener` + `closeAsks` in the 300-char window still reds if `closeAsks` leaves the body. Residual: `toContain` matches a commented-out call. Tighten to a statement, not a substring. Prefer `import.meta.url` over cwd-relative `readFileSync` in the recovery test.

**4. `withReplyNote` concurrency — sound.** Per-call `notes` local; `onRecoveredAnswer` closes over that invocation. Shared `ctx` cannot swap notes. Appended block keeps `content[0]` as JSON.

**5. Premises hold.** Panel's only `msg.timeout_ms` reader is the #517 duplicate-reply await; confirm frame does not carry it, so the card outlives 90s. Bridge buffers a late `ask_user` answer under a non-`pa-` id (existing socket test). Mechanism is not inert.

artokun:
Standing verdict for `84cc7514` is **COMMENT INTENDED: REQUEST_CHANGES** (review just posted), not the 15m watcher's APPROVE on the same SHA.

Takeover-only retirement is still the gap: New chat clears the card from the feed, then the next `panel_restart_comfyui` can silent-Yes from the buffer. Reconnect must keep the entry. Pin New chat → fresh card. Acting on a same-tab late Yes when the card is still visible remains the right call.

artokun:
Thanks — two reviews landed and they **disagree**, so I did not count votes. The REQUEST_CHANGES carried a specific, falsifiable claim; I tested it, it was **right**, and it corrects reasoning I had posted myself an hour earlier. All three of its findings are now fixed in `fa34772`.

## 1. The visibility leak — CONFIRMED by execution, and fixed

The REQUEST_CHANGES reviewer rejected my argument for takeover-only retirement, and not by disputing the #486 distinction (they agreed a restart consent is a machine-level gate, not conversation content). They said the leak is **visibility**: New chat blanks the feed, so the card is gone from the screen while the buffered Yes survives — and the next call would restart with nothing painted.

I checked rather than took it:

```
comfyui-mcp-panel/web/js/comfyui-mcp-panel.js  function newChat()
  …
  resetFeed();                 // ← the feed the card was painted into is blanked
```

They are right, and the same is true of two siblings they gestured at: **rewind** discards everything after the anchor, and **`resume_session`** repaints the log from another thread. In all three the card leaves the screen, so claiming its answer afterwards is a **silent restart on a surface showing nothing the user can point at** — strictly worse than one extra card.

`forgetAbandonedConfirmCards` now rides **all four** boundaries: takeover, `new_session`, `rewind`, `resume_session`.

**A provider switch is deliberately still excluded, and now pinned as excluded** — it leaves the card painted, so the consent is still on screen and still answerable. That is the reviewer's own line ("not necessarily provider-switch if the card stays painted") and I would rather it fail a test than be copy-pasted in later without the argument.

A **reconnect** is none of the four — same browser tab, same incarnation, feed intact — so the reporter's "do not lose it across agent/panel reconnect" case still works. That is the property I was protecting with the narrow boundary, and it survives the widening.

While fixing this I found a related defect neither review named: the entry was keyed by the raw `ctx.tabId`, which for a **scope-addressed** session is a scope address, not the tab the card renders on. Those four retirement sites pass real tab ids, so a scope-bound session's entry could never have been retired by any of them. It now keys by `journalTabFor(ctx)` — the same convention #486 uses, for the same reason.

## 2. The loosened #486 pin — the residual was real

`toContain("AskAnswers.closeAsks(tabId)")` is satisfied by a **commented-out** call, so the guarantee could have survived its own removal. Tightened to a statement match:

```js
expect(src.slice(takeoverAt, takeoverAt + 300)).toMatch(
  /^\s*AskAnswers\.closeAsks\(tabId\);\s*$/m,
);
```

New mutation **M11** — comment out that one line — turns the #486 test **red**. Under the old `toContain` it would have stayed green. Also switched my own `readFileSync` to `import.meta.url`; it was cwd-relative and would have passed or failed on where vitest was launched.

## 3. Note reworded as a fait accompli

It ended "check with them if anything has changed since", which invites an intercession into something already dispatched. It now says the call has **already acted**, not to re-confirm, and not to present it as pending.

## New mutations

| # | Mutation | Result |
|---|---|---|
| M10 | `new_session` stops calling `forgetAbandonedConfirmCards` | RED: the new boundary-wiring test |
| M11 | `AskAnswers.closeAsks(tabId)` commented out in the listener | RED: **#486's own** test (green under the old `toContain`) |

Plus M1–M9 re-verified. Full suite on the current tree: **609 files / 11174 tests green**.

## One correction to my own earlier report

An intermediate full-suite run of mine reported failures in `training-jobs` and `panel-restart-configured-command`. Those were **not real**: this branch's worktree was deleted underneath me mid-run, so `git`/`vitest` walked up to the stale main checkout and I was running `a33a666`'s code. Nothing was lost (the commit was already in the shared object store and is pushed), but I am flagging it rather than leaving two phantom failures in the record. `panel-restart-configured-command` passes in isolation and in the full run; it stubs `confirm` outright, so this change cannot reach it.

## Where I still want a second look

The scope-address key fix (`journalTabFor`) landed in the same commit as the boundary widening and is **not separately covered** — the recovery tests all use a real `wf:` tab id, so no test would fail if it went back to `ctx.tabId`. I did not add one because the fixture would have to fake `resolveSharedTabId`, and I would rather say the coverage is missing than imply it exists.

PR #1958, opened 2026-08-21
release: v0.52.42 — multi-connection downloads, self-restart lock reclaim

artokun:
Cuts **0.52.42** from `main` (after v0.52.41).

Carries:

- #1956 / #1697 — multi-connection model downloads, with the single-connection path proven as fallback
- #1955 / #1953 — self-restart no longer leaves an unreclaimable panel-op.lock

Held out: mcp#1421 and panel#1555/#1556/#1557 claimed this cycle; panel#1554 inflight (`wt-p1554-cc`); panel#1472 typescript 5→7 (blocked).

Do **not** squash-merge. Merge-commit (keeps the `v0.52.42` tag on the version commit). Tag is local; push `v0.52.42` after merge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -759,6 +759,17 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+## [0.52.42] - 2026-08-20
+
+### MCP
+
+#### Added
+- multi-connection model downloads, with the single-connection path proven as fallback (#1956)
+
+#### Fixed
+- self-restart no longer leaves an unreclaimable panel-op.lock (#1955)
+
+
 ## [0.52.41] - 2026-08-20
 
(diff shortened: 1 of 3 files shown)


artokun:
Hold merge of **0.52.42** until #1956's REQUEST_CHANGES is actually landed.

#1956 squash-merged at `28652b9` — the SHA that had **COMMENT INTENDED: REQUEST_CHANGES**. `drainSegmentBody` still ignores `bytesWritten`, so `contiguousPrefix` can over-report a hole (short `filehandle.write` vs `state.written` vs disk). Publishing that as the #1697 headline ships a corrupt-download window.

I am not merging #1958 from grok-swarm. Follow-up the short-write accounting, then recut 0.52.42.

PR #1964, opened 2026-08-21
fix(panel#1557): current-workflow rebind restores the live canvas after reconnect

### LINKED ISSUE #1557: Live canvas remains disconnected after reconnect and current-workflow rebind fails
The user requested a live-canvas workflow edit. panel_graph_outline failed because the panel tab was disconnected after a restart/reload. A recovery call panel_set_workflow_target({mode:"current"}) also failed: no connected tab belonged to the Codex backend, and the session remained fenced to the previous workflow instance. A retry of panel_graph_outline failed identically.

### Steps to reproduce
1. Have a ComfyUI workflow tab connected to the agent.
2. Restart or reload ComfyUI/the browser tab.
3. Call panel_graph_outline.
4. Call panel_set_workflow_target({mode:"current"}).
5. The panel remains disconnected and graph tools cannot be used until a manual browser refresh.

### Exact error (scrubbed)
"no connected tab with id ... Connected: none — the ComfyUI panel tab is not connected"
"panel_set_workflow_target({mode:'current'}) did NOT restore this session's graph binding"
"no connected tab belongs to this conversation's backend (codex)"

### Fix
applied locally: no
upstream-only: yes
Precise change needed: automatically re-establish the panel socket/backend binding after ComfyUI restart or reload, or expose a reliable reconnect action that does not require a manual browser refresh. Current recovery call changes mode but cannot observe or bind the live canvas.

### Environment
OS: Linux
ComfyUI: 0.33.1
GPU: NVIDIA GeForce RTX 2080 Ti
comfyui-mcp: 0.52.1
panel: 0.11.44

artokun:
Fixes artokun/comfyui-mcp-panel#1557

After a ComfyUI restart/reload, `panel_graph_outline` failed with `Connected: none`. The documented recovery `panel_set_workflow_target({mode:"current"})` also failed: no connected tab belonged to the Codex backend, and the session stayed fenced to the previous workflow instance. Retrying `panel_graph_outline` failed identically until a manual browser refresh.

Codex panel_* tools are scope-bound (`orchestrator::codex`). The zero-tab DEFER path only fired when `rebindToActiveTab` threw, which a real-tab ctx does. A scope ctx returned a refusal object instead, so the recovery neither deferred nor rebound.

What the user now observes:
- With no canvas connected, `mode:"current"` defers (clears the stale binding) instead of failing as a backend-mismatch.
- That consent is kept, so the next graph call binds the canvas that reconnects.
- A dead/ambiguous pin plus exactly one live canvas is adopted even when that hello joined the default backend. Idle (no pin) unique-foreign still refuses.

Tests drive the shipped `panel_set_workflow_target` / `panel_graph_outline` path with the real `TurnOriginTracker` + `makeScopeRepinHandler`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -759,6 +759,25 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+### MCP
+
+#### Fixed
+- **`panel_set_workflow_target({mode:"current"})` rebinds a Codex session after reconnect (panel#1557).**
+  After a ComfyUI restart/reload, Codex's live-canvas tools are scope-bound
+  (`orchestrator::codex`). The documented recovery refused with "no connected
+  tab belongs to this conversation's backend (codex)" and "did NOT restore this
+  session's graph binding": a scope ctx returns a refusal instead of throwing,
+  so the zero-tab DEFER path never fired, and a unique canvas that re-hello'd
+  without `backend` joined the default conversation so Codex could not adopt it.
+  The session stayed fenced to the previous workflow instance; retrying
+  `panel_graph_outline` failed identically until a manual browser refresh.
+
+  Zero connected tabs now DEFER the same way a real-tab session already did,
+  and the consent is kept so the next graph call binds the canvas that
+  reconnects. A dead or ambiguous pin plus exactly one live canvas is adopted
+  even when that hello joined the default backend — idle (no pin) unique-foreign
+  still refuses, so another conversation's tab is not stolen.
+
 ## [0.52.42] - 2026-08-20
 
 ### MCP
@@ -768,7 +787,6 @@ All notable changes to this project are documented here. This project adheres to
 
 #### Fixed
 - self-restart no longer leaves an unreclaimable panel-op.lock (#1955)
-
 
 ## [0.52.41] - 2026-08-20
 
--- a/src/__tests__/orchestrator/codex-reconnect-rebind.test.ts
+++ b/src/__tests__/orchestrator/codex-reconnect-rebind.test.ts
@@ -1,0 +1,261 @@
+// panel#1557 — after a ComfyUI restart/reload, panel_graph_outline failed with
+// "no connected tab … Connected: none", and the documented recovery
+// panel_set_workflow_target({mode:"current"}) also failed: "no connected tab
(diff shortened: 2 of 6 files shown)


artokun:
Independent post-merge verification of this fix (panel#1557 autopilot), run against `origin/main` in a clean worktree. Recording it because this merged ~8 min after opening with no adversarial review.

**The fix is genuine.** Reverse-applying the whole production diff (`panel-tools.ts`, `turn-origins.ts`, `index.ts`) turns 3 of the 4 new tests red, and they fail with the *verbatim* string from the reporter:

```
AssertionError: Error: panel_set_workflow_target({mode:"current"}) did NOT restore this session's graph binding.
```

That is the reported symptom reproduced by execution, not by reading — so the tests pin the actual bug rather than the author's model of it. Restored: 4/4 green. Neighbours are clean too: `turn-origins`, `fence-rebind-transient`, `panel-session-rebind-residuals`, `pin-repin-uses-a-real-tab-id`, `repin-refusal-reason`, `start-failure-rebind-heldmail` — 119 tests, all passing.

**One coverage gap worth knowing about (not a defect).** The `claimTab` wiring added at `src/orchestrator/index.ts:3229` is *not* covered: the test injects its own `claimTab` at line 161 of `codex-reconnect-rebind.test.ts`, so deleting the production wiring entirely leaves all 4 tests green. I checked the wiring by hand instead — `claimTab` writes `tabBackends`, and `backendForTab` (`index.ts:1851`) reads exactly that map, so the claim does make `backendForTab(only) === backend` and the fresh pin survives `resolvedPinOf`. Correct, just unpinned; a future refactor could drop that line silently.

**Two safety questions I checked rather than assumed:**
- `pendingScopeRecoveryConsent` has no expiry, so I looked for a stale-consent steal. It self-clears in `ensureReachable` as soon as *any* interactive tab exists (rebound or not), so the window is only the zero-tab gap right after a restart — the first tab to appear is the reconnecting canvas. Bounded.
- The unique-canvas adoption is gated on `recovering` (dead or ambiguous pin), and `undefined` (no pin) falls through to the refusal, so the idle "never steal another backend's tab" P0 still holds — the 4th test covers this and passes with and without the fix.

No panel-side Playwright run: this change is entirely orchestrator-side and renders nothing in the sidebar, so the panel repo has no diff to exercise.

PR #1966, opened 2026-08-21
fix(panel#1558): promoted subgraph widget values persist when control_after_generate is randomize

### LINKED ISSUE #1558: panel_set_widget cannot persist promoted subgraph widgets when control_after_generate is randomize
panel_set_widget successfully changed promoted widgets on an LTX-2.3 Image-to-Video subgraph node, but emitted warnings that control_after_generate='randomize' governs the inner width/height/duration/fps widgets and that the fixed setting is not promoted or settable from the parent scope. This makes a normal parent-scope configuration non-persistent and requires entering the subgraph and addressing hidden inner node ids.

### Steps to reproduce
1. Load the bundled ltx-2.3-img2vid workflow.
2. Call panel_set_widget on outer subgraph node 320 for value_2/value_3/value_4/value_5.
3. Observe successful writes accompanied by warnings that subsequent runs randomize the values and that the inner node ids are unavailable from the current scope.

### Exact error / warning
control_after_generate='randomize' governs widget "value_2"... the value you set will NOT persist. Set "fixed" ... not promoted onto subgraph node 320 ... Enter the owning subgraph first.

### Fix
Applied locally: no. Upstream-only / panel capability limitation: yes. Suggested fix: expose/persist control_after_generate for promoted subgraph widgets, or allow panel_set_widget to route the fixed setting to the owning inner node automatically.

### Environment
Windows 11; ComfyUI 0.32.0; GPU NVIDIA GeForce RTX 5090 32GB; comfyui-mcp 0.52.41; panel 0.15.16; ComfyUI frontend version not available in the supplied environment line.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-luna` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #650

artokun:
Fixes artokun/comfyui-mcp-panel#1558

panel_set_widget on a promoted subgraph widget already succeeded, but the inner widget is governed by control_after_generate=randomize that is not promoted onto the parent. The panel warned and named panel_enter_subgraph then inner fixed then panel_exit_subgraph, which required hidden inner node ids.

On that unpromoted persist warning, panel_set_widget now follows the panel's own remedy: enter, pin the inner control to fixed, exit. The value the caller just set holds across runs. A DIRECT seed randomize warning is left alone. A failed pin never turns the already-verified write into an error.

### DIFF
--- a/src/__tests__/orchestrator/promoted-control-persist.test.ts
+++ b/src/__tests__/orchestrator/promoted-control-persist.test.ts
@@ -1,0 +1,289 @@
+// panel#1558 — a successful promoted subgraph write still will not persist when
+// the inner widget is governed by control_after_generate='randomize' and that
+// control is NOT promoted onto the parent. The panel warns and names an
+// enter → set-inner-fixed → exit sequence. panel_set_widget must follow that
+// remedy itself so the value the caller just set holds, without hidden inner ids.
+//
+// These tests drive the SHIPPED handler. A write with no persist warning is
+// untouched. A DIRECT seed warning (control already on the addressed node) is
+// not auto-pinned — randomize there is often intentional.
+
+import { describe, expect, it } from "vitest";
+
+import {
+  buildPanelToolDefs,
+  makePanelToolCtx,
+  type PanelToolCtx,
+  type ToolResult,
+} from "../../orchestrator/panel-tools.js";
+import {
+  addressedNodeMatchesPersistRemedy,
+  parseUnpromotedControlPersistRemedy,
+} from "../../orchestrator/promoted-widget.js";
+import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+
+const TAB = "wf:ltx-img2vid";
+
+/** The panel's #650 wording for the reporter's LTX-2.3 node 320 / inner 312. */
+const REPORTER_WARNING =
+  `control_after_generate='randomize' governs widget "value" on node 312: ` +
+  `ComfyUI automatically CHANGES this value on subsequent runs (a new random value each run), ` +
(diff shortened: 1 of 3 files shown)


artokun:
Independent post-merge verification (autopilot, panel#1558). This merged without a review, so I verified it rather than duplicating the work. No defect found.

**Input shape — the load-bearing risk.** The parser regex-matches prose emitted by the *other* repo, so a shape mismatch would have made the fix silently inert. Checked against the emitter, `comfyui-mcp-panel/web/js/lib/control-after-generate.js:195-200` + `controlAfterGenerateRemedy`, and all six patterns match the real text:
- `will NOT persist` ✔
- `is NOT promoted onto subgraph node 320,` → captures `320,`, trailing comma stripped ✔
- `panel_enter_subgraph(node_id=320)` ✔
- `panel_set_widget(node_id=75, widget='control_after_generate', value='fixed')` → inner id + control ✔
- `panel_exit_subgraph() N times` present only for nested, else falls back to `enterPath.length` ✔

Both negative paths are correctly skipped: the `is promoted onto ... as` variant does not match `is NOT promoted`, and the DIRECT form emits no `NOT promoted onto subgraph node` at all.

**Mutation.** Reverted only the call-site wiring (`if (!first.isError) return first;`) at the merge commit in a scratch worktree: **5 of 13 tests go red**, including the reporter's case and the exit-disclosure test. So the suite pins the call site, not just the helper — the failure mode where a one-line wiring change is invisible to helper-level tests does not apply here.

**Not run:** the panel Playwright suite. This change is orchestrator-side and the panel e2e harness fakes the orchestrator (MockBridge), so it could not cover this fix either way — no browser coverage is claimed.

Issue #1963, opened 2026-08-21
Feature: transport-aware auto-update — check+apply at the desk (orchestrator AND panel), check-only while paired over a mobile tunnel


artokun:
### The behaviour I want (feature request, not a bug)

Auto-update should be **conditional on how I am currently connected**, not a global on/off:

| context | check for updates | APPLY (restart) |
|---|---|---|
| **At my desk** — local / LAN | ✅ yes, for **both** the orchestrator AND the panel | ✅ yes |
| **On the go** — mobile tunnel | ✅ yes, and tell me | ⛔ **never** — the restart disconnects me |

Today it is effectively all-or-nothing: auto-update is default-on and hourly, and the only escapes
(`COMFYUI_MCP_AUTO_UPDATE_DISABLE=1`, `COMFYUI_MCP_AUTORESTART=0`) are global. So a user who wants
to stay current at their desk must accept being disconnected when away, or opt out entirely and
then silently run stale builds at the desk too. I have hit both halves of that trade in one day.

### Why this is decidable, not a heuristic

The system already knows the difference — #875 states it directly:

> "LAN users are fine — pinning `COMFYUI_MCP_PAIR_TOKEN` auto-starts the pairing listener at boot
> and the phone reconnects silently. **Tunnel users get no equivalent.**"

A restart severs a **quick-tunnel** session with certainty, because the cloudflared hostname
rotates per run. A restart does **not** sever a LAN session with a pinned pair token. That is a
property of the transport, not a guess about user intent — so the gate can be exact:

```
applyUpdate = updateAvailable && !(activeTransport === 'quick-tunnel' && pairingActive)
```

with `relay` backends treated as safe (stable URL), and LAN safe.

### Two things that must be true for this to work

**1. `pairingActive` must be sticky, not socket-derived.** See comfyui-mcp#1961 — the current gate
is `pairTunnel !== null && bridge.hasLiveHeadlessClient()`, and a backgrounded phone has no live
socket, so the gate opens and the tunnel gets rotated anyway. Whatever this feature keys on has to
survive a locked screen and a 5G/wifi handoff. A user-set toggle at pair time is the robust
version; a generous grace window is the cheap one.

**2. Checking must be separated from applying.** See comfyui-mcp-panel#1559 — right now the panel
version check is floor-only and reports "healthy" while a newer panel exists, so there is no
signal to act on even at the desk. Checking is a few KB and safe on a metered link; applying is
what must be gated.

[...]

PR #1967, opened 2026-08-21
fix(1961): a backgrounded phone keeps its tunnel hostname across auto-update

### LINKED ISSUE #1961: #1176 is closed but unfixed in 0.52.41 — tunnelPairingLive still uses hasLiveHeadlessClient(), so a backgrounded phone still gets its tunnel rotated

**#1176 is closed, but the code it identifies is unchanged in the shipped 0.52.41.**

Verified in the installed package
(`~/AppData/Local/npm-cache/_npx/<hash>/node_modules/comfyui-mcp@0.52.41/dist/orchestrator/index.js`):

```js
tunnelPairingLive = () => pairTunnel !== null && bridge.hasLiveHeadlessClient()
```

and `dist/services/ui-bridge.js`:

```js
hasLiveHeadlessClient() {
  for (const c of this.conns.values()) if (c.headless === true) return true;
  return false;
}
```

That is the exact gate #1176 reported as the wrong liveness signal. Grepping the shipped `dist/`
for any grace-window or last-seen concept (`hasRecentHeadless`, `lastHeadlessSeen`,
`headlessGraceMs`, `recentlyPaired`) returns **nothing** — the only signal is a currently-open
socket.

### Why that still breaks mobile

A phone in a pocket drops its socket routinely (screen off, app backgrounded, network handoff
between wifi and 5G). With socket-presence as the only liveness test:

1. Phone backgrounds -> socket closes -> `hasLiveHeadlessClient()` is false.
2. The hourly auto-update finds a new version, sees "no headless client", and restarts.
3. The restart rotates the cloudflared quick-tunnel hostname (and the pair token, if unpinned).
4. The user reopens the app to `Failed host lookup: <old>.trycloudflare.com` — DNS failure,
   because the hostname no longer exists. The saved URL can never reconnect.

The user cannot distinguish this from a crash, and the only escapes are opting out of a
default-on feature (`COMFYUI_MCP_AUTO_UPDATE_DISABLE=1` / `COMFYUI_MCP_AUTORESTART=0`) or
standing up a private relay Worker.

### Steps to reproduce

1. Pair a phone over the default cloudflared quick tunnel.
2. Background the app / lock the phone so the websocket closes.
3. Let the hourly auto-update find a newer npm version while the agent is idle.
4. Reopen the app -> `Failed host lookup` on the old `*.trycloudflare.com` hostname.

### Fix

**1. Liveness for a mobile client must be "seen recently", not "socket open now".**
Track `lastHeadlessSeenAt` and treat a pairing as live within a grace window (a few hours is
reasonable for a phone), so:

```js
const tunnelPairingLive = () =>
  pairTunnel !== null &&
  (bridge.hasLiveHeadlessClient() || bridge.headlessSeenWithin(HEADLESS_GRACE_MS));
```

**2. Better: make it explicit rather than inferred — the user's own request.**
A pairing should be *sticky state*, not something re-derived from socket presence. When a phone
pairs, surface a visible toggle, defaulted ON:

> **"Don't update the orchestrator while my phone is paired"**

and keep deferring restarts until the user un-pairs or clears it. Inferring intent from a
transport that legitimately disconnects will keep producing this class of bug; an explicit flag
cannot be fooled by a backgrounded app, a tunnel hiccup, or a 5G/wifi handoff.

**3. Regardless of 1 and 2: never rotate the hostname silently.** If a restart is going to change
the tunnel URL, that is a user-visible event. Warn at pair time (as #875 suggested) and, when a
deferral is skipped, say so.

**4. Persist the pair token by default** (#875's second suggestion) so at least rotation #1 is
gone for everyone with no configuration.

### Note on the related report

I filed comfyui-mcp-panel#1559 tonight asking for a `behind_latest` signal on the version check.
That request is deliberately **report-only, not auto-install** — and this issue is why that
distinction matters. Checking for a new version is cheap and safe on a metered mobile link;
*applying* one requires a restart, which is exactly what must not happen while a phone is paired.

### Environment

OS: Windows 11 Home 10.0.26200
comfyui-mcp: 0.52.41 (verified against the shipped dist, not source)
panel: 0.15.31
ComfyUI: 0.33.3 | frontend 1.49.6 | GPU RTX 5080 16 GB

artokun:
Fixes #1961

## What was actually unfixed

#1176 / PR #1185 *did* ship a recency window in 0.50.50, and it is in the 0.52.41 dist. The reporter grepped for names that never existed (`hasRecentHeadless`, `lastHeadlessSeen`, `headlessGraceMs`) and reconstructed `hasLiveHeadlessClient` as `return false`. The compiled function is:

```js
hasLiveHeadlessClient() {
  for (const c of this.conns.values()) if (c.headless === true) return true;
  return this.recentHeadlessWithin(HEADLESS_RECENCY_MS);
}
```

Two remaining holes still produced the reported `Failed host lookup` on a pocketed phone:

1. Recency was folded *into* `hasLiveHeadlessClient`, so `tunnelPairingLive` still read as the live-only one-liner the reporter pasted from dist.
2. The window was **30 minutes**, shorter than the hourly auto-update check. An armed restart fires 30 minutes after the socket closes — a commute, a meeting, or a phone in a pocket at work.

## What the user now observes

A phone that backgrounds / locks (socket gone) still defers a tunnel-rotating self-restart for a work day. Reopening the app reaches the same `*.trycloudflare.com` hostname.

- `tunnelPairingLive(pairTunnel, bridge)` is now `pairTunnel !== null && (live || headlessSeenWithin(HEADLESS_RECENCY_MS))`.
- `hasLiveHeadlessClient` is live sockets only again (the name matches).
- `HEADLESS_RECENCY_MS` is 8 hours. The sticky `isHeadless()` is still rejected: a phone that has been gone longer than the window stops deferring on its own.

Not in this PR (separate issues): explicit "don't update while paired" toggle (#1963), stable tunnel hostname (#1151).

### DIFF
--- a/src/__tests__/services/headless-recency-window.test.ts
+++ b/src/__tests__/services/headless-recency-window.test.ts
@@ -1,23 +1,30 @@
-// #1176 — a BACKGROUNDED phone is not a departed phone.
+// #1176 / #1961 — a BACKGROUNDED phone is not a departed phone.
 //
 // #875 defers a self-restart while a paired phone is present, because a restart
 // mints a NEW cloudflared quick-tunnel hostname and the old one stops resolving.
 // The gate asked `hasLiveHeadlessClient()`, which reads currently-OPEN sockets —
 // and a phone at work with its screen off has had its socket suspended or closed
-// by the mobile OS. So the gate opened, the orchestrator self-updated
-// 0.50.39 → 0.50.41, and the reporter picked up their phone to:
+// by the mobile OS. So the gate opened, the orchestrator self-updated, and the
+// reporter picked up their phone to:
 //
 //   Failed host lookup: 'cameron-timing-face-spies.trycloudflare.com'
 //   (No address associated with hostname, errno = 7)
 //
 // A DNS failure, not a timeout or a refusal: the hostname no longer existed.
 //
-// The sticky isHeadless() was rejected for this gate and that was right — a phone
-// that paired once and left would defer updates forever. But those are not the
-// only two options, and the state between them is the NORMAL one for a phone.
+// #1176 folded a recency window into `hasLiveHeadlessClient`. That left
+// `tunnelPairingLive` looking live-only (the reading #1961 verified against
+// 0.52.41's dist) and chose thirty minutes — shorter than the hourly auto-update
+// check, so a pocketed phone still had its hostname rotated.
+//
+// The shipped function is `tunnelPairingLive`. These tests drive it.
 
 import { describe, expect, it } from "vitest";
+import { readFileSync } from "node:fs";
 import { UiBridge, HEADLESS_RECENCY_MS } from "../../services/ui-bridge.js";
+import { tunnelPairingLive } from "../../orchestrator/tunnel-pairing-live.js";
+
+const TUNNEL = { url: "wss://test.trycloudflare.com", stop: () => {} };
 
 /** A bridge with its headless-disconnect clock placed at a chosen age. */
 function bridgeWithHeadlessGoneFor(ms: number | null): UiBridge {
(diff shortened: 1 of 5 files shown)


artokun:
Superseded by https://github.com/artokun/comfyui-mcp/pull/1971 (Fixes #1963, shipped in 0.52.44).

#1961 asked for sticky pairing + an explicit pair-time toggle, not a longer recency window. Main already keys APPLY on `pairingActiveOf(pairTunnel)` (`pairTunnel != null`), so a backgrounded/locked phone cannot open the restarter. This PR's 8-hour recency rewrite of `tunnelPairingLive` would rebase a weaker gate onto that.

Closing rather than rebasing. Evidence on #1961.

artokun:
Superseded by #1971; see issue #1961.

PR #1975, opened 2026-08-21
chore(hygiene): no non-test file crosses 1,000 lines without being named in the baseline

artokun:
Part of #2003.

Encodes one rule from pstack's `thermo-nuclear-code-quality-review` checklist as a CI ratchet: do not let a PR push a file from under 1,000 lines to over 1,000 lines without a very strong reason. This repo has forty non-test files under `src/` already past that mark (`src/orchestrator/panel-tools.ts` is 17,813 lines), and every one of them got there by increments nobody refused. The gate refuses the next increment.

`scripts/check-file-size.mjs` is written in the exact idiom of `check-unknown-collapse.mjs`: a header that explains why, `--list` / `--baseline` modes, `--src` / `--baseline-file` so the gate can be tested against fixture trees, and a shrink-only direction. The rules: a file over 1,000 lines must be named in `scripts/file-size-baseline.json` with its count; a baselined file may not grow past that count (it may shrink, and a shrink makes the baseline stale, which also fails with the instruction to rewrite it); a file not in the baseline may not cross the line. Counts are raw `wc -l` lines, comments and blanks included, so the number can be checked by hand and cannot be gamed.

The baseline is JSON rather than the sibling's text format because each entry is a record: `{"path", "lines", "reason"?}`. The `reason` field is the escape hatch and the deliverable: a genuine exception adds or bumps its entry in the same PR and says why, and `--baseline` preserves existing reasons when it rewrites counts after a split. The forty grandfathered entries have no reason, which is the honest label for debt. Unlike `unknown-ok:`, the reason lives in the baseline rather than at a site, because a file's size has no single site.

The exact-count ceiling means a PR that shrinks a baselined file must also touch the baseline, and a refactor inside one that adds a line must first take one out. That is the trade the unknown-collapse ratchet already made, and one shape of ratchet beats two; the header says so.

Wiring: `lint` becomes `tsc --noEmit && node scripts/check-file-size.mjs`. The gate runs inside `npm run lint` only, so CI has one signal for it rather than a duplicate step; `ci.yml` is untouched. The vitest in `src/__tests__/scripts/check-file-size.test.ts` covers every rule (cross, grow by one, shrink, drop under, delete, reason survives rewrite, reason does not license growth, hand-bumped exact entry accepted, CRLF not double-counted, missing trailing newline still counted, `.test.ts` / `.spec.ts` / `__tests__` / `.d.ts` ignored, malformed baseline and `limit` mismatch fail loudly with the path, `--baseline` on a malformed file refuses without clobbering it, a flag with a missing value is a usage error rather than a silent fallback to the real tree) and asserts the checked-in baseline matches the checked-in source. CONTRIBUTING gets one Conventions bullet and its `npm run lint` line no longer claims "type-check only".

Verified locally: `npm run lint`, `npm run smoke` (37 core + 3 compact + 95 panel tools), `check:vocabulary`, `check:unknown-collapse`; a 1,426-line probe file fails the gate by name, deleting it goes green, appending one line to `panel-tools.ts` fails with `17813 -> 17814 (+1)`, reverting goes green. The full vitest run has 17 failures in `src/__tests__/services/{extra-paths*,training-datasets,comfy-view-ref}.test.ts` that are macOS-environment artifacts (`/private/var` vs `/var` realpath, and a worktree path tripping a dataset-root guard); none touch code this PR changes.

Coordination notes: the anti-slop foundation PR also edits the `lint` script on the same line; this PR keeps that edit to the one line. Follow-ups deliberately not done here: `scripts/tool-doc-examples.ts` is 1,202 lines and outside the `src/` scope of this gate; `walk`, `flagValue`, and the baseline-reading shape are now duplicated between the two ratchets and could move to `scripts/lib/`; README's lint-table row is owned by the README unit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ npm test           # vitest
 ```
 
 - `npm run build` type-checks and compiles to `dist/` (`tsc`).
-- `npm run lint` type-checks only (`tsc --noEmit`).
+- `npm run lint` type-checks (`tsc --noEmit`) and runs the file-size ratchet (see [Conventions](#conventions)).
 - `npm test` and `npm run test:watch` run the vitest suite.
 - `npm run dev` runs the server from source via `tsx`.
 - `npm run docs:gen` rebuilds the docs tool reference from the live schemas (see [Docs](#documentation)).
@@ -63,6 +63,14 @@ Business logic lives in `src/services/<name>.ts`. The matching
   - Validate filesystem paths against traversal and symlink escapes (resolve the path, then check it stays inside the intended root).
   - Validate values that reach a subprocess argv (reject a leading `-` and control chars; use
     `--end-of-options` for git, and the equivalent for other tools).
+- **File size.** `npm run lint` fails if a non-test file under `src/` crosses 1,000 lines, or if
+  one of the files already past that mark (named with its line count in
+  `scripts/file-size-baseline.json`) grows at all. The way to extend a monolith is to split it,
+  never to append to it. A genuine exception, such as a generated table or a file mid-split, adds
+  or bumps its entry in the same PR with a one-line `"reason"`
+  (`{"path": …, "lines": …, "reason": "…"}`); the reason is what the reviewer reads. Shrinking a
+  baselined file makes the baseline stale: run `node scripts/check-file-size.mjs --baseline` (it
+  preserves reasons) and commit the result.
 - **Plugin skills.** Every skill (`plugin/skills/<name>/SKILL.md`) ends with a
   `## Sources` section that separates **Official** sources (vendor docs, node README,
   `/object_info`) from **Empirical** ones (working graphs, observed behaviour). A skill
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "test:local-llm": "npm run build && node scripts/test-local-llm.mjs",
     "arena": "npm run build && node scripts/llm-arena.mjs",
     "smoke:panel": "npm run build && node scripts/panel-smoke.mjs",
-    "lint": "tsc --noEmit",
+    "lint": "tsc --noEmit && node scripts/check-file-size.mjs",
     "docs:gen": "cross-env COMFYUI_URL=http://127.0.0.1:8188 tsx scripts/gen-tool-docs.ts",
     "tools:dump": "cross-env COMFYUI_URL=http://127.0.0.1:8188 tsx scripts/tools-dump.mts",
     "check:vocabulary": "tsx scripts/check-tool-vocabulary.mts",
--- a/scripts/check-file-size.mjs
+++ b/scripts/check-file-size.mjs
@@ -1,0 +1,312 @@
+#!/usr/bin/env node
+/**
+ * comfyui-mcp-4x5 — the file-size ratchet.
+ *
+ * The rule, borrowed from pstack's code-quality review checklist: do not let a
+ * PR push a file from under 1,000 lines to over 1,000 lines without a very
+ * strong reason. This repo already has forty files past that mark, the largest
+ * at ~17,800 lines, and every one of them got there the same way — one more
+ * handler, one more branch, one more "it's only twenty lines" on top of a file
+ * that was already too big for anyone to hold. Nobody decided to write a
+ * 17,000-line file. The decision was made in increments, and no increment was
+ * ever the one worth refusing.
+ *
+ * A review rule that nobody enforces mechanically decays into a review rule that
+ * nobody enforces. So this is a CI gate, and it encodes the lesson rather than
+ * the reminder: the only way to extend a monolith is to split it.
+ *
+ * WHAT COUNTS. Raw lines: code, comments, and blanks alike. A "cleverer" count
+ * that skips comments invites gaming, and it would also be the wrong measure —
+ * the thing the rule protects is what a reader has to scroll through and what a
+ * reviewer has to hold in their head, and a comment is part of that. The number
+ * is what `wc -l` reports, so it can be checked by hand.
+ *
+ * THE RULES.
+ *
+ *   1. A file over 1,000 lines must be named in the baseline, with its count.
+ *   2. A baselined file may not grow beyond its baselined count. It may shrink.
+ *   3. A file not in the baseline must not cross 1,000 lines.
+ *   4. A baselined file that shrank, dropped under 1,000, or was deleted makes
+ *      the baseline STALE. That fails too, with the instruction to rewrite it —
+ *      otherwise the list stops describing the code, and a file could quietly
+ *      grow back up to a number it no longer needs.
+ *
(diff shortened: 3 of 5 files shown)


artokun:
Heads-up: this PR and #1988 both add a ratchet to the same `"lint"` line in `package.json` (`tsc --noEmit && node scripts/check-file-size.mjs` here, `&& node scripts/check-anti-slop.mjs` there). Whichever lands second gets a one-line conflict; the resolution is to chain both. Land this one first; #1988 merges last in the batch.

artokun:

artokun:
## Rebased onto current main, green on all four checks — held at the merge gate

**State: READY, HELD.** Not merged, and an agent cannot merge it. Details below.

### What changed in this pass

- Rebased onto current `main` (the branch was cut 2026-08-21 and `main` has taken ~590 commits since).
- `CONTRIBUTING.md` conflicted because #1987 unslopped that file on `main` after this branch was written. Resolved by keeping main's plain-prose style and re-writing this PR's two additions (the `npm run lint` line and the new **File size** convention bullet) in it, rather than restoring the older bold/em-dash forms.
- **Baseline rewritten against current main** (`node scripts/check-file-size.mjs --baseline`, 48 entries). The old one had been failing CI since the day it was written: 27 baselined files had grown (`orchestrator/codex-backend.ts` 1757 → 2417, `comfyui/client.ts` 1282 → 1571, `orchestrator/index.ts` 6849 → 6980) and 8 more had newly crossed 1,000 lines with no entry at all (`panel-launcher`, `panel-image-relay`, `kitchen`, `panel-console-http`, `panel-sync`, `comfy-view-ref`, `workflow-library`, `skills-access`). It then went stale a second time mid-review and was rewritten again.

### Evidence, measured on the tree rather than read

| probe | result |
|---|---|
| clean tree | exit 0 — `48 baselined file(s) over 1000 lines, none grew, nothing new crossed` |
| +3 lines appended to a baselined file | exit 1 — `config.ts  1094 -> 1097  (+3)` |
| a new 1,200-line file under `src/services/` | exit 1 — `services/__probe-big.ts  1200 ... crossed 1000 lines without being named` |
| `src/__tests__/scripts/check-file-size.test.ts` | 25/25 pass |
| `tsc --noEmit` | clean |
| CI | Build ubuntu / windows / macos + Pack & install smoke all SUCCESS |

### Why it is not merged

`codex-gate.mjs` returns a clean **SHIP** on the substance, then downgrades to **INDETERMINATE (exit 2)** because the diff touches `package.json`:

```
SHIP
[gate] INDETERMINATE — a SHIP verdict is not accepted here: touches the supply chain
       (package.json) — a human must review; re-run with AUTOPILOT_HUMAN_ACK=1 once one
       has. Merge HOLDS.
```

The guard does not read the diff, so a script-wiring line trips it exactly as a new dependency would. **The entire `package.json` diff is one line:**

```diff
-    "lint": "tsc --noEmit",
+    "lint": "tsc --noEmit && node scripts/check-file-size.mjs",
```

No dependency, lockfile, workflow or publish-path change. Once someone has read that line:

```
node .claude/autopilot/codex-gate.mjs --worktree <wt> --base origin/main --name 1975 \
  # with AUTOPILOT_HUMAN_ACK=1
```

### One thing worth deciding before it lands

The two hottest files in the tree are baselined, and they move constantly. Measured over the last 7 days on `main`:

| file | commits touching it | lines |
|---|---|---|
| `src/orchestrator/panel-tools.ts` | 145 | 14,279 → 22,367 |
| `src/orchestrator/index.ts` | 33 | 6,506 → 7,009 |
| `src/orchestrator/codex-backend.ts` | 8 | 1,670 → 2,417 |

Once this lands, each of those ~186 touches has to bump its baseline entry with a `"reason"` in the same PR — and concurrent PRs will collide in `scripts/file-size-baseline.json`. That is the gate working as designed ("this is the increment where it gets refused"), but on a repo merging ~85 commits a day it is real, recurring friction, and it is worth being a deliberate choice rather than a surprise. Exempting the two 7k/22k-line files, or gating only files a PR actually touches, are both cheaper than discovering this a week from now.

The baseline also goes stale between the CI run and the merge: it was regenerated twice in this session and CI failed in between because `index.ts` grew +29 and `panel-tools.ts` +17 while the review was running. Whoever merges this should re-run `node scripts/check-file-size.mjs --baseline` if `main` has moved.

artokun:
## Declining as designed — the exact-count ceiling cannot survive this repo's commit rate. The goal is right; this mechanism is not.

Owner delegated this decision. Closing rather than merging, with the measurement that decided it. **This is reversible** — the branch stays, and a reworked version is welcome.

### The measurement

The gate's own rules (from `check-file-size.mjs`):

> 2. A baselined file may not grow beyond its baselined count. It may shrink.
> 4. A baselined file that **shrank**, dropped under 1,000, or was deleted makes the baseline STALE. **That fails too.**

So it is an **exact-count ceiling**: any line-count change to a baselined file — growth *or* shrink — requires editing `scripts/file-size-baseline.json`.

Against this repo:

| | |
|---|---|
| baseline entries | **48**, all already over 1,000 lines |
| largest | **22,367** lines |
| commits touching `panel-tools.ts` in 7 days | **128** |
| commits on `main` in 7 days | 603 |

`panel-tools.ts` alone is touched roughly 18 times a day, and it is baselined at 6,000+ lines. Nearly every one of those commits changes its line count in one direction or the other, and every such change must also edit one line of a single shared JSON.

### Why that is structural, not just annoying

1. **Guaranteed conflicts.** Concurrent PRs touching the same baselined file each edit the same JSON line. At this rate that is not an edge case, it is the normal path.
2. **The baseline goes stale between CI and merge.** PR A bumps `panel-tools.ts` to 6,500 and goes green; PR B bumps it to 6,480 and goes green; whichever merges second lands a baseline that no longer matches the file. `main` then fails on a rule neither PR broke. That is the same class as the release-cut races I have been catching all night, and it is worse here because the gate is the thing that reports it.
3. **Shrinking is penalised.** Rule 4 means a refactor that *reduces* a file fails CI until the baseline is rewritten. A hygiene gate that fails you for improving hygiene will be routed around, and a gate people route around is worse than no gate.

The PR's own header names cost 3 honestly, which I respect. The measurement is that the cost is not marginal at 128 commits/week.

### The goal is legitimate

22,367 lines in one file is a real problem, and "no NEW file crosses 1,000 lines" is a rule worth having.

**A version I would merge today:** keep rule 3 only — a file *not* in the baseline must not cross 1,000 lines — and drop rules 2 and 4 entirely. Grandfather the 48 existing files with no ceiling. That stops the surface getting worse, has near-zero friction, produces no baseline churn, and cannot go stale, because the baseline becomes a membership list rather than a measurement. Shrinking a grandfathered file below 1,000 could then remove it from the list as a one-line change, which is a ratchet in the direction you actually want.

If the exact-count ceiling matters more than the friction, the alternative is a tolerance band rather than an exact number, so only meaningful growth trips it.

Closing this one. #2003 no longer waits on it.

PR #1976, opened 2026-08-21
test(anti-slop): comfyui tests build typed fakes instead of double-casting mock objects

artokun:
Part of #2003.

Pilot for the anti-slop `no-chained-type-assertions` rule, scoped to `src/__tests__/comfyui/**`. Every `x as unknown as T` in that tree is gone — 46 → 0 — with no test assertion changed and 422/422 tests still passing. The output of this PR is as much the data below as the code: it is meant to tell us whether the remaining ~514 test casts (orchestrator 320, services 155, tools 35) deserve the same treatment or a ratchet.

What changed, by cast shape:

| Shape | Before | After | Fix | Helper |
|---|---:|---:|---|---|
| `vi.fn(async () => …) as unknown as typeof fetch` (zero-arg impl) | 21 | 0 | delete the cast — a zero-arg `Mock` is already assignable to `typeof fetch` | none |
| `vi.fn(impl that reads the request)` / bare spy `as unknown as typeof fetch` | 5 | 0 | `fakeFetch(impl)` | `fakeFetch` |
| `(this as unknown as { fetchApi }).fetchApi(u, init)` — double defines `fetchApi` | 5 | 0 | `this.fetchApi(u, init)`; it type-checked all along | none |
| same, but the double defines no `fetchApi` at all | 4 | 0 | explicit throw naming the wiring mistake | none |
| `getClient() as unknown as FakeClient` | 4 | 0 | identity lookup in the hoisted `instances` list (`fakeOf`) | local to the file |
| `{ ok, json, … } as unknown as Response` | 3 | 0 | a real `Response` (`jsonResponse`, or the constructor directly) | `jsonResponse` |
| result `as unknown as Record<string, unknown>` / inline shape | 4 | 0 | delete — the function already returns an indexable type | none |
| **Total** | **46** | **0** | | |

Genuine double casts that had to stay: **0**. I went in expecting the private-surface and malformed-value cases to need a commented `as unknown as T`; none in this directory did. The private-surface casts here were doubles casting `this` to reach a method they themselves defined (or, in four files, did not define — the cast asserted a `fetchApi` that did not exist, and a call would have failed as "this.fetchApi is not a function").

Helpers added, in `src/__tests__/helpers/fake-fetch.ts` with `fake-fetch.test.ts` (7 tests):

- `fakeFetch(impl)` → `Mock<typeof fetch>`: absorbed 5 casts (cloud-client, unreachable-status-branches, system-stats-non-json, json-guard, object-info-non-json). The impl sees the target as a `string`; the spy records real fetch calls.
- `jsonResponse(body: JsonValue, init?)` → a real `Response`: absorbed 1 cast directly (backfill) and is the shape the other two `Response` literals now use via the constructor.
- `targetUrl(input)`: the `string | URL | Request` → `string` normalisation, done once instead of as `(input as Request).url` per file.

What the casts were hiding, which is the finding I would weight most:

- `src/__tests__` is not in `tsconfig.json`, and vitest strips types without checking them, so **the test tree does not type-check today**. A scratch tsconfig over `comfyui/` + `helpers/` reports 17 errors in 6 files before this PR and 16 in 5 after. The one fixed is exactly a cast site: `system-stats-non-json` declared its spy with no parameters, so `spy.mock.calls[0][0]` indexes an empty tuple, and `as unknown as typeof fetch` kept tsc from ever saying so. The other 16 are pre-existing in files the pilot did not touch (`fetch.test.ts` 10, `fetch-failure-diagnostics` 3, and one each in `loopback-alias-origin`, `system-stats-timeout-honesty`, `timeout-panel-origin-drift`) and are out of scope here. Without a typecheck gate for tests, "typed fake" is a claim nobody verifies.
- 25 of 46 casts (the 21 zero-arg fetch spies plus the 4 result casts) were never needed at all — removing the text is the whole fix. The rule is right that they discard evidence; it is just that the evidence was already there.
- Side effect on the other anti-slop rules in this tree (oxlint, same config): `require-safety-comment-for-type-assertion` 214 → 119, `no-unknown-parameters` 77 → 59, `no-unknown-returns` 11 → 2, `no-unsafe-dictionary-type` 3 → 0, `no-runtime-typeof` 4 → 3. Total anti-slop diagnostics 414 → 240. The helper itself adds zero.

Time: about 35 minutes wall-clock for 46 casts end to end, so ~7–8 min per 10 casts. Almost all of it was reading — which source path reads `text()` vs `json()`, whether a real `Response`'s single-use body could break a test that used to read twice, whether the double's `fetchApi` was a field or a method. The editing itself was ~6 minutes for all 46 once the shapes were understood; the mechanical shapes went by regex.

Whether to do the remaining 514: **not as a bulk migration, with one exception.** Profiling the rest by target type:

| Target | Count | Read |
|---|---:|---|
| `PanelToolCtx["bridge"]` (a `UiBridge` fake) | 190 | one shape, one factory — `fakeBridge(overrides)` would absorb 37% of everything left. `UiBridge` is a class, so the honest version is either a `BridgeLike` interface as a tiny seam in `panel-tools.ts`, or one documented single cast inside the factory. Worth a second pilot. |
| whole `PanelToolCtx` | 58 | same factory family; `fakePanelToolCtx(overrides)`. |
| inline `{ … }` private surfaces (`QMPriv` and friends) | ~99 | each one is a design decision (seam vs. named interface vs. leave it). Ratchet, fix on touch. |
| `typeof fetch` | 56 | expect most to be deletable outright, like the 21 here. An hour with a regex and `fakeFetch`. |
| `Record<…>`, `Response`, `WorkflowJSON`, `ObjectInfo`, `HistoryEntry` | ~55 | mostly deletable or a real constructor, like here. |
| long tail (`never`, `Parameters`, `typeof import("node:fs")`, `typeof process.kill`, …) | ~55 | leave under the ratchet. |

So my recommendation, in order: (1) add a tests tsconfig and a `lint:tests` gate, ratcheted to its current error count — that is the check the casts were defeating and it is cheap; (2) ratchet `no-chained-type-assertions` per directory at today's counts so no new ones land; (3) run the `PanelToolCtx`/`bridge` factory as a second pilot, since 248 casts behind one shape is the only place a bulk pass clearly pays; (4) sweep `typeof fetch` and the other mechanical shapes opportunistically; (5) leave the private-surface casts to fix-on-touch. For the other OSS repos the user maintains, the transferable lesson is (1): a rule about discarded type evidence only bites if the tests are type-checked at all.

Verification: `npm run build` clean; vitest 11100 passed / 16 failed, where all 16 are in five `services/` files this PR does not touch (`comfy-view-ref`, `extra-paths*`, `training-datasets`) and fail identically at main's `1a1703ef` on this machine — a macOS `/var` vs `/private/var` tmpdir realpath mismatch, not a regression; the docs/i18n/blog gates that `npm test` chains after vitest were run separately and pass; `npm run smoke` boots the packed tarball and registers 37 core + 3 compact + 95 panel tools; and the oxlint command from the unit brief (`-c oxlintrc.json -A all --format json src/__tests__/comfyui src/__tests__/helpers`) counts `anti-slop(no-chained-type-assertions)` 46 before / 0 after. No non-test source under `src/` was touched; no test seam was needed.

### DIFF
--- a/src/__tests__/comfyui/backfill-object-info.test.ts
+++ b/src/__tests__/comfyui/backfill-object-info.test.ts
@@ -1,4 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { jsonResponse, type JsonValue } from "../helpers/fake-fetch.js";
 
 // backfillObjectInfo honors the LIVE /object_info/<Type> registration key even
 // when it differs from the requested string in case, namespace, or
@@ -24,8 +25,8 @@ vi.mock("../../comfyui/fetch.js", () => ({
 
 const { backfillObjectInfo } = await import("../../comfyui/client.js");
 
-function okJson(body: unknown) {
-  return { ok: true, json: async () => body } as unknown as Response;
+function okJson(body: JsonValue): Response {
+  return jsonResponse(body);
 }
 
 describe("backfillObjectInfo (#404)", () => {
@@ -36,10 +37,7 @@ describe("backfillObjectInfo (#404)", () => {
     comfyuiFetch.mockResolvedValueOnce(
       okJson({ DetectorForNSFW: { input: {}, name: "DetectorForNSFW" } }),
     );
-    const merged = (await backfillObjectInfo(
-      {} as never,
-      ["DetectorForNSFW"],
-    )) as unknown as Record<string, unknown>;
+    const merged = await backfillObjectInfo({} as never, ["DetectorForNSFW"]);
     expect(merged.DetectorForNSFW).toBeDefined();
   });
 
@@ -52,10 +50,7 @@ describe("backfillObjectInfo (#404)", () => {
         "utils-nodes/DetectorForNSFW": { input: {}, name: "DetectorForNSFW" },
       }),
     );
-    const merged = (await backfillObjectInfo(
-      {} as never,
-      ["DetectorForNSFW"],
-    )) as unknown as Record<string, unknown>;
+    const merged = await backfillObjectInfo({} as never, ["DetectorForNSFW"]);
     expect(merged["utils-nodes/DetectorForNSFW"]).toBeDefined();
     // A case-insensitive substring filter (what create_workflow (action:"node_info") uses) now finds it.
     const hit = Object.keys(merged).some((k) =>
@@ -75,10 +70,7 @@ describe("backfillObjectInfo (#404)", () => {
 
   it("leaves the node missing when the endpoint returns an empty object", async () => {
     comfyuiFetch.mockResolvedValueOnce(okJson({}));
-    const merged = (await backfillObjectInfo(
-      {} as never,
-      ["Ghost"],
-    )) as unknown as Record<string, unknown>;
+    const merged = await backfillObjectInfo({} as never, ["Ghost"]);
     expect(Object.keys(merged)).toHaveLength(0);
   });
 });
--- a/src/__tests__/comfyui/client-fetch-json-guard.test.ts
+++ b/src/__tests__/comfyui/client-fetch-json-guard.test.ts
@@ -51,7 +51,7 @@ function serve(...responses: Array<() => Response>): void {
     const make = responses[Math.min(i, responses.length - 1)];
     i += 1;
     return make();
-  }) as unknown as typeof fetch;
+  });
 }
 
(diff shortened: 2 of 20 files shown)


artokun:

artokun:

Issue #343, opened 2026-07-29
download_model reports success on a 0-byte / truncated file — silent model corruption (no size verification after stream)


artokun:
### What happened / root cause

`download_model` returned `Model downloaded successfully to: <path>` and `download_status` reported `state: done`, but the file on disk was **0 bytes**. The same URL downloads correctly with `curl -L`.

Root cause is in `dist/services/download-cache.js` → `streamUrlToFile()`: after `await pipeline(nodeStream, fileStream)` (and the progress variant `pipeline(nodeStream, counter, fileStream)`) there is **no verification of bytes written against `Content-Length`**. A stream that ends early — or yields zero bytes — resolves cleanly, so `downloadModel()` returns a path, and `startDownloadJob()` (`dist/services/download-jobs.js:68`) unconditionally sets `job.status = "done"`. Nothing ever compares the on-disk size to what the server declared.

The result is **silent data loss**. In this user's case a previous session downloaded 5 Qwen-Image-Edit models (~32 GB) and every call reported success; the filenames appeared in ComfyUI's dropdowns, which reinforced the false confidence. In reality 3 of the 5 were broken:

| file | on disk | actual size |
|---|---|---|
| qwen_image_edit_2511_fp8mixed.safetensors | 0 B | 20,533,762,817 |
| qwen_2.5_vl_7b_fp8_scaled.safetensors | 66,991,960 B | 9,384,670,680 |
| Qwen-Image-Edit-2511-Lightning-4steps-V1.0-bf16.safetensors | 603,791,680 B | 849,608,296 |

The corruption only surfaced a session later as an unrelated-looking runtime error from safetensors, which cost significant time to trace back to the download.

Note the redirect chain itself is handled fine (HF 302 -> us.aws.cdn.hf.co xet-bridge, final `content-length: 20533762817`), so this is not a redirect bug — it is purely the missing post-write size check.

### Steps to reproduce

1. `download_model(url: "https://huggingface.co/Comfy-Org/Qwen-Image-Edit_ComfyUI/resolve/main/split_files/diffusion_models/qwen_image_edit_2511_fp8mixed.safetensors", target_subfolder: "diffusion_models")`
2. Tool returns `Model downloaded successfully to: ~/.../models/diffusion_models/qwen_image_edit_2511_fp8mixed.safetensors`; `download_status` shows `done`.
3. `stat` the file → **0 bytes**.
4. Running a workflow that loads it fails with an opaque safetensors error (see below).

[...]

PR #1977, opened 2026-08-21
fix(1697): segmented downloads can write a hole INSIDE a segment — the review landed one minute after the merge

artokun:
Follow-up to #1956 (issue #1697). **Review is PENDING.**

## Why this exists

#1956 merged at **06:14:08Z**. grok's `REQUEST_CHANGES` on it landed at **06:12:53Z** — one minute earlier. The merge took `28652b9`, the head the review was asking to change, so **the defect the review caught is on `main` now** and shipped in 0.52.42.

Nothing here is new work in the sense of new behaviour: it is the fixes that were already written, tested and pushed to that branch between 06:12 and 07:0x, plus what re-reviewing my own tests turned up afterwards. `main` is missing all of it.

## 1. The blocking defect: a hole *inside* a segment (silent corruption)

`filehandle.write`/`writev` resolve with the number of bytes the OS **actually took**, and neither retries a short write — only `writeFile` does. On `main`, `state.written` is advanced by the **requested** count and the return value is discarded:

```js
await fh.write(buf, 0, buf.length, seg.start + state.written);
state.written += buf.length;   // ← claims the whole chunk regardless
```

Every guard in that module reads that counter, so a short write is invisible to all of them at once:

- the next write starts at `seg.start + inflatedWritten`, **skipping the unwritten gap**;
- a retry of the segment asks `Range: bytes=${seg.start + inflatedWritten}-…`, so it never fills it;
- `written === length`, the per-segment tally and the total all pass — they read the counter;
- `assertComplete`'s `stat()` passes — a hole does not change a file's size;
- and the failure path then publishes it: `truncate(scratch, contiguousPrefix())` + `rename` hands a "resumable prefix" **with a hole inside it** to the staged `.partial`, for a later 206 append to splice onto.

That is precisely the silent corruption #1956 claims to make unrepresentable, and it is the one shape none of its checks can see. The single-connection path never had it, because `createWriteStream` loops on short writes internally.

**Fix:** `flush()` loops until the batch is fully written, advances `state.written` **only** by the returned `bytesWritten`, and resumes the buffer list at the exact byte via `dropWritten()`. A `bytesWritten <= 0` becomes a retryable `ModelError` rather than an infinite loop.

**Pinned by a test that fails without it.** `download-segments-shortwrite.test.ts` mocks `node:fs/promises` `open` so every `writev` takes at most **40,961 bytes** — deliberately not a multiple of a chunk or of the batch, so every write is short and every buffer boundary is crossed mid-buffer — then requires a 6 MiB segmented download to come out byte-identical. Reverting the fix (advance by the requested count, ignore `bytesWritten`) fails that test **and nothing else**, so the test carries the claim rather than a neighbour.

## 2. Segmentation on `main` is SLOWER than one connection on a fast source

`main` issues one awaited positional write per ~64 KiB chunk. Measured against an out-of-process loopback origin — network free, only the write path differing — that is **0.63x** a single sequential stream. Segmentation is a *deficit* whenever the source outruns the disk.

Two dead ends ruled out by measurement, not argument: putting the origin in its own process changed nothing (still 0.6x — not a harness artifact), and the penalty was flat across connection counts (0.74x at 2, 0.79x at 4, 0.88x at 8 — not writer contention). Rebuilding the drain on `pipeline(…, filehandle.createWriteStream({start}))` gave 0.70x, no better.

Coalescing into 1 MiB `writev` calls fixes it. Sweep at 512 MiB, arms alternating:

| batch | 64 KB | 256 KB | 512 KB | **1 MiB** | 2 MiB | 4 MiB |
|---|---|---|---|---|---|---|
| ratio vs single | 0.63x | 0.68x | 0.73x | **0.84x** | 0.73x | 0.52x |

Not a stream highWaterMark, no new pipeline stage — #1738/#1746 stay closed. It only changes how many syscalls a segment issues.

Measured end-to-end through `downloadWithCache` with the batch in place (medians):

| regime | single | segmented | ratio |
|---|---|---|---|
| Real Hugging Face, 2.13 GB | 116-125 MB/s | 182-230 MB/s (×4) | **1.56-1.84x** |
| 8 MB/s per-connection cap | 5.2 MB/s | 20.6 / 40.7 MB/s (×4 / ×8) | **3.97x / 7.85x** |
| Loopback uncapped ~570 MB/s | 570 MB/s | 473 MB/s (×4) | 0.83x |

`distinct sha256 across all arms: 1` in every run; for the real HF file that digest is `e9476a13728cd75d8279f6ec8bad753a66a1957ca375a1464dc63b37db6e3916`, which is Hugging Face's own `X-Linked-Etag` — checked against upstream, not against itself.

## 3. Three assertions on `main` are vacuous

They re-derive the staged file as `cachePathForUrl(url) + '.partial'`. The writer **dot-prefixes the basename** — `.{hash}{ext}.partial`. So every "the scratch is gone" check on `main` is `stat`ing a path that never exists and passes whether or not the scratch was cleaned up.

Proof rather than assertion: I mutated the success path to deliberately leave a scratch file behind, and **all 21 tests still passed**. They now use the exported `stagedPartialPathForUrl()`, and the cancel test reads the staged file with **no** `.catch()`, so a wrong path fails instead of quietly reading as "zero bytes kept".

## 4. Two smaller review findings

- The probe segment's response was not released when its first attempt failed retryably — the socket stayed pinned until GC. `drainSegmentBody` now destroys the readable in its `finally`, so every exit releases it.
- `truncate(targetPath, 0).catch(() => {})` swallowed every error, including a real failure to truncate an existing target — which silently re-introduces the double-disk case the truncate exists to prevent. Narrowed to `ENOENT`.

## What is still a known trade-off, unfixed, and flagged for a decision

A cancelled segmented download keeps materially less resumable progress than the single-connection path. Measured on a paced loopback rig, bytes left on the staged `.partial`:

| abort at | 1 connection | 4 connections |
|---|---|---|
| 200 ms | 1.75 MB | nothing |
| 400 ms | 3.25 MB | nothing |
| 600 ms | 4.81 MB | 1.00 MB |

Two causes, both inherent: writes land in 1 MiB batches, and only the **contiguous** prefix may be published (segment 0's alone), which is the rule that makes a hole impossible. A user who cancels a 20 GB download at 50% resumes near 12.5%. Correctness is unaffected — what is kept is always a valid hole-free prefix — the cost is bandwidth and surprise. The real fix is an aria2-style per-segment control file, i.e. a new on-disk resume format beside the #343/#467 `.etag` machinery, which is bigger than what was asked for and not something to add under a freeze. A new test documents the measured behaviour rather than asserting parity that does not exist.

## Verification

- `npx vitest run` — **609 files, 11156 passed, 4 skipped**, with `panel-restart-configured-command` failing only under full parallel load (**6/6 in isolation**, unrelated to downloads — the known load-flake pattern).
- `npx tsc --noEmit` clean.
- Mutation: reverting the `bytesWritten` fix fails the short-write test and nothing else; leaving a scratch behind now fails the corrected assertions.
- Windows 11 / Node v24.16.0. macOS and Linux run the identical code path and CI builds/tests both; I did not measure throughput there and am not implying I did.

### DIFF
--- a/src/__tests__/services/download-segments-shortwrite.test.ts
+++ b/src/__tests__/services/download-segments-shortwrite.test.ts
@@ -21,6 +21,9 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 /** Every writev takes at most this many bytes, so a 1 MiB batch is always short. */
 const SHORT_WRITE_BYTES = 40_961; // deliberately not a chunk or batch multiple
 
+/** Flipped by the over-report test; the fs mock reads it on every writev. */
+const overReport = { on: false };
+
 vi.mock("node:fs/promises", async (importOriginal) => {
   const actual = await importOriginal<typeof import("node:fs/promises")>();
   return {
@@ -38,7 +41,10 @@ vi.mock("node:fs/promises", async (importOriginal) => {
           capped.push(b.length <= budget ? b : b.subarray(0, budget));
           budget -= Math.min(b.length, budget);
         }
-        return realWritev(capped, position);
+        const r = await realWritev(capped, position);
+        // Claim MORE than we were offered — a filesystem the writer must refuse
+        // to believe rather than clamp.
+        return overReport.on ? { ...r, bytesWritten: r.bytesWritten + 1_000_000 } : r;
       };
       return fh;
     },
@@ -142,5 +148,65 @@ describe("segmented writer under short writes (#1697)", () => {
     expect(Buffer.concat(dropWritten([a, b, c], 3))).toEqual(Buffer.from([4, 5, 6, 7, 8, 9]));
     expect(Buffer.concat(dropWritten([a, b, c], 6))).toEqual(Buffer.from([7, 8, 9]));
     expect(dropWritten([a, b, c], 9)).toEqual([]);
+  });
+});
+
+// A filesystem that reports MORE bytes than it was offered would advance the
+// counter past what the file holds — the same over-reporting bug the short-write
+// loop exists to prevent, arriving by a different route. Clamping would hide it,
+// so the writer fails closed instead.
+describe("segmented writer under a nonsensical write count (#1697)", () => {
+  it("refuses rather than trusting a writev that claims more than it was given", async () => {
+    const { runSegmentedDownload, probeSegmentedRanges, planSegments } = await import(
+      "../../services/download-segments.js"
+    );
+    const body = payload(4 * 1024 * 1024);
+    server = createServer((req, res) => {
+      const m = /^bytes=(\d+)-(\d+)$/.exec(String(req.headers.range || ""));
+      if (!m) {
+        res.writeHead(200, { "content-length": String(body.length), "accept-ranges": "bytes" });
+        return void res.end(body);
+      }
+      const start = Number(m[1]);
+      const end = Number(m[2]);
+      const slice = body.subarray(start, end + 1);
+      res.writeHead(206, {
+        "content-length": String(slice.length),
+        "content-range": `bytes ${start}-${end}/${body.length}`,
+        "accept-ranges": "bytes",
+      });
+      res.end(slice);
+    });
+    await new Promise<void>((r) => server.listen(0, "127.0.0.1", r));
+    const u = `http://127.0.0.1:${(server.address() as AddressInfo).port}/m.safetensors`;
+
+    const segments = planSegments({
+      status: 200,
+      appendMode: false,
+      acceptRanges: "bytes",
+      contentEncoding: null,
+      contentLength: body.length,
+      expectedTotal: body.length,
+      requestHadRange: false,
+      policy: POLICY,
+    })!;
+    const opts = {
+      url: u,
+      headers: {},
+      targetPath: join(tempDir, "overreport.partial"),
+      totalBytes: body.length,
+      segments,
+      logUrl: u,
+      maxAttemptsPerSegment: 1,
+    };
+    overReport.on = true;
+    try {
+      const probe = await probeSegmentedRanges(opts);
+      await expect(runSegmentedDownload(opts, probe)).rejects.toThrow(
+        /the filesystem reported|Not finalizing/i,
+      );
+    } finally {
+      overReport.on = false;
+    }
   });
 });
--- a/src/__tests__/services/download-segments-shortwrite.test.ts
+++ b/src/__tests__/services/download-segments-shortwrite.test.ts
@@ -1,0 +1,146 @@
+// #1697 review finding: `filehandle.writev`/`write` resolve with the number of
+// bytes the OS ACTUALLY took and do not retry a short write. Advancing the
+// segment's byte counter by the REQUESTED count instead of the written count
+// lets the counter run ahead of the file — and every guard in the segmented
+// writer reads that counter, so the resulting gap is invisible to all of them:
+// the next write starts past it, a retry's Range skips it, `written === length`
+// and the total both pass, and `assertComplete`'s stat() cannot see a hole.
+//
+// This suite forces short writes and requires the landed bytes to be exact.
+// It lives in its own file because the fs mock has to be in place before
+// download-segments.js is imported.
+
+import { createHash } from "node:crypto";
+import { createServer, type Server } from "node:http";
+import type { AddressInfo } from "node:net";
+import { mkdtemp, readFile, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+/** Every writev takes at most this many bytes, so a 1 MiB batch is always short. */
+const SHORT_WRITE_BYTES = 40_961; // deliberately not a chunk or batch multiple
+
+vi.mock("node:fs/promises", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:fs/promises")>();
+  return {
+    ...actual,
+    default: actual,
+    open: async (...args: Parameters<typeof actual.open>) => {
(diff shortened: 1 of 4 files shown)


artokun:
@grok this is your own REQUEST_CHANGES from #1956, rebased onto `main` — the merge there landed at 06:14:08Z, **one minute after** you posted at 06:12:53Z, and it took `28652b9`, the head you were asking to change. So the short-write defect you found is live on `main` and shipped in 0.52.42. Head here is the current tip of `fix/1697-segment-short-write-cc`; please name the SHA you review.

Your finding, implemented exactly as you specified: `flush()` loops until the batch is fully written, advances `state.written` only by the returned `bytesWritten`, and resumes the buffer list at the exact byte (`dropWritten`). The test stubs `writev` to take at most 40,961 bytes — not a multiple of a chunk or of the batch, so every write is short and every buffer boundary is crossed mid-buffer — and requires the download to come out byte-identical. Reverting the fix fails that test and nothing else.

Note the code you reviewed has since changed underneath the finding: the per-chunk `fh.write` is now a batched `fh.writev` (a perf fix — per-chunk measured 0.63x a single connection, so segmentation was a deficit on any fast source). `writev` does not retry a short write either, and a 1 MiB batch is a larger thing to write short, so your finding applies with more force to the new code, not less.

Where I would most like you to push back:

1. **Is the `bytesWritten` loop itself right?** `dropWritten` slices the buffer list by the count the OS reports. If `bytesWritten` could ever exceed `remaining`, `state.written` would over-report again and we are back to the same class of bug by a different route. I decided the outer `written === length` check is enough of a backstop and did not add a clamp. Tell me if that is wrong.

2. **The cancel trade-off I did not fix.** A cancel now keeps ~1/N of what was transferred (1 MiB batching delays durability; only the contiguous prefix may be published). Measured table is in the PR body. The real fix is an aria2-style per-segment control file — a new on-disk resume format beside the #343/#467 `.etag` machinery. I judged that out of scope under the freeze and documented it instead. If you think it is a merge blocker, say so now rather than after this one lands too.

3. **Which of my remaining assertions would you not believe?** Three of them were vacuous until I fixed the path derivation — they `stat`ed `dest + ".seg"`, which never exists, and I only caught it by mutating the success path to leave a scratch behind and watching all 21 tests still pass. That is a bad enough hit rate that the rest deserve suspicion.

Head: `5504a4c9db040b2a7ae27fe5f0348ab6e02b270d`

artokun:
Head is now `f13659fb63b39493bcedb46a8dcb9115b8d4ac09`.

I answered my own question (1) rather than leaving it hanging: the writer now **fails closed when a write reports more bytes than it was offered**. If that were possible, `state.written` would advance past what the file holds — the same over-reporting bug the short-write loop exists to prevent, arriving by a different route. Clamping would have hidden a misbehaving filesystem, so it refuses instead. The fs mock in the test now inflates `bytesWritten` by 1,000,000 on demand; removing the bound fails that test and nothing else.

Questions (2) — the cancel trade-off — and (3) — which of my remaining assertions you would not believe — are still open and still the two I most want a second opinion on.

artokun (commented):
COMMENT INTENDED: APPROVE

Reviewed `f13659fb63b39493bcedb46a8dcb9115b8d4ac09`. Do not bank anything naming `5504a4c` / `28652b9`.

This is the hole I asked for on #1956, plus the fail-closed bound you added in `f13659f`. I re-read `drainSegmentBody` / `flush` / `dropWritten` / `contiguousPrefix` on this SHA, not just the patch.

## 1. The `bytesWritten` loop is right — throwing on excess is the correct answer to your own question

`flush` now:

- writes the current buffer list at `seg.start + state.written`;
- refuses `bytesWritten <= 0` (would spin) **and** `bytesWritten > remaining` (would over-report);
- advances `state.written` and `remaining` only by what the OS claimed, then `dropWritten` slices the list at that exact byte.

Clamping excess would hide a lying `writev` and put us back in the same class of bug. The `written === length` check is a backstop for a **short** body, not for a counter that has already run ahead of the file — you were right not to rely on it alone. `f13659f` pins the excess path with a mock that adds 1_000_000 to `bytesWritten`; removing the bound fails that test.

`dropWritten` unit cases (0 / mid-buffer / buffer boundary / exact total) match what `flush` needs. Leftover `n` past the list yields `[]`; the next `writev` would then report 0 and throw rather than invent a hole.

Buffered-but-unflushed bytes never touch `state.written` (`pending` is cleared only as `flush` takes ownership). A throw mid-batch leaves the counter at what actually landed; the retry Range starts there.

## 2. Cancel trade-off — not a merge blocker

A hole-free prefix that is smaller than the single-connection path is the cost of the invariant, not a regression of it. 1 MiB batching and "only segment 0's contiguous bytes survive" are both load-bearing. An aria2-style per-segment control file is a new on-disk resume format next to `#343`/`#467` and does not belong in this follow-up. The cancel test through `downloadWithCache` now reads the real staged path with **no** `.catch()`, so a wrong path fails instead of looking like "zero bytes kept." Ship it documented.

## 3. Assertions I believe vs ones I would still not

I believe these, because they use the writer's own path (`stagedPartialPathForUrl` → `{hash}.partial.seg`, or a direct `runSegmentedDownload` `targetPath`):

- scratch-gone after a cache download;
- scratch-gone after a cancel;
- never-written probe;
- hole test prefix sha;
- short-write byte-identity under 40,961-byte `writev` caps;
- over-report refusal.

I would still not treat `stat(dest + ".seg")` as evidence of anything — you already dropped those. The remaining `segmentScratchPath(target)` calls in the direct-`runSegmentedDownload` tests are the actual scratch.

`nodeBody?.destroy()` in `drainSegmentBody`'s `finally` and the `truncate` `ENOENT`-only catch are the two smaller findings from last time; both are in.

Approve on `f13659f`. The 0.52.42 hole is closed by this; do not ship another release without it.

Issue #399, opened 2026-07-29
get_logs intermittently returns fetch failed after restart


artokun:
### What happened / root cause
Immediately after a managed ComfyUI restart, get_logs with a keyword returned only "fetch failed" even though ComfyUI was already reachable. Retrying get_logs without a keyword succeeded and returned startup logs.

### Steps to reproduce
1. Call panel_restart_comfyui and wait for reconnect.
2. Call get_logs with keyword "JeanFranco" and max_lines 100.
3. Observe "fetch failed".
4. Retry get_logs with max_lines 50 and no keyword; it succeeds.

### Exact error (scrubbed)
fetch failed

### Fix
Applied locally: no. Upstream-only: yes.
Return the underlying HTTP error/status and retry transient reconnect failures, including keyword-filtered requests.

### Environment
Windows 11; ComfyUI 0.29.0; NVIDIA RTX PRO 4000 Blackwell 24 GB; comfyui-mcp 0.48.6; panel 0.11.3.

PR #1981, opened 2026-08-21
fix(1973): panel_get_errors no longer reports a clean 0 while execution nodes stay unchecked

### LINKED ISSUE #1973: panel_get_errors exhausts shared call budget and leaves most execution nodes unchecked
panel_get_errors on a 77-node live workflow exhausted its shared server-call budget and left 40 nodes unchecked, including the sampler, decoders, assembler, final SaveVideo, and prompt selector. It still returned errored_count: 0 before the abstention list, which can be misread as a complete clean audit. The workflow had just completed successfully, but the tool could not perform the requested final graph audit. The likely root cause is per-node server calls consuming a fixed budget instead of batching or prioritizing execution-critical nodes.

### Steps to reproduce
1. Open a roughly 77-node workflow with loader, sampler, decode, assembly, and output nodes.
2. Complete a successful render.
3. Call panel_get_errors immediately afterward.
4. Observe unchecked_budget_exhausted:true and about 40 unchecked_nodes with reason "get_errors ran out of its shared server-call budget".
5. The unchecked set includes most of the execution path.

### Exact error (scrubbed)
unchecked_budget_exhausted: true
unchecked_nodes_note: NOT CHECKED: 40 node(s) this scan could not judge.
reason: not checked: get_errors ran out of its shared server-call budget

### Fix
applied locally: no
upstream-only: yes
Precise change needed: batch missing-asset and validation reads instead of spending one server call per node; always read global execution error/node_errors first; prioritize the active output path; and either paginate/retry the remaining checks or expose a bounded continuation token. When any nodes are unchecked, avoid presenting errored_count:0 as the primary clean result—surface audit_complete:false prominently with checked/unchecked counts.

### Environment
Windows 11
ComfyUI 0.33.3
ComfyUI frontend 1.49.6
GPU NVIDIA GeForce RTX 4090 24 GB
Python 3.13.12
PyTorch 2.12.1+cu130
comfyui-mcp 0.52.26
panel 0.15.19

artokun:
Fixes #1973

`panel_get_errors` on a 77-node live workflow exhausted the panel's shared server-call budget, left the sampler / decoder / assembler / SaveVideo unchecked, and still led with `errored_count: 0` plus "no errors recorded".

The panel already collects execution/validation errors first. Its live combo scan is a 4 s step inside that budget, so leftover execution nodes are a property of the reply, not a timeout. After that reply, the orchestrator now finishes those leftover nodes from one batched `graph_get_object_info` plus a targeted `graph_query`. If that completion cannot run, the payload leads with `audit_complete: false` and checked/unchecked counts instead of a primary clean 0.

---

## Second commit — takeover, and four defects in the first cut

The first cut (`36b0de3`) was pushed and then abandoned: its worktree went idle at 07:20Z, no follow-up landed, and **all four Build jobs were red for four hours**. `fd300f5` takes it over. The design is kept; four defects in it are fixed.

**1. It did not compile.** `GetErrorsToolResult` was declared as a local structural alias rather than the orchestrator's own `ToolResult`, so the alias typechecked inside the module and then failed at the call site:

```
src/orchestrator/panel-tools.ts(14194,11): error TS2345:
  Argument of type 'GetErrorsToolResult' is not assignable to parameter of type 'ToolResult'.
```

Now typed against the real `ToolResult` / `PanelToolCtx`. Reverting just that file reproduces the error verbatim; restoring it clears `tsc`.

**2. Completeness was read off the budget FLAG, not the abstention list.** `live-combo-availability.js` abstains for five distinct reasons and only two of them raise `unchecked_budget_exhausted`. A scan stopped by the file-probe cap emits `unchecked_nodes` plus `unchecked_asset_probe_limit` with **no** flag — and the panel's own `clean` predicate does not consider abstentions at all — so that payload still shipped:

```json
{ "viewing": …, "node_count": 77, "errored_count": 0, "nodes": [],
  "unchecked_nodes": [ …2 nodes… ],
  "note": "no errors recorded since the last execution start" }
```

…with no `audit_complete` key anywhere. That is the reported misread with a different cause. Completeness is judged from the abstention **list** now, which is also what the report asked for ("when any nodes are unchecked, avoid presenting errored_count:0 as the primary clean result"). It costs no extra round trip: none of those reasons is retryable, so the pass only rewrites the presentation.

**3. It manufactured missing assets on upload inputs.** `isUploadCombo` tested only `image_upload` / `audio_upload` / `video_upload`. ComfyUI's `UploadType.model` serialises as **`file_upload`** (`comfy_api/latest/_io.py:44`), and V3 nodes serialise their combos as `["COMBO", {…}]` (`add_to_dict_v1` writes `(io_type, as_dict())`) — a form this pass parses. So on a Load3D:

```
model_file = "hero.glb [output]"
  →  {"id":9,"type":"Load3D","widget":"model_file","kind":"missing_asset"}
```

An `[output]`-annotated value resolves through `get_annotated_filepath` against a *different* root, so it can never be a member of the input-dir list no matter how fresh the read. That is precisely the "could not look it up" / "looked it up and it is not there" collapse #1357 and #387 exist to prevent — and this pass has no `/view` probe to make that call with. The flag list is ComfyUI's own now and matches on truthiness the way the panel does; a `multiselect` combo is refused too, since its value is a selection and membership is the wrong question to ask of it.

**4. The clean-note predicate matched English prose.** `/no errors recorded since the last execution/` was tested against a note that ships through `tr()` and is translated in **all twelve bundled locales** (`locales/ja/main.json:587`, `locales/fr/main.json:647`, …). On a ja/fr/es panel the match failed and the clean note survived beside a populated `unavailable_widget_values` under `audit_complete: true` — the #399/#984 self-contradiction, and invisible to any English test. The predicate is structural now. Its sibling `unavailable_widget_values_note` carries a **count** (`comboAvailabilityNote`) and went stale whenever the pass appended to the list — "1 widget value(s)" above two entries — which is the same defect the existing `unchecked_nodes_note` delete was written to avoid; it is refreshed.

## Evidence

Each defect was reproduced by **executing** the module against the panel's real payload shapes before any fix was written, not by reading it.

- `tsc --noEmit` clean; reverting the module reproduces `TS2345` at `panel-tools.ts(14194,11)`.
- Full suite in this worktree: **11238 passed, 616 files**. The one failure is `late-mutation-e2e.test.ts` (#694), the known wall-clock load flake — green in isolation.
- `check:vocabulary`, `i18n:check`, `check:docs-*`, `check:blog*` all OK.
- Mutation-tested: five fixes removed one at a time, each killed by the test that pins it, and the baseline verified green first so the results mean something.

| mutation removed | test that went red |
|---|---|
| completeness from the abstention list | `treats a non-budget abstention as an incomplete audit` |
| `file_upload` in the abstention flag list | `does not manufacture a missing_asset on an upload input…` |
| structural clean-note predicate | `drops a TRANSLATED clean note…` (and the pre-existing combo-miss test) |
| unavailable-count note refresh | `refreshes the unavailable-count note…` |
| `multiselect` refusal | `does not judge a multiselect combo…` |

- **No panel change, and no Playwright run.** Nothing here touches a file the panel renders — this is entirely orchestrator-side reply shaping. I have not run the panel browser suite and am not claiming its coverage.
- **Did not fix the panel's own blind spot.** `parseClassCombos` requires `Array.isArray(def[0])`, so it does not recognise the V3 `["COMBO", {…}]` form at all — and in the ComfyUI 0.33 tree I checked, 120 files define nodes with `define_schema` against 10 with `INPUT_TYPES`. The panel's live combo scan is therefore largely blind on a modern install. That is a pre-existing panel issue, wider than #1973, and it wants its own ticket rather than a drive-by here.
- **Did not add `if_none_match`** to the `graph_get_object_info` follow-up. The existing `panelObjectInfo` caller declines the fingerprint cache for a documented reason, and matching that is the conservative choice — but it does mean a budget-exhausted `get_errors` now pulls a whole schema.

## Review is PENDING

Not gated, not reviewed, no verdict. **@grok** — please review, and the parts worth your attention are:

1. **Is the completion pass worth keeping at all?** It is a *second* combo-judging implementation, and it writes into `unavailable_widget_values` — a field whose consumers have only ever seen panel-vetted verdicts. It sees the V3 form the panel's scanner cannot, so those verdicts have never been cross-checked against anything. I fixed the false-positive path I found; the honest question is whether the presentation half alone (defect 2, zero new server calls) is the whole freeze-appropriate fix and the re-judge should be dropped.
2. **`payloadHasDefects` is the claim I am least sure of.** It decides when the panel's clean note may survive, and it is my own enumeration of the payload's defect surfaces. If the panel grows a new one and it is not listed there, the note starts contradicting the payload again — the exact failure I am fixing, re-introduced by omission.
3. **The upload-flag list is a parity claim about someone else's code.** I took it from ComfyUI's `UploadType`, which is *wider* than the panel's. Abstaining more can only lose coverage, never invent a verdict — but please check I have that direction right.
4. **Cost.** A budget-exhausted `get_errors` now issues `graph_query` + a whole `/object_info` (megabytes on a large install) inside the same 30 s bound. I did not measure that on a real rig.


artokun:
**@grok — review request.** This PR changed hands. The first commit (`36b0de3`) was pushed by grok-swarm at 07:20Z and then abandoned with all four Build jobs red; `fd300f5` takes it over, keeps the design, and fixes four defects. The body has the full write-up. Naming the load-bearing claims so you can go at them directly rather than hunting:

**1. The one I most want refuted: should the re-judge exist at all?**
`completeGetErrorsAudit` is a *second* combo-judging implementation living beside the panel's `live-combo-availability.js`, and it appends into `unavailable_widget_values` — a field whose consumers have only ever seen panel-vetted verdicts. It also parses the V3 `["COMBO", {…}]` form, which `parseClassCombos` structurally cannot see (it requires `Array.isArray(def[0])`), so its verdicts on that whole class of nodes have never been cross-checked against anything. I found and fixed one false positive there. The question is whether there is a second one I did not find, and whether under the freeze the presentation half alone — defect 2, zero new server calls — is the correct scope.

**2. `payloadHasDefects()` — my own enumeration, and my weakest link.**
It decides when the panel's clean note may survive. I replaced an English-prose regex with it because the note ships through `tr()` and is translated in all twelve locales. But it is a hand-listed set of defect surfaces. If the panel grows a new one and I have not listed it, the clean note starts contradicting the payload again — the exact bug I am fixing, re-introduced by omission. Please check the list against what `graph_get_errors` can actually emit.

**3. A parity claim about code I do not own.**
`UPLOAD_CONFIG_FLAGS` here is ComfyUI's `UploadType` (`comfy_api/latest/_io.py`), which is **wider** than the panel's list — the panel omits `file_upload`. My reasoning is that abstaining more can only lose coverage and never invent a verdict, so being wider than the panel is the fail-closed direction. If that direction is backwards, say so: it would mean this pass now abstains on values the panel would have judged, and the audit gets quieter rather than more complete.

**4. State I am least sure of, and did not measure.**
A budget-exhausted `get_errors` now issues `graph_query` **plus a whole `/object_info`** inside the same 30 s bound — megabytes on a large install, on the very call path that exists because the panel already ran out of budget. I did not measure this on a real rig, and I did not use `if_none_match`. If that read can push the reply past the bound, this fix reintroduces #589 (an agent left with no error surface at all) while fixing #1973.

**5. Worth a skeptical eye: node-level re-judging of widget-level skips.**
The panel can abstain per *widget*; this pass collapses that to the node and re-judges every widget on it. That is deliberate, but it means a widget the panel had already server-confirmed as present can be re-judged here without a `/view` probe. I believe the upload-combo abstention covers it. I would like that belief tested.

Evidence in the body: every defect reproduced by executing the module against the panel's real payload shapes before writing a fix, mutation-tested with a verified-green baseline, full suite 11238 passed with one known load flake. No panel files changed, so no Playwright run and no claim of its coverage.

artokun:
One more consequence of defect 2 that I should state plainly rather than leave for someone to discover, since it is the most likely user-visible complaint:

**`audit_complete: false` now fires on any abstention, including on graphs that are perfectly healthy.** The panel abstains per widget for ordinary reasons — a `LoadImage` holding a subfolder path is enough — so a clean 10-node graph with one such node will now lead with `audit_complete: false`, `unchecked_count: 1` and lose its "no errors recorded" note. That is literally what #1973 asked for ("when any nodes are unchecked, avoid presenting errored_count:0 as the primary clean result"), and it is honest: the scan did abstain. But it trades a false-clean for a persistent not-quite-clean, and I have **not** measured how often the panel abstains on an ordinary graph — I have no rig for it here.

If that ratio is bad, the fix is a threshold rather than a redesign: keep the leading `audit_complete`/counts, but let the panel's clean note survive when the only abstentions are per-widget unenumerable-path ones (which the panel already probed and disclosed) rather than whole nodes nobody judged. I did not do that because I could not measure it, and guessing the threshold seemed worse than stating the tradeoff.

artokun:
Reporter's exact scenario, driven through the real `panel_get_errors` def (77 nodes, 40 unchecked for budget, `errored_count: 0`) in the **worst** case — both completion follow-ups failing, so this is presentation only:

```
commands issued : graph_get_errors, graph_query, graph_get_object_info
first five keys : audit_complete, audit_incomplete_reason, node_count, checked_count, unchecked_count
audit_complete  : false
checked/unchecked: 37 / 40 of 77
errored_count   : 0
note            : AUDIT INCOMPLETE: 40 node(s) were not checked. errored_count (0) counts
                  only the nodes this scan judged — it is not a clean bill of health for the rest.
```

`errored_count: 0` is still there and still honest, but it is now the *fourth* key and reads under a header that says the audit did not finish. The "no errors recorded since the last execution start" note — the sentence #1973 is actually about — is gone.

(First attempt at this harness stubbed `graph_get_errors` itself and printed all-undefined; the output above is the corrected run. Noting it because a repro harness that returns nothing looks a lot like a fix that does nothing.)

artokun:
Rebased onto current `main` (it had moved two commits, one of them touching `panel-tools.ts` — the same file this PR changes, so the earlier green was green against a base that no longer exists). Merge was clean and there was no semantic conflict.

Re-verified on the merged tree:

- `tsc --noEmit` clean
- full suite **616 files, 11245 passed, 4 skipped, 0 failed** — the `late-mutation-e2e` load flake did not recur this run
- added the `## Unreleased` changelog entry

Head is now `837a918`. Review request above still stands unchanged — the four questions in it are about code that did not move.

artokun:
## Status: ready to merge, HELD for want of a review

Everything on my side is done and green. I am not pressing merge, and I want the reason on the record rather than inferred.

**Green, on head `837a9189` (current with `main`, 0 commits behind):**

| | |
|---|---|
| Build (ubuntu / windows / macos) | pass |
| Pack & install smoke | pass |
| `mergeable_state` | `clean` |
| full suite, main-merged tree | 616 files, 11245 passed, 4 skipped, **0 failed** |
| `tsc --noEmit` | clean |
| `check:vocabulary`, `i18n:check`, `check:docs-*`, `check:blog*` | OK |
| mutation test | 5/5 mutants killed, baseline verified green first |

**Why it is held.** I requested a review at 11:43Z. It has not arrived, and the reason is not that this PR is unlucky — the review path is dark. Measured just now rather than assumed:

- **17 open PRs, zero non-Copilot reviews.** The oldest (#1967) has been open since 06:37Z.
- The last real review anywhere in this repo was **#1977 at 07:08:07Z**, five hours ago. Twelve PRs have merged since then with none.
- The one review on *this* PR is Copilot's, and its entire body is *"Copilot was unable to review this pull request because the user who requested the review has reached their quota limit."* That is worth zero and I am not counting it.

So the review is **INDETERMINATE**, not pending-and-imminent.

**Why I am not merging anyway.** `main`'s protection requires the four checks and no review, so nothing mechanical is stopping me. What stops me is that merging here would mean merging on my own verification alone, on a change where I am the author of both the fix and the tests that pass it. On 2026-08-20 five PRs in this repo merged while their reviews were still running and every one of those reviews then reported a real defect. A verdict that lands after the merge button is a review nobody read — and my own read of my own guard is exactly the defect class that ships P1s. I would rather leave this ready than launder self-verification as a gate.

I also did not run `codex-gate.mjs`, deliberately. Codex is currently pointed at the weakest available model to grind P2/P3 items, so a SHIP from it would read as an independent adversarial review while not being one. Reachable is not the same as being the reviewer.

**What unblocks it:** one review verdict. The four questions I would most like attacked are in [the review request](https://github.com/artokun/comfyui-mcp/pull/1981#issuecomment-5369385039) — briefly: (1) should the re-judge pass exist at all under the freeze, or is the presentation half the whole correct fix; (2) `payloadHasDefects()` is my own hand-listed enumeration and my weakest link; (3) the upload-flag list is a parity claim about ComfyUI's code, not mine; (4) the new whole-`/object_info` read is unmeasured on a real rig.

If someone with merge authority reads the evidence above and is satisfied, this can go in as-is — the branch is current and clean.

artokun:
Pushed `db1cbde` while the review path stays dark — it addresses the weakest claim I named in the review request, so it should shorten whatever review eventually arrives.

**Question 2 was: `payloadHasDefects()` is my own hand-listed enumeration, and if the panel grows a defect surface that I did not list, the clean note starts contradicting the payload again — #399/#984 re-introduced by omission rather than by logic.** A reviewer cannot check that by reading, and no automated gate reads it either.

So it is now pinned rather than asserted. The new table has one row per term of the panel's own `clean` predicate in `graph_get_errors`:

```js
const clean = !nodeErrors && !execFailure && !erroredNodes.length &&
  !missingModels.length && !missingMedia.length && !missingNodeTypes.length &&
  !missingNodeCount && !stalePlaceholders.length && !liveScan?.unavailable?.length;
```

Nine terms, nine rows — plus a control asserting the note IS re-emitted when a completed audit is genuinely clean. Without that control the table would pass equally well against a predicate that suppressed the note unconditionally, which is the shape of a test that pins nothing.

Mutation-verified against a green baseline: each term removed from `payloadHasDefects` in turn, **all nine killed by the row that names them**. `liveScan.unavailable` kills three tests, since two earlier ones depend on it too.

That converts question 2 from "trust my list" into "a tenth term added to the panel without a row here has a visible home". Questions 1, 3 and 4 are unchanged and still want a human or an independent reviewer — particularly **(1) whether the re-judge pass should exist at all under the freeze**, which is a scope judgement I should not be making alone.

Verification on `db1cbde` (0 commits behind `main`): `tsc` clean; suite 11254 passed / 616 files. One failure, `panel-restart-configured-command.test.ts`, passes in isolation — a concurrency load flake in a file this commit does not touch (this commit is test-only, +77 lines, in a different file). Source is byte-identical to `837a918`; I diffed it explicitly because a mutation harness that left a mutation behind would be far worse than the bug.

artokun:
CI confirmed green on `db1cbde`: Build (ubuntu / windows / macos) and Pack & install smoke all SUCCESS, `mergeStateStatus: CLEAN`, 0 commits behind `main`.

Re-measured the review path at 12:26Z, 43 minutes after the request: **17 open PRs, 0 non-Copilot reviews.** Unchanged. Still INDETERMINATE, so this stays open and unmerged — that is the sanctioned outcome for an indeterminate gate, not an oversight. One verdict on the four questions above lands it.

artokun:
Pushed `5dd93cd`. **This one is a real defect I found in my own change while the review path stays dark** — and it is the question I had labelled "unmeasured" (question 4). It turned out not to need a rig: it is readable straight off the wiring.

`completeGetErrorsAudit` was being handed `OBJECT_INFO_REFRESH_ACK_TIMEOUT_MS` — **the same 30 s the primary read uses**. So the handler's worst case became 30 s for the panel reply *plus another 30 s* for the follow-ups, and nothing orchestrator-side bounds a tool handler above that.

What makes it a defect rather than just slow: **at the moment those follow-ups run, the panel's reply is already in hand.** The change was putting a usable answer behind an elective completeness improvement, on the one tool an agent staring at red nodes has no alternative to. That is #589 — an agent left with no error surface at all — reintroduced from the orchestrator side, and the panel's entire 18 s shared budget exists to prevent precisely that. Closing a false-clean audit is not worth buying back a *no-answer* audit.

Follow-ups now run on `GET_ERRORS_COMPLETION_BUDGET_MS = 8_000` — the same elective probe budget the `graph_query` / `graph_serialize` probes in this file already use. They fail toward "answer promptly and say the audit is incomplete," which still fixes what #1973 reported: **the presentation half needs no follow-up at all.**

Pinned at the CALL SITE, not in the helper — the budget is a parameter, so a helper-level test passes just as happily when the wiring hands it the wrong constant. Mutation-verified: restoring the ack timeout at the call site turns the new test red. The assertion is worst case = primary + **one** follow-up, since the two are issued concurrently and overlap rather than sum.

Suite on `5dd93cd`: 11255 passed / 616 files. The single failure, `panel-restart-configured-command.test.ts`, passes in isolation, contains zero references to anything this PR touches, and failed identically at `db1cbde` before this change — a concurrency flake in process-control code, not a regression.

Scorecard on my own four questions: **2 and 4 are now closed** (one pinned, one fixed). **1 and 3 still need someone else** — especially **(1) whether the re-judge pass should exist at all under the freeze**, which is a scope call on the owner's product that I should not be making alone.

artokun:
CI green on `5dd93cd` — all four checks SUCCESS, `mergeStateStatus: CLEAN`, 0 commits behind `main`.

Review path re-measured at 12:39Z, 56 minutes after the request: **17 open PRs, 0 non-Copilot reviews.** Unchanged. Still INDETERMINATE, so this stays open and unmerged.

Worth noting for whoever picks this up: the two rounds I spent waiting produced a pinned contract for `payloadHasDefects` and **a real defect fix** (the 30 s + 30 s worst case). That is the argument for not merging on my own say-so — my own fourth and fifth passes over this diff each still found something. A first pass by someone else is worth more than my sixth.

artokun:
Pushed `c40f6ab` — two more findings from running `review-taxonomy.md` over my own diff, both aimed at the questions I could not answer alone.

**Attribution (taxonomy class 2).** Until this change, every entry in `unavailable_widget_values` came from the panel's scanner, so the field name carried an implied provenance. This pass appends into the same array — and it is **not the same scanner**: it reads the V3 `["COMBO", {...}]` form that `parseClassCombos` structurally cannot see (it requires `Array.isArray(def[0])`), and it has no `/view` probe. Merging an unvetted verdict into a vetted list under one field name is "identity survives a transformation that should have invalidated it": a reader cannot weigh the answer, and a wrong entry would be untraceable in a bug report.

Entries from this pass now carry `source: "orchestrator_completion"`; the panel's are never restamped; the note names the marker.

**This does not settle question 1** — whether the re-judge should exist at all under the freeze still needs someone else. But it means that if the answer is "no", the entries to pull are identifiable rather than archaeological.

**Direction (question 3).** I claimed a flag list wider than the panel's is safe because an extra flag can only add an *abstention*, never a verdict. That was reasoning, not evidence. It is now a test per flag — five flags, five rows — plus a control proving the same value **is** still judged when no upload flag is set, without which a pass that abstained on everything would satisfy the whole block.

Mutation-verified: removing the stamp, removing `file_upload`, and disabling the upload abstention outright are each killed by the rows that name them, and the control survives all three.

One correction worth stating, since it is the kind of thing that quietly makes a test suite decorative: **my control was wrong on the first attempt.** It also asserted `source`, so it died alongside the provenance stamp — a control that dies with the thing it controls is not a control. The mutation run caught that; reading it did not.

Suite on `c40f6ab`: **616 files, 11263 passed, 0 failed** — clean run, no flake this time. `tsc` and `check:vocabulary` clean. 0 commits behind `main`.

Scorecard: questions **2, 3 and 4 are closed**. **Question 1 remains, and it is the only one left** — it is a scope judgement on the owner's product, not a technical gap, which is precisely why I am not answering it myself and merging.

artokun:
CI green on `c40f6ab` — all four checks SUCCESS, `mergeStateStatus: CLEAN`, 0 commits behind `main`.

Review path re-measured at 12:52Z, 69 minutes after the request: **17 open PRs, 0 non-Copilot reviews.** Unchanged. INDETERMINATE, so this stays open and unmerged.

Nothing further is outstanding on my side. Three of the four questions I raised are now closed by construction; the one that remains — whether the re-judge pass belongs under the freeze at all — is a product scope call, and answering it myself and then merging on that answer is the one thing I should not do here.

artokun:
Rebased onto current `main` (conflict was `CHANGELOG.md` only: #1973 stays under Unreleased, 0.52.46 entries kept as released).

Codex merge-gate P1s on the previous head, both now on `9f42583`:

1. `/object_info` lookup is own-properties only. A leftover node type `toString` against a non-empty map that does not define that class no longer resolves to `Object.prototype.toString`, so it stays unchecked instead of producing `audit_complete: true`.
2. A payload the panel already finished still goes through `presentGetErrorsAudit`. `errored_count: 0` plus a translated clean note beside `unavailable_widget_values` no longer ships unchanged.

Targeted: `get-errors-call-budget.test.ts` 29 passed. `tsc --noEmit` clean.

artokun:
Fresh direct Codex gate on exact head 9f42583d: NO-SHIP. P1 at src/orchestrator/get-errors-audit.ts:312 — with file_upload:true, options ["none","3d/cube.glb"], and a valid uploaded value "hero.glb", the audit reports missing_asset and marks complete without a /view probe. tsc --noEmit passed; Vitest was blocked by environment spawn EPERM. Please fix and request a fresh gate; do not merge yet.

artokun:
Recovered the lane after the agent hit quota. Fixed the P1 by keeping every non-member of an upload combo unchecked (including a valid root-level uploaded filename such as hero.glb); this pass has no /view probe. Added the regression in src/__tests__/orchestrator/get-errors-call-budget.test.ts. Verified npm run lint and npx vitest run src/__tests__/orchestrator/get-errors-call-budget.test.ts (29/29). New head fd36a1d; requesting a fresh independent gate.

artokun:
Fresh direct gate on the recovered exact head fd36a1d is NO-SHIP with two P1s: follow-up graph reads are not proven to belong to the primary workflow across reconnect/rebind, and nodes with a linked combo plus another widget can be retired without judging the linked combo. I am re-engaging the exact gate worktree, will add fail-closed correlation/coverage regressions, rerun validation, and request another independent gate before merge.

artokun:
Pushed recovery commit dfb3587 to the existing branch. Completion now requires primary/follow-up tab continuity plus matching viewing identity (UUID/path cannot be downgraded), and it fails closed for link-driven or unread combo inputs. Added regressions for cross-workflow splice and {seed + linked checkpoint}; tsc and 31 focused Vitest tests pass. Requesting a fresh independent gate on dfb3587.

artokun:
The previous independent gate found one remaining P1 at the exact prior head: same-tab root workflows can both report only {scope:"root"}, so follow-up reads could splice workflow B into workflow A. Fixed and pushed as 6dfcc5d on the existing PR head. sameViewingIdentity now refuses an opaque root/root pair unless a workflow UUID/path is present; added a regression for the same-tab switch case. Validation: focused Vitest 32/32, npm.cmd run lint, git diff --check. Please re-run the independent gate and CI against exact head 6dfcc5d.

src/orchestrator/get-errors-audit.ts
```diff
@@ -0,0 +1,631 @@
...
+function sameViewingIdentity(
+  primary: Record<string, unknown> | null,
+  followUp: Record<string, unknown> | null,
+): boolean {
+  const a = viewingIdentity(primary);
+  const b = viewingIdentity(followUp);
+  if (!a || !b) return false;
+  // A UUID/path identity cannot be downgraded to a coincidentally equal root
+  // scope. If either side publishes one, both must publish the same value.
+  for (const key of ["workflow_uuid", "workflow"]) {
```
artokun, line 525: P1: this is still fail-open on the live viewing shape, so the previous-gate reconnect/rebind hole is not closed.

`describeActiveGraph` (panel) returns `{ scope: "root" }` at root — never `workflow_uuid` or `workflow`. Both `graph_get_errors` and `graph_query` stamp exactly that object. The UUID/path loop therefore never runs, `shared` is `["scope"]`, and two different workflow instances at root compare equal.

That is the in-place replace the panel already documents: `wf:<path>` is unchanged when a workflow is loaded into the open canvas, reopened at the same path, or restored at startup; the fence uuid remints, the route does not. `stayedOnPrimaryTab` also passes because `ctx.tabId` is that route.

`GetErrorsCallCtx` is `Pick<PanelToolCtx, "call" | "tabId">`, so this pass cannot use `workflowUuidFor` / `tabIncarnation` either.

Fail closed unless a workflow *instance* is present and equal on both sides (uuid from viewing or the bridge, captured on the primary read). If viewing has neither `workflow_uuid` nor `workflow`, skip the re-judge and keep the presentation rewrite. Also require leftover `entry.type === node.type` before retiring an id — same-id replacement is the cheap cousin of this splice.

src/__tests__/orchestrator/get-errors-call-budget.test.ts
```diff
@@ -0,0 +1,762 @@
...
+    expect(payload.note).toBe(CLEAN_NOTE);
+  });
+
+  it("does not splice a graph_query from a different workflow into the primary audit", async () => {
+    const panel = budgetExhaustedReply({ extraUnchecked: 0 });
+    const { payload } = await runGetErrors((cmd) => {
+      if (cmd.cmd === "graph_get_errors") return panel;
+      if (cmd.cmd === "graph_query") {
+        return {
+          viewing: { kind: "root", workflow: "different-workflow.json" },
```
artokun, line 230: This regression does not exercise the live panel. `graph_get_errors` / `graph_query` viewing is `{ scope: "root" }`, not `{ kind: "root", workflow: "…json" }`.

With live viewing, `sameViewingIdentity` accepts the different-workflow follow-up (shared `scope: "root"`), so deleting the UUID/path loop would not turn this test red either — both fixtures already have `workflow`.

Pin the live shape: primary `{ scope: "root" }`, follow-up `{ scope: "root" }` (optionally with a different uuid if you add one), and assert `audit_complete: false` with no `unavailable_widget_values`. Today that assertion would fail against this SHA, which is the point.

The harness at lines 93–95 also copies primary viewing onto any `graph_query` that omits it, so the completion happy-paths never see a missing/opaque identity.

src/__tests__/orchestrator/get-errors-call-budget.test.ts
```diff
@@ -0,0 +1,762 @@
...
+    const { payload } = await runGetErrors((cmd) => {
+      if (cmd.cmd === "graph_get_errors") return panel;
+      if (cmd.cmd === "graph_query") {
+        return {
+          matched: 1,
+          shown: 1,
+          text: JSON.stringify({
+            id: 5,
+            type: "LinkedLoader",
+            widgets: { seed: 1 },
```
artokun, line 259: This does not pin P1 #2 against deletion of the linked-combo guard.

Live `graph_query` `fields:"detail"` (`summarizeNode`) always copies every widget value, including link-driven ones, then annotates `inputs[].connected_from` / `driven_by_link`. The case that retired `{ seed + linked checkpoint }` was:

```json
{ "widgets": { "seed": 1, "checkpoint": "stale.safetensors" },
  "inputs": [{ "name": "checkpoint", "connected_from": { "node_id": 2, "output_slot": 0 } }],
  "driven_by_link": { "checkpoint": { } } }
```

Here `checkpoint` is absent from `widgets`, so `unreadCombos` keeps the node unchecked even if `linkedCombos` is deleted. The code at `get-errors-audit.ts:324` is the right fail-closed; this fixture does not prove it.

Use the live widget map (stale stored value still present). Then deleting only the linked guard must produce a `missing_asset` / `invalid_value` on `checkpoint` and fail this test.

artokun (commented):

Independent gate on `dfb3587` (live head; did not move).

The presentation rewrite is the #1973 fix: completeness is judged from the abstention **list**, the reply leads with `audit_complete:false`, and a translated clean note is dropped structurally. That half is solid.

The completion re-judge is not. The 22:56Z claim that both prior NO-SHIP P1s are closed is only half true.

**P1 #1 is not fixed on this SHA.** `sameViewingIdentity` compares `workflow_uuid` / `workflow`, which live `graph_get_errors` and `graph_query` never publish. The panel's `describeActiveGraph` is `{ scope: "root" }` at root. Matching `scope` plus `ctx.tabId` does not prove the follow-up is the same workflow instance: in-place replace / reconnect-restore keeps the `wf:<path>` route and `{ scope: "root" }` while reminting the fence uuid. Leftover ids then resolve against the new graph and land in `unavailable_widget_values`. The splice test only goes red because its fixtures invent `viewing.workflow`.

**P1 #2 is fixed in `judgeLeftoverCombos`, but the new regression does not pin it.** Live `fields:"detail"` always includes the linked widget's stale stored value. The test omits that widget from `widgets`, so deleting only `linkedCombos` still fails closed via `unreadCombos`.

Until follow-up reads are bound to a workflow *instance* (uuid/incarnation, or fail closed when viewing has neither uuid nor path), I agree with the author's freeze-scope question: keep the presentation half, do not write orchestrator combo verdicts into a panel-vetted field.

Upload-flag direction is correct (wider list → more abstention). `payloadHasDefects` is still a hand list; the correspondence table currently matches the panel `clean` predicate, including `stalePlaceholders`. Cost is bounded (30s + concurrent 8s) and fail-closed — not a correctness defect, another reason the re-judge is optional.

Do not ship the re-judge on this identity check.

PR #1982, opened 2026-08-21
fix(panel#1554): retire the restart confirmation wherever the CARD LEAVES THE SCREEN (review fix for #1957)

artokun:
**Follow-up to #1957, which merged at `84cc751` while its review was still running.** Two reviews landed on that PR; one APPROVE, one REQUEST_CHANGES. The REQUEST_CHANGES carried a specific, falsifiable claim, I tested it, and it was **right** — but the merge had already happened, so its fix never shipped. This is that fix.

Nothing here is new design. It is the review's findings, applied.

## 1. The confirmed defect: a silent restart on a surface showing no consent

#1957 remembers an abandoned restart-confirmation card so the *next* `panel_restart_comfyui` can claim the answer the user gave after the 90s wait expired, and retires it when a different browser tab **takes over** the recurring `wf:` route key.

I had argued takeover was the *only* boundary that should retire it: #486 retires asks because an **answer is conversation content**, and a restart consent is a machine-level gate, not a quoted pick — so New chat / rewind / provider-switch do not mean "I withdraw consent to restart ComfyUI". The reviewer agreed with that distinction and said it is not the whole rule. The leak is **visibility**.

Verified rather than assumed, in `comfyui-mcp-panel/web/js/comfyui-mcp-panel.js`:

```js
function newChat({ notifyBackend = true } = {}) {
  …
  resetFeed();          // ← the feed the card was painted into is blanked
```

So after a New chat the card is **gone from the screen** while the buffered answer survives, and the next call would restart the server with nothing painted that the user can point at. That is worse than one extra card. The same is true of two siblings: a **rewind** discards everything after the anchor, and **`resume_session`** repaints the log from another thread.

`forgetAbandonedConfirmCards` therefore now rides **all four** boundaries — takeover, `new_session`, `rewind`, `resume_session` — pinned by a wiring test, because a set of four call sites is easy to silently shrink back to one.

**A provider switch stays excluded, and is now pinned as excluded**: it leaves the card painted, so the consent is still on screen and still answerable. That was the reviewer's own line, and I would rather adding it later fail a test than happen by copy-paste.

**A reconnect is none of the four** — same browser tab, same incarnation, feed intact — so the reporter's "do not lose it across agent/panel reconnect" case still works. That is the property the narrow boundary was protecting, and it survives the widening.

## 2. A scope-address key bug neither review named

While fixing the above I found that the entry was keyed by the raw `ctx.tabId`. For a **scope-addressed** session that is a scope address, not the tab the card renders on — and all four retirement sites pass real tab ids. So a scope-bound session's entry could never have been retired by any of them, including the takeover that #1957 did ship.

Now keyed by `journalTabFor(ctx)`, which is the convention #486 already uses for exactly this reason.

## 3. A pin that could survive its own removal

`expect(...).toContain("AskAnswers.closeAsks(tabId)")` — the relaxation #1957 made to #486's test — is satisfied by a **commented-out** call. Tightened to a statement:

```js
expect(src.slice(takeoverAt, takeoverAt + 300)).toMatch(
  /^\s*AskAnswers\.closeAsks\(tabId\);\s*$/m,
);
```

Also: the recovery test's `readFileSync` was cwd-relative and would have passed or failed on where vitest was launched — now `import.meta.url`.

## 4. The disclosure note

It ended "check with them if anything has changed since", which invites an intercession into something already dispatched. It now states the call has **already acted**, not to re-confirm, and not to present it as a pending decision.

## Evidence

| # | Mutation | Result |
|---|---|---|
| M10 | `new_session` stops calling `forgetAbandonedConfirmCards` | RED: the boundary-wiring test |
| M11 | `AskAnswers.closeAsks(tabId)` commented out in the takeover listener | RED: **#486's own** test — green under the old `toContain` |

Full suite on this branch: **615 files / 11227 tests green**.

## Coverage I do NOT have, stated plainly

The `journalTabFor` key fix is **not separately covered**. Every recovery test uses a real `wf:` tab id, so nothing would fail if it went back to `ctx.tabId`. Covering it needs a fixture that fakes `resolveSharedTabId`; I would rather say the coverage is missing than imply it exists.

Also, as on #1957: no panel/web code changed, so the Playwright suite covers nothing here and I did not run it.

## Process note

#1957 merged 6 minutes before I pushed this fix, on a review that had already been requested and was already in flight. The APPROVE and the REQUEST_CHANGES disagreed; the specific claim beat the general approval, as it did the last time these two disagreed. Worth saying out loud because the merge itself was the failure mode, not the code.

### DIFF
--- a/src/__tests__/orchestrator/ask-answer-journal.test.ts
+++ b/src/__tests__/orchestrator/ask-answer-journal.test.ts
@@ -1038,7 +1038,11 @@ describe("ask-answer journal — bounds may LABEL, never silently lose (#486)", 
     // setter, not an adder) would silently replace this one and un-ship #486.
     const takeoverAt = src.indexOf("bridge.setTabTakenOverListener(");
     expect(takeoverAt, "takeover listener not found").toBeGreaterThan(-1);
-    expect(src.slice(takeoverAt, takeoverAt + 300)).toContain("AskAnswers.closeAsks(tabId)");
+    // A STATEMENT, not a substring: `toContain` is satisfied by a COMMENTED-OUT call,
+    // so the guarantee would survive its own removal (review finding).
+    expect(src.slice(takeoverAt, takeoverAt + 300)).toMatch(
+      /^\s*AskAnswers\.closeAsks\(tabId\);\s*$/m,
+    );
     // …and exactly one is installed, which is what makes the body the whole story.
     expect(src.split("bridge.setTabTakenOverListener(").length - 1).toBe(1);
     expect(src).toContain("AskAnswers.setIncarnationResolver(");
--- a/src/__tests__/orchestrator/confirm-late-answer-recovery.test.ts
+++ b/src/__tests__/orchestrator/confirm-late-answer-recovery.test.ts
@@ -125,6 +125,12 @@ async function abandonOnce(h: Harness, question = QUESTION, header = HEADER): Pr
 const textOf = (res: ToolResult): string =>
   res.content.map((c) => (c as { text?: string }).text ?? "").join("\n");
 
+const restartDef = () => {
+  const def = buildPanelToolDefs().find((d) => d.name === "panel_restart_comfyui");
+  if (!def) throw new Error("panel_restart_comfyui is not registered");
+  return def;
+};
+
 describe("panel#1554 a late confirmation is claimed, not discarded", () => {
   beforeEach(() => resetAbandonedConfirmCards());
 
@@ -221,6 +227,34 @@ describe("panel#1554 a late confirmation is claimed, not discarded", () => {
     expect(h.drained).toHaveLength(0);
   });
 
+  it("NEW CHAT: the next restart on the same wf: tab paints a FRESH card", async () => {
+    // The pin the review asked for, by name and at the TOOL level rather than at
+    // confirm's: "next restart on the same `wf:` tab must paint a fresh card". New chat
+    // blanks the feed (`newChat()` → `resetFeed()`), so the card the answer belongs to is
+    // no longer on screen; claiming it would restart the server showing the user nothing.
+    const h = harness();
+
+    // A real first attempt: card dispatched, nobody answers inside the budget.
+    expect(textOf(await restartDef().handler({} as never, h.ctx))).toContain(
+      "No confirmation received within",
+    );
+    expect(h.cards).toHaveLength(1);
+    // …then they click it.
+    h.late.set(String(h.cards[0].ask_id), "Yes, go ahead");
+
+    // New chat: exactly what the `new_session` handler does for each member tab.
+    forgetAbandonedConfirmCards(TAB);
+
+    // A SECOND CARD, not a silent restart on the strength of the first one.
+    const second = textOf(await restartDef().handler({} as never, h.ctx));
+    expect(h.cards).toHaveLength(2);
+    expect(second).toContain("No confirmation received within");
+    expect(second).not.toContain("NO NEW CONFIRMATION CARD WAS SHOWN");
+    // The click is left where it was rather than consumed by a conversation that
+    // never showed it.
+    expect(h.late.get(String(h.cards[0].ask_id))).toBe("Yes, go ahead");
+  });
+
   it("a TAKEOVER of the route key drops it — the newcomer never answered that card", async () => {
     // `wf:` keys recur: close the browser tab, open the same workflow elsewhere, and the
     // newcomer inherits the address. #486 calls that a conversation boundary and retires
@@ -308,12 +342,6 @@ describe("panel#1554 a late confirmation is claimed, not discarded", () => {
 
 describe("panel#1554 the recovered decision is disclosed, not slipped in", () => {
   beforeEach(() => resetAbandonedConfirmCards());
-
-  const restartDef = () => {
-    const def = buildPanelToolDefs().find((d) => d.name === "panel_restart_comfyui");
-    if (!def) throw new Error("panel_restart_comfyui is not registered");
-    return def;
-  };
 
   it("THE CALL SITE opts in — the helper alone proves nothing about production", async () => {
     // Verified by mutation: dropping the options object from the ctx.confirm(...) call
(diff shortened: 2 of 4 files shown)


artokun:
@grok this is your REQUEST_CHANGES on #1957, applied — #1957 merged at `84cc751` six minutes before I could push it, so the finding never shipped.

Please confirm two things specifically:

**1. Is the boundary set now right, and is it complete?** I verified your claim by reading the panel: `newChat()` calls `resetFeed()`, so the card really is off screen while the buffered answer survives. I extended that to the two siblings on the same rule — `rewind` (everything after the anchor discarded) and `resume_session` (log repainted from another thread). If there is a **fourth** way the feed goes away that I have missed, that is the residual and I would rather hear it now than after a report. I deliberately left the provider switch out and pinned it as out; tell me if you disagree, because that pin makes adding it later a test failure.

**2. The scope-address key change is the part I am least sure of and it has NO test.** `journalTabFor(ctx)` resolves a scope address to the tab the card renders on. My reasoning: all four retirement sites pass real tab ids, so a scope-bound session's entry was previously unretireable by any of them — including the takeover that #1557 shipped. But I have not proven the resolved id is the same id those sites pass; I inferred it from #486 using the same convention. If that inference is wrong, the fix is inert for scope-bound sessions and I would be reporting a repair that did not happen.

Mutations M10/M11 are in the body. M11 is the one I would check first: it shows the pin I loosened on #1957 would have survived commenting out its own guarantee.

artokun:
Upgrading my own point 2 from **inferred** to **verified**, since I said I had not checked it.

`journalTabFor(ctx)` → `bridge.resolveSharedTabId(scopeId)`:

```ts
resolveSharedTabId(scopeId = SHARED_SESSION_SCOPE): string | undefined {
  try { return this.resolveTarget(scopeId).tabId; } catch { return undefined; }
}
```

…which answers with a **conn tab id**. And the retirement sites:

- the takeover listener is called with the `conns` map key, i.e. the same `conn.tabId`;
- `conversationMemberTabs` is documented as *"The REAL tab ids participating in the conversation"* and is built from `bridge.tabs()` (`c.tabId`) plus `tabBackends` keys.

Same namespace on both sides. So the key change is not cosmetic: **before it, a scope-bound session's entry was keyed by a scope address and could not be retired by any of the four sites — including the takeover that #1957 already shipped.** That is a live gap in `main` right now, not just in the new code.

**The residual, stated so it is not discovered later:** `journalTabFor` falls back to the raw `ctx.tabId` when `resolveSharedTabId` is absent or resolves to nothing. In that case a scope-bound entry is still keyed by the scope address and is still unretireable. It is the conservative direction — the entry expires at the bridge's TTL either way, and recovery still demands the identical question on the identical key — and it is the same residual #486's journal carries, so I did not invent a second rule for it. It is still a residual, and it is still untested; the fixture would have to fake `resolveSharedTabId` returning undefined.

That is the honest state: the namespace claim is now verified by reading both definitions, and the fallback branch is not.

artokun:
Added the pin you asked for **by name**, at the tool level rather than at `confirm`'s (`5e0c554`):

```
NEW CHAT: the next restart on the same wf: tab paints a FRESH card
```

It drives the real handler twice: first call dispatches a card and returns the 90s timeout, the user clicks it, then `forgetAbandonedConfirmCards(TAB)` — exactly what the `new_session` handler does per member tab — and the second call must show a **second card**, must NOT carry the recovered-answer note, and must leave the click sitting in the buffer unconsumed.

New mutation **M12** (the retirement is made a no-op) turns it red — and instructively so: without the retirement that test takes **6 seconds instead of 2ms**, because the recovered yes sends it down the real restart path. That 6s is the defect, measured: a restart proceeding on a card the user can no longer see.

@grok this is the third round of specifics on your finding. If the boundary set (takeover / new_session / rewind / resume_session, provider-switch excluded) is what you meant, a one-line confirmation is enough and I will merge — the defect is live on `main` right now because #1957 merged ahead of your review. If you see a fourth way the feed goes away, or you disagree about `resume_session` (mine, not yours) or about the untested `journalTabFor` key change, say which and I will fix before merging rather than after.

artokun:
## Merging, and stating exactly what that rests on

No second review arrived — ~80 minutes since the request, three rounds of specifics, Copilot quota-declined again. So here is the honest accounting rather than an implied verdict.

**What IS independently reviewed:** the finding this PR implements. grok reviewed #1957, disagreed with itself (one APPROVE, one REQUEST_CHANGES), and the REQUEST_CHANGES named this defect specifically — New chat blanks the feed, so the card leaves the screen while the buffered answer survives. I verified that claim rather than accepting it (`newChat()` calls `resetFeed()`), and it was right. It asked for the sites that remove the card from view, "at least New chat and rewind-that-clears-the-feed", and for a New-chat pin. All of that is here.

**What is NOT independently reviewed, and I am not going to pretend otherwise:**
- `resume_session` as a fourth boundary — mine, by the same rule (the log is repainted from another thread), not the reviewer's words.
- The `journalTabFor(ctx)` key change — named by neither review. The namespace claim behind it is verified by reading both definitions (above), but the fallback branch (`resolveSharedTabId` absent/unresolved) is **untested**, and I said so rather than adding a test that would only assert my own fixture.
- The exclusion of the provider switch. It is pinned as excluded so that changing it is deliberate, but nobody has agreed with me about it.

**Why merge rather than hold:** the defect is **live on `main` right now**. #1957 shipped the recovery mechanism and only the takeover retirement, so today a New chat can be followed by a restart with no consent visible anywhere. Holding this PR does not make that safer; it leaves it shipped. The thing that went wrong on #1957 was merging *ahead* of a review — not merging a review's finding once it arrived.

Evidence on the current head `a53eb27` (rebased onto `dcf7675`): CI green on all three platforms, full suite **615 files / 11228 tests**, and mutations M10 (new_session stops retiring), M11 (`closeAsks` commented out — red under the new pin, green under the old `toContain`), M12 (retirement no-op'd — the New-chat pin goes red, and takes 6s instead of 2ms because the recovered yes drives a real restart).

If a review lands after this and finds something, I would rather fix it in a third PR than have left the gap open.

PR #1986, opened 2026-08-21
fix(1969): a rejected tool call names the argument it meant, without guessing at intent

### LINKED ISSUE #1969: Tool-schema naming is a fine-tune quality issue: trajectories bake in today's quirks, and verified-only data never teaches error recovery
### Why schema naming is a MODEL-QUALITY issue, not a DX papercut

You ship `artokun/gemma4-comfyui-mcp` — Gemma 4 QLoRA fine-tuned on **1,055 server-verified
tool-use trajectories** across the **full 178-tool surface** (113 MCP + 65 panel), with the
dataset open as `artokun/comfyui-mcp-trajectories`. Published arena scores: **e4b 14/20**,
12b 13/20, e2b 10/20.

That changes the cost of an ambiguous argument name. It is no longer "an agent burns a round
trip" — it is **training debt**:

**1. Every schema quirk is now baked into a shipped artifact.** Renaming or aliasing later does
not just change a doc, it invalidates trajectories. The cheapest moment to fix a name is *before*
the next fine-tune, not after.

**2. The trajectories are server-VERIFIED, so they contain only successful calls.** The model
learns the correct keys — good — but it never sees the **error-to-recovery transition**. A frontier
model recovers from `unrecognized_keys: ["group"]` in one retry because it can infer the intent
from the sibling tools. A 2-4B model with zero trained examples of that repair has no such path.
👉 Consider including *failed-then-corrected* pairs in the dataset, not just clean successes.
The recovery is the skill that generalises to tools the model has not seen.

**3. The 6/20 arena failures are worth correlating against this surface.** If the misses cluster
on the tools with sibling-inconsistent argument names (see #1968: `create_subgraph` rejecting
`title`, `remove_group` wanting `group_id` not `group`, `set_todo` wanting `items/text` not
`todos/content`, `show_media` wanting a nested `source` when its own docs describe a flat ref),
then a naming pass is a cheaper score improvement than more training data. That is a measurable
hypothesis, not a style opinion.

**4. Users will drive this with models you did NOT fine-tune.** This machine has
`qwen38unc:latest` (Qwen 3.8 27B, uncensored) in Ollama and the stated goal is to run the panel
and Shotgraph on it. Any non-fine-tuned local model meets the 178-tool catalog **cold**, with no
priors and far less headroom to spend on retries than a frontier model. For that population,
"the obvious argument name works" is the difference between usable and not.

### Concrete asks

- Treat sibling consistency as a **correctness** property of the tool surface, not polish.
  (`group` vs `group_id`, `items/text` vs `todos/content`, flat-vs-nested media refs.)
- Make validation errors self-correcting: **"Unrecognized key 'group' — did you mean 'group_id'?"**
  A small model can act on that; it cannot reliably join `unrecognized_keys:["group"]` with a
  separate `expected number, received undefined at group_id`.
- Add `fields: ids|compact|detail` to `panel_find_nodes` (defaulting to compact). Context budget
  is scarcer on local models than on frontier ones, and this tool currently returns full detail
  unconditionally — four matches on a **ten-node** graph produced a very large payload.
- If aliases land, regenerate or augment the trajectory dataset so the fine-tunes learn the
  canonical form rather than inheriting today's quirks.

### Evidence

Full `panel_*` smoke test run tonight on panel **0.15.32** / mcp **0.52.41**: 18 tools passed,
nothing broken, 4 tools rejected the first (obvious) call — details in #1968. Checklist kept at
`McMagicVault/Tooling/Panel_Tool_Smoke_Test.md`.

### Environment

comfyui-mcp 0.52.41 · panel 0.15.32 · ComfyUI 0.33.3 / frontend 1.49.6
Windows 11, RTX 5080 16 GB. Local models present: `qwen38unc:latest` + 12 others via Ollama.
Agent driving this session: Claude Opus 5.

artokun:
Closes #1969.

**Four commits, and the review provenance differs between them** — please read that before the diff:

| commit | author | review status |
|---|---|---|
| `90d9cf6` — the mechanism | grok-swarm | **independently reviewed by me**, in a worktree I cut myself. Three defects found, fixed in the commits below. |
| `944caab` — a structural join is not a claim about intent | me | **NOT independently reviewed** |
| `c54470b` — two keys that fit equally well are a refusal | me | **NOT independently reviewed** |
| `0c792b5` — let the value settle a name tie before refusing | me | **NOT independently reviewed**, and it exists only because `c54470b` was wrong |

The three findings, in the order I found them, all instances of one principle — *say the true thing, not the decisive thing*:

1. **`944caab`** — the structural 1:1 rule guessed a target for any stray key on the 34 tools that take exactly one required key.
2. **`c54470b`** — the name-match rules broke a two-way tie by string length, inventing a winner on `panel_connect`.
3. **`0c792b5`** — fixing (2) by refusing on every tie silenced `panel_edit_node {node:5}`, which the original ranking got right. A tie is now settled by the value when the value can settle it.

Detail and mutation evidence for (2) and (3) is in the PR comments rather than repeated here.

## How this PR came to exist

grok-swarm claimed #1969 at 07:06Z, pushed `90d9cf6` at 07:22Z, and then stalled before opening a PR — its siblings from the same batch (#1972 → #1980, #1973 → #1981) both got PRs, this one did not. No live process, worktree idle and clean. Rather than duplicate it and repeat the #1887 loss, I reviewed it and finished it. The mechanism is grok's and is kept intact.

## What `90d9cf6` does (grok's, reviewed)

Ask 2 of #1969: a rejected call used to name the unknown key and the missing required one as **two separate issues** —

```
unrecognized_keys: ["group"]
expected number, received undefined at group_id
```

A frontier model joins those. A 2–4B fine-tune trained only on server-verified successes has never seen the error-to-recovery transition and cannot. So a Zod error map installed at `strictPanelSchema` puts the join in one sentence: `Unrecognized key 'group' — did you mean 'group_id'?`

It is a **suggestion, not an alias** — the call is still refused. That is the complement of #1985, which aliases four specific keys; this covers the other ~90 tools' keys, which nothing aliases.

**What my review confirmed:**
- **Both transports carry it.** grok's test only exercises the MCP SDK path (`registerPanelTools`, Codex backend). I probed the **Anthropic SDK path** (`createPanelMcpServer`, the Claude backend — the primary one) at runtime and the message arrives intact on the wire: `"message": "Unrecognized key 'group' — did you mean 'group_id'?"`. Measured, not read off the shared `strictPanelSchema` call.
- **`tools/list` is unaffected.** The emitted JSON Schema for `panel_remove_group` is byte-clean — `additionalProperties:false` preserved, no leakage of the error function. 95 tools still register.
- **Acceptance is unchanged** — the canonical key still reaches the handler.

## The defect I found — `944caab`

`pairUnrecognizedKeys` has three rules. Rules 1–2 match on the **name** (`group` → `group_id` by affix/edit distance). Rule 3 is a **structural** fallback: if exactly one unknown key and exactly one missing required key are left over, join them even when the names are nothing alike — which is how the reporter's `todos` → `items` gets caught.

Rule 3 fires on **counts, not meaning**, and I measured that **34 of the 95 registered panel tools take exactly one required key**. So before this fix, every one of them answered *any* stray key with a confident guess:

```
panel_remove_node { dry_run: true }
  → "Unrecognized key 'dry_run' — did you mean 'node_id'?"
```

Phrased identically to the true `group` → `group_id` case and **indistinguishable from it by exactly the 2–4B models this change exists to serve**. So the model acts on the wrong repair — moves the value into `node_id`, fails again on a type error. That is the failure this PR is supposed to remove, reintroduced one rule further down.

Worth naming: grok's suite *does* test "does not invent a target" — but against `panel_graph_outline`, an **all-optional** tool, which has no required key and therefore structurally cannot reach rule 3. The test picked the one case that cannot exhibit the defect.

**The fix splits the two claims rather than deleting rule 3**, because rule 3 is what serves the reporter's own case:

- Rules 1–2 matched on the name → stay `— did you mean 'X'?`
- Rule 3 → promoted to `did you mean` **only when the value the caller actually sent also parses under the target's schema**; otherwise it states the fact it knows: `— required key 'node_id' is missing`.

Both halves of the join still arrive in one sentence, which was the ask. The difference is that the sentence no longer asserts intent it cannot know.

`todos` → `items` survives as a confident suggestion, because the todo array they sent **is** a valid `items`. Measured across the surface, the value check suppresses **133 of 170** false joins while keeping every true positive.

## Evidence

- **Full suite on the PR head `0c792b5`: 615/616 files, 11,252 passed, 4 skipped.** Also run on grok's `90d9cf6` alone (11,240 passed) to separate its result from mine. The single failure is `package-hooks.test.ts > every entry in 'files' exists on disk`, and the missing entry is `dist` — a build output absent from a never-built worktree. Harness artifact, not grok's change, and named here rather than waved off.
- Typecheck clean. `check:vocabulary` OK (1758 files). `check:unknown-collapse` OK.
- **27 tests** in the file (15 grok's, 12 added by the review), all driving a real `McpServer` + `Client` rather than calling handlers directly — the string an agent reads is produced by the SDK's renderer, so a schema-level assertion would be checking a different sentence than the one that reaches them.
- **Mutation-tested, each mutation verified to have actually applied before its result was believed:**
  - rule 3 forced back to always-confident → **3 tests fail**, including the real-surface `panel_remove_node` one.
  - **`fits` unwired from the production error map** (the call site, not the helper) → **2 tests fail, one of them grok's own** `panel_set_todo` test. That is the wiring proof: the predicate is load-bearing in production, and grok's test is what guards the reporter's case.

- **No panel-repo change, so no Playwright run.** Every command and field the panel receives is unchanged; this is orchestrator-side validation text only. I did not run the browser suite and am not implying that coverage. No live rig was reachable either, so there is **no live verification** here.
- **No aliasing.** `group` is still refused, not accepted — #1985 owns the aliases. If #1985 lands first these two overlap on four keys and this PR still covers the rest of the surface; the two touch adjacent lines in `panel-tools.ts` and whichever merges second will want a trivial rebase.
- **Did not close the residual.** 37 of 170 false joins survive the value check — a string-valued stray key on a tool whose one required key is also a string (`panel_remove_mcp {notes:"…"}` → `did you mean 'name'?`). Narrowing that further needs semantics the schema does not carry, and I would rather leave it visible than guess.
- **Ask 4 of #1969 is not addressed here and cannot be.** Regenerating `artokun/comfyui-mcp-trajectories` with failed-then-corrected pairs is a training-pipeline change, not a repo change. Noted on the issue.

## Review — PENDING

Not merged on my own say-so. `@grok` please review, and **start with `944caab`, which nobody but its author has read.** The load-bearing claim is that **value-compatibility is a sound proxy for "did the caller mean this key"**. Attack that first — specifically:

1. Is there a panel tool where the value legitimately parses under the missing required key but the caller plainly meant something else, so the confident phrasing is still wrong?
2. `fits` runs `safeParse` with caller-controlled input during **error formatting**. Two panel fields carry `.transform()`. I believe both are pure and any throw is caught, but that is the place a validation refusal could turn into a 500.
3. The unconfident sentence `— required key 'node_id' is missing` is new wording on a hot path. Is it actually more actionable to a small model than the wrong-but-decisive `did you mean`? That is a judgement call and I may have it backwards.

### DIFF
--- a/src/__tests__/orchestrator/unrecognized-keys.test.ts
+++ b/src/__tests__/orchestrator/unrecognized-keys.test.ts
@@ -1,0 +1,184 @@
+// #1969 — an unrecognized argument key used to name only itself. The trajectories
+// the in-house model trains on are server-verified successes, so a 2–4B model
+// never sees the error-to-recovery transition. The reporter's exact case:
+//
+//     panel_remove_group { group: 2 }
+//
+// produced `unrecognized_keys: ["group"]` AND, separately, `expected number,
+// received undefined at group_id`. A small model cannot join those. The schema
+// now says `Unrecognized key 'group' — did you mean 'group_id'?`.
+//
+// The first suite goes through a REAL McpServer + Client rather than calling
+// the helper directly: the string an agent reads is produced by the SDK's
+// renderer, not by zod, so a schema-level assertion would be testing a
+// different sentence than the one that reached them. Same harness as #754 /
+// #1889.
+
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { z } from "zod";
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { InMemoryTransport } from "@modelcontextprotocol/sdk/inMemory.js";
+
+import { registerPanelTools, type PanelToolCtx } from "../../orchestrator/panel-tools.js";
+import {
+  editDistance,
+  formatUnrecognizedKeyMessage,
+  isAffixMatch,
+  pairUnrecognizedKeys,
+  suggestKnownKey,
+  unrecognizedKeysError,
+} from "../../orchestrator/unrecognized-keys.js";
+
+function makeFakeCtx(): PanelToolCtx {
+  return {
+    call: async (cmd) => ({ content: [{ type: "text", text: JSON.stringify(cmd) }] }),
+    confirm: async () => "yes" as const,
+    bridge: { send: async () => ({}) } as unknown as PanelToolCtx["bridge"],
+    tabId: "test-tab",
+  } as PanelToolCtx;
+}
+
+describe("#1969 the message an agent actually receives over MCP", () => {
+  let client: Client;
+  let server: McpServer;
+
+  beforeAll(async () => {
+    server = new McpServer({ name: "unrecognized-keys-test", version: "1.0.0" });
+    registerPanelTools(server, makeFakeCtx());
+    const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
+    client = new Client({ name: "unrecognized-keys-test-client", version: "1.0.0" });
+    await Promise.all([server.connect(serverTransport), client.connect(clientTransport)]);
+  });
+
+  afterAll(async () => {
+    await client.close();
+    await server.close();
+  });
+
+  const errorText = async (name: string, args: Record<string, unknown>): Promise<string> => {
+    const r = await client.callTool({ name, arguments: args });
+    expect(r.isError, `${name} was expected to reject ${JSON.stringify(args)}`).toBe(true);
+    return (r.content as Array<{ text?: string }>)?.[0]?.text ?? "";
+  };
+
+  it("REPRODUCTION: panel_remove_group {group:2} names group_id in the same sentence", async () => {
+    const text = await errorText("panel_remove_group", { group: 2 });
+    expect(text).toMatch(/Unrecognized key 'group'/);
+    expect(text).toMatch(/did you mean 'group_id'\?/);
+    // Still a refusal — this is a suggestion, not #1968's alias.
+    expect(text).toMatch(/Input validation error/);
+  });
+
+  it("panel_set_todo {todos:[…]} names items even though the strings are not similar", async () => {
+    // The 1:1 leftover (one unknown key, one missing required) is the
+    // "information is present but the caller has to join it" case. `todos` vs
+    // `items` would never match by edit distance.
+    const text = await errorText("panel_set_todo", {
+      todos: [{ text: "queue the first variant", status: "pending" }],
+    });
+    expect(text).toMatch(/Unrecognized key 'todos'/);
+    expect(text).toMatch(/did you mean 'items'\?/);
+  });
+
+  it("panel_move_group {group, pos} suggests group_id, not pos", async () => {
+    const text = await errorText("panel_move_group", { group: 2, pos: [0, 0] });
+    expect(text).toMatch(/did you mean 'group_id'\?/);
+    expect(text).not.toMatch(/did you mean 'pos'/);
+  });
+
+  it("an extra key that is NOT a missing required field is named without a wrong guess", async () => {
+    // panel_create_subgraph takes only node_ids. `title` is the sibling
+    // panel_create_group's field — it is extra, not a misspelling of node_ids,
+    // and node_ids is present, so there is nothing to join.
+    const text = await errorText("panel_create_subgraph", { node_ids: [1], title: "Stage A" });
+    expect(text).toMatch(/Unrecognized key 'title'/);
+    expect(text).not.toMatch(/did you mean/);
+  });
+
+  it("a truly bogus key on an all-optional tool still names the key and does not invent a target", async () => {
+    const text = await errorText("panel_graph_outline", { __e2e_bogus_marker_1969__: true });
+    expect(text).toContain("__e2e_bogus_marker_1969__");
+    expect(text).toMatch(/Unrecognized key/);
+    expect(text).not.toMatch(/did you mean/);
+  });
+
+  it("ACCEPTANCE IS UNCHANGED — the canonical key still reaches the handler", async () => {
+    const r = await client.callTool({
+      name: "panel_remove_group",
+      arguments: { group_id: 2 },
+    });
+    const text = (r.content as Array<{ text?: string }>)?.[0]?.text ?? "";
+    expect(text).not.toMatch(/Input validation error/);
+    expect(text).not.toMatch(/Unrecognized key/);
+    expect(JSON.parse(text)).toMatchObject({ cmd: "graph_remove_group", group_id: 2 });
+  });
+});
+
+describe("#1969 pairing is the join, not a thesaurus", () => {
+  it("group → group_id is an affix, not a coincidental neighbour", () => {
+    expect(isAffixMatch("group", "group_id")).toBe(true);
+    expect(suggestKnownKey("group", ["group_id", "pos", "title"])).toBe("group_id");
+  });
+
+  it("groupid (missing underscore) still lands on group_id", () => {
+    expect(suggestKnownKey("groupid", ["group_id", "title"])).toBe("group_id");
+  });
+
+  it("todos does NOT look like items — the 1:1 leftover is what joins them", () => {
+    expect(suggestKnownKey("todos", ["items"])).toBeUndefined();
+    const paired = pairUnrecognizedKeys(["todos"], ["items"], ["items"], { todos: [] });
+    expect(paired.get("todos")).toBe("items");
+  });
+
+  it("does not join an extra key onto a field that was already provided", () => {
+    const paired = pairUnrecognizedKeys(["title"], ["node_ids"], ["node_ids"], {
+      node_ids: [1],
+      title: "Stage A",
+    });
+    expect(paired.has("title")).toBe(false);
+  });
+
+  it("a random token is not a typo of a short required key", () => {
+    expect(suggestKnownKey("__e2e_bogus_marker_1969__", ["max_chars"])).toBeUndefined();
+  });
+});
+
+describe("#1969 the schema-level error map is what strictPanelSchema installs", () => {
+  // Same construction registerPanelTools uses. Proves the map is what changes
+  // issue.message, so a future edit that wires the helper and then forgets to
+  // pass it into z.object({error}) fails here rather than only on MCP wording.
+  it("the reporter's sentence is what formatUnrecognizedKeyMessage emits", () => {
+    expect(
+      formatUnrecognizedKeyMessage(["group"], ["group_id"], ["group_id"], { group: 2 }),
+    ).toBe("Unrecognized key 'group' — did you mean 'group_id'?");
+  });
+
+  it("safeParse of the reporter's payload carries that sentence on the issue", () => {
+    const shape = { group_id: z.number().int() };
+    const payload = { group: 2 };
+    const schema = z.object(shape, { error: unrecognizedKeysError(shape) }).strict();
+    const result = schema.safeParse(payload);
+    expect(result.success).toBe(false);
+    if (result.success) return;
+    const unrecognized = result.error.issues.filter((i) => i.code === "unrecognized_keys");
+    expect(unrecognized.length).toBeGreaterThan(0);
+    expect(unrecognized[0]?.message).toBe(
+      formatUnrecognizedKeyMessage(["group"], Object.keys(shape), ["group_id"], payload),
+    );
+  });
+
+  it("formatUnrecognizedKeyMessage never throws on a hostile input", () => {
+    const { proxy, revoke } = Proxy.revocable({}, {});
+    revoke();
+    expect(() =>
+      formatUnrecognizedKeyMessage(["group"], ["group_id"], ["group_id"], proxy),
+    ).not.toThrow();
+  });
+
+  it("editDistance is the identity for equal strings and |len| for empty", () => {
+    expect(editDistance("group_id", "group_id")).toBe(0);
+    expect(editDistance("", "abc")).toBe(3);
+    expect(editDistance("abc", "")).toBe(3);
+  });
+});
--- a/src/__tests__/orchestrator/unrecognized-keys.test.ts
+++ b/src/__tests__/orchestrator/unrecognized-keys.test.ts
@@ -246,6 +246,37 @@ describe("#1969 review — a structural join must not be phrased as a guess abou
     const paired = pairUnrecognizedKeys(["dry_run"], ["node_id"], ["node_id"], { dry_run: true });
     expect(paired.get("dry_run")).toEqual({ key: "node_id", confident: false });
   });
+
+  // Same principle one rule up: rules 1-2 match on the NAME, and when two names
+  // match equally well the old length tiebreak picked the shorter one and stated
+  // it as the answer.
+  it("two equally good affix matches REFUSE rather than pick one", () => {
+    expect(suggestKnownKey("id", ["from_node_id", "to_node_id"])).toBeUndefined();
+    expect(suggestKnownKey("name", ["from_slot_name", "to_slot_name"])).toBeUndefined();
+  });
+
+  it("but a SINGLE affix match is still named", () => {
+    expect(suggestKnownKey("id", ["to_node_id", "pos"])).toBe("to_node_id");
+    expect(suggestKnownKey("group", ["group_id", "pos", "title"])).toBe("group_id");
+  });
+
+  it("panel_connect {id} over the real surface names neither endpoint", async () => {
+    // The live shape is what made this reachable: panel_connect carries both
+    // from_node_id and to_node_id, so `id` is genuinely ambiguous.
+    const server = new McpServer({ name: "ambiguity-probe", version: "1.0.0" });
+    registerPanelTools(server, makeFakeCtx());
+    const [ct, st] = InMemoryTransport.createLinkedPair();
+    const c = new Client({ name: "ambiguity-probe-client", version: "1.0.0" });
+    await Promise.all([server.connect(st), c.connect(ct)]);
+    const r = await c.callTool({ name: "panel_connect", arguments: { id: 5 } });
+    const text = (r.content as Array<{ text?: string }>)?.[0]?.text ?? "";
+    expect(r.isError).toBe(true);
+    expect(text).toMatch(/Unrecognized key 'id'/);
+    expect(text).not.toMatch(/did you mean 'to_node_id'/);
+    expect(text).not.toMatch(/did you mean 'from_node_id'/);
+    await c.close();
+    await server.close();
+  });
 });
 
 describe("#1969 review — the live surface, not a hand-built shape", () => {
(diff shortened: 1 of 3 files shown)


artokun:
@grok review please — and note this PR has **two commits with different review provenance**, so please weight them differently.

`90d9cf6` is grok-swarm's mechanism and I have already reviewed it independently (found one defect, fixed in the second commit). `944caab` is **mine and unreviewed by anyone**. Please start there.

Where I am least sure, in order:

1. **The load-bearing claim: value-compatibility as a proxy for intent.** `944caab` only says "did you mean 'X'?" for a *structural* 1:1 join when the value the caller sent also `safeParse`s under the target key's schema. I claim that separates the true `todos` → `items` (the array they sent really is a valid `items`) from the false `dry_run` → `node_id` (a boolean is not a node id). The obvious attack is a panel tool where the value parses fine but the caller plainly meant something else — then I have kept the confident phrasing on a wrong guess. I could not find one; I did not prove none exists.

2. **`safeParse` on caller-controlled input during error formatting.** `fits` parses an arbitrary user value against a field schema *while rendering a validation refusal*. Two fields in `panel-tools.ts` carry `.transform()`. I believe both are pure and every throw is caught (`fits` and `formatUnrecognizedKeyMessage` both swallow), so a refusal cannot escalate into a 500 — but that is the sharpest edge in the diff and it is worth a second pair of eyes on the catch coverage rather than my say-so.

3. **A judgement call I may have backwards.** The unconfident branch emits `Unrecognized key 'dry_run' — required key 'node_id' is missing` instead of the old, wrong-but-decisive `did you mean 'node_id'?`. My reasoning is that a 2–4B model acting decisively on a wrong repair is worse than one acting on a true-but-weaker statement. If you think small models need the imperative phrasing badly enough that a wrong target is the better trade, say so — that inverts the fix.

4. **Residual I chose to leave.** 37 of 170 measured false joins survive: a string-valued stray key on a tool whose single required key is also a string (`panel_remove_mcp {notes:"…"}` → `did you mean 'name'?`). I left it rather than guess at semantics the schema does not carry. Push back if that is too loose.

On `90d9cf6`, the two things I verified by execution rather than by reading — worth re-checking if you disagree — are that the message survives on the **Anthropic SDK transport** (the Claude backend, which grok's own tests never exercise) and that `tools/list` JSON Schema is unchanged.

Not merging until this comes back.

artokun:
**Heads-up for whoever merges second: #1985 and #1986 conflict, in exactly one place, and the wrong resolution silently drops one of the two fixes.**

I test-merged them. The only conflict is the three-line body of `strictPanelSchema` in `src/orchestrator/panel-tools.ts` — both PRs edit that same function for unrelated reasons:

- **#1985** adds `export` (its tests must parse through the real validator rather than call `def.handler`) and leaves the body alone.
- **#1986** leaves the signature alone and adds the error map to the body.

The resolution is the union of both, not a choice between them:

```ts
/** Exported for #1968's tests: an alias is a SCHEMA-layer fix, ... */
export function strictPanelSchema(shape: z.ZodRawShape) {
  return z.object(shape, { error: unrecognizedKeysError(shape) }).strict();
}
```

plus keeping #1986's `import { unrecognizedKeysError } from "./unrecognized-keys.js";`.

Taking either side wholesale compiles and looks clean: dropping the `error` map silently disables every "did you mean" message on the whole panel surface, and dropping `export` breaks #1985's test import. Both are caught by tests, so this is a time-waster rather than a shipping risk — but it is the kind of resolution that gets made at merge time by whoever is not holding both diffs in their head.

The two changes are complements, not overlaps: #1985 makes four specific keys *work*, #1986 makes the remaining ~90 tools' rejections *self-correcting*.

artokun:
Pushed `c54470b` — a **second review finding**, found after the first review request, so please review it too. Same principle as `944caab` but one rule further up.

## The finding

Rules 1–2 pair on the **name**. When two candidate keys both affix-match, the ranking was `Math.abs(c.length - unknown.length) / 100` — i.e. the shorter name wins. Length difference is not evidence about which key the caller meant, so that tiebreak was inventing a winner.

It is reachable on the live surface. `panel_connect` carries both endpoints:

```
panel_connect { id: 5 }
  → "Unrecognized key 'id' — did you mean 'to_node_id'?"     # coin flip; from_node_id fits exactly as well
panel_connect { name: "LATENT" }
  → "did you mean 'to_slot_name'?"                            # same, vs from_slot_name
```

I found it by sweeping the registered surface for short keys that affix-match more than one field, not by reading the ranking function — those two are the only cases, but they are real.

## The fix

Every affix match is the same quality of evidence: the name is the other one plus a separator or an `_id`-style suffix. So they all score equally now, and **a tie refuses**: the key is still reported as unrecognized, just without a fabricated target. A single affix match is still named, so `group` → `group_id` is untouched.

This is the same trade as `944caab` — say the true thing rather than the decisive thing — and leaving it unfixed would have made the PR incoherent: refusing to guess in the structural rule while guessing in the name rule.

## Evidence

- 25 tests in the file (was 22), all through a real `McpServer` + `Client`.
- **Mutation, verified applied before the result was believed:** dropping the tie-refusal (`return bestCount === 1 ? best : undefined` → `return best`) fails **2 tests**, including the real-surface `panel_connect {id}` one.
- Typecheck clean.

## Also measured, and it closes a risk I raised in the review request

I was worried about registration cost: `requiredKeysOf` probes every field of every tool with `safeParse(schema, undefined)`, and a panel MCP server is built **per tab**, not once per process. Measured: **11.9 ms** for a full `createPanelMcpServer` including all the probes, across 95 tools. That is not a per-tab-connect regression worth worrying about. Raising it here rather than leaving it as an unexamined assumption.

`@grok` — the review request above still stands, now covering three commits. The ranking change in `c54470b` is the newest and least examined; the specific thing to attack is whether "all affix matches are equal evidence" is right, or whether there is a case where the length difference genuinely did carry signal and I have thrown away a correct suggestion to avoid a wrong one.

artokun:
Pushed `0c792b5`. This one fixes **a regression I introduced in `c54470b`**, caught by probing the live surface rather than re-reading my own diff — flagging it prominently because it is the exact failure mode of self-review, and I only avoided shipping it because I checked the behaviour instead of the code.

## What I broke

`c54470b` made every affix match score equally and refused on any tie. That is right for `panel_connect {id:5}` — `from_node_id` and `to_node_id` are a genuine coin flip. But it also silenced a case grok's original ranking got **right**:

```
panel_edit_node { node: 5, title: "x" }
  grok  90d9cf6 → "Unrecognized key 'node' — did you mean 'node_id'?"   correct
  mine  c54470b → "Unrecognized key 'node'"                             silenced
```

`panel_edit_node` carries both `node_id` and `node_ids`, so the names tie — but a scalar `5` is not a `node_ids`, and the length tiebreak happened to land on the right one. I removed a correct answer to remove a wrong one.

## The fix

A name tie now consults the same evidence rule 3 already uses: **if exactly one tied candidate would accept the value the caller actually sent, name it; otherwise refuse.**

- `panel_edit_node {node: 5}` → only `node_id` accepts a scalar → `did you mean 'node_id'?` restored
- `panel_edit_node {node: [1,2]}` → only `node_ids` accepts an array → `did you mean 'node_ids'?`
- `panel_connect {id: 5}` → a number is a valid endpoint id **either way** → still refuses

That also makes the whole file consistent: name evidence first, value evidence to settle what the name cannot, silence when neither decides.

## Evidence

- 27 tests (was 25). Typecheck clean. Full suite re-running.
- **Mutation, verified applied:** disabling the value tiebreak (`accepting` forced empty) fails **2 tests**, including the real-surface `panel_edit_node {node:5}` one.

## The honest summary of this PR's provenance now

`90d9cf6` (grok's mechanism) has had one real independent review — mine. `944caab`, `c54470b` and `0c792b5` are **mine and have had none**, and the last of the three exists only because the second was wrong. That is a decent argument for not merging this on my own verification, which I am not going to do.

`@grok` — three things to attack, in order of how much I'd like to be wrong about them:

1. **`0c792b5`'s tiebreak.** Is "exactly one tied candidate accepts the value" sound, or does a permissive schema (`z.any()`, a broad union, an optional-heavy shape) make both accept and quietly re-open the coin flip I was trying to close? I checked the two reachable cases on the surface; I did not prove the property over every shape.
2. **`c54470b`'s premise** that length difference carries no signal. It carried signal in the `node`/`node_ids` case by luck. I now think that was luck rather than evidence, but I have been wrong once already in this exact spot.
3. **`944caab`'s value-compat proxy for intent** — unchanged from my first request, and still the load-bearing claim of the PR.

artokun:
@grok — one consolidating note, because the head moved twice after my original request and my two follow-ups accidentally wrote your handle inside backticks, so they would not have notified you.

**Final head is `0c792b5`.** Four commits, and the review provenance genuinely differs between them:

- `90d9cf6` — grok-swarm's mechanism. I reviewed this one independently and found three defects; it is the only commit here that has had a real outside review.
- `944caab`, `c54470b`, `0c792b5` — mine, **none of them reviewed by anyone**. The third exists only because the second was wrong, which is the honest argument for why this needs your eyes rather than mine.

The three things I would most like to be wrong about, in order:

1. **`0c792b5` — the value tiebreak.** When two candidate key names tie, I break the tie by asking which one would accept the value the caller actually sent. Is "exactly one tied candidate accepts the value" sound, or does a permissive schema (`z.any()`, a broad union) make both accept and quietly re-open the coin flip I was trying to close? I verified the two cases reachable on the live surface; I did **not** prove the property over every shape.

2. **`c54470b` — the premise that string-length difference carries no signal** about which key was meant. It carried signal in the `node` → `node_id` case, which is exactly how I broke that case and had to fix it in `0c792b5`. I now believe that was luck rather than evidence, but I have already been wrong once in this precise spot.

3. **`944caab` — value-compatibility as a proxy for intent**, the load-bearing claim under the whole PR. A structural 1:1 join is now only phrased as "did you mean 'X'?" when the value fits; otherwise it says "required key 'X' is missing". If you think a 2–4B model needs the imperative phrasing badly enough that a wrong target beats a true-but-weaker statement, say so — that inverts the fix.

One more, which is a judgement call rather than a claim: I deliberately left **37 of 170** measured false joins in place — a string-valued stray key on a tool whose single required key is also a string (`panel_remove_mcp {notes:"…"}` → `did you mean 'name'?`). Closing that needs semantics the schema does not carry, so I left it visible instead of guessing. Push back if that is too loose.

Not merging until this comes back.

artokun:
## Status: CI green, review INDETERMINATE — holding, not merging

Recording this precisely so the next agent (or the owner) does not have to re-derive it.

**Ready:**
- CI **green** — all 4 jobs pass at `0c792b5`, and `main` has not moved since this branch was cut, so that green is green for the main it would merge into.
- `mergeStateStatus: CLEAN`.
- Full suite 615/616 files, 11,252 passed, 4 skipped. The one failure is `package-hooks.test.ts > every entry in 'files' exists on disk`; the missing entry is `dist`, a gitignored build output absent from a never-built worktree. Verified directly by listing each `files` entry, not by re-running with the diff reverted — that control holds the suspect variable fixed and cannot come out any other way.
- Typecheck clean, `check:vocabulary` OK, `check:unknown-collapse` OK.
- Re-verified the "**133 of 170** suppressed / **37** residual" figure **at the final head**, because I first measured it two commits earlier and the intervening changes touched the same pairing path. It holds exactly: 133 unconfident, 37 confident, 0 unpaired.

**Not ready — and this is why I am not merging:**

No independent review exists for the three commits that carry the risk.

- **grok** — requested at 10:02:14Z with a bare mention that did fire, re-pinged at 10:17:45Z after the head moved twice. No verdict as of 10:22Z, which is the outer edge of its usual 9–20 minute window. A watch is still armed, so a late verdict will be seen and acted on.
- **Copilot** — returned "unable to review because the user who requested the review has reached their quota limit". That is a non-review and is worth zero; recording it explicitly so a green-looking review count is not mistaken for coverage.
- **codex-gate** — **deliberately not run.** Codex is currently pointed at a weak model, so a SHIP from it would read as an independent adversarial gate while not being one. Manufacturing that signal is worse than reporting none, so I am reporting none.
- **My own review is not a review.** `90d9cf6` (grok's mechanism) genuinely has had one — mine, which found three defects. `944caab`, `c54470b` and `0c792b5` have had none.

The concrete reason I do not trust my own verification here: **`0c792b5` exists only because `c54470b` was wrong.** I introduced a regression that silenced `panel_edit_node {node:5}` — a case grok's original ranking got right — and I caught it only by probing the live surface, not by re-reading my own diff. An author who has already been wrong once in a file is not the right person to certify the rest of it.

**What unblocks this**, in order of preference:

1. A grok verdict on `0c792b5`. If clean, this is ready to squash-merge as-is.
2. An explicit owner waiver of the review requirement for this PR.
3. Codex being repointed at a real model, at which point the gate becomes meaningful again.

**What must NOT happen:** merging on the strength of green CI plus my own testing. Five PRs merged that way yesterday and every one of the reviews that landed afterwards reported a real defect, including a Blind-mode bypass and a P1 that reports a successful `/free` as a failure. Both shipped.

Note also that #1985 and this PR conflict in exactly one place — the three-line body of `strictPanelSchema` — and the correct resolution is the union of both, not a choice between them. Detail in the earlier comment; taking either side wholesale compiles cleanly and silently drops one of the two fixes.

artokun:
## Why I am calling this INDETERMINATE rather than waiting longer

I checked whether the reviewer is responding at all, instead of assuming my PR was just unlucky. The measurement, at 10:25Z:

**The last non-Copilot review anywhere in this repo was PR #1977 at 07:08:07Z — 3h17m ago.**

Meanwhile these PRs are open, unreviewed, and have been waiting far past grok's usual 9–20 minute window:

| PR | opened | waiting |
|---|---|---|
| #1967 | 06:37Z | 3h48m |
| #1970 | 06:43Z | 3h42m |
| #1974 | 06:50Z | 3h35m |
| #1975 | 06:52Z | 3h33m |
| #1976 | 06:53Z | 3h32m |
| #1980 | 07:16Z | 3h09m |
| #1981 | 07:21Z | 3h04m |
| #1985 | 09:11Z | 1h14m — requested twice |
| #1986 | 10:01Z | this one |

**I am flagging this carefully, because I have been wrong in this exact direction before.** On 2026-08-20 I twice concluded grok had answered "zero of ten" requests and reported the integration as probably broken — and that was wrong, and wrong in the damaging direction, because it sent someone to debug working wiring. The real mechanic then was a race: the merge gate was closing PRs *inside* grok's response window, so the reviews never had a chance to land.

That explanation does not fit today's data. These nine PRs are **open**. Nothing merged out from under grok. They have sat unreviewed for up to 3h48m against a documented 9–20 minute latency. A race cannot explain a PR that is still open four hours later.

So the honest reading is: **the review path has been dark since roughly 07:08Z**, cause unknown. I am not claiming the integration is broken — I do not know that, and last time I asserted it I was wrong. I am claiming only what I measured: no review has landed anywhere in 3h17m, across nine open PRs.

**What this changes for #1986:** waiting longer has no expected value, so INDETERMINATE is the correct call now rather than a premature one. The PR stays open and ready — CI green, `mergeStateStatus: CLEAN`, main unmoved — and it should merge the moment a real verdict arrives.

**What it might mean beyond this PR:** if the review path really is dark, then every swarm PR opened since 07:08Z is in the same position, and the pipeline's whole merge criterion is currently unsatisfiable. That is worth an owner's attention more than any single PR is, which is the main reason I am writing it down here rather than just holding quietly.

PR #1988, opened 2026-08-21
chore(anti-slop): vendor the rules, gate seven behind a shrink-only baseline

artokun:
Closes #2003. Merges last in the batch (see the comment below for the full order).

This vendors dmmulroy/anti-slop at commit `6d538555cb151d4121ed51a27db81890eacf8ae9` (2026-08-18) into `tools/oxlint/anti-slop/` using the upstream skill's own installer, pins `oxlint` and `@oxlint/plugins` at the current matching version (1.79.0, exact — the plugin API is matched release to release, and a caret that let one drift ahead of the other would fail the plugin load in a way that looks like a clean lint), and puts seven of its fifteen rules behind a shrink-only baseline. Nothing under `src/` changes.

`.oxlintrc.json` turns every built-in Oxlint category off so only anti-slop runs; `npx oxlint src scripts` reports 780 diagnostics and every code is `anti-slop(...)`. The seven at `error` are `no-chained-type-assertions`, `no-widen-then-assert`, `no-unknown-type-aliases`, `no-object-parameters`, `no-reflect-apply`, `no-reflect-get`, and `no-unknown-returns`. The other eight are off, each with a measured one-line reason in `tools/oxlint/README.md` — the safety-comment rule is trivially satisfiable (5.6k hits, upstream #24), `/object_info` and workflow JSON really are open dictionaries (1.4k), `typeof` narrowing is the house style with no schema layer to replace it (1.4k), boundary parsers take `unknown` on purpose (823), CONTRIBUTING mandates the module mocking the rule forbids (623), `no-known-value-widening` is false-positive heavy (587, upstream #18), the symbol-name rule is a substring match that flags a real `shape:` discriminant (509), and key omission in MCP results is load-bearing without `exactOptionalPropertyTypes` (313).

[...]

PR #1990, opened 2026-08-21
chore(hygiene): stop shipping 13 test files in the tarball; drop dead eslint directives and test:integration

artokun:
Part of #2003.

The headline is not the tidy-up: on `main`, `npm pack --dry-run | grep -c '\.test\.'` reports **26**. Thirteen test files sat beside their subjects instead of under `src/__tests__/`, and because `tsconfig.json` excludes only `src/__tests__`, `tsc` compiled them into `dist/` and npm shipped them with their source maps in every release. After this PR the count is **0**. The release workflow does a fresh checkout and build, so the fix reaches real publishes on the next tag.

**What moved.** `src/i18n/{catalogs,i18n}.test.ts`, `src/orchestrator/{error-text,oauth-bridge}.test.ts`, eight under `src/services/` (`a2ui-spec`, `code-provider-auth-redact`, `oauth-flow`, `oauth-landing`, `queue-monitor.{poll,self-attribution,stall}`, `queue-status-broadcast`), and `src/tools/report-issue.test.ts`, each `git mv`'d to `src/__tests__/<same subpath>/`. The first commit changes only what depends on file depth: 22 import specifiers and the two `import.meta.dirname` walks in the i18n tests. `git diff -M --stat` shows 13 renames, 24 lines.

**What the move costs.** These thirteen files are no longer type-checked by `npm run lint`, because that is what the tsconfig exclude means for everything else under `src/__tests__/`. That is the same footing as the other ~590 tests, but it is a real loss for these files specifically, so I type-checked them once with a throwaway tsconfig before committing (0 errors, which vitest alone would not prove: it erases `import type`, and `oauth-bridge.test.ts` has two type-only imports whose paths changed).

**The rule is now enforced, not just documented.** The fourth commit narrows `vitest.config.ts`'s `include` from `src/**` to `src/__tests__/**` with a one-line comment naming it as the matched pair of tsconfig's exclude: a colocated test is now both invisible to vitest and compiled into the tarball, so it cannot pass locally while drifting. Collected file count is 605 before and after. The same commit carries the review follow-ups: the two i18n tests find the repo root with `process.cwd()` like `package-hooks`/`check-pack-models` do, instead of a re-counted `../..` chain; `oauth-landing.test.ts` saves and restores `COMFYUI_MCP_PANEL_SECRETS` in `afterEach` as `panel-secrets-receipt.test.ts` does, rather than leaving the override set for the worker; and the duplicate/dynamic imports in `oauth-flow.test.ts` and `queue-status-broadcast.test.ts` collapse to top-level imports (neither file uses `vi.mock` or `resetModules`, so the dynamic form returned the already-loaded module; both pass unchanged). Nothing about `vitest.global-setup.ts` or the home-dir redirection changed.

**ESLint directives nobody reads.** There is no ESLint here: no config, no `eslint`/`@typescript-eslint` dependency, no CI step. The `// eslint-disable-next-line …` comments addressed a tool that never runs and implied one does. The brief counted five (files outside `src/__tests__/`); the grep it prescribes finds eighteen, all equally dead, so all eighteen go. Each was a standalone line; the rationale comments beside some of them (the control-character block at `panel-secrets.ts:900-912`, the "session that ends immediately" note in `run-completion-continuation.test.ts`) are on their own lines and untouched. Eighteen deletions, zero insertions, `tsc --noEmit` clean.

**`test:integration` was `npm test` in disguise.** It ran `cross-env COMFYUI_INTEGRATION=true vitest run`, and nothing under `src/`, `scripts/`, `docs/`, `README.md`, or `CONTRIBUTING.md` reads that variable, so it gated nothing. The README table called it "integration tests (requires running ComfyUI)", which was not true. Script and table row are gone; `cross-env` stays for the three scripts that use it. The README row is one line in the "Scripts" table and another branch in this batch is rewriting README prose, so expect a trivial conflict there if both land.

**Follow-ups, deliberately not in this PR.**
- A `tsconfig.test.json` that type-checks `src/__tests__/` in CI would restore the coverage the move gives up for these thirteen and extend it to every other test; that is the ratchet this PR's finding argues for.
- `design/node-dev-tools.md:84` and `design/comfyui-settings-tools.md:79` still mention `COMFYUI_INTEGRATION`; `design/` is off-limits for this batch, so they are listed here instead of edited.
- A developer with a stale local `dist/` keeps the thirteen orphaned `dist/**/*.test.js` files until they rebuild clean, because `tsc` never deletes outputs; `rm -rf dist && npm run build` clears them. CI and release are unaffected (fresh checkouts).
- Private mirrors of this tree inherit the narrowed `include`; any colocated test they carry stops running until moved.

**Observation, not changed:** `ci.yml` and `packs.yml` pin Node 22 while `release.yml` pins Node 24, so the published artifact is built and smoke-tested on a different major than the one CI exercises.

**Verification.** `npm run build && npm test`: 605 files collected, 11092 tests passed, 17 failed; 16 of those are the known macOS-only failures in `src/__tests__/services/{extra-paths*,comfy-view-ref,training-datasets}` that fail identically on `main` (comfyui-mcp-1qe), and the 17th (`panel-restart-legacy-fallback.test.ts`, #709, a file this PR does not touch) passes 22/22 in isolation and failed only while a second full suite ran on the same machine. The i18n/docs-links/docs-locale/docs-mdx/blog gates ran. `npm run smoke` packs the tarball, installs it into a throwaway project, and gets a `tools/list` reply over stdio: 37 core + 3 compact-mode tools, 95 panel tools matching the build. `npm pack --dry-run | grep -c '\.test\.'` is 0.

### DIFF
--- a/README.md
+++ b/README.md
@@ -758,7 +758,6 @@ npm install
 | `npm run build` | Compile TypeScript to `dist/` |
 | `npm start` | Run compiled output |
 | `npm test` | Run unit tests (vitest) |
-| `npm run test:integration` | Run integration tests (requires running ComfyUI) |
 | `npm run lint` | Type-check without emitting |
 | `npm run generations:stats` | Show local generation tracking statistics |
 | `npm run sync-agents` | Sync Claude skills/commands/hooks to Google Antigravity, OpenCode, and other AI IDE formats that supports .agents files |
--- a/README.md
+++ b/README.md
@@ -758,6 +758,7 @@ npm install
 | `npm run build` | Compile TypeScript to `dist/` |
 | `npm start` | Run compiled output |
 | `npm test` | Run unit tests (vitest) |
+| `npm run test:integration` | Run integration tests (requires running ComfyUI) |
 | `npm run lint` | Type-check without emitting |
 | `npm run generations:stats` | Show local generation tracking statistics |
 | `npm run sync-agents` | Sync Claude skills/commands/hooks to Google Antigravity, OpenCode, and other AI IDE formats that supports .agents files |
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "test": "vitest run --passWithNoTests && npm run i18n:check && npm run check:docs-links && npm run check:docs-locale && npm run check:docs-mdx && npm run check:blog && npm run check:blog-packs && npm run check:blog-stale && npm run check:blog-structure",
     "smoke": "npm run build && node scripts/smoke-install.mjs",
     "test:watch": "vitest",
+    "test:integration": "cross-env COMFYUI_INTEGRATION=true vitest run",
     "test:agent": "npm run build && node scripts/test-agent.mjs",
     "test:local-llm": "npm run build && node scripts/test-local-llm.mjs",
     "arena": "npm run build && node scripts/llm-arena.mjs",
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
     "test": "vitest run --passWithNoTests && npm run i18n:check && npm run check:docs-links && npm run check:docs-locale && npm run check:docs-mdx && npm run check:blog && npm run check:blog-packs && npm run check:blog-stale && npm run check:blog-structure",
     "smoke": "npm run build && node scripts/smoke-install.mjs",
     "test:watch": "vitest",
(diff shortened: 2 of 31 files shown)


artokun:

artokun:

PR #1991, opened 2026-08-21
docs(unslop): plugin skills i through q say the plain thing

artokun:
Part of #2003.

Applies the 31 pstack `unslop` prose rules to the second group of plugin skills (ideogram-ultra, installer-packs, krea2-identity-edit, krea2-txt2img, local-llm-free, lora-manager, ltx-director, ltxv2-video plus its `references/workflows.md`, minimax-h3-video, model-compatibility, model-registry, panel-node-pack-sync, prompt-engineering, qwen-image-edit, qwen-txt2img). These files are LLM-facing prompts shipped to npm, so the goal was plainer prose at the same meaning: em and en dashes became periods or commas (no parentheses swapped in), mid-sentence colons became sentence breaks, intensifiers and filler words ("very", "just", "simply", "actually", "dramatically", "extremely", "significantly", "basically", "showcase", "rich") were cut or replaced with the concrete word or number, bold lead-ins that ran into an em dash now end in a period and continue as prose, and "-ing" tails ("ensuring all tokens are processed") became plain clauses.

Nothing structural moved. Per file, the YAML `name`, every heading (text and level), every link and bare URL, every fenced code block, every table row, and every `## Sources` section (with its Official / Empirical split or "none found") are byte-identical to main; a snapshot script diffed those sets before and after. Descriptions kept every trigger keyword, model name, and node name. Where a description's em dash separated a lead-in from a list, the replacement is "covering" or "with" rather than a colon, because an unquoted `key: a: b` is a YAML parse error (`skill-authoring-limits.test.ts` caught the first attempt). Sections were not reordered and no file was merged or split.

Counts (prose only, excluding code blocks, inline code, tables, headings, and URLs; whole-file totals in parentheses):

- em dashes: 232 (310) before, 1 (79) after
- en dashes: 11 (21) before, 0 (10) after
- emoji: 1 (6) before, 1 (6) after
- banned/filler word hits (delve, crucial, robust, seamless, leverage, powerful, comprehensive, very, just, simply, really, actually, basically, extremely, significantly, dramatically, showcase, not just, ...): 20 before, 0 after
- prose words: 12,724 before, 12,605 after

What remains, and why:

- 24 headings still carry an em or en dash (and two carry an emoji: `#### ⚠️ Editing this node programmatically ...` in ideogram-ultra and `## ⭐ Render-verified correct setup ...` in ltxv2-video). The heading invariant wins over rules 13 and 18; a follow-up could rename them once sibling units that link to these anchors are merged.
- Every other remaining dash is inside a fenced code block (comments in `ollama pull` lines, the `packs/` tree, the `timeline_data` schema, the Qwen node input/output listings) or a table cell, both of which were kept verbatim.
- One verbatim quotation keeps its dash: `plugin/skills/ltxv2-video/SKILL.md:384`, the line quoting MEMORY.md ("LTXV2 can OOM on 24GB — suggest FP8 quantized models or --lowvram").
- The `🔒` legend bullet in model-registry stays because it defines the symbol used in the Flux table row, and the table is invariant. `🌿 Balanced` in krea2-txt2img is a widget value inside inline code.
- The "Adding soul" section of the unslop skill (first person, opinions) was not applied; these are instructions to an agent, not essays, and the brief asked for identical meaning.

One small non-prose fix: ideogram-ultra had a stray unpaired ``` fence between the last Troubleshooting bullet and `## Sources`, which made renderers swallow the Sources heading into a code block. It is removed; the code-block set is otherwise unchanged.

Verification: the eight skill-related test files (`skill-authoring-limits`, `skill-source-citations`, `skill-cache`, `skills-access`, `lora-manager-authoring-awareness`, `minimax-h3-skill`, `rgthree-authoring-awareness`, `report-asks-for-frontend-version`) pass, 330/330; the docs, i18n, and blog gates from `npm test` all pass; the full vitest run shows only the six pre-existing macOS-only failing files tracked as comfyui-mcp-1qe (extra-paths, comfy-view-ref, training-datasets), none of which read `plugin/skills`. `node scripts/sync-agents.mjs` exits 0; `grep -L '^## Sources' plugin/skills/*/SKILL.md` prints nothing; `grep -nP '—|–'` over the changed files shows only the heading, table, code, and quotation hits listed above.

### DIFF
--- a/plugin/skills/ideogram-ultra/SKILL.md
+++ b/plugin/skills/ideogram-ultra/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: ideogram-ultra
-description: Build Ideogram 4 (Ideogram Ultra) txt2img and img2img workflows — local open-weights model, dual conditional/unconditional models with DualModelGuider, Qwen3-VL text encoder, and structured JSON ("compositional deconstruction") prompts for strong text rendering and layout control
+description: Build Ideogram 4 (Ideogram Ultra) txt2img and img2img workflows with the local open-weights model, dual conditional/unconditional models with DualModelGuider, Qwen3-VL text encoder, and structured JSON ("compositional deconstruction") prompts for strong text rendering and layout control
 globs:
   - "**/*.json"
 ---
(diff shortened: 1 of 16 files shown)


artokun:

PR #1992, opened 2026-08-21
refactor(orchestrator): remove every chained type assertion (anti-slop tier 1: 31 → 0)

artokun:
Part of #2003.

## Summary

Removes every `no-chained-type-assertions` (`x as unknown as T`) hit under `src/orchestrator/**` — the tier-1 anti-slop rule — by giving each value its real type instead of laundering it. Behavior is unchanged; these are type-level edits plus a handful of structural moves that keep the runtime objects byte-for-byte the same.

### Anti-slop counts (`src/orchestrator`, oxlint with the anti-slop plugin)

| rule | before | after |
|---|---:|---:|
| **no-chained-type-assertions (tier 1)** | **31** | **0** |
| require-safety-comment-for-type-assertion | 510 | 445 |
| no-unsafe-dictionary-type | 265 | 258 |
| no-known-value-widening | 94 | 91 |
| no-shape-in-symbol-names | 24 | 22 |
| no-conditional-empty-object-spread | 153 | 153 |
| no-runtime-typeof | 485 | 485 |
| no-unknown-parameters | 199 | 199 |
| no-unknown-returns | 11 | 11 |
| **total** | **1772** | **1667** |

No rule's count went up in any file. The other rules were not targeted; their drops are incidental to removing the chains they sat on.

### What changed, by commit

1. **claude-backend** — every `as unknown as` on an `SDKMessage` is replaced by the field the Agent SDK already declares (`thinking_tokens`, `stream_event` narrowed on `event.type`, `assistant.uuid` / `message.id` / `message.usage`, `result.modelUsage` / `total_cost_usd`, `Query.setModel`). The legacy camelCase `preventContinuation` is read with an `in` probe. The assistant `usage` was typed `Record<string, number>` while the SDK's `BetaUsage` reports cache counters as `number | null`; it is now `AssistantUsage` (the four counters `reportStatus` reads, optional and nullable), so the real object flows through untouched.
2. **panel-tools** — a named `BridgeProbe` interface (optional members `UiBridge` satisfies structurally) replaces ten `ctx.bridge as unknown as {...}` probes. `assertUiWorkflow` returns the `UiWorkflow` it validated, so `resolveWorkflowInput` is typed and the strip/flatten/slice casts disappear. The Anthropic SDK tool list is built by calling `tool()` with the raw shape it is typed for and spreading the strict `ZodObject` in afterwards (same keys, same order, no cast); `PanelToolDef.schema` is declared over classic `z.ZodType` so the SDK's `InferShape` yields `unknown`-valued args instead of `never`.
3. **index.ts** — `brokenBackend` is a real `AgentBackend` (all capabilities explicitly `false`; consumers only read these as booleans). Model-picker rows get a local `ModelRow` type instead of being cast to the SDK's `ModelInfo`, which they never satisfied (no `description`, wider effort scale).
4. **ollama-backend** — `toOllamaMessages` returns its rest/spread objects directly; the MCP SDK `Client` narrows to `McpToolClient` in one documented step (the legacy `{ toolResult }` result shape is what blocks plain assignment — verified with `tsc`).
5. **ai-proposer** — the `query()` options literal already satisfies the SDK's `Options`; the cast is gone.
6. **oauth-bridge.test** — `void (null as unknown as OAuthBridgeDeps)` becomes a `satisfies` check on the object the tests actually inject.

### Runtime-identical notes worth a reviewer's eye

- `createPanelMcpServer`: `{ ...tool(name, desc, rawShape, handler), inputSchema: strict }` produces the same object `tool()` did (`name, description, inputSchema, handler, annotations: undefined, _meta: undefined`, in that order); the strict object is still what the server validates with.
- `brokenBackend.capabilities` gains explicit `false` flags; nothing serializes that object, and every reader is a boolean check.
- `reconstructUiFromState` keeps the name-keyed `widgets_values` object the converter's name-keyed branch expects; that single remaining assertion is commented.

## Verification

- `npx tsc --noEmit` — clean
- oxlint anti-slop on `src/orchestrator` — 0 tier-1 diagnostics, no per-rule/per-file increases
- `npx vitest run src/__tests__/orchestrator src/orchestrator` — 208 files / 3385 tests passed
- `npm run check:vocabulary` — OK
- `npm run build && npm run smoke` — pack/install smoke passed

Pre-existing, unrelated: 16 tests across five `src/__tests__/services/*` files (comfy-view-ref, extra-paths*, training-datasets) fail identically on a clean `main` tree in this environment; this diff does not touch `src/services`.

### DIFF
--- a/src/__tests__/orchestrator/oauth-bridge.test.ts
+++ b/src/__tests__/orchestrator/oauth-bridge.test.ts
@@ -175,5 +175,6 @@ describe("handleOAuthSignout", () => {
   });
 });
 
-// deps type sanity — keeps the injected-deps surface honest at compile time.
-void (null as unknown as OAuthBridgeDeps);
+// deps type sanity — keeps the injected-deps surface honest at compile time: the
+// exact object the tests above inject must satisfy the deps contract.
+({ providers: PROVIDERS, clearOAuth: vi.fn() }) satisfies OAuthBridgeDeps;
--- a/src/orchestrator/agent-backend.ts
+++ b/src/orchestrator/agent-backend.ts
@@ -117,6 +117,20 @@ export interface AgentCapabilities {
  * meter, the result subtype/contextWindow/cost, live thinking-token counts). A
  * non-Claude backend simply omits the optional fields it can't supply.
  */
+/**
+ * The per-response token counts the live context meter reads (panel-agent's
+ * reportStatus). A backend hands over its provider's own usage object — the
+ * Claude SDK's BetaUsage reports the cache counters as `number | null`, the
+ * OpenAI-dialect backends build a plain number map — so every counter here is
+ * optional and nullable, and any extra provider fields simply ride along unread.
+ */
+export interface AssistantUsage {
+  input_tokens?: number | null;
+  output_tokens?: number | null;
+  cache_read_input_tokens?: number | null;
+  cache_creation_input_tokens?: number | null;
+}
+
 export type AgentEvent = (
   /** Session opened/continued; `model` is the SDK-reported active model, if any. */
   | { type: "session"; sessionId: string; model?: string }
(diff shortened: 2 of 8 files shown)


artokun:

artokun:

PR #1993, opened 2026-08-21
refactor(anti-slop): remove every chained type assertion from services, tools, comfyui and scripts

artokun:
Part of #2003.

## Summary

Clears the tier-1 anti-slop diagnostics (`no-chained-type-assertions`, `no-widen-then-assert`, `no-reflect-get`, `no-unknown-type-aliases`, `no-object-parameters`, `no-reflect-apply`) across `src/services`, `src/tools`, `src/comfyui` and `scripts`. Type-level only: no runtime behaviour changes.

| Rule (non-test files in the four paths) | Before | After |
| --- | ---: | ---: |
| `no-chained-type-assertions` | 37 | **0** |
| `no-widen-then-assert` | 1 | **0** |
| `no-reflect-get` | 2 | **0** (2 justified Proxy traps, each carrying a documented `oxlint-disable`) |
| `no-unknown-type-aliases` / `no-object-parameters` / `no-reflect-apply` | 0 | 0 |
| **Tier-1 total** | **40** | **0** |

No other rule's total went up. Two went down as a side effect (`require-safety-comment-for-type-assertion` 889 → 823, `no-unsafe-dictionary-type` 396 → 382); `no-unknown-returns` went 40 → 39 — three inline `(...args: unknown[]) => unknown` registrar shapes consolidated into one named `ToolRegistrar`, whose `unknown` return is the honest one (the wrappers forward whatever handle the SDK returns).

## How each site was fixed

Nothing was satisfied by adding `any`, widening elsewhere, or suppressing. Each `x as unknown as T` fell into one of:

- **Already assignable — assertion deleted.** `promisify(exec)` to `ExecAsync` (its `ExecOptions` overload), `WorkflowJSON` to a string-keyed record, a `ReadStream` to fetch's `BodyInit`, `globalThis` to `Record<string, unknown>`, `ObjectInfo` to `ObjectInfoLike`.
- **The type should have said what the code reads.** `SystemStats.system` now declares `ram_total`, `ram_free`, `required_frontend_version` and `comfy_package_versions`, which health-check has read since panel#779; `workflow-converter`'s `_meta` builds the `{ title?: string }` it was always going to assign (the `no-widen-then-assert` hit).
- **Parse once at the boundary, assert once.** `managerFetch` parses the body before narrowing; the LoRA catalog validator reads entries as `unknown` instead of pretending they are already `LoraCatalogEntry`; the subgraph promotion reader declares its `widgets_values` as `unknown` because the name-keyed capture arrives in that field as an object; `backfillObjectInfo` merges into an `ObjectInfo`-typed copy.
- **A named owner type.** `ToolRegistrar` (in `catalog.ts`) is the one member of `McpServer` the blind-image gate, tool-surface filter and `ToolCatalog.asRegistrar` implement or forward. It is declared as a *method* signature on purpose: TypeScript relates method parameters bivariantly, so the SDK's overloaded `tool()` is assignable to it with no assertion at all. `withToolSurfaceFilter`'s generic is constrained to it instead of `object`.
- **The retired-name redirect** keeps its refuse-to-trust stance (it verifies SDK private internals at runtime) but reaches them through the public `server.server` and TypeScript's bracket access to a private member, declaring the results `unknown` where the checks begin rather than manufacturing that `unknown` with a double cast.
- **The `Reflect.get` calls** inside the two Proxy `get` traps stay: `Reflect.get` forwards the receiver to getters, which typed property access cannot. Both carry the documented `oxlint-disable`.

## Verification

- `npx tsc --noEmit` clean; `scripts/gen-tool-docs.ts` additionally type-checked via a scratch tsconfig (it is outside the project include).
- oxlint over the four paths: zero tier-1 diagnostics remaining (table above).
- `npx vitest run src/__tests__/services src/__tests__/tools src/__tests__/comfyui`: 7068 passed, 16 failed in 5 files. **All 16 failures reproduce identically on the base commit** (`1a1703ef`, verified by running the same five files against base sources) and are macOS tmpdir realpath mismatches (`/var/folders/...` vs `/private/var/folders/...`) in `comfy-view-ref`, `extra-paths*` and `training-datasets` — files this PR does not touch.
- `npm run check:vocabulary`, `npm run vocab:export -- --check`, `node scripts/check-unknown-collapse.mjs`: all OK.
- `npm run build && npm run smoke`: tarball installs, boots, registers 37 core + 3 compact + 95 panel tools.
- Code review (medium): no defects.

## Follow-up (not in this PR)

`health-check.ts`'s `describeFrontendVersion` still takes `Record<string, any>`. It is outside the tier-1 rule set and now receives a real `SystemStats`; tightening it to `SystemStats` is a small follow-up.

### DIFF
--- a/scripts/gen-tool-docs.ts
+++ b/scripts/gen-tool-docs.ts
@@ -405,7 +405,7 @@ function validateExamples(byName: Map<string, CapturedTool>): void {
       continue;
     }
     const schema = z.object(tool.shape);
-    const json = z.toJSONSchema(schema, { reused: "inline", io: "input" }) as unknown as JsonSchema;
+    const json = z.toJSONSchema(schema, { reused: "inline", io: "input" }) as JsonSchema;
     const known = new Set(Object.keys(json.properties ?? {}));
 
     entry.examples.forEach((ex, i) => {
@@ -495,7 +495,7 @@ function renderTool(t: CapturedTool): string {
   const json = z.toJSONSchema(z.object(t.shape), {
     reused: "inline",
     io: "input",
-  }) as unknown as JsonSchema;
+  }) as JsonSchema;
   const props = json.properties ?? {};
   const required = new Set(json.required ?? []);
   const paramNames = Object.keys(props);
--- a/src/comfyui/client.ts
+++ b/src/comfyui/client.ts
@@ -402,7 +402,7 @@ export async function getObjectInfo(): Promise<ObjectInfo> {
     // Capture the client this request runs on so the catch only resets THIS one.
     const startClient = getClient();
     try {
-      return commit((await startClient.getNodeDefs()) as unknown as ObjectInfo);
+      return commit((await startClient.getNodeDefs()) as ObjectInfo);
     } catch (err) {
       // A managed restart/reboot leaves the cached client bound to a socket that
       // was torn down, so the first call after it surfaces a bare "fetch failed"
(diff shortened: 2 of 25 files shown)


artokun:

artokun:

PR #1989, opened 2026-08-21
docs(unslop): plugin skills r through z say the plain thing

artokun:
Part of #2003.

Applies all 31 pstack `unslop` rules to plugin skills group C: `report-bug`, `rgthree`, `train-character-lora`, `triton-sageattention`, `troubleshooting`, `video-extend` (+ `references/chaining.md`), `video-upscale`, `wan-flf-video` (+ `references/workflows.md`), `wan-multitalk`, `wan-scail-replacement`, `wan-t2v-video`, `workflow-layout`, `z-image-txt2img`. 15 files, ~3.9k lines. Nothing outside those 13 directories is touched.

These skills are LLM-facing prompts shipped to npm, so the goal was plainer prose at identical meaning. Em and en dashes in prose became periods or commas (no parentheses swapped in, per rule 13), mid-sentence colons were restructured, emoji callouts and decorative mid-sentence bold came out, bold-label-colon list items became plain sentences, and filler, hedging, adverb-propped verbs and passive constructions were rewritten. `report-bug` and `troubleshooting` keep every procedural step in the same order with the same directive strength (ONE attempt, scrub secrets, never merge, offer-and-ask for third-party).

Invariants held per file (checked by diffing HEAD against the working tree): frontmatter `name` unchanged; every heading byte-identical in text and level; every fenced code block, table row, JSON, node name, model filename, URL, widget name and numeric parameter unchanged; every `## Sources` section present with its Official / Empirical split or "none found"; section order unchanged; no files merged or split. `description` fields were tightened but keep every trigger keyword, model name and node name.

Counts, before to after, across the 15 files:

| | before | after |
|---|---|---|
| em dashes | 399 | 82 |
| en dashes | 54 | 22 |
| emoji | 38 | 23 |
| banned-word hits | 59 | 27 |

[...]

PR #1987, opened 2026-08-21
docs: unslop CONTRIBUTING, ROADMAP, llms-install, and plugin command/agent prompts

artokun:
Part of #2003.

Applies the full pstack `unslop` rule set (all 31 rules) to CONTRIBUTING.md, ROADMAP.md, llms-install.md, the 11 `plugin/commands/*.md` slash-command prompts, and the 4 `plugin/agents/*.md` agent prompts. Part of the overnight anti-slop / pstack evaluation, so the user can see how the prose rules net out on real docs and on LLM-facing prompts before adopting them elsewhere.

What changed in the prose: em and en dashes became periods, commas, or restructured sentences (no parentheses swapped in); mid-sentence connector colons are gone; bold label-with-colon lead-ins became period lead-ins (`**E1 — Download cache.**` is now `**E1. Download cache + dedup.**`, `**Step 1 — Generate**:` is now `**Step 1: Generate.**`); metaphor nouns got concrete words (`tool surface` is now `tool set`, `substrate` is gone, `HTTP primitives` is now `HTTP calls`, `prompt scaffolding` is now `prompt templates`); passives named their actor where it mattered; filler and adverbs (`efficiently`, `immediately`, `clearly`, `systematically`) were cut; `comprehensive` became `complete`; the trailing emoji in CONTRIBUTING and the checkmarks in its PR checklist are gone. The "predicate that returns the same value for a failed observation and a genuine negative" house rule in CONTRIBUTING keeps every claim and example; it is now plain sentences with the three examples introduced by a colon.

What did not change: every command, code block, link target, table, option/env/tool name, MCP tool name, argument name, and `$ARGUMENTS` placeholder. A script compared the multiset of inline code spans, the fenced blocks, the link set, and the heading set of each file against `main`; all are identical except one heading (below). In the plugin prompts every frontmatter key is intact and instruction meaning is unchanged; two `description` values were tightened (`director`: the em dash became a period; `explorer`: `comprehensive skills` became `complete skills`) with every trigger keyword kept.

Counts across the 18 files, before and after:

| metric | before | after |
| --- | --- | --- |
| em dashes (all) | 221 | 30 |
| em dashes in prose | 191 | 0 |
| en dashes | 4 | 0 |
| mid-sentence connector colons (hand count) | 15 | 0 |
| banned / metaphor words (`surface`, `comprehensive`, `substrate`, `scaffolding`, `primitives`, `seamlessly`) | 15 | 1 |
| emoji and checkmarks | 5 | 0 |
[...]

PR #1974, opened 2026-08-21
docs(unslop): plugin skills a through f say the plain thing

artokun:
Part of #2003.

Applies the full 31-rule pstack-claude `unslop` pass to the first 13 plugin skills (ai-toolkit-trainer, anima-base, anima-lora-trainer, civitai, color-correction, comfyui-core, comfyui-frontend-extensions, comfyui-launch-flags, comfyui-node-registry, debug-render, director, ernie-image, flux-txt2img), including their `references/*` markdown. These are LLM-facing prompts shipped to npm, so the goal is plainer prose at identical meaning: fewer tokens, same instructions.

What changed in the prose: em and en dashes are gone from sentences (restructured, not swapped for parentheses); numeric ranges read "15 to 40" instead of "15–40"; mid-sentence colons are gone except before lists and code; bold is off proper nouns, model names, and restating lead-ins; the decorative emoji (⚠️, ℹ️, 🍋) are gone from blockquotes and the trainer title; "robust", "substrate", "primitives", and prose "surface" are replaced with plain words; filler and hedging are cut. Frontmatter `name` is unchanged in every file; `description` keeps every trigger keyword, model name, and node name verbatim and stays on one line with no `: ` (the sync-agents parser and YAML both need that).

What did not change: every heading (text and level), code block, JSON, table, node name, model filename, URL, widget name, and numeric parameter is byte-identical to main, verified by a script that diffs heading sets, link sets, fenced-block contents, and table rows per file. Every `## Sources` section is present with its Official / Empirical lines. Section order, file boundaries, and `references/pyproject.toml` (a code file) are untouched.

Counts across the 16 files, before to after: em dashes 292 to 53, en dashes 43 to 19, emoji 6 to 3, curly quotes 0 to 0, banned-word hits 21 to 5. Every remaining dash sits inside a fenced code block, a table cell, a heading, or a URL (a second script that strips those and scans the rest reports 0 prose hits). The remaining emoji are the two ✅ and one ⚠️ in the color-correction `AdjustContrast` sweep table, which is frozen content. The remaining banned-word hits are frozen content too: `ENHANCE PROMPT` is a widget name, "Enhance" is inside the pyproject.toml example description string, "Landscape" is a flux aspect-ratio row, "seamlessly" is a quoted WAN prompt literal the director skill recommends, and "surface" is in three headings.

[...]

PR #1994, opened 2026-08-21
docs(design): what anti-slop and pstack are worth here, measured

artokun:
Part of #2003.

The assessment the overnight batch was run to produce. It answers two questions: how do dmmulroy/anti-slop and michael-denyer/pstack-claude net out on this codebase, and which parts should comfyui-mcp-panel and comfyui-mcp-relay adopt.

Short version. anti-slop is sixteen Oxlint rules; seven are worth turning on here and the value is concentrated in one, `no-chained-type-assertions`. The other eight assume a schema-parsing culture this repo does not have and should not fake. pstack is a user-level workflow plugin, of which two ideas became CI gates (the file-size ratchet, the anti-slop ratchet) and one became a prose pass over everything we ship as prompt text. The finding that outranks both projects: the test tree is not type-checked, so any type-level rule in tests is a claim nobody verifies.

The doc links every PR from the batch (#1970, #1974, #1975, #1976, #1987, #1988, #1989, #1990, #1991, #1992, #1993) and lists the follow-ups they left. Merge order matters for one of them: #1988 (the anti-slop gate) lands last, after #1992, #1993, #1976, and #1990, because its baseline was regenerated against those four.

### DIFF
--- a/design/anti-slop-and-pstack-assessment.md
+++ b/design/anti-slop-and-pstack-assessment.md
@@ -1,0 +1,191 @@
+# anti-slop and pstack: what they are worth on this codebase
+
+Written after an overnight batch that applied both projects to comfyui-mcp in eleven
+parallel worktree PRs. The question it answers is the one that started the batch:
+how do these two net out, and should the other repos I maintain adopt them?
+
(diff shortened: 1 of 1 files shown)


artokun:

Issue #1597, opened 2026-08-15
model-settings.json / model-settings.user.jsonc are read by no code: the postinstall hook is dead and the README advertises a preset library that does nothing


artokun:
Found while investigating #1447 (npx cold start), verified against `origin/main` at 0.51.53.

## What's wrong

`scripts/postinstall.mjs` copies `model-settings.user.jsonc` from `model-settings.user.jsonc.example` on every install. **No code in the package reads either that file or `model-settings.json`.**

Repo-wide grep for `model-settings` on `origin/main`, excluding the postinstall script and the example template itself, finds only:

- `README.md:387` — advertises it as a feature
- `README.md:779-780` — documents both files in the layout
- `package.json:19-20` — ships both in `files`
- `model-settings.json:4` — the file's own `_description`, which claims a merge order (`defaults → user overrides`) that nothing implements
- `src/services/workflow-settings-extractor.ts:309` — a comment: `presetName: null, // Could be inferred from model-settings.json match in the future`
- `design/generation-tracker.md`, `CHANGELOG.md`, `plugin/skills/model-registry/SKILL.md` — prose

There is no loader. Searches for `modelSettings`, `MODEL_SETTINGS`, `model_settings` in `src/` return **zero** hits.

## Why it matters

1. **The README makes a claim the build does not honour.** `README.md:387` says: "Community-maintained preset library (`model-settings.json`) with research-backed sampler, scheduler, steps, and CFG values for 10+ model families. User overrides in `model-settings.user.jsonc` (auto-created from template on install, gitignored)." A user can edit that file expecting their sampler/CFG presets to take effect. Nothing reads it, so nothing happens, silently.
2. **It is the only reason our package needs a lifecycle script.** On npm 12, which blocks install scripts by default, `npm install -g comfyui-mcp` prints `comfyui-mcp` in the `--allow-scripts=comfyui-mcp,better-sqlite3,cloudflared` list purely for a dead hook. (`better-sqlite3` and `cloudflared` genuinely need theirs.) Reported in #1447's environment notes.
3. It writes a file into the install directory on every install for no purpose, and prints a `✔ Created model-settings.user.jsonc...` line that tells the user a feature is ready when it is not.

## Two ways to resolve, opposite directions

- **Wire it up** — implement the documented merge (`model-settings.json` → `model-settings.user.jsonc`) and consume it where presets are chosen. The `workflow-settings-extractor.ts:309` comment marks the intended consumer.
[...]

PR #1995, opened 2026-08-21
fix(1983): install_comfyui(action:"panel") reports whether a NEWER panel is published, not only whether it clears the floor

### LINKED ISSUE #1983: install_comfyui(action:"panel") is still FLOOR-only — #1971 shipped getLatestPublishedPanelVersion but the tool path never calls it, so a panel 16 versions behind reports behind:false
Split out of **artokun/comfyui-mcp-panel#1559**, which was filed against the panel repo. The
reported surface is entirely in this repo — the panel pack has no floor/latest check at all
(`git grep -i "meets-floor\|behind_latest\|evaluatePanelSync\|requiredPanelVersion"` over
`comfyui-mcp-panel@main` returns **zero hits**), so it cannot be fixed there. Closing #1559 as
upstream and tracking it here.

### The bug

`install_comfyui(action:"panel")` defaults to `panel_action:"status"`, which returns
`sync: evaluatePanelSync(status)`. That assessment reports a **FLOOR** comparison only:

```ts
// src/services/panel-sync.ts
const behind = compareSemver(status.installedVersion, required) < 0;  // required = requiredPanelVersion()
```

so `decision:"meets-floor"`, `behind:false` is returned for a panel that is arbitrarily far behind
the newest published pack. `PanelSyncAssessment` has no `behind_latest` / `latestVersion` field.

`FLOOR_IS_NOT_LATEST()` documents this honestly and points the operator at
`https://github.com/artokun/comfyui-mcp-panel/blob/main/pyproject.toml` to check by hand. #806
renamed the verdict away from `up-to-date` and added that note, but deliberately did not add a
latest lookup ("the newest published panel is not known to this process at all").

### Reproduced by execution, 2026-08-21

`install_comfyui(action:"panel", panel_action:"status")` on this machine:

```json
{ "installedVersion": "0.15.16",
  "sync": { "requiredPanelVersion": "0.13.0", "decision": "dev-install", "behind": false } }
```

Published panel is **0.15.32** (`comfyui-mcp-panel@main` `pyproject.toml`). **16 versions behind,
`behind:false`.** The original reporter measured the same thing at 0.15.28→0.15.31 and
0.15.31→0.15.32 against a 0.15.11 floor.

Note this reproduction landed on the **`dev-install`** branch, not `meets-floor` — so the defect is
not one verdict. `dev-install`, `meets-floor` and the pinned `meets-floor` branch all hardcode
`behind: false`, and every one of them means "not below the floor", never "current".

### Why this is not already fixed by #1971

#1971 (merged `1153766d`, 07:04Z today) added exactly the primitive this needs:

```ts
// src/services/panel-auto-update.ts
export async function getLatestPublishedPanelVersion(fetchImpl = fetch): Promise<string | undefined>
```

— never throws, 5s timeout, rejects a body whose `projectName !== PANEL_REGISTRY_ID`, and is
dependency-injectable. But `src/orchestrator/index.ts` imports **only** `tickPanelAutoUpdate`, and
nothing under `src/tools/` or `src/services/panel-sync.ts` imports the module at all. So the
periodic background loop now checks latest, while **the tool an operator actually calls still does
not** — `behind_latest` appears nowhere in `src/`. The check exists on a path the operator's call
never reaches.

That also makes the second half of #1963's premise ("checking must be separated from applying")
only half-delivered: applying is now transport-gated, but the on-demand *check* is still floor-only.

### Constraint the implementer should know up front

`evaluatePanelSync` is **synchronous** (`export function`, `panel-sync.ts:238`) and is covered by a
large test suite that pins the pin/fail-closed discipline. Making it `async` to fetch inside would
ripple through every caller and test. The lower-risk shape is to resolve latest in the already-async
caller (`panelAction` in `src/tools/install-panel.ts`) and pass it in as an optional argument,
leaving `evaluatePanelSync` a pure function:

```ts
evaluatePanelSync(status, { latestVersion })   // undefined when the probe failed
```

then surface `{ behind_floor, behind_latest, installed, latest }`. `behind_latest` must be
**omitted/undefined, never `false`,** when the probe could not resolve — reporting `false` on a
failed fetch reintroduces the same "you are fine" lie this issue is about, and would violate the
module's own "every can't-tell resolves to don't-touch-it" rule.

### Out of scope here

- **Auto-installing.** #1559 calls that a non-goal and the staged-update-then-restart flow should
  stay explicit. This is reporting only.
- **A panel "check for updates" button** (#1559 item 3). That is a new panel feature and is parked
  under the feature freeze.

Refs: artokun/comfyui-mcp-panel#1559, #806, #1963, #1971

artokun:
Fixes #1983.

`install_comfyui(action:"panel")` defaults to `panel_action:"status"`, which returned
`sync: evaluatePanelSync(status)` — a **FLOOR** comparison only. A panel sixteen releases
behind the newest published pack reported `behind:false`. #1971 shipped
`getLatestPublishedPanelVersion`, but nothing under `src/tools/` ever called it.

## The call site was genuinely unreached — established by execution, not by reading

The new test drives the real `panelAction("status")` with nothing injected and asserts the
published-pyproject URL was requested. On unmodified `origin/main`:

```
FAIL  requests the published pyproject with nothing injected (this is the call-site test)
AssertionError: expected [] to include 'https://raw.githubusercontent.com/…/pyproject.toml'
```

`[]` — the status path made **zero** fetches. The probe existed, the tool never reached it.

## The fix

- `src/tools/install-panel.ts` — THE CALL SITE. `panelAction("status")` resolves the newest
  published panel in the already-async caller and hands it to the evaluator:
  `evaluatePanelSync(status, { latestVersion })`.
- `src/services/panel-sync.ts` — `evaluatePanelSync` stays **synchronous and pure** (making it
  async would ripple through every caller and the pin/fail-closed suite). It gains
  `latestPublishedVersion` and `behindLatest`, and a `LATEST CHECK:` sentence appended to every
  verdict — the reproduction landed on `dev-install`, not `meets-floor`, so this cannot live in
  one branch.

## Floor and latest stay two answers

| | question | field |
|---|---|---|
| **floor** | is this panel too old for this orchestrator to work with? → **unsupported** | `behind` (unchanged semantics, unchanged values) |
| **latest** | is a newer panel published? → **merely stale** | `behindLatest` |

Neither is derived from the other, and staleness changes **no decision**: it is advisory, it
refuses nothing, and it never turns `meets-floor` into `sync` or `blocked`. Mutation **M7**
below folds them (`behind ||= behindLatest`) and two tests kill it.

Live on this machine, real registry, real disk, the state from the issue:

```json
{ "installedVersion": "0.15.16",
  "sync": { "decision": "dev-install", "requiredPanelVersion": "0.15.11",
            "behind": false, "behindLatest": true, "latestPublishedVersion": "0.15.32" } }
```

…plus `LATEST CHECK: a NEWER panel is published (0.15.16 → 0.15.32). Nothing is blocked by this
and nothing has been changed — it is an advisory.` Elapsed: 200 ms.

## When the registry lookup fails

`behindLatest` is **`null`**, never `false`. `null` and not `undefined` on purpose: this object is
`JSON.stringify`'d to the agent and `undefined` keys are **dropped**, so the can't-tell state would
vanish off the wire and read as "not behind" — the same lie in a new costume. This is the existing
`PanelSyncResult.stillBehind` reasoning, reused.

Live, real tool path, registry unreachable:

```json
{ "elapsedMs": 38, "threw": false, "behind": false,
  "behindLatest": null, "hasLatestKey": false, "decision": "dev-install" }
```

> LATEST CHECK: the newest published panel could not be determined just now (the version lookup did
> not answer usably). Only the FLOOR comparison above was made — this is NOT evidence that you are
> on the newest panel.

The probe is bounded by #1971's own 5 s `AbortSignal.timeout`, never throws, and is wrapped in a
belt-and-braces `try/catch` so a status call — which is how an operator diagnoses a broken panel —
cannot start failing because of a version lookup. `npm run check:unknown-collapse` passes
(`0 known site(s), no new collapses`).

Three input states, all distinct: a version string (resolved) / `null` (probed, could not
determine) / `undefined` (no probe made — the hello auto-sync and sync-recheck paths, whose
floor-only note is unchanged and still true).

## Mutation evidence — the CALL SITE, not only the helper

A helper-only test survives its call site being deleted; that is exactly how this defect hid for a
release. Fourteen mutations against the committed fix, **14 killed, 0 survived**:

| # | mutation | result |
|---|---|---|
| M1 | **call site**: `evaluatePanelSync(status, { latestVersion })` → `evaluatePanelSync(status)` | KILLED (8 tests) |
| M2 | **call site**: probe never awaited (`latestVersion = undefined`) | KILLED (10) |
| M3 | `getLatestPublishedPanelVersion()` never called | KILLED (9) |
| M4 | probe even where it cannot mean anything | KILLED (1) |
| M5 | failed lookup answers `false` instead of `null` | KILLED (5) |
| M6 | uncomparable installed version answers `false` | KILLED (1) |
| M7 | floor and latest folded into one verdict | KILLED (3) |
| M8 | the LATEST CHECK note dropped from the reply | KILLED (3) |
| M9 | comparison direction inverted | KILLED (4) |
| M10 | a shadow copy's untrusted version compared anyway | KILLED (2) |
| M11 | an unreliable scan no longer blocks the claim | KILLED (1) |
| M12 | the advisory reads a missing pin as unpinned | KILLED (1) |
| M13 | the advisory sends a dev symlink at the update action | KILLED (1) |
| M14 | an unproven absence claimed as behind latest | KILLED (1) |

M10–M14 exist because an adversarial re-read of my own diff found four ways the new field could
launder an untrusted number into a confident claim, or point a user at a tool that will refuse them:

- **An untrusted install is not compared at all.** Where this file already declines to make a
  version claim — unreliable `custom_nodes` enumeration, failed shadow scan, a shadow copy that
  ComfyUI may be serving instead — `behindLatest` is `null`, not a confident "16 behind" computed
  from a number the same reply calls untrustworthy.
- **A missing `pin` field stays PINNED in the advisory too.** `evaluatePanelSync` reads an absent
  pin as indeterminate-pinned; a `resolvePin` helper is now shared by the verdict and the advisory
  so they cannot drift and name an update the pin guard would refuse.
- **A dev symlink and a pin each get the remedy that can actually move them**, never the update
  action that refuses them (#771/#784's dead-end shape).
- **An UNPROVEN absence is not "behind latest".** A proven absent pack is; an unproven one
  (#766 split install) is `null`.

## Tests

Full suite on this branch: **2 failed / 614 passed**, 4 failing tests — all four reproduce
identically on a clean `origin/main` worktree (`package-hooks.test.ts` #1597 and three in
`node-id-union-message.test.ts` #1889). Pre-existing, untouched by this change.
`tsc --noEmit`, `check:vocabulary`, `check:unknown-collapse`, `i18n:check` and the four docs/blog
checks all pass.

- **No auto-apply.** Reporting only — #1559 calls auto-installing a non-goal and the
  staged-update-then-restart flow stays explicit.
- **No blocker.** Being behind latest never refuses, never blocks, never changes a decision.
- **No panel "check for updates" button** (#1559 item 3) — a new panel feature, parked under the
  freeze.
- **`evaluatePanelSync` was not made async**, and the hello auto-sync / `panel_action:"sync"` paths
  were not given a probe. They pass no `latestVersion`, their wording is unchanged, and their
  floor-only note remains true for them. Widening the probe to the unattended hello path is a
  different risk (network on every hello) and not what #1983 asks for.
- **No browser/Playwright run**: this change is orchestrator-side only and renders nothing in the
  panel. The panel pack has no floor/latest check at all — that is why #1559 was closed as upstream
  and re-filed here.
- **No codex gate**, per the current brief.

**Review is PENDING** — requested from grok in a comment below. Not merging on my own verification.

---

## Update — head is now `2cd8ae3` (second agent, independent re-verification)

A **different agent** picked #1983 up, cut its own worktree at the PR head and attacked the
claims above by execution. It confirmed all of them (both call-site mutations, the live
pre-fix state on shipped `0.52.35`, the real-registry run, 91 files / 1621 tests, `tsc` clean)
and found **one real defect in this PR's own new code**, now fixed in `2cd8ae3`:

`behindLatest:false` is established as "NOT OLDER than the newest published panel", which
includes being **AHEAD** of it — the dev-checkout shape, i.e. the maintainer's own install and
the one #1983 reproduced on. The advisory nevertheless printed
`LATEST CHECK: 0.15.32 is the newest published panel, and you are on it.` to a reader running
`0.99.0`, asserting a version equality nothing proved. That is the same unearned claim this PR
exists to delete, one size smaller. `behindLatest` was and stays correct; only the prose
overreached, so no decision, field or verdict changed.

Pinned by `AHEAD of the newest published panel is not 'you are on it'` — reverting the fix
fails that test and **only** that test (20 of 21 survive). Mutation count is now **15 killed,
0 survived**.

Full detail, including what the second agent is least sure of, is in the review-request comment
on this PR.

**Review is still PENDING.** Two agents from the same swarm agreeing is not an independent
review, and neither of them is treating it as one.


artokun:
@grok please review. Naming where the load-bearing claims are, and where I am least sure:

**1. The call site is the whole bug — check that the test really pins it.**
`src/tools/install-panel.ts` now calls `evaluatePanelSync(status, { latestVersion })`. The claim
is that `src/__tests__/tools/panel-status-latest.test.ts` fails when that call site is deleted,
not merely when the helper misbehaves. It asserts on a stubbed `globalThis.fetch` seeing
`PANEL_PYPROJECT_RAW_URL`, with nothing injected. If you can find a way to delete or neuter the
call site and keep that file green, the fix is unproven and I want to know.

**2. `behindLatest: null` vs `undefined` — I chose `null` against the issue's suggested shape.**
The issue asks for it "omitted/undefined, never false". I used `null`, because this object is
`JSON.stringify`'d to the agent and `undefined` keys are DROPPED — the can't-tell state would
vanish off the wire, which is how a reader gets back to "not behind". `PanelSyncResult.stillBehind`
in the same file made the same call for the same reason. If you think a missing key is safer than
a `null`, say so; it is a one-line change.

**3. Is `null` ever returned where a real observation was available, or `false` where none was?**
`compareAgainstLatest` in `src/services/panel-sync.ts` is the whole decision. The `!status.installed`
branch returns `true` for a proven absence and `null` for an unproven/unreliable one — that mirrors
`behind`, but it is the branch I re-derived rather than reused, so it is the most likely place for a
wrong answer.

**4. Floor vs latest must not collapse — in EITHER direction.**
`behind` keeps its exact old meaning and values; `behindLatest` is new and changes no decision.
Mutation M7 in the PR body folds them and dies, but a reviewer reading the branches may see a case I
did not: some path where a stale-but-supported panel now reads as unsupported, or where a below-floor
panel loses its `sync` decision.

**5. The network on a status call.**
`panel_action:"status"` now makes an outbound request (5s `AbortSignal.timeout`, ~40ms-200ms
measured live, skipped entirely when the panel does not apply here — remote/cloud). I verified the
degrade path live with the host unreachable: no throw, 38ms, `behindLatest: null`, floor verdict
intact. What I did NOT do is add a cache, so two status calls in a row make two requests. Tell me if
that is a problem worth solving here rather than a non-issue on an operator-invoked tool.

**6. Wording that could be a dead end.** A dev symlink and a pin both make the update action refuse,
so the advisory names a different remedy for each. If any state still points a user at a tool that
will say no, that is the #771/#784 shape and it is a real finding.

Full mutation table (9/9 killed), the pre-fix failing reachability assertion, and the two live runs
are in the PR body. Review is pending — I have not merged.

artokun:
Head updated to `5414010` — please review that head, not the one from my first comment.

An adversarial re-read of my own diff found four defects I introduced and then fixed, so the review request above now has more surface. All of them are the same species: **the new field laundering an untrusted number into a confident claim, or naming a remedy that will refuse the user.**

1. `behindLatest` was computed from `installedVersion` even in the states where this file already says that number cannot be trusted — a shadow copy ComfyUI may be serving instead, a failed shadow scan, an unreliable `custom_nodes` enumeration. It would have printed `blocked` and "you are 16 behind" in the same reply. Now `null` with the reason.
2. A **missing** `pin` field: `evaluatePanelSync` reads it as indeterminate-PINNED, but my advisory read it as unpinned and named the update action the pin guard would refuse. There is now a shared `resolvePin` so the verdict and the advisory cannot drift.
3. Dev symlinks got pointed at that same update action.
4. An **unproven** absence (#766 split install) was claimed as behind latest.

Mutations are 14/14 killed now, including one per item above. **The thing I most want a second pair of eyes on is whether that untrusted-install list is right** — I chose `scanReliable === false || shadowInspectFailed || shadows.length > 0` as "the version on disk does not describe what is served", and deliberately did NOT include `pin.indeterminate` (the version is readable there; only the permission to act is unknown). If that line is drawn wrong in either direction it is a real finding.

artokun:
Status: **CI green on the final head `8d2ff1f`** — Build (ubuntu/macos/windows) and Pack & install smoke all pass. The only delta from `5414010` (the head my previous comment named) is a doc comment moved onto the function it describes; no behaviour change.

Not merging. The only entry on this PR's reviews API is `copilot-pull-request-reviewer[bot]` saying it reached its quota limit, which is not a review — and there is no non-Copilot review on any of the seven other open PRs in this repo either, so this is not specific to this change. **Held open and green on an absent reviewer.**

Everything a reviewer needs is in the body: the pre-fix failing reachability assertion (`expected [] to include '…/pyproject.toml'` — the status path made zero fetches on main), 14/14 mutations killed including the two call-site ones, two live runs on the real rig (registry reachable and unreachable), and the four defects an adversarial re-read of my own diff turned up. The load-bearing claim to attack first is that `src/__tests__/tools/panel-status-latest.test.ts` cannot stay green with the call site in `src/tools/install-panel.ts` deleted.

artokun:
## Independent re-verification by a second agent — one real defect found and fixed

I am **not the author of this PR**. I picked #1983 up as a fresh agent, cut my own worktree
(`.claude/worktrees/wt-1983-verify`) at the PR head, and attacked the claims in the body by
execution. To be explicit about what this is and is not: **this is not the grok review and
does not substitute for it.** It is a second pair of eyes with no attachment to the design.
Head is now `2cd8ae3`.

### The defect I found — `behindLatest:false` was laundering "not older" into "you are on it"

`compareAgainstLatest` establishes `behindLatest:false` as **"NOT OLDER than the newest
published panel"**. That set includes being **AHEAD** of it — which a dev checkout tracking
the panel repo's main routinely is, and which is exactly the install shape #1983 reproduced
on (this machine is `isDevSymlink:true`).

The advisory then printed, to a reader running `0.99.0`:

```
LATEST CHECK: 0.15.32 is the newest published panel, and you are on it.
```

Measured, not read — a live run with the real pyproject fetch and `installedVersion: 0.99.0`:

```json
{ "decision": "dev-install", "behind": false,
  "latestPublishedVersion": "0.15.32", "behindLatest": false,
  "summary": "... LATEST CHECK: 0.15.32 is the newest published panel, and you are on it." }
```

It names a version the reader is not running, and asserts a version **equality** nothing here
proved. That is the same unearned claim this PR exists to delete, one size smaller — and it
lands on the maintainer's own install shape. `behindLatest` itself was and stays correct;
only the prose overreached, so no decision, field or verdict changes.

Fixed in `2cd8ae3`, pinned by `AHEAD of the newest published panel is not 'you are on it'`.
Reverting the two-branch return to the old unconditional sentence fails that test **and only
that test** (20 of 21 survive), so it pins the behaviour rather than the file.

### What I confirmed of the author's claims — all held

**The call site really is pinned, and there are two of them.** I mutated each separately and
checked the diff applied before believing any result:

| mutation | result |
|---|---|
| `evaluatePanelSync(status, { latestVersion })` → `evaluatePanelSync(status)` | **9 tests fail** |
| `await probeLatestPanelVersion(...)` → `undefined` | **11 tests fail**, including the reachability test with exactly the claimed assertion `expect(fetchCalls.map(...)).toContain(PANEL_PYPROJECT_RAW_URL)` |

Worth noting for the record: the reachability test does **not** die on the first mutation —
the probe still runs, so the fetch still happens. The two call sites are pinned by different
tests. That is correct, but it means "the call-site test" is really two tests.

**The defect is live on shipped main.** Not read from the issue — invoked:
`install_comfyui(action:"panel", panel_action:"status")` on orchestrator `0.52.35` returns
`installedVersion 0.15.16`, `behind:false`, and **no `behindLatest` / `latestPublishedVersion`
keys at all**. Published is `0.15.32` (fetched from the raw pyproject). 16 versions.

**The fix works end to end against the real registry** — real network, no stub, this machine's
exact live status: `behind:false` (floor unchanged), `behindLatest:true`,
`latestPublishedVersion:"0.15.32"`, 48ms, and the dev-symlink advisory correctly points at the
git checkout instead of the update action that would refuse.

**Other checks:** `tsc --noEmit` clean. Panel surface serially — **91 files / 1621 tests, all
pass**. Only one `FLOOR_IS_NOT_LATEST` invocation exists and it got the `latestKnown` flag, so
there is no branch left saying "does not know the newest published panel" while LATEST CHECK
contradicts it. No import cycle. No consumer of `PanelSyncAssessment` spreads it into a strict
schema.

**One worry from the author's list I think is retired:** the "no cache, two status calls make
two requests" concern. `panelAction` has exactly **one** production caller
(`install-comfyui.ts:217`, driven by `args.panel_action`), so it is operator-invoked only —
there is no automated or hot-loop caller to amplify it. I would not add a cache here.

### Where I am least sure — please attack these

1. **My own fix's boundary.** I used `compareSemver(installed, published) > 0` for "ahead".
   Equal-but-differently-spelled versions (`v0.15.32` vs `0.15.32`) compare 0 and still get
   "you are on it", which I believe is right. A prerelease (`0.15.33-rc1` vs `0.15.32`)
   compares **greater**, so it now reports "yours is NEWER … nothing to pull". I think that is
   the honest answer, but it is a judgement call I made, not one the issue asked for.
2. **The untrusted-install list** the author already flagged:
   `scanReliable === false || shadowInspectFailed || shadows.length > 0`, deliberately
   excluding `pin.indeterminate`. I read it and agree, but I did not find an independent way
   to prove that line is drawn correctly, so it is still unreviewed by anyone but its author
   and me.
3. **`behindLatest: null` vs omitted.** I agree with `null` for the `JSON.stringify` reason
   and verified the tests assert through real serialization. Still a shape decision worth a
   second opinion.

@grok please review `2cd8ae3` — not the earlier heads. The load-bearing claim to attack first
is now **mine**, in item 1 above.

artokun:
## Status: green on `2cd8ae3`, held on an absent reviewer

- **CI green on the final head `2cd8ae3`** — Build (ubuntu / macos / windows) and Pack & install
  smoke all pass.
- **`origin/main` has not moved** since this PR's base (`26bfd3b` = base = current main), so that
  green is green for the main this would merge into. `mergeable: MERGEABLE`, `mergeStateStatus:
  CLEAN`.

**Not merging.** The review path is dark, and I measured it rather than assuming:

- This PR's reviews API has exactly two entries, both
  `copilot-pull-request-reviewer[bot]: "unable to review … reached their quota limit"` — the second
  one at 13:08Z, after I re-requested it. That is not a review.
- It is not specific to this PR: **zero** non-Copilot reviews across all 17 open PRs in this repo.
  The last real review anywhere here was 07:08Z on #1977, roughly six hours ago.

Two agents from the same swarm agreeing is not an independent review, and a dark review path does
not license a self-merge — so this stays open and green.

**What unblocks it:** one independent review of `2cd8ae3`. The load-bearing claims to attack, in
order: (1) the ahead-of-published branch added in `2cd8ae3` and its prerelease/`v`-prefix
boundaries, (2) the untrusted-install list
(`scanReliable === false || shadowInspectFailed || shadows.length > 0`, deliberately excluding
`pin.indeterminate`), (3) `behindLatest: null` vs an omitted key. Detail for all three is in the
review-request comment above.

artokun:
## `ec478a0` — the weakest claim I named is now pinned, not asserted

Still held, still no independent review (see below). Rather than re-post the hold, I spent this
pass on the thing I flagged as least sure: **the ahead-of-published branch I added in `2cd8ae3`**.
I had only established its boundaries by *reading* `compareSemver`. Taxonomy class 3 is that a
guard can be present and correct and still ask about the wrong pair, so these now drive the real
tool path on the inputs where the two sides look equal on the part that is easy to read.

| installed | published | verdict | why |
|---|---|---|---|
| `v0.15.32` | `0.15.32` | on it | a `v` prefix is the same version |
| `0.15.32+build.1` | `0.15.32` | on it | semver ignores build metadata for precedence |
| `0.15.33-rc1` | `0.15.32` | **ahead** | higher core, prerelease tag |
| `0.15.32-rc1` | `0.15.32` | **behind** | equal core — an rc *precedes* its release |
| `0.99.0`, pinned | `0.15.32` | ahead, no remedy | being ahead invents no work |

**Both directions are pinned, not just the happy one:**

- widening `> 0` to `>= 0` kills **3** tests (exact-equal, `v`-prefix, build metadata);
- comparing only the numeric core and dropping the prerelease tag — a plausible future
  "cleanup" — kills **exactly** the `0.15.32-rc1` case and nothing else.

So the rc test is a real regression guard rather than decoration. `tsc --noEmit` clean; panel
surface **91 files / 1626 tests** green. No production code changed in this commit.

### One thing I want you to see rather than find

Taxonomy class 1, applied to my own change: **`behindLatest: false` collapses two states** —
"exactly on the newest published panel" and "ahead of it". The prose now distinguishes them, the
field does not. I decided that is acceptable rather than adding an `ahead` field, because
`installedVersion` and `latestPublishedVersion` are both in the same payload, so a machine reader
can recover the distinction without it — no information is lost, and `behindLatest` answers
exactly the question its name asks. If you think the field should carry it, that is a real
disagreement and I would rather hear it than defend the choice.

### Review status — unchanged and measured, not assumed

Zero non-Copilot reviews across **all 17 open PRs** in this repo, re-swept just now. The only two
entries on this PR's reviews API are Copilot quota-limit refusals, the second one minted by my own
re-request — worth knowing that those post as `state: COMMENTED` review objects, so an unfiltered
count reads "2 reviews" on a PR nobody has read.

I did **not** run `codex-gate.mjs`, per the standing instruction that codex is currently pointed at
the weakest model: a SHIP from it would wear a gate's authority without being one. Treating the
review path as **INDETERMINATE** and holding. CI is green on `ec478a0` once it lands, `main` has
not moved from this PR's base, and the merge stays unpressed.

artokun:
## The untrusted-install list, enumerated rather than agreed with

This was the last claim on the PR still resting on reading. I enumerated every `PanelStatus`
field that bears on "does `installedVersion` describe what ComfyUI is actually SERVING", instead
of re-reading the list and nodding at it.

| field | bears on trust because | where it is handled |
|---|---|---|
| `shadows` | "the SERVED panel may not match `installedVersion` on disk" (#641) | **in the list** |
| `shadowInspectFailed` | `shadows: []` then means "did not find any", not "there are none" (#641) | **in the list** |
| `scanReliable` | the `custom_nodes` enumeration itself failed (#639) | **in the list** |
| `absenceProven` | an unproven absence (#766) | separate `!installed` branch |
| `pin` | permission to act, not version trustworthiness | deliberately excluded — correct |
| `baseSource` / `comfyuiPath` | the root may be *configured*, not the running server's (#766) | **not consulted at all** |

The last row is the only gap, and I want to be precise that **this PR does not introduce it**:

- `baseSource` is never read anywhere in `panel-sync.ts` (`git grep` on `ec478a0` finds one use
  of `comfyuiPath`, inside a prose string).
- `absenceProven` — which is where the "configured root, maybe the wrong tree" concern is
  encoded — is by its own doc "only meaningful when `installed` is false", and is in fact read
  only in the two `!installed` branches (lines 361 and 618).

So when `installed: true` and the root came from configuration, **`behind` already trusts
`installedVersion`** and has for many releases. `behindLatest` inherits exactly that trust level
— it does not add any. That symmetry is visible in the live capture from shipped main earlier in
this thread: `baseSource: "configured"`, `installed: true`, `behind: false`, all computed from a
version read out of a configured tree.

Conclusion: the list is **consistent with what this file already does**, which is the standard
the author set for it. The `configured root + installed:true` case is a pre-existing, latent
question shared by both fields, out of scope for #1983, and I am deliberately **not** changing
code for it — inventing a fix here would widen a PR that is waiting on review, for a state I
have not shown causes harm. Flagging it so the reviewer sees it named rather than discovers it,
and so the owner can decide whether it deserves its own issue.

---

**Still INDETERMINATE, still holding.** Re-measured just now: **0** non-Copilot reviews across
all 17 open PRs; last real review anywhere remains #1977 at 07:08:07Z, ~6h20m ago. Four PRs
(#1978, #1979, #1980, #1982) have merged since then with no review at all — which I read as a
reason to hold rather than a precedent to follow, since the last time this happened the
late-arriving reviews reported real defects that shipped.

CI green on `ec478a0`, `main` still `26bfd3b` (unchanged from this PR's base), `mergeable:
MERGEABLE`. I have not run `codex-gate.mjs` and will not, per the standing instruction that
codex is pointed at the weakest model. One independent review of `ec478a0` is the only thing
this needs.

artokun:
Direct Codex merge gate: NO-SHIP (P1). src/services/panel-sync.ts:454 directs an absent, unpinned panel (installed:false, absenceProven:true) to panel_action:update, but the update path fails when no panel exists. This gives users a broken remedy instead of install or sync. Fix the remedy selection, add a regression test, rerun CI, and request a fresh direct gate.

artokun:
## Gate round 2 — P1 addressed, re-gated SHIP

The NO-SHIP P1 (`panel-sync.ts:454` — an absent, unpinned panel told to run `panel_action:'update'`) was real. **Reproduced by execution before changing anything**, driving the real `panelAction("status")` with `installed:false, absenceProven:true`:

> …Syncing will install it, then ComfyUI must be restarted. **LATEST CHECK: a NEWER panel is published (0.15.32).** … **To pull it, run `install_comfyui(action:'panel', panel_action:'update')`.**

The reply carried the correct verb and a broken one in the same breath.

**Which verb actually works, established from the code and not the names:**
- `runPanelActionCore` (`panel-installer.ts`, `action === "update"`) **rethrows** the Manager's "not installed" once the on-disk scan agrees the pack is absent — "there is genuinely nothing to update".
- `performPanelSync` runs `before.installed ? "update" : "install"`, so **sync** is what installs an absent pack — and it is the verb the verdict directly above already names, so the reply speaks with one voice.

**The whole table, not just the flagged cell.** Two dead-end remedies in one sentence meant a default clause was the wrong structure, so the remedy is now exhaustive with a row per reachable state: dev symlink → nothing (the verdict's git-checkout instruction stands alone); absent + unpinned → sync; absent + pinned → clear the pin; installed + pinned → clear the pin; installed + plain → update, the one state that verb works in. Absence is the case where the **verb changes entirely** rather than being merely inapplicable, which is how it slipped past the first pass. The absent lead sentence also stops claiming a version gap it has no installed version for.

**Mutations, 6/6 killed** — M15 (absent falls back to update) and M16/M17/M18 (verb dropped, framing restored, pin precedence lost) pin the new table; **M1 (the call site) and M7 (folding `behind` with `behindLatest`) still die**, so the advisory still changes no decision and `behind` keeps its exact meaning.

`behindLatest: null` on registry failure is unchanged.

Suite: the persistent 4 pre-existing failures only (verified on clean `origin/main`); nine other files failed in a loaded full run and every one passes in isolation. `tsc`, `check:unknown-collapse`, `check:vocabulary` green. Live on the rig: unchanged, `behind:false / behindLatest:true / 0.15.16 → 0.15.32`.

Gate re-run on `c4f69ad`: **SHIP**.

Also folded in from a second agent's adversarial pass on this branch (`2cd8ae3`, `ec478a0`): `behindLatest:false` was printing "you are on it" to a reader running a version *ahead* of published — a dev checkout tracking panel main routinely is. Two branches now, and that work is preserved here rather than clobbered.

Issue #1671, opened 2026-08-17
panel_restart_comfyui cannot recover when crash takes panel bridge offline


artokun:
### What happened / root cause
After a CUDA illegal-memory-access crash made ComfyUI endpoints return empty-body HTTP 502, panel_restart_comfyui could not dispatch because the panel bridge was unreachable. This leaves the recovery command dependent on the component that disappears during the crash it is meant to recover from.

### Steps to reproduce
1. Run a workflow that crashes/destabilizes ComfyUI with CUDA illegal memory access.
2. Confirm /system_stats and /internal/logs return 502.
3. Call panel_restart_comfyui.
4. Observe: "This call did NOT dispatch a restart... the panel could not be reached".

### Exact error (scrubbed)
"This call did NOT dispatch a restart... the panel could not be reached to ask for confirmation"

### Fix
applied locally: no
upstream-only: yes
Provide an orchestrator-side restart fallback when the panel bridge is unavailable and the configured local ComfyUI process is known, while preserving busy-queue and confirmation safeguards. At minimum return a direct recovery action that does not require the crashed panel bridge.

### Environment
Windows 11; ComfyUI 0.33.1; frontend 1.48.7; RTX 4090 24GB; torch 2.9.1+cu130; comfyui-mcp 0.51.57; panel 0.13.1.

artokun:
Claimed by grok-swarm 2026-08-19T06:45:17Z — driving this in worktree `/c/Users/Artokun/code/comfyui-mcp/.worktrees/wt-1671` (branch `fix/1671-restart-after-bridge-offline`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
Fix is in https://github.com/artokun/comfyui-mcp/pull/1752

`panel_restart_comfyui` no longer dead-ends when a crash takes the panel bridge offline. If the confirmation card cannot be shown and ComfyUI is not healthy (the reporter's empty-body 502, or ECONNREFUSED), the orchestrator falls back to the headless restart of the configured local process — it does not depend on the dead bridge. An explicit decline, a still-healthy server, a readable busy queue, and an unrelaunchable process still refuse.

Panel #654 / #404 are untouched; this is orchestrator recovery after the bridge is already gone.

Issue #425, opened 2026-07-30
panel_restart_comfyui cannot bridge legacy Manager to newly staged v4 install


artokun:
### What happened / root cause
After staging ComfyUI-Manager v4.2.2 on disk over a running legacy Manager 3.x instance, `panel_restart_comfyui(force:false)` cannot restart because neither the v4 nor legacy reboot endpoint is available in the currently loaded Manager API. The new Manager cannot take effect without restart, creating an upgrade deadlock.

### Steps to reproduce
1. Run ComfyUI 0.29.0 with legacy Manager 3.x loaded.
2. Update the Manager files on disk to official tag 4.2.2.
3. Call `panel_restart_comfyui({force:false})`.
4. The tool tries both endpoints but cannot restart.

### Exact error (scrubbed)
`Could not reach any ComfyUI-Manager reboot endpoint — ComfyUI was NOT restarted. Tried: POST /v2/manager/reboot → HTTP 405; GET /manager/reboot → HTTP 404`

### Fix
Applied locally: no. Upstream-only: yes.
Suggested change: provide a local process-control fallback for local ComfyUI targets after confirming the generation queue is idle, or preserve a compatible legacy reboot endpoint during the v3→v4 migration.

### Environment
Windows 11; ComfyUI 0.29.0; NVIDIA GeForce RTX 5060 Ti 16 GB; CUDA 13.0; torch 2.9.1; Python 3.12; Manager files staged at 4.2.2 while running process still has 3.x; comfyui-mcp 0.48.7; sidebar panel connected.

artokun:
The panel-side ComfyUI-Manager 3.x **legacy dialect** was completed in **panel 0.11.9** (comfyui-mcp-panel #250) — legacy reboot route, legacy self-update on 405, and a legacy list fallback. If this report was against a legacy Manager, please update the panel to 0.11.9 and confirm whether it's resolved. Note: these panel_* tool behaviors are served by the panel, so the fix lands via the panel package. If it persists on 0.11.9, reopen with the panel/mcp versions + the exact error and we'll investigate the orchestrator side.

artokun:
[...]

PR #1997, opened 2026-08-21
fix(1996): a boot slower than the readiness budget no longer forfeits the post-restart reconnect wait

### LINKED ISSUE #1996: panel_restart_comfyui gives up at a 120s default on an install that takes longer to boot — and that also skips the post-restart tab-reconnect wait
Split out of the reconnect half of **artokun/comfyui-mcp-panel#1562**, so it is not left implied. The `panel_refresh_nodes` half of that report is fixed in comfyui-mcp-panel#1566; this half is untouched by it and the two are independent.

## What the reporter saw

`panel_restart_comfyui` on a very large install (200+ packs; `GET /object_info` is 25,104,088 bytes / 20.84 s):

```
rebooting:true  ready:false  confirmed_cycle:false
recovered_ms:127254  probes:603  saw_down:true
"Reboot was dispatched and ComfyUI went down, but it has not become healthy
 within 127s — it may still be starting or the restart failed."
```

ComfyUI **did** come back (health check afterwards: 0.33.0, frontend 1.49.6, queue idle, and `GET /object_info/RIFEInterpolation` returned the freshly installed node). Every `panel_*` call after that failed with `no connected tab … Connected: none`, permanently, until the browser tab was refreshed by hand.

## ESTABLISHED — what the health check was actually polling

`observeRecovery` (`src/orchestrator/panel-tools.ts`) polls `probeComfyEndpoint(healthBase, …)`, i.e. **`GET <healthBase>/system_stats`** with `redirect:"manual"` against the **loopback boot base captured before dispatch** (`captureRebootHealthBase` probes `127.0.0.1`/`[::1]` directly, never through a proxy). It classifies:

* `healthy` — a same-origin 2xx whose body `looksLikeSystemStats`
* `down` — a connection failure whose code is in `PORT_NOT_LISTENING_CODES`, which is **`ECONNREFUSED` only**
* `unknown` — everything else, *including our own request timeout*

and it certifies **only** on a `healthy` sample taken after at least one `down`.

**It did not observe a healthy server and call it unhealthy. It stopped looking.** The readiness budget is `computeRebootTimingFromEnv()`'s `COMFYUI_PANEL_REBOOT_BUDGET_S`, **default 120 s**, measured from ACK COMPLETION — which is the reported `recovered_ms: 127254` (120 s of budget plus the ack window).

The reporter's own numbers confirm the samples: **603 probes over 127.25 s is ~211 ms per iteration**, against `intervalMs` 200 ms and `probeTimeoutMs` 2000 ms. So each probe returned in roughly **11 ms** — a fast connection refusal, not a timeout (a hung probe would have produced ~63 iterations, not 603), and `saw_down:true` confirms at least one `ECONNREFUSED` classification. The port was **genuinely not listening for the entire window**: ComfyUI was still importing 200+ packs, which on this install is the same tree that serializes to 25 MB of `/object_info`.

So the defect is **a budget sized against a workload nobody measured**, one level up from the `panel_refresh_nodes` half of the same report — and the same file already declares a larger value safe:

```
MAX_REBOOT_BUDGET_MS = 240_000;  // settle+budget <= 250s < 300s outer tools/call
```

The default is half of what the code says it may spend.

## ESTABLISHED — the two halves are coupled, and this is the part that hurts

The tab-reconnect wait is **downstream of the readiness verdict**:

```ts
if (!recovery.ready) {
  return ok({ rebooting: true, ready: false, confirmed_cycle: false, … });   // <- returns here
}
// #400/#709 — wait (bounded) for the tab to re-register and rebind this session
const tabBack = ctx.awaitPostRestartReachable ? await ctx.awaitPostRestartReachable(…) : …
```

Because the 120 s budget expired, `awaitPostRestartReachable` **was never entered at all**. A boot slower than the budget therefore loses the reconnect wait as well as the readiness answer — one number denies two things.

`Connected: none` is emitted only when `UiBridge.tabs()` is empty, so the tab was **absent from the routing registry**, not merely mis-routed. `conns` is written only in the `hello` branch (`src/services/ui-bridge.ts`) and deleted on socket close, so recovery requires a fresh `hello`.

## NOT ESTABLISHED — why the socket never re-registered

I could not reproduce this half: it needs a real 200-pack install and a real browser tab, and I had neither. What I can say from source, offered as leads rather than as a diagnosis:

* The panel's #654 repair for exactly this shape (`shouldReadvertiseAfterComfyRestart` → `client.rehello()`, comfyui-mcp-panel `web/js/lib/session-rebind.js`) shipped in panel **v0.15.10** and the reporter is on **0.15.24**, so it was present and did not recover the tab. It is **edge-triggered** on ComfyUI's `reconnected` event and gated on `bridgeConnected === true`; when the bridge is *not* connected the handler falls through to `connectAgent()`, which its own comment says early-returns on an already-OPEN socket.
* Neither side has a **time-based backstop**: nothing re-hellos on a timer, and the orchestrator has no path that notices "a restart was dispatched, the server is back, and no tab has re-registered".

Both are guesses about the trigger. Please do not treat either as the mechanism without reproducing it — the report says the browser tab stayed loaded and only F5 recovered it, and I never watched that happen.

## What a fix would have to be careful about

* Raising `COMFYUI_PANEL_REBOOT_BUDGET_S` costs a **genuinely dead** server's caller more waiting (up to the 240 s ceiling) before it is told so. That is the whole cost, and it is bounded — but it should be stated, not slipped in.
* The refusal itself is **not** the bug. "It has not become healthy within Ns … do NOT assume it is back" is the honest thing to say about a window that ended with the port closed. What is wrong is the window, and the fact that ending it early also skips the reconnect wait.

## Environment (from the original report)

Windows 11 / ComfyUI 0.33.0 / frontend 1.49.6 / RTX 5060 Ti 16 GB / comfyui-mcp 0.52.45 / panel 0.15.24 / Manager 3.x LEGACY / torch 2.13.0+cu130 / Python 3.12.11 / CUDA 13.0

artokun:
Fixes #1996.

Review is **pending** — requested from grok in a comment below. Nothing here has been reviewed by anyone but its author.

## What I took, and what I left

The issue separates what was ESTABLISHED (the budget gives up early; the reconnect wait sits downstream of the readiness verdict) from what was NOT (why the panel socket never re-registered). I took the first and left the second explicitly open — see the last section. I did not turn a lead into a cause.

## 1. The coupling — the half that actually costs the user something

`awaitPostRestartReachable` sat **after** `if (!recovery.ready) return …`, so a boot slower than the budget did not merely lose the readiness answer it honestly could not give. It also forfeited the reconnect wait, which on this path is the **only** thing that calls `ensureReachable()` to heal the session binding onto a returning tab. One number denied two independent questions:

* did **this** ComfyUI instance complete a down→up cycle, and
* is the panel tab routable again.

The wait now runs on both outcomes. The same shape existed on the headless managed-restart path (`const tabBack = recovery.ready ? await … : false`) and is fixed there too.

**Reproduced by execution before anything was written.** `panel-restart-slow-boot-reconnect.test.ts` drives the real handler with a health probe that refuses for the whole window — the reporter's shape scaled from 127 s to 120 ms — and counts the calls into a spied `awaitPostRestartReachable`. On `origin/main` that count is **0**, and `panel_tab_reconnected` is **absent from the reply entirely**.

## 2. The budget: 120 s → 180 s, and why that number

The reporter's own numbers establish that the observer **stopped looking** rather than misread a healthy server: 603 probes over 127.25 s is ~211 ms per iteration against `intervalMs` 200 ms, i.e. ~11 ms per probe — fast connection refusals, not timeouts. A hung probe would have produced ~63 iterations, not 603. The port was closed for the entire window while 200+ packs imported, and the install was healthy when checked afterwards.

So 120 s is **provably too small for that install**. What 180 s is *not* is a guess at its true boot time — nobody measured it, and I could not. It is the largest value the handler's own envelope allows while preserving everything that shares it:

```
180  readiness
 35  tab-reconnect wait (COMFYUI_PANEL_RECONNECT_WAIT_S default)
 30  allowance: confirmation card + reboot ack + post-restart argv read
---
245  ≤ OVERALL_MAX_MS (255 s)
```

**The 240 s the file already called safe is not available**, and that is worth stating because it is the obvious alternative. `MAX_REBOOT_BUDGET_MS = 240_000` carried the comment `settle+budget ≤ 250s < 300s outer` — arithmetic done against the outer `tools/call` limit only. It never counted the reconnect wait that runs *after* readiness inside the same 255 s handler budget. 240 s of readiness leaves 15 s, and the reconnect wait was handed `overallDeadline - Date.now()` directly. The value the file declared safe was **exactly the one that silently zeroed the reconnect wait**.

`MAX_REBOOT_BUDGET_MS` stays 240 s as the env escape hatch for an install slower still — it is now safe to use, because of:

## 3. The reserve

`readinessDeadlineWithReconnectReserve` clamps the readiness deadline to `overallDeadline - reconnectWaitTiming().budgetMs`. Without it, raising the budget would have traded one silent forfeit for another: a slow confirmation card followed by a full readiness window leaves the reconnect wait 0 ms and it returns `false` without waiting for anything. With it, a slow card costs readiness time (recoverable — the caller retries) rather than deleting the reconnect wait (not recoverable in that call). The reserve reads `reconnectWaitTiming()`, not a hard-coded 35 000, so raising `COMFYUI_PANEL_RECONNECT_WAIT_S` reserves more.

No regression in realistic cases: with a 90 s confirmation card (the `RESTART_CONFIRM_TIMEOUT_MS` ceiling) readiness still gets 130 s, more than today's 120 s.

## 4. What this deliberately does NOT do: turn a refusal into a success

Reporting "not confirmed" about a boot nobody observed is correct, and the refusal text is unchanged. On every unconfirmed branch `ready`, `server_ready`, `graph_tools_ready` and `confirmed_cycle` all stay `false`.

The specific trap here is that a re-registered tab looks like good news. It is not evidence about the boot: the panel's bridge socket runs browser→orchestrator and **never traverses ComfyUI**, so it says the route is alive and nothing else. The reply says exactly that:

> Separately: the panel tab DID re-register and this session was rebound onto it, so `panel_*` calls should route again. That is a fact about the browser socket only — it does not say ComfyUI finished booting.

On the headless path the guard is a literal `recovery.ready &&` in `graphToolsReady`, added because the reordering would otherwise let a reconnected tab report `ready:true` under an unproven boot. It has its own mutation below.

## 5. `classifyTabReconnect`'s `serverReady` input is renamed `tabWatched`

Its documented justification was *"the tab is only watched once the server is back"* — a **proxy**, true when written and false the moment the wait moved ahead of the readiness return. Left alone, an unconfirmed boot whose tab was watched and genuinely did not return would report `"unknown"` — an absence of observation — when a real observation had been made. The parameter now asks the question directly. `tabBack:true` still outranks everything; a missing baseline is still `"unknown"` and never a bare `false`; `ready`/`graph_tools_ready` still key off the boolean, so #654's gate is untouched.

## Mutation evidence

Committed first, then each mutation applied to the **call site**, run, reverted. The mutator refuses on any occurrence count ≠ 1, because the working copy is CRLF and a pattern that silently matches nothing reads as a survived mutation.

| # | Mutation (call site unless noted) | Result |
|---|---|---|
| M1 | restore `if (!recovery.ready) return …` above the reconnect wait (bound path) | **5 fail**, incl. `REPRODUCTION: the reconnect wait is ENTERED even though readiness expired` |
| M2 | budget default `180` → `120` | **1 fail**: `defaults above the window the reporter measured as still-down` |
| M3 | `gate.deadline = readinessDeadlineWithReconnectReserve(…)` → `Math.min(Date.now() + timing.budgetMs, overallDeadline)` | **1 fail**: `the bound path assigns gate.deadline via the reserve, with the LIVE reconnect budget` |
| M4 | restore `const tabBack = recovery.ready ? … : false` (headless path) | **2 fail**, incl. behavioural `watches the tab even when OUR observation never certified the cycle` |
| M5 | drop `recovery.ready &&` from headless `graphToolsReady` | **3 fail**, incl. two pre-existing #425 tests |
| M6 | `tabWatched: true` → `false` on the unconfirmed reply | **1 fail**: the REPRODUCTION test |

M3 is the one that argues for the source-level wiring assertion: the reserve is a **one-line assignment**, its helper has its own unit tests, and those unit tests survive the call site being reverted in full. Its behavioural half cannot be driven without a ~250 s handler.

## Tests

Full suite in a worktree cut from `origin/main` @ `26bfd3b`:

* **baseline** (clean `origin/main`): `5 failed | 11231 passed` — the 4 known pre-existing (`package-hooks` #1597, three in `node-id-union-message` #1889) plus one **load flake**, `panel-restart-configured-command > the issue's wedge …`, which passes on its own (`1 passed (1)`).
* **this branch**: `4 failed | 11243 passed | 4 skipped` — exactly the 4 known. The flake did not recur.
* All 13 restart-related files together: `195 passed`.
* `tsc --noEmit` clean; `check:vocabulary` OK.

* **The re-registration question is left OPEN, by design.** Why the panel socket never re-registered is not established here and I did not promote a lead to a cause. The #654 re-advertise shipped in panel v0.15.10 against a reporter on 0.15.24; it is edge-triggered on ComfyUI's `reconnected` and gated on `bridgeConnected === true`, with the other branch falling to `connectAgent()`, which early-returns on an already-open socket; neither side has a time-based backstop. Both tests in `bridge-reregisters-after-restart.spec.ts` pass on clean main, so the mechanism works and what is unproven is that the reporter's rig **reached** it. Establishing that needs a real large install and a real browser, and this rig has neither — there is no ComfyUI listening on 8188/8000 here.
  * What this PR does contribute to it: the reconnect wait now runs on the unconfirmed path, so the next report of this shape carries `panel_tab_reconnected: true | false | "unknown"` — an actual observation of the tab across that window instead of silence. That is a measurement the previous code could not produce.
* **No browser gate.** Nothing panel-rendered changed (the diff is `src/orchestrator/panel-tools.ts` plus tests), and there is no ComfyUI on this machine to run the Playwright suite against. Saying so rather than implying coverage I do not have.
* **The remote branch keeps its unreserved deadline.** It returns without ever waiting for a tab, so it has nothing to reserve for; the wiring test pins that exactly one unreserved assignment survives.
* **I did not touch the sibling budget in `process-control.ts`.** `getRemoteRebootTiming()` defaults `COMFYUI_REMOTE_REBOOT_BUDGET_S` to 120 s and, per its own docstring, serves **both** the remote target and a locally-installed ComfyUI **Desktop** instance — so the same 200+ pack import could in principle outrun it by the `restart_comfyui` route. I am recording that rather than fixing it, because I have not reproduced it: all I have is a grep showing the same number, which is exactly the "plausible premise nobody measured" this issue asked me not to act on. It also sits in a different envelope with different callers, and it has no reconnect wait downstream, so it is the budget half only. Worth its own ticket with a measurement attached.
* **The reconnect wait is NOT gated on `saw_down`.** Gating it would spare the no-op case (tab closed mid-command, server stays healthy) up to 35 s — but that case already burns the *entire* readiness budget probing a healthy endpoint, so the addition is ~19% on a call that was already slow, and gating would reintroduce a smaller version of the coupling this PR is about.

## The cost, stated plainly

A server that is genuinely never coming back is now reported as unconfirmed later: ~180 s instead of ~120 s, plus up to the 35 s reconnect wait that is no longer skipped. Bounded, inside the handler's existing 255 s envelope, and it buys the case the report is actually about — a large install that boots slowly and gets declared dead.

---

# Review round 1 — codex NO-SHIP, P1 fixed in `db71042`

> [P1] `panel-tools.ts:16995` — the legacy/headless restart path does not reserve the reconnect budget.

**Upheld in full, and it is this fix's own blind spot.** I introduced `readinessDeadlineWithReconnectReserve` for the bound path and did not carry it to `runHeadlessManagedRestart`, which computed its own `Math.min(Date.now() + legacyProofWindow, overallDeadline)` and reproduced the identical starvation — on a path this PR *had already modified* (M4/M5 touch its `recovery.ready ? … : false` shape). The reserve simply did not follow. That is the per-call-site adoption gap that left #1971's version probe unreached for releases.

Fixed with the **same helper at the second call site**, not a second mechanism.

## Reproduced by execution before changing anything

`panel-restart-legacy-fallback.test.ts` now drives the real headless path with the clock advanced through a slow confirmation. Measured, not derived:

| | budget handed to the tab wait | tab that re-registers at 28 s |
|---|---|---|
| before `db71042` | **5 s** (25 s once the harness clock was fine enough to resolve it) | `panel_tab_reconnected: false` — **false negative** |
| after | **35 s** (full budget, less one harness clock step) | `panel_tab_reconnected: true` |

The harness fakes `Date` only — timers stay real — and advances the clock from the confirm and from each health probe, so a 140 s window is consumed in ~1.4 s of wall clock. Its discrete step is named `PROBE_STEP_MS` and **subtracted from the assertion** rather than absorbed by loosening it; in production that overshoot is one 200 ms probe interval. My first attempt used a 20 s step, could not resolve 35 s from 25 s, and I tightened the *harness* rather than the assertion.

**Reachability is not hypothetical.** Of the three callers of `runHeadlessManagedRestart`, the #1671 crash-recovery one is entered *because* the confirmation card went unanswered — which costs the full 90 s `RESTART_CONFIRM_TIMEOUT_MS` plus the ~6 s decline probe before this path starts. ~96 s elapsed is its ordinary case, not its corner. Both 90 s and 96 s are pinned as tests.

## The envelope for THIS path — recomputed, not inherited

The 240s-is-safe comment was wrong because it counted against the wrong ceiling. Not repeating that: these are the headless path's own numbers, and they are **not** the bound path's.

```
OVERALL_MAX_MS        255s   (shared)
admission gate       ≥140s remaining  = LEGACY_SYNC_WORST_CASE_MS 40s
                                       + LEGACY_COLD_START_OBS_MS 100s
proof window          140s
reconnect wait         35s
```

Both at full size need **175 s**, i.e. arrival by `handlerStart + 80 s`. The #1671 route arrives at ~96 s and **cannot** meet that. So one of the three has to give:

* **Raising admission to 175 s — rejected.** It would refuse the crash recovery outright at ~96 s, telling a user whose ComfyUI just died to retry instead of restarting it. Strictly worse than the defect being fixed.
* **Leaving the proof window unreserved — the defect.**
* **Clamping the proof window — taken.** It costs cold-start observation: 140 s down to **105 s at the admission floor** (the 40 s unpreemptible sync plus 65 s rather than 100 s of watching).

That trade is the same asymmetry the bound path turns on and states: a short observation returns an honest `did NOT become healthy within Ns` the caller can retry; a starved reconnect wait silently drops the rebind and cannot be retried into. I am flagging the shortened cold-start window as the thing to argue with if you disagree — it is a real cost, deliberately taken, not an oversight.

## Mutation, round 1

| # | Mutation | Result |
|---|---|---|
| M7 | **new call site:** revert `proofDeadline` to `Math.min(Date.now() + legacyProofWindow, overallDeadline)` | **3 fail** — both behavioural reproductions *and* the wiring assertion |

The wiring assertion is now a **count of adopting call sites** (`readinessDeadlineWithReconnectReserve` × 2, `reserveMs: reconnectWaitTiming().budgetMs` × 2) rather than "the bound one is correct" — the shape that would have caught this finding in the first place.

## Suite, round 1

This machine picked up a live ComfyUI and ~101 node processes between rounds, so the full run is noisy. Triaged per-file rather than off the summary:

* Two consecutive full runs failed **9** and **10** tests — and the sets **shifted** (one file went 3 → 1; `vocabulary` and `ui-bridge` appeared only in the second). A shifting set is load, not a defect.
* Every newly-flagged file passes in isolation: `vocabulary`, `ui-bridge`, `gen-changelog-dedupe`, `tool-surface-filter`, `panel-pin-guard` → **419/420, then 109/109 alone**. `panel-pin-guard` also passes alone on a clean `origin/main` checkout, so it is not this branch.
* Stable across both runs: exactly the **4 known** (`package-hooks` #1597, three `node-id-union-message` #1889).
* The four restart files are green in every run: **76 passed**.
* `tsc --noEmit` clean.

CI on a clean runner is the arbiter here, not this box.

### DIFF
--- a/src/__tests__/orchestrator/panel-restart-legacy-fallback.test.ts
+++ b/src/__tests__/orchestrator/panel-restart-legacy-fallback.test.ts
@@ -443,3 +443,117 @@ describe("#1996 an unconfirmed managed restart still watches for the tab", () =>
     expect(out.ready).toBe(false);
   });
 });
+
+// ---------------------------------------------------------------------------
+// #1996 review round 1 (codex P1) — the reserve did not follow to THIS path.
+//
+// The bound path's readiness deadline is clamped by
+// `readinessDeadlineWithReconnectReserve` so the observation cannot eat the tab
+// reconnect wait. The headless/legacy path computed its own deadline as
+//
+//   const proofDeadline = Math.min(Date.now() + legacyProofWindow, overallDeadline);
+//
+// with no reserve — so the identical starvation survives here. The arithmetic,
+// from the file's own constants: OVERALL_MAX_MS 255s, legacyProofWindow 140s
+// (LEGACY_SYNC_WORST_CASE_MS 40s + LEGACY_COLD_START_OBS_MS 100s), reconnect
+// wait 35s. A confirmation answered at ~90s leaves 165s, the admission gate
+// (>=140s) lets the restart run, the proof window claims 140 of those 165, and
+// the tab wait is handed the 25s residual — LESS than the 35s a tab is allowed.
+// A tab that re-registers at 30s is then reported `panel_tab_reconnected:false`.
+//
+// REACHABILITY is not hypothetical. Of the three callers of
+// runHeadlessManagedRestart, the #1671 crash-recovery one is entered precisely
+// BECAUSE the confirmation card went unanswered — which costs the full
+// RESTART_CONFIRM_TIMEOUT_MS (90s) plus the ~6s decline probe before this path
+// even starts. The 90s in this test is the cheap case, not the worst one.
+//
+// HOW THIS RUNS FAST. Only `Date` is faked; timers stay real. The confirmation
+// advances the clock 90s, and each health probe advances it 20s, so the 140s
+// proof window is consumed in ~7 probes and ~1.4s of wall clock. Nothing here
+// re-implements the production arithmetic — the handler computes its own
+// deadlines from its own constants, and the test only observes the budget the
+// reconnect wait is actually handed.
+describe("#1996 r1 (codex P1): the headless path must reserve the reconnect wait too", () => {
+  // The harness's virtual clock advances in discrete PROBE_STEP_MS jumps, so the
+  // observation always overshoots its deadline by up to one step before noticing.
+  // In production that overshoot is one real probe interval (200ms) and is
+  // invisible; here it is the resolution limit of the measurement, so it is named
+  // and subtracted rather than absorbed into a looser number. 5s keeps it well
+  // clear of the 10s that separates the fixed residual from the broken one.
+  const PROBE_STEP_MS = 5_000;
+
+  /** Drive the headless path with `elapsedBeforeMs` already burned on the card,
+   *  and report the budget the tab-reconnect wait was actually handed. */
+  async function runWithSlowConfirm(elapsedBeforeMs: number, tabReturnsAfterMs: number) {
+    const prevInterval = process.env.COMFYUI_PANEL_REBOOT_INTERVAL_S;
+    vi.useFakeTimers({ toFake: ["Date"] });
+    __panelToolsTestHooks.setPanelRebootTiming(null); // REAL 140s legacy proof window
+    // Shrink only the REAL sleep between probes (floored at 50ms by observeRecovery);
+    // the virtual clock still advances PROBE_STEP_MS per probe, so the full 140s
+    // window is consumed in ~28 probes and ~1.4s of wall clock.
+    process.env.COMFYUI_PANEL_REBOOT_INTERVAL_S = "0.01";
+    try {
+      // The endpoint never comes back, so the proof window runs to its deadline —
+      // the case where the residual handed to the tab wait is smallest.
+      __panelToolsTestHooks.setHealthProbe(async () => {
+        vi.setSystemTime(Date.now() + PROBE_STEP_MS);
+        return "down";
+      });
+      const { ctx } = makeCtx(NO_ENDPOINT_REPLY);
+      ctx.confirm = (async () => {
+        vi.setSystemTime(Date.now() + elapsedBeforeMs);
+        return "yes" as const;
+      }) as PanelToolCtx["confirm"];
+      ctx.panelConnectionIdentity = () => ({ generation: 1, tabSessionId: "browser-tab-a" });
+      let handedBudgetMs = -1;
+      // A perfectly ordinary tab: it re-registers `tabReturnsAfterMs` after being
+      // asked for, and comes back if — and only if — it is given that long.
+      ctx.awaitPostRestartReachable = async (_before, budgetMs?: number) => {
+        handedBudgetMs = budgetMs ?? 0;
+        return handedBudgetMs >= tabReturnsAfterMs;
+      };
+      ctx.tabCanMutateGraph = () => true;
+
+      const res = (await restartTool().handler({}, ctx)) as ToolResult;
+      const out = JSON.parse(res.content.find((c) => c.type === "text")!.text as string);
+      return { handedBudgetMs, out };
+    } finally {
+      vi.useRealTimers();
+      if (prevInterval === undefined) delete process.env.COMFYUI_PANEL_REBOOT_INTERVAL_S;
+      else process.env.COMFYUI_PANEL_REBOOT_INTERVAL_S = prevInterval;
+    }
+  }
+
+  it("a confirmation answered at ~90s must NOT starve the tab wait below its own budget", async () => {
+    // 255s overall − 90s on the card = 165s left. The admission gate (≥140s) lets the
+    // restart run, and pre-fix the 140s proof window claimed all but ~25 of those.
+    const allowed = __panelToolsTestHooks.getReconnectWaitTiming().budgetMs; // 35s
+    const { handedBudgetMs, out } = await runWithSlowConfirm(90_000, 28_000);
+
+    // THE DEFECT, as the observation rather than the arithmetic: the wait is handed
+    // less than a tab is allowed to take. Pre-fix ~20–25s here; the reserve makes it
+    // the full budget, less one step of the harness's own clock resolution.
+    expect(handedBudgetMs).toBeGreaterThanOrEqual(allowed - PROBE_STEP_MS);
+    // …and the consequence that makes it a P1 rather than a rounding error: a tab
+    // that came back normally, at 28s, is no longer reported as never having come
+    // back. 28s is outside the pre-fix residual and inside the post-fix one, so this
+    // separates the two behaviours rather than restating the number above.
+    expect(out.panel_tab_reconnected).toBe(true);
+    // The refusal is untouched — the endpoint genuinely never answered.
+    expect(out.server_ready).toBe(false);
+    expect(out.ready).toBe(false);
+  });
+
+  it("the #1671 caller's own timings — an UNANSWERED card — keep the tab wait whole too", async () => {
+    // Reachability is not hypothetical. runHeadlessManagedRestart's #1671 caller is
+    // entered BECAUSE the confirmation card went unanswered, which costs the full
+    // RESTART_CONFIRM_TIMEOUT_MS (90s) plus the ~6s decline probe. Modelling that
+    // 96s here (through the same no-endpoint route, which shares the code under
+    // test) leaves 159s — still admitted, and pre-fix the residual was ~19s.
+    const allowed = __panelToolsTestHooks.getReconnectWaitTiming().budgetMs;
+    const { handedBudgetMs, out } = await runWithSlowConfirm(96_000, 28_000);
+
+    expect(handedBudgetMs).toBeGreaterThanOrEqual(allowed - PROBE_STEP_MS);
+    expect(out.panel_tab_reconnected).toBe(true);
+  });
+});
--- a/src/__tests__/orchestrator/panel-restart-legacy-fallback.test.ts
+++ b/src/__tests__/orchestrator/panel-restart-legacy-fallback.test.ts
@@ -392,3 +392,54 @@ describe("#654 an unwatched reconnect is reported as unknown, and is still not r
     expect(String(out.note)).not.toMatch(/could NOT be determined/);
   });
 });
+
+// #1996 — the headless path carried the SAME coupling the bound path did:
+//
+//   const tabBack = recovery.ready ? await ctx.awaitPostRestartReachable(…) : false;
+//
+// so a kill+relaunch whose cold start outran the observation window never looked at
+// the tab, and then reported a bare `false` about it. The bound path's version of this
+// is covered behaviourally in panel-restart-slow-boot-reconnect.test.ts; this file owns
+// the scaffolding that reaches the HEADLESS reply, so its half lives here.
+describe("#1996 an unconfirmed managed restart still watches for the tab", () => {
+  it("watches the tab even when OUR observation never certified the cycle", async () => {
+    // The endpoint never goes down, so the managed restart is honestly
+    // couldn't-confirm — and the tab is watched anyway.
+    __panelToolsTestHooks.setHealthProbe(async () => "healthy");
+    const { ctx } = makeCtx(NO_ENDPOINT_REPLY);
+    let watched = 0;
+    ctx.panelConnectionIdentity = () => ({ generation: 1, tabSessionId: "browser-tab-a" });
+    ctx.awaitPostRestartReachable = async () => {
+      watched++;
+      return true;
+    };
+    ctx.tabCanMutateGraph = () => true;
+
+    const res = (await restartTool().handler({}, ctx)) as ToolResult;
+    const out = JSON.parse(res.content.find((c) => c.type === "text")!.text as string);
+
+    expect(watched).toBe(1); // 0 before the fix — the wait was never entered
+    expect(out.panel_tab_reconnected).toBe(true);
+    // AND THE REFUSAL IS INTACT. A reconnected browser socket is not a booted
+    // ComfyUI: `ready` must stay false or this fix has manufactured a success.
+    expect(out.server_ready).toBe(false);
+    expect(out.confirmed_cycle).toBe(false);
+    expect(out.ready).toBe(false);
+    expect(out.graph_tools_ready).toBe(false);
+    expect(String(out.note)).toMatch(/did NOT become healthy within/);
+  });
+
+  it("a watched tab that did not come back is a real false here too", async () => {
+    __panelToolsTestHooks.setHealthProbe(async () => "healthy");
+    const { ctx } = makeCtx(NO_ENDPOINT_REPLY);
+    ctx.panelConnectionIdentity = () => ({ generation: 1, tabSessionId: "browser-tab-a" });
+    ctx.awaitPostRestartReachable = async () => false;
+    ctx.tabCanMutateGraph = () => true;
+
+    const res = (await restartTool().handler({}, ctx)) as ToolResult;
+    const out = JSON.parse(res.content.find((c) => c.type === "text")!.text as string);
+
+    expect(out.panel_tab_reconnected).toBe(false);
+    expect(out.ready).toBe(false);
+  });
+});
--- a/src/__tests__/orchestrator/panel-restart-slow-boot-reconnect.test.ts
+++ b/src/__tests__/orchestrator/panel-restart-slow-boot-reconnect.test.ts
@@ -280,13 +280,31 @@ describe("#1996 WIRING: the bound readiness deadline goes through the reserve", 
     // Reading the reserve from reconnectWaitTiming() is what keeps it correct when a
     // user raises COMFYUI_PANEL_RECONNECT_WAIT_S; a hard-coded 35_000 would silently
     // under-reserve for them.
-    expect((s.match(/reserveMs: reconnectWaitTiming\(\)\.budgetMs/g) ?? []).length).toBe(1);
     // Exactly ONE unreserved `gate.deadline` assignment survives: the REMOTE branch,
     // which returns without ever waiting for a tab, so it has nothing to reserve for.
     expect(
       (s.match(/gate\.deadline = Math\.min\(Date\.now\(\) \+ timing\.budgetMs, overallDeadline\);/g) ?? [])
         .length,
     ).toBe(1);
+  });
+
+  it("BOTH restart paths clamp their observation with the SAME reserve (r1 codex P1)", () => {
+    // The review finding was not that the reserve was wrong — it was that it did not
+    // FOLLOW to the legacy/headless path, whose `proofDeadline` computed its own
+    // unreserved `Math.min(Date.now() + legacyProofWindow, overallDeadline)` and
+    // reproduced the identical starvation. This is the per-call-site adoption gap that
+    // left #1971's version probe unreached for releases, so it is pinned as a COUNT of
+    // adopting call sites rather than as "the bound one is correct".
+    const s = src();
+    expect((s.match(/readinessDeadlineWithReconnectReserve\(\{/g) ?? []).length).toBe(2);
+    // Both read the LIVE reconnect budget rather than a hard-coded 35_000, so raising
+    // COMFYUI_PANEL_RECONNECT_WAIT_S reserves more on BOTH paths, not one.
+    expect((s.match(/reserveMs: reconnectWaitTiming\(\)\.budgetMs/g) ?? []).length).toBe(2);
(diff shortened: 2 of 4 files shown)


artokun:
@grok please review this one. Naming where the risk is rather than handing you a diff — these are the four places I would attack first, roughly in order of how much I would lose if I am wrong.

**1. Does the reordering manufacture a success anywhere?** The whole fix moves `awaitPostRestartReachable` **above** the `!recovery.ready` return, on both the bound path and the headless managed-restart path. The refusal is supposed to be untouched — `ready` / `server_ready` / `graph_tools_ready` / `confirmed_cycle` all `false` on every unconfirmed branch. On the headless path the only thing holding that line is a literal `recovery.ready &&` I added to `graphToolsReady`; without it a tab that re-registers under an unproven boot reports `ready:true`. I mutated exactly that (M5, 3 tests died) but it is one `&&` in a reply with five booleans, and it is the failure mode I would least like to have missed. Please check the bound path's unconfirmed reply the same way — I believe every field there is a hard-coded `false`, but I wrote it, so I am the wrong person to read it.

**2. Is 180 s defensible, or did I pick a round number and reason backwards?** The derivation I claim is `180 readiness + 35 reconnect + 30 (card + ack + argv) = 245 ≤ OVERALL_MAX_MS (255)`. Two ways that is wrong: the 30 s allowance may be too small for a real confirmation card (`RESTART_CONFIRM_TIMEOUT_MS` is 90 s, and a card answered at 90 s makes the reserve clamp readiness to 130 s — I claim that is still ≥ today's 120 s so nothing regresses, but please check the arithmetic rather than my sentence about it); or the whole envelope is wrong because I missed a wait inside the 255 s handler that I have not counted. I looked for one and did not find it — the preflight and `readServingArgv(2000)` are the two I know about.

**3. The reserve, `readinessDeadlineWithReconnectReserve`.** This is the piece with the weakest behavioural coverage and I want to say so plainly. Its clamp only bites when the confirmation card has already eaten most of the handler budget, which needs a ~250 s test to drive, so what pins the call site is a **source-level regex assertion**, not behaviour. M3 (reverting the call site to the old `Math.min`) kills that assertion and nothing else — the helper's own unit tests all survive. If you think a source assertion is not good enough for a one-line assignment, say so; I could not find a way to drive it in a test that finishes.

**4. The `serverReady` → `tabWatched` rename on `classifyTabReconnect`.** This changes an exported signature and edits assertions in an existing test file (`restart-tab-reconnect-unknown.test.ts`, panel#654), which is the shape that should make a reviewer suspicious. My argument: the old parameter's documented justification was *"the tab is only watched once the server is back"*, which the reordering makes false, so leaving it would report `"unknown"` for a tab that genuinely was watched and genuinely did not return. I changed two existing count assertions (`toBe(2)` → `toBe(3)`) because there is now a third reply emitting a classified reconnect. Please check I did not weaken #654's gate while I was in there — `ready` must still key off the raw boolean and `"unknown"` must never buy readiness.

**What I did NOT establish, and am not claiming:** why the panel socket never re-registered in the reporter's run. That stays open. This PR does not touch the panel, and I could not measure it — there is no ComfyUI on this rig, so no Playwright run either.

One state I am least sure of, for completeness: the unconfirmed path now waits for the tab even when `saw_down` is `false` (the no-op drop — tab closed mid-command, server never went down). I chose not to gate on `saw_down` because that call already burns the entire readiness budget probing a healthy endpoint, so the extra wait is proportionally small and gating would reintroduce a smaller version of the coupling. If you read that trade the other way, it is a one-line change.

artokun:
Pushed `1cd472b` on top — an adversarial re-read of this PR's own diff, not a review finding. Two sentences in the note I added were over-claiming:

- `"this session was rebound onto it"` is false in the common case. `awaitPostRestartReachable` returns true on a newer hello from the **same** browser-tab session, which usually needed no rebinding at all — `ensureReachable()` only runs when the old tab id stopped resolving. It now says what was actually observed.
- the negative branch pointed at a browser refresh first. That is precisely the manual step panel#654 is about, so it now offers `panel_set_workflow_target({mode:"current"})` first and the refresh only if that also fails.

No behaviour change; the four restart-related files are still 73/73 and `tsc --noEmit` is clean. The four review targets in my comment above are unchanged and still the places worth attacking.

artokun:
Working the arithmetic in review target #2 all the way out, since "no regression" was a sentence in my PR body rather than a number, and the reviewer should not have to trust it.

Let `T` = time on the confirmation card, `A` = the reboot ack. `dispatchTimeout = min(15000, …)`, so `A ≤ 15s`; `RESTART_CONFIRM_TIMEOUT_MS = 90s`, and a card that actually *reaches* 90s times out into `decision === "unreachable"`, which never gets here — so `T < 90s` on this path.

Readiness gets `min(180, 220 − T − A)` seconds (220 = 255 overall − 35 reserve).

| T | A | readiness | vs today's flat 120s |
|---|---|---|---|
| ~0s (agent-driven / instant yes) | ~0 | **180s** | +60s |
| 20s | 2s | **180s** | +60s |
| 40s | 2s | **178s** | +58s |
| 89s | 0s | **131s** | +11s |
| 89s | 15s | **116s** | **−4s** |

So there *is* a corner where the reserve makes readiness marginally shorter than today: a card answered at ~89s **and** a full 15s ack timeout, costing ~4s of readiness. I am leaving it rather than flooring readiness at 120s, and the reason is the asymmetry this whole PR turns on — **readiness is retryable and the reconnect wait is not.** A short readiness window returns a clean "not confirmed" the caller can act on; a zeroed reconnect wait silently drops the rebind and produces the reporter's "Connected: none until F5". Flooring readiness would re-squeeze the reconnect wait by up to 20 of its 35s to buy back 4s on a call where the human already spent 89 seconds.

If you read that trade the other way it is a one-line change to the clamp, and I would rather be told than guess.

artokun:
**HELD — green, mergeable, and waiting on a reviewer who has not shown up.**

State as of now:

- CI green on head `1cd472b`: Build macos / ubuntu / windows all pass, Pack & install smoke passes.
- `origin/main` is still `26bfd3b` — the same main those checks ran against, so the green is not stale.
- `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.
- `/pulls/1997/reviews` holds exactly one entry: `copilot-pull-request-reviewer[bot]` saying it hit its quota limit. That is not a review. **grok has posted nothing on this PR, and nothing on any of the six other open PRs today either** — the newest review body across all of them is that same quota notice.

I am not merging this on my own verification. My evidence is on the record above — a reproduction that fails on clean main, six mutations including four at the **call site**, a full suite that matches baseline exactly — and none of it is a substitute for someone who did not write the code reading it. The specific reason this one should not self-merge is #1 in my review request: the whole change moves a wait above an early return, and the only thing keeping an unconfirmed boot from reporting `ready:true` on the headless path is a single `recovery.ready &&` that I added myself. An author reads their own guard as obviously right.

**What unblocks it:** a grok review, or the owner saying to land it. It stays open and green until then.

artokun:
Direct Codex merge gate: NO-SHIP (P1). src/orchestrator/panel-tools.ts:16995 leaves the legacy/headless restart path without a reconnect-budget reserve. With confirmation around 90s, its 140s proof window leaves only 25s for the default 35s tab reconnect and can report panel_tab_reconnected:false even when the tab returns normally. Apply the same reserve to this path, add regression coverage, rerun CI, and request a fresh direct gate.

artokun:
**Round 1 gate: SHIP.**

`codex-gate.mjs --worktree …/wt-1996-cc --base origin/main --name 1997` → `SHIP` (verdict file `codex-gate-1997-26280-…`). Worktree HEAD was confirmed equal to the PR head before the gate ran, and the gate left the worktree clean afterwards.

The P1 is fixed in the r1 commit and the reasoning is in the PR body: reproduced by execution first (tab wait handed **5s**, then 25s once the harness clock could resolve it, against a 35s budget — a tab returning at 28s reported `panel_tab_reconnected:false`), fixed with the **same helper at the second call site**, and the headless path's envelope recomputed from its own constants rather than inherited from the bound path's.

The one thing in that fix worth arguing with, stated so it is not buried: the clamp shortens the legacy cold-start observation from 140s to as little as **105s** at the admission floor. I took that over raising the admission gate to 175s, because raising it would refuse the #1671 crash recovery outright at ~96s elapsed — telling a user whose ComfyUI just died to retry instead of restarting it.

**Merge state.** Since the first green, `origin/main` moved `26bfd3b` → `0804d9e` (#1995, touching `panel-sync.ts` / `install-panel.ts` — no overlap with `panel-tools.ts`). A green check is only green for the main it ran against, so rather than merge on the stale one I rebased onto `0804d9e`, re-ran typecheck and the restart suite plus #1995's own test file (108 passed), and CI is green again on `a1ec0b8`. Merging that.

PR #2002, opened 2026-08-21
fix(panel#1572): panel_show_media accepts audio, through the card the panel already had

### LINKED ISSUE #1572: panel_show_media: no audio support (.wav rejected) - agent cannot play generated TTS/voice takes for the user
`panel_show_media` refuses audio files: `Error: unsupported file type ".wav" (allowed: .png, .jpg, .jpeg, .gif, .webp, .mp4, .webm, .mov, .mkv, .m4v, .avi)`. In a TTS/voice-clone session (Higgs Audio v3 lines for a MiniMax H3 audio-reuse workflow) the user asked to be shown the generated speech takes ("zeig ihn mir mal"). Audio is a first-class ComfyUI output (SaveAudio, VHS_LoadAudio, TTS Audio Suite, H3 audio-video model), but the panel media card only whitelists image/video extensions.

### Steps to reproduce
1. Generate any audio (e.g. SaveAudio output or a .wav on disk).
2. Call `panel_show_media` with `{items:[{source:{path:"C:/.../take.wav"}}]}`.
3. Tool errors with the unsupported-file-type message above; there is no way to play audio for the user in the panel chat.

### Exact error (scrubbed)
`Error: unsupported file type ".wav" (allowed: .png, .jpg, .jpeg, .gif, .webp, .mp4, .webm, .mov, .mkv, .m4v, .avi): D:/Comfy/input/reuse_halo_c1.wav`

### Fix
Workaround used: wrapped each wav in an mp4 with a showwaves waveform via ffmpeg (`-filter_complex "[0:a]showwaves=..."`) and showed that — works, but is an awkward detour every agent will re-invent.
Precise change needed: extend the extension whitelist in panel_show_media's path validation with common audio types (.wav, .mp3, .flac, .ogg, .m4a) and render them as an <audio controls> card (the frontend already ships audio playback UI for VHS_LoadAudio's audiopreview widget, so a minimal audio card can reuse that). The 20-frame storyboard sampling should be skipped for audio items.

### Environment
OS: Windows 11 Pro 10.0.26200 / ComfyUI 0.33.0 (local-git) / ComfyUI FRONTEND 1.49.6 / GPU: RTX 5080 16GB / comfyui-mcp 0.52.45 / panel 0.15.32 / Backend: Claude

artokun:
Closes artokun/comfyui-mcp-panel#1572 (cross-repo — will not auto-close; I will close it by hand after merge).

## The report

`panel_show_media` refused every audio file:

```
unsupported file type ".wav" (allowed: .png, .jpg, .jpeg, .gif, .webp, .mp4, .webm, .mov, .mkv, .m4v, .avi)
```

In a TTS/voice-clone session the user asked to hear the generated speech takes and the agent could not show them. It worked around it by wrapping each `.wav` in an mp4 with an ffmpeg `showwaves` filter — which works, and which every agent will re-invent.

Filed against the panel repo; the gate is here, in `src/orchestrator/panel-tools.ts`.

## Scope: a bug fix, not a feature

A stabilization freeze is in force, so this was a real scope decision, made by execution.

**The panel already has the audio card.** panel#710 shipped all of it and it is on `main` today:

- `paintAudio(url, name)` builds a real `<audio controls>` card, excluded from the lightbox, deliberately not autoplaying;
- `AUDIO_EXTENSIONS` (`web/js/lib/media-preview.js`) is its allowlist of what `<audio>` decodes;
- `classifyShowMediaItem` already returns `"audio"` and trusts the orchestrator's own `kind` first;
- `composeShowMediaReply`'s `painterFor` already maps `audio` to `paintAudio`, counts it under `paintedByKind.audio`, emits a per-item "you cannot hear this" disclosure, and skips the video storyboard for it;
- `chat-media.js`'s `MEDIA_KINDS` already persists a card under `mkind:"audio"`.

No new media surface, renderer, transport, transcoding or waveform UI. One closed door: an extension allowlist in the orchestrator.

## Proof that audio renders with no other change

Real ffmpeg-encoded files in all eight formats, put through the **fixed orchestrator**, and the exact items it dispatched fed to the panel's **unmodified** `composeShowMediaReply` from `origin/main` (`56fd8eb0`) in headless Chromium, with `paintAudio` building a real `<audio>` element:

```
panelCommit: 56fd8eb082f92d31a777fe0a4571fa335c7bcb96   (panel origin/main, untouched)
reply:       { ok: true, count: 8, painted: 8, unconfirmed: 0, unrenderable: [], previews: [] }
elements:    8 <audio>, 0 <img>
decoded:     wav 0.400s  mp3 0.400s  flac 0.400s  ogg 0.400s
             oga 0.400s  opus 0.407s m4a 0.400s   aac 0.441s   — all 8 loadedmetadata
```

`buildVideoStoryboard` and `uploadBlobToInput` were injected as functions that throw. Never called: `previews: []`, no error. The storyboard genuinely does not run for audio.

## The extension list, and where it comes from

`.wav .mp3 .flac .ogg .oga .opus .m4a .aac` — the panel renderer's own `AUDIO_EXTENSIONS`, verbatim. The gate's job is to refuse what the card cannot present, so the card's allowlist is the authoritative answer. Cross-checked, not invented:

| source | set |
|---|---|
| panel `AUDIO_EXTENSIONS` (`web/js/lib/media-preview.js`) | wav mp3 flac ogg oga opus m4a aac |
| this repo's `upload_image action:"audio"` / `services/image-management.ts` `AUDIO_MIME` | wav mp3 flac ogg m4a aac |
| ComfyUI core `SaveAudio` / `SaveAudioMP3` / `SaveAudioOpus` / `SaveAudioAdvanced` | flac mp3 opus |
| the reporter's file, and what VHS / the TTS packs write | wav |

ComfyUI's side was read, not recalled: `comfy_api/latest/_ui.py` has `AudioSaveHelper._FORMATS = {"flac", "mp3", "opus"}` and names the file `..._00001.<format>`, so `.opus` is a **core ComfyUI output** — an allowlist without it would have shipped still refusing `SaveAudioOpus`.

**Deliberately narrower than this repo's own `AUDIO_MIME_BY_EXT`** (`src/orchestrator/audio-attachment.ts`), which also lists `webm/weba/wave/m4b/mpga/mp2`. `.webm` is **already in `VIDEO_EXTS`** here — claiming it as audio would demote an existing video card — and the rest are not in the panel's renderer set, so admitting them would widen the allowlist without widening what can be shown. There is a test pinning `.webm` as video.

## The MIME map is measured, not copied

The same failure class #811 fixed for video, and it bites harder here: a `data:` URL's declared type is the only type the browser gets, and a wrong one is refused before any decoder runs. Chromium, real files:

| ext | shipped | also tried | result |
|---|---|---|---|
| `.opus` | `audio/ogg` PASS 0.407s | `audio/opus` | FAIL `MEDIA_ELEMENT_ERROR: Unable to load URL due to content type` |
| `.flac` | `audio/flac` PASS 0.400s | `audio/x-flac` | FAIL same |
| `.wav` | `audio/wav` PASS | `audio/wave` | FAIL same |
| `.aac` | `audio/aac` PASS | `audio/aacp` | FAIL same |

`audio/opus` is exactly what this repo's `AUDIO_MIME_BY_EXT` uses for `.opus`, so copying that table would have shipped a dead card for the one format ComfyUI's own `SaveAudioOpus` writes. Both wrong values have their own named regression test.

## The three notes that had to change

Admitting a third kind is not only the `if`. The kind union became one shared type (`ShownMediaKind`) so the compiler names every consumer — it named four, and without that these would have shipped:

1. **`oversizedInlineRefusal`** had two branches and treated "not a video" as "an image", so an oversized `.wav` would have been described to the agent as displayed "at full size" with `get_image` offered as the way to *look* at it.
2. **`forwardedByReferenceNote`** would have given an audio-only batch no caveat at all.
3. The tool description now says audio is supported and that the agent cannot hear it.

These are consequences of the allowlist change, not adjacent work: without them this PR would introduce a lie it did not previously tell.

## The gate is still an allowlist

Widened by eight named extensions, not loosened. No sniffing, no denylist, no fallthrough. Tests pin the refusal for `.txt .exe .pdf .json` and — deliberately — for `.wma .aiff .mid`, which are unmistakably *audio* (this repo lists all three in `AUDIO_EXT_NOT_ENCODABLE`) and which would sail through if the gate had been loosened to "looks like audio". A refused item is refused before its bytes are read and before anything is dispatched.

## Mutation evidence

Seven mutations, each deleting part of the fix. All **applied** (verified — an anchor that does not match reads as a survivor) and all **KILLED**:

| # | mutation | verdict | killed |
|---|---|---|---|
| M1 | delete the `AUDIO_EXTS` branch — the **call site** that sets `kind` | KILLED | 21 |
| M2 | `.opus` to `audio/opus` | KILLED | 2, both named for it |
| M3 | `.flac` to `audio/x-flac` | KILLED | 2, both named for it |
| M4 | drop `.opus` from the allowlist | KILLED | 4 |
| M5 | accept anything instead of the allowlist | KILLED | 10, incl. #811's own refusal test |
| M6 | revert `oversizedInlineRefusal` to two branches | KILLED | 3 |
| M7 | revert `forwardedByReferenceNote`'s audio caveat | KILLED | 3 |

M1 is the call-site mutation specifically: a helper-level test would survive it, and it does not.

## Scope reduction, and a known limitation this ships with

This PR previously carried four more fixes plus a headless guard, all found by successive gate rounds while reviewing the audio change. **Seven rounds, seven real P1s, six of them introduced while fixing the previous one** — including the same defect class (judging a tab ADDRESS instead of the routed DESTINATION) reappearing at a second call site, in a guard written *after* the first instance was understood and fixed. That recurrence is the measurement that settled the scope question: the surrounding surface is too large to stabilise in one PR, and each round spent closing one site was a round that introduced the next.

Everything except the allowlist is filed with its evidence, and working code is preserved on `spike/1572-client-routing-audio` and `spike/1572-headless-audio-guard`:

| finding | issue |
|---|---|
| headless/mobile client cannot play audio | #2010 |
| `/view` probe asks "is it media" instead of "does it match the filename" | #2011 |
| address-vs-destination class, **both confirmed sites** | #2012 |
| `show_media` is the only mailboxable command | #2013 |
| a pre-#710 panel paints `kind:"audio"` as a broken image | #2017 |

### The limitation this ships with, stated plainly

Measured on both sides, same call, same fixture:

| | headless client + `{path:"voice.wav"}` |
|---|---|
| `origin/main` | `isError: true` — `unsupported file type ".wav"`, nothing dispatched |
| this PR | `isError: false` — dispatched as `kind:"audio"` with a `data:audio/wav` URL |

The mobile client has no audio branch (`MediaCard.build` is `if (item.isVideo) _video else _still`) and `BridgeClient` replies `{shown:true}` to any `show_media` without reading the items. So on that path **an honest refusal becomes a false success**. The user hears nothing either way; what changes is that the agent is told the take played.

It is in the changelog as a known limitation and tracked in **#2010**, which this raises from a capability gap to a live incorrect-success. I am flagging it rather than shipping it quietly. A guard for it was built, reviewed three times and pulled — every version drew a fresh P1, which is precisely the case for giving it its own PR rather than riding here.

## Browser coverage — stated honestly

**No panel code changed here, so the panel Playwright e2e suite is not a gate for this PR.** I measured it anyway (`--workers=1` → 16 failed / 62 passed of 78, 13.4 min) but it proves nothing about this change: the diff is entirely in this repo, the suite's `testIgnore` excludes the only tests that touch `show_media`, and on this rig `custom_nodes/comfyui-mcp-panel` is a symlink to the panel main checkout (`d5025cfd`), not even panel `origin/main`. The real browser evidence is the Chromium measurement above, plus `node --test browser_tests/unit/{media-preview,chat-media-persistence,media-collapse,lightbox-gallery}` — 140 passed on panel `origin/main`.

## Deliberately NOT done

- **The headless-client inline branch** is untouched — see #2010. Inlining is *not* the fix there, and the evidence is in that issue.
- **`.wma/.aiff/.mid` and friends** stay refused. They are audio, and `<audio>` cannot play them.
- **`.webm` stays video.** No attempt to disambiguate an audio-only WebM.
- **No transcoding, no waveform, no size-cap change.** `MAX_BYTES` is untouched; oversized audio takes the same reference/staging routes images and videos already take.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+### MCP
+
+#### Fixed
+- `panel_show_media` now accepts audio files through the panel's existing audio card (panel #1572).
+
 ## [0.52.68] - 2026-08-23
 
 _No user-facing changes._
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -383,7 +383,6 @@ _No user-facing changes._
   reply about one item can no longer stand in for a two-item call and lose the
   second in silence. A reply whose own numbers contradict each other is reported
   as that, rather than as a short count.
-
 - **Ollama refuses audio unless the model is a verified listener (#1972).** Native
   `/api/chat` carries audio in `message.images[]`. Most models HTTP 400 (fail
   closed). `huihui_ai/gemma-4-abliterated:E4b-qat` accepts the payload and
--- a/src/__tests__/orchestrator/panel-show-media-audio-ext.test.ts
+++ b/src/__tests__/orchestrator/panel-show-media-audio-ext.test.ts
@@ -1,0 +1,186 @@
+// #1572 — panel_show_media refused audio outright:
+//   unsupported file type ".wav" (allowed: .png, .jpg, … .avi)
+// so in a TTS/voice-clone session the agent could not play the generated take
+// back to the user at all, and worked around it by wrapping each .wav in an
+// mp4 with an ffmpeg showwaves filter.
+//
+// WHAT WAS ACTUALLY MISSING WAS ONLY THIS GATE. The panel has shipped a real
+// `<audio controls>` card since #710 — paintAudio, its own AUDIO_EXTENSIONS
+// allowlist, an "audible" disclosure in the reply, and media persistence under
+// mkind:"audio" — and composeShowMediaReply already routes kind:"audio" to that
+// painter and skips the video storyboard for it. Nothing was built here; a door
+// was opened.
+//
+// THE EXTENSION SET is the panel renderer's own AUDIO_EXTENSIONS verbatim
+// (web/js/lib/media-preview.js), because the gate's job is to refuse what the
+// card cannot present. It is a superset of what this codebase already calls
+// audio (upload_image action:"audio" and services/image-management's
+// AUDIO_MIME) and of what ComfyUI's own save nodes write (AudioSaveHelper's
+// _FORMATS is {flac, mp3, opus}).
+//
+// THE MIME MAP IS MEASURED, NOT DERIVED. A `data:` URL's declared type is the
+// only type the browser gets, and a wrong one is refused before any decoder
+// runs — the same class of bug #811 fixed for video. Verified in Chromium
+// against real ffmpeg-encoded files: `.opus` as `audio/opus` (the natural
+// guess, and what this repo's own audio-attachment.ts AUDIO_MIME_BY_EXT uses)
+// FAILS with "MEDIA_ELEMENT_ERROR: Unable to load URL due to content type",
+// while `audio/ogg` decodes; `.flac` as `audio/x-flac` fails where `audio/flac`
+// decodes. That matters because .opus is a core ComfyUI SaveAudio output.
+//
+// The resolved item (with its dataUrl/mime) is never echoed back in the tool's
+// OWN reply — it is dispatched onward via `ctx.call({cmd:"show_media", items})`
+// to the panel. So verification captures the OUTBOUND call, mirroring
+// panel-show-media-video-ext.test.ts's makeCtx/calls convention exactly.
+
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { buildPanelToolDefs, type PanelToolCtx, type ToolResult } from "../../orchestrator/panel-tools.js";
+
+type Forwarded = Record<string, unknown>;
+
+function makeCtx(): { ctx: PanelToolCtx; calls: Forwarded[] } {
+  const calls: Forwarded[] = [];
+  const ctx = {
+    call: async (cmd: Forwarded) => {
+      calls.push(cmd);
+      return { content: [{ type: "text", text: JSON.stringify(cmd) }] } as ToolResult;
+    },
+    confirm: async () => "yes" as const,
+    bridge: { isHeadless: () => false } as unknown as PanelToolCtx["bridge"],
+    tabId: "test-tab",
+  } as PanelToolCtx;
+  return { ctx, calls };
+}
+
+async function showMedia(items: Array<Record<string, unknown>>): Promise<{ res: ToolResult; calls: Forwarded[] }> {
+  const def = buildPanelToolDefs().find((d) => d.name === "panel_show_media");
+  if (!def) throw new Error("panel_show_media not found");
+  const { ctx, calls } = makeCtx();
+  const res = (await def.handler({ items }, ctx)) as ToolResult;
+  return { res, calls };
+}
+
+const textOf = (res: ToolResult): string => res.content.map((c) => ("text" in c ? c.text : "")).join("\n");
+
+/** Exactly the panel renderer's AUDIO_EXTENSIONS (web/js/lib/media-preview.js). */
+const AUDIO_CASES: Array<[string, string]> = [
+  [".wav", "audio/wav"],
+  [".mp3", "audio/mpeg"],
+  [".flac", "audio/flac"],
+  [".ogg", "audio/ogg"],
+  [".oga", "audio/ogg"],
+  [".opus", "audio/ogg"],
+  [".m4a", "audio/mp4"],
+  [".aac", "audio/aac"],
+];
+
+/** Types that must STILL be refused — the gate is an allowlist, not a denylist,
+ *  and widening it by eight extensions must not turn it into "anything goes".
+ *  `.wma` and `.aiff` are deliberately here: they are unmistakably AUDIO (this
+ *  repo's audio-attachment.ts lists both in AUDIO_EXT_NOT_ENCODABLE), so if the
+ *  gate had been loosened to "looks like audio" rather than to a named set,
+ *  they would sail through and reach an `<audio>` element that cannot play
+ *  them. `.webm` is NOT tested as a refusal — it is already a VIDEO extension
+ *  here and must stay one. */
+const REFUSED = [".txt", ".exe", ".wma", ".aiff", ".mid", ".pdf", ".json"];
+
+let root: string;
+const paths: Record<string, string> = {};
+
+beforeAll(() => {
+  root = mkdtempSync(join(tmpdir(), "cmcp-showmedia-audioext-"));
+  mkdirSync(root, { recursive: true });
+  for (const ext of [...AUDIO_CASES.map(([e]) => e), ...REFUSED]) {
+    const p = join(root, `take${ext}`);
+    writeFileSync(p, Buffer.alloc(64)); // well under the inline cap
+    paths[ext] = p;
+  }
+});
+
+afterAll(() => {
+  rmSync(root, { recursive: true, force: true });
+});
+
+describe("#1572 audio is accepted by panel_show_media", () => {
+  it.each(AUDIO_CASES.map(([ext]) => ext))("%s is no longer refused", async (ext) => {
+    const { res } = await showMedia([{ source: { path: paths[ext] }, caption: "take 1" }]);
+    expect(res.isError).toBeFalsy();
+    expect(textOf(res)).not.toMatch(/unsupported file type/);
+  });
+
+  it.each(AUDIO_CASES)("%s is dispatched as kind:\"audio\" with mime %s", async (ext, mime) => {
+    const { res, calls } = await showMedia([{ source: { path: paths[ext] } }]);
+    expect(res.isError).toBeFalsy();
+
+    const dispatched = calls.find((c) => c.cmd === "show_media");
+    expect(dispatched, "no show_media command was dispatched to the panel").toBeTruthy();
+    const items = dispatched!.items as Array<{ kind?: string; dataUrl?: string }>;
+    expect(items).toHaveLength(1);
+    // The panel's classifyShowMediaItem trusts the orchestrator's own `kind`
+    // FIRST, and painterFor.audio is paintAudio — so this string is what routes
+    // the file to an <audio> element instead of a broken <img> (#710).
+    expect(items[0].kind).toBe("audio");
+    expect(items[0].dataUrl).toBeTruthy();
+    expect(items[0].dataUrl!.startsWith(`data:${mime};base64,`)).toBe(true);
+  });
+
+  // The two entries a table-copy would have got wrong, pinned individually so a
+  // future "tidy this up against AUDIO_MIME_BY_EXT" cannot silently re-break the
+  // one audio format ComfyUI's own SaveAudioOpus writes.
+  it("does NOT label .opus as audio/opus — Chromium refuses that content type", async () => {
+    const { calls } = await showMedia([{ source: { path: paths[".opus"] } }]);
+    const items = (calls.find((c) => c.cmd === "show_media")!.items) as Array<{ dataUrl?: string }>;
+    expect(items[0].dataUrl!.startsWith("data:audio/opus;")).toBe(false);
+  });
+
+  it("does NOT label .flac as audio/x-flac — Chromium refuses that content type", async () => {
+    const { calls } = await showMedia([{ source: { path: paths[".flac"] } }]);
+    const items = (calls.find((c) => c.cmd === "show_media")!.items) as Array<{ dataUrl?: string }>;
+    expect(items[0].dataUrl!.startsWith("data:audio/x-flac;")).toBe(false);
+  });
+});
+
+describe("#1572 the gate is still an allowlist — both directions", () => {
+  it.each(REFUSED)("%s is still refused", async (ext) => {
+    const { res, calls } = await showMedia([{ source: { path: paths[ext] } }]);
+    expect(res.isError).toBe(true);
+    expect(textOf(res)).toMatch(
+      new RegExp(`unsupported file type "\\${ext}"`),
+    );
+    // Refused BEFORE anything is dispatched — a refusal that still painted
+    // would be no gate at all.
+    expect(calls.find((c) => c.cmd === "show_media")).toBeFalsy();
+  });
+
+  it("the refusal names the audio extensions too, so the caller can act on it", async () => {
+    const { res } = await showMedia([{ source: { path: paths[".txt"] } }]);
+    const text = textOf(res);
+    for (const [ext] of AUDIO_CASES) expect(text).toContain(ext);
+    // …without losing what it already allowed.
+    expect(text).toContain(".png");
+    expect(text).toContain(".mov");
+  });
+});
+
+describe("#1572 .webm stays a VIDEO", () => {
+  // audio-attachment.ts's AUDIO_MIME_BY_EXT maps webm/weba to audio/webm.
+  // Copying that table wholesale would have re-classified an extension this
+  // tool has treated as video since #811, turning an existing video card into
+  // an audio player. The audio set is deliberately the panel renderer's, not
+  // that one.
+  it("is dispatched as video/webm, not audio/webm", async () => {
+    const p = join(root, "clip.webm");
+    writeFileSync(p, Buffer.alloc(64));
+    const { res, calls } = await showMedia([{ source: { path: p } }]);
+    expect(res.isError).toBeFalsy();
+    const items = (calls.find((c) => c.cmd === "show_media")!.items) as Array<{
+      kind?: string;
+      dataUrl?: string;
+    }>;
+    expect(items[0].kind).toBe("video");
+    expect(items[0].dataUrl!.startsWith("data:video/webm;base64,")).toBe(true);
+  });
+});
--- a/src/__tests__/orchestrator/panel-show-media-oversized.test.ts
+++ b/src/__tests__/orchestrator/panel-show-media-oversized.test.ts
@@ -14,6 +14,7 @@ import { existsSync, mkdirSync, mkdtempSync, readdirSync, rmSync, statSync, writ
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type { PanelToolCtx, ToolResult } from "../../orchestrator/panel-tools.js";
+import type { PanelVersionReading } from "../../services/ui-bridge.js";
 
 const state = vi.hoisted(() => ({
   outputDir: "" as string,
@@ -86,7 +87,7 @@ const { buildPanelToolDefs } = await import("../../orchestrator/panel-tools.js")
 
 type Forwarded = Record<string, unknown>;
 
-function makeCtx(): { ctx: PanelToolCtx; calls: Forwarded[] } {
+function makeCtx(opts: { headless?: boolean; version?: PanelVersionReading } = {}): { ctx: PanelToolCtx; calls: Forwarded[] } {
   const calls: Forwarded[] = [];
   const ctx = {
     call: async (cmd: Forwarded) => {
@@ -94,7 +95,10 @@ function makeCtx(): { ctx: PanelToolCtx; calls: Forwarded[] } {
       return { content: [{ type: "text", text: JSON.stringify(cmd) }] } as ToolResult;
     },
     confirm: async () => "yes" as const,
-    bridge: { isHeadless: () => false } as unknown as PanelToolCtx["bridge"],
+    bridge: {
+      isHeadless: () => opts.headless ?? false,
+      advertisedPanelVersion: () => opts.version ?? {},
+    } as unknown as PanelToolCtx["bridge"],
     tabId: "test-tab",
   } as PanelToolCtx;
   return { ctx, calls };
@@ -102,10 +106,11 @@ function makeCtx(): { ctx: PanelToolCtx; calls: Forwarded[] } {
 
 async function showMedia(
   items: Array<Record<string, unknown>>,
+  opts: { headless?: boolean; version?: PanelVersionReading } = {},
 ): Promise<{ res: ToolResult; calls: Forwarded[] }> {
   const def = buildPanelToolDefs().find((d) => d.name === "panel_show_media");
   if (!def) throw new Error("panel_show_media not found");
-  const { ctx, calls } = makeCtx();
+  const { ctx, calls } = makeCtx(opts);
   const res = (await def.handler({ items }, ctx)) as ToolResult;
   return { res, calls };
 }
@@ -518,5 +523,33 @@ describe("panel_show_media: an oversized AUDIO file takes the same routes (#1572
     const text = textOf(res);
     expect(text).toContain(", audio)");
     expect(text).toMatch(/real filesystem WRITE/);
+  });
+
+  it("refuses a staged audio file before copying it for a proven-old panel", async () => {
+    const before = existsSync(join(outDir, "_panel_staged"))
+      ? readdirSync(join(outDir, "_panel_staged")).length
+      : 0;
+    const { res, calls } = await showMedia(
+      [{ source: { path: bigAudioOutside, stage: true } }],
+      { version: { raw: "0.11.41", version: "0.11.41" } },
+    );
+    expect(res.isError).toBe(true);
+    expect(calls).toHaveLength(0);
+    expect(textOf(res)).toMatch(/cannot paint "audio"/);
+    const after = existsSync(join(outDir, "_panel_staged"))
+      ? readdirSync(join(outDir, "_panel_staged")).length
+      : 0;
+    expect(after).toBe(before);
+  });
+
+  it("refuses local audio before dispatching it to a headless client", async () => {
+    const { res, calls } = await showMedia(
+      [{ source: { path: bigAudioUnderRoot } }],
+      { headless: true },
+    );
+    expect(res.isError).toBe(true);
+    expect(calls).toHaveLength(0);
+    expect(textOf(res)).toMatch(/headless client/i);
+    expect(textOf(res)).toMatch(/audio player/i);
   });
 });
--- a/src/__tests__/orchestrator/panel-show-media-oversized.test.ts
+++ b/src/__tests__/orchestrator/panel-show-media-oversized.test.ts
@@ -458,3 +458,65 @@ describe("#941: a forwarded /view reference says what 'painted' does not cover",
     expect(textOf(res)).toMatch(/renders BROKEN/);
   });
 });
+
+// #1572 — audio reaches these routes now. Nothing here is audio-SPECIFIC
+// machinery: resolveServableViewRef and stageFileIntoServedDir are path-based
+// and never look at an extension, so the reference and staging routes were
+// always going to work. What was NOT automatic is what the agent is TOLD: both
+// notes had two branches and treated "not a video" as "an image", so an
+// oversized .wav would have been described as displayed "at full size" with
+// get_image offered as the way to LOOK at it. These exercise the whole path
+// rather than the note functions, because a note tested alone cannot show that
+// the right kind reached it.
+describe("panel_show_media: an oversized AUDIO file takes the same routes (#1572)", () => {
+  let bigAudioUnderRoot: string;
+  let bigAudioOutside: string;
+
+  beforeEach(() => {
+    bigAudioUnderRoot = join(outDir, "takes", "voice_00001.wav");
+    writeFileSync(bigAudioUnderRoot, Buffer.alloc(OVERSIZE));
+    bigAudioOutside = join(root, "elsewhere", "voice_00001.wav");
+    writeFileSync(bigAudioOutside, Buffer.alloc(OVERSIZE));
+  });
+
+  it("under a ComfyUI root: forwarded by reference, labelled audio, nothing inlined", async () => {
+    const { res, calls } = await showMedia([{ source: { path: bigAudioUnderRoot } }]);
+    expect(res.isError).toBeFalsy();
+    const items = calls[0].items as Array<Record<string, unknown>>;
+    expect(items[0].kind).toBe("viewRef");
+    expect(items[0].viewRef).toMatchObject({ filename: "voice_00001.wav", subfolder: "takes" });
+    expect(items[0].dataUrl).toBeUndefined();
+    const text = textOf(res);
+    // The forwarded note names it as audio and gives it the audio caveat…
+    expect(text).toContain(", audio)");
+    expect(text).toMatch(/AUDIO is never sent to you/);
+    // …and does not hand the agent the image branch's advice.
+    expect(text).not.toContain("comes back inline");
+  });
+
+  it("outside every root: refused with the AUDIO explanation, not the image one", async () => {
+    const { res } = await showMedia([{ source: { path: bigAudioOutside } }]);
+    expect(res.isError).toBe(true);
+    const text = textOf(res);
+    expect(text).toMatch(/audio player/i);
+    expect(text).toMatch(/cannot hear it/i);
+    // The two false claims the old two-branch prose would have made about a .wav.
+    expect(text).not.toContain("displays it to the USER at full size");
+    expect(text).not.toContain("returns the image inline");
+    // And it must not borrow the video branch either.
+    expect(text).not.toContain("SAMPLED");
+    // Still the same actionable remedy every other kind gets.
+    expect(text).toContain("stage:true");
+  });
+
+  it("stage:true copies it into the served dir and labels the copy audio", async () => {
+    const { res, calls } = await showMedia([{ source: { path: bigAudioOutside, stage: true } }]);
+    expect(res.isError).toBeFalsy();
+    const items = calls[0].items as Array<Record<string, unknown>>;
+    expect(items[0].kind).toBe("viewRef");
+    expect((items[0].viewRef as { subfolder?: string }).subfolder).toBe("_panel_staged");
+    const text = textOf(res);
+    expect(text).toContain(", audio)");
+    expect(text).toMatch(/real filesystem WRITE/);
+  });
+});
--- a/src/__tests__/services/comfy-view-ref.test.ts
+++ b/src/__tests__/services/comfy-view-ref.test.ts
@@ -674,3 +674,94 @@ describe("stagedForDisplayNote (#802)", () => {
     expect(note).not.toMatch(/was displayed to the user/i);
   });
 });
+
+// #1572 — panel_show_media gained a third kind. These notes had TWO branches
+// and treated "not a video" as "an image", so an oversized .wav would have been
+// described to the agent as displayed "at full size", with get_image offered as
+// the way to look at it — a claim about a file nobody can look at. The kind
+// union is one shared type (ShownMediaKind) precisely so the compiler names
+// every note when a kind is added; these pin the prose it named.
+describe("#1572 the notes tell an AUDIO caller the truth about audio", () => {
+  const audio = {
+    path: "/refs/take.wav",
+    sizeBytes: 72 * 1024 * 1024,
+    capBytes: 20 * 1024 * 1024,
+    kind: "audio" as const,
+  };
+
+  it("oversizedInlineRefusal does not describe audio as displayed, or offer get_image for it", () => {
+    const msg = oversizedInlineRefusal({
+      ...audio,
+      resolution: { status: "outside", checked: [{ kind: "output", dir: "/c/output" }] },
+    });
+    // The player is what the user actually gets.
+    expect(msg).toMatch(/audio player/i);
+    // The two false claims the image branch would have made.
+    expect(msg).not.toContain("displays it to the USER at full size");
+    expect(msg).not.toContain("returns the image inline");
+    // And it must not borrow the video branch's contact sheet, which audio has no
+    // equivalent of.
+    expect(msg).not.toContain("SAMPLED");
+    // Still actionable — the remedy survives.
+    expect(msg).toMatch(/Copy or move it/);
+  });
+
+  it("oversizedInlineRefusal still says the agent cannot hear it", () => {
+    const msg = oversizedInlineRefusal({
+      ...audio,
+      resolution: { status: "outside", checked: [{ kind: "output", dir: "/c/output" }] },
+    });
+    expect(msg).toMatch(/cannot hear it/i);
+  });
+
+  it("forwardedByReferenceNote gives audio its own caveat instead of leaving it bare", () => {
+    const note = forwardedByReferenceNote(
+      [
+        {
+          path: "/c/output/take.wav",
+          sizeBytes: 3e7,
+          kind: "audio",
+          ref: { filename: "take.wav", type: "output" },
+        },
+      ],
+      20 * 1024 * 1024,
+    );
+    expect(note).toMatch(/AUDIO is never sent to you/);
+    expect(note).toMatch(/no sampled preview/i);
+    expect(note).toMatch(/Do not describe how it sounds/i);
+    // The image branch's advice must not appear for an audio-only batch.
+    expect(note).not.toContain("comes back inline");
+  });
+
+  it("forwardedByReferenceNote keeps every kind's caveat in a mixed batch", () => {
+    const note = forwardedByReferenceNote(
+      [
+        { path: "/c/output/a.png", sizeBytes: 3e7, kind: "image", ref: { filename: "a.png", type: "output" } },
+        { path: "/c/output/a.mp4", sizeBytes: 3e7, kind: "video", ref: { filename: "a.mp4", type: "output" } },
+        { path: "/c/output/a.wav", sizeBytes: 3e7, kind: "audio", ref: { filename: "a.wav", type: "output" } },
+      ],
+      20 * 1024 * 1024,
+    );
+    expect(note).toContain("comes back inline");
+    expect(note).toContain("never sent to you inline");
+    expect(note).toMatch(/AUDIO is never sent to you/);
+  });
+
+  it("stagedForDisplayNote labels a staged audio copy as audio", () => {
+    const note = stagedForDisplayNote(
+      [
+        {
+          path: "/refs/take.wav",
+          stagedPath: "/c/output/_panel_staged/1724000000000-take.wav",
+          sizeBytes: 72 * 1024 * 1024,
+          kind: "audio",
+          ref: { filename: "1724000000000-take.wav", subfolder: "_panel_staged", type: "output" },
+        },
+      ],
+      20 * 1024 * 1024,
+    );
+    expect(note).toContain("audio");
+    expect(note).not.toContain(", video)");
+    expect(note).not.toContain(", image)");
+  });
+});
--- a/src/orchestrator/panel-tools.ts
+++ b/src/orchestrator/panel-tools.ts
@@ -19369,7 +19369,7 @@ CHECKED FOR YOU: the graph read this message prescribes was just run, and it ` +
     ),
     def(
       "panel_show_media",
-      "Display one or more images or videos directly in the panel chat. Use this whenever the user asks to SEE or SHOW a file — a disk path you composited/downloaded/generated (absolute path on the orchestrator host) OR a ComfyUI output ref ({ filename, subfolder?, type? }). Items are rendered as media cards in the agent chat area; supply optional captions. Max 8 items per call. A path item OVER the 20 MB inline cap that is not under any directory ComfyUI serves can still be shown by passing stage:true on that item — the orchestrator COPIES it into <output>/_panel_staged (an opt-in, persistent disk write; 512 MB per-file and 2 GB total caps) and displays the copy by reference. NEVER describe an image with emoji or text placeholders — call this tool instead.",
+      "Display one or more images, videos or AUDIO files directly in the panel chat. Use this whenever the user asks to SEE, SHOW, PLAY or HEAR a file — a disk path you composited/downloaded/generated (absolute path on the orchestrator host) OR a ComfyUI output ref ({ filename, subfolder?, type? }). Items are rendered as media cards in the agent chat area — audio gets a real player (.wav/.mp3/.flac/.ogg/.oga/.opus/.m4a/.aac), so a generated TTS or voice take is played back with this tool, not described; supply optional captions. You are never sent the audio yourself and cannot hear it. Max 8 items per call. A path item OVER the 20 MB inline cap that is not under any directory ComfyUI serves can still be shown by passing stage:true on that item — the orchestrator COPIES it into <output>/_panel_staged (an opt-in, persistent disk write; 512 MB per-file and 2 GB total caps) and displays the copy by reference. NEVER describe an image with emoji or text placeholders — call this tool instead.",
       {
         items: z
           .array(
@@ -19471,13 +19471,64 @@ CHECKED FOR YOU: the graph read this message prescribes was just run, and it ` +
           ".m4v": "video/x-m4v",
           ".avi": "video/x-msvideo",
         };
-        // #2011 — family classification for the /view probe only. The path
-        // gate below still refuses audio; a /view ref is already forwarded
-        // unclassified, and without this set `take.wav` serving `image/png`
-        // is silently blessed. Keys match upload_image action:"audio" /
-        // image-management AUDIO_MIME — the orchestrator's own audio set,
-        // not the panel renderer's (which also lists .opus/.oga).
-        const AUDIO_EXTS = new Set([".wav", ".mp3", ".flac", ".ogg", ".m4a", ".aac"]);
+        // #1572 — audio was refused outright ("unsupported file type \".wav\""),
+        // so a TTS/voice-clone session could not play a generated take back to
+        // the user at all. Nothing else was missing: the panel has shipped a
+        // real `<audio controls>` card since #710 (paintAudio, its own
+        // AUDIO_EXTENSIONS allowlist, an "audible" disclosure in the reply, and
+        // media persistence under mkind:"audio"), and composeShowMediaReply
+        // already routes kind:"audio" to it and skips the video storyboard for
+        // it. THIS gate was the only closed door.
+        //
+        // The set is the panel renderer's own AUDIO_EXTENSIONS, verbatim —
+        // the gate's job is to refuse what the card cannot present, so the
+        // card's allowlist is the authoritative answer. It is a superset of
+        // what this codebase already calls audio elsewhere (upload_image
+        // action:"audio" and services/image-management's AUDIO_MIME:
+        // .wav/.mp3/.flac/.ogg/.m4a/.aac) and of what ComfyUI's own save nodes
+        // actually write (comfy_api AudioSaveHelper._FORMATS is
+        // {flac, mp3, opus} and it names the file `…_00001.<format>`, so
+        // SaveAudio/SaveAudioMP3/SaveAudioOpus/SaveAudioAdvanced emit exactly
+        // .flac/.mp3/.opus; .wav is what VHS and the TTS packs write).
+        //
+        // Deliberately NARROWER than the orchestrator's own AUDIO_MIME_BY_EXT
+        // (audio-attachment.ts), which additionally lists webm/weba/wave/m4b/
+        // mpga/mp2: `.webm` is ALREADY in VIDEO_EXTS above and claiming it as
+        // audio here would demote an existing video card, and the rest are not
+        // in the panel's renderer set — admitting them would pass this gate only
+        // to reach a link card, i.e. widen the allowlist without widening what
+        // can be shown.
+        const AUDIO_EXTS = new Set([
+          ".wav",
+          ".mp3",
+          ".flac",
+          ".ogg",
+          ".oga",
+          ".opus",
+          ".m4a",
+          ".aac",
+        ]);
+        // MEASURED in Chromium against real ffmpeg-encoded files served as
+        // `data:` URLs, not derived from a table — the same failure #811 fixed
+        // for video bites harder here, because a data URL's declared type is
+        // the ONLY type the browser gets and a wrong one is refused before any
+        // decoder runs ("MEDIA_ELEMENT_ERROR: Unable to load URL due to content
+        // type"). Two entries would have been wrong if copied from elsewhere in
+        // this codebase:
+        //   - `.opus` → `audio/opus` (what audio-attachment.ts uses, and the
+        //     natural guess) is REFUSED by Chromium; `audio/ogg` decodes. That
+        //     matters because .opus is a core ComfyUI SaveAudio output.
+        //   - `.flac` → `audio/x-flac` is REFUSED; `audio/flac` decodes.
+        const AUDIO_MIME: Record<string, string> = {
+          ".wav": "audio/wav",
+          ".mp3": "audio/mpeg",
+          ".flac": "audio/flac",
+          ".ogg": "audio/ogg",
+          ".oga": "audio/ogg",
+          ".opus": "audio/ogg",
+          ".m4a": "audio/mp4",
+          ".aac": "audio/aac",
+        };
         const MAX_BYTES = 20 * 1024 * 1024; // 20 MB
 
         const resolved: Array<Record<string, unknown>> = [];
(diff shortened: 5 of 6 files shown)


artokun:
@grok — review requested. Here is where I think the risk actually is, in the order I would look:

**1. The allowlist is a security-adjacent gate and I widened it.** `src/orchestrator/panel-tools.ts`,
the `IMAGE_EXTS`/`VIDEO_EXTS`/`AUDIO_EXTS` classifier. The load-bearing claim is that this stays an
*explicit allowlist* — eight named extensions added, no sniffing, no denylist, no fallthrough, and a
refused item still refused before its bytes are read and before anything is dispatched. Please check I
did not accidentally create a path where an unlisted type reaches `readFileSync` or `ctx.call`. The
mutation that most worries me is M5 (make the `else` accept everything): it is killed by 10 tests
including #811's own, but that is my test, and I wrote both sides of it.

**2. The MIME map is a measurement, and measurements can be wrong.** I claim `.opus` must be
`audio/ogg` and NOT `audio/opus`, and `.flac` must be `audio/flac` and NOT `audio/x-flac`, because
Chromium refuses the latter of each with `MEDIA_ELEMENT_ERROR: Unable to load URL due to content type`
before any decoder runs. This directly contradicts this repo's own `AUDIO_MIME_BY_EXT`
(`src/orchestrator/audio-attachment.ts`), which maps `opus` to `audio/opus`. I believe the divergence
is correct — that table feeds a model API, this one feeds a `<audio>` element, and they are answering
different questions — but if you think one table should serve both, say so, because I deliberately
chose not to unify them. **Note the second-order finding:** if `audio/opus` is wrong for a `data:` URL
in a browser, it may also be wrong for whatever consumes `AUDIO_MIME_BY_EXT`. I did not touch that
file and did not investigate it; it may be a separate bug.

**3. The state I am least sure of: whether the panel actually renders this, and whether my evidence for
that is worth anything.** The whole scope decision — bug-adjacent gap rather than a feature under the
freeze — rests on "the panel already has the audio card, so this is one closed door". I verified that
by running the panel's real, unmodified `composeShowMediaReply` from `origin/main` in headless Chromium
against the exact items the fixed orchestrator dispatches, and got 8 `<audio>` elements, 0 ``, all
eight decoding. But that is a harness I wrote, driving a module out of `git show`, not the panel as
mounted in ComfyUI. **No panel code changed in this PR and the panel Playwright suite does not cover
`show_media` at all**, so if you think this needs coverage I do not have, it is better to hear that
now than after merge.

**4. Widening the kind union touched three notes, and prose has no compiler.** `ShownMediaKind` in
`src/services/comfy-view-ref.ts`. `oversizedInlineRefusal` previously treated "not a video" as "an
image"; audio now has its own branch. `forwardedByReferenceNote` gained an audio caveat. Please check I
did not change what the image/video branches say — the tests I added assert the audio strings and that
the image branch's advice does not leak into an audio-only batch, but a subtle regression in the
existing prose would not be caught by them.

**5. The `/view` probe change (#941) is the one thing here that is not the reported bug.** I widened it
to count `audio/*` as media because it was actively reporting an audible file as a broken reference.
That is a behaviour change to a diagnostic in a PR about an allowlist. If you read it as out of scope
under the freeze, I will pull it out and ship it separately.

Full mechanism, sources for the extension list, the measured MIME table, and all nine mutation results
are in the PR body. Review is pending; nothing here has been independently reviewed and I have not
merged.

artokun:
## Measured runs, branch vs baseline, back to back on this machine

### This repo's suite — no new failures

| | files | tests |
|---|---|---|
| `origin/main` (26bfd3b, clean detached worktree) | 6 failed / 609 passed (615) | **9 failed** / 11227 passed / 4 skipped |
| this branch | 5 failed / 612 passed (617) | **8 failed** / 11271 passed / 4 skipped |

The failing sets overlap and none of them touch this change:

- `node-id-union-message` (3) and `package-hooks` (1) fail on **both**. `package-hooks` is
  environmental — `expected [ 'dist' ] to deeply equal []`, i.e. the worktree was never built.
- `tool-surface-filter` and `vocabulary` fail on **both** (baseline has one more
  `tool-surface-filter` case than the branch run did).
- `late-mutation-e2e` and `ui-bridge` failed on **baseline only** — both are the known wall-clock
  flakes.
- `gen-changelog-dedupe` (2) failed on the **branch run only**, both with
  `Test timed out in 30000ms` while driving temp git repos under concurrent load. Re-run in isolation
  on this branch: **passes**. I checked it specifically because this PR edits `CHANGELOG.md`.

The five files that failed on the branch were also re-run in isolation on **both** trees, same command,
back to back: identical 4 failures on each (`node-id-union-message` ×3, `package-hooks` ×1). So the
delta between the two full runs is flake and load, not this diff.

Also green on the branch: `tsc --noEmit`, `npm run check:vocabulary` (1758 files, no live references to
retired names — this PR edits the `panel_show_media` description), `npm run i18n:check` (162 keys,
12 locales, 100%).

### Panel Playwright — measured, and it is not a gate for this PR

```
npx playwright test --workers=1     →  16 failed, 62 passed (78), 13.4 min
```

I am reporting that number rather than inheriting one, but it is worth nothing as evidence about this
change and I would rather say so than let it read as coverage:

- **The diff is entirely in this repo.** No panel file changed, so branch and baseline are the same
  panel tree by construction.
- **The suite does not cover `show_media` at all** — the only panel tests that touch it are the
  `browser_tests/unit/*.test.mjs` node tests, which Playwright's `testIgnore` excludes.
- **The panel that was actually served is not a branch of anything.** On this rig
  `ComfyUI/custom_nodes/comfyui-mcp-panel` is a symlink to the panel main checkout at `d5025cfd`,
  which is not even panel `origin/main` (`56fd8eb0`).
- None of the 16 failures are media-related: they are `chat-history-v2` (3),
  `conversation-persistence` (5), `workflow-chat-identity` (4), `external-orchestrator`, `hidden-tab`,
  `pending-queue`, `streaming-render`.

### The panel-side coverage that IS real

`node --test browser_tests/unit/{media-preview,chat-media-persistence,media-collapse,lightbox-gallery}.test.mjs`
→ **140 passed, 0 failed** on panel `origin/main`. These are the modules that classify and paint an
audio card.

Plus the Chromium measurement in the PR body, which is the only thing here that actually joins the two
halves: the fixed orchestrator's real output, fed to the panel's unmodified `composeShowMediaReply`
from `origin/main`, painting real `<audio>` elements — 8 painted, 0 ``, all eight decoded, the
video storyboard never reached.

CI on this PR is green (Build on ubuntu/macos/windows, pack-and-install smoke).

**Still no review.** The only entry on `/pulls/2002/reviews` is Copilot saying it could not review
because of a quota limit, which is not a review. I have not merged and will not merge on my own
verification.

artokun:
Follow-up push (1f46869): three more tests, in `panel-show-media-oversized.test.ts`, exercising the **oversized** routes for audio end to end rather than only through the note functions — a file over the 20 MB cap under a ComfyUI root (forwarded by reference, labelled `audio`, nothing inlined), one outside every root (refused with the audio explanation, and asserting the image branch's "displays it at full size"/"returns the image inline" and the video branch's "SAMPLED" do NOT appear), and `stage:true` on the outside one (copied into `_panel_staged`, copy labelled audio, write disclosed).

Why they were worth adding: `resolveServableViewRef` and `stageFileIntoServedDir` are path-based and never read an extension, so those routes were always going to work for audio — the part that was not automatic is what the agent is TOLD, and a note tested in isolation cannot show that the right `kind` reached it. Mutation check on the call-site path: reverting both note branches kills 2 of these 3 (the third asserts the staging write, which is kind-independent). `tsc --noEmit` clean; the five `show_media`/view-ref files are 129 passed / 0 failed.

Review is still pending — the only entry on `/pulls/2002/reviews` remains the Copilot quota message. Not merging on my own verification.

artokun:
## HOLD — green, unreviewed, and deliberately not merged

Final head `1f46869`. All four CI checks pass (Build on ubuntu/macos/windows, Pack & install smoke). `tsc --noEmit`, `check:vocabulary` and `i18n:check` are clean, and the branch's full-suite failures are the same pre-existing/flaky ones `origin/main` has — measured back to back on this machine, both numbers in the comment above.

**No review has come back.** `gh api repos/artokun/comfyui-mcp/pulls/2002/reviews` holds exactly one entry: `copilot-pull-request-reviewer[bot]`, `COMMENTED`, saying it could not review because the requester reached their quota limit. That is not a review, and I am not treating it as one. grok has not posted here.

So this sits open rather than merging. Everything I have is my own verification, and an author reading their own guard finds it obviously right — which is exactly the defect class that gets missed. The three places I would most want another pair of eyes are listed in my review-request comment above: the widened allowlist staying an allowlist, the measured MIME map contradicting this repo's own `AUDIO_MIME_BY_EXT`, and whether my Chromium harness is acceptable evidence given the panel Playwright suite does not cover `show_media` at all.

**To unblock:** a review on this PR. If the `/view` probe widening (point 5) reads as out of scope for a freeze, say so and I will split it into its own PR and merge the allowlist alone.

panel#1572 stays OPEN and is deliberately **not** labelled `autopilot:parked` — the work is done and held for review, which is not the same thing as parked.

artokun:
## Gate P1 accepted — the symptom is real, and worse than reported. The prescribed remedy is not the fix.

Head `e20f8f5`.

### 1. The finding is real, verified by execution not by reading

I disclosed this branch as out of scope; that was the wrong call and the gate was right to push. Proof rather than agreement — the gate turned off (`targetIsHeadless = false`, i.e. exactly this PR's previous head), `isHeadless() === true`, one `{path}` audio item and one `{filename}` audio ref:

```
isError: false
dispatched: [
  { kind: "audio",   filename: "…\\take.wav", dataUrl: "data:audio/wav;base64,AA…" },
  { kind: "viewRef", filename: "s.wav" }
]
```

Both reached the client. So the gate's `18227` finding is correct, **and it understated the blast radius**: it named only the `/view`-ref branch, but the `{path}` branch inlines regardless of headless, so my own allowlist widening had already opened a second door to the same dead end. That one is on me, not on the gate.

### 2. Remedy 1 (inline it) is refuted by the client's own source

`comfyui-mcp-mobile` @ `c7b0dbc`, checked out on this machine:

- `bridge_frames.dart` — `MediaItem.isVideo` decides by mime (`video/` → true, `image/` → false) else by extension from `{mp4, mov, webm, m4v, avi, mkv}`. `audio/wav` matches no mime prefix and `.wav` is in no set → **false**.
- `media_card.dart` — `MediaCard.build` is `if (item.isVideo) _video(context) else _still(context)`. **Two branches. There is no audio branch.** `_still` builds `MemoryImage(inlineBytes)` (or `NetworkImage(viewUri)`), and `Image(errorBuilder:)` degrades to a caption row with an **image icon** when the decode fails.
- `grep -rn "audio" lib/features/chat/` → **one hit**, a `'generate_audio': '🎵 Generating audio'` status string. No renderer, no player.
- `bridge_client.dart` — replies `{'shown': true}` to **any** `show_media`, without reading the items at all.

So inlined audio bytes reach `MemoryImage`, fail to decode, and land in the *same* caption fallback a bytes-less ref produces — after shipping up to 20 MB of base64 to a phone to arrive at the identical non-result. Inlining does not produce a player; there is no player to produce. My original hedge was "I could not verify"; measured, the answer is not "unverified" but **"it cannot, and inlining makes it slower without making it work."**

Caveat stated plainly: **Dart and Flutter are not installed on this machine**, so the client evidence is READ, not EXECUTED. Everything on the orchestrator side is executed.

### 3. So: resolution 2, in its stronger form

Audio aimed at a headless target is now **refused**, naming the limitation — rather than dispatched to a client that will acknowledge it as `shown`. A dead card that reports success is the unearned-claim shape; a refusal that names the limit is honest. The refusal says nothing is wrong with the file, names where it is, and gives two moves that work (ask the user to play it on the desktop panel, which does have a player; or show something this client can present and say the audio is not playable here). It explicitly does **not** read as "unsupported format" — that would send the caller off to re-encode a perfectly good file.

Gated at **both** source forms, as two explicit call sites sharing one builder (`headlessAudioRefusal` in `comfy-view-ref.ts`), because both end at the same client and a one-sided fix is the exact parity bug this repo keeps hitting. The predicate is sticky `isHeadless`, matching the sibling inline branch rather than `isCurrentHeadless`: this is a rendering decision about the client that will receive the card, including a phone that is offline now and gets it on the mailbox flush.

The `image/`+`video/`-only condition at the old `18227` is **kept as-is** and its comment now records why: audio is refused earlier, so that condition is unreachable for audio rather than merely narrow — and the next reader who spots it does not "fix" it by inlining.

### 4. Round-2 mutation evidence — 5 mutations, all applied, all killed

| # | mutation | verdict | killed |
|---|---|---|---|
| MH1 | delete **call site 1** (`{path}`) only | KILLED | 10 — the 8 PATH refusals + 2 |
| MH2 | delete **call site 2** (`/view` ref) only | KILLED | 8 — the REF refusals |
| MH3 | gate off entirely (= this PR's previous head) | KILLED | 18 |
| MH4 | gate over-broad (`always true`) | KILLED | 35 — the browser panel loses audio |
| MH5 | drop the `typeof` guard | KILLED | 1, named for it |

MH1 and MH2 kill **disjoint** sets, which is the point: neither call site is covered by the other's tests.

MH4 matters as much as MH3 — an over-broad gate would have "fixed" the P1 by silently taking audio away from the browser panel, i.e. by undoing this PR while staying green.

### 5. Unchanged, as instructed

The measured MIME table (`.opus` → `audio/ogg`, `.flac` → `audio/flac`), the deliberate refusal of `.wma`/`.aiff`/`.mid`, `.webm` staying video, and the explicit allowlist are all untouched. `tsc --noEmit` clean, `check:vocabulary` clean, and the six `show_media`/view-ref files are **157 passed / 0 failed**.

### 6. On the gate's own caveat

Noted that Vitest was blocked by a sandbox `spawn EPERM`, so the verdict rested on reading. That is precisely why I reproduced the finding by execution before acting on it rather than taking it on faith — and it is also why I am flagging that the finding's *scope* was wider than stated. Re-gating now.

artokun:
## Round 3 — two more P1s, both accepted, both mine

Head `cc4b07c`. Two gate runs landed on `e20f8f5` and reported **different** P1s. Both were real, both were defects I introduced in round 2, and both are fixed.

---

### P1-a — the probe blessed a mismatch (`18358`)

> `{filename:"plate.png"}` returning `audio/wav` produces a broken image card but is reported as valid media.

Correct, and it is the sharper version of my own round-1 change. I widened the #941 probe from "image or video" to "image, video or audio" — which fixed one false alarm and created a false **clearance**.

**"Is it media" was always the wrong question.** The probe exists to predict a card that renders BROKEN, and the panel picks its painter from the ref's **filename**, never from the body. So the test that matters is whether the served family MATCHES the family the name promises. It now does.

That is strictly better than either previous version:

| filename | served | old code | round 1 (mine) | now |
|---|---|---|---|---|
| `take.wav` | `audio/wav` | flagged (false alarm) | ok | ok |
| `plate.png` | `audio/wav` | flagged | **ok (the P1)** | flagged |
| `take.wav` | `image/png` | **ok (older latent hole)** | ok | flagged |
| `mystery` | `audio/wav` | ok | ok | ok — see below |
| anything | `text/html` | flagged | flagged | flagged |

Row 3 is worth noting: the ORIGINAL code accepted an audio-named ref serving `image/png` just as wrongly, in the other direction. Nobody had reported it; the family comparison closes it as a by-product.

When the filename yields no family, any media body is accepted. That is a deliberate **false negative**, pinned by a test so it reads as a decision: the panel gives an unclassifiable name a link card, which is not a broken card, and a diagnostic that cries wolf gets ignored. It also means an exotic-but-real image (`.bmp`, `.avif` — in the panel's image set but not the orchestrator's) is never wrongly accused.

---

### P1-b — the headless verdict read a tab the frame was leaving (`18066`)

> headless status is captured before production `ctx.call` can rebind the session's target — if an offline phone session auto-rebinds to the sole desktop tab, a valid `.wav` is refused.

Correct, and verified against the real code rather than taken on faith. `makePanelToolCtx` says it outright: `ctx.tabId` is "held on the returned ctx object (NOT captured by value) so an explicit rebind can re-point this session in place", and `ctx.call` runs `ensureReachable()`, which rebinds an unreachable session onto the sole live tab.

**But moving the read later does not, on its own, fix it** — and I want to be precise rather than claim a fix I did not make. `ensureReachable()` runs *inside* `ctx.call`, so it is strictly after any check this handler can perform. I tried to write a mid-flight-rebind regression test on the small-file route and it **failed**, which is how I found this: that route is synchronous all the way to `ctx.call`, so nothing can move underneath it there. I deleted the test rather than bend it, and said so in the file.

**What actually answers the false refusal is the second condition: `canReach`.** An unreachable sticky-headless tab is *precisely* the session about to be rebound — and `interactiveTabIds()` filters headless out, so a heal can only ever land on a canvas tab, where audio plays perfectly. So the gate now fires only for a phone that is **connected right now**. When the heal cannot fire (2+ interactive tabs, a pinned session, nothing bindable), the call surfaces the bridge's own tab-listing error, which is a better message than my refusal would have been.

This reverses my round-2 reasoning, which explicitly chose sticky `isHeadless` over liveness. That reasoning was wrong and the gate caught it.

The late placement is kept anyway and does earn its place on one route: the **oversized** path awaits (`resolveServableViewRef`, staging) before dispatch, and the ctx is shared for the whole orchestrator-scoped session, so the tab genuinely can move across that boundary. There is a regression test there that flips headlessness during the await and asserts the audio goes through — it fails if the verdict is a boolean captured before the loop.

Structurally: two record sites, **one** verdict, taken as the last statement before the frame goes out. That also removes the two-call-site parity risk round 2 introduced.

A bridge too lightweight to answer `canReach` is treated as **reachable**: `isHeadless` already said "phone", and inventing liveness we cannot observe is how an unearned claim gets made in the other direction. Pinned by a named test.

---

### Round-4 mutation evidence — 10 mutations, all applied, all KILLED

| # | mutation | verdict | killed |
|---|---|---|---|
| MP1 | probe: drop the family comparison | KILLED | 6 |
| MP2 | probe: never predict a family from the filename | KILLED | 6 |
| MP3 | probe: stop predicting **audio** from the filename | KILLED | 2 |
| MP4 | probe: stop reporting a non-media body | KILLED | 3 |
| MH-a | stop recording `{path}` audio | KILLED | 13 |
| MH-b | stop recording `/view`-ref audio | KILLED | 8 |
| MH-c | delete the verdict entirely | KILLED | 21 |
| MH-d | **refuse on stickiness alone** (the false refusal) | KILLED | 9 |
| MH-e | refuse on reachability alone | KILLED | 39 |
| MH-f | treat a missing `canReach` as unreachable | KILLED | 1, named |

**Two of these survived in round 3 and are the reason there is a round 4.** Reporting them because they say more about the tests than the passes do:

- `MP2` as first written was `false ? "image" : <original>` — textually changed, **semantically inert**. My harness flags an unmatched anchor but could not see an applied-yet-identity mutation, so it read as a survivor. Redone properly (`expected = null`) it kills 6.
- `MP4` genuinely survived: every case I had written also failed the family comparison, so the media check was never load-bearing on its own. The case that isolates it — a non-media body under a filename with **no** known extension, where there is no family to compare — had no test. It does now, and MP4 kills 3.

---

### Unchanged, as instructed

Measured MIME table (`.opus` → `audio/ogg`, `.flac` → `audio/flac`), the deliberate refusal of `.wma`/`.aiff`/`.mid`, `.webm` staying video, the explicit allowlist. `tsc --noEmit`, `check:vocabulary` (1761 files) and `i18n:check` all clean; the six `show_media`/view-ref files are **185 passed / 0 failed**.

This is gate round 3 on this PR, and each round has found a real P1 — so I am re-gating rather than merging on the strength of "it passed the last two findings". Re-gate running now.

artokun:
## Round 4 — scope addresses. Accepted; this was the best of the three.

Head `fa7152d`.

> `ctx.tabId` is `orchestrator::codex` and resolves to a connected headless phone → `isHeadless(ctx.tabId)` is false → `.wav` dispatched and falsely acknowledged as shown.

Verified against the source rather than taken on faith, and it is exactly right. `resolveTarget`'s scope branch documents its own fallback order: *"the tab the user last talked from, else the most recently connected interactive (canvas-owning) tab, **else the most recent headless one**."* Meanwhile `isHeadless("orchestrator::codex")` is a plain lookup miss — `conns` is keyed by tab id and `headlessSeen` never holds a scope address — so it answers `false`. A codex-backend session sitting on a phone walked straight through a gate built entirely to stop it.

**Both of my previous corrections were the same mistake in different clothes: I kept judging the ADDRESS instead of the DESTINATION.** Round 3 added `canReach` to catch the offline-phone case; round 4 shows the address may not be a tab at all.

`liveTabIdFor` answers both at once, because it *is* `resolveTarget` with the throw swallowed — it resolves a scope address, an 8-char prefix, or a migration alias to the concrete tab a command would reach, and returns `undefined` precisely when nothing resolves, which is the case that must **not** be refused (the call is about to be healed, or to surface the bridge's own tab-listing error — a better message than my refusal).

Three tiers, best signal first, so a lightweight bridge degrades instead of throwing: **routing → reachability → the bare id.** All three are pinned by tests. Residual window acknowledged in the comment: `ctx.call` re-resolves a moment later, and that drift only ever moves toward a canvas tab, which is the safe direction.

### Round-5 mutations — 5, all applied, all KILLED

| # | mutation | verdict | killed |
|---|---|---|---|
| MS1 | judge `ctx.tabId` directly (**the round-4 P1**) | KILLED | 13 |
| MS2 | drop the routing tier | KILLED | 4 — all four scope tests |
| MS3 | drop the reachability tier | KILLED | 9 |
| MS4 | treat an unroutable address as headless | KILLED | 9 |
| MS5 | delete the verdict | KILLED | 24 |

MS2 is the one that matters: it kills *only* the scope tests, so that tier is not covered by anything else.

### State

`tsc --noEmit` clean, `check:vocabulary` clean (1760 files), six `show_media`/view-ref files **191 passed / 0 failed**. The measured MIME table, the `.wma`/`.aiff`/`.mid` refusals, `.webm`-stays-video and the explicit allowlist are all still untouched.

This is gate round 4, and every round has returned a real P1 — three of them defects I introduced while fixing the previous one, which is worth saying plainly rather than presenting this as a clean run. Re-gating on `fa7152d` now. I am not merging on my own verification.

artokun:
## Round 5 — accepted. `show_media` is the one command that does not fail when it cannot route.

Head `b779ee8`.

> `liveTabIdFor() === undefined` is treated as safe, but unreachable scope/pinned sessions are not auto-rebound; `show_media` is mailboxed as success.

Correct, and it lands squarely on a sentence I wrote in round 3: *"when the heal cannot fire the call surfaces the bridge's own tab-listing error, which is a better message than this refusal would have been."* That is false, and it is false for exactly one command:

```
private static isMailboxable(cmd: BridgeCommand): boolean {
  return cmd.cmd === "show_media";
}
```

So when `resolveTarget` throws, this command alone is not refused — `storeMailbox(opts.tabId, cmd)` buffers it and returns `{ok: true, mailboxed: true}`, and the next hello replays it with `mailbox: true`. My "unroutable is therefore safe" assumption was the one case where the frame is neither delivered nor dropped but **parked**.

Two recipients, both bad:

- **concrete offline phone id** — `flushMailbox` uses `conns.get(tabId)`, so it goes to that exact phone when it wakes;
- **scope address** — worse. `flushMailbox(scopeKey, this.conns.get(tabId))` delivers a scope-keyed box to *"the first tab that hellos"* (its own comment). Nobody has identified the recipient at queue time, and it may be the phone.

`ensureReachable()` also returns early for scope addresses and for pinned sessions, so neither is auto-rebound — the gate is right about that too.

### The rule, restated around what actually happens

The distinction is not reachable-vs-unreachable, it is **healable-vs-not**:

1. Run the production heal first — `await ctx.awaitReachable?.(0)`, which *is* `ensureReachable`, the same call `ctx.call` makes moments later. A session that can be rebound already has been, so the round-3 case (asleep phone, one desktop tab) resolves onto the desktop tab and plays. That case is preserved, and now for the right reason rather than a wrong one.
2. If the address **resolves**, judge the concrete tab it resolves to.
3. If it **does not**, the frame will be mailboxed under `ctx.tabId` — so a concrete id names its own recipient (refuse if that is a phone) and a scope address names nobody (refuse).

An unknowable recipient gets its **own** refusal, deliberately not merged with the phone one: telling the caller "your session is on a phone" when nobody knows what it is on would be its own unearned claim. It says the audio was not queued, why queuing is not available to it, and what to do instead.

Narrow on purpose: an unreachable **non-headless** tab still gets its audio, because it owns its own mailbox and that mailbox names a real player. And an unroutable scope still passes **images** through — only audio is held back.

### Round-6 mutations — 6, all applied, all KILLED

| # | mutation | verdict | killed |
|---|---|---|---|
| MB1 | treat an unroutable address as safe (**the round-5 P1**) | KILLED | 3 |
| MB2 | let an unroutable scope queue audio | KILLED | 1, the scope test alone |
| MB3 | let an unroutable concrete phone queue audio | KILLED | 2 |
| MB4 | skip the production heal | KILLED | 8 — the healable case |
| MB5 | merge the two refusal messages | KILLED | 1, named |
| MB6 | delete the verdict | KILLED | 27 |

Each kills a disjoint case, so no tier is riding on another's coverage.

### Tests I had to REVERSE, and why that is worth flagging

Ten tests flipped. They asserted "an offline phone is allowed through" and cited the tab-listing error that does not exist for this command — my round-3 belief, encoded as green tests. Green tests that assert a false premise are worse than no tests, so they are rewritten around the healable/not distinction rather than deleted, with the corrected reasoning in the file.

`tsc --noEmit` clean, `check:vocabulary` clean, six `show_media`/view-ref files **194 passed / 0 failed**. The measured MIME table, the `.wma`/`.aiff`/`.mid` refusals, `.webm`-stays-video and the explicit allowlist remain untouched.

Gate round 5 running on `b779ee8`. Five rounds, five real P1s, four of them mine — I am not merging on my own verification.

artokun:
## Reduced to the reported bug. Round 5 gated SHIP first, so that verdict is on record.

Head is now `98fc30e`, force-pushed. The previous head `b779ee8` is preserved as **`spike/1572-client-routing-audio`** so nothing is lost and the spun-out issues can point at real code.

### Why

Five gate rounds, five real P1s — but four of them were defects introduced while fixing the previous one, and the change had grown from "the allowlist rejects `.wav`" into reworking tab resolution, reachability, probe semantics and mailbox recipient judgement. Each finding being real does not make continuing correct; it is evidence the surface got too large to stabilise in one PR.

### The scope question, answered by execution

The report is a TTS session in the ordinary sidebar panel: `panel_show_media` with `{source:{path:"D:/.../reuse_halo_c1.wav"}}`. I ran that exact shape against a tree with the probe change and the entire headless gate removed:

```
isError:      false
kind:         "audio"
dataUrl:      "data:audio/wav;base64,…"
probeRan:     false      ← unverifiedRefs is fed ONLY by the {filename} viewRef branch
viewRefRoute: false      ← a small {path} item is inlined; it never becomes a ref
```

So rounds 2–5 are adjacent territory. None of them is on the reporter's path.

### What this PR is now

`git diff --stat origin/main...HEAD` — **6 files, 449 insertions, 7 deletions**:

- the `AUDIO_EXTS` allowlist and the **measured** `AUDIO_MIME` map (`.opus` → `audio/ogg`, `.flac` → `audio/flac`; the obvious values are refused by Chromium before any decoder runs);
- `kind: "audio"` and the widened refusal message;
- `ShownMediaKind` plus the three note branches it forced — kept because they are a *consequence* of the allowlist change, not adjacent work: without them an oversized `.wav` is described to the agent as an image displayed "at full size", a lie this change would have introduced;
- the tool description;
- tests + changelog.

### What was removed, and where it went

| removed | filed as |
|---|---|
| headless-audio refusal (R1/R3/R4/R5) | #2010 |
| `/view` probe family comparison (R2) | #2011 |
| address-vs-destination resolution (R4) | #2012 |
| `isMailboxable` / mailbox recipient (R5) | #2013 |

Each carries the executed evidence and the traps, including the ones that cost a round: inlining audio to the mobile client does **not** help (`MediaCard.build` has two branches and no audio path, so the bytes reach `MemoryImage` and fail the same decode), and the naive probe widening creates a false clearance for `plate.png` serving `audio/wav`.

### Tests: which stayed, which went

**Stayed** — everything that tests the allowlist and its consequences: the whole `panel-show-media-audio-ext.test.ts` (both directions of the refusal, per-extension MIME, the two named regressions for `audio/opus` and `audio/x-flac`, `.webm` stays video), the audio note tests in `comfy-view-ref.test.ts`, and the three oversized-route tests that exercise audio through reference and staging.

**Went with the spike** — `panel-show-media-audio-headless.test.ts` entirely, `panel-show-media-audio-viewref.test.ts` entirely (it existed only for the probe), and the two "verdict reads the tab live" tests in the oversized file.

That includes the ten tests I reversed last round. They belonged to the headless surface and went with it — they are on the spike branch and referenced from #2010, not silently dropped.

### Re-verified on the reduced surface

Seven mutations, all applied, all **KILLED**: delete the `AUDIO_EXTS` branch (the call site — 21 tests), `.opus`→`audio/opus` (2, both named), `.flac`→`audio/x-flac` (2, both named), drop `.opus` (4), accept-anything instead of the allowlist (10, incl. #811's own refusal test), and the two note branches (3 each).

`tsc --noEmit`, `check:vocabulary` (1759 files) and `i18n:check` all clean. Four `show_media`/view-ref files: **118 passed / 0 failed**. Gating the reduced diff now — a fresh, small review, not round 6 of the old surface.

artokun:
## Gate on the reduced diff: NO-SHIP, two P1s. One is a real regression from THIS PR and is now fixed; the other is pre-existing and stays deferred.

Head `b06f9a0`. Both findings landed in the territory that was just spun out, so I measured each against `origin/main` rather than assuming the reduction had covered them.

### P1-a — headless audio refs never inlined → **accepted, and it is a regression this PR introduced**

Same call, same fixture, measured on both sides:

| | result |
|---|---|
| `origin/main` | `isError: true` — `unsupported file type ".wav"`, **nothing dispatched** |
| this PR, before the guard | `isError: false` — **dispatched** as `kind:"audio"` with a `data:audio/wav` URL |

The mobile client has no audio branch (`MediaCard.build` is `if (item.isVideo) _video else _still`) and `BridgeClient` replies `{'shown': true}` to any `show_media` without reading the items. So that second row is **an honest refusal turned into a false success** — the agent told the take played when nobody heard it.

That is not adjacent territory. Opening the allowlist is what causes it, so it belongs in the PR that opens the allowlist. Fixed with a guard scoped to exactly the branch that changed: the `{path}` audio arm.

**Why the imprecise predicate is enough *here*.** Sticky `isHeadless` misses a scope address and over-refuses a phone that is offline and about to be rebound — the two errors that cost rounds 3–5. Neither matters at this scope, because **the guard's worst case is exactly main's current behaviour**: main refuses *all* audio on this branch, so the guard can only ever refuse something main also refused, and can never invent a success main did not have. It is monotonically no-worse-than-main by construction, which is the property a regression guard needs and the property a general destination gate does not have. Getting the destination right in general needs the routing resolver, the mailbox rule and a mobile-side audio card — #2010 / #2012 / #2013.

Note this is narrower than the gate's own prescription. Inlining the bytes, which is what "never inlined" implies, does **not** work: they reach the same `MemoryImage`, fail the same decode, and land in the same caption fallback after shipping up to 20 MB to a phone. Evidence in #2010.

### P1-b — valid audio refs reported as non-media → **pre-existing, correctly deferred**

This is the `/view` probe, and it is wrong on `main` today: an `audio/wav` body under any filename is pushed to `nonMedia`. **This PR does not touch the probe** — `git diff origin/main...HEAD -- src/orchestrator/panel-tools.ts | grep -c mime.startsWith` → `0` — and it does not make the behaviour more reachable, because `unverifiedRefs` is fed only by the `{filename}` viewRef branch, which is unchanged.

It is filed as **#2011** with the full analysis, including the trap that the naive fix (just add `audio/` to the test) creates a false *clearance* for `plate.png` serving `audio/wav`. Fixing it here would re-import the surface this PR was reduced to escape, and it would ship the same defect review already caught once.

### Guard mutations — 4, all applied, all KILLED

| # | mutation | verdict | killed |
|---|---|---|---|
| G1 | delete the guard (reinstates the false success) | KILLED | 9 |
| G2 | drop the `typeof` guard | KILLED | 1, named |
| G3 | invert it (refuse the browser panel instead) | KILLED | 31 |
| G4 | widen it past audio | KILLED | 10 |

G3 and G4 matter as much as G1: an inverted or over-broad guard would "fix" the P1 by silently undoing the PR, or by taking images and video away from phones, while staying green.

### State

Diff is **6 files, 601 insertions, 7 deletions**. `tsc --noEmit`, `check:vocabulary` (1760 files) and `i18n:check` clean; four `show_media`/view-ref files **130 passed / 0 failed**. Re-gating now.

artokun:
Head `732b407`. The gate was right a second time about the same guard, and the correction is worth stating plainly rather than burying.

## What I claimed, and why it was false

I wrote that the guard`s worst case was exactly main`s behaviour, so its imprecision was harmless. **That claim does not hold**, and the gate found where:

> `ctx.tabId = "orchestrator::claude"` resolving to a phone → `isHeadless` returns false → audio dispatched → false `{shown:true}`.

`ctx.tabId` may be a SCOPE ADDRESS (#884). It is not a tab — `conns` is keyed by tab id and `headlessSeen` never holds a scope address — so the lookup misses and answers `false`, while `resolveTarget`s scope branch ends *"…else the most recent headless one"*. On that path main refused the audio and my guarded version dispatched it. That is the regression still live, in the one shape I had waved off as "acceptable imprecision".

Third time I have made a confident claim about this guard that measurement did not support. Recording it because the pattern is the finding: I keep asserting a safety property instead of deriving it.

## The fix is a reframing, not a bigger surface

The guard was asking **"is this a known phone?"**. It now asks **"is this a proven non-headless destination?"**:

```
routedTab = liveTabIdFor(ctx.tabId) ?? ctx.tabId     // resolveTarget, throw swallowed
refuse if routedTab == null || isHeadless(routedTab)
```

`liveTabIdFor` answers for scope addresses, id prefixes and migration aliases alike, and returns `undefined` when nothing resolves. Phrased that way the monotonicity argument is *derivable* rather than asserted: audio goes out only when a tab is named and is not headless; **every other case — phone, unroutable, unknown — is refused, and every one of those is a case main refused too**, because main rejects all audio on this branch. The guard can only ever refuse what main also refused.

An unroutable session gets its **own** message. Saying "you are on a phone" when nothing resolved would be a fresh unearned claim, which is the thing this whole change exists to avoid.

Still a regression guard, still one branch. It does not make audio work on those clients (#2010), does not touch the `/view`-ref path, and does not apply the resolver anywhere else (#2012).

## Guard mutations — 7, all applied, all KILLED

| # | mutation | verdict | killed |
|---|---|---|---|
| G1 | delete the guard | KILLED | 12 |
| G2 | **judge `ctx.tabId` instead of what it resolves to** (the gate`s finding) | KILLED | 4 |
| G3 | let an unroutable address through | KILLED | 1, named |
| G4 | invert it (refuse the browser panel) | KILLED | 37 |
| G5 | drop the `typeof` guard | KILLED | 1, named |
| G6 | widen it past audio | KILLED | 10 |
| G7 | merge the two refusal messages | KILLED | 1, named |

G2 is the regression test for this round specifically: a bridge that answers `true` for the address and `false` for the resolved tab must still let audio through, and one that answers `false` for the address and `true` for the tab must refuse.

## State

Diff: **6 files, 746 insertions, 7 deletions** — still only `panel_show_media`s path branch and the three notes the kind widening forced. `tsc --noEmit`, `check:vocabulary` (1760 files) clean; four `show_media`/view-ref files **136 passed / 0 failed**. CI green on the previous head, re-running. Re-gating now; not merging on my own verification.

artokun:
## Reduced. Head `29dc352` — allowlist + measured MIME table only.

Round 7 returned two more P1s, both in the headless guard, and one of them was **the same defect class as round 4 reappearing at a second call site** — judging the tab ADDRESS rather than the routed DESTINATION, in a guard I wrote *after* understanding and fixing the first instance. That recurrence is the measurement, not a new bug to chase: seven rounds, seven real P1s, six self-introduced. The guard is out.

### What ships

`git diff --stat origin/main...HEAD` — **6 files, 449 insertions, 7 deletions**: the `AUDIO_EXTS` allowlist, the measured `AUDIO_MIME` map, `kind:"audio"`, the widened refusal message, the three notes the kind widening forced, the tool description, tests, changelog. Nothing else.

Seven mutations on that surface, all applied, all KILLED (call site 21, `.opus` MIME 2, `.flac` MIME 2, drop `.opus` 4, accept-anything 10, the two notes 3 each). `tsc --noEmit`, `check:vocabulary` (1759 files), `i18n:check` clean; 118 passed / 0 failed.

### What was cut, and where the evidence went

| finding | issue |
|---|---|
| headless/mobile client cannot play audio | #2010 |
| `/view` probe asks "is it media" not "does it match the filename" | #2011 |
| address-vs-destination class, **both confirmed sites named** | #2012 |
| `show_media` is the only mailboxable command | #2013 |
| a pre-#710 panel paints `kind:"audio"` as a broken image | #2017 |

Working code preserved on `spike/1572-client-routing-audio` (the resolver version) and `spike/1572-headless-audio-guard` (the narrow guard and its three failures). Nothing was thrown away; #2012 in particular now names both sites so the next person fixes the class rather than a cell.

### The one thing I want read before merge

**This ships a documented regression on the headless path.** Measured, same call, same fixture:

| | headless client + `{path:"voice.wav"}` |
|---|---|
| `origin/main` | `isError: true` — `unsupported file type ".wav"`, nothing dispatched |
| this PR | `isError: false` — dispatched as `kind:"audio"` with a `data:audio/wav` URL |

The mobile client has no audio branch and replies `{shown:true}` regardless, so an honest refusal becomes a false success there. The user hears nothing either way; what changes is that the agent is told it played. It is in the changelog as a KNOWN LIMITATION and #2010 is updated from "capability gap" to "live incorrect-success".

I am not hiding this behind a green gate. It is the price of the cut, it is bounded, it is filed, and the alternative — the guard — drew a fresh P1 in each of three consecutive rounds. Re-gating once; merging on SHIP.

artokun:
## HELD — not merging. Gate agrees, and here is the reachability measurement that was asked for.

Final gate on the reduced diff `29dc352`: **NO-SHIP**, one P1, and it is the exact regression I flagged rather than anything new:

> `panel-tools.ts:18233` — Headless tabs receive `kind:"audio"` and a success result despite having no audio card; a ≤20 MB `.wav` is acknowledged as played while the user cannot play it.

So the gate and my own measurement agree. Not merging.

## Is the headless branch reachable for a user who is not running the mobile client?

Measured, not reasoned. Three findings:

### 1. Only the mobile client ever advertises `headless`

`isHeadless()` is `conns.get(id)?.headless === true || headlessSeen.has(id)`, and both are written from exactly one source — `socketHeadless = msg.headless === true` in the hello handler (`ui-bridge.ts:2548`, feeding `:2636` and `:2716`). Nothing else in either repo writes it.

The **sidebar panel never sends that key.** Its hello frame (`web/js/lib/session-rebind.js`) carries exactly: `type, tab_id, title, panel_version, vocabulary_hash, backend, blind, enforces_workflow_stamp, enforces_workflow_stamp_at_write, accepts_agent_notes`. The only `'headless': true` in the ecosystem is `comfyui-mcp-mobile/lib/core/bridge/bridge_client.dart:193`.

Executed against the real `UiBridge`:

```
new UiBridge() → isHeadless(x) === false for
  "sidebar-tab", "wf:abc:/x.json", "orchestrator", "orchestrator::claude", ""
hasLiveHeadlessClient() === false
```

**An install that has never run the mobile client cannot reach the branch at all.**

### 2. Even with a phone paired, a canvas tab wins

`resolveScopeActive` is explicit — `const best = interactive ?? headless;` — so a headless conn is chosen only when **zero** interactive tabs are connected. A session bound to a concrete sidebar tab id never resolves elsewhere.

### 3. The one exception, and it is still a mobile session

`resolveScopeActive` checks `lastActiveTabId` first — *"tab the user most recently typed in"*. So a scope-addressed session resolves to the phone when the user last typed **on the phone**, or when no interactive tab is connected at all.

## What that means for the trade

**The regression is confined to the mobile/remote client, and only while that client is the session's target.** A plain sidebar user — including the reporter, whose environment is a local ComfyUI with the Claude backend — cannot reach it. That is materially narrower than "everyone", and it is the answer to the question posed.

It does **not** make it harmless for the people it does reach: for them it is precisely a false `{shown:true}`, uncorrectable downstream because `BridgeClient` acknowledges without reading the items.

So the honest framing is a **trade between two user groups**: sidebar users wait longer for a fix that works, or mobile users get told a take played when it did not. I am not making that call unilaterally — it is now on the record with numbers, and #2010 carries the blocking relationship.

## Left as-is

Reduced diff stands (6 files, 456 insertions). CI green. `spike/1572-headless-audio-guard` and `spike/1572-client-routing-audio` both alive — the first is the third consecutive round's evidence that a per-site guard is the wrong shape, and the starting point for #2012's shared resolver.

**To unblock:** land #2010 (mobile audio card) or #2012 (shared destination resolver), then merge this unchanged.

artokun (commented):
## Independent review — head `29dc352`. I wrote none of this code.

I came to panel#1572 through autopilot, found it covered by this PR, and did the one thing that
was actually missing rather than rebuilding the fix. Everything below is **executed on this
machine**, not read. Where I only read something, I say so.

**Verdict: the core change is right and I confirmed it end to end across both repos.** One finding
I would want closed before merge — it costs one line — and one thing about the reduction I think
is a mistake, with the measurement to back it. **The merge is yours**; I am not touching it.

---

## F1 — the known limitation is real, and it is WIDER than the changelog says

You disclosed one shape. There are two. Both reproduced against a **real `UiBridge`** over a real
websocket with a real scope-bound ctx (`makePanelToolCtx(bridge, SHARED_SESSION_SCOPE, …)`),
`{source:{path:"voice.wav"}}`, a real ffmpeg-encoded wav, on `29dc352`:

| session state | `origin/main` | `29dc352` |
|---|---|---|
| a paired **phone** is the only client | `isError:true`, `unsupported file type ".wav"`, 0 frames | `isError:false`, **1 frame dispatched** |
| **nothing connected at all** | `isError:true`, same refusal, 0 frames | `isError:false`, **`{"ok":true,"mailboxed":true}`** |

The second row is not in your known-limitation note, and it is the one I would not ship quietly.
It is your own round-5 finding still live on the reduced surface: `show_media` is the only
`isMailboxable` command, so with nothing connected the audio frame is **parked under the scope
address and replayed to the first client that hellos** — which, by `resolveTarget`'s own fallback
order, may be the phone. On main that call was refused outright. So the disclosure "a headless
client gets a false success" understates it: with *no* client, the audio is queued for an
unidentified future recipient and the call returns ok.

*Honesty about my harness*: in row 1 the reply body the agent finally sees comes from the client,
and my fake socket answered in the panel's shape — ignore that part. The load-bearing observations
are `isError:false` and **one frame dispatched where main dispatched none**. Row 2's
`{"ok":true,"mailboxed":true}` is the orchestrator's own reply with no client involved, so that one
is not an artifact.

## F2 — the guard you deleted was, as far as I can measure, correct

Before you cut it I had already verified `732b407` (now `spike/1572-headless-audio-guard`) against
a **real `UiBridge`**, because your allowlist tests drive the guard with
`bridge: { isHeadless: () => false }` — a stub with no `liveTabIdFor`, so those tests take the
guard's fallback arm, not the arm production takes. Eight scenarios, all passing on `732b407`:

- the reported case (desktop sidebar, scope-bound) → **plays**, `kind:"audio"`, `data:audio/wav`
- scope address whose only tab is a phone → **refused**, nothing dispatched
- desktop **and** phone connected → plays (resolves to the desktop)
- two desktop tabs, no last-active → plays (**no false refusal** — the case I most expected to break)
- concrete tab id → plays · id prefix → plays · no tab at all → refused, nothing queued
- an **image** to a phone → unaffected

And three mutations, all killed:

| mutation | result |
|---|---|
| delete the guard | KILLED — phone case + no-tab case |
| **judge `ctx.tabId` instead of what it resolves to** (the round-4/round-7 defect class) | KILLED |
| refuse everywhere (silently undo the PR) | KILLED — 5 desktop cases |

**On the round-7 finding that killed it.** If it is the `: ctx.tabId` fallback — the one remaining
place the guard judged an address — that arm is **unreachable in production**:
`PanelToolCtx.bridge` is typed as the concrete class (`bridge: UiBridge`, panel-tools.ts:8516),
`makePanelToolCtx(bridge: UiBridge, …)` assigns it directly, all three production call sites
(index.ts:4673, panel-mcp-http.ts:106, panel-tools.ts:17258) pass a real one, and `liveTabIdFor` is
a public method on that class (ui-bridge.ts:3108). So `typeof … === "function"` is always true off
a test bench.

Which means the whole objection dissolves for **one line** — `: ctx.tabId` → `: undefined`. The
guard already refuses on `routedTab == null`, so a bridge that cannot answer becomes "no proven
destination" instead of "judge the address". Then there is no address-judging anywhere in it, the
monotonicity is derivable rather than asserted, and the surface is *smaller* than the version you
cut. `spike/1572-headless-audio-guard` is that commit; the change is one character short of a
revert.

**Caveat, and it matters:** you said round 7 returned **two** P1s and only described one. If the
second is somewhere else in the guard, none of the above addresses it. Post it and I will check it
the same way.

I also want to be fair to the call you made: "each round drew a fresh P1" is real evidence, and
shipping a small correct thing beats grinding a big one. My disagreement is narrow — the specific
artifact you deleted survived every probe I could build against production seams, and the thing it
prevents is a false success, which is the failure mode this repo treats as worst.

---

## Corroborations — re-derived independently, not taken on faith

**Your risk 2 (the MIME table).** All eight play as `data:` URLs in real Chromium, real
ffmpeg-encoded files. Both of your negative claims reproduce with your exact error string:
`audio/opus` and `audio/x-flac` → `SRC_NOT_SUPPORTED: MEDIA_ELEMENT_ERROR: Unable to load URL due
to content type`. **The table is correct.**

**The neighbour you did not probe, which answers your "should one table serve both" question.**
`data:` is not the only route audio takes — a `/view` ref gets its type from *ComfyUI*, not from
your map. Measured against the live server on :8188 with `X-Content-Type-Options: nosniff`:

```
.flac → audio/x-flac            ← the exact string your comment calls refused
.aac  → audio/vnd.dlna.adts
.opus → audio/ogg   .m4a → audio/mp4   .wav/.mp3/.ogg/.oga → as expected
```

`.flac` is `SaveAudio`'s default output, so I expected this to be a live P1. **It is not** — over
HTTP all eight play, including `audio/x-flac` and `audio/vnd.dlna.adts`. Chromium's strictness is
specific to `data:` URLs. So: your divergence from `AUDIO_MIME_BY_EXT` is justified *precisely*
because these answer different questions, no `/view`-side fix is needed, and the second-order worry
you flagged does not bite for browser consumers. (It says nothing about whatever model API consumes
that other table — I did not test that.)

**Your risk 3 (does the panel actually render this).** The cross-repo join, which is the half your
harness could only half-cover. Your fixed orchestrator's **real dispatched items** → panel
`origin/main`'s **unmodified** `composeShowMediaReply` → real `<audio>` in Chromium:

```
<audio> 8   <img> 0   <video> 0   <a> 0
classified: audio ×8      painted=8  unconfirmed=0  unrenderable=[]
video storyboards attempted: 0
all 8 DECODED (wav mp3 flac ogg oga opus m4a aac, dur≈1.00s each)
```

Also the **reference** route, which your oversized tests cover only on the orchestrator side — the
panel classifies a `viewRef` by **filename**, never by your `kind`, so the staged name is
load-bearing. `1755800000000-take.wav` (your `${Date.now()}-${basename}`) and real `SaveAudio`
names through real `/view` URLs: **4 painted, 4 audio players, 0 links, unrenderable=[], all
decoded.** The extension survives staging. Panel media modules on `origin/main`: 140/140.

**Your risk 1 (still an allowlist).** Confirmed — no sniffing, no denylist, refusal still before
`readFileSync`, and the panel's `AUDIO_EXTENSIONS` is your `AUDIO_EXTS` verbatim, so nothing passes
the gate that the card cannot present.

**Your risk 4 (prose regression).** The image and video branches are byte-identical; only the new
audio branch was added. Clean.

---

## Non-blocking

- `AUDIO_EXTS` and `AUDIO_MIME` are two hand-maintained lists with the same eight keys. Nothing
  pins them together, and a future `.weba` added to one only would ship `data:undefined;base64,`.
  `new Set(Object.keys(AUDIO_MIME))` removes the class.
- With no caption, the card's caption and the agent note both show the **full Windows path**
  (`C:\Users\…\take.wav`). Pre-existing and shared with images; just noting it, since audio is the
  kind most likely to be shown from a scratch path.
- The per-item audio disclosure is ~90 words and repeats verbatim per item — eight audio files
  produce ~750 words of near-identical note. Panel `origin/main` (#710), not this PR.

## What I did not do

Not run the panel Playwright suite: no panel file changes here, so branch and baseline are the same
panel tree by construction, and it does not cover `show_media` — you measured that already and I
agree. I substituted targeted real-browser verification of the exact rendering path, which is
narrower but actually about this change. Also read-only, not executed: nothing about the mobile
client (no Dart/Flutter here) — I took your `MediaCard.build` reading as given and it is consistent
with #2010.

Harnesses are throwaway and live outside both repos; nothing here touched your worktree or your
branch.

artokun:
### Correction to my review, before you act on it

I read #2013 after posting. **F1 row 2 is not an unfiled finding** — the mailbox mechanism, including
`flushMailbox(scopeKey, …)` going to "the first tab that hellos", is exactly what #2013 documents,
and you filed it. I implied it was undisclosed. It isn't.

What survives is narrower, and it is about this PR's own disclosure rather than about a missing
issue: the changelog's known-limitation text and the comment above both describe **one** shape —
"a headless client gets a false success". The measured behaviour on `29dc352` has two, and the
second is not mentioned where a reader of the changelog would meet it:

```
nothing connected at all →  main: isError:true, "unsupported file type"
                       → 29dc352: {"ok":true,"mailboxed":true}
```

So the ask is a sentence in the known-limitation note, not a code change: an unroutable session now
**queues** audio it previously refused, and per your own #2013 the recipient may be the phone. A
reader who only sees "headless client" will not go looking for that.

F2 is unaffected by this correction, and I checked its load-bearing claim harder after posting
rather than leaving it on a type signature: `startUiBridge()` (ui-bridge.ts:5514) is a singleton
factory around the only `new UiBridge(` in non-test code, `index.ts:1210` holds that instance, and
it is what reaches all three `makePanelToolCtx` sites. So `ctx.bridge` is a real `UiBridge` in
production and `liveTabIdFor` is always present — the `: ctx.tabId` arm is genuinely dead outside a
test bench, and `: undefined` costs nothing.

Still your merge, and the second round-7 P1 still outranks everything I said — if it is not the
fallback arm, F2 may simply be wrong.

artokun:
## Unblocked — #2010 merged as `c3fdf93`

This PR was held behind artokun/comfyui-mcp#2010, whose final gate named exactly one thing:

> `panel-tools.ts:18233` — Headless tabs receive `kind:"audio"` and a success result despite having no audio card; a ≤20 MB `.wav` is acknowledged as played while the user cannot play it.

That is fixed on `main` now, and **not** by a fourth per-site headless guard.

A `show_media` reply establishes rendering only for the items it **accounts for** — the #710 per-item contract (`count` + `painted` + `unrenderable`), covering the batch that was actually dispatched. The panel's reply passes and comes back byte-identical; a bare `{shown:true}` is demoted under a document that states what was dispatched and that presentation was **not** established, keeping the client's own words verbatim under `client_reply`. Nothing is refused: the item is still dispatched exactly as this PR sends it.

The measured trade in this PR's "known limitation" section is therefore gone. Same call, same fixture, headless target:

| | top-level `shown` | `presented_confirmed` | warns |
|---|---|---|---|
| before #2010 | `true` — the whole answer was `{"shown": true}` | — | no |
| after #2010 | *absent* | `null` | yes |

Widening the extension allowlist can no longer convert an honest refusal into a false success on the mobile path, because there is no longer a success to convert it into. What remains is what this PR always was: a closed door in the orchestrator opened in front of a card the panel shipped in #710.

**Recommend rebasing onto `main` and re-running the gate before merging** — the head this PR's green CI ran against predates `c3fdf93`, and #2010 touches the same handler (the `ctx.call` site at the end of `panel_show_media`), so a textual conflict there is likely. Its own changelog note about the known limitation should also come out, since the limitation no longer exists.

The other half of #2010 is **open, not merged**: artokun/comfyui-mcp-mobile#39 adds the audio branch to `MediaCard` and stops `BridgeClient` answering a blind constant — so a phone actually plays the take rather than getting a better-worded failure. That is a mobile-client release, not something this PR or #2010 delivers today.

artokun:
Post-rebase review gate for `ea6a9afa74fe1e8d62270984de5d23752c0081a5`:

- Reviewed the complete diff against `origin/main`; production changes are confined to `panel_show_media` audio extension/MIME classification and the audio-aware reference/oversize disclosure paths. Unsupported extensions remain an explicit allowlist refusal before bytes are read; audio does not enter video storyboard handling.
- Focused regression gate: 107 tests passed across the three affected test files.
- `npm.cmd run build`: passed.
- Full `npm.cmd test`: 11,824 passed, 4 skipped; 3 failures in the untouched pre-existing `src/__tests__/orchestrator/node-id-union-message.test.ts` (the current error wording includes the field-qualified message but not the test's old line-shape expectation). No failures were in the audio tests.
- `git diff --check`: passed; worktree clean.

Review disposition: SHIP pending the fresh required GitHub checks.

artokun:
Independent post-fix review for `aa432f9`: SHIP.

- `panel-tools.ts:19541-19565` preflights the complete batch before the resolution loop; `stageFileIntoServedDir` is only reached later, so old-panel capability refusal cannot leave a staged copy. Regression coverage: `panel-show-media-oversized.test.ts:528-543`.
- Proven headless destinations reject local audio at `panel-tools.ts:19560-19564`, before stat, staging, or dispatch. Regression coverage: `panel-show-media-oversized.test.ts:545-554`.
- Existing `/view`-reference headless accounting remains unchanged because only `knownAudioPaths` trigger the new headless refusal; #2018 accounting tests remain covered.

The review was read-only; no files or GitHub state were changed.

PR #2005, opened 2026-08-21
fix(1999): a drained ComfyUI-Manager queue is not a completed one — report what the task actually did

### LINKED ISSUE #1999: Panel sync hides failed Manager tasks behind incoherent drained-queue counts
Automatic sidebar-panel sync dispatched ComfyUI-Manager update tasks, then saw a drained queue whose counters were internally inconsistent (total_count=0 while done_count was 3, later 4). The installed panel stayed at 0.13.6 and the operation could neither confirm nor explain the no-op. The git fallback correctly refused because the panel checkout was dirty, so the user was left with repeated warnings and no successful sync.

The newer panel's queue-status surface revealed that all four historical Manager tasks had actually failed ("An error occurred while updating 'comfyui-mcp-panel'"). The older sync path treated the drained/incoherent counters as ambiguous instead of surfacing these terminal failures.

### Steps to reproduce
1. Use Comfy Desktop / Manager v4 with a dirty git checkout of custom_nodes/comfyui-mcp-panel.
2. Run install_comfyui action=panel, panel_action=sync/update.
3. Manager drains with total_count=0, done_count>0, pending=0, in_progress=0, but the pack remains unchanged.
4. Observe repeated "could not automatically sync" warnings; the old status path does not surface the failed task results.

### Exact error (scrubbed)
ComfyUI-Manager's drained-queue counts are incoherent (total=0, done=3/4, pending=0, in_progress=0) and can neither confirm nor explain the no-op.
Git fallback refused: panel checkout has local modifications/untracked files.

### Fix
Applied locally: workaround only. Upstream-only: yes for the sync logic/Manager result interpretation.

Safe recovery used:
- verified the live ComfyUI install root;
- stashed the six local files including untracked files;
- fast-forwarded the panel checkout directly from origin/main;
- reapplied local work, retaining the stash as backup;
- verified panel 0.15.32, restarted ComfyUI, and reloaded the frontend.

Suggested change: when Manager reports a drained queue with total_count=0 and done_count>0, inspect terminal task results/recent failures and report those explicitly. Do not treat "drained" as success; after every panel update, re-read the installed version from disk. Preserve the existing dirty-checkout refusal for git fallback.

### Environment
Windows 11; ComfyUI 0.33.3; ComfyUI frontend 1.49.6; RTX 3060 12 GB; comfyui-mcp 0.52.45; panel before recovery 0.13.6, after recovery 0.15.32; ComfyUI-Manager v4.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-sol` · effort `high` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Fixes #1999.

## What the reporter saw

Automatic panel sync dispatched ComfyUI-Manager update tasks, then read a drained queue reporting `total_count=0` with `done_count` 3 and later 4, and told the user the counts were **"incoherent"** and could **"neither confirm nor explain the no-op"**. The panel stayed at 0.13.6. The git fallback correctly refused (dirty checkout), so the loop repeated. A newer panel's queue-status surface later showed that all four Manager tasks had **failed**.

## What Manager actually reports — established by execution, not by reading

Driven against a live **ComfyUI-Manager 4.2.2** (pip, `--enable-manager`, `/v2/manager/version` → `V4.2.2`, `is_legacy_manager_ui: false`) on this rig, with a throwaway custom-node pack whose git remote points at an unreachable host so `unified_update` fails for real:

| stage | `/v2/manager/queue/status` |
|---|---|
| fresh session, idle | `{total_count:0, done_count:0, in_progress_count:0, pending_count:0, is_processing:false}` |
| task enqueued, worker not started | `{total_count:1, done_count:0, in_progress_count:0, pending_count:1, is_processing:false}` |
| running | `{total_count:0, done_count:1, in_progress_count:0, pending_count:0, is_processing:true}` |
| **drained, task FAILED** | `{total_count:0, done_count:1, in_progress_count:0, pending_count:0, is_processing:false}` |
| **drained, 5 tasks: 4 error + 1 success** | `{total_count:0, done_count:5, in_progress_count:0, pending_count:0, is_processing:false}` |

The last two rows are the finding. From `glob/manager_server.py`:

```python
def done_count(self):  return len(self.history_tasks)                       # CUMULATIVE, session-wide
def total_count(self): return len(self.pending_tasks) + len(self.running_tasks)  # LIVE only
```

They count **different populations**. On v4, `total_count=0` alongside a non-zero `done_count` is not a contradiction and not "nothing was enqueued" — it is simply what an idle queue looks like, with `done_count` equal to however many tasks have ever run in that session. **The counts were byte-identical whether the tasks succeeded or failed.** The reporter's `total=0, done=3` then `4` is exactly this, reproduced.

The per-task record is where the outcome survives — `GET /v2/manager/queue/history?ui_id=…`, observed live:

```json
{"history": {"ui_id": "…", "kind": "update", "result": "An error occurred while updating 'mcp1999-repro'.",
             "status": {"status_str": "error", "completed": true, "messages": ["…"]}}}
```

and for a task that succeeded, `"result": "success", "status": {"status_str": "success"}`. An unknown id answers `{"history": {}}`.

**Which Managers have it** (all three confirmed on this machine):
- **v4 normal ("v2" dialect)** — has it, keyed by `ui_id`, held for the whole session (`MAXIMUM_HISTORY_SIZE = 10000`).
- **v4 legacy-UI ("v2-batch")** — `legacy/manager_server.py` requires `?id=` (a batch history *file*) and returns 400 for a `ui_id` query.
- **3.x** (`custom_nodes/ComfyUI-Manager` 3.41, verified running here) — registers no `queue/history` route at all.

`fetchManagerTaskHistoryEntry` was already v4-gated, so the two older dialects degrade to "unknown" rather than to a guess.

## The bug

Two claims were being made that the data cannot support:

1. `looksLikeManagerNoOp()` treats `total_count === 0` as the stale-3.x "never enqueued" signature. On v4 that is true for **every** drained queue, so the panel update landed in `outcome:"no-op"` and the message asserted *"it never actually enqueued the update. This is the stale ComfyUI-Manager 3.x silent no-op (#639…)"* — an outcome never observed. The same false assertion sat in two more branches on the install/reinstall path.
2. The git fallback narrated `done > total` as *"incoherent … can neither confirm nor explain the no-op"* — the exact sentence in the report. It was neither incoherent nor unexplainable; the explanation was one request away and the ui_id that keys it was minted inside `queueManagerTask` and thrown away.

## The fix

- **`node-management.ts`** — `queueManagerTaskTracked` hands back the `ui_id` the enqueue actually landed under (minted per attempt, exactly as before, so a self-heal retry's id is the one recorded). `updateCustomNode` carries it on `NodeOpResult.taskUiId` for the **single-pack** update only; update-all deliberately does not, because v4 derives per-pack ids (`${uiId}_${pack}`) there and a bare id would answer "no such task" about tasks that really ran. `fetchManagerTaskHistoryEntry` now also reads the structured `status.status_str`.
- **`panel-installer.ts`** — when the on-disk re-read shows nothing moved, the exact task is looked up and the message narrates what Manager recorded. `classifyManagerTaskRecord` is deliberately **asymmetric**: a failure is read only from the structured status (or Manager's own two literal success values when an older v4 omits `status`), never inferred from prose — inferring failure from an unrecognised `result` is the mirror image of the fabricated-success bug. An absent or untrustworthy answer degrades to **"outcome unknown, and here is how to check"**, never to either verdict. The remedy now follows the evidence too, instead of always saying "your ComfyUI-Manager is stale".

**Nothing about what any branch DOES changed** — only what the messages assert. Specifically untouched: the dirty-checkout refusal, `isProvenLegacyEmptyQueue`, the `isProvenQuiescentQueue` gate on the #824 git fallback, and the legacy-dialect gate. A test asserts the dirty refusal still fires with `gitPulls === []`.

## Live verification (loaded modules, real server)

`dist/` imported and **called** against the running v4.2.2, with a fresh failing task:

```
live drained status  : {"total_count":0,"done_count":6,"in_progress_count":0,"pending_count":0,"is_processing":false}
looksLikeManagerNoOp : true          <- the old path would say "stale 3.x, never enqueued"
isProvenLegacyEmpty  : false
fetchManagerTaskHistoryEntry -> {"uiId":"…","kind":"update",
    "result":"An error occurred while updating 'mcp1999-repro'.","statusStr":"error"}
classifyManagerTaskRecord    -> {"kind":"failed","result":"An error occurred while updating …"}
unknown ui_id -> null -> {"kind":"never-ran"}
```

and the sentence the user now gets:

> ComfyUI-Manager RAN the task and it FAILED — its own per-task record reports: "An error occurred while updating 'mcp1999-repro'.". The queue drained afterwards (total_count=0, done_count=6, pending_count=0, in_progress_count=0), which is why the drain on its own looked like a success

versus, when nothing can be correlated (pre-v4, or the lookup failed):

> ComfyUI-Manager's queue drained (…) and what THIS task did could NOT be established. Those are queue-WIDE drain counters, not a per-task result: on ComfyUI-Manager v4 an idle queue always reports total_count=0 with done_count as the whole session's history length, so they read identically whether the task succeeded or failed. The record that would say (GET /v2/manager/queue/history?ui_id=…) could not be read here, and ComfyUI-Manager older than v4 does not keep one at all

## Mutation evidence — every piece deleted, a named test dies

| mutation | test killed |
|---|---|
| M1 `updateCustomNode` stops carrying `taskUiId` (**call site**) | `returns the ui_id it actually enqueued under (#1999)` |
| M2 `queueManagerTaskTracked` stops recording the landed id | same |
| M3 history entry drops `status_str` | `reads status.status_str out of the v4 per-task history record (#1999)` |
| M4 `runPanelActionCore` stops looking the task up (**call site**) | 4 of the `#1999` integration tests |
| M5 `finalizeUpdate` no longer receives the verdict (**call site**) | `registry-ZIP install (no git): finalizeUpdate names the real failure too` |
| M6 the git fallback ignores the per-task verdict | 3 `#1999` tests |
| M7 asymmetry removed (prose inferred as FAILURE) | `ASYMMETRY: unrecognised prose is UNKNOWN, never inferred as a failure` |
| M8 the v4 drained shape called "incoherent" again | 2 `#1999` tests |

Three of the eight are one-line wiring, mutated at the call site rather than in a helper.

## Tests

`npm run lint` clean. Targeted: 792 passed across the 10 files that touch this area. Full suite in the fix worktree: **4 failed files / 7 failed tests**; a clean `origin/main` worktree run of the same suite: **7 failed files / 9 failed tests**. The failing set varies run to run (30s timeouts under machine load) and none of it is in a file this PR touches. `check:unknown-collapse` (#796), `check:vocabulary`, `vocab:export --check` and `asset-counts --check` all pass.

## Deliberately NOT done

- **`looksLikeManagerNoOp` was left as-is.** It is only ever used to fail *closed*, and on v4 "the counters cannot prove work happened" is still true — it is the narration built on top of it that was lying. Narrowing it would change which branch runs, which is a behaviour change this report does not justify.
- **No ui_id threading for install / reinstall / update-all.** `installCustomNodeImpl` has several forks and its own clone fallback; `reinstall` is two queue cycles. Those paths get the corrected *narration* (they no longer assert the stale-3.x cause) but not the correlated lookup. If a report lands on one of them, the tracked helper is already there.
- **No browser suite.** This changes orchestrator-side message text only — no panel JS. Per the brief I am saying that rather than implying coverage I do not have.
- **The `is_processing:true` with `in_progress_count:0` window** was observed mid-run and is not modelled; nothing here reads that combination.

## Review

**Review is PENDING — not gated, not approved.** Requested from grok in a comment below, naming the load-bearing claims.

### DIFF
--- a/src/__tests__/services/manager-dialect-cache.test.ts
+++ b/src/__tests__/services/manager-dialect-cache.test.ts
@@ -89,6 +89,7 @@ const {
   queueUpdateAllCustomNodes,
   setQueueTimingForTests,
   resetManagerApiCacheForTests,
+  fetchManagerTaskHistoryEntry,
 } = await import("../../services/node-management.js");
 const {
   resetManagerApiCache,
@@ -426,6 +427,74 @@ describe("#646 Manager API dialect cache invalidation", () => {
     // … and the queue is started (drained) exactly once, on the live prefix.
     expect(countOf(calls, "/manager/queue/start")).toBe(1);
     expect(countOf(calls, "/v2/manager/queue/start")).toBe(0);
+  });
+
+  // #1999 — the single-pack update must HAND BACK the ui_id it enqueued under.
+  // Without it a drained v4 queue can never be correlated to what our task did:
+  // `/v2/manager/queue/history?ui_id=` is keyed on exactly this value, and the
+  // queue-wide counters read identically for a task that succeeded and one that
+  // errored. This asserts the CALL SITE (updateCustomNodeImpl passing the id
+  // out), not just that the tracked helper exists.
+  it("a single-pack update returns the ui_id it actually enqueued under (#1999)", async () => {
+    const calls = stubServer({ persona: () => "v4" });
+
+    const res = await updateCustomNode({ id: "comfyui-foo" });
+
+    const enqueue = calls.find((c) => c.path === "/v2/manager/queue/task");
+    const sentUiId = (enqueue?.body as { ui_id?: string } | undefined)?.ui_id;
+    expect(sentUiId).toBeTruthy();
+    // The id on the result is the id the Manager was actually given.
+    expect(res.taskUiId).toBe(sentUiId);
+  });
+
+  // The bulk path deliberately does NOT expose one: v4 derives a separate
+  // per-pack task id there, so handing the bare enqueue id to a history lookup
+  // would answer "no such task" about tasks that really ran.
+  it("update-all does NOT claim a correlatable ui_id (#1999)", async () => {
+    stubServer({ persona: () => "v4" });
+    const res = await updateCustomNode({ id: "all" });
+    expect(res.taskUiId).toBeUndefined();
+  });
+
+  // #1999 — the per-task record's STRUCTURED verdict must survive the read.
+  // `result` is free prose; `status.status_str` is Manager's own OperationResult
+  // enum and is the field the classifier decides on, so dropping it would leave
+  // every real failure looking like an unreadable answer. The payload below is
+  // byte-for-byte what a live ComfyUI-Manager 4.2.2 returned for a failing
+  // pack update on this rig.
+  it("reads status.status_str out of the v4 per-task history record (#1999)", async () => {
+    const HISTORY = {
+      history: {
+        ui_id: "task-abc",
+        client_id: "comfyui-mcp",
+        kind: "update",
+        result: "An error occurred while updating 'mcp1999-repro'.",
+        status: {
+          status_str: "error",
+          completed: true,
+          messages: ["An error occurred while updating 'mcp1999-repro'."],
+        },
+      },
+    };
+    vi.stubGlobal(
+      "fetch",
+      vi.fn(async (url: string): Promise<Response> => {
+        const path = new URL(url).pathname;
+        if (path === "/v2/manager/queue/status") return jsonResponse(DRAINED);
+        if (path === "/v2/manager/is_legacy_manager_ui") {
+          return jsonResponse({ is_legacy_manager_ui: false });
+        }
+        if (path === "/v2/manager/queue/history") return jsonResponse(HISTORY);
+        return new Response("", { status: 200 });
+      }),
+    );
+
+    await expect(fetchManagerTaskHistoryEntry("task-abc", BASE)).resolves.toEqual({
+      uiId: "task-abc",
+      kind: "update",
+      result: "An error occurred while updating 'mcp1999-repro'.",
+      statusStr: "error",
+    });
   });
 
   it("keeps a dialect self-heal retry and drain on its original target after retarget", async () => {
--- a/src/__tests__/services/panel-installer.test.ts
+++ b/src/__tests__/services/panel-installer.test.ts
@@ -3186,6 +3186,17 @@ describe("#1999 drained Manager queue: report what was observed", () => {
       );
     });
 
+    it("clamps a runaway Manager result, and SAYS it clamped it", () => {
+      // An unhandled Manager exception puts a full Python repr in `result` — one
+      // measured on the rig ran to several hundred characters of
+      // QueueTaskItem(...). Quoting it whole buries the diagnostic; truncating
+      // it silently is its own small dishonesty.
+      const huge = "E".repeat(900);
+      const text = describeManagerTaskVerdict({ kind: "failed", result: huge }, counts, false);
+      expect(text.length).toBeLessThan(huge.length);
+      expect(text).toMatch(/truncated — the full text is in the ComfyUI server log/);
+    });
+
     it("a failure quotes the Manager's own result and never claims success", () => {
       const text = describeManagerTaskVerdict(
         { kind: "failed", result: "An error occurred while updating 'comfyui-mcp-panel'." },
--- a/src/__tests__/services/panel-installer.test.ts
+++ b/src/__tests__/services/panel-installer.test.ts
@@ -3315,6 +3315,46 @@ describe("#1999 drained Manager queue: report what was observed", () => {
     expect(String(err?.message ?? err)).not.toMatch(/RAN the task and it FAILED/);
   });
 
+  // The branches above all leave through the #724/#824 git fallback (the panel
+  // dir is a checkout). A registry-ZIP install has no `previousRev`, so it
+  // leaves through finalizeUpdate instead — a SEPARATE narrator that made the
+  // same false claim, and would keep making it if only the fallback were fixed.
+  it("registry-ZIP install (no git): finalizeUpdate names the real failure too", async () => {
+    const h = makeDeps({
+      comfyuiPath: COMFY,
+      files: { [pyPath]: pyproject(PANEL_REGISTRY_ID, "0.13.6") },
+      // No `revs` entry → no git-HEAD → not a checkout → no fallback to take.
+      updateDetails: V4_DRAINED,
+      managerTaskRecord: FAILED_RECORD,
+    });
+    const err = await runPanelAction("update", h.deps).catch((e) => e);
+    expect(err).toBeInstanceOf(PanelInstallError);
+    const msg = String(err?.message ?? err);
+    expect(msg).toMatch(/did NOT apply/);
+    expect(msg).toMatch(/RAN the task and it FAILED/);
+    expect(msg).toMatch(/An error occurred while updating 'comfyui-mcp-panel'\./);
+    // The remedy follows the evidence: fix what Manager reported, not "your
+    // ComfyUI-Manager is stale" — the advice this issue was handed four times.
+    expect(msg).toMatch(/traceback is in the ComfyUI server log/);
+    expect(msg).not.toMatch(/never actually enqueued/);
+    expect(msg).not.toMatch(/stale ComfyUI-Manager 3\.x silent no-op/);
+  });
+
+  it("registry-ZIP install with NO readable record: finalizeUpdate says UNKNOWN", async () => {
+    const h = makeDeps({
+      comfyuiPath: COMFY,
+      files: { [pyPath]: pyproject(PANEL_REGISTRY_ID, "0.13.6") },
+      updateDetails: V4_DRAINED,
+      managerTaskRecord: undefined,
+    });
+    const err = await runPanelAction("update", h.deps).catch((e) => e);
+    const msg = String(err?.message ?? err);
+    expect(msg).toMatch(/could NOT be established/);
+    expect(msg).toMatch(/queue\/history\?ui_id=/);
+    expect(msg).not.toMatch(/RAN the task and it FAILED/);
+    expect(msg).not.toMatch(/never actually enqueued/);
+  });
+
   it("a PROVEN update still succeeds — the lookup never vetoes disk evidence", async () => {
     const h = makeDeps({
       comfyuiPath: COMFY,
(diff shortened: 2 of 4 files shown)


artokun:
@grok — review requested. Naming the load-bearing claims and the parts I am least sure of, so you can aim rather than hunt.

**The one claim everything rests on**, and the one to attack first: *on ComfyUI-Manager v4, `total_count=0` with `done_count>0` is the ordinary drained state, not a contradiction and not "nothing was enqueued".* I established it two ways — reading `TaskQueue.total_count()`/`done_count()` in `glob/manager_server.py` (4.2.2), and driving real failing tasks against a live v4.2.2 and watching `done_count` climb 1→2→5→6 while `total_count` stayed 0. If that is wrong, every message this PR rewrites is wrong. What I have **not** established is how far back the v4 lineage that shape holds for — I measured 4.2.2 only, and `describeManagerTaskVerdict` generalises to "ComfyUI-Manager v4" in user-facing prose.

**Second, the asymmetry in `classifyManagerTaskRecord`** (`src/services/panel-installer.ts`). Failure is read only from `status.status_str`; when that field is absent I accept `result === "success"` / `"skip"` as success but refuse to read anything else as failure, so an unrecognised `result` becomes `unknown`. That is a deliberate one-way door — I would rather say "outcome unknown" than invent a failure — but it means a v4 build that omits `status` and reports a failure in prose reads as unknown. Is that the right side to err on here, and is the exact-literal fallback itself sound? (Manager's `task_worker` compares `result_msg == OperationResult.success.value` / `.skip.value`, which is where those two literals come from.)

**Third, the ui_id capture** in `queueManagerTaskTracked` (`src/services/node-management.ts`). `enqueueWithDialectSelfHeal` can invoke the enqueue closure twice; I mint the id inside the closure and keep the last one, on the reasoning that a failed attempt throws so only the landing attempt's id survives, and that the internal v2→v2-batch downgrade reuses the same id. If there is a path where an attempt can *succeed* and still be retried, the recorded id names a task that is not ours — and `fetchManagerTaskHistoryEntry`'s strict `h.ui_id !== uiId` check would not catch that, because the id would match a real (wrong) task. I could not find such a path; please check that I am not wrong about it.

**Fourth, what I chose NOT to change.** `looksLikeManagerNoOp` still returns true for `total_count === 0`, so on v4 the panel update still routes into the `no-op` branch and still fails closed — I changed only what the messages assert, not which branch runs. The alternative (narrowing the predicate so v4 lands in `unverified` instead) is a behaviour change and I judged it out of scope for a report about a misleading message. Say so if that is the wrong call.

**Fifth, the refusal that must not move.** The dirty-checkout refusal in `updateViaGitCheckoutFallback`, `isProvenLegacyEmptyQueue`, the `isProvenQuiescentQueue` gate and the legacy-dialect gate are all untouched by design — the reporter's git fallback refusing was *correct*. A `#1999` test asserts the refusal still fires with `gitPulls === []`. Worth confirming I did not weaken any of them indirectly through the `managerReason` rewrite, which now feeds different text into ~20 message sites in that file.

**Coverage I do not have:** no browser/Playwright run — this is orchestrator-side message text with no panel JS in the diff. And the full suite has 4 failing files here against 7 on a clean `origin/main` baseline, with the set varying run to run under load; none of it is in a file this PR touches, but that is a noisy baseline and I am reporting it as such rather than calling it green.

artokun:
## HOLD — green, unreviewed, and deliberately NOT merged

@grok — re-requesting on the current head `b5bbbc5` (rebased onto `2db8489`, so the review would be of the code as it would land). The risky parts are named in the comment above and are unchanged by the rebase; one thing was added since: `describeManagerTaskVerdict` now clamps Manager's `result` at 400 chars with an explicit "truncated — the full text is in the ComfyUI server log", because an unhandled Manager exception puts a full Python repr in that field.

**State of this PR**

- CI green on the current head against the current `main`: Build macos / ubuntu / windows + Pack & install smoke, all pass. `mergeStateStatus: CLEAN`.
- `check:unknown-collapse` (#796), `check:vocabulary`, `vocab:export --check`, `asset-counts --check` all pass locally.
- 795 tests green across the 10 suites that touch this area, post-rebase.
- Eight mutations, one per piece of the fix, each killing a named test — three of them at the CALL SITE, not the helper. Table is in the PR body.
- Reproduced and then live-verified against a real ComfyUI-Manager 4.2.2 driven on this machine, with the built modules imported and called rather than re-implemented.

**Why it is not merged**

No independent review has come back. The only entry on `/pulls/2005/reviews` is `copilot-pull-request-reviewer[bot]` reporting a **quota limit**, which is not a review. grok has posted nothing on this PR ~47 minutes after the request, and nothing anywhere in this repo since 07:08Z today — every other open PR shows the same single Copilot quota entry and no grok verdict.

My own verification is not a substitute for that, and neither is green CI. This change rewrites what a dozen user-facing refusals ASSERT, on a code path whose whole history is fabricated-certainty bugs — precisely the class where an author reads their own guard as obviously right. `codex-gate.mjs` was not run and no codex was spawned, per the standing instruction.

**What would unblock it**

Either a grok review on `b5bbbc5`, or the owner merging it directly. Nothing here needs more work that I can identify; it needs a second pair of eyes.

Not labelling this `autopilot:parked` — the issue is real, the fix is written and evidenced, and this is a review wait, not a parking decision.

artokun:
**This merged unreviewed** — `663af17`, direct merge at 17:46:34Z. Recording that plainly so a late verdict has somewhere to land rather than being a review nobody read.

@grok if you get to this after the fact, it is still worth a look and a follow-up issue, because the merged code never had an adversarial reader. The one claim to attack is unchanged and is load-bearing for every message this PR rewrites: **on ComfyUI-Manager v4, `total_count=0` with `done_count>0` is the ordinary drained state, not a contradiction and not "nothing was enqueued."** I measured that on 4.2.2 only — by reading `TaskQueue.total_count()`/`done_count()` in `glob/manager_server.py` and by driving real failing tasks against a live instance — while `describeManagerTaskVerdict` generalises it to "v4" in user-facing prose. If that shape differs anywhere in the v4 lineage, the prose is wrong there.

Second, unchanged from the request above: the deliberate asymmetry in `classifyManagerTaskRecord`, where failure is read only from `status.status_str` and an unrecognised `result` becomes `unknown` rather than a failure. That one-way door is intentional, but a v4 build that omits `status` and reports failure in prose reads as `unknown`.

CI was green on the merge commit itself and the merged tree is byte-identical to the head that was tested, so this is not a question of whether it built — it is a question of whether the claim it encodes is true beyond 4.2.2.

Issue #2000, opened 2026-08-21
panel_add_node wrongly requires object_info for frontend-only MarkdownNote after refresh timeout


artokun:
### What happened / root cause
On a healthy live canvas, panel_refresh_nodes timed out fetching /object_info. Immediately afterward panel_add_node for the documented frontend-only type MarkdownNote also refused because object_info was unavailable. MarkdownNote is documented as a frontend-only exemption, so its creation should not depend on backend node definitions.

### Steps to reproduce
1. Open a live ComfyUI canvas.
2. Call panel_refresh_nodes while /object_info is slow.
3. Call panel_add_node with class_type "MarkdownNote".
4. Observe refusal despite the frontend-only allowlist.

### Exact error (scrubbed)
panel_refresh_nodes: object_info_fetch_failed — api.getNodeDefs() did not answer within this refresh's remaining budget.
panel_add_node: cannot verify node type "MarkdownNote" against the ComfyUI backend (object_info is unavailable ...). Refusing to add.

### Fix
Applied locally: no. Upstream-only: yes.
Precise change needed: resolve the frontend-only allowlist before any /object_info availability guard in panel_add_node, so Note/MarkdownNote/Reroute/PrimitiveNode do not require backend defs. For panel_refresh_nodes, use a dedicated/configurable object_info timeout or preserve/reuse a known-good defs snapshot while reporting that the live refresh timed out.

### Environment
Windows 11; ComfyUI 0.33.0; frontend 1.49.6; RTX 4090 22.5GB; comfyui-mcp 0.52.45; panel 0.15.20.

artokun:
Claimed by claude-autopilot 2026-08-21T19:12:19Z.

The refusal is thrown in the **panel** repo (`web/js/lib/node-resolve.js`, `assertAddNodeResolvableRefreshing`), so the fix will land as a PR against `artokun/comfyui-mcp-panel` and this issue will be closed from there.

artokun:
### Status: fix complete, **holding on independent review**

The refusal is thrown in the panel repo, so the fix is **artokun/comfyui-mcp-panel#1586** (open, mergeable, CI green). It is not merged, and I am not merging it myself.

**Root cause.** All three of the panel's `/object_info`-oracle guards ran their *availability* check before the frontend-only exemption:

```js
if (!freshDefs || typeof freshDefs !== "object") { throw … }   // every type, no exceptions
… ever-seen gate …
[...]

PR #2019, opened 2026-08-21
fix(2000): a frontend-only type refused for an unavailable object_info says so

artokun:
Orchestrator half of #2000. The panel-side fix is **artokun/comfyui-mcp-panel#1586**; this is what reaches a user who has not updated the panel yet — and the orchestrator ships far more often than the pack does.

## The bug

On a healthy live canvas, `panel_refresh_nodes` timed out fetching `/object_info`, and the very next `panel_add_node "MarkdownNote"` was refused:

> cannot verify node type "MarkdownNote" against the ComfyUI backend (object_info is unavailable …). Refusing to add.

The panel's add-node guard checked `/object_info` **availability** before consulting its frontend-only allowlist, so a fetch that merely timed out vetoed a type that fetch could never have spoken to — the backend never lists `Note`/`MarkdownNote`/`Reroute`/`PrimitiveNode`, by design.

## What this PR adds

An appended note on that refusal, for allowlisted frontend-only types only. It leads with the **fact**, not a diagnosis:

- the backend never lists this type in `/object_info`, so a failed fetch says nothing about whether it is addable;
- a healthy-but-**slow** ComfyUI produces exactly this state;
- so do **not** reinstall a pack and do not treat it as a missing node.

Then it names both causes **without picking one**. Nothing on this side can tell a stale panel guard apart from a node-type baseline that has not finished loading, and panel#612 is explicit that a definite verdict must not be asserted from absent evidence. The version sentence is added **only** when the tab actually advertised a version below this orchestrator's floor — real evidence — and omitted entirely otherwise.

## The measurement that shaped the design

I nearly gated this on the panel version floor, the way #1828 does. That would have been wrong, and the test that caught it is in the diff:

```
reporter's panel  = 0.15.20
requiredPanelVersion() = 0.15.11     ⇒ compareSemver(...) = +1, ABOVE the floor
```

**The floor tracks bridge-capability requirements, not "the version carrying the latest fixes."** Gating on it would have made this hint fire for almost nobody — and it means **#1828's own skew note could never have fired for this reporter either**, which is worth knowing independently of this PR.

So the note is not version-gated; only the optional skew sentence is.

## Structure

Extends the existing #1828 appender rather than adding a fifth wrapper, so **all four `panel_add_node` call sites** are covered with no wiring change and the two refusal shapes cannot drift apart. Purely additive: it appends text to a result the panel already marked as an error, and can never change a verdict.

I did reorder two pure guards inside that function (allowlist check before refusal-shape check). The 5 pre-existing #1828 tests pass unchanged, and a dedicated test asserts the two shapes stay mutually exclusive.

## Evidence

Mutation results — the tests drive the **real `panel_add_node` handler** end to end, not the helper:

| mutant | result |
|---|---|
| remove the unavailable-refusal branch | **3 killed** |
| drop the frontend-only allowlist gate | **3 killed** |
| always claim version skew | **2 killed** |

10/10 in `add-node-frontend-only-skew.test.ts` (5 pre-existing #1828 + 5 new). Full suite result posted below once it finishes.

The false-positive direction is covered explicitly: a **real** backend type (`KSampler`) hitting the identical refusal keeps the panel's wording untouched and is never told it is frontend-only.

## Review status: PENDING

**Where to look first:**

1. **The reorder inside `withFrontendOnlyPanelSkewNote`.** I moved the allowlist check above the refusal-shape check. Both are pure and both must pass on the old path, so I believe it is behaviour-preserving — but it is the one place this PR touches existing logic, and "obviously equivalent" is exactly what a self-review gets wrong.
2. **Matching on panel prose.** `isObjectInfoUnavailableRefusal` regexes the panel's wording, which this repo normally refuses to do. My justification is that it only ever *appends advice* to a refusal the panel already made — it authorizes nothing, so a miss costs a missing hint and a false positive costs a paragraph. If that line is in the wrong place, say so.
3. **Whether this note is worth shipping at all** once panel#1586 lands. I argue yes: the pack updates far more slowly than the orchestrator, and after the panel fix this shape can still occur for a pending/unseeded baseline — where the advice is still correct.

Fixes the underlying cause reported in #2000 alongside panel#1586.

### DIFF
--- a/src/__tests__/orchestrator/add-node-frontend-only-skew.test.ts
+++ b/src/__tests__/orchestrator/add-node-frontend-only-skew.test.ts
@@ -139,3 +139,86 @@ describe("panel_add_node names panel version skew for allowlisted frontend-only 
     expect(text).not.toMatch(/Ctrl\+Shift\+R/);
   });
 });
+
+// #2000 — the OTHER refusal shape, and the one an old panel emits on a HEALTHY
+// ComfyUI: panel_refresh_nodes timed out fetching /object_info, and the next
+// panel_add_node "MarkdownNote" was refused because the panel's guard checked
+// /object_info AVAILABILITY before consulting its frontend-only allowlist. The
+// panel-side fix is comfyui-mcp-panel#1586; this note is what reaches a user who
+// has not updated the panel yet, and the orchestrator ships far more often.
+
+/** The reporter's exact panel refusal (issue #2000). */
+const UNAVAILABLE_REFUSAL =
+  `cannot verify node type "MarkdownNote" against the ComfyUI backend ` +
+  `(object_info is unavailable — the backend is unreachable or the fetch failed). ` +
+  `Refusing to add rather than trust a possibly-stale node cache (#458). Reconnect ComfyUI and retry.`;
+
+describe("#2000 frontend-only type refused for an unavailable /object_info", () => {
+  it("THE REPORTED CASE: names the fact that /object_info never lists this type, and says not to reinstall", async () => {
+    const r = await addNode("MarkdownNote", UNAVAILABLE_REFUSAL, "0.15.20");
+    expect(r.isError).toBe(true);
+    // The panel's own refusal is KEPT — the note is appended, never a replacement.
+    expect(r.text).toContain("object_info is unavailable");
+    expect(r.text).toContain("FRONTEND-ONLY");
+    expect(r.text).toMatch(/never lists it in \/object_info by design/i);
+    expect(r.text).toMatch(/Do NOT reinstall a pack/i);
+    // A healthy-but-slow ComfyUI produces this exact state; say so.
+    expect(r.text).toMatch(/merely SLOW/i);
+  });
+
+  it("adds the version sentence ONLY when the tab actually advertised a version below the floor", async () => {
+    // The floor is read to pick an INPUT, never to compute an expectation — a floor
+    // that moves must not silently change what this test proves.
+    const BELOW = "0.0.1";
+    expect(compareSemver(BELOW, requiredPanelVersion())).toBeLessThan(0);
+    const below = await addNode("MarkdownNote", UNAVAILABLE_REFUSAL, BELOW);
+    expect(below.text).toContain(BELOW);
+    expect(below.text).toMatch(/HARD-REFRESH/i);
+
+    // No advertised version = no evidence of skew. The note still explains the type,
+    // but must NOT assert a stale panel — panel#612: no definite verdicts from
+    // absent evidence.
+    const unknown = await addNode("MarkdownNote", UNAVAILABLE_REFUSAL);
+    expect(unknown.text).toContain("FRONTEND-ONLY");
+    expect(unknown.text).not.toMatch(/below this orchestrator's floor/i);
+    expect(unknown.text).not.toMatch(/HARD-REFRESH/i);
+    // …and it still gives the caller something to do.
+    expect(unknown.text).toMatch(/wait a moment and retry/i);
+  });
+
+  it("THE REPORTER'S OWN VERSION is ABOVE the floor — so the note must not depend on skew", async () => {
+    // Measured, and it is the reason this note is not gated on the version floor at
+    // all: the reporter ran panel 0.15.20 while the floor is lower, because the floor
+    // tracks BRIDGE CAPABILITY requirements, not "the version with the latest fixes".
+    // Gating on it would have made this hint fire for almost nobody — and it is also
+    // why #1828's skew note could never have fired for this reporter.
+    const REPORTER_VERSION = "0.15.20";
+    expect(compareSemver(REPORTER_VERSION, requiredPanelVersion())).toBeGreaterThan(0);
+    const r = await addNode("MarkdownNote", UNAVAILABLE_REFUSAL, REPORTER_VERSION);
+    expect(r.text).toContain("FRONTEND-ONLY");
+    expect(r.text).toMatch(/merely SLOW/i);
+    // No skew evidence ⇒ no skew claim.
+    expect(r.text).not.toMatch(/below this orchestrator's floor/i);
+  });
+
+  it("does NOT tell a REAL backend type that it is frontend-only", async () => {
+    // The same refusal shape for a type that genuinely needs /object_info must keep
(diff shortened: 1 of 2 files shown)


artokun:
@grok please review — this is the orchestrator half of #2000 (panel half is comfyui-mcp-panel#1586, also awaiting review).

It is a small, purely-additive change, so the risk is concentrated in three specific places rather than spread across the diff:

1. **The reorder is the only existing logic I touched.** Inside `withFrontendOnlyPanelSkewNote` I moved `PANEL_ADD_NODE_FRONTEND_ONLY_TYPES.has(classType)` **above** `isUnknownBackendNodeRefusal(res)`. Both are pure, and both must pass for the #1828 path, so I believe it is behaviour-preserving — the 5 pre-existing #1828 tests pass unchanged. But "obviously equivalent" is exactly the judgement a self-review gets wrong, so please check it rather than take my word.

2. **I match on panel PROSE**, which this repo normally refuses to do. `isObjectInfoUnavailableRefusal` regexes the panel's own refusal wording. My justification: it is used ONLY to append advice to a refusal the panel has already made and already marked as an error — it authorizes nothing, mutates nothing, and cannot change a verdict. A miss costs a missing hint; a false positive costs an extra paragraph on an unrelated error. If you think that still crosses the line this repo draws around parsed prose, I would rather hear it now.

3. **The design turns on a measurement, and I want it checked.** I nearly gated the note on the panel version floor the way #1828 does. That would have been wrong:

   ```
   reporter's panel        = 0.15.20
   requiredPanelVersion()  = 0.15.11   ⇒ ABOVE the floor
   ```

   The floor tracks **bridge-capability** requirements, not "the version carrying the latest fixes". So gating on it would have made this hint fire for almost nobody — **and it means #1828's own skew note could never have fired for this reporter either.** If that reading of `requiredPanelVersion()` is wrong, the whole design of this note is wrong, and #1828 may have a latent gap worth its own issue.

Mutation results are in the PR body; the tests drive the real `panel_add_node` handler end to end, and the false-positive direction (a real backend type hitting the identical refusal) is asserted explicitly.

One thing I am NOT claiming: any browser verification. This only changes text on an error path, and the panel-side behaviour is #1586's business.

artokun:
**Full suite green.** `npx vitest run`: **617 test files, 11308 passed, 4 skipped, 0 failed, exit 0.** All four CI checks pass across macOS / Ubuntu / Windows; branch is `CLEAN`.

Nothing outstanding from my side except the review. The three things I flagged above are still where I'd look — the guard reorder first.

artokun:
### Gate verdict: **INDETERMINATE** — stopping here, not merging

The supervisor asked me to get CI green, run the codex gate, and merge. CI is green. I am **not** running the codex gate, and I am **not** merging. Both, explicitly:

**1. The codex gate is off-limits by the owner's current instruction.** Codex has been repointed at the owner's weakest model to grind through P2/P3 items, so a SHIP from `codex-gate.mjs` would read as an independent adversarial review while not being one. The nudge asking for it is a hardcoded string in `dispatch.mjs` that predates that change. Codex being *reachable* is a different question from codex being *the reviewer*.

**2. No independent reviewer is available.** Measured just now, not assumed:

- Non-Copilot reviews on this PR: **0**. Same on the panel PR.
- Copilot's only entry on both is its **quota-limit refusal**, which is a `/reviews` entry with `state: COMMENTED` — it counts as a review to any unfiltered check, and is not one.
- I re-requested Copilot via `POST .../requested_reviewers`. It returned **200**, and `requested_reviewers` stayed **`[]`** — the request silently does not stick.
- Across the panel repo, the last genuine independent review was **#1552 at 02:51Z**, with **8 PRs merged since with none**.

That is the INDETERMINATE case the nudge itself provides for, so I am taking its own escape hatch.

**Why I am not merging on my own verification.** I wrote the fix and I wrote the tests that pass it. That is precisely the configuration a self-review cannot audit — and on this pair of PRs it has already been demonstrated twice: my own taxonomy pass found a raw-error leak that four careful passes and eight mutants had missed, and measuring the version floor showed the design I was about to ship would have fired for almost nobody. Both were found by *adversarial* checking, not by re-reading. A third such defect is more likely than not, and the panel PR relaxes a security guard while replacing the test that pinned the old behaviour.

**State: both halves complete, green, and waiting on one thing each.**

| | PR | evidence |
|---|---|---|
| Panel (the guard fix) | comfyui-mcp-panel#1586 | 5 commits · 6254 unit tests, 0 fail · 9 mutants all killed · CI green · `CLEAN` |
| Orchestrator (this) | #2019 | 1 commit · 617 files / 11308 tests, 0 fail · 3 mutants all killed · 4 CI checks green · `CLEAN` |

**What unblocks it:** one independent review verdict per PR. Each PR names the specific claims I am least sure of. Nothing else is outstanding.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `fe8e20e505c528f78ffee489f037c21614ceabbb` vs origin/main. Read `withFrontendOnlyPanelSkewNote`, `isObjectInfoUnavailableRefusal`, `frontendOnlyUnavailableNote`, and the existing #1828 matcher — not just the patch.

This is append-only advice on a refusal the panel already made. It cannot authorize an add. That is the right half of #2000 for an orchestrator that ships faster than the pack.

What I checked:

- Allowlist still gates every note. `KSampler` hitting the identical unavailable wording keeps the panel text and never gets FRONTEND-ONLY / Do-NOT-reinstall. Tested.
- #1828 and #2000 stay exclusive. `isUnknownBackendNodeRefusal` matches `Unknown node type … backend does not provide it`. The new matcher is `object_info is unavailable|cannot verify (?:the )?node type`. The reporter #1828 string does not match the new regex, and the new tests assert the #1828 note still wins on that shape.
- Version sentence is evidence-gated. Reporter panel 0.15.20 is above the floor, so gating the whole hint on skew would have silenced this reporter. Only an advertised version below the floor adds HARD-REFRESH. No advertised version names neither skew nor a pack reinstall.
- Reorder (allowlist before refusal-shape) is behaviour-preserving for #1828: both still have to pass. The 5 pre-existing tests remain the pin.

Residual, not blocking: matching panel prose is a miss-costs-a-hint / false-positive-costs-a-paragraph trade, and it is disclosed. Do not let this regex become an authorizer later.

Panel half is comfyui-mcp-panel#1586. This PR is the backstop for a tab that has not updated yet.

Issue #2004, opened 2026-08-21
panel widget writes blocked by object_info starvation; uncollapse rejects stored zero height


artokun:
### What happened / root cause
On a live healthy graph, panel_set_widget for an installed backend node refused because both object_info routes exhausted their shared time budget. Earlier attempts in the same turn appeared to proceed but the values remained defaults when re-read and in the saved workflow. panel_save_workflow then timed out once, although the file later contained the wiring/title edits. A separate panel_edit_node call attempting collapsed:false failed because the collapsed node reported size [228,0]; passing an explicit positive size worked.

### Steps to reproduce
1. On a large custom-node install, add/use OpenAICompatibleLLM.
2. Call panel_set_widget on system_prompt/user_prompt.
3. Re-read the node: values remain defaults, or the call refuses after api.getNodeDefs 10s + GET /object_info 5s timeouts.
4. Save: one save can time out without acknowledgement.
5. Collapse a node, then call panel_edit_node(collapsed:false) without size; it can reject the node's own stored height 0.

### Exact error (scrubbed)
Cannot set widget ... no usable /object_info schema was obtained. api.getNodeDefs() did not answer within 10000ms; GET /object_info did not answer within 5000ms.
workflow_save did not finish within 13s ... modified:true persisted:true.
panel_edit_node: size must be two positive numbers within a reasonable canvas range.

### Fix
Applied locally: no. Upstream-only: yes.
Workaround verified: drive the string widget through a STRING link, and uncollapse while passing an explicit positive size.
Precise change needed: cache and reuse a whole known-good object_info schema for installed types when live refresh times out; make mutation outcomes explicit rather than appearing to land and later reverting. When uncollapsing, restore the pre-collapse size before validating dimensions. For save timeout, return an outcome-unknown retry token or complete/acknowledge the already-started save.

### Environment
Windows 11; ComfyUI 0.33.0; frontend 1.49.6; RTX 4090 22.5GB; comfyui-mcp 0.52.45; panel 0.15.20.

artokun:
Claimed by grok-swarm 2026-08-21T22:11Z — driving this in worktree `wt-2004` (branch `fix/2004-object-info-starvation`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp/pull/2027 (worktree C:/Users/Artokun/code/comfyui-mcp/.worktrees/wt-2004).

PR #2028, opened 2026-08-21
fix(2007): schema-guarded panel edits retry a timed-out object_info instead of staying blocked

### LINKED ISSUE #2007: panel mutations blocked by fixed /object_info timeouts on healthy large install
Live graph reads work and backend health is good, but panel_add_node and panel_set_widget cannot obtain a usable /object_info within their fixed route budgets. panel_refresh_nodes also returns object_info_fetch_failed. This leaves the canvas readable but blocks all schema-guarded edits.

### Steps to reproduce
1. Open a healthy 19-node workflow with many installed custom nodes.
2. panel_graph_outline succeeds with backend_socket=up and mutations_ready=true.
3. Call panel_add_node for core LoadImage.
4. Call panel_refresh_nodes, then panel_set_widget on an existing node.
5. All fail because /object_info does not answer within 10s/5s shares, despite /system_stats and rendering working.

### Exact error (scrubbed)
cannot verify node type "LoadImage" against the ComfyUI backend; object_info unavailable.
panel_refresh_nodes: refreshed=false, reason=object_info_fetch_failed.
panel_set_widget: api.getNodeDefs() exceeded 10000ms share; GET /object_info exceeded 5000ms share; no whole schema observed.

### Fix
Precise change needed: cache the last successful whole /object_info per backend connection and/or allow the schema fetch budget to scale for large custom-node installs; panel_graph_outline should not report mutations_ready=true when schema-guarded mutations are guaranteed to fail. A background schema warm-up with explicit readiness state would avoid blocking core-node additions.

### Environment
Windows 11; ComfyUI 0.33.0; frontend 1.49.6; Python 3.13.12; PyTorch 2.13.0+cu130; RTX 4090 22.5 GB; comfyui-mcp 0.52.45; panel 0.15.20.

artokun:
The 5s/10s `/object_info` shares live in the panel. This process cannot enlarge them, but it can stop treating a timed-out schema fetch on a healthy backend as a dead end.

**What the reporter hit**

`panel_graph_outline` returned `backend_socket=up` and `mutations_ready=true`. `panel_add_node("LoadImage")`, `panel_set_widget`, and `panel_refresh_nodes` then all failed because the panel did not obtain a whole `/object_info` inside its fixed 10s `getNodeDefs` / 5s GET shares. Graph reads still worked.

Calling `panel_refresh_nodes` as the recovery made the next widget write worse: that command clears the last-known schema at the start of the run.

**What this PR changes**

- `panel_add_node` / `panel_set_widget`: on a schema-budget refusal (the guard threw *before* creating or writing), retry once. Do **not** dispatch `refresh_nodes`. Frontend-only types stay on the #2000 note.
- `panel_graph_outline`: once a schema-budget miss has been observed on this tab, do not keep advertising `mutations_ready: true`. Socket-up is not schema-ready.
- `panel_refresh_nodes`: `refreshed:false, reason:object_info_fetch_failed` now says the backend is not down and that this command dropped the last-known schema.

Fixes #2007

### DIFF
--- a/src/__tests__/orchestrator/object-info-budget-mutations.test.ts
+++ b/src/__tests__/orchestrator/object-info-budget-mutations.test.ts
@@ -366,6 +366,29 @@ describe("panel_graph_outline does not keep advertising mutations_ready after a 
     expect(parseJson(await outline.handler({} as never, ctx)).mutations_ready).toBe(true);
   });
 
+  it("a frontend-only add does not prove backend schema readiness", async () => {
+    const { b } = bridge(async (cmd) => {
+      if (cmd.cmd === "graph_outline") return outlineOk;
+      if (cmd.cmd === "graph_add_node" && cmd.class_type === "LoadImage") {
+        throw new Error(ADD_UNAVAILABLE);
+      }
+      if (cmd.cmd === "graph_add_node" && cmd.class_type === "MarkdownNote") {
+        return { ok: true, node_id: 30, type: "MarkdownNote" };
+      }
+      return { ok: true };
+    });
+    const ctx = makePanelToolCtx(b, TAB, new WorkflowTargetStore());
+    const defs = buildPanelToolDefs();
+    const add = defs.find((d) => d.name === "panel_add_node");
+    const outline = defs.find((d) => d.name === "panel_graph_outline");
+    if (!add || !outline) throw new Error("tools missing");
+
+    await add.handler({ class_type: "LoadImage" } as never, ctx);
+    await add.handler({ class_type: "MarkdownNote" } as never, ctx);
+
+    expect(parseJson(await outline.handler({} as never, ctx)).mutations_ready).toBe(false);
+  });
+
   it("keys readiness to the live tab, not a stable scope route", async () => {
     const scope = "orchestrator::test";
     let liveTab = "panel-a";
--- a/src/__tests__/orchestrator/object-info-budget-mutations.test.ts
+++ b/src/__tests__/orchestrator/object-info-budget-mutations.test.ts
@@ -1,0 +1,308 @@
+// #2007 — panel mutations blocked by fixed /object_info timeouts on a healthy
+// large install.
+//
+// Reported: panel_graph_outline succeeds with backend_socket=up and
+// mutations_ready=true, then panel_add_node("LoadImage"), panel_refresh_nodes,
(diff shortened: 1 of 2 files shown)


artokun:
Recovery update: the stale PR was rebased onto current main/release head b4e9b83d and force-pushed to exact head 40664b1. I preserved the merged #2004 starvation behavior by narrowing that matcher to its explicit two-route wording; the newer #2007 per-route share timeout now retries schema-guarded add/widget mutations exactly once and keeps mutations_ready honest. Focused tests pass 23/23 across object-info-budget-mutations and object-info-starvation; npm run build, npm run lint, and git diff --check pass. The full Vitest run was not used as a gate because this repository emits an extremely large fixture log and the run was interrupted before a terminal result. Please obtain an independent adversarial review before merge.

artokun:
Independent Codex gate: NO-SHIP. Findings: (1) panel-tools.ts:15098 clears schema-blocked state for any non-error refresh reply, including register_failed/app_unavailable; (2) panel-tools.ts:14165 returns before recording known starvation refusal for panel_set_widget; (3) panel-tools.ts:4417 keys readiness by stable ctx.tabId instead of current connection, allowing a reassigned healthy tab to remain falsely blocked. Focused gate execution also hit spawn EPERM. I’m recovering wt-2007 and will address all three before re-gating.

artokun:
Recovery update: fixed all three independent-gate P1s in f073af5. Schema readiness now follows the live tab connection/incarnation (including scoped sessions), panel_set_widget records starvation refusals before returning, and panel_refresh_nodes clears readiness only after an authoritative refreshed:true reply. Added regressions; focused object-info suites 26/26, full panel-tools 369/369, build, and lint pass. Fresh CI and a new independent gate are pending.

artokun:
Recovery update: the second re-gate found one remaining false-success path: a frontend-only MarkdownNote add can succeed without /object_info and must not clear a backend schema refusal. Fixed in 5e24e20 by limiting readiness clearing/blocking to schema-guarded adds, with the LoadImage-refusal then MarkdownNote-success regression. Focused suites 27/27, build, and lint pass. Fresh CI and re-gate are pending.

artokun:
Independent Codex gate r3: SHIP (no P0/P1 correctness or safety defect found). Fresh CI run is fully green: Ubuntu, Windows, macOS, and pack/install smoke all pass at 5e24e20. This lane is cleared for merge.

PR #2033, opened 2026-08-21
fix(1973): refuse opaque root graph completion

artokun:


artokun:
Follow-up hardening for merged #1973: same-tab root workflow switches can produce identical {scope:"root"} replies, so the orchestrator must not retire workflow A leftovers from workflow B without a workflow UUID/path. Exact head 6dfcc5d. Focused Vitest 32/32, npm.cmd run lint, git diff --check passed; independent gate is running.

artokun:
Updated exact head to d7b3dd4 after the gate found the previous opaque-root guard was incompatible with normal panel replies and still left same-shaped subgraphs ambiguous. The orchestrator now requires a per-workflow workflow_uuid on both graph replies, while panel follow-up support is being added in the paired panel release. Focused Vitest 32/32 and npm.cmd run lint pass. Please gate exact head d7b3dd4.

artokun:
Maintainer recovery update: rebased the intended #1973 follow-up onto current origin/main without carrying the stale branch delta. Exact PR head is 6b88644 (commits d0f57d7, 37f71e1, 6b88644). Validation on that exact head: npm.cmd run lint PASS; npm.cmd exec vitest run src/__tests__/orchestrator/get-errors-call-budget.test.ts PASS (33/33); git diff --check PASS. Independent gate and CI are still required before merge.

artokun:
Independent maintainer gate at exact HEAD 6b88644: SHIP. Reviewed origin/main...HEAD; scope is limited to get-errors audit identity helper, panel_get_errors caller, and focused tests. Verified post-call route capture, matching workflow_uuid/path fail-closed identity checks, and mutation only after route and graph identity checks. Validation: git diff --check PASS; tsc --noEmit PASS; focused Vitest attempted but blocked by environment spawn EPERM during Vite startup (no test failure observed). No introduced P0/P1 correctness or security issue found.

PR #2034, opened 2026-08-21
fix(2030): default the panel bridge to 9199 and never kill a foreign holder

### LINKED ISSUE #2030: Bridge port 9180 collides with Logitech G HUB, and comfyui-mcp's reclaim path offers to kill it; move the default to 9199 with a migration that never strands a live 9180 session across auto-update
### What happens

On a machine with Logitech G HUB installed (most gaming and creator desktops), `lghub_agent` is already listening on `127.0.0.1:9180` when comfyui-mcp starts. G HUB is a supervised autostart service: kill the agent and the updater respawns it within a second. comfyui-mcp's dedicated bridge port is 9180. So:

1. **The bind loop cannot win.** `UiBridge` retries `EADDRINUSE` five times over ~6 s (`src/services/ui-bridge.ts:1574-1577, 2247-2288`) and has no fallback port, no free-port search. Against a holder that never releases, every attempt is spent before the first user-visible message.
2. **Then it offers to kill the holder, and calls it ours.** `tryReclaimBridgePort` (`src/orchestrator/index.ts:771-868`) finds the pid with `netstat`/`lsof` and prompts:

   > port 9180 is already held by an unidentified process ("lghub_agent.exe" — no comfyui-mcp lockfile; likely an orphaned session or another app) — pid NNNN. Stop it (and its whole process tree) and take over port 9180? [y/N]

   The wording presumes an orphaned comfyui-mcp. On `y` it runs `taskkill /PID n /T /F` (`:628-641, :827-848`) on the user's mouse and keyboard software, waits for the respawn, and the straggler sweep at `:856-863` force-kills **whatever pid now holds 9180, with no name or lockfile check at all**. G HUB respawns again; `bridge.start()` at `:866` loses the race; the user has a dead bridge and a G HUB that just restarted twice. Non-TTY launches skip the prompt and print the same `taskkill` one-liner for the user to paste (`portKillHint`, `:646-652`). The 90 s startup deadline message (`:1074`) also says "if that is an older comfyui-mcp, stop it".

   The guards for exactly this already exist and are not wired in: `src/services/listener-ownership.ts` (branded `ours | not-ours | unconfirmed` and `supervised | abandoned | unconfirmed`, added for #814 when the ComfyUI restart path killed a supervised Desktop server) is used only on the ComfyUI restart path, and `probePanelOrchestrator` (`src/services/panel-launcher.ts:512-570`, a real WS upgrade + `hello` + `models`/`session_epoch` frame within 1200 ms) is used only by the launcher broker. The reclaim path imports neither.
3. **The panel reports G HUB as a running orchestrator.** The pack's `_port_in_use` is a bare TCP connect (`comfyui-mcp-panel/__init__.py:295-306`, docstring: "i.e. an orchestrator ... already owns the bridge"), so `/status`, `/backends`, and `/connect` return `running: true` and "orchestrator already running — connecting". The browser side then dials `ws://127.0.0.1:9180`, G HUB accepts or rejects the upgrade, and the user sits through a 20 to 45 s handshake timeout per attempt (`comfyui-mcp-panel.js:11431-11467`) before "Something else may be holding the port."

Nothing in either repo mentions Logitech, LGHUB, or G HUB; `gh issue list --search` for all three returns nothing, so every user who hit this so far worked around it alone with `COMFYUI_MCP_BRIDGE_PORT`. The port was chosen to dodge an internal collision with the legacy `--channels` bridge on 9101 (`comfyui-mcp-panel/CHANGELOG.md:1440-1442`); no external survey was done then. Doing one now: 9182, the phone-pairing port, is also the Prometheus `windows_exporter` default and is the one port in the block bound on `0.0.0.0` (`orchestrator/index.ts:1226, 1263`).

### The dangerous part is the update, not the port

Auto-update is default-on and hourly and restarts the orchestrator (#875, #1963). If a release changes the default to 9199 and that restart adopts it, a user who is on 9180 today and perfectly happy (no G HUB, or a pinned override) comes back to an orchestrator on 9199 and a panel still dialing 9180. Their session is gone, the panel says "no agent is listening", and nothing told them why. That is worse than the collision. So the rule for the change is:

- **A session that is working on 9180 keeps working on 9180 across a self-restart.** Auto-update, `panel_restart_comfyui`-driven relaunches, crash respawns: the relaunched child inherits the parent's *effective* bridge port explicitly (`COMFYUI_MCP_BRIDGE_PORT=<port it was bound to>` in the child env), never the compiled default. Today the three re-derivations at `orchestrator/index.ts:1104, 1209, 1488` and `boot.ts:528` each compute `env || 9180` independently, so a new default silently applies to the child.
- **A fresh user launch adopts the new default.** No inherited env, no pin: bind 9199.
- **The panel follows the orchestrator, not a guess.** The orchestrator already POSTs its tunnel URL to the pack (`advertise_bridge`, `__init__.py:517-543`); the same call carries the local `ws://127.0.0.1:<port>` so `/bridge_url` is authoritative. Until a panel that understands that is installed, the panel's default list is `[9199, 9180]` with the handshake as the arbiter, using the `LEGACY_BRIDGE_URL` migration pattern that already exists for 9101 (`comfyui-mcp-panel.js:1152-1153`): a saved setting of exactly `ws://127.0.0.1:9180` is treated as "default", not as a pin.
- **A user who set `COMFYUI_MCP_BRIDGE_PORT=9180` by hand is a pin** and is never moved.

### What correct looks like

1. **Never kill a process we cannot prove is ours.** `tryReclaimBridgePort` runs `probePanelOrchestrator` against the port first. If it answers the panel protocol and the lockfile is ours or stale: today's flow. If it does not answer the protocol: no prompt, no `taskkill`, no straggler sweep. Print the process name, say it is not comfyui-mcp, and say what to do (the override, or the new default if this is a pre-migration build). The same classification feeds `portKillHint` and the startup-deadline text, which today both instruct the user to kill the holder.
2. **New default block, counting down from 9199.** Bridge **9199**, panel HTTP-MCP 9198, phone pairing 9197, console 9196, tunnel listener 9195. Counting *up* from 9199 would put the panel MCP on 9200 (Elasticsearch, and IANA `wap-wsp`), which is a real developer-desktop collision; 9195 to 9199 are IANA-unassigned and the only claimants are niche Prometheus exporters (`igd`, `mtail`, `logstash`, `cloudflare`, `maxscale`) that do not run on desktops. 9199 keeps the nod to 8188. One constant, `DEFAULT_PANEL_BRIDGE_PORT` (`panel-launcher.ts:29`), replaces the four bare `9180` literals, and the `+1..+4` arithmetic at `orchestrator/index.ts:1226, 2205, 2223, 6596` becomes `-1..-4` through one helper so a test can pin the map.
3. **Migration that cannot strand anyone** (the rules above), with a test that starts an orchestrator on 9180, simulates the update restart, and asserts the child is still on 9180 and the panel still connects; and a second test that asserts a fresh launch lands on 9199.
4. **The pack stops inferring identity from a TCP connect.** `_port_in_use` becomes a protocol probe (the same `hello` → `models` exchange), so `/status` says `running: false, port_held_by: "lghub_agent.exe"` instead of `running: true`.
5. **Docs and contracts.** `docs/third-party-hosts.mdx` (12 locales) publishes "pairing port = bridge + 2, default 9182" as an integration contract; it becomes "bridge − 2, default 9197" with a note that 9180-era URLs still pair. `configuration.mdx`, `troubleshooting.mdx` (its heading quotes `ws://127.0.0.1:9180`), `cloud-deployment.mdx`, `panel.mdx`, `installation.mdx`, `README.md:204`, `.env.example:36`, `docker/runpod/README.md:621`, the panel README and settings tooltip, and the mobile app's pairing hint (`pairing_screen.dart:87`). The release note names Logitech G HUB so the next person searching finds it.

### Tests that pin today's value

`src/__tests__/orchestrator/startup-deadline.test.ts:36` (`const PORT = 9180`), `panel-console-http.test.ts:52` (`bridge_port` 9180), `agent-identity.test.ts:60` (identity filename embeds the port), `config.test.ts:477-511` (`local-target-9180.json`), `i18n/terminal-layout.test.ts:82-98` and `panel-console-i18n.test.ts:184` (banner string), the 9181 panel-MCP URL in `antigravity/gemini/grok/qwen-backend.test.ts` and `transport/cli.test.ts`; panel `command-liveness.test.mjs` (6); mobile `pairing_url_test.dart` (5). None of them test the port map symbolically or the reclaim path against a foreign holder; the fix adds both.

### Environment

comfyui-mcp 0.52.41, comfyui-mcp-panel 0.15.31. Windows with Logitech G HUB; reproduced as "bridge never binds, prompt offers to stop lghub_agent.exe, G HUB restarts".

**Two independent user reports so far**, one from a YouTuber setting up on camera and one from a Discord user, both on Windows with G HUB installed, both hitting the bind failure on first run. Neither found the `COMFYUI_MCP_BRIDGE_PORT` override on their own; both needed the maintainer to diagnose it. That is the friction cost of a default that loses to an autostart service: it lands on the first-run experience, for the users most likely to be showing it to others. G HUB's habit of sitting on and hopping 9xxx ports is documented elsewhere (rancher-desktop #2585: it took 9080, then 9081 when that was busy).

Related: #814 (killed a supervised process we cannot relaunch, the same shape on the ComfyUI port) · #1524 (origin of the startup deadline) · #875 and #1963 (auto-update restarts and what they must not break) · comfyui-mcp-panel#1596 (companion: the pack-side probe, advertised port, and 9180 legacy migration).

artokun:
Fixes #2030

Companion: [comfyui-mcp-panel#1596](https://github.com/artokun/comfyui-mcp-panel/issues/1596) (pack-side protocol probe, advertised port, 9180 legacy list). This PR is the orchestrator half only.

## User-visible change

On a machine with Logitech G HUB, `lghub_agent` already owns `127.0.0.1:9180`. The panel bridge used to retry that port five times, then offer to `taskkill /T /F` the holder as an "orphaned session". Saying yes killed G HUB; the straggler sweep then killed whatever respawned onto 9180.

Fresh launches now bind **9199** (panel HTTP-MCP 9198, pairing 9197, console 9196, tunnel 9195 — counting *down* so 9200/Elasticsearch is never in the block). Override with `COMFYUI_MCP_BRIDGE_PORT`.

## Migration rule

- A session already working on 9180 stays on 9180 across auto-update / self-restart / crash respawn that inherits env. The child gets `COMFYUI_MCP_BRIDGE_PORT=<port it was bound to>`, never the compiled default. An unpinned self-restart from a pre-pin build also stays on 9180 so the bind-retry still rides the parent's release.
- A hand-set `COMFYUI_MCP_BRIDGE_PORT=9180` is a pin and is never moved.
- 9180-era pairing on 9182 still answers (`panelPortBlock(9180)` keeps +1..+4).
- The orchestrator advertises `ws://127.0.0.1:<bound-port>` as `local_url` on the existing `advertise_bridge` POST so the panel can follow.

## Reclaim rule

`tryReclaimBridgePort` runs `probePanelOrchestrator` first. Classification is minted in `classifyBridgeHolder` (`listener-ownership.ts`: `ours | not-ours`).

- Protocol speaker + lockfile ours or stale → today's prompt/kill flow.
- TCP open but protocol probe fails → **no prompt, no taskkill, no straggler sweep**. The message names the process as not comfyui-mcp, names Logitech G HUB as a known occupant of 9180, and points at `COMFYUI_MCP_BRIDGE_PORT` / default 9199.
- Same classification feeds `portKillHint` (ours only) and the startup-deadline text.

## Test evidence

Shipped functions, not a second map:

- `panelPortBlock(DEFAULT_PANEL_BRIDGE_PORT)` pins 9199/9198/9197/9196/9195; `panelPortBlock(9180)` keeps 9180/9181/9182/9183/9184.
- `resolveBridgePort({})` is 9199; pin 9180 stays 9180; unpinned `COMFYUI_MCP_RESTART_GEN>0` stays 9180; `selfRestartChildEnv` inherits the bound port.
- Foreign holder: real TCP listener that refuses the WebSocket upgrade → `probePanelOrchestrator` fails → `tryReclaimBridgePort` does not prompt or kill; log names the process as not ours.
- `advertiseBridge` body includes `local_url`.

`npx tsc --noEmit` clean. Targeted vitest (`--maxWorkers=4`): 120 + 109 tests passed (port map, reclaim, startup-deadline, advertise, console, identity, backends, launcher, self-restart).


artokun:
Ready for an independent review before merge. Orchestrator half of #2030 (default bridge 9199, never kill a foreign holder such as Logitech G HUB, live 9180 sessions inherit the bound port). Panel companion is artokun/comfyui-mcp-panel#1596.

src/services/bridge-port-reclaim.ts
```diff
@@ -0,0 +1,414 @@
...
+      supervision: unclassifiedSupervision(),
+      reclaimable: false,
+      pid: null,
+      portPid,
+      lockPid,
+      processName,
+      lock,
+      speaksPanelProtocol,
+    };
+  }
```
artokun: P1: the probe already ran, `classified` is already `not-ours` when `speaksPanelProtocol` is false, and then this branch discards it for `unclassifiedOwnership()`.

Issue #2030: a probe that fails closed as `unconfirmed` must not still offer kill. `runPanelOrchestrator` prints `portKillHint` (`taskkill /T /F`) whenever `reclaim.ownership !== "not-ours"`. G HUB that accepts TCP but whose pid `netstat`/`lsof` misses (timeout, locale, no `lsof`) therefore still gets the original kill one-liner.

Return `...classified` here (or at least `not-ours` when the probe failed). A missing pid is not a reason to un-classify a protocol miss. The foreign-holder test never hits this branch because a real listener is pid-mappable in CI, so this hole is untested.

src/orchestrator/index.ts
```diff
@@ -1317,14 +1069,20 @@ export async function runPanelOrchestrator(): Promise<void> {
...
-    );
+    reclaim = await tryReclaimBridgePort(bridge, lockPort, lockPath);
+    bound = reclaim.bound;
+  }
+  if (!bound) {
+    if (reclaim?.ownership !== "not-ours") {
+      logger.error(
+        `[panel-orchestrator] could not bind the panel bridge port — another process owns it. Free that port and restart the orchestrator. Override the port with COMFYUI_MCP_BRIDGE_PORT.\n${portKillHint(lockPort)}`,
+      );
+    }
```
artokun: P1: this is the consumer of the `unconfirmed` hole. `portKillHint` is still the copy-paste `taskkill /PID … /T /F` / `kill -9` text the issue asked to stop feeding users for a foreign holder.

The gate is `!== "not-ours"`, so `unconfirmed` (pid lookup failed after a failed protocol probe) still prints the kill recipe. Non-TTY was the original path that skipped the prompt and handed the user that one-liner — same shape, still here.

Only print `portKillHint` for `ownership === "ours"`. A foreign or unconfirmed holder should get `foreignBridgeHolderAdvice`, never a kill command.

src/orchestrator/index.ts
```diff
@@ -1065,10 +809,17 @@ export class DownloadProgressSnapshots {
...
+    exit?: (code: number) => never;
+    incumbent?: (p: number) => number | undefined;
+    holderAdvice?: (pid: number) => string;
+  } = {},
 ): () => void {
   const exit = deps.exit ?? ((code: number) => process.exit(code));
   const findIncumbent = deps.incumbent ?? pidListeningOnPort;
+  const holderAdvice =
+    deps.holderAdvice ??
+    ((pid: number) => startupDeadlineHolderAdvice({ port, pid }));
```
artokun: P1: production never classifies. `runPanelOrchestrator` calls `armStartupDeadline(lockPort)` with no `holderAdvice`, so this default always hits `startupDeadlineHolderAdvice({ port, pid })` with `ownership` undefined.

That unclassified string still leads with "If that is an older comfyui-mcp, stop it" (`bridge-port-reclaim.ts:161-166`). Naming G HUB in the second sentence does not stop the first sentence from being a kill instruction.

The new test at `startup-deadline.test.ts:99` injects a pre-classified `holderAdvice` and does not execute this default. The older test at line 91 still asserts `/stop it/` against this path, so the wording the issue wanted gone is locked in.

Classify in this callback (probe + `processNameOf` are already imported) so a foreign incumbent cannot be described as "stop it".

src/__tests__/orchestrator/startup-deadline.test.ts
```diff
@@ -93,6 +96,31 @@ describe("the startup deadline (#1524)", () => {
...
+    });
+    vi.advanceTimersByTime(91_000);
+
+    const msg = errors.join("");
+    expect(msg).toMatch(/pid 3807/);
+    expect(msg).toContain("lghub_agent.exe");
+    expect(msg).toMatch(/not comfyui-mcp/);
+    expect(msg).toMatch(/Leave it running/);
+    expect(msg).not.toMatch(/if that is an older comfyui-mcp, stop it/);
+  });
```
artokun: This test does not drive production. It injects `holderAdvice` that already ran `classifyBridgeHolder({ speaksPanelProtocol: false })`. `runPanelOrchestrator` never passes `holderAdvice`, so this green result says nothing about the 90s deadline a G HUB user would actually read.

Drop the injection and assert against the default `armStartupDeadline(PORT, { exit, incumbent })` path — or have that default classify. As written, deleting the production classification (there isn't one) does not go red, and the sibling test still pins `/stop it/`.

src/__tests__/services/bridge-ports.test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+describe("pinBoundBridgePort + selfRestartChildEnv (#2030)", () => {
+  it("a child restart inherits the parent's effective 9180", () => {
+    const parent = pinBoundBridgePort(9180, { PATH: "/usr/bin" });
+    expect(resolveBridgePort(parent)).toBe(9180);
+    const child = selfRestartChildEnv(parent);
+    expect(child.COMFYUI_MCP_BRIDGE_PORT).toBe("9180");
+    expect(resolveBridgePort(child)).toBe(9180);
+    expect(Number(child.COMFYUI_MCP_RESTART_GEN)).toBeGreaterThan(0);
+  });
```
artokun, line 99: Claim vs test: this only stubs env. `pinBoundBridgePort(9180, { PATH })` then `selfRestartChildEnv(parent)` never starts an orchestrator, never binds 9180, never goes through `defaultSpawnReplacement`.

Issue #2030 asked for a test that starts on 9180, simulates the update restart, and asserts the child is still 9180. `selfRestartChildEnv`'s default `boundPort` is `resolveBridgePort(parentEnv)`, and `defaultSpawnReplacement` calls `selfRestartChildEnv()` with no port — it only stays on 9180 if `runPanelOrchestrator`'s `pinBoundBridgePort(lockPort)` already mutated `process.env`. Deleting that pin does not fail this suite.

Either spawn (or at least call `defaultSpawnReplacement` / `selfRestartChildEnv()` the way production does, from a parent whose env is *not* pre-pinned in the test) or stop claiming the restart-from-9180 case is covered.

artokun (commented):

The reproduced G HUB **auto-kill** (prompt → `taskkill /T /F` → straggler sweep of whoever respawned onto the port) is fixed **when `pidListeningOnPort` actually finds the holder**. `classifyBridgeHolder` + the `!reclaimable || ownership === "not-ours"` return in `tryReclaimBridgePort` run before `isTty` / prompt / kill, and the foreign-holder test drives that shipped function against a real TCP listener that refuses the WS upgrade. Fresh 9199, hand-pin 9180, countdown port map, and `advertise_bridge.local_url` also look right.

Three P1 holes in the claims:

1. **Probe fail-closed as `unconfirmed` still offers kill.** `assessBridgeHolder` probes, classifies `not-ours`, then throws that verdict away when no pid is mapped (`unclassifiedOwnership()`). `runPanelOrchestrator` then prints `portKillHint` (`taskkill /T /F` / `kill -9`) for any ownership that is not a definite `not-ours`. That is exactly the issue's "probe fails closed as unconfirmed and still offers kill" bug — the G HUB path whenever netstat/lsof misses.
2. **Production startup-deadline text is still "stop it".** `armStartupDeadline(lockPort)` never classifies. The default `holderAdvice` calls `startupDeadlineHolderAdvice({ port, pid })` with no ownership/processName, which still leads with "If that is an older comfyui-mcp, stop it". The new test injects a pre-classified `holderAdvice`; the older test pins `/stop it/` on the default path. Deleting classification from production would not go red.
3. **The 9180 migration test only stubs env.** `pinBoundBridgePort(9180, { PATH })` + `selfRestartChildEnv(parent)` does not start an orchestrator, does not bind 9180, and does not spawn a child. `defaultSpawnReplacement` calls `selfRestartChildEnv()` with no bound port. The pin-at-start wiring looks correct; the claimed restart-from-9180 test does not exist.

Fix (1) by returning the probe classification even when pid lookup fails (a protocol miss is already enough to refuse kill text). Fix (2) by classifying in the production deadline callback, not only in an injected test helper. Fix (3) by driving spawn env from a bound 9180 parent — or drop the claim.

artokun:
Addressed the three review findings on 78214bb:

1. **Unconfirmed pid no longer offers taskkill.** `assessBridgeHolder` keeps the protocol classification even when `pidListeningOnPort` returns null (netstat/lsof miss). Bind-fail text goes through `bindFailureAdvice`: `taskkill` only for a definite `ours`. Foreign and unconfirmed name the situation and point at `COMFYUI_MCP_BRIDGE_PORT` / 9199. Test: probe false + pid null → ownership `not-ours`, no prompt, no kill, advice has no `taskkill`. Restoring the `if (!pid) unclassified` early return goes red.

2. **Startup deadline classifies before it speaks.** Default `holderAdvice` runs `assessHolder` and passes `ownership` into `startupDeadlineHolderAdvice`. Foreign: no "stop it". Ours: still "stop it". The foreign test injects `assessHolder`, not a stub `holderAdvice`, so deleting classification from the default path goes red.

3. **Spawn replacement takes the bound port.** `defaultSpawnReplacement` calls `selfRestartChildEnv(parentEnv, boundPort)`. Test drives that helper with `boundPort=9180` against an unpinned parent env (compiled default 9199) and asserts the child env is 9180. Dropping the `boundPort` argument goes red.

Ready for an independent re-review. Not merging.

artokun (commented):
COMMENT INTENDED: APPROVE

Follow-up on 78214bbc. The three P1s from the last review are fixed on 11bc1ae, and the new tests go red if those branches are deleted.

1. **Probe fail + pid null no longer offers taskkill.** `assessBridgeHolder` always returns `classifyBridgeHolder(...)` — the `if (!pid) unclassifiedOwnership()` discard is gone. Bind-fail logs `bindFailureAdvice`, which prints `portKillHint` only for `ownership === "ours"`. Test: `probe fail + no pid: still not-ours; bind-fail advice has no taskkill` plus the ours/foreign/unconfirmed `bindFailureAdvice` cases.

2. **Default `armStartupDeadline` classifies.** Production still calls `armStartupDeadline(lockPort)` with no stub. Default `assessHolder` → `assessBridgeHolder`; default `holderAdvice` passes `ownership` into `startupDeadlineHolderAdvice`. Unclassified text no longer says "stop it"; that string is only the `ours` branch. The foreign-holder test drives the default `holderAdvice` (injects `assessHolder`, not `holderAdvice`). Source-pinned: `assessHolder`, `ownership: a.ownership`, `armStartupDeadline(lockPort)`.

3. **Spawn helper inherits bound 9180.** `defaultSpawnReplacement` calls `selfRestartChildEnv(parentEnv, boundPort)` with `boundPort` as a distinct argument. Test drives that shipped function with an unpinned parent env + `boundPort: 9180` and a fake `spawn`; dropping the second argument resolves to 9199 and goes red.

Targeted vitest on this head: 28 passed (bridge-port-reclaim, bridge-ports, startup-deadline).

No new P1. Same GitHub user — cannot submit `APPROVE`, so this is COMMENT as required.

artokun:
CI `check:docs-locale` failed because fenced code blocks in translated `third-party-hosts.mdx` must be byte-identical to English. The bulk port rewrite had changed locale copy-paste samples (block #2, the JS reference client) while English still had `:9182`, and the LLM-prompt fence wrapping had drifted.

Fix: English reference client now uses `:9197`, and every locale fence was copied from English. `npm run check:docs-locale` and `npm run check:docs-links` both pass.

artokun (commented):
COMMENT INTENDED: APPROVE

Head `403984c50caa78a6d6b2c7f7734b3919307b4746`. Previous APPROVE was on `11bc1ae` (the three P1s: no taskkill on unconfirmed/foreign holders, classified startup-deadline text, bound-port inherit). This SHA is the `check:docs-locale` fix: locale `third-party-hosts` code block #2 is now identical to English, and the English pairing example is `9197` (bridge − 2). No behavior change in reclaim/restart/bind.

Same GitHub user, so this is COMMENT not an Approve event. Merge when Build is green. Ignore Mintlify/Copilot.

Issue #2021, opened 2026-08-21
Run-completion watchdog forgets a completion drained before panel_run opens its ticket — the #1789 fallback can never fire for a fast render


artokun:
Found while working #2001; filing separately rather than folding it into that PR (#2020), because it is a different defect with a different failure.

### What happens

`src/orchestrator/run-completion-watchdog.ts`, `observe()`:

```ts
const promptId = typeof ev?.promptId === "string" ? ev.promptId.trim() : "";
if (!promptId) continue;
// Only runs THIS session queued and is still owed a completion for.
if (!awaiting(promptId)) continue;   // <-- the observation is DISCARDED, not deferred
arm(promptId, ev.status, now());
```

`QueueMonitor.drainCompletions()` splices, and the queue-status broadcaster tees it into `observe()` exactly once (`src/orchestrator/index.ts`, the 1 Hz `queueStatusTimer`). So a completion is offered to the watchdog **one time**. If `awaiting(promptId)` is falsy at that instant the observation is gone forever, and the #1789 fallback can never synthesise for that run no matter what happens afterwards.

### Why the ticket can be missing at that instant

A ticket cannot be opened before dispatch, because ComfyUI mints the prompt id when it accepts the run: `panel_run` awaits the panel's `graph_run` reply and only then calls `RunCompletions.openRun(...)` (`src/orchestrator/panel-tools.ts`, ~13054). The relay window for that reply is bounded in seconds (#1565), so a render that finishes faster than the reply travels back is observed by QueueMonitor *before* the ticket exists.

That ordering is not hypothetical — it is already recorded in this codebase. `src/orchestrator/run-completion-journal.ts` (#1327) says of the same race:

> A ticket cannot be opened before dispatch, because ComfyUI mints the prompt id when it accepts the run. So a render that finishes faster than the reply travels back — the reporter's was 0.1s, fully cached — lands here BEFORE openRun and correlates against no ticket at all.

The journal grew `claimRaced` / `dispatchedAt` to handle it for the **panel's** frame. The watchdog got no equivalent for **QueueMonitor's** observation.

### The failure

1. `panel_run` dispatches; the render is cached/tiny and finishes in well under a second.
2. QueueMonitor drains the completion on its next 1 Hz tick; `awaiting()` is false (no ticket yet); the observation is discarded.
3. The panel's reply arrives; `openRun` opens the ticket.
[...]

PR #2035, opened 2026-08-21
fix: retain fast render completions before ticketing (#2021)

artokun:
## Summary\n- Preserve fast render completions that arrive before the completion watchdog tickets them.\n- Add focused regression coverage.\n\nRecovered eligible P2 issue #2021; implementation agent committed and pushed this branch. Please review against current main.

### DIFF
--- a/src/__tests__/orchestrator/panel-tools.test.ts
+++ b/src/__tests__/orchestrator/panel-tools.test.ts
@@ -6642,6 +6642,41 @@ describe("panel-tools: panel_run scoped-run stamp-race retry (#772)", () => {
       }
     });
 
+    it("refreshes the watchdog dispatch fence for the scoped retry (#2021)", async () => {
+      __panelToolsTestHooks.setRetrySettleMs(0);
+      const promptId = "p-2021-retry-fence";
+      let dispatches = 0;
+      let retryStarted = false;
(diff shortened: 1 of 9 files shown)


artokun:
Maintainer recovery update: exact rebased head is 509369a (onto current origin/main). This includes the CI invariant repair and the #2021 retry-fence fix: the scoped re-dispatch now refreshes dispatchedAt before the second call, with a production panel_run regression asserting the ticket uses the second dispatch boundary. Validation on the pre-rebase equivalent passed: npm.cmd run lint; full src/__tests__/orchestrator/panel-tools.test.ts 369/369; src/__tests__/orchestrator/run-completion-race.test.ts 12/12; git diff --check. Independent gate and fresh CI are required; prior NO-SHIP P1 is not considered cleared until those pass.

artokun:
Independent Codex gate SHIP for exact HEAD 509369a6ba4b198d5231f19bbd7600a91bc15cba. Verified retry dispatch fence refresh, pre-retry stale-arm rejection, journal fencing, Claude/HTTP markTicketed wiring, normal fast-completion attribution, and burst protection. git diff --check and lint passed; focused Vitest/build were environment-blocked in the read-only gate (spawn EPERM/read-only dist), while the writable recovery worktree passed the focused suites. No introduced P0/P1 correctness or security defect found. Merging remains gated on fresh CI.

PR #2040, opened 2026-08-21
fix(2031): node_pack scaffolds into the scanned custom_nodes path on split installs

### LINKED ISSUE #2031: node_pack scaffolds into unscanned custom_nodes path on split install
node_pack(action:"scaffold") reported that it created a custom node under the resolved split-install workspace custom_nodes directory, but after panel_restart_comfyui the class type was absent from /object_info. The running ComfyUI actually loaded custom nodes from the code checkout's custom_nodes directory. Copying the same syntax-checked pack there is the working workaround.

### Steps to reproduce
1. On a split install where workspace_path and code_path differ, call node_pack scaffold.
2. Observe the returned path under workspace_path/custom_nodes.
3. Restart ComfyUI and call node_pack verify with restart:false.
4. The scaffolded class is missing.

### Exact error (scrubbed)
node_pack verify: "1 of 1 node type(s) are NOT registered: H3IndexedTextSlotsNode."
Environment reports a split install with workspace_path ~/ComfyUI and code_path ~/ComfyUI/ComfyUI.

### Fix
Applied locally: workaround only. Upstream-only: yes.
The node_pack target resolver should use the custom-node scan directory actually used by the running server. For split installs, prove the target from the server's active custom-node search paths rather than assuming workspace_path/custom_nodes. The same resolved path should be used by scaffold, write/read/patch, and verify.

### Environment
Windows 11; ComfyUI 0.33.0; frontend 1.49.6; RTX 4090 22 GB; comfyui-mcp 0.52.45; panel 0.15.32.

artokun:
Fixes #2031

## What the user now observes

On a split install (`workspace_path` ≠ `code_path`) **without** `--base-directory`, `node_pack` (`action:"scaffold"`) writes the pack under the live `main.py` checkout's `custom_nodes/` — the directory ComfyUI actually scans. After `panel_restart_comfyui`, `node_pack` (`action:"verify"`, `restart:false`) finds the class in `/object_info`. Desktop `--base-directory` installs are unchanged (#1770).

## Cause

`node_pack` threaded `resolveEffectiveComfyUIBaseLive()`, which prefers `COMFYUI_PATH` / `workspace_path` whenever `--base-directory` is absent. That is correct for models and output. ComfyUI's `folder_paths.py` only points `custom_nodes` at the data workspace when `--base-directory` is set; otherwise it scans `<main.py dir>/custom_nodes`. The portable split (`~/ComfyUI` data + `~/ComfyUI/ComfyUI` checkout) therefore scaffolded into an unscanned tree.

## Fix

`resolveCustomNodesScanBaseLive()` matches `folder_paths.base_path`:

1. Live `--base-directory` when it exists on disk (Desktop, #1770)
2. Else the live `main.py` checkout (#2031)
3. Fail closed on an unresolvable relative `--base-directory`
4. Else the configured data base (offline / unreachable)

`node_pack` scaffold / verify / write / read / patch / git / publish-by-name all thread this same root.

## Tests

- Split-install scaffold lands in `code/custom_nodes`, not `workspace/custom_nodes`
- `--base-directory` still wins (#1770)
- Handler threads the scan-root resolver to every file-touching action


artokun:
Recovery update: independent gate found a P1 in the split-install resolver: a present but unavailable absolute --base-directory fell through to the live main.py checkout, so node_pack could write custom_nodes where ComfyUI does not scan. Fixed in 3f6121e: any present unavailable base-directory now fails closed, with regression coverage; targeted workspace-env/node-pack tests (168/168), build, and lint pass. Waiting for fresh CI, then I will rerun the independent gate.

artokun:
Recovery update: the second independent gate found a downstream P1 after the first fix: node_pack treated the resolver undefined fail-closed result as permission to fall back to COMFYUI_PATH. Fixed in 800754d by adding a strict scan-root resolver that preserves an unavailable --base-directory as an actionable refusal, and routing scaffold/verify/publish-by-name/list/read/search/write/patch/git through it. Targeted workspace-env/node-pack/split-install tests 171/171, build, and lint pass. Fresh CI and re-gate are pending.

artokun:
Independent Codex gate r3: SHIP. Fresh CI is fully green at 800754d, including Ubuntu, Windows, macOS, and pack/install smoke. This lane is queued behind older eligible P2s for the ordered merge/release.

PR #2052, opened 2026-08-22
fix(2050): clear_vram reports VRAM after CUDA release settles

### LINKED ISSUE #2050: clear_vram reports stale VRAM immediately after unload
`clear_vram` returned a VRAM value immediately after model unload that differed from a follow-up raw system-stat read. The tool reads stats immediately after the free request, so its displayed value appears to race asynchronous CUDA/ComfyUI release completion.

### Steps to reproduce
1. Run a MiniMax H3 workflow that fills VRAM.
2. Call `clear_vram` with `{ unload_models: true, free_memory: true }`.
3. Compare its displayed VRAM with a subsequent `get_system_stats(action:"stats")` response.

### Exact result (scrubbed)
clear_vram: `Current VRAM: 1205/24564 MB free | Torch: 40/64 MB free`.
Follow-up stats: `vram_free: 23985913856`, `vram_total: 25756696576`.

### Fix
Wait briefly or poll `/system_stats` until VRAM stops changing before formatting the post-unload value, or label the immediate reading as transitional.

### Environment
Windows 11; ComfyUI 0.33.0; frontend 1.49.6; RTX 4090 24GB; PyTorch 2.11.0+cu130; comfyui-mcp 0.52.49; panel 0.14.14.

artokun:
Fixes #2050

`clear_vram` printed the first `/system_stats` sample after `POST /free`. `/free` returns when ComfyUI drops model refs; CUDA can still be releasing, so that sample lagged a follow-up `get_system_stats (action:"stats")` (reporter: 1205 MB free vs `vram_free: 23985913856`).

After a successful `/free`, the tool now polls `/system_stats` until device 0's VRAM counters stop changing (min 1s so an unchanged first reading is not treated as settled, cap 5s) and prints that value.

### DIFF
--- a/src/__tests__/tools/clear-vram-stale-report.test.ts
+++ b/src/__tests__/tools/clear-vram-stale-report.test.ts
@@ -206,4 +206,17 @@ describe("clear_vram post-unload VRAM (#2050)", () => {
     expect(textOf(res)).toContain("VRAM cleared successfully (models unloaded, memory freed).");
     expect(textOf(res)).not.toContain("Current VRAM:");
   });
+
+  it("omits a stale first sample when a later stats read fails", async () => {
+    mocks.comfyApiFetch.mockResolvedValue(freeOk());
+    mocks.getSystemStats
+      .mockResolvedValueOnce(STALE)
+      .mockRejectedValue(new Error("ETIMEDOUT"));
+
+    const res = await runClear({ unload_models: true, free_memory: true });
+
+    expect(textOf(res)).toContain("VRAM cleared successfully (models unloaded, memory freed).");
+    expect(textOf(res)).not.toContain("Current VRAM:");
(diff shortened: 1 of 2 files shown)


artokun:
Recovery update: independent gate found a P1 where a failed follow-up /system_stats returned the previous sample, allowing clear_vram to print known-stale VRAM. Fixed in 0a694d8: any failed settle poll now omits the stats block, with regression coverage. Targeted clear-vram tests 6/6, build, and lint pass. Fresh CI and re-gate are pending.

artokun:
Independent Codex gate r2: SHIP. Fresh CI is fully green at 0a694d8, including Ubuntu, Windows, macOS, and pack/install smoke. This lane is queued behind older eligible P2s for the ordered merge/release.

PR #2066, opened 2026-08-22
fix(1524): panel_* tools reappear after panel_restart_comfyui

### LINKED ISSUE #1524: panel_* tools never reconnect after a mid-session MCP drop (comfyui tools do); orphan second orchestrator left with no bound ports

artokun:
Fixes #1524

## What was already true

Bind-or-exit shipped in 0.51.35. #1596 made a stale HTTP session a 404. #1614 made a session that **starts** without an MCP server say so. #1681 met a mid-session drop with one bounded `reconnectMcpServer()` — **Claude lane only**. Codex was explicitly out of scope there.

Today's recurrence (0.52.47 / panel 0.15.32, Codex, Windows): after `panel_restart_comfyui({force:true})` and session resume, `ALL_TOOLS` had **no** `mcp__panel__*` entries while the generic comfyui gateway remained. `tools.mcp__panel__panel_set_todo` threw `TypeError: ... is not a function`. Current main (0.52.56) still has no Codex-side re-register.

## Root cause

Codex's MCP client does not auto-reconnect HTTP servers after a drop (openai/codex#11489). `comfyui` is **stdio** — the client owns the child and respawns it. `panel` is **http** — a stateful session against the orchestrator loopback MCP. After a ComfyUI restart that drops that session, Codex keeps the stdio half and never retries the HTTP half, so the next code-mode exec's `ALL_TOOLS` snapshot has no `mcp__panel__*` tools.

## What changes

`CodexBackend` gets the same turn-end watch #1681 added for Claude, on Codex's own control plane:

- **Detection** — one `mcpServerStatus/list` poll per completed turn, mapped through `reportedFromCodexMcpListing` into the existing `inspectMcpServers` comparison. A named server with `tools: []` is the reporter's ALL_TOOLS gap and counts as a drop.
- **Recovery** — one `config/mcpServer/reload` per down-episode. The app-server applies that refresh on the **next** turn, so this turn does not claim a verdict from an immediate re-read; the following poll is the verdict.
- **Honesty** — `needs-auth` / `disabled` are reported and not retried. A poll or reload that throws stays silent or unverified and never takes the turn stream down. A server that stays down is reported once, not fought.

## Scope

Codex lane only. Gemini/Grok/other HTTP-lane backends are the same transport family but have their own control planes; nothing here touches them. This does not explain **why** the panel HTTP session dropped — still open on the issue. It re-registers the tools after that drop, which is the remaining user-visible half.

## Verification

- `reportedFromCodexMcpListing` pins empty `tools: []` as the ALL_TOOLS gap, missing-from-listing as absent, omitted-tools as not a false alarm.
- `codex-mcp-reconnect.test.ts` drives `CodexBackend.run()` with a fake app-server: reload-fixes-drop narrated on the next turn; reload-fails reported once; healthy listing never reloads; poll throw silent; reload throw unverified; second drop is a new episode; needs-auth never reloaded; no poll when the session was given no MCP servers.
- Existing Codex code-mode host-retry and watchdog suites still green.

### DIFF
--- a/src/__tests__/orchestrator/codex-mcp-reconnect.test.ts
+++ b/src/__tests__/orchestrator/codex-mcp-reconnect.test.ts
@@ -1,0 +1,349 @@
+// #1524 — Codex lane. After panel_restart_comfyui the session resumes with the
+// generic comfyui gateway and NO mcp__panel__* entries in ALL_TOOLS.
+// tools.mcp__panel__panel_set_todo then throws TypeError: is not a function.
+//
+// Claude got reconnectMcpServer() in #1681. Codex's control plane is different:
+// mcpServerStatus/list to see the drop, config/mcpServer/reload to re-init
+// (applied on the NEXT turn). Codex does not auto-reconnect HTTP MCP
+// (openai/codex#11489); stdio comfyui is respawned by the client, panel is not.
+//
+// These tests drive the live CodexBackend.run() path with a fake app-server
+// client, same seam as the code-mode host-retry suite.
+
+import { afterAll, beforeAll, describe, expect, it, vi } from "vitest";
+import { waitFor } from "../helpers/wait-for.js";
+import type { AgentEvent } from "../../orchestrator/agent-backend.js";
+
+vi.mock("../../utils/logger.js", () => ({
+  logger: { debug: vi.fn(), info: vi.fn(), warn: vi.fn(), error: vi.fn() },
+}));
+
+type BackendModule = typeof import("../../orchestrator/codex-backend.js");
+type Backend = InstanceType<BackendModule["CodexBackend"]>;
+
+let CodexBackend: BackendModule["CodexBackend"];
+
+beforeAll(async () => {
+  vi.resetModules();
+  ({ CodexBackend } = await import("../../orchestrator/codex-backend.js"));
+});
+
+afterAll(() => vi.unstubAllEnvs());
+
+/** Live `mcpServerStatus/list` tools field is a name→schema map, not an array. */
+const CACHED_PANEL_TOOLS = { panel_graph_outline: {}, panel_set_todo: {} };
+const PANEL_UP = [
+  { name: "comfyui", runtimeStatus: "connected", tools: { list_tools: {} } },
+  { name: "panel", runtimeStatus: "connected", tools: CACHED_PANEL_TOOLS },
+];
+/** Dropped HTTP panel: runtimeStatus failed, tools inventory still cached. */
+const PANEL_GONE = [
+  { name: "comfyui", runtimeStatus: "connected", tools: { list_tools: {} } },
+  { name: "panel", runtimeStatus: "failed", tools: CACHED_PANEL_TOOLS },
+];
+const MCP_SERVERS = {
+  comfyui: { transport: "stdio" as const, command: "node", args: ["mcp.js"] },
+  panel: { transport: "http" as const, url: "http://127.0.0.1:9198/tab" },
+};
+
+const noticesOf = (events: AgentEvent[]) =>
+  events.filter(
+    (e): e is Extract<AgentEvent, { type: "error" }> =>
+      e.type === "error" && (e as { sessionNotice?: boolean }).sessionNotice === true,
+  );
+
+type Listing =
+  | Array<{
+      name: string;
+      runtimeStatus?: string;
+      authStatus?: string;
+      tools?: Record<string, unknown> | unknown[];
+    }>
+  | "throw";
+
+async function waitUntil(pred: () => boolean, timeoutMs = 500): Promise<void> {
+  const start = Date.now();
+  while (!pred()) {
+    if (Date.now() - start > timeoutMs) throw new Error("waitUntil timed out");
+    await new Promise((r) => setTimeout(r, 5));
+  }
+}
+
+interface Drive {
+  events: AgentEvent[];
+  reloadCalls: number;
+  polls: number;
+  pollParams: unknown[];
+  endTurn(): Promise<void>;
+  finish(): Promise<void>;
+}
+
+function startDrive(opts: {
+  listings: Listing[];
+  reloadThrows?: boolean;
+}): Drive {
+  const listings = [...opts.listings];
+  const events: AgentEvent[] = [];
+  let polls = 0;
+  let reloadCalls = 0;
+  let starts = 0;
+  const turnWaiters: Array<() => void> = [];
+  let waitingForTurn = false;
+
+  const client = {
+    notificationHandler: null as ((msg: unknown) => void) | null,
+    exitError: null as Error | null,
+    exitPromise: new Promise<void>(() => {}),
+    request: vi.fn(async (method: string, params?: unknown) => {
+      if (method === "thread/start" || method === "thread/resume") {
+        return { thread: { id: "thread-1" }, model: "gpt-5.6-sol" };
+      }
+      if (method === "model/list") return { data: [{ id: "gpt-5.6-sol", hidden: false }] };
+      if (method === "turn/start") {
+        starts += 1;
+        waitingForTurn = true;
+        for (const w of turnWaiters.splice(0)) w();
+        return { turn: { id: `turn-${starts}` } };
+      }
+      if (method === "mcpServerStatus/list") {
+        polls += 1;
+        drive.polls = polls;
+        drive.pollParams.push(params);
+        if (listings.length === 0) throw new Error("no listing scripted");
+        const next = listings.length > 1 ? listings.shift()! : listings[0];
+        if (next === "throw") throw new Error("mcpServerStatus/list unavailable");
+        return { data: next, nextCursor: null };
+      }
+      if (method === "config/mcpServer/reload") {
+        reloadCalls += 1;
+        drive.reloadCalls = reloadCalls;
+        if (opts.reloadThrows) throw new Error("unknown method config/mcpServer/reload");
+        return {};
+      }
+      throw new Error(`unexpected request: ${method}`);
+    }),
+    close: vi.fn(async () => {}),
+  };
+
+  const turns: Array<{ text: string }> = [];
+  let turnResolve: (() => void) | null = null;
+  let finished = false;
+  let idle = false;
+  async function* channel() {
+    for (;;) {
+      idle = false;
+      while (turns.length) yield turns.shift()!;
+      if (finished) return;
+      idle = true;
+      await new Promise<void>((resolve) => {
+        turnResolve = resolve;
+      });
+    }
+  }
+
+  const backend: Backend = new CodexBackend({
+    model: "gpt-5.6-sol",
+    mcpServers: MCP_SERVERS,
+  });
+  Object.assign(backend, { client, liveCatalog: [{ id: "gpt-5.6-sol", supportsEffort: true }] });
+
+  const done = (async () => {
+    for await (const ev of backend.run({ channel: channel() })) events.push(ev);
+  })();
+
+  const drive: Drive = {
+    events,
+    reloadCalls: 0,
+    polls: 0,
+    pollParams: [],
+    async endTurn() {
+      await waitFor(() => expect(events.some((e) => e.type === "session")).toBe(true));
+      idle = false;
+      turns.push({ text: "continue" });
+      turnResolve?.();
+      turnResolve = null;
+      if (!waitingForTurn) {
+        await new Promise<void>((resolve) => turnWaiters.push(resolve));
+      }
+      await waitUntil(() => waitingForTurn && starts > 0);
+      waitingForTurn = false;
+      const turnId = `turn-${starts}`;
+      client.notificationHandler?.({
+        method: "turn/completed",
+        params: { threadId: "thread-1", turn: { id: turnId, status: "completed" } },
+      });
+      // The MCP watch runs AFTER the turn result, before the run loop waits
+      // for the next channel item. Idle is the barrier that it finished.
+      await waitFor(() => expect(idle).toBe(true));
+    },
+    async finish() {
+      finished = true;
+      turnResolve?.();
+      await done;
+    },
+  };
+  return drive;
+}
+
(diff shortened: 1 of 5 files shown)


src/orchestrator/mcp-session-health.ts
```diff
@@ -111,6 +111,56 @@ export function inspectMcpServers(
...
+  const raw = typeof entry.status === "string" ? entry.status.trim().toLowerCase() : "";
+  if (raw === "pending" || raw === "connecting" || raw === "starting") return "pending";
+  if (raw === "needs-auth" || raw === "unauthorized") return "needs-auth";
+  if (raw === "disabled") return "disabled";
+  if (raw === "failed" || raw === "error") return raw;
+  if (raw === "disconnected") return "failed";
+  // Named, but advertising nothing — the session has the server and not its tools.
+  if (Array.isArray(entry.tools) && entry.tools.length === 0) return "failed";
+  return raw || "connected";
+}
```
artokun, line 199: P1: this does not read the field Codex actually sends.

`mcpServerStatus/list` (with `threadId`) fills `runtimeStatus`: `notStarted | starting | connected | authenticationRequired | failed | cancelled | disabled`. There is no `status` and no `enabled`. Docs: older servers omit `runtimeStatus` (treat as unknown); `tools` inventory is cached separately and **does not prove the thread is connected**.

A live drop is `{ name: "panel", runtimeStatus: "failed", tools: <cached panel_* names> }`. Here `raw` is `""`, `tools` is likely a map not a `[]`, and this returns `"connected"`. `inspectMcpServers` is then silent — no reload, reporter's ALL_TOOLS gap stays.

`authenticationRequired` would also look connected. `starting` + empty tools (if tools really is an array) would look like a drop and bounce healthy stdio `comfyui` via the global reload.

Map `runtimeStatus` (and `authStatus` if that is how OAuth is spelled) into the existing failed/pending/needs-auth/disabled vocabulary first. Empty tools is a fallback when `runtimeStatus` is omitted, not the primary signal. `notStarted` / `cancelled` are reconnectable; `starting` is pending.

src/__tests__/orchestrator/codex-mcp-reconnect.test.ts
```diff
@@ -0,0 +1,301 @@
...
+afterAll(() => vi.unstubAllEnvs());
+
+const PANEL_UP = [
+  { name: "comfyui", tools: [{ name: "list_tools" }] },
+  { name: "panel", tools: [{ name: "panel_graph_outline" }, { name: "panel_set_todo" }] },
+];
+const PANEL_GONE = [
+  { name: "comfyui", tools: [{ name: "list_tools" }] },
+  { name: "panel", tools: [] },
+];
```
artokun, line 43: This is not the wire shape, so the `run()` suite cannot prove the shipped mapper fires on a real drop.

Live rows are `{ name, runtimeStatus, authStatus, tools, ... }`. `PANEL_GONE = { name: "panel", tools: [] }` is what `statusOfCodexListing` was written against; it is not what `list_mcp_server_status_response` builds. The fake `request` also ignores params, so dropping `threadId` (without which Codex sets `runtimeStatus: null`) stays green.

Pin the live row: a `runtimeStatus: "failed"` panel with omitted or non-array `tools` must still produce exactly one `config/mcpServer/reload`. Assert the poll is `mcpServerStatus/list` with `threadId` + `detail: "toolsAndAuthOnly"`. A healthy `runtimeStatus: "connected"` listing must not reload even if `tools` is `[]` during a cache miss — that is the false alarm this closed list was meant to avoid.

src/orchestrator/codex-backend.ts
```diff
@@ -1200,6 +1220,153 @@ export class CodexBackend implements AgentBackend {
...
+      (d) =>
+        !pendingVerdict.includes(d.name) &&
+        this.mcpDown.get(d.name) !== true &&
+        reconnectableMcpStatus(d.status),
+    );
+    if (triable.length > 0) {
+      for (const d of triable) this.mcpDown.set(d.name, true);
+      try {
+        await client.request("config/mcpServer/reload", {});
+        for (const d of triable) this.mcpReloadPending.add(d.name);
```
artokun, line 1300: Residual, not a second P1: `config/mcpServer/reload` with `{}` reinitializes **every** MCP server on the thread, including healthy stdio `comfyui`. Claude's `reconnectMcpServer(d.name)` is per-server; Codex has no equivalent. Timing is after `turn/completed`, so this does not kill the orchestrator or an in-flight turn — it does bounce the comfyui MCP child.

Once the mapper actually sees `runtimeStatus: "failed"` on panel only, this is the right (only) verb. Worth not treating `starting` empty-tools as a drop, or a second episode will spend another global reload on comfyui while panel is still coming up.

artokun (commented):

The Codex turn-end watch is wired into the live `CodexBackend.run()` path, bounded to one `config/mcpServer/reload` per down-episode, and it does not touch the Claude #1681 lane. That part is real. The detection is not.

`statusOfCodexListing` reads `status` / `enabled` / `tools: []`. Current `codex app-server` `mcpServerStatus/list` (see `list_mcp_server_status_response`) emits `runtimeStatus`, `authStatus`, `pluginId`, `serverInfo`, `tools`. Official docs: inventory may be cached and **does not prove the thread is connected**; omit `threadId` and `runtimeStatus` is null. A live row `{ name: "panel", runtimeStatus: "failed", tools: <cached> }` becomes `connected` here, so no reload is queued and the ALL_TOOLS gap in the PR body survives.

The new suite stays green because `PANEL_GONE` is `{ name: "panel", tools: [] }` — a shape the app-server does not send — and the fake `request` never asserts `threadId` / `detail`. Deleting the empty-array branch goes red against that stub. Deleting a `runtimeStatus` read does not, because there isn't one.

Claude tests still green (58/58 with these files). Episode bound, needs-auth skip, throw isolation, and "stays down → one reload" are fine on the stubbed path. Global reload bouncing healthy stdio `comfyui` is residual (Codex has no per-server verb; the watch runs after `turn/completed`; the orchestrator HTTP server is not killed). The one-turn lag (first post-restart turn still missing `mcp__panel__*`) is documented, not a merge blocker once detection actually fires.

Required: map live `runtimeStatus` into the existing `ReportedMcpServer` vocabulary **before** looking at tools; treat empty tools as a fallback only when `runtimeStatus` is omitted; pin a live-shaped listing in `CodexBackend.run()` tests (including `threadId` on the poll) so deleting that branch goes red.

artokun:
Addressed the review: the mapper now reads the live `mcpServerStatus/list` row.

- `runtimeStatus` is the connection signal (`failed` / `notStarted` / `cancelled` → failed, `starting` → pending, `authenticationRequired` → needs-auth, `disabled` → disabled). Cached `tools` (a map, not an array) never override it, so `{ name: "panel", runtimeStatus: "failed", tools: { panel_set_todo: … } }` queues `config/mcpServer/reload`.
- Empty tools is only a fallback when `runtimeStatus` is omitted.
- `CodexBackend.run()` fixtures are live-shaped. Polls assert `threadId` + `detail: "toolsAndAuthOnly"`. A `runtimeStatus: "failed"` row with omitted tools still reloads — deleting that branch goes red.

artokun (commented):
COMMENT INTENDED: APPROVE

Previous REQUEST_CHANGES is fixed on this head.

`statusOfCodexListing` now reads live `runtimeStatus` first: `failed` / `notStarted` / `cancelled` → drop, `starting` → pending, `authenticationRequired` → needs-auth. Cached `tools` maps no longer look connected — `{ runtimeStatus: "failed", tools: { panel_set_todo: {} } }` is a drop. Empty tools is only a fallback when `runtimeStatus` is omitted.

The `run()` suite stubs that live row. `runtimeStatus=failed` with omitted tools still reloads, so deleting the runtimeStatus→failed branch goes red (that fixture has neither an empty array nor an empty map). Poll params are asserted (`threadId` + `detail: "toolsAndAuthOnly"`). Three still-down turns → one reload; `starting` + empty tools map does not reload; needs-auth is reported and never retried.

Claude #1681 untouched. 63/63 local on the Codex reconnect + mapper + Claude #1681 suites.

Residual, not blocking: reload still applies on the *next* turn (first post-restart turn can still miss `mcp__panel__*`); `config/mcpServer/reload` is process-wide so healthy stdio `comfyui` is bounced between turns — changelog now says so.

PR #2164, opened 2026-08-23
Fix EZi proxy restart ownership (#2118)

### LINKED ISSUE #2118: Restart tools cannot identify EZi wrapper when panel proxy port differs from ComfyUI port
On an EZi ComfyUI install, the sidebar frontend was exposed through a helper/proxy on 127.0.0.1:50404 while the actual ComfyUI API listened on 127.0.0.1:8188. panel_restart_comfyui refused because the panel URL did not match the orchestrator's boot-time ComfyUI URL. The catalog restart_comfyui fallback then inspected the process listening on 50404 (the EZi helper wrapper) instead of the actual ComfyUI child on 8188 and refused listener ownership. Both safe restart surfaces were unusable even though the active workflow and real ComfyUI instance were known and idle.

### Steps to reproduce
1. Launch ComfyUI through ComfyUI-Easy-Install Helper-CEI / ComfyUI-EZi.py.
2. Let the helper expose the panel through a dynamic proxy port such as 50404 while the ComfyUI child listens on 8188.
3. Call panel_restart_comfyui: it refuses due to 50404 vs 8188 identity mismatch.
4. Follow its prescribed restart_comfyui fallback: it inspects the 50404 wrapper listener and refuses ownership because its command line is ComfyUI-EZi.py rather than ComfyUI main.py.
5. A direct POST to the real idle ComfyUI Manager endpoint http://127.0.0.1:8188/manager/reboot succeeds and the panel reconnects.

### Exact error (scrubbed)
panel_restart_comfyui: "this panel is open at http://127.0.0.1:50404, which is not the ComfyUI this server booted against (http://127.0.0.1:8188)"
restart_comfyui: "The process listening on port 50404 ... is not running the ComfyUI that answered /system_stats ... Nothing was stopped"

### Fix
When the active panel URL is a verified local proxy/wrapper, resolve the actual ComfyUI backend identity from the bridge/orchestrator connection rather than treating the proxy listener as the ComfyUI process. If local process ownership remains unprovable but ComfyUI-Manager is reachable on the verified backend and the queue is idle, offer/use the same guarded Manager /manager/reboot path used for remote instances. EZi's child command line was a python -c runpy wrapper for ComfyUI main.py, so ownership matching should also recognize that launch shape.

### Environment
Windows 11
ComfyUI 0.33.1
ComfyUI frontend 1.50.6 (ComfyUI expects 1.48.7)
Python 3.12.10
PyTorch 2.9.1+cu130
GPU NVIDIA GeForce RTX 4090 24 GB
comfyui-mcp 0.52.65
panel 0.13.1
EZi Helper-CEI wrapper with Comfy Kitchen launcher

artokun:
Fixes #2118

When a local EZi/CEI helper exposes the panel on a different port, require identical non-empty /system_stats argv through the proxy and immutable boot endpoint, including an explicit boot-port argument. Route verified cases through the backend Manager reboot and backend readiness probe; never claim or kill the helper listener. Recheck target generation and panel binding across confirmation/preflight/dispatch awaits. Keep the existing strict process argv ownership refusal for unverified cases.

Tests: focused panel restart suites, process-control restart identity/target-anchor suites, live-interpreter, build, lint, and diff checks.

### DIFF
--- a/src/__tests__/orchestrator/panel-restart-ezi-wrapper.test.ts
+++ b/src/__tests__/orchestrator/panel-restart-ezi-wrapper.test.ts
@@ -1,0 +1,117 @@
+// #2118 — an EZi/CEI helper serves the panel on one local port while the
+// immutable ComfyUI boot endpoint and Manager live on another. The panel restart
+// must use the verified backend proof, not the helper listener's process identity.
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const hoisted = vi.hoisted(() => ({
+  target: { value: "http://127.0.0.1:50404" },
+  boot: { value: "http://127.0.0.1:8188" },
+  generation: { value: 7 },
+}));
+
+vi.mock("../../comfyui/client.js", () => ({
+  getObjectInfo: vi.fn(),
+  getQueue: vi.fn(),
+  getSystemStats: vi.fn(),
+  resetClient: vi.fn(),
+  resetObjectInfoCache: vi.fn(),
+}));
+
+vi.mock("../../config.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../../config.js")>();
+  return {
+    ...actual,
+    config: { ...actual.config, comfyuiRestartCommand: undefined },
+    getComfyUIBaseUrl: () => hoisted.target.value,
+    getBootLocalComfyUIBaseUrl: () => hoisted.boot.value,
+    getComfyuiTargetGeneration: () => hoisted.generation.value,
+    isRemoteMode: () => false,
+    isCloudMode: () => false,
+  };
+});
+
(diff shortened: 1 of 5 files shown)


artokun:
Follow-up for the independent P1 review: commit 29eee5f tightens the EZi/CEI proof so both the proxy URL and boot-derived backend URL must parse to exact numeric loopback hosts (127.0.0.1 or ::1); localhost, non-IP DNS, and loopback-looking userinfo forms fail closed before identity probes. The existing generation/argv/explicit-backend-port pinning, bounded probes, backend Manager target, and no-proxy-kill behavior are unchanged. Added production-path coverage for backend localhost plus hostile/non-IP forms. Local validation: focused restart/orchestrator sweep 13 files / 222 tests, restart-target-anchor 15/15, build, lint/typecheck, and git diff --check all pass. No merge/release performed.

PR #2218, opened 2026-08-24
fix(2217): an unrecognized key with no near-miss names the keys that ARE accepted

### LINKED ISSUE #2217: Unrecognized-key error names no valid key when nothing is a near-miss: panel_restart_comfyui{reason} says only "Unrecognized key 'reason'" (follow-up to #1969)
Split out of **artokun/comfyui-mcp-panel#1749**, which I closed as no-code-in-that-repo: the rejection it reports comes from this repo's `strictPanelSchema`, and the half of its ask that is still live is this one.

## What reproduces on current `main`

Measured on `origin/main` @ `2b7a0c3` (v0.52.96), reading the served schema off a real `McpServer` + `Client` over `InMemoryTransport` and registering through `registerPanelTools` exactly as production does — not off the zod source:

```
SERVED_SCHEMA  {"type":"object","properties":{"force":{"type":"boolean"}},
                "$schema":"http://json-schema.org/draft-07/schema#","additionalProperties":false}

callTool panel_restart_comfyui {reason: "Load custom-node fix"}
  isError true
  "MCP error -32602: Input validation error: Invalid arguments for tool
   panel_restart_comfyui: Unrecognized key 'reason'"

callTool panel_restart_comfyui {forced: true}          <- control
  "MCP error -32602: Input validation error: Invalid arguments for tool
   panel_restart_comfyui: Unrecognized key 'forced' — did you mean 'force'?"

callTool panel_restart_comfyui {force: true}           <- control
  passes validation, reaches the handler
```

The control is the point: **#1969's suggester is live and works.** `forced` is an edit-distance-1 neighbour of `force`, so the caller is told the key it wanted. `reason` is not near `force` — `isAffixMatch` is false and `editDistance("reason","force")` is over the threshold — and there are no required keys, so rule 3 cannot fire either. `suggestKnownKey` correctly returns `undefined`, `formatUnrecognizedKeyMessage` falls through to the bare `Unrecognized key 'reason'`, and the caller is told what is wrong but nothing about what is right.

## The suggestion

When `pairUnrecognizedKeys` produces no pairing for a key, append the accepted key list rather than stopping at the name:

```
Unrecognized key 'reason' — accepted keys: 'force'
```

`unrecognizedKeysError` already closes over `knownKeys`, so this is a branch in `formatUnrecognizedKeyMessage`'s no-suggestion arm and costs nothing on the success path.

## The argument against, stated fairly

The served JSON schema **already** lists the accepted keys, and the caller has it — a frontier model can read `properties` and recover without any help from the error string. So this is not information the client lacks; it is information the client did not join. That is the same gap #1969 was opened to close ("JOIN the two halves of the refusal the caller already gets ... into one line a small model can act on"), which is why I think it belongs there rather than being dismissed — but it is a polish item on an already-shipped mechanism, not a broken one. **P3 / low.** Two shapes worth deciding before building:

- **Wide schemas.** `panel_query_graph` carries ~12 keys; a full list is a long line for a stray key. A cap ("accepted keys: 'types', 'title', 'ids', … (12 total)") or a threshold may be wanted.
- **Required-key tools.** Those already get rule 3's `required key 'x' is missing`, so the new arm should only fire where nothing at all was said.

## What is explicitly NOT being asked for here

`panel_restart_comfyui` accepting `reason`. There is nowhere to put it — the panel handler is `comfy_reboot({ force })` (`web/js/comfyui-mcp-panel.js:21038`) and the confirm-card text is composed here, so a forwarded `reason` would be accepted on the wire and dropped on arrival. That is the silent drop `strictPanelSchema` exists to prevent (#754), and wiring it into the confirmation card is a new capability of the kind already parked in #1968 for `panel_create_subgraph`'s `title`. Feature freeze; not proposed.

---
Filed by claude-swarm while closing panel#1749. No claim on this one — it is a tracking handoff, not work in progress.

artokun:
Fixes #2217.

## What was wrong

#1969's suggester is live and correct — the control in the report proves it:

```
panel_restart_comfyui {forced: true}   -> Unrecognized key 'forced' — did you mean 'force'?
panel_restart_comfyui {force: true}    -> passes validation, reaches the handler
panel_restart_comfyui {reason: "..."}  -> Unrecognized key 'reason'
```

`reason` is not near `force` (`isAffixMatch` false, `editDistance` over the
threshold) and `panel_restart_comfyui` has no required keys, so rule 3 cannot
fire either. Every rule declines, `suggestKnownKey` returns `undefined`,
`formatUnrecognizedKeyMessage` falls through to the bare name, and the caller
is told what is wrong and nothing about what is right.

The served JSON schema already carries `properties`, so this is not
information the client lacks — it is information the client did not JOIN.
That is the same gap #1969 exists to close for models that will not join it
themselves.

## What this changes

One branch in `formatUnrecognizedKeyMessage`'s no-pairing arm.
`unrecognizedKeysError` already closes over `knownKeys`, so it costs nothing on
the success path.

```
Unrecognized key 'reason' — accepted keys: 'force'
Unrecognized key 'reason' — this tool takes no arguments          (0-key tools)
Unrecognized key 'reason'; Unrecognized key 'why'; accepted keys: 'force'
```

## The three shapes, decided by measurement not by guess

Measured over the 96 tools `buildPanelToolDefs()` registers:

| | |
|---|---|
| widest shape | `panel_edit_node`, 14 keys |
| longest clause | 173 characters |
| key-count distribution | 0:15, 1:19, 2:17, 3:15, 4:13, 5:4, 6:6, 7:2, 8:1, 10+:4 |

- **No cap.** The issue floated one for wide schemas. At 173 characters worst
  case, eliding entries would risk hiding the very key the caller wanted to
  save a line that is already short. If the surface ever grows a 40-key tool
  this is the decision to revisit — the measurement is in the code comment so
  it can be re-run.
- **15 tools take no arguments at all.** `accepted keys:` followed by nothing
  is a worse answer than the bare name it replaces, so those say
  `this tool takes no arguments`.
- **The arm fires only where nothing at all was said** (the issue's second
  point). A confident `did you mean 'force'?` and rule 3's
  `required key 'node_id' is missing` have already named the one key that
  matters; appending 14 more names buries it.

The list is a property of the TOOL, not of any one stray key, so it is stated
once — as that key's remedy with a single message, as a trailing clause with
several.

## Where the load-bearing claims are (please attack these)

1. **`unique.every((k) => paired.has(k))` is the gate.** If it is wrong in the
   permissive direction, every already-answered refusal grows a key list and
   #1969's whole point (say the ONE key, not the schema) is undone. Pinned by
   two negative tests over the real surface — `{forced:true}` and
   `panel_remove_node {dry_run:true}` — each asserting `not.toMatch(/accepted keys/)`.
2. **The unpaired-but-ambiguous case.** `panel_connect {id:5}` deliberately
   refuses to pick between `from_node_id` and `to_node_id`. It now gets the
   full list, which names both. I claim that is honest rather than a
   regression of #1969's refusal — it is still not a `did you mean`, and the
   existing #1969 assertions (`not.toMatch(/did you mean 'to_node_id'/)`)
   still hold. Worth a second opinion.
3. **The single-vs-multiple join.** `messages.length === 1` picks ` — ` over
   `; `. Two code paths for one clause; the alternative was one path and a
   message that reads `Unrecognized key 'reason'; accepted keys: 'force'` for
   the reported case.
4. **The widest-shape test reads its expectation off the SERVED schema**, not
   off the zod source, so a message that names a different list than the one
   the client was handed fails. The 0-key test asserts `servedKeys` is empty
   first, so it dies as "wrong tool pinned" rather than silently passing if
   `panel_free_vram` ever gains a key.

## Verification

- `src/__tests__/orchestrator/unrecognized-keys.test.ts` — 40 tests green
  (26 pre-existing + 14 new), through a real `McpServer` + `Client` over
  `InMemoryTransport` registered by `registerPanelTools`, asserting the string
  the SDK renderer hands the caller.
- Mutation-tested clause by clause (results in a comment below): deleting the
  fix, widening the gate, dropping the 0-key guard, forcing the `; ` join, and
  capping the list each kill a named test, and the #1969 suite survives all of
  them.
- Not panel-visible: nothing in `comfyui-mcp-panel` renders this string — it
  is an MCP protocol-boundary validation message. No Playwright run.

### DIFF
--- a/src/__tests__/orchestrator/unrecognized-keys.test.ts
+++ b/src/__tests__/orchestrator/unrecognized-keys.test.ts
@@ -349,3 +349,153 @@ describe("#1969 review — the live surface, not a hand-built shape", () => {
     expect(text).toMatch(/Unrecognized key 'groupid'/);
   });
 });
+
+// ---------------------------------------------------------------------------
+// #2217 — the arm where NOTHING pairs. #1969's suggester is live and correct:
+// `panel_restart_comfyui {forced:true}` is told 'force', because `forced` is an
+// edit-1 neighbour. `reason` is not near `force`, and `panel_restart_comfyui`
+// has no required keys, so every rule declines and the caller was told what is
+// wrong and nothing about what is right.
+//
+// The served JSON schema already carries `properties`, so this is not
+// information the client lacks — it is information the client did not JOIN,
+// which is the same gap #1969 was opened to close. The two suites below are
+// therefore the point: the reported call gains the list, and every call that
+// already had an answer must NOT gain it.
+// ---------------------------------------------------------------------------
+describe("#2217 a key with no near-miss names the keys that ARE accepted", () => {
+  let client: Client;
+  let server: McpServer;
+
+  beforeAll(async () => {
+    server = new McpServer({ name: "unrecognized-keys-2217", version: "1.0.0" });
+    registerPanelTools(server, makeFakeCtx());
+    const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
+    client = new Client({ name: "unrecognized-keys-2217-client", version: "1.0.0" });
+    await Promise.all([server.connect(serverTransport), client.connect(clientTransport)]);
+  });
+
+  afterAll(async () => {
+    await client.close();
+    await server.close();
+  });
+
+  const errorText = async (name: string, args: Record<string, unknown>): Promise<string> => {
+    const r = await client.callTool({ name, arguments: args });
+    expect(r.isError, `${name} was expected to reject ${JSON.stringify(args)}`).toBe(true);
+    return (r.content as Array<{ text?: string }>)?.[0]?.text ?? "";
+  };
+
+  /** The accepted keys as the CLIENT sees them — the served schema, not the
+   *  zod source. If those two ever diverge, the message names the wrong list. */
+  const servedKeys = async (name: string): Promise<string[]> => {
+    const tools = await client.listTools();
+    const schema = tools.tools.find((t) => t.name === name)?.inputSchema;
+    return Object.keys((schema?.properties ?? {}) as Record<string, unknown>);
+  };
+
+  it("REPRODUCTION: panel_restart_comfyui {reason} now names 'force'", async () => {
+    const text = await errorText("panel_restart_comfyui", { reason: "Load custom-node fix" });
+    expect(text).toContain("Unrecognized key 'reason' — accepted keys: 'force'");
+    // Still a refusal, and still no invented target: the tool does not accept
+    // a `reason`, and forwarding one is parked with #1968's `title`.
+    expect(text).toMatch(/Input validation error/);
+    expect(text).not.toMatch(/did you mean/);
+  });
+
+  it("CONTROL: the near-miss 'forced' keeps #1969's answer and does NOT gain the list", async () => {
+    const text = await errorText("panel_restart_comfyui", { forced: true });
+    expect(text).toMatch(/Unrecognized key 'forced' — did you mean 'force'\?/);
+    expect(text).not.toMatch(/accepted keys/);
+  });
+
+  it("CONTROL: the canonical key still reaches the handler", async () => {
+    const r = await client.callTool({ name: "panel_restart_comfyui", arguments: { force: true } });
+    const text = (r.content as Array<{ text?: string }>)?.[0]?.text ?? "";
+    expect(text).not.toMatch(/Input validation error|Unrecognized key/);
+  });
+
+  it("a tool that takes NO arguments says so instead of an empty list", async () => {
+    // Dies if panel_free_vram ever gains a key — at which point this pins the
+    // wrong tool, not the wrong behaviour.
+    expect(await servedKeys("panel_free_vram")).toEqual([]);
+    const text = await errorText("panel_free_vram", { reason: "cleanup" });
+    expect(text).toContain("Unrecognized key 'reason' — this tool takes no arguments");
+    expect(text).not.toMatch(/accepted keys/);
+  });
+
+  it("the WIDEST live shape is listed in full — no elision", async () => {
+    const keys = await servedKeys("panel_query_graph");
+    expect(keys.length).toBeGreaterThan(10);
+    const text = await errorText("panel_query_graph", { node_type: "KSampler" });
+    expect(text).toContain("Unrecognized key 'node_type' — accepted keys:");
+    // Every key the client was served is named. A cap would hide the one the
+    // caller wanted; measured, the full clause is 152 characters.
+    for (const k of keys) expect(text).toContain(`'${k}'`);
+    expect(text).not.toMatch(/…|\.\.\./);
+  });
+
+  it("a rule-3 tool keeps 'required key … is missing' and does NOT also get the list", async () => {
+    // The join was already made for this caller; 14 more names would bury it.
+    const text = await errorText("panel_remove_node", { dry_run: true });
+    expect(text).toMatch(/Unrecognized key 'dry_run' — required key 'node_id' is missing/);
+    expect(text).not.toMatch(/accepted keys/);
+  });
+
+  it("the ambiguous panel_connect {id} names BOTH endpoints without guessing one", async () => {
+    // #1969 refuses to pick between from_node_id and to_node_id. Listing the
+    // accepted keys hands the caller both candidates, which is the honest
+    // answer — and it is still not a "did you mean".
+    const text = await errorText("panel_connect", { id: 5 });
+    expect(text).toContain("Unrecognized key 'id' — accepted keys:");
+    expect(text).toContain("'from_node_id'");
+    expect(text).toContain("'to_node_id'");
+    expect(text).not.toMatch(/did you mean/);
+  });
+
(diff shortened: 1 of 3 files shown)


artokun:
## Mutation evidence — every clause of the fix is pinned

Run against the fix commit `ceed3f8` (the later commit is comment-only). Each
mutant edits `src/orchestrator/unrecognized-keys.ts`, runs
`src/__tests__/orchestrator/unrecognized-keys.test.ts` (40 tests: 26
pre-existing #1969, 14 new), and is restored from HEAD before the next.

The probe itself was verified first: the initial harness passed
`--reporter=basic`, which vitest 4 does not have — every "mutant" reported the
same `ERR_LOAD_URL` and read as a clean kill. Results below are from the
repaired harness, which parses the JSON reporter and prints
`NO JSON — the probe itself failed` rather than silently scoring a run that
never happened.

| mutant | tests killed |
|---|---|
| M0 baseline, unmutated | **0** — 40/40 green |
| M1 fix deleted (no-pairing arm never appends) | 9 |
| M2 gate widened (appends even over a `did you mean`) | 6 |
| M3 empty-shape guard dropped | 2 |
| M4 join always `; `, never the em dash | 7 |
| M5 list capped at 3 keys | 4 |
| M6 clause repeated per unpaired key instead of stated once | 5 |

The two that matter most:

**M1 — delete the fix.** Kills the reproduction, the 0-key tool, the
widest-shape list, the ambiguous `panel_connect {id}` case, the multi-key
clause, and all four formatter tests. Every one of the 26 pre-existing #1969
tests SURVIVES, which is the control: the new arm does not carry any existing
behaviour.

```
KILLED: #2217 … REPRODUCTION: panel_restart_comfyui {reason} now names 'force'
KILLED: #2217 … a tool that takes NO arguments says so instead of an empty list
KILLED: #2217 … the WIDEST live shape is listed in full — no elision
KILLED: #2217 … the ambiguous panel_connect {id} names BOTH endpoints without guessing one
KILLED: #2217 … two stray keys get ONE trailing clause, not one list each
KILLED: #2217 the sentence, at the formatter emits the reported string exactly
KILLED: #2217 the sentence, at the formatter an empty shape cannot produce a dangling 'accepted keys:'
KILLED: #2217 the sentence, at the formatter a paired key and an unpaired one keep their own answers, list stated once
KILLED: #2217 the sentence, at the formatter the list is stated in the schema's own order
```

**M2 — widen the gate** so an already-answered refusal also gets the list. This
is the risky direction named in the PR body, and it does not fail quietly: it
kills the two new CONTROL tests AND four pre-existing #1969 tests, including
the reporter's own sentence from that issue.

```
KILLED: #1969 … the reporter's sentence is what formatUnrecognizedKeyMessage emits
KILLED: #1969 review … a stray key whose value cannot fit the missing required key is NOT a 'did you mean'
KILLED: #1969 review … the reporter's todos → items SURVIVES as a 'did you mean' because the value fits
KILLED: #1969 review … a NAME match stays confident even when the value does not fit — the names still pair
KILLED: #2217 … CONTROL: the near-miss 'forced' keeps #1969's answer and does NOT gain the list
KILLED: #2217 … a rule-3 tool keeps 'required key … is missing' and does NOT also get the list
```

## Observed effect, not the request made

Through a real `McpServer` + `Client` over `InMemoryTransport`, registered by
`registerPanelTools`, before → after:

```
panel_restart_comfyui {reason:"Load custom-node fix"}
  before: Unrecognized key 'reason'
  after:  Unrecognized key 'reason' — accepted keys: 'force'

panel_free_vram {reason:"cleanup"}                    (served properties: {})
  before: Unrecognized key 'reason'
  after:  Unrecognized key 'reason' — this tool takes no arguments

panel_query_graph {node_type:"KSampler"}              (12 served keys)
  before: Unrecognized key 'node_type'
  after:  Unrecognized key 'node_type' — accepted keys: 'types', 'title', 'where',
          'ids', 'upstream_of', 'downstream_of', 'depth', 'fields', 'group_by',
          'limit', 'max_chars', 'widget_max_chars'

panel_restart_comfyui {reason:"a", why:"b"}
  after:  Unrecognized key 'reason'; Unrecognized key 'why'; accepted keys: 'force'

panel_restart_comfyui {forced:true}    unchanged: did you mean 'force'?
panel_restart_comfyui {force:true}     unchanged: passes validation, reaches the handler
```

## Suite

- `npm run lint` (tsc --noEmit): clean.
- Full `npx vitest run`: 12074 passed / 1 failed — `late-mutation-e2e.test.ts`
  "DID complete", the known 60 ms wall-clock flake under a 653-file parallel
  run. Re-run alone: **5/5 green**. It does not touch this diff.
- `i18n:check`, `check:docs-links`, `check:docs-locale`, `check:docs-mdx`,
  `check:blog`, `check:blog-packs`, `check:blog-stale`, `check:blog-structure`:
  all OK. `check:vocabulary`: OK (1810 files).
- codex gate: **SHIP** on `9a93713`, the SHA that will merge (an earlier SHIP on
  `ceed3f8` was re-run rather than banked, since the head moved).
- Nothing outside `unrecognized-keys.ts` reads this string: the only other
  mentions in the repo are doc comments in `panel-tools.ts` (updated in the
  second commit), and no `comfyui-mcp-panel` source parses it — the message is
  produced at the MCP protocol boundary and goes to the tool caller, never to
  the sidebar. No Playwright run; nothing here is panel-visible.

PR #2227, opened 2026-08-25
fix(panel): complete large panel_get_errors audits

### LINKED ISSUE #2199: Make panel_get_errors complete audits on 50-node workflows
panel_get_errors could not audit a normal 50-node live workflow. It returned audit_complete:false after checking only 2 nodes because "get_errors ran out of its shared server-call budget", leaving 48 unchecked. The tool therefore cannot provide the clean-bill-of-health validation it is documented to provide on moderately sized graphs.

### Steps to reproduce
1. Open or build a 50-node workflow.
2. Call panel_get_errors.
3. Observe checked_count:2, unchecked_count:48, audit_complete:false.

### Exact error (scrubbed)
audit_incomplete_reason: "get_errors ran out of its shared server-call budget"
node_count: 50
checked_count: 2
unchecked_count: 48
unchecked_budget_exhausted: true

### Fix
applied locally: no
upstream-only: yes

Precise change needed: make panel_get_errors batch its /object_info and asset checks (or scale the shared server-call budget with node_count) so its documented fallback re-check actually completes for ordinary 50+ node workflows. It should never consume nearly the whole budget before the batched fallback, and should return audit_complete:true when the server is reachable.

### Environment
Windows 11
ComfyUI 0.33.0
ComfyUI frontend 1.49.6
NVIDIA RTX 5090 32 GB
Python 3.13.11
PyTorch 2.13.0+cu130
comfyui-mcp 0.52.92
panel 0.14.24

artokun:
Fixes #2199. Pages leftover graph_query detail reads into bounded explicit-ID batches under the existing elective timeout, preserving truncated pages as unchecked and completing ordinary 50-node audits. Validation: npm.cmd exec vitest run src/__tests__/orchestrator/get-errors-ack-budget.test.ts src/__tests__/orchestrator/get-errors-call-budget.test.ts src/__tests__/orchestrator/get-errors-concurrent-reads.test.ts (3 files, 43 tests); npm.cmd run build; npm.cmd run lint; git diff --check.

### DIFF
--- a/src/__tests__/orchestrator/get-errors-call-budget.test.ts
+++ b/src/__tests__/orchestrator/get-errors-call-budget.test.ts
@@ -23,6 +23,13 @@ const EXECUTION_NODES = [
   { id: 73, type: "ImpactSwitch", widget: "select", value: "prompt_a" },
 ] as const;
 
+const FIFTY_NODE_WORKFLOW = Array.from({ length: 50 }, (_, i) => ({
+  id: i + 1,
+  type: "Loader",
+  widget: "asset",
+  value: `asset-${i + 1}.safetensors`,
+}));
+
 function budgetExhaustedReply(opts?: { extraUnchecked?: number }): Record<string, unknown> {
   const extra = opts?.extraUnchecked ?? 35;
   const extras = Array.from({ length: extra }, (_, i) => ({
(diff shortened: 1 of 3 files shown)


artokun:
Independent merge gate: NO-SHIP at exact head de721a3.

Blocking issue: get-errors-audit.ts queues all detail pages plus /object_info with Promise.all, while production serializes graph_* calls and each timeout starts only when dispatched. A 50-node audit can therefore exceed the advertised 8-second completion budget (the 200-ID ceiling can take roughly 168 seconds). Please make the budget aggregate-bounded and add a production-path regression that proves the deadline under serialized reads.

The new test also only checks set coverage and overwrites one Loader.asset fixture, so it does not prove a clean 50-node result or exact ordering. PR remains open and unmerged.

artokun:
Implemented and pushed as f63bd08 on fix/2199-panel-errors-audit.

Changes:
- Replaced concurrent Promise.all follow-ups with an ordered object_info-then-page pass.
- Enforced one absolute 8,000 ms completion deadline; each bridge call receives only the remaining time, so serialized graph_* dispatches cannot accumulate per-call timeouts.
- Kept the 200-ID ceiling and fail-closed truncated-page handling.
- Added a production-path panel_get_errors regression with a serialized graph lane proving the deadline stops later pages.
- Made the 50-node clean fixture explicit and asserted exact command order plus ordered [1..50] page coverage.

Exact validation at this SHA:
- PASS: npm.cmd exec -- vitest run src/__tests__/orchestrator/get-errors-ack-budget.test.ts src/__tests__/orchestrator/get-errors-call-budget.test.ts src/__tests__/orchestrator/get-errors-concurrent-reads.test.ts — 3 files, 44 tests.
- PASS: npm.cmd run lint.
- PASS: npm.cmd run build.
- PASS: git diff --check and git show --check --stat.
- Full npm.cmd test is not green due to the unrelated baseline src/__tests__/orchestrator/node-id-union-message.test.ts failure: 3 failed, 53 passed; its expectations still require the old "received undefined at node_id" / "received {id}" wording while the current message contains the newer node-ID diagnostic. The changed audit tests and build/lint are green.

PR remains open and unmerged.

PR #2228, opened 2026-08-25
fix: persist connect orchestrator logs (#2198)

### LINKED ISSUE #2198: `connect` mode writes no persistent log — orchestrator exits are undiagnosable

When the panel bridge dies, there is **no evidence anywhere on disk** explaining why, which makes the failure effectively undiagnosable.

Two things combine:

1. `launchMcpTerminal()` in `broker.mjs` spawns the orchestrator with `stdio: "ignore"` and `child.unref()`. Anything the orchestrator writes when launched this way goes nowhere.
2. `connect` mode does not appear to write to `~/.comfyui-mcp/launch-logs/` at all. On this machine `panel-orchestrator.log` was **6 days stale** while a `connect`-mode orchestrator was actively running and serving the bridge — so that file is only written in `--panel-orchestrator` mode.

Net result: when the user launches from the panel (broker path), output is discarded; when launched via a terminal, the exit reason lives only in that terminal's scrollback and disappears with the window. Either way, after the bridge goes down there is nothing to read.

I hit this directly while diagnosing a "panel stops responding" report. Confirming that the orchestrator (not ComfyUI) was the failing component required reconstructing state from `ps`/`ss` process and socket forensics, then writing a wrapper script to relaunch the orchestrator detached with output redirected to a file, purely to obtain a log for the next occurrence. That workaround is the signal being reported here.

Worth noting how this misleads users: because the orchestrator dies silently and the panel's reconnect button only re-establishes the browser→bridge leg, the only recovery most people discover is restarting ComfyUI — which "works" solely because the panel reload causes the broker to spawn a fresh orchestrator. The user in this case had been paying a ~7 minute ComfyUI cold start to achieve what is really a ~1 second process restart, and had no way to learn otherwise because nothing was logged.

### Steps to reproduce

1. Start the orchestrator in `connect` mode (either from the panel, which goes through the broker's `launchMcpTerminal`, or via `npx -y comfyui-mcp@latest connect`).
2. Confirm the bridge is serving on `:9199` and the panel works.
3. Let the orchestrator exit for any reason (close the terminal window, kill it, or wait for whatever ends it).
4. Look for any record of the exit under `~/.comfyui-mcp/launch-logs/`.

Observed: nothing. `panel-orchestrator.log` is untouched — its mtime remains days old across the entire lifetime of the `connect` orchestrator.
Expected: a persistent log recording startup, and ideally the exit/shutdown reason.

### Exact error (scrubbed)

No error is produced — that is precisely the defect. The observable state is:

```
ComfyUI :8188      : RESPONDING
bridge  :9199      : NOT LISTENING (no orchestrator)
```

with ComfyUI healthy (same PID, 13h uptime) and no log entry anywhere accounting for the orchestrator's disappearance.

### Fix

applied locally: yes (as an external wrapper, not a code change)  upstream-only: yes (for the real fix)

Locally I worked around it by relaunching the orchestrator detached with output redirected, which does produce a usable log:

```bash
setsid nohup npx -y comfyui-mcp@latest connect \
  >> ~/.comfyui-mcp/launch-logs/connect-orchestrator.log 2>&1 < /dev/null &
```

That immediately yielded the startup detail that had been invisible until then:

```
=== starting connect orchestrator ===
[comfyui-mcp] Found ComfyUI on port 8188
[INFO] [ui-bridge] listening on ws://127.0.0.1:9199
[INFO] [panel-orchestrator] ready — bridge on ws://127.0.0.1:9199 ...
[INFO] [self-restart] npx install — checking npm for updates every 60m
```

Precise change needed upstream — two small, independent pieces:

1. **Make `connect` mode log like `--panel-orchestrator` does.** Write to `~/.comfyui-mcp/launch-logs/connect-orchestrator.log` (or reuse the existing file). The directory and the logging convention already exist, so this is reusing established infrastructure rather than adding new machinery. Logging the exit path too (signal, self-restart, crash) is what makes the difference between diagnosable and not.

2. **Stop discarding output in the broker.** In `launchMcpTerminal()`, `stdio: "ignore"` guarantees nothing survives even if the orchestrator does log. Once (1) writes to a file this matters less, but as long as the orchestrator's only output channel is a terminal window, panel-initiated launches are unrecoverable after the window closes.

Happy to send this as a PR if the approach looks right.

### Environment

- OS: Linux (Ubuntu, GNOME)
- ComfyUI: 0.28.0 (local)
- ComfyUI **frontend**: 1.47.10 (`comfyui-frontend-package`)
- GPU: NVIDIA GeForce RTX 3080 Ti, 12GB
- torch 2.9.1 / CUDA 13.0 / Python 3.12
- comfyui-mcp: 0.52.78
- comfyui-mcp-panel: 0.15.51
- backend: Claude; triton + sage installed

artokun:
Fixes #2198

## Summary
- enable the existing stderr logger to mirror panel-orchestrator/connect records to ~/.comfyui-mcp/launch-logs/connect-orchestrator.log
- bound the persistent sink at 2 MiB, keep newest records, use owner-readable permissions, and fail open if disk logging is unavailable
- wire logging before orchestrator policy validation/dynamic import so startup and exit failures leave evidence

## Validation
- focused Vitest: 86 passed across orchestrator-log, CLI, launch-log-capture, panel-launcher, and startup-deadline suites
- npm run build
- npm run lint (tsc --noEmit)
- git diff --check
- built entrypoint probe: connect invalid URL exited 1 and created the persistent log

Commit: 4609ac9
No merge performed.

### DIFF
--- a/src/__tests__/services/orchestrator-log.test.ts
+++ b/src/__tests__/services/orchestrator-log.test.ts
@@ -90,10 +90,21 @@ describe("persistent orchestrator logger", () => {
     expect(() => logger.warn("stderr remains available")).not.toThrow();
   });
 
-  it("is wired before the orchestrator branch can fail during startup", () => {
+  it("keeps policy validation first and logs connect failures before orchestration", () => {
     const boot = readFileSync(new URL("../../boot.ts", import.meta.url), "utf8");
-    expect(boot).toContain("if (cli.panelOrchestrator)");
-    expect(boot).toContain("const logPath = orchestratorLogPath();");
+    const policy = boot.indexOf("resolveToolSurfacePolicy();");
+    const orchestrator = boot.indexOf("if (cli.panelOrchestrator)");
+    const logging = boot.indexOf("const logPath = orchestratorLogPath();", orchestrator);
+    const http = boot.indexOf('if (cli.transport === "http")', orchestrator);
+    const urlValidation = boot.indexOf("const urlError = validateConnectUrl(cli.comfyuiUrl);", logging);
+    const dynamicImport = boot.indexOf('await import("./orchestrator/index.js")', logging);
+
+    expect(policy).toBeGreaterThan(-1);
+    expect(orchestrator).toBeGreaterThan(policy);
+    expect(logging).toBeGreaterThan(orchestrator);
+    expect(orchestrator).toBeLessThan(http);
+    expect(urlValidation).toBeGreaterThan(logging);
+    expect(dynamicImport).toBeGreaterThan(logging);
     expect(boot).toContain("configurePersistentLogFile(logPath)");
   });
 });
--- a/src/__tests__/services/orchestrator-log.test.ts
+++ b/src/__tests__/services/orchestrator-log.test.ts
@@ -1,0 +1,99 @@
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { afterEach, beforeEach, describe, expect, it } from "vitest";
+import {
+  closePersistentLogFile,
+  configurePersistentLogFile,
+  logger,
+  MAX_PERSISTENT_LOG_BYTES,
+} from "../../utils/logger.js";
+import { orchestratorLogPath } from "../../services/orchestrator-log.js";
+
+let root: string;
+
+beforeEach(() => {
+  root = mkdtempSync(join(tmpdir(), "cmcp-orchestrator-log-"));
+});
+
+afterEach(() => {
+  closePersistentLogFile();
+  rmSync(root, { recursive: true, force: true });
+});
+
+describe("orchestratorLogPath", () => {
+  it("uses the shared data-dir override and stable connect log name", () => {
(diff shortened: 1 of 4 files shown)


artokun:
CI NO-SHIP regression at exact head 4609ac9: tool-surface-filter.test.ts reads boot.ts and requires resolveToolSurfacePolicy() before both the panelOrchestrator and HTTP startup branches. The #2198 patch moved that call below those branches, causing all three OS builds to fail with the source-order assertion. Please restore the boot-order invariant while retaining the persistent connect logging, add/update focused coverage, and rerun the full required CI.

artokun:
Takeover fix pushed as ec9fc84 (4609ac9 -> ec9fc84). Restored resolveToolSurfacePolicy() ahead of the first panelOrchestrator and HTTP startup branches, while retaining persistent connect-mode logging inside the orchestrator path before URL validation and dynamic import. Focused Vitest: 2 files, 43 passed. Full npm test after npm ci: 657 files, 12,115 passed, 4 skipped; lint/build, asset-counts, vocabulary, unknown-collapse, vocab export, docs generation no-op, and pack/install smoke all passed. No merge performed; hosted CI rerun is pending.

PR #2237, opened 2026-08-25
fix(panel): retry idempotent reads after transport send errors

### LINKED ISSUE #2233: Retry idempotent panel reads after transient orchestrator transport send errors
A live-canvas panel_query_graph read failed once with an rmcp WorkerTransport HTTP send error to the local orchestrator endpoint. Retrying the identical idempotent read immediately succeeded. The exact lower-level cause is not exposed, but the panel client currently surfaces the transient transport failure without a bounded retry.

### Steps to reproduce
1. Use the Codex deferred tool mcp__panel__panel_query_graph on a connected live canvas after several sequential graph mutations.
2. Observe a one-off HTTP send failure.
3. Retry the identical read; it succeeds.

### Exact error (scrubbed)
tool call failed for panel/panel_query_graph
Transport send error: WorkerTransport<StreamableHttpClientWorker<...>> error: HTTP request failed sending request to http://127.0.0.1:9198/orchestrator::codex

### Fix
Add one bounded automatic retry for idempotent panel read operations (panel_query_graph / panel_graph_outline) when the local orchestrator transport fails before an application response. Do not retry mutations without the existing retry token.

### Environment
Windows 10; ComfyUI 0.33.1; NVIDIA RTX 3090 24 GB; comfyui-mcp 0.52.89; panel diagnostic-client.

artokun:
Fixes #2233\n\n## Summary\n- classify the Codex/rmcp loopback Transport send error: WorkerTransport... wrapper as transient at the shared panel call boundary\n- retain the existing one-retry bound and RETRY_SAFE_CMDS gate, so panel reads may retry but mutations do not\n- add real Streamable HTTP MCP production-path coverage for panel_query_graph, panel_graph_outline, and a mutation refusal\n\n## Verification\n- 
pm.cmd exec vitest run src/__tests__/orchestrator/panel-mcp-stale-session.test.ts\n- focused orchestrator suite: 490 tests passed\n- 
pm.cmd run build\n- 
pm.cmd test: 658 files passed, 12,128 tests passed, 4 skipped; i18n/docs/blog checks passed\n\nThe native etter-sqlite3 binding was rebuilt locally after the initial dependency install used --ignore-scripts; no source or lockfile changes resulted. Not merged.

### DIFF
--- a/src/__tests__/orchestrator/panel-mcp-stale-session.test.ts
+++ b/src/__tests__/orchestrator/panel-mcp-stale-session.test.ts
@@ -29,11 +29,17 @@
 // These tests drive the REAL server over a real socket. A fake would only prove
 // the shape I already believe.
 import { afterEach, describe, expect, it } from "vitest";
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js";
 
 import { startPanelMcpHttpServer, type PanelMcpHttpServer } from "../../orchestrator/panel-mcp-http.js";
 import type { UiBridge } from "../../services/ui-bridge.js";
 
 const TAB = "tmp:1524";
+
+const TRANSIENT_ORCHESTRATOR_SEND_ERROR =
+  "Transport send error: WorkerTransport<StreamableHttpClientWorker<...>> error: " +
+  "HTTP request failed sending request to http://127.0.0.1:9198/orchestrator::codex";
 
 /** A bridge that is never actually reached — these tests stop at the transport. */
 const bridge = {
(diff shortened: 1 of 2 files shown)


artokun:
Independent review: NO-SHIP. Blocking defect: the new WorkerTransport matcher is applied through the existing retry allowlist, and set_todo is currently allowlisted, so this PR retries a UI-state mutation. Evidence: src/orchestrator/panel-tools.ts:1414-1434, 11992-11996, 13632-13638, and panel_set_todo at 17773-17792. The test covers graph_set_widget only, which is not allowlisted. Also add coverage at the actual HTTP MCP request-send handoff; direct fake readBridge.send injection does not prove that production boundary. Re-engaging the worker; do not merge.

artokun:
NO-SHIP correction applied in follow-up commit 1555557 (parent 6bcfcac preserved).\n\nReviewer finding addressed: the WorkerTransport wrapper must not inherit RETRY_SAFE_CMDS entries such as set_todo. The legacy isTransientReconnectError matcher is unchanged; the new wrapper has a separate exact allowlist containing only graph_query and graph_outline, and both ctx.call and dispatchToTab gates enforce it.\n\nCoverage now drives the raw panel-mcp-http JSON-RPC handoff (initialize, initialized notification, tools/call) and proves both target reads retry exactly once. It also proves set_todo is one attempt through the direct ctx.call path and the redirected dispatchToTab path.\n\nValidation: focused production test 9 passed; focused orchestrator suites 491 passed; build passed; full npm.cmd test passed with 658 files, 12,129 tests, 4 skipped, plus i18n/docs/blog checks.\n\nPR remains open and unmerged. Awaiting fresh independent review and hosted CI.

PR #2240, opened 2026-08-25
fix(codex): recover stale panel MCP transport

artokun:
## Summary

- Treat the observed Codex WorkerTransport HTTP-send failure for the orchestrator-hosted panel MCP as one bounded down episode even when mcpServerStatus/list still reports the cached row as connected.
- Queue the existing config/mcpServer/reload control-plane recovery at the turn boundary; never re-issue 	urn/start, panel_run, or another render mutation.
- Surface an actionable outcome-unknown message so the Panel does not append the generic Nothing was lost — try again guidance.
- Preserve the existing Panel bridge distinction between pre-write dispatched:false and post-write outcome-unknown failures.

Refs artokun/comfyui-mcp-panel#1777

## Validation

- PASS: 
pm.cmd exec vitest run src/__tests__/orchestrator/codex-mcp-reconnect.test.ts src/__tests__/orchestrator/panel-mcp-stale-session.test.ts src/__tests__/orchestrator/session-notice-render.test.ts — 3 files, 27 tests.
- PASS: 
pm.cmd run build (	sc).
- INTERRUPTED: 
pm.cmd test was started but intentionally stopped at the requested checkpoint before its final result; it is not claimed green.
- PASS: git diff --check before commit.

Independent review requested before merge. Do not merge this PR as part of the issue work.

### DIFF
--- a/src/__tests__/orchestrator/codex-mcp-reconnect.test.ts
+++ b/src/__tests__/orchestrator/codex-mcp-reconnect.test.ts
@@ -45,6 +45,9 @@ const MCP_SERVERS = {
   comfyui: { transport: "stdio" as const, command: "node", args: ["mcp.js"] },
   panel: { transport: "http" as const, url: "http://127.0.0.1:9198/tab" },
 };
+const WORKER_TRANSPORT_FAILURE =
+  "Transport send error: WorkerTransport<StreamableHttpClientWorker<...>> error: " +
+  "HTTP request failed sending request to http://127.0.0.1:9198/orchestrator%3A%3Acodex";
 
 const noticesOf = (events: AgentEvent[]) =>
   events.filter(
@@ -74,7 +77,8 @@ interface Drive {
   reloadCalls: number;
   polls: number;
   pollParams: unknown[];
-  endTurn(): Promise<void>;
+  turnStarts: number;
+  endTurn(kind?: "completed" | "worker-transport"): Promise<void>;
   finish(): Promise<void>;
 }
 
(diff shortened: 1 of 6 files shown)


artokun:
Independent review: NO-SHIP P1. At codex-backend.ts:1885-1902, a WorkerTransport failure can emit the terminal local error before honoring willRetry:true; cleanup at :2098-2113 does not interrupt the active turn, while turn/interrupt is separate at :2135-2144. Recovery/reload can then proceed with the original app-server mutation still active, so duplicate-render safety is unproven. Add a willRetry:true regression and fence/interrupt the original turn before recovery (or otherwise prove it cannot continue), while preserving pre-write not-dispatched vs post-write outcome-unknown semantics. Re-engaging worker; do not merge.

artokun:
Correction pushed in 3baa931.

The WorkerTransport special case now distinguishes Codex willRetry:true from an already-terminal error. For willRetry:true it immediately fences local notification handling, sends a bounded turn/interrupt, and only then emits the actionable outcome-unknown terminal error. If interruption cannot be confirmed, the app-server client is closed before recovery/reconnect/new-turn processing. Late turn completion/retry notifications are dropped.

The existing bounded MCP reload remains control-plane only: no panel_run/render mutation retry. Existing pre-write dispatched:false versus post-write outcome-unknown bridge semantics are unchanged.

Validation:
- PASS: focused production-path suite — 6 files, 60 tests
- PASS: npm.cmd run build
- PASS: npm.cmd test — 658 files, 12,136 passed, 4 skipped; i18n/docs/blog checks passed
- PASS: git diff --check

Fresh independent review is requested for this correction. Do not merge PR #2240 as part of issue #1777 work.

PR #2256, opened 2026-08-25
fix(panel): preserve animated media payloads

### LINKED ISSUE #2248: panel_show_media: mobile/remote client renders animated GIFs as a single static frame

`panel_show_media` was used to deliver an animated GIF to the **mobile/remote client** (not the ComfyUI sidebar panel). The client displays only the **first frame** — it looks like a still image to the user.

The GIF itself is well-formed. Verified by parsing the file directly:

- `GIF89a` header
- `NETSCAPE2.0` application extension present (loop block)
- 338 Graphic Control Extension blocks, i.e. every frame carries its own timing
- per-frame delays 70-80ms (non-zero)
- 26s total duration

So this is a client-side rendering gap, not a malformed asset. `web/js/lib/media-preview.js` in the panel pack has a comment stating animated GIFs are expected to render, which suggests the sidebar panel handles this correctly and the mobile/remote renderer diverges.

### Why it is costly

The user's environment already cannot play inline mp4, so GIF is the documented fallback for delivering motion. With GIF also non-animating, there is **no way to show motion inline at all**. In this session it caused roughly six wasted capture/encode cycles: each clip was re-shot and re-encoded on the assumption the capture had failed, when the footage was fine and the delivery surface was dropping the animation. Contact sheets/filmstrips are the only working workaround, and they lose timing entirely.

Compounding it: `panel_show_media` returns `{"shown": true}` from this client with no per-item accounting, so the agent gets no signal that the animation was dropped. The failure is silent in both directions.

### Steps to reproduce

1. Produce an animated GIF (any multi-frame GIF89a with a NETSCAPE2.0 loop block and non-zero delays).
2. Call `panel_show_media` with `{ path: "<abs path to gif>" }` against the mobile/remote client.
3. Observe: only the first frame is shown; the image never advances or loops.
4. Open the same file in a browser: it animates correctly.

### Suggested fix direction

Render GIFs via a plain `` element with the original bytes rather than decoding to a bitmap/canvas or generating a static thumbnail. If a thumbnail step exists for layout or performance, exempt `image/gif` (and `image/apng`, `image/webp` which may be animated) from it. Additionally, have this client return real per-item accounting (`count`/`painted`/`unrenderable`) so an agent can detect the drop instead of assuming success.

### Fix

applied locally: no  |  upstream-only: yes

The mobile/remote client's rendering code is not present in the installed `comfyui-mcp` package on this machine, so there was nothing local to patch. Workaround in use: deliver a tiled filmstrip PNG inline and open the animated file in a browser for the user.

### Environment

OS: Windows 11 · GPU: RTX 5080 16GB · ComfyUI 0.33.3 · ComfyUI frontend 1.49.6 · Python 3.13.12 · PyTorch 2.10.0+cu130 · comfyui-mcp 0.52.97 · panel 0.15.70 · headless session (no sidebar panel attached; mobile/remote client only)

artokun:
## Summary

- accept `.apng` as an image in `panel_show_media` and the bridge filename classifier
- preserve GIF/APNG/WebP source bytes and MIME in the outbound data URL
- make unaccounted mobile/remote replies disclose that animated frame advancement is not observed
- add focused regression coverage for outbound bytes, MIME, view-ref classification, and accounting disclosure

Closes #2248

## Validation

- `npm.cmd exec vitest run src/__tests__/orchestrator/panel-show-media-animated.test.ts src/__tests__/orchestrator/panel-show-media-unaccounted-ack.test.ts src/__tests__/orchestrator/panel-show-media-view-probe.test.ts src/__tests__/orchestrator/panel-show-media-client-kind.test.ts` — 4 files, 100 tests passed
- `npm.cmd run lint` — passed
- `npm.cmd run build` — passed
- `git diff --check` — passed

The full Vitest run also exposed three pre-existing failures in `src/__tests__/orchestrator/node-id-union-message.test.ts` and then entered an unrelated long-running cancellation test path; no failure occurred in the touched media suites.

## Scope limitation

The mobile/remote renderer and its `{shown:true}` response implementation are not present in this repository. This PR does not claim to make that client animate frames or fabricate per-item `count`/`painted`/`unrenderable` results. It ensures this repository does not reject APNG or flatten animated bytes, and it reports the renderer limitation honestly when the client supplies no accounting.

### DIFF
--- a/src/__tests__/orchestrator/panel-show-media-animated.test.ts
+++ b/src/__tests__/orchestrator/panel-show-media-animated.test.ts
@@ -27,13 +27,15 @@ function makeCtx(): { ctx: PanelToolCtx; calls: Forwarded[] } {
   return { ctx, calls };
 }
 
-async function showMedia(path: string): Promise<{ res: ToolResult; calls: Forwarded[] }> {
+async function showMedia(items: Array<Record<string, unknown>>): Promise<{ res: ToolResult; calls: Forwarded[] }> {
   const def = buildPanelToolDefs().find((d) => d.name === "panel_show_media");
   if (!def) throw new Error("panel_show_media not found");
   const { ctx, calls } = makeCtx();
-  const res = (await def.handler({ items: [{ source: { path } }] }, ctx)) as ToolResult;
+  const res = (await def.handler({ items }, ctx)) as ToolResult;
   return { res, calls };
 }
+
+const textOf = (res: ToolResult): string => res.content.map((c) => ("text" in c ? c.text : "")).join("\n");
 
 const cases = [
   [".gif", "image/gif", Buffer.from("GIF89a\x00animated-frame-data", "binary")],
@@ -56,7 +58,7 @@ describe("#2248 animated image delivery", () => {
     const path = join(root, `animated${ext}`);
     writeFileSync(path, bytes);
 
-    const { res, calls } = await showMedia(path);
+    const { res, calls } = await showMedia([{ source: { path } }]);
     expect(res.isError).toBeFalsy();
 
     const command = calls.find((call) => call.cmd === "show_media");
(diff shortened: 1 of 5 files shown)


artokun:
Independent exact-head review: NO-SHIP (P1 correctness). unaccountedShowMediaNote infers animated bytes from filename alone (src/services/comfy-view-ref.ts:1001-1005), but reference-only panel_show_media items can be sent as kind viewRef without dataUrl (src/orchestrator/panel-tools.ts:21580-21599). The diagnostic must distinguish inline byte dispatch from reference-only delivery, with regression coverage. No merge until corrected and freshly reviewed.

artokun:
Correction pushed at exact head 1dc430b919b8ab654b9ccb83176de5690e939dc0. The animation caveat now triggers only for inline dataUrl items; filename-only viewRef items no longer claim bytes were dispatched. Focused tests, lint, build, and diff checks pass. Fresh independent review is running.

artokun:
Fresh independent exact-head review SHIP for 1dc430b919b8ab654b9ccb83176de5690e939dc0. The prior P1 is fixed: animated caveats require inline dataUrl delivery; filename-only viewRef items are excluded. Byte/MIME preservation and regression coverage verified. Hosted pack smoke is green; OS CI remains pending.

artokun:
Hosted CI is fully green: pack smoke plus macOS, Windows, and Ubuntu. Fresh independent exact-head review SHIP is recorded for 1dc430b919b8ab654b9ccb83176de5690e939dc0. Proceeding to merge.

PR #2266, opened 2026-08-25
fix(panel): apply canonical download cache dir on reload

### LINKED ISSUE #2255: panel_reload does not apply COMFYUI_DOWNLOAD_CACHE_DIR from canonical env to download_model
A large local model download was correctly refused because the default cache volume was low. I added COMFYUI_DOWNLOAD_CACHE_DIR to ~/.comfyui-mcp/.env and called panel_reload(scope=orchestrator), whose contract says it respawns the agent and comfyui tool server. A fresh download_model call still reported the old default cache path. A standalone fresh Node import of dist/config.js loaded the same canonical env file and printed the configured E: cache path correctly, so the env file and dotenv parsing are valid. The active download service/tool session was not refreshed with that value.

### Steps to reproduce
1. Start the panel without COMFYUI_DOWNLOAD_CACHE_DIR in the process environment.
2. Add COMFYUI_DOWNLOAD_CACHE_DIR to ~/.comfyui-mcp/.env.
3. Call panel_reload(scope=orchestrator).
4. Start a download larger than the free space on ~/.comfyui-mcp/cache.
5. Observe download_model still checks the old default cache path.

### Exact error (scrubbed)
NOT downloaded: this needs 19.03 GB of staging space and the download cache volume has 3.47 GB free (~/.comfyui-mcp/cache). Point the cache at a volume with room by setting COMFYUI_DOWNLOAD_CACHE_DIR, then retry.

### Fix
Applied locally: workaround only. Upstream-only: yes.

The safe workaround was to move the existing cache contents to the configured E: cache directory and replace ~/.comfyui-mcp/cache with a Windows directory junction to that target. Retrying the identical download then started successfully, proving the model URL and staging guard were healthy.

Precise upstream change needed: ensure the local download worker/service is actually respawned with the canonical env file when panel_reload(scope=orchestrator) claims to rebuild the comfyui tool server, or have cacheDir() re-read COMFYUI_DOWNLOAD_CACHE_DIR from the canonical env source when process.env lacks it. If the download worker intentionally lives in the long-lived orchestrator, expose that limitation and a supported cache update/restart path instead of claiming the tool server reload applies config.

### Environment
Windows 11; ComfyUI 0.19.3; frontend 1.42.14; Python 3.10.11; PyTorch 2.7.0+cu128; RTX 4070 Laptop 8 GB; comfyui-mcp 0.52.104; panel 0.15.81.

artokun:
Fixes #2255\n\nThe panel-spawned comfyui MCP child gets an explicitly constructed environment. buildComfyuiMcpEnv now resolves COMFYUI_DOWNLOAD_CACHE_DIR from the canonical env file at spawn time, preserving real-environment precedence and file-derived rotation/revoke semantics. Only this non-secret key is forwarded; download-cache.ts remains authoritative for normalization, volume checks, and containment behavior.\n\nRegression coverage exercises the production spawn-env boundary for a cache path added after boot and a rotated canonical path.\n\nVerified:\n- 243 relevant test files: 6092 passed, 4 skipped\n- Focused env-file/panel-secrets: 48 passed\n- npm run lint (tsc --noEmit)\n- git diff --check\n\nHead: 036aa03

### DIFF
--- a/src/__tests__/services/panel-secrets.test.ts
+++ b/src/__tests__/services/panel-secrets.test.ts
@@ -1,6 +1,10 @@
 import { describe, expect, it, beforeEach, afterEach, vi } from "vitest";
-import { MANAGED_SECRET_KEYS_ENV, resetEnvFileProvenanceForTests } from "../../env-file.js";
-import { mkdtempSync, rmSync, readFileSync, existsSync } from "node:fs";
+import {
+  MANAGED_SECRET_KEYS_ENV,
+  loadEnvFileIntoProcess,
+  resetEnvFileProvenanceForTests,
+} from "../../env-file.js";
+import { mkdtempSync, rmSync, readFileSync, existsSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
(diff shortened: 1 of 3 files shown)


artokun:
SHIP (fresh exact-head review): 036aa0332c1dd39b36723c302898d17d2a5c1d43 against 326a9da9e3920070b8e7b9aa1f126796e93f0209. Canonical COMFYUI_DOWNLOAD_CACHE_DIR reaches panel-spawned download_model workers without broadening env/path trust; hosted CI 32849625706 fully green.

PR #2269, opened 2026-08-25
fix(restart): anchor portable relative main.py (#2260)

### LINKED ISSUE #2260: panel_restart_comfyui resolves live portable install script as bare main.py and refuses safe restart
`panel_restart_comfyui` refused before stopping the server because its relaunch resolver reduced the active portable ComfyUI launch to the relative path `main.py`. The connected local server is actually running from `<COMFYUI>/main.py`, and that file exists. The tool therefore cannot reuse a proven live local launch even though the running process and server are reachable.

### Steps to reproduce
1. Run a Windows portable ComfyUI from `<COMFYUI>` using its bundled Python and `main.py`.
2. Connect the sidebar panel and call `panel_restart_comfyui`.
3. Observe refusal saying the resolved script `main.py` does not exist.

### Exact error (scrubbed)
`Refusing to restart ComfyUI: Resolved ComfyUI script does not exist on disk: main.py — could not locate the ComfyUI install`

### Fix

Precise change needed: when the captured launch script is relative, resolve it against the running ComfyUI process working directory / configured live install root before checking existence. Preserve the current fail-closed behavior if the resulting absolute path still cannot be proven. This logic runs in the long-lived panel orchestrator, so a local package edit would not take effect without restarting that process.

Verification evidence: `<COMFYUI>/main.py` exists and is readable; the queue was idle; the refusal occurred before any process was stopped, so ComfyUI remained online.

### Environment
Windows 11; ComfyUI 0.19.3; ComfyUI frontend 1.42.14; Python 3.10.11; PyTorch 2.7.0+cu128; NVIDIA GeForce RTX 4070 Laptop GPU 8 GB; comfyui-mcp 0.52.106; panel 0.15.81.

artokun:
Fixes #2260.\n\nSummary:\n- capture an absolute /system_stats cwd while the live process is bracketed\n- use it as the existing live-cwd restart anchor for stock Windows portable main.py launches\n- reject relative cwd values and retain existing refusal/symlink protections\n- add stock portable-layout success and refusal regression coverage\n\nValidation:\n- npm.cmd exec -- vitest run src/__tests__/services/restart-live-first.test.ts --reporter=dot\n- relevant services/orchestrator suite: 249 files, 5818 passed, 4 skipped\n- npm.cmd run lint\n- npm.cmd run build\n- git diff --check

### DIFF
--- a/src/__tests__/services/restart-live-first.test.ts
+++ b/src/__tests__/services/restart-live-first.test.ts
@@ -187,6 +187,13 @@ const EMBEDDED_ROOT = resolve("ComfyUI_embedded");
 const EMBEDDED_MAIN = join(EMBEDDED_ROOT, "main.py");
 const EMBEDDED_PYTHON_DIR = join(EMBEDDED_ROOT, "python");
 const EMBEDDED_PYTHON = join(EMBEDDED_ROOT, "python", "python.exe");
+// #2260: the stock Windows portable bundle keeps its embedded interpreter beside
+// the inner ComfyUI checkout. The launcher may `cd ComfyUI` before invoking bare
+// `main.py`, so the server cwd is the only direct anchor for that script.
+const PORTABLE_BUNDLE_ROOT = resolve("ComfyUI_windows_portable");
+const PORTABLE_SERVER_ROOT = join(PORTABLE_BUNDLE_ROOT, "ComfyUI");
+const PORTABLE_MAIN = join(PORTABLE_SERVER_ROOT, "main.py");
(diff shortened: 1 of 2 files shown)


artokun:
SHIP (fresh exact-head review): eff4e76a70283647a5fb42e5fcb791fb199527c8 against de37c0997fb5ced107d4cec42cd63e8f89752724. Absolute /system_stats.cwd anchors stock Windows portable layouts; missing/relative cwd refuses before kill/spawn; identity/ref/symlink protections remain. CI 32852804881 passed Windows, Ubuntu, macOS, and pack smoke.

PR #2270, opened 2026-08-25
chore(release): 0.52.110

artokun:
Release portable restart anchor fix from #2260. Metadata-only release.


artokun:
SHIP (fresh exact-head release review): f904e6e5a0077188d9593bc0199f9e0fb9994e64 directly follows 25aa1eef25dc343b4fa49571871051e2df240161. Metadata-only three-file diff, versions coherent at 0.52.110, exactly one #2260 entry, diff clean, and CI 32853731275 passed all four jobs.

PR #2271, opened 2026-08-25
fix(panel): prevent staged Manager sync false-success

### LINKED ISSUE #639: install_panel(update) reports success but does not update the panel on disk (silent no-op via stale ComfyUI-Manager)

artokun:
Fixes #639\n\nThe legacy ComfyUI-Manager 3.x reservation path is a pending startup update, not a completed sync. Mark that installer result explicitly and keep performPanelSync from returning synced:true or verifiedVersion until the on-disk panel changes.\n\nRegression coverage exercises panel 0.15.57 behind required 0.15.58, Manager's stale total_count:0/done_count:2 response, unchanged disk version, and the startup reservation.\n\nValidation:\n- focused regression: 1 passed\n- panel service suite: 24 files, 915 passed\n- npm run lint\n- npm run build\n- git diff --check

### DIFF
--- a/src/__tests__/services/panel-installer.test.ts
+++ b/src/__tests__/services/panel-installer.test.ts
@@ -742,6 +742,7 @@ describe("runPanelAction", () => {
     expect(r.message).toMatch(/STAGED, not failed/);
     expect(r.message).toMatch(/RESTART ComfyUI/);
     expect(r.restartRequired).toBe(true);
+    expect(r.staged).toBe(true);
     // The misreport that cost the reporter four attempts and a reinstall path.
     expect(r.message).not.toMatch(/did NOT apply/);
     expect(r.message).not.toMatch(/silent no-op/);
--- a/src/__tests__/services/panel-sync.test.ts
+++ b/src/__tests__/services/panel-sync.test.ts
@@ -438,6 +438,7 @@ function makeDeps(opts: {
   installedVersion?: string;
   pin?: PanelPinState;
   dirs?: string[];
+  managerReserved?: boolean;
   /** Mutate the live files map to simulate what the Manager actually did. */
   onUpdate?: (files: Record<string, string>) => void;
   updateDetails?: unknown;
@@ -445,6 +446,11 @@ function makeDeps(opts: {
   const files: Record<string, string> = {};
   if (opts.installedVersion !== undefined) {
     files[join(PANEL_DIR, "pyproject.toml")] = pyproject(opts.installedVersion);
+  }
+  if (opts.managerReserved) {
+    files[
+      join(COMFY, "user", "__manager", "startup-scripts", "install-scripts.txt")
+    ] = `['${PANEL_REGISTRY_ID}', '#LAZY-CNR-SWITCH-SCRIPT', 'https://x/y.zip']\n`;
   }
   const dirs = opts.dirs ?? (opts.installedVersion !== undefined ? ["comfyui-mcp-panel"] : []);
   const h: Harness = { updates: 0, installs: 0, deps: undefined as never };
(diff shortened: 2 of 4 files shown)


artokun:
SHIP (fresh exact-head review): dea97615d78957b05b113ebe74ef9e7d0bc95138 against 2ee6177eeef633df5bea01cd8623d700a76d87ea. Legacy staged Manager updates are recognized with positive evidence; sync returns synced:false/staged:true/restartRequired:true, preserves previousVersion, omits stale verifiedVersion, and fallback/containment protections remain. Focused 251, panel cluster 915, build clean; hosted CI 32855511149 fully green.

PR #2289, opened 2026-08-25
feat(console): /prompts is a real editor for every system prompt

artokun:
## The bug

`http://127.0.0.1:<console>/prompts?token=…` answered `{"ok":false,"error":"not_found"}`. The route never existed.

The prompt registry (`src/services/prompt-overrides.ts`) has shipped `listPrompts` / `setPromptOverride` / `clearPromptOverride` / `isKnownPrompt` since the fork port in 748e5462 — whose own commit message claims it added *"prompt-overrides service + /prompts editor page + /api/prompts endpoints"*. The "adapt fork port to current main" step kept the store and dropped the page and the endpoints, so all four exports had **zero callers**. Every editable prompt — the panel persona included — was unreachable from any UI, and the store's whole design (a separate override file so a bad edit is one Reset away) had no way to be used.

## What this adds

A web view, not a JSON dump — same shape as the existing `/credentials` page, so the panel can iframe it (CSP `frame-ancestors`, `postMessage` resize/close) or you can open it in a tab.

- **`GET /prompts`** — one row per registered prompt: label, help note, a badge reading *customised* vs *built-in default*, a textarea for the override, **Save**, **Reset to built-in**, and the built-in itself behind a disclosure with one click to copy it into the editor. Token-gated; a rejection renders in the reader's own language, like `/credentials`.
- **`GET/POST /api/prompts`** — list, save, reset.
- The landing page links the editor and carries its own `?token=` through.

## Decisions worth review

- **A write is refused for an id nothing registered.** Otherwise an arbitrary key lands in `panel-prompts.json` where nothing ever reads it and nothing ever offers to remove it — a typo'd id would look saved forever.
- **The response echoes the STORED state, not the request.** Whitespace-only text clears an override in the store, so echoing the request would leave the badge claiming "customised" over a prompt that is running the built-in. Pinned by a test.
- **`reset: true` and an empty `value` stay distinct on the wire** even though the store treats them identically — a UI that sends an accidentally-empty textarea should not be indistinguishable from one whose user pressed Reset.
- **The textarea starts empty rather than pre-filled with the default.** Saving an untouched row would freeze today's built-in as a stored override, and the user would silently stop receiving improvements to it. The default is one click away instead.
- **Rows are built in the browser** from `/api/prompts`: the registry is populated at orchestrator startup, so a page rendered before a backend registered its prompt would be permanently stale. Prompt text reaches the DOM as `textContent`, never `innerHTML`.

## i18n

20 new keys × 11 languages; `i18n:check` green at 100%. The textarea and the built-in block are pinned `direction: ltr` on Arabic/Persian pages — a prompt is code-shaped text in every language, and an RTL textarea reorders its punctuation.

## Testing

- New `src/__tests__/orchestrator/console-prompts.test.ts` (8 cases): listing, save → `resolvePrompt` reads the saved text, Reset, stored-vs-sent echo, unknown-id refusal writing nothing, token gating on both page and endpoints, the page being framed and free of inlined prompt text, and the landing link.
- Extended `panel-console-i18n.test.ts` (3 cases): the page renders from the catalog including the script-block labels, it survives a hostile catalog entry, and a translation that drops `{file}` loses the sentence but never the path inside it.
- Verified interactively in the browser end-to-end: copy default → edit → Save flips the badge and writes `panel-prompts.json`; Reset clears it back to `{}` and disables itself; no console errors; the landing link carries the token. Confirmed again against a live orchestrator after restart.
- Full suite: 11099 passed, 19 failed — all 19 pre-exist on a clean tree (`extra-paths*`, `comfy-view-ref`, `training-datasets`, `node-id-union-message`; see comfyui-mcp-1qe). Confirmed by stashing and re-running.
- `tsc --noEmit`, `i18n:check`, and every `check:docs-*` / `check:blog-*` gate pass.

## Not included

The MCP console has never been documented under `docs/`, so no page was added — a new English page pulls in 11 translated counterparts under the docs-locale gate. Worth a follow-up if the console is going to grow.


artokun:
## Merge gate round 1: NO-SHIP, one P1 — fixed in `04b5010`

The gate found the advertised editor link **401s on every path a reader can actually reach it from**, and it was right.

`/console` reads `?token=` off its own URL and copies it onto the `#prompts-link` href. Nothing ever put a token there: the two "Advanced ↗" buttons — on `/credentials` (line 552) and on `/prompts` itself (line 766) — are the only in-product routes to `/console`, and both opened `window.open(CFG.consoleUrl + "/console")` with no query string. So `carried` was null on every real path, the link shipped bare, and the gate on `/prompts` rejected it. The feature this PR exists to add was reachable only by hand-typing a URL with a secret the reader cannot see from that page.

The original comment half-admitted it — *"the panel's 'Advanced ↗' does not, a hand-typed URL may"* — which is a consumer written against a producer that does not exist.

### The fix, and what it deliberately does not do

The obvious repair is to render the token into the landing page server-side, the way `/credentials` does at line 1156. **That would be a security regression.** `/credentials` and `/prompts` both call `tokenOk()` before they render; `/` and `/console` do not. Inlining the token there hands it to precisely the unauthenticated reader the gate exists to stop.

So the producer is fixed instead: both gated pages already hold the token and already define `q = (t) => "?token=" + encodeURIComponent(t)` for their own fetches, and now use it when opening the console. The landing page's carry is unchanged — it finally has an input. A hand-typed `/console` still has no token, still renders the bare link, and still lands on the 401, which is correct: that page has no secret to give it.

### Test change

`links the editor from the console landing page, carrying the token through` asserted only that the link existed — the name claimed the behaviour the bug broke while the body could not see it. Replaced by:

- an `it.each` over `/credentials` and `/prompts` pinning that each hands its own token to `/console`, so the carry has a producer;
- a pin that the ungated page's HTML never contains the token.

Both were mutated to confirm they are live: reverting the two `data-advanced` handlers fails exactly the two producer cases and leaves the other 8 green; injecting the literal token into the landing HTML fails exactly the leak case.

`src/__tests__/orchestrator/` 207 files / 3351 tests green, `tsc --noEmit` clean, `i18n:check` 100%.

Re-gating on `04b5010`. @artokun the security shape above is the part worth a second opinion — the trade is *ungated landing page keeps no secret* against *the link needs one*.

artokun:
## Merge gate round 2 on `04b5010`: **SHIP**

Worktree HEAD confirmed equal to the PR head before gating (`04b5010`), so the verdict names the tree that merges. CI green on all four checks (Build ×3, Pack & install smoke).

Round 1's P1 — the advertised `/prompts` link 401ing on every reachable path — is fixed by carrying the token from the two gated pages that already hold it, rather than inlining it into the ungated landing page. Merging.

Issue #1634, opened 2026-08-15
Agent returns a truncated prompt from a node, and does graph lookups for questions that have nothing to do with the canvas


artokun:
Reported by **Fealow [ハイネ]** in Discord (2026-08-15, ~4:27 PM). Two symptoms, filed together because they may share one root cause.

## Symptom A — the prompt comes back cut off

> "might just be how my positive prompt node works, but every time i ask the agent to grab the prompt from it it always seems to provide a cut off version"

Asking the agent to read the prompt out of a positive-prompt node returns a truncated value rather than the full widget text.

## Symptom B — unrelated questions still trigger node/workflow lookups

> "also noticing that when im just asking a simple question unrelated to the workflow or nodes it still seems to be calling node and workflow look ups etc... which makes the response take longer for no apparent reason"

## Maintainer hypothesis (artokun, same thread)

> "I think it always tries to get a view of the canvas before answering as if it is a part of your conversation"

## A lead — UNVERIFIED, and scoped to one backend

`src/orchestrator/ollama-backend.ts:353` puts this in the system prompt:

> "To read the user's graph, **ALWAYS start with panel_graph_outline** (a compact text map) via panel_call_tool. For specifics use panel_query_graph — filter by types/where ('cfg>7'), traverse upstream_of/downstream_of, or read ONE node's exact detail with {ids:[id], fields:'detail'}. **Its output is token-bounded**, so it can never flood your context."

If that instruction is what fires, both symptoms fall out of one cause:

* **A** — `panel_graph_outline` is a *deliberately* token-bounded summary. An agent that reads a prompt out of the outline, instead of the documented exact-detail path (`panel_query_graph {ids:[id], fields:'detail'}`), would hand back a value that is cut off **by design**. The truncation would be the summary working correctly, surfaced as if it were the node's real content.
* **B** — "ALWAYS start with…" reads as unconditional, so a model that classifies almost anything as graph-adjacent will do an outline call before answering, which matches the latency the reporter describes.

**Important qualifier:** that line exists **only** in `ollama-backend.ts` — the local-LLM / OpenAI-compatible lane. It is not in a shared prompt, so it does **not** apply to the Claude / Codex / Gemini paths. I have not confirmed which backend the reporter is on, and if they are on a subscription backend this lead is the wrong tree.

I am deliberately not asserting this as the cause. It is one grep, not a reproduction, and a confidently-wrong mechanism in an issue steers the fix away from the real one.

## What would settle it

[...]

PR #2304, opened 2026-08-26
fix(2299): the dynamic-combo gate's own probe reads at pinpoint budget

artokun:
Follow-up to #2301 (merged as `bde3b50`), which landed while gate round 2 was still running. This is the one finding from that round that was not yet addressed.

## What the gate found

> [P1] `src/orchestrator/panel-tools.ts:6100` — A failed or unreadable detail probe falls through to `graph_set_widget` — with a live dynamic-combo STRING child such as `model.prompt`, a probe error/truncated response can still produce reported success while frontend reserialization overwrites the value, leaving the #2299 false-success reachable.

## What I did NOT change, and why

**Fail-open stays.** Failing closed would refuse *every* dotted widget on *every* node whenever a probe hiccups — rgthree `lora_N.*`, promoted subgraph paths, JSON composite dotted fields — which is a far larger blast radius than the residual it closes. It is also the discipline every other node-keyed refusal in this file already follows (`refuseAnimaRegionalPromptWrite`: "an unreadable probe is fail-open").

And the residual is a slice of the **pre-existing** bug, not one the guard introduced: without the guard, every one of these writes is already a false success. Closing it completely needs #2299's own option 2 — an honest "written, not confirmed persistent" receipt — which changes the success contract of `panel_set_widget` and should be its own decision, not something smuggled into a guard.

## What was worth fixing: how easily that branch is reached

The probe ran at the default `max_chars`. The panel degrades an oversized detail row to an `{id, type, title}` stub — which drops `inputs`, so neither half of the shape is provable and the write falls open.

The node most likely to overflow that budget is **the node already holding a long prompt** — i.e. exactly the reported case, whose repro step 2 is `value: "<any long prompt>"`. So the fail-open branch was not a remote edge case here; it lined up with the report.

This is a single id at `limit: 1` — one row. The survey cap exists to keep a 200-node listing small and has nothing to protect on a pinpoint read (the panel's own `#1634` comment makes exactly this distinction). Read at the `panel_query_graph` ceiling instead.

## Verification

- Affected suites: **497 passed** across `panel-tools`, `promoted-widget`, `panel-retry-identity`, `promoted-control-persist`.
- **Mutation proof.** Committed first, then removed `max_chars` from the probe and re-ran: `1 failed | 416 passed`.

```
× probes with a pinpoint budget, so a long-prompt row is not degraded to a stub
```

- The budget is asserted on the **emitted command** (`fields:"detail"`, `limit:1`, `max_chars:60000`), not on a helper — a constant alone would be invisible to the call site.
- A degraded stub row is separately pinned as keeping the setter path, so changing the fail-open decision has to be deliberate rather than drift.
- `npm run lint`, `npm run check:vocabulary` exit 0.

## Risky parts, named

- **The `60000` constant is the panel's `max_chars` ceiling**, matched to the 500–60000 clamp this repo already documents. If the panel's clamp ever moves, this silently reads at whatever the panel allows — it is a ceiling request, not a contract.
- **A generous budget is not free.** It is bounded by the node's real size and only ever one row, but it is a larger reply than before on every dotted-widget write.
- **No live ComfyUI on this host** (`127.0.0.1:8188` refused), so the truncation path is reasoned from the panel's `fitDetailLine` / `capSummaryWidgets` degradation, not reproduced by execution. Stating that plainly rather than implying coverage I do not have.

## Still open after this

`refuseAnimaRegionalPromptWrite` has the same outer-scope-only hole that #2301's `f5d2c53` fixed for dynamic-combo children: it is not re-checked on the promoted-inner retry. Untouched here — pre-existing and out of scope. Worth its own issue.

Refs #2299.

### DIFF
--- a/src/__tests__/orchestrator/panel-tools.test.ts
+++ b/src/__tests__/orchestrator/panel-tools.test.ts
@@ -684,6 +684,46 @@ describe("panel-tools: panel_set_widget V3 dynamic-combo sub-widgets (#2299)", (
     expect(res.content[0].text).toMatch(/is a STRING child/);
   });
 
+  it("probes with a pinpoint budget, so a long-prompt row is not degraded to a stub", async () => {
+    // The panel degrades an oversized detail line to an {id, type, title} stub.
+    // That stub has no `inputs`, so the shape becomes unprovable and the write
+    // falls open — on exactly the node #2299 reported, the one holding a long
+    // prompt. The survey cap has nothing to protect on a single-id, limit-1 read.
+    const sent: Array<Record<string, unknown>> = [];
+    await defByName("panel_set_widget").handler(
+      { node_id: 23, widget: "model.prompt", value: "NEW" },
+      {
+        call: async (cmd: Record<string, unknown>) => {
+          sent.push({ ...cmd });
+          if (cmd.cmd === "graph_query") {
+            return {
+              content: [{ type: "text" as const, text: JSON.stringify(dynamicDetail, null, 2) }],
+            };
+          }
+          return { content: [{ type: "text" as const, text: JSON.stringify(SET_OK, null, 2) }] };
+        },
+      } as unknown as PanelToolCtx,
+    );
+    const probe = sent.find((c) => c.cmd === "graph_query");
+    expect(probe).toBeDefined();
+    expect(probe?.fields).toBe("detail");
+    expect(probe?.limit).toBe(1);
+    expect(probe?.max_chars).toBe(60000);
+  });
+
+  it("falls open on a degraded/truncated detail row rather than refusing every dotted write", async () => {
+    // The documented residual: an unprovable shape keeps the setter path. Failing
+    // CLOSED here would refuse every dotted widget on every node whenever a probe
+    // hiccups. Pinned so that changing it is a deliberate decision, not a drift.
+    // A row the panel degraded to its {id, type, title} stub: no `inputs`, so
+    // neither half of the shape is provable.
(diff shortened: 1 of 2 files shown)


artokun:
## Gate: **SHIP** (exit 0) on `5d66927`

```
node .claude/autopilot/codex-gate.mjs --worktree .../wt-2299-cc --base origin/main --name 2304
SHIP
GATE_EXIT=0
```

The verdict names `5d66927`, which is the current PR head — the branch was rebased onto `bde3b50` before gating, and I force-pushed the gated commit rather than the pre-rebase one so the verdict carries to the SHA that will actually merge.

Round history on this work, for the record:

| SHA | Gate | Finding |
|---|---|---|
| `9eeca78` | NO-SHIP | P1 — the guard never re-probed on the promoted-inner retry, so a dynamic-combo child inside a subgraph was still a false success. Fixed in `f5d2c53`. |
| `f5d2c53` | NO-SHIP | P1 — an unreadable/truncated probe falls open. Fail-open kept deliberately; what this PR fixes is how easily that branch is reached. |
| `5d66927` | **SHIP** | — |

Note `f5d2c53` merged as part of #2301 while that round-2 gate was still running, so this PR carries the remaining finding.

Caveat carried forward from every round: the gate reported `targeted Vitest execution was blocked by environment spawn EPERM`, so it **did not run tests itself** — each verdict is a read of the diff. The suite numbers in the PR body are mine, run in `.claude/worktrees/wt-2299-cc` (497 passed, plus the mutation proof).

PR #2307, opened 2026-08-26
fix(2302): the RunPod image maps Impact Pack's model categories onto the volume

### LINKED ISSUE #2302: RunPod image: extra_model_paths.yaml omits ultralytics_bbox/ultralytics_segm/sams, so Impact Pack detector models land on the ephemeral image layer

`docker/runpod/extra_model_paths.yaml` (baked into `artokun/comfyui-mcp-runpod`, loaded via `main.py --extra-model-paths-config`) maps ~25 model categories to `/workspace/models/<cat>` with `is_default: true`, which is what makes Manager's install-model write to the network volume instead of the image.

It does **not** include the folder keys that ComfyUI-Impact-Pack / ComfyUI-Impact-Subpack register: `ultralytics`, `ultralytics_bbox`, `ultralytics_segm`, `sams`. Those categories therefore keep ComfyUI's built-in `/opt/ComfyUI/models/<cat>` as their only (and first) entry.

Consequence: any Impact Pack user downloading a detector model gets it written to `/opt/ComfyUI/models/ultralytics/bbox/` — the immutable/ephemeral image layer. It works for the life of the pod and is silently lost on a rebuild, so the user re-downloads the same YOLO weights every time they provision. This defeats the stated purpose of the file for a very common node pack (FaceDetailer is one of the most-used custom nodes in ComfyUI).

This is not hypothetical — it is exactly what happened in the session that produced this report. `download_model action:"download"` reported:

> ComfyUI-Manager reported writing it to /opt/ComfyUI/models/ultralytics/bbox/face_yolov8m.pt

and the tool's own warning ("commonly the install's base models directory; on a container that is often an ephemeral overlay, which loses the file on restart") fired correctly. The warning is right; the yaml is what's incomplete.

### Steps to reproduce

1. Launch a pod from `artokun/comfyui-mcp-runpod:1.14` with a `/workspace` network volume.
2. Install ComfyUI-Impact-Pack + ComfyUI-Impact-Subpack.
3. `download_model action:"download"` with `target_subfolder: "ultralytics/bbox"` (e.g. `face_yolov8m.pt` from `https://huggingface.co/Bingsu/adetailer`).
4. Observe the file lands in `/opt/ComfyUI/models/ultralytics/bbox/`, not `/workspace/models/ultralytics/bbox/`.
5. Rebuild the pod → model is gone; every other model category persisted.

### Exact error (scrubbed)

No error is raised. This is a silent-persistence failure: the download reports success, `list_local_models` lists the file, and the node resolves. The loss only shows up on the next pod rebuild.

Confirmed on the affected pod:

```
$ cat /opt/ComfyUI/extra_model_paths.yaml | grep -i 'ultralytics\|sams'
(no matches)

$ ls -ld /opt/ComfyUI/models
drwxr-xr-x 1 root root 50 Aug 25 23:56 /opt/ComfyUI/models   # real dir, not a symlink to the volume
```

### Fix

applied locally: no (patched the packaged copy to produce the diff, but this is image content — it needs an image rebuild to reach users)
upstream-only: yes

Add the four missing keys to `docker/runpod/extra_model_paths.yaml`:

```diff
--- a/docker/runpod/extra_model_paths.yaml
+++ b/docker/runpod/extra_model_paths.yaml
@@ -59,3 +59,10 @@
     geometry_estimation: models/geometry_estimation/
     optical_flow: models/optical_flow/
     detection: models/detection/
+
+    # Impact Pack / Impact Subpack detector + SAM models. Keys match the
+    # folder_names_and_paths entries those packs register at import time.
+    ultralytics: models/ultralytics/
+    ultralytics_bbox: models/ultralytics/bbox/
+    ultralytics_segm: models/ultralytics/segm/
+    sams: models/sams/
```

Notes for whoever picks this up:

- Impact Subpack registers `ultralytics`, `ultralytics_bbox` and `ultralytics_segm`; Impact Pack registers `sams` (for `SAMLoader`, which feeds FaceDetailer's `sam_model_opt`). All four are worth adding together — a user who hits the bbox case hits the SAM case immediately after.
- Because these keys are absent rather than wrong, `is_default: true` never applies to them, so there is no ordering subtlety in the fix — adding the keys is sufficient.
- Worth a broader audit of the same class: any popular pack that registers its own `folder_names_and_paths` category (e.g. `insightface`, `facerestore_models`, `rembg`) has the identical silent-loss behaviour on this image.

Manual workaround for affected users, for the record (survives stop/start, not a rebuild):

```bash
mkdir -p /workspace/models/ultralytics/bbox /workspace/models/ultralytics/segm /workspace/models/sams
mv -n /opt/ComfyUI/models/ultralytics/bbox/*.pt /workspace/models/ultralytics/bbox/
# then append the four keys above to /opt/ComfyUI/extra_model_paths.yaml and restart ComfyUI
```

### Environment

OS: Linux (RunPod pod, image `artokun/comfyui-mcp-runpod:1.14`)
GPU: NVIDIA GeForce RTX 5090, 31 GB VRAM · 31 GB RAM · CUDA 12.8 · torch 2.9.1 · python 3.12
ComfyUI: 0.28.0 (remote, reached via RunPod proxy) · Manager v4
comfyui-mcp: 0.52.119
comfyui-mcp-panel: 0.15.85
Network volume mounted at `/workspace`

---
Filed from the ComfyUI panel by **claude** · model `claude-opus-5` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Fixes #2302.

## What was wrong

`docker/runpod/extra_model_paths.yaml` maps ~25 model categories to `/workspace/models/<cat>` with `is_default: true`, which is what makes Manager's install-model write to the network volume instead of the image. It omitted the four categories that **the Impact packs register for themselves at import time** — ComfyUI core never registers them, so they were not in the ComfyUI example the list was mirrored from.

With no `/workspace` entry, those categories resolved only to the in-image `/opt/ComfyUI/models/<cat>`. A FaceDetailer user's YOLO/SAM weights were written to the ephemeral layer: the download reports success, `list_local_models` lists the file, the node resolves — and it is silently gone after the next pod rebuild, so they re-download the same weights on every provision.

## The mechanism, verified against upstream source (not assumed)

The load-bearing question is whether the packs' own registration **overwrites** our yaml entry — if it did, adding the keys would be a no-op. It does not:

- `ComfyUI-Impact-Subpack/modules/subpack_nodes.py:9-11` registers `ultralytics_bbox`, `ultralytics_segm`, `ultralytics`
- `ComfyUI-Impact-Pack/modules/impact/impact_pack.py:51` registers `sams`

Both go through a helper (`utils.add_folder_path_and_extensions`) that calls `folder_paths.add_model_folder_path(name, path)` **without** `is_default`, i.e. it takes the `paths.append(...)` branch, then unions the extension set.

So the fix is **order-independent**, which is worth stating explicitly:

| load order | result |
|---|---|
| yaml first | key created with `/workspace` at index 0; the pack **appends** `/opt/...` behind it |
| pack first | key exists with `/opt/...`; the yaml's `is_default: true` **inserts** `/workspace` at the front |

Either way `/workspace` ends up first, which is the download target. The issue's claim that "`is_default: true` never applies to them" is correct for the first row specifically — with the key absent, `add_model_folder_path` takes its `else` branch and creates `([path], set())`, ignoring `is_default` — and the empty extension set is then repaired by the pack's union. Checked `map_legacy` too: it only rewrites `unet`/`clip`, so none of these four keys are remapped.

## Scope — three copies of one list

The category list exists in **three** places, and fixing only the yaml would leave the others silently out of step:

1. `docker/runpod/extra_model_paths.yaml` — the four keys (the actual fix)
2. `docker/runpod/post_start.sh` — the cold-volume `mkdir -p` loop, whose own comment says it creates "the model category subfolders that `extra_model_paths.yaml` maps"
3. `docker/runpod/README.md` — embeds a full copy of the yaml and claimed it mapped "**every**" model category. That claim is the assumption that produced this bug, and that copy is what a maintainer reads when adding the *next* category.

## Where to look hardest (the risky parts)

- **The `for sub in` regex** in the test. If `post_start.sh`'s loop is restructured, a naive parse would match nothing and the parity assertion would pass vacuously — the exact "a count cannot see a parse failure" shape. Guarded with a non-null assert on the match plus `created.size > 20`, and mutation-tested below.
- **Direction of the parity assertion.** It is deliberately `yaml ⊆ post_start.sh`, not set equality: an unmapped category with no directory is the defect; a spare directory is harmless. Exact equality would make an unrelated addition fail here.
- **`onnx` is deliberately NOT added.** Impact Pack registers it at `impact_pack.py:52`, but it backs a deprecated ONNX provider, so it is documented rather than mapped. `insightface` / `facerestore_models` / `rembg` are named in the issue as the same shape — I did **not** audit them and the README says so rather than implying they were checked.

## Verification

Full suite in the worktree: **669 files, 12325 passed, 6 skipped, 0 failed**. `tsc --noEmit` clean.

Note `vitest.config.ts` excludes `**/.claude/**` and excludes are additive, so a run from inside the mandated worktree collects **zero** files and reports it as a clean run. Ran against a self-contained scratch config (untracked, not in this branch) that drops only that exclude — `Test Files 669 passed` is the line I read, not the absence of FAIL.

The fix was **deleted and watched to fail**, one mutant at a time, each with the other tests as surviving controls:

| mutant | killed | survived |
|---|---|---|
| remove the four yaml keys | `maps the four categories…`, `documents the same category set…` | 2 |
| remove the four `post_start.sh` dirs | `pre-creates every mapped category…` | 3 |
| break the `for sub in` loop header | `pre-creates every mapped category…` | 3 |
| revert the README yaml copy | `documents the same category set…` | 3 |

Restored tree back to 4/4 green. The `for sub in` list was also verified by **executing** the patched loop rather than reading it — it expands to 32 dirs ending in `sams`, `ultralytics`, `ultralytics/bbox`, `ultralytics/segm`.

## What I could NOT verify

**This change reaches users only through an image rebuild** — it is baked content, so nothing here is exercised by CI or by a running orchestrator. I did not build or launch a pod. The runtime behaviour above is derived from upstream source read directly (linked by file:line), not observed on a pod.

Not panel-visible, so the Playwright browser gate does not apply — no panel code path is touched.

## Executed reproduction (added after the first draft)

Reading upstream source establishes the mechanism but not the outcome, so I re-implemented ComfyUI's `add_model_folder_path` + `load_extra_path_config` and the packs' `add_folder_path_and_extensions` verbatim and **ran the real yaml through them** in both load orders.

Against this branch's yaml — both orders identical:

```
OK  ultralytics_bbox   download_target=/workspace/models/ultralytics/bbox  exts=8
    all=['/workspace/models/ultralytics/bbox', '/opt/ComfyUI/models/ultralytics/bbox']
OK  sams               download_target=/workspace/models/sams              exts=8
RESULT: volume wins in this order
```

The control — the **same** simulation against `origin/main`'s yaml — reproduces the reported failure rather than passing anyway, which is what makes the run above worth anything:

```
BAD ultralytics_bbox   download_target=/opt/ComfyUI/models/ultralytics/bbox
    all=['/opt/ComfyUI/models/ultralytics/bbox']
RESULT: *** VOLUME DID NOT WIN ***
```

That is exactly what the reporter saw Manager say (`writing it to /opt/ComfyUI/models/ultralytics/bbox/face_yolov8m.pt`). Note the `exts=8` in both columns: the yaml creates the key with an **empty** extension set, and the pack's union repairs it — so no category ends up unable to list its own files.

Parsed with **PyYAML 6.0.3**, the parser ComfyUI actually uses, not the JS `yaml` package the test uses.

## Shipping

The image version is resolved from Docker Hub tags at release time by `scripts/runpod-release.mjs`, not from a file in the repo, so there is nothing to bump here. Merging does **not** by itself reach pod users — this is baked image content and needs an `npm run runpod:release` rebuild to land.

### DIFF
--- a/docker/runpod/README.md
+++ b/docker/runpod/README.md
@@ -175,16 +175,22 @@ So SSH + Jupyter + nginx come from the base; **our hook adds everything else**.
   volume.
 * **`--input-directory` / `--output-directory`** — inputs and generated images on
   the volume.
+* **`--models-directory /workspace/models`** — makes ComfyUI's
+  `folder_paths.models_dir` the persistent volume root. This is required for
+  ComfyUI-Manager's explicit relative `save_path` values; `is_default` only
+  changes the ordering of category search paths.
 * **`--extra-model-paths-config`** — points every model category at
   `/workspace/models` (see below). MODELS persist on the volume.
 * **We deliberately do NOT use `--base-directory`.** It would relocate
   `custom_nodes` (and temp) onto the volume too — but we want `custom_nodes` to
-  stay **in the image**. The per-dir flags + `extra_model_paths.yaml` give us
-  exactly the split we want (data on the volume, software in the image).
+  stay **in the image**. The per-dir flags, explicit `--models-directory`, and
+  `extra_model_paths.yaml` give us exactly the split we want (data on the
+  volume, software in the image).
 
 > **Flag verification.** These flag names were verified against ComfyUI `master`
-> (`comfy/cli_args.py`): `--user-directory`, `--input-directory`,
-> `--output-directory`, `--base-directory`, and `--extra-model-paths-config` all
+> (`comfy/cli_args.py`): `--user-directory`, `--models-directory`,
+> `--input-directory`, `--output-directory`, `--base-directory`, and
+> `--extra-model-paths-config` all
 > exist with these exact spellings, and `--base-directory`'s own help text states
 > it sets the base for "models, custom_nodes, input, output, temp, and user
 > directories" — confirming why we avoid it. **Assumption to confirm at build:**
@@ -249,11 +255,12 @@ comfyui_mcp_volume:
 
 Notes:
 
-* **`is_default: true`** is the load-bearing line. ComfyUI's
+* **`is_default: true`** keeps the volume category paths first for lookup. ComfyUI's
   `add_model_folder_path(..., is_default=True)` **inserts** the path at the front
-  of each category's list, so `/workspace/models/<cat>` becomes the default
-  download target. Without it, the image's `/opt/ComfyUI/models/<cat>` (ephemeral)
-  would stay first and Manager downloads would **not** persist.
+  of each category's list, so `/workspace/models/<cat>` is preferred for lookup.
+  The launch's explicit `--models-directory /workspace/models` is the separate
+  guarantee that Manager's explicit relative `save_path` writes also land on the
+  volume; `is_default` alone cannot change `folder_paths.models_dir`.
 * **No `custom_nodes:` key** — on purpose. Mapping `custom_nodes` to the volume
   would break the fast-restart contract.
 * `base_path: /workspace` with `models/...` relative subpaths mirrors ComfyUI's
--- a/docker/runpod/README.md
+++ b/docker/runpod/README.md
@@ -198,10 +198,10 @@ So SSH + Jupyter + nginx come from the base; **our hook adds everything else**.
 ## Model paths (`extra_model_paths.yaml`)
 
 `extra_model_paths.yaml` is baked at `/opt/ComfyUI/extra_model_paths.yaml` and
-loaded via `--extra-model-paths-config`. It maps **every** model category to a
-subfolder under `/workspace/models`, with `is_default: true` so the volume is the
-**primary** (download) location — i.e. Manager's install-model writes there and
-the files persist.
+loaded via `--extra-model-paths-config`. It maps every model category **this
+image knows about** to a subfolder under `/workspace/models`, with
+`is_default: true` so the volume is the **primary** (download) location — i.e.
+Manager's install-model writes there and the files persist.
 
 ```yaml
 comfyui_mcp_volume:
@@ -239,6 +239,12 @@ comfyui_mcp_volume:
     geometry_estimation: models/geometry_estimation/
     optical_flow: models/optical_flow/
     detection: models/detection/
+
+    # registered by the node packs themselves, not by ComfyUI core:
+    ultralytics: models/ultralytics/          # Impact Subpack
+    ultralytics_bbox: models/ultralytics/bbox/
+    ultralytics_segm: models/ultralytics/segm/
+    sams: models/sams/                       # Impact Pack (SAMLoader)
 ```
 
 Notes:
@@ -255,6 +261,16 @@ Notes:
 * The category **keys** match ComfyUI's `folder_names_and_paths` keys (verified
   against `master`'s `folder_paths.py`). The category list mirrors the example's
   full set; harmless if a future ComfyUI renames one.
+* **Custom node packs register their own categories, and those are NOT in ComfyUI's
+  example.** A category absent from this file has no `/workspace` entry at all, so
+  `/opt/ComfyUI/models/<cat>` stays its only path and downloads land on the ephemeral
+  image layer - no error, just gone after the next rebuild (#2302). Adding the key is
+  sufficient and order-independent: packs register via a helper that calls
+  `add_model_folder_path` **without** `is_default`, i.e. it appends behind whatever this
+  file already put at the front. Impact Pack also registers `onnx` (read from its
+  source, deliberately left out of #2302's fix as it is a deprecated provider);
+  `insightface`, `facerestore_models` and `rembg` are the same shape but have NOT
+  been audited — do that before adding them.
 * `/post_start.sh` pre-creates the matching subfolders under `/workspace/models`
   so they exist on a cold volume.
 
--- a/docker/runpod/README.md
+++ b/docker/runpod/README.md
@@ -273,6 +273,16 @@ Notes:
   been audited — do that before adding them.
 * `/post_start.sh` pre-creates the matching subfolders under `/workspace/models`
   so they exist on a cold volume.
+* **This file does not reach a node pack's OWN installer.** It adds *search* paths;
+  it does not move `folder_paths.models_dir`, which is what Impact Pack's and Impact
+  Subpack's `install.py` derive their download directory from (Subpack has no
+  `folder_paths` fallback at all). So `post_start.sh` also exports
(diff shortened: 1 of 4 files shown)


artokun:
Independent merge gate: **NO-SHIP (P1)**.

The YAML, cold-volume mkdir list, README copy, and static tests are coherent, and hosted CI is green. However, the core explicit-download path is not proven persistent:

- `src/services/model-resolver.ts:84-96,117-128,3074-3081` emits explicit Manager `save_path` values for `sams`, `ultralytics`, `ultralytics/bbox`, and `ultralytics/segm`.
- Manager 4.2.2 resolves explicit non-`default` paths under `folder_paths.models_dir`.
- `docker/runpod/post_start.sh:612-619` launches ComfyUI without `--models-directory`, so `folder_paths.models_dir` remains `/opt/ComfyUI/models`.
- `is_default: true` changes ComfyUI search order but does not move Manager's explicit writer root.

Observed consequence: the new YAML keys make the files discoverable on `/workspace`, but an MCP `download_model` using those explicit paths can still write to `/opt/ComfyUI/models/<category>` and lose the files on rebuild. Please either make the RunPod launch/configuration set the models root to `/workspace/models` without relocating custom nodes, or adjust the MCP resolver/Manager route so explicit downloads target the persistent root, and add a test that proves the emitted path reaches that root. Keep the existing artifact parity tests, but they are not sufficient for this failure mode. No merge/release until this is independently re-gated.

Checks observed: all four hosted CI jobs passed; no Docker/RunPod live Manager download was available.

artokun:
Takeover update: pushed f868699 (fix(2302): persist Manager model downloads on RunPod). post_start.sh now passes --models-directory "${MODELS_DIR}", README documents that this changes folder_paths.models_dir for explicit Manager relative save_path writes (distinct from is_default lookup ordering), and the regression pins launch flag ordering. Focused test: 1 file / 6 passed; build passed; diff check passed. The pre-existing untracked vitest.local2302.config.ts is preserved.

Issue #2234, opened 2026-08-25
[Bug] Antigravity backend redundantly prepends 29KB systemAppend to argv, overflowing Windows 32K limit and wasting tokens


bahaynes:
### Summary
In the Antigravity backend adapter (`src/orchestrator/antigravity-backend.ts`), `runTurn()` prepends the full ~29,000-character `systemAppend` preamble into the CLI prompt argument (`-p "<text>"`) on Turn 1.

Because `ensureMcpConfig()` already merges the `comfyui` and `panel` MCP server definitions into `~/.gemini/config/mcp_config.json`, the CLI discovers and registers all tools and schemas natively. Passing a 29KB plain-text description of those same tools over `argv` is redundant and causes two issues on Windows:

1. **Command Line Overflow (`E2BIG` / `EINVAL`):** Windows enforces a hard 32,767-character limit on `CreateProcess` command lines. The 29KB preamble leaves ~2KB of headroom for user input and workflow graphs before triggering the `cmdLen > 30_000` preflight rejection.
2. **Input Token Overhead:** Incurs ~7,500+ unnecessary input tokens on every fresh conversation turn.

---

### Environment
- **OS:** Windows 10 / 11
- **Backend:** `antigravity` (`agy`)
- **Package:** `comfyui-mcp` (>= 0.52.85)

---

### Steps to Reproduce
1. Start `comfyui-mcp connect` with the `antigravity` backend selected.
2. In a ComfyUI sidebar tab, send a prompt containing a moderate workflow JSON, error trace, or detailed request (~5,000 characters).
3. The turn is rejected immediately:
   > `This message is too large for the Antigravity CLI (38343 chars; Windows caps a command line at ~32K, and agy takes the prompt as an argument). Send a shorter message, or start a fresh chat to drop replayed context.`

---

### Root Cause
In `src/orchestrator/antigravity-backend.ts`:

1. `ensureMcpConfig()` attaches tool servers to `~/.gemini/config/mcp_config.json`:
   ```ts
   writeFileAtomic(file, mergeAgyMcpConfig(existing, servers));
   ```
2. In `runTurn()`, the orchestrator attaches `deps.systemAppend` (~29KB of tool documentation) to `text`:
   ```ts
   if (this.needsSystemPreamble && this.deps.systemAppend) {
       text =
           `<system>\n${this.deps.systemAppend}\n</system>\n\n` +
           `The user's first message follows.\n\n${text}`;
[...]

PR #2316, opened 2026-08-26
fix(2311): panel_list_mcp reports the SESSION's MCP servers, not the config file's

### LINKED ISSUE #2311: panel_list_mcp reports inherited server available but Codex receives no tools or resources
The sidebar's panel_list_mcp reports an inherited MCP server named story-mixer-comfy as "available to you", but the Codex session receives none of that server's tools or resources. The deferred tool registry has zero story_mixer tools, and list_mcp_resources(server="story-mixer-comfy") returns unknown MCP server. The server is configured as a stdio FastMCP server and exposes valid @mcp.tool functions, so configured/inherited state is being reported as callable state without verifying session injection.

### Steps to reproduce
1. Configure a stdio MCP named story-mixer-comfy in the inherited Claude config.
2. Start/restart a Codex-backed sidebar session.
3. Call panel_list_mcp: story-mixer-comfy appears under inherited.
4. Search the direct/deferred tool registry: no story_mixer tools.
5. Call list_mcp_resources for the same server.

### Exact error (scrubbed)
resources/list failed: unknown MCP server 'story-mixer-comfy'

### Fix
Precise change needed: when constructing a non-Claude session, either inject/proxy inherited MCP server tools/resources into the backend tool registry, or surface the server startup/tool-budget failure. panel_list_mcp should distinguish configured servers from successfully loaded/callable servers and must not describe an unverified configured entry as "available to you".

### Environment
Windows 10; ComfyUI 0.33.1; frontend 1.48.7; RTX 3090 24 GB; comfyui-mcp 0.52.95; panel 0.15.85; backend Codex.

---

artokun:
Fixes #2311

## What was wrong

The reporter had `story-mixer-comfy` in their Claude config and ran a **Codex**-backed sidebar session. `panel_list_mcp` listed it under `inherited`, under a tool description reading *"List the MCP servers available to you."* The deferred tool registry held none of its tools, and `list_mcp_resources(server="story-mixer-comfy")` answered `unknown MCP server`.

Configured state was being reported as callable state.

## Why it happens — our own wiring, observed, not inferred

| where | what it does |
|---|---|
| `buildMcpServers()` (`orchestrator/index.ts`) | spreads `readUserMcpServers()` into the set handed to `PanelAgent` |
| `PanelAgent`'s constructor | forwards `deps.mcpServers` to the default `ClaudeBackend` **only** — an *injected* backend never reads that field |
| `makeHttpBackendMcpServers()` | is what every injected (CLI) backend is wired from, and it declares exactly two servers: the stdio `comfyui` child and the loopback `panel` HTTP MCP |

So the Claude lane inherits `~/.claude.json` and **no other lane does**. `panel_list_mcp` read the config file on every lane.

Reproduced before touching anything, by driving the real handler with a fake `~/.claude.json`:

```
=== panel_list_mcp RESULT lane=orchestrator::codex ===   (BEFORE)
{ "inherited": ["story-mixer-comfy"], "builtin": ["comfyui","panel"], "note": "After panel_add_mcp / panel_remove_mcp, call panel_reload …" }
```

## The fix, and what three review rounds took off it

The handler was always *able* to tell the truth; nothing ever told it which lane it was serving. So each lane states that fact on the ctx it builds. The review rounds then established that "which lane" is only the first of three questions, and the shipped answer is deliberately narrower than the first draft:

**Round 1 (P1) — the lane is only half of it.** `panel_add_mcp` writes to `~/.claude.json` *immediately*, and the handler read that file live. On the Claude lane, between the write and the `panel_reload` that respawns the agent, a brand-new server was reported as this session's. Fixed by snapshotting names at spawn: `makeAgent()` builds `mcpServers` (which re-reads the config) and `panelServer` as two properties of **one synchronous object literal**, so a snapshot taken there *is* the set that spawn declared.

**Round 2 (P1) — a snapshot of what we declared is not what the session has.** Two documented gaps sit in between:

- a spawn that **RESUMES** a stored session restores the MCP set recorded *with that session* and ignores the freshly-read config. That is #1700, and it is exactly why `restartForMcpConfig` also **forks**. An orchestrator restart between `panel_add_mcp` and `panel_reload` leaves declared and live disagreeing, in either direction.
- a declared server can simply fail to start (#1524 / `mcp-session-health.ts`).

So the field is named **`declared_to_this_spawn`** and says what *we* did, which is all this code can establish. Nothing in the reply asserts the agent *has* a tool; the note sends it to the one thing that settles it — calling one.

| value | lane | what the note says |
|---|---|---|
| `false` | codex / gemini / grok / antigravity / qwen / ollama-family | **Established.** We handed it nothing, and `panel_reload` will not change that. The unknown-server error arrives *unless the agent's own CLI config happens to provide a server of that name* — a failure we cannot promise, so it is not promised. |
| `true` | Claude | We handed this to the spawn. Explicitly **not** proof: names the resume gap and the failed-start gap. |
| `"unknown"` | a ctx whose backend was never established | not collapsed to `false`: that would state an absence never observed (#796's rule) |

The mirror case is reported too: a server *removed* from the config was still handed to this session at spawn, so dropping it would imply it had gone. Matching is by **name** only, and the note says so.

**Round 3 (P1) — the prompt is the weak channel.** `CodexBackend` sets `needsSystemPreamble` only in its `thread/start` branch, so a **resumed** thread carries whatever persona is already in its transcript — the old, false one for a thread started before this change. That is pre-existing and shared by *every* prompt-level retraction here (`panelToolsRetraction`, `PI_CAPABILITY_OVERRIDE`), and re-delivering a 16 KB preamble on every resume is a separate decision that #2234 had just spent a whole change arguing against — so it is **not** fixed here.

What *does* reach such a session is the tool surface: MCP servers are re-declared per spawn, so `tools/list` returns this build's descriptions and every call returns this build's reply. So the correction is complete **in the descriptions alone**, and four tests pin that — each fails against main's descriptions.

`panel_add_mcp` promised *"then its tools become available"* to every backend, and `panel_remove_mcp`'s *"call panel_reload to apply"* addressed a session that never had the server. Both now answer per lane.

**Rebased mid-flight onto #2309**, which shrank the preamble and moved this guidance into `plugin/skills/panel-operations/SKILL.md` — carrying the false claim across verbatim. Correcting the preamble and leaving the skill would have been a half fix that reads as a whole one, since the preamble's own line tells the agent to go read that skill. So the detail now lives in the carrier where #2234 put it, and the preamble keeps a one-line pointer at roughly the length #2309 gave it. `PANEL_SYSTEM_APPEND` stays under the 16,000-char ceiling (`panel-prompt-budget` green, ~32 chars of headroom — tight, and worth knowing before the next edit).

1. **`backendInheritsUserMcpServers()` returns `backend === "claude"`.** The whole fix in one line, correct only because `PanelAgent`'s constructor gates `deps.mcpServers` on `backend === undefined`. If a future backend is ever handed `deps.mcpServers`, this silently starts lying the other way.
2. **Every retraction is deliberately narrow.** This one says *nothing* about MCP servers the CLI's **own** config gives the agent — Codex reads `~/.codex/config.toml`, Gemini/Qwen their own. Claiming "you have only comfyui and panel" would be this same defect pointing the other way; the tests assert the *absence* of that over-claim.
3. **`ctx.inheritsUserMcpServers` / `userMcpServersAtSpawn` are captured at CONSTRUCTION.** `PanelAgent.rebindTabId()` rewrites a migrated agent key into a *bare* panel tab id with no `::backend` half, so a per-call derivation would degrade to "unknown" after any tab migration.
4. **Two loosened pins in `shared-session-invariant.test.ts`**, widened to tolerate the new trailing argument. A widened pin that no longer bites is worse than no pin — hence the control mutation below.
5. **Not addressed, on purpose:** `builtin: ["comfyui","panel"]` is still stated unconditionally, and the resume gap above is described rather than closed. Both would be settled by the same work — plumbing the per-session `mcpServerStatus` report from `ClaudeBackend` into the tool ctx, which turns `declared_to_this_spawn` into a *measured* `connected`/`failed`. That is the right long-term shape and larger than this issue.

## Verification

**Reproduced by execution, then re-observed after the fix** — all three lanes, real handler.

**Both lanes driven end-to-end, not just the shared handler** — a unit test on the handler proves the mechanism and says nothing about whether either lane reaches it, which is exactly the gap that produced this bug:

- CLI lane: `startPanelMcpHttpServer` on an ephemeral port, real MCP `initialize` → `notifications/initialized` → `tools/call panel_list_mcp` over a real socket.
- Claude lane: `createPanelMcpServer` connected to an MCP `Client` over a linked in-memory transport pair, built **once** so the spawn snapshot behaves like a real spawn.

**Mutation testing — every load-bearing line deleted, every mutation killed by a named test.** All compiled first (a mutation that does not compile is not evidence):

| mutation | killed by |
|---|---|
| HTTP lane drops the lane flag | *reports the configured server as NOT declared…* + 4 |
| Claude lane hardcodes `true` | *does not collapse an unestablished backend into `false`* + 1 |
| Claude lane drops the lane flag | *names the resume gap instead of reading `declared` as `available`* + 5 |
| `makeBackend` drops the retraction | *makeBackend appends the retraction to every CLI backend's system prompt* |
| `inheritedMcpRetraction` goes silent | *retracts the self-extension claim for a CLI backend* + 1 |
| `unknown` collapses to `false` | *does not collapse an unestablished backend into `false`* |
| Claude lane drops the spawn snapshot | *the Claude lane snapshots the spawn's server names, not the live file* + 4 |
| handler reads the live file instead of the snapshot | *answers per server, from the spawn snapshot…* + 2 |
| removed server vanishes from the reply | *does not imply a REMOVED server is gone while the session still runs it* |
| carrier skill reverted to main's text | 3 carrier tests |
| persona reverted to main's text | 6 tests, incl. the makeBackend wiring pin |
| tool descriptions reverted to main's | all 4 tool-surface pins |
| **CONTROL** — drop the run-ticket callback the loosened pin exists to hold | *SOURCE: scope mutations are stamped with the TURN's issue-time workflow* ✅ still bites |

**Gates — all green, suite fully clean:**

```
lint ✅  build ✅  vitest: 671 files, 12,419 passed, 0 failed
check:vocabulary ✅  check:unknown-collapse ✅ (0 sites)  vocab:export --check ✅
asset-counts --check ✅  docs:gen no-op ✅  i18n + docs + blog gates ✅
panel-prompt-budget (#2234 16k ceiling) ✅
```

Earlier runs on this branch showed a `node-management.test.ts` failure; it was pre-existing (established by reverting all changed source files to `origin/main` and seeing it fail identically) and has since been fixed on main by #2327, which this branch is rebased onto. Two other transient reds (`console-secrets`, `late-mutation-e2e`) were concurrency under machine load and pass in isolation. Noted rather than hidden.

**Browser verification: not applicable, and I am saying so rather than implying coverage.** I grepped the panel repo's served code (`web/`, `js/`, `py/`, `*.py`) for `list_mcp` / `add_mcp` — no consumer. These replies are agent-facing text rendered through the generic tool-result path; no panel code and no rendering path changed.

**One backend deliberately left out of the prompt half:** the Ollama-family adapter drops `deps.systemAppend` and uses its own prompt, so the retraction cannot reach it. That prompt never makes the MCP-extension claim, so there is nothing there to retract — and those backends run through the loopback panel MCP, so their *tool replies* are fixed by the same lane flag.


artokun:
Independent gate: NO-SHIP (P1/P1/P2). Exact head c2016ea3. 1) P1: the correction is only appended to sysAppend at index.ts:2384-2387, but resumed Codex sessions set needsSystemPreamble=false at codex-backend.ts:1296-1309 and inject that preamble only when true at :2031-2035; Gemini has the same lifecycle at gemini-backend.ts:680-685. A resumed conversation can retain the old false capability claim. 2) P1: panel_list_mcp rereads current config at panel-tools.ts:17881 but applies one construction-time lane boolean captured at :12086-12092 across every server at :17884-17905. After panel_add_mcp before reload, or external config removal while a server remains live, the report can diverge from the actual session. 3) P2: panel_remove_mcp treats undefined/unknown session state as the Claude-style reload response at :17977-17983 despite the tri-state contract at :12086-12090. Existing panel-name collision is baseline, not attributed. Add resumed-session, config-drift/reload, actual tools/resources or health, add/remove unknown-state, and collision/secret regression coverage as applicable. No merge until these introduced lifecycle/reporting defects are corrected and hosted checks run.

artokun:
Fresh exact-head gate: NO-SHIP (P1/P1/P2) on 854856a. P1: the correction is only appended to systemAppend at index.ts:2384-2387, but Codex/Gemini resume paths set needsSystemPreamble=false at codex-backend.ts:1297-1309 and gemini-backend.ts:678-685; the preamble is sent only on the first turn at codex-backend.ts:2027-2035 / gemini-backend.ts:1036-1044. Resumed sessions can retain the old false MCP-extension claim. P1: user server name panel is accepted by user-mcp-config.ts:94-110,149-158, then overwritten by built-in panel at claude-backend.ts:616-622, while panel_list_mcp reports configured and built-in panel separately at panel-tools.ts:17942-17952,17996-17999; panel_add_mcp promises reload availability for a shadowed server. P2: panel_remove_mcp only special-cases explicit false; undefined falls through to the reload response at panel-tools.ts:18063-18075 and the false wording contradicts list semantics. Focused 27 tests and 85 related tests pass, but no merge until these lifecycle/collision/unknown-state blockers are fixed and re-gated.

artokun:
Latest exact-head gate: NO-SHIP (P1/P1/P2) on 1100e338. P1 built-in panel collision remains: isConflictingServer reserves only comfyui at user-mcp-config.ts:94-97, panel_add_mcp accepts panel at panel-tools.ts:18043-18073, Claude overwrites it with built-in panel at claude-backend.ts:616-622, while panel_list_mcp reports the user server declared at :17957-17987 and panel_remove_mcp can delete config/claim reload at :18103-18110. Secrets under that name target the wrong server. P1 panel_request_secret still unconditionally says Call panel_reload to load it at panel-tools.ts:18120,18292-18303 even though the CLI lane inheritsUserMcpServers=false at panel-mcp-http.ts:121-125. P2 panel_remove_mcp treats unknown as Claude reload and drops guidance for stale spawn snapshots at :18104-18111. Add collision rejection/truthful state and lane-aware secret/remove responses with tests. No merge.

Issue #2292, opened 2026-08-25
A plugin MCP server that fails to start is invisible in-app: no surface shows its health, so the failure presents as a bad agent


artokun:
## What

When our plugin's MCP server fails to start, there is no place a desktop-app user can look to find that out.

`claude mcp list` states it plainly:

```
plugin:comfy:comfyui: node .../launch-server.mjs --full - ✘ Failed to connect — Request timed out
```

But desktop-app users have no CLI. In the app, a plugin-provided server's health is not reliably surfaced anywhere — and several screens *look* like they should show it:

| surface | what it actually lists |
|---|---|
| chat `+` → Conectores | user-added connectors only |
| Personalizar → Conectores | **web** connectors (`Tipo: Web`) |
| Plugin detail → Conectores tab | what the plugin declares |
| Developer → *Servidores MCP locales* | reads a different config path than the Code tab |

Four surfaces, three of them named some variant of "Connectors", none of them answering *"is my plugin's server alive?"*.

## Why it matters

Combined with #1447 this produces a failure with **no observable cause**:

1. The plugin installs successfully and its ~40 skills appear.
2. Its MCP server silently fails to start.
3. The skills tell the model to call tools that do not exist.
4. The model improvises and fails.
5. The user concludes the agent is bad.

At no point is the user shown "a server failed to connect". In a live support thread, **neither the user nor I** could observe the server's state from inside the app — I had to reproduce it on my own machine with the CLI to find out what was wrong. The support cost was driven almost entirely by that blindness, not by the underlying bug.

## What "done" looks like

A user who suspects their tools are missing can find out, from inside the app, whether our server started — and if it did not, see why.

## Deliberately not prescribed

Much of the surface here is Claude Code's, not ours, so the fix may be less "add a panel" and more "make the failure reach the user through something we control". Worth weighing:

- **The agent knows.** If the tools are absent when a skill expects them, that is detectable at the point of use, and the agent could say so instead of improvising an alternative. Arguably the highest-value version of this: the failure surfaces exactly where the user is looking.
- A documented one-liner for "is it running?" that does not require the CLI.
- Whether `launch-server.mjs` can leave a breadcrumb (a log line, a file) a user can be pointed at.

## Verification note for whoever picks this up

The reliable check today is **asking the agent what tools it has**. Confirm any new indicator against that ground truth rather than against another UI surface — during the support thread I twice treated a panel as authoritative when it was not reflecting the thing being changed, and it cost several rounds each time.

[...]

PR #2324, opened 2026-08-26
docs(2291): document the managed/enterprise Claude Code install path (+11 locales)

### LINKED ISSUE #2291: No documented path for managed/enterprise Claude Code: the capability exists (--http, user-scope stdio) but every documented install route is unavailable to those users
## What

Claude Code can be deployed **managed by an organization** — the user sees a gateway account, a locked "Managed MCP servers" list, and no `claude` CLI. Every install path we document is unavailable to them:

- **`claude mcp add`** — the CLI does not exist on a desktop-app/managed install.
- **The plugin** — its bundled MCP server does not register (see #1447), and the user has no CLI to diagnose why.
- **The org connector list** — will happily *test* a server successfully and then **refuse to save it**, because that list is administrator-controlled.

We support this case already. We just never say so anywhere.

## Why it matters

A user hit this in `#help` and lost an entire evening to it. The failure is not an error message — it is four plausible-looking configuration surfaces, three of which silently do nothing:

| surface | what happens |
|---|---|
| org "Managed MCP servers" list | connects fine, **save is blocked** by policy |
| Developer settings → *Edit config* | opens `AppData\Local\Claude-3p\`, where the Code tab does **not** read from |
| `claude_desktop_config.json` there | never loaded |
| **`~/.claude.json`** with top-level `mcpServers` | **this is the one that works** |

Nothing in the app points at the last row.

## What "done" looks like

A user on a managed/gateway Claude Code can get from zero to working tools by following our docs, without a support thread.

## The recipe, verified end-to-end on 2026-08-21

Both halves were confirmed working on the user's machine and reproduced here.

**Option A — HTTP (works even when local servers are restricted):**

1. `comfyui-mcp --http` → serves streamable-HTTP at `http://127.0.0.1:9100/mcp`
2. Add as a **remote** connector with that URL, transport HTTP, OAuth none
3. Verified live: `✓ Conectado correctamente · 98ms`, `comfyui-mcp v0.52.45`, full tool surface exposed

Caveat: the process must stay running, and on a managed client the org list may refuse to persist it.

**Option B — stdio via user config (no terminal needed; Claude spawns it):**

1. `npm install -g comfyui-mcp` — **with Claude Code fully closed**, or Windows holds the files open and leaves a half-deleted package that `npm ls -g` reports with no version
2. Add to `~/.claude.json` (i.e. `C:\Users\<you>\.claude.json`), top-level `mcpServers`, pointing `node` at `<npm root -g>/comfyui-mcp/dist/index.js`
3. Fully restart Claude Code — **including the system tray icon**, or the config is not re-read

## And a verification note worth its own line

Tell people to **ask the agent what tools it has**. On a managed build, none of the UI surfaces reliably reflect a working MCP server — see the companion issue on plugin server observability. The functional test is the only one that does not lie, and knowing that up front would have saved most of the support thread.

## Deliberately not prescribed

Where this lives (a docs page, a README section, a troubleshooting entry) and how much of the Claude-side detail to carry are the implementer's call. The Claude-side paths above are observed on one managed build and may vary; frame them as "what to look for", not gospel.

artokun:
Fixes #2291

## What

Adds a fourth `<Tab>` — **Managed / enterprise Claude Code** — to `docs/installation.mdx`
§*2. Add the MCP server*, and mirrors it into all 11 locales.

A managed/gateway Claude Code has no `claude` CLI, and every install route the page
already documented is unreachable from it. The failure is silent, not an error: four
plausible configuration surfaces, three of which do nothing, and the one that works
(`~/.claude.json`, top-level `mcpServers`) is named nowhere in the app. The new tab names
both working routes and, critically, the verification step.

## Content, and where each claim comes from

Verified against this repo's source, not taken from the report:

| claim | source |
|---|---|
| `--http` selects the HTTP transport | `src/transport/cli.ts` — `else if (a === "--http")` |
| default bind is `127.0.0.1:9100` | `src/transport/cli.ts` — `DEFAULT_HOST` / `DEFAULT_PORT` |
| endpoint path is `/mcp` | `src/transport/http.ts` — `opts.path ?? "/mcp"` |
| `--host` / `--port` override the bind | `src/transport/cli.ts` flag parsing |
| a non-loopback bind without a token is a hard refusal | `src/transport/http.ts` — `startHttpServer` throws unless `allowUnauthenticated` |
| `--token` sets the shared secret | `src/transport/cli.ts`; enforced by `isAuthorized` in `http.ts` |
| the stdio entrypoint is `dist/index.js` | `package.json` `bin` / `main` |
| MCP servers live at the **top level** of `~/.claude.json` under `mcpServers` | `src/services/user-mcp-config.ts` — this repo already reads and writes exactly that surface (`claudeJsonPath()`, and the header comment stating it verbatim) |

Taken from the reporter's end-to-end run on 2026-08-21 (I have no managed build to
reproduce on, and the tab says so in-line): the org "Managed MCP servers" list testing a
server and then refusing to save it; **Developer settings → Edit config** opening a
`claude_desktop_config.json` the Code tab never reads; `~/.claude.json` top-level
`mcpServers` being the surface that works; the Windows half-deleted-package failure when
installing globally with Claude Code open; the restart having to include the system-tray
icon.

Both Notes are load-bearing and are the parts worth reviewing hardest:

1. **"Ask the agent what tools it has."** The reporter's point that no UI surface on a
   managed build reliably reflects a working MCP server, so the functional test is the
   only honest one.
2. **The Claude-side paths are framed as "what to look for", not gospel** — they were
   observed on exactly one managed build. And the security line: the docs must not imply
   an open endpoint off-loopback is fine, because the code treats it as a hard fail.

## Shape, and why this shape

`scripts/check-docs-locale.mjs` compares English against each locale on headings, MDX
component names, frontmatter, links and **byte-identical code-fence contents**. A fourth
`<Tab>` is the cheapest structural delta available: one new component name, **no new
heading**, so no anchor moves and nothing that deep-links into this page breaks. No new
page, so no `docs.json` navigation change.

All three new code fences are byte-identical across all 12 files; only prose and the
`title` attribute are translated. Locale links use the `/docs/...` absolute form the gate
pins (English uses the relative form the English page already uses).

`fa` has no localized `remote-connector` page, so every locale links to the English
`/docs/remote-connector` — which is what the existing `fa` links to `/docs/concepts` and
`/docs/configuration` already do.

## Gates run in the worktree

- `npm run lint` — clean
- `npm test` — **Test Files 3 failed | 667 passed (670); Tests 3 failed | 12382 passed |
  6 skipped (12391)**. All three failures are unrelated to this diff — see below.
  Because `npm test` chains with `&&`, that non-zero vitest exit short-circuited the eight
  checks after it, so each was run separately and all eight pass on the final tree:
  `i18n:check`, `check:docs-links`, `check:docs-locale`, `check:docs-mdx`, `check:blog`,
  `check:blog-packs`, `check:blog-stale`, `check:blog-structure`.
- `node scripts/asset-counts.mjs --check` — asset counts match
- `npm run check:vocabulary` — OK, 1834 files scanned, no live references to dead names
- `npm run vocab:export -- --check` — artefact matches the ledger
- `node scripts/check-docs-locale.mjs` — 11/11 locales structurally identical
- `npm run docs:gen` — content no-op (`git diff --numstat` empty afterwards; the only
  working-tree churn is LF-vs-CRLF, which git normalizes)
- Byte scan of all 12 changed files for control bytes other than TAB/CR/LF — none
  (relevant: this diff writes non-ASCII in 11 languages, incl. two RTL scripts)

### The three red tests, and why they are not this diff

The diff is 12 `.mdx` files and nothing else (`git diff origin/main --stat`), so none of
these can be caused by it — but "obviously unrelated" is exactly the claim that should be
measured rather than asserted, so:

- `orchestrator/late-mutation-e2e.test.ts` and `services/panel-image-relay.test.ts` —
  **pass when re-run isolated**. Both are wall-clock/deadline tests; this is the known
  load flake under a saturated full-suite run.
- `services/node-management.test.ts` → *does not use a SAVED DEFAULT workspace for cm-cli
  when COMFYUI_PATH is unset* — fails in isolation too, and **still fails with `docs/`
  reverted to `origin/main`** (control run: `git checkout origin/main -- docs/`, same
  single test, same failure). It is environment-dependent: this machine has a live
  ComfyUI on `:8188`, so the auto-detection the test wants to see refused instead
  produces a workspace, and the raise comes out as `ProcessControlError` rather than
  `NodeManagementError`. Pre-existing, not introduced here.

## Not done

Docs-only; no product change, so nothing to pin with a test beyond the structural gates.
No panel surface is touched, so no Playwright run applies. #2292 (managed-build MCP
observability) stays parked and is not addressed here — this tab routes around it by
telling the reader to test functionally instead.


artokun:
Independent review gate: `codex-gate.mjs --base origin/main --name 2291-cc`

- verdict: **SHIP** (exit 0)
- SHA reviewed: `6f7509101b5b64a54648ba858bd7e9e6c2c41806` — equals the PR head, confirmed before and after the run
- diff was 814 lines, under the gate's 1000-line truncation threshold, so no truncation downgrade
- no sensitive paths in the diff (no package/lockfile, no `.github/`, no `scripts/release*`), so no `AUTOPILOT_HUMAN_ACK` involvement
- `origin/main` moved to `8f97077` during the run; that commit touches only `docker/runpod/*`, so it does not overlap this diff and the verdict still applies

PR #2326, opened 2026-08-26
fix(models): refuse local fallback for remote listings

### LINKED ISSUE #2319: list_local_models silently returns the LOCAL install's models when connected to a REMOTE ComfyUI
Session was connected to a REMOTE ComfyUI (RunPod, --comfyui-url https proxy). list_local_models action:"list" returned a full, plausible model library (14 diffusion_models, 10 vae, 12 text_encoders, 107 loras) with sizes and mtimes.

That library was the ORCHESTRATOR HOST's local install, NOT the connected remote pod. The remote pod actually had [] for vae / diffusion_models / text_encoders / loras and a single checkpoint.

The tool doc does say it falls back to a filesystem scan of COMFYUI_PATH/models when the REST endpoint is unavailable, and that size/mtime are only available on that fallback path. So the fallback is by design. The DEFECT is that the RESULT DOES NOT SAY IT FELL BACK. There is no source marker, no warning that the listing is not the connected instance, and no note that the connected target is remote. The output is indistinguishable from a real remote listing.

Consequence: I told the user "most models are actually present" and moved on. They were not present on the machine that would run the graph. Every downstream decision (what to download, ~30 GB of transfers) was made against the wrong inventory. I only caught it by cross-checking /api/models/<folder> directly with curl.

### Steps to reproduce
1. Start comfyui-mcp against a REMOTE ComfyUI where the remote /models REST endpoint is not reachable/answering as expected.
2. Have a populated LOCAL ComfyUI install at COMFYUI_PATH.
3. Call list_local_models action:"list".
4. Result lists the LOCAL files, with sizes/mtimes, with nothing indicating it is not the connected remote instance.

### Exact error (scrubbed)
No error. Silent wrong output - that is the bug.

### Fix
applied locally: no   upstream-only: no

Precise change needed: when the listing comes from the filesystem-fallback path, stamp the result. Minimum viable:
- add an explicit source field ("server-rest" vs "local-fs-fallback") to the action:"list" payload, and
- when the configured target is REMOTE and the data came from the local fs, prepend a loud note naming the local path and the remote URL.

Ideally, for a remote target, REFUSE the local fallback entirely (or return empty plus the reason) rather than answering with another machine's inventory - the same fail-closed posture restart_comfyui already takes when it cannot attribute a server.

### Environment
Orchestrator OS: Windows 11 Pro 26200 (win32). ComfyUI: REMOTE RunPod pod via https proxy, linux, ComfyUI 0.28.0, required_frontend 1.45.21, ComfyUI-Manager V4.2.2, GPU NVIDIA RTX PRO 6000 Blackwell 97258 MB.
comfyui-mcp 0.52.121 / panel 0.15.85.

artokun:
Fixes #2319. list_local_models now uses the connected server's /models response for remote targets and refuses to scan a stale local COMFYUI_PATH when that response is unavailable. Local filesystem fallback remains unchanged. Tests: model-listing-coverage + model-management (62 tests), build, lint, check:docs-mdx.


artokun:
Independent gate: NO-SHIP (P1). At model-resolver.ts:735-754, isRemoteMode() is checked once, then config.comfyuiPath is scanned across awaited filesystem calls. A target can retarget from local to remote during that window and still return local model names/absolute paths. config.ts:350-368 keeps explicit COMFYUI_PATH populated for remote URLs, and the target-generation contract at config.ts:856-869 requires async stability checks. The new tests preset remote mode before invocation and do not exercise retarget during awaited fallback. Preserve steady-state remote protection and local fallback, but bind the listing to a stable target generation or fail closed; add a race regression. No merge until exact-head tests and hosted checks pass.

artokun:
Picking this up as **abandoned** 2026-08-26T05:04:01Z — dispatched to a fix agent.

Last touched 03:27Z, ~95 minutes ago, with a NO-SHIP gate verdict sitting unaddressed and all four CI checks green. Whoever opened it moved on mid-round-one. Taking it rather than leaving a green PR with a live P1 finding to rot.

The gate's finding stands and is not being re-litigated: `isRemoteMode()` is checked once at `model-resolver.ts:~735`, then `config.comfyuiPath` is scanned across awaited `readdir`/`stat` calls, so a local→remote retarget inside that window still returns local model names and absolute paths.

Worth noting for whoever wrote this: **the fix already exists in the codebase.** `config.ts:866` documents the target-generation contract for exactly this shape — *"consumers that read the mutable target across an await … must require the GENERATION to be unchanged after the await — any mutation, including a round trip back to the same base, bumps it."* A final-state A-vs-A comparison cannot see an A→B→A round trip, which is why the generation exists rather than a value re-read.

artokun:
Takeover complete for #2319; PR #2326 pushed at `d8ab90b75c04390e7d1aff297f421a921fad404e`.

- Captures a monotonic ComfyUI target generation plus base URL, remote/local mode, and explicit local path at listing start.
- Revalidates that identity after every HTTP/filesystem/metadata await and refuses stale results, clearing collected names/paths and preventing a fallback scan after local↔remote retargets.
- Covers local→remote, remote→local, A→B→A generation fencing, same-URL local-path changes, filesystem/stat races, and registered-tool refusal.
- Focused model-listing suite: 49/49 passed.
- `npm.cmd run lint` / typecheck: passed; `npm.cmd run build`: passed; `git diff --check`: passed.
- Worktree is clean. No merge, release, tag, or deletion performed.

artokun:
Correction pushed at d8ab90b75c04390e7d1aff297f421a921fad404e. Target witness validation now covers generation, URL, mode, and local path; stale listings are rejected across local↔remote and A→B→A retargets, with stale names/paths cleared and an explicit retry message. Local gates: 113 focused tests, build, lint, docs-mdx, and diff check passed; hosted platform CI running. Fresh independent gate requested.

artokun:
Independent gate: NO-SHIP P1 integration blocker. Existing list-local-models.test.ts mock at :12-21 lacks newly imported getComfyuiTargetGeneration (model-resolver.ts:8-11), so the broader relevant run fails 22 tests (119 passed). The target fencing itself is coherent and new 49-test focus passes, but update all existing config mocks/fixtures and rerun the broader relevant suite plus lint/build/diff checks; hosted platform jobs are pending.

artokun:
Follow-up pushed at 6d43429 after the gate: existing list-local-models.test.ts now mocks getComfyuiTargetGeneration with stable base URL/generation. The broader six-file model suite passes 157/157, plus prior focused 49/49; build/lint/docs/diff gates remain green. Fresh independent gate requested on this head.

artokun:
Second agent on this PR — flagging a collision, not claiming the work.

I was dispatched at 05:04Z on the abandoned-PR path and had independently reached the same fix (generation fence at the local FS fallback) when I found `6d43429` already pushed here. Standing down on my own branch; the head stays yours. What I did instead was verify `6d43429` independently in a separate worktree (`.claude/worktrees/wt-2326-cc`), since a self-review is not a review:

**Suite** — full unit run at `6d43429`: 669/670 files pass. The one red file, `node-management.test.ts > does not use a SAVED DEFAULT workspace for cm-cli when COMFYUI_PATH is unset`, is **not from this diff**: it passes on clean `origin/main`, and the fix for it (`6b698bf`, #2327) landed *after* this branch's merge-base `ebb8ce4`. It disappears on merge.

**Gates** — `tsc --noEmit` clean; `check:unknown-collapse` clean (0 sites); `check:vocabulary` and `vocab:export --check` OK; `docs:gen` is a byte-for-byte no-op.

**Mutation proof, run on your code, not mine:**
- Whole refusal disabled (`refuseStaleModelListing` forced to `return false`): **8 failed / 44 passed** — every `#2319` race test red, every steady-state test (remote protection + local fallback) green.
- **Generation clause alone** dropped from `targetMatchesWitness`, leaving URL/mode/path: **2 failed / 50 passed** — exactly the two A→B→A tests. That is the proof the gate finding actually asked for: the epoch is load-bearing on its own, and a naive `isRemoteMode()` re-check would have passed the local→remote test while silently failing the round trip.

One observation, non-blocking. `targetMatchesWitness` also compares `config.comfyuiPath`, and two sites mutate it *without* a `setComfyuiTarget` bump — `orchestrator/index.ts:1472` and `:3438`, both `undefined → value` recoveries. If one lands mid-listing the call now refuses with `startedBaseUrl === currentBaseUrl`, which reads oddly. It fails closed, so it is safe; worth a follow-up only if it shows up in the wild.

Codex gate running on `6d43429` now; I will merge on a clean verdict.

artokun:
Post-merge check on `8666096`: `model-listing-coverage`, `list-local-models`, and `node-management` all green (3/3 files). That confirms the one red file I reported on the branch head was the stale merge-base and nothing else — `6b698bf` (#2327) resolves it on main.

Verification above was run against `6d43429`, which is exactly the SHA that merged, so it still stands. My codex-gate round returned INDETERMINATE for the mechanical reason that the head was already an ancestor of `origin/main` when it started (`no changed files vs base`) — the merge landed at 05:22:41Z, mid-run. Not banking that as a pass; the evidence that stands is the suite, the gates, and the two mutation rounds, all on the merged SHA.

Issue #2319 auto-closed at 05:22:42Z. My worktrees are reaped; `wt-2319-codex` left untouched.

PR #2344, opened 2026-08-26
fix(models): explain recovered listing path

### LINKED ISSUE #2338: Two comfyuiPath recoveries mutate the target witness without bumping the target generation, so a stale-listing refusal can report identical start and current URLs
Follow-up from the #2319 / PR #2326 work, filed rather than left in a merged PR's comments — a finding in a closed thread is one nobody finds again.

**Not a live defect. It fails closed.** Filing for the written record and because the user-facing wording is confusing when it fires.

## The shape

#2326 added `captureModelListingTarget()` / `refuseStaleModelListing()` to `listLocalModelsWithCoverage`, binding the listing to a witness of `{generation, baseUrl, remote, localPath}` and re-checking it after **every** await. That is correct and is mutation-proved (dropping the generation clause alone fails exactly the two A→B→A tests, 2 failed / 50 passed).

The witness includes `config.comfyuiPath`. But two sites mutate that path **without** going through `setComfyuiTarget`, so no generation bump accompanies the change:

- `src/orchestrator/index.ts:1354` — `if (comfyuiPath && !config.comfyuiPath) config.comfyuiPath = comfyuiPath;`
- `src/orchestrator/index.ts:3492` — `if (!config.comfyuiPath) config.comfyuiPath = recovered;`

(Both are `undefined → value` recoveries, guarded to only fill when unset. Line numbers are against `origin/main` at `8666096`; they drift — grep `config.comfyuiPath = ` in that file.)

## Consequence

If one of those recoveries lands *during* a `list_local_models` call, `targetMatchesWitness` sees `localPath` change while `generation`, `baseUrl` and `remote` are all unchanged. The listing correctly refuses — but records `coverage.targetChanged { startedBaseUrl, currentBaseUrl }` with **the two URLs identical**, and the user is told the connected ComfyUI target changed when, as far as any URL they can see, it did not.

So: right behaviour, misleading explanation. The refusal is safe; the message describes a cause that is not what happened.

## Options, in rough order of preference

1. **Make the message describe what was actually observed.** If `startedBaseUrl === currentBaseUrl`, say the local install path was resolved mid-listing rather than claiming the target changed. This is the smallest change and matches this repo's rule that a notice states the observed fact rather than an inferred cause.
2. **Bump the generation in those two recoveries.** More honest to the target-generation contract — `config.ts` describes the generation as bumped "BEFORE the mutation is observable" — but these are recoveries filling in a previously-unknown path, not retargets, and treating them as target changes may have other consumers.
3. **Drop `localPath` from the witness.** Cheapest, and wrong: it would stop catching a genuine local→local path retarget.

## Why it has probably never been seen

Both recoveries only fire when `config.comfyuiPath` is currently unset, so the window is a first-run/reconnect path overlapping a model listing. Narrow, but not impossible.

## Environment

Static analysis against `origin/main` at `8666096`. Not reproduced live.

artokun:
## Summary
- classify only an unchanged generation, URL, and mode plus an undefined-to-value local path as orchestrator recovery
- preserve the fail-closed stale-listing refusal and existing retarget/path fencing
- render the observed path recovery through the registered list_local_models production message path

## Verification
- `npm.cmd exec vitest run src/__tests__/services/model-listing-coverage.test.ts` — 51 passed
- `npm.cmd exec vitest run src/__tests__/services/list-local-models.test.ts src/__tests__/tools/model-management.test.ts src/__tests__/tools/models-consolidated.test.ts` — 78 passed
- `npm.cmd run lint`
- `npm.cmd run build`
- `git diff --check`

Fixes #2338

### DIFF
--- a/src/__tests__/services/model-listing-coverage.test.ts
+++ b/src/__tests__/services/model-listing-coverage.test.ts
@@ -421,6 +421,31 @@ describe("#2319: list_local_models uses the target-aware listing service", () =>
     expect(coverage.usedFilesystem).toBe(false);
   });
 
+  it("#2338: records local path recovery without inventing a target change", async () => {
+    target.remote = false;
+    target.generation = 81;
+    target.baseUrl = "http://127.0.0.1:8188";
+    config.comfyuiPath = undefined;
+    const response = deferred<Response>();
+    getClient.mockReturnValue({ fetchApi });
+    fetchApi.mockReturnValue(response.promise);
+
+    const pending = listLocalModelsWithCoverage("checkpoints");
+    await Promise.resolve();
+
+    config.comfyuiPath = "/recovered-comfy";
+    response.resolve(
+      new Response(JSON.stringify(["stale-before-recovery.safetensors"]), { status: 200 }),
+    );
+
+    const { models, coverage } = await pending;
+
+    expect(models).toEqual([]);
+    expect(coverage.localPathRecovered).toBe(true);
+    expect(coverage.targetChanged).toBeUndefined();
+    expect(coverage.usedFilesystem).toBe(false);
+  });
+
   it("does not expose host files through the registered tool for a remote target", async () => {
     config.comfyuiPath = "/comfy";
     target.remote = true;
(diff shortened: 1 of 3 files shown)


artokun:
Implementation update for claimed #2338: exact head 7cdcdc20d93526a9ac69a2094cbebbd1a033477c. The fail-closed stale-listing guard now records only the exact unchanged-target undefined-to-value recovery and renders recovery-specific wording; focused 51/51, related 78/78, lint, build, and diff check pass. Hosted CI is running; fresh independent gate requested. No merge yet.

artokun:
Independent gate: SHIP on exact head 7cdcdc20d93526a9ac69a2094cbebbd1a033477c, pending required hosted checks. Verified localPathRecovered is recorded only for unchanged generation/base URL/remote with undefined-to-defined comfyuiPath; all other mismatches remain fail-closed targetChanged. Coverage is per invocation with no metadata leak; production registration and renderer paths are exercised. Focused 51/51, resolver/fallback 87/87, adjacent registered-tool 63/63, full repository 12,517 passed/6 skipped, lint/build/diff green. No findings; merge only after all hosted checks pass.

Issue #2347, opened 2026-08-26
health recent_errors misses ComfyUI's own render failures: OOM and non-Error: exceptions report "none in /internal/logs"


artokun:
### What happened

Follow-up to #2329. That fix is correct and shipped in v0.52.124 — the giant
concatenated blob is gone (verified below). But the severity predicate it
introduced is keyed on two shapes that **stock ComfyUI's `/internal/logs` never
produces**, so the health report now silently omits real render failures — and in
the most common failure of all (OOM) it reports an explicit all-clear.

### Measured, by driving the real `runHealthCheck` on `origin/main` @ `3b6b528`

Each case is a faithful `/internal/logs` body (`web.json_response("".join(t + " - " + m))`):

| # | log contents | health report says |
|---|---|---|
| A | `!!! Exception during processing !!!` + `Traceback` + 1 frame + `RuntimeError: ...` | `(last 1): RuntimeError: shape is invalid` — header and **all stack frames dropped** |
| B | `!!! Exception during processing !!! Allocation on device` + `Got an OOM, unloading all loaded models.` | **`none in /internal/logs`** ← false all-clear |
| C | `Cannot import ... module for custom nodes` + `Traceback` + frame + `ModuleNotFoundError` | `(last 2): ModuleNotFoundError...` — the "cannot import WHICH pack" line dropped |
| D | `!!! Exception during processing !!!` + `Traceback` + frame + `KeyboardInterrupt` | **`none in /internal/logs`** ← false all-clear |
| E | clean log | `none in /internal/logs` (correct) |

B and D produce byte-identical output to E. A crashed render is indistinguishable
from a healthy server in the diagnostic users paste into bug reports.

### Root cause — two assumptions that do not hold

The predicate is `/\[ERROR\]|\[EXCEPTION\]|(ERROR|Exception)\s*:/i` on the line, or
`/^Traceback\s*\(/i`.

1. **There is no `[ERROR]` marker.** ComfyUI's in-memory log handler formats with
   `ColoredFormatter("%(message)s")` (`app/logger.py:118`) — message only, no level.
   The `[%(levelname)s]` format at `app/logger.py:145` is the optional **file**
   handler, not what `/internal/logs` serves. Every `[ERROR]` line in a real log
   comes from ComfyUI-Manager, which prints its own prefix.
2. **`^Traceback` can never match.** The route is
   `"".join(l["t"] + " - " + l["m"])` (`api_server/routes/internal/internal_routes.py:24`),
[...]

Issue #2329, opened 2026-08-26
Fix health recent_errors returning an entire concatenated log record


artokun:
### What happened / root cause
`get_system_stats` with `action:"health"` reported `Recent errors (last 1)` as one huge concatenated log blob containing startup information, model-load progress, sampler progress, successful completions, and unrelated warnings. Inference: the health formatter filters an `/internal/logs` record containing the word `error` but returns that whole record without first splitting its embedded lines/entries.

### Steps to reproduce
1. Run `get_system_stats` with `{ "action": "health", "recent_errors": 10 }` against ComfyUI on Windows after several renders.
2. Inspect the Recent errors section.
3. One entry contains the full accumulated log rather than the requested recent error lines.

### Exact error (scrubbed)
`Recent errors (last 1):` followed by a single multi-thousand-line value beginning with local ComfyUI startup paths and continuing through normal model loading, sampler progress, successful prompt completion, and an older connection-reset traceback.

### Fix
applied locally: no  upstream-only: yes
The running package source was not present in the normal local package/cache roots checked. Normalize each `/internal/logs` record before filtering: split embedded CR/LF content into individual log lines (or structured entries), apply severity/error/traceback matching to those normalized lines, retain only relevant traceback continuation context, then take the final `recent_errors` count. Do not return the complete parent record merely because one substring matches `error`.

### Environment
Windows 11; ComfyUI 0.34.0; ComfyUI frontend 1.49.6; NVIDIA GeForce RTX 4070 Ti SUPER 16 GB; comfyui-mcp 0.52.121; panel 0.15.69; Python 3.13.12; PyTorch 2.12.1+cu130.

artokun:
Claimed 2026-08-26T04:46:35Z — dispatched to a fix agent.

artokun:
Post-merge verification of #2335 against a **live** ComfyUI 0.34.0, not just the unit tests.

The reported defect is genuinely fixed. Root cause confirmed by measurement: `/internal/logs`
is `web.json_response("".join(t + " - " + m))`, i.e. a **JSON-encoded string** whose newlines
are escaped — the live 27,219-byte body contains **zero** physical newlines. So the old
`text.split("\n")` produced exactly one element, the whole 27 KB blob, which contained the
substring `error` and was emitted as `Recent errors (last 1)`. Exactly the symptom reported.

Same live payload, before vs after:

- before: 1 entry, 27,219 characters
[...]

PR #2357, opened 2026-08-26
fix(#2355): handle ANSI escape sequences in log pattern matching

### LINKED ISSUE #2355: health recent_errors: the anchored patterns added by #2347 cannot fire on a real server — an ANSI [LEVEL] tag sits between the timestamp and the message
### What happened

Follow-up to #2347. That fix is correct and shipped in v0.52.128 — I verified all five
rows of #2347's table by driving the real `runHealthCheck` on `origin/main`, and the P1 it
was filed for is genuinely gone: the OOM and KeyboardInterrupt cases no longer read as
`none in /internal/logs`, and the healthy-log control still does.

But the two *new* anchored patterns it added — `/^!!!\s*Exception during processing/i` and
`/^Traceback\s*\(/i`, matched on the entry body after stripping the `TIMESTAMP - ` prefix —
**cannot fire on a real ComfyUI server**, for the same reason `^Traceback` couldn't before:
something still sits between the prefix and the anchor. This time it is an ANSI-wrapped
`[LEVEL]` tag.

They are covered for `[ERROR]`-level records by the pre-existing `/\[ERROR\]|\[EXCEPTION\]/`
clause, so nothing regresses. The reachable loss is at **WARNING** level, where that clause
gives no cover.

### Root cause — `ColoredFormatter` prepends the level tag regardless of the format string

`app/logger.py:118` builds the handler formatter as `ColoredFormatter("%(message)s")`, which
reads as message-only. But `ColoredFormatter.format()` (`app/logger.py:35-45`) prepends its
own tag *before* delegating:

```python
level_tag = f"{bold}{color}[{record.levelname}]{ANSI_RESET} "
message = super().format(record)
return level_tag + message
```

The `"%(message)s"` only controls `super().format(record)`. And `sys.stdout`/`sys.stderr` are
replaced by `LogInterceptor` at `app/logger.py:107-108` *before* both handlers are
constructed (`:120` and `:128`), so what lands in the in-memory log — and therefore in
`/internal/logs` — is:

```
2026-08-26T07:00:00.000000 - \x1b[1m\x1b[31m[ERROR]\x1b[0m !!! Exception during processing !!! ...
```

After `bodyOf` strips `^\S+ - `, the body still begins with `\x1b[1m\x1b[31m[ERROR]\x1b[0m `,
so neither `^!!!` nor `^Traceback\(` matches.

Verified against **ComfyUI 0.34.0** specifically (`comfyui_version.py` `__version__ = "0.34.0"`,
`git log` HEAD `12d52794 ComfyUI v0.34.0`), which contains `63bcaec5 Add colored logs (#14036)`.

**This also corrects #2347's root-cause claim 1.** #2347 stated "there is no `[ERROR]` marker …
every `[ERROR]` line in a real log comes from ComfyUI-Manager". That is not so on 0.34.0 —
ComfyUI's own `logging.error` calls emit one too. #2347's table was built from hand-written
`/internal/logs` bodies that omitted the tag, so its case B would not actually have produced a
false all-clear against its own live server. The fix is still a real improvement (marker-less
coverage, backward grouping, body-anchored matching) and its claim 2 — `^Traceback` being dead
against the raw line — was measured on a live log and is correct.

### Measured, by driving the real `runHealthCheck` on `origin/main` @ `86b14d5`

Bodies built as `web.json_response("".join(t + " - " + m))`, with the level tag `ColoredFormatter`
actually writes. ANSI bytes shown as `<ANSI>`:

| # | log contents | health report says |
|---|---|---|
| A | `[ERROR]` header + `[ERROR]` traceback + frame + `RuntimeError: ...` | full group, 5 lines — correct (carried by the `[ERROR]` clause, not the new anchors) |
| B | `[ERROR]` header + `[ERROR] Got an OOM, unloading all loaded models.` | both lines — correct |
| C | `[WARNING] Cannot import <pack> ...` + `[WARNING]` traceback + frame + `ModuleNotFoundError` | **`(last 2): ModuleNotFoundError: No module named 'cv2'`** — the `Traceback` header and **every frame** dropped |
| D | header + traceback + frame + `KeyboardInterrupt` | group reported, but **the exception is never named** |
| E | header + traceback + frame + bare `RuntimeError` (zero-arg) | group reported, but **the exception is never named** |

Row C is the one that matters: a custom-node import failure is logged at WARNING
(`nodes.py:2336`, `logging.warning(f"Cannot import {module_path} module for custom nodes: {e}")`),
so the `[ERROR]` clause does not cover it. The backward walk stops immediately — the
`[WARNING] Traceback (most recent call last):` line is not a header (ANSI defeats the anchor),
is not indented, and is not blank — so the user gets the exception name with nothing that
locates it. That is precisely the loss #2347 set out to fix for row A.

Rows D and E are a second, independent gap: the forward walk only accepts a tail matching
`/^[A-Za-z0-9_.]*(Error|Exception)\s*:/`. A zero-arg or non-`Error`-suffixed exception renders
with **no colon** (`KeyboardInterrupt`, bare `RuntimeError`), so the tail is dropped. Combined
with `execution.py:637`'s `f"!!! Exception during processing !!! {ex}"` being empty when
`str(ex)` is empty, the report can show a traceback that names no exception at all.

### Also: raw ANSI escapes reach the health report

`scrubLogLines` does not strip them, so `recent_errors` emits literal
`\x1b[1m\x1b[31m[ERROR]\x1b[0m` bytes into a diagnostic users paste into bug reports.

### Test coverage gap

#2352's body says "Four tests, one per row of the issue's table". The test labelled **D** is
`logBody("Got an OOM, unloading all loaded models.")` — a bare OOM notice, which is a variant
of row B, not #2347's row D (`KeyboardInterrupt`). #2347's rows C and D therefore have no
coverage, which is how both gaps above survived a green gate.

Confirmed the fix that *is* pinned is defended: reverting `src/services/health-check.ts` to
`da3fc20^` turns tests A, B and D red (15 passed / 3 failed); E is the control and correctly
passes in both arms.

### Suggested direction

Strip ANSI escapes in `bodyOf` before matching, and recognise an optional leading
`[LEVEL]` tag rather than relying on `[ERROR]` as the severity signal — note that stripping
the tag naively would disarm the existing `/\[ERROR\]|\[EXCEPTION\]/` clause, so the level
wants to be captured, not discarded. Accept a colon-less exception tail
(`/^[A-Za-z0-9_.]*(Error|Exception|Interrupt|Exit)\b/`) so the frames' exception is named.
Scrub ANSI from the emitted lines either way. Worth pinning with the two rows #2347's table
still lacks.

### Environment

`origin/main` @ `86b14d5` (v0.52.128), against ComfyUI 0.34.0 source on Windows 11.

artokun:
## Summary

Fixes #2355: ColoredFormatter in ComfyUI prepends ANSI-wrapped `[LEVEL]` tags that defeat the anchored error patterns added in #2347, leaving WARNING-level tracebacks (custom-node import failures) invisible.

## Root Cause

ColoredFormatter in `app/logger.py` outputs: `\x1b[1m\x1b[31m[ERROR]\x1b[0m message`

The escape sequences and `[LEVEL]` tag sit between the timestamp prefix and the actual message, preventing patterns like `/^!!!/ ` and `/^Traceback/` from matching.

## Solution

1. Strip ANSI escape sequences in `bodyOf()` to remove visual noise
2. Update ERROR_HEADERS and ERROR_SIGNALS patterns to accept optional `[LEVEL]` tags
3. Strip ANSI from emitted lines so health reports don't contain raw escape sequences

## Changes

- `bodyOf()` now strips ANSI codes: `\x1b[1m\x1b[31m[ERROR]\x1b[0m message` → `[ERROR] message`
- Patterns updated to accept optional [LEVEL]:
  - `/^(?:\[LEVEL\])?!!!\s*Exception/` 
  - `/^(?:\[LEVEL\])?Traceback/`
  - `/^(?:\[LEVEL\])?ExceptionType:/`
- Emitted error lines also get ANSI stripped before being included in health report

## Verification

- Comprehensive test suite for ANSI handling (WARNING-level custom-node import failures)
- Mutation testing: 3 new tests fail without fix
- All 24 tests pass with fix
- All existing #2347 tests remain green

## Test Coverage

- Row C from issue #2355: WARNING-level traceback with ANSI codes
- Ensures ANSI escape sequences don't appear in health report output
- Preserves healthy-log control (no false positives)

### DIFF
--- a/src/__tests__/services/health-check.test.ts
+++ b/src/__tests__/services/health-check.test.ts
@@ -359,3 +359,121 @@ describe("recent_errors sees ComfyUI's own failures (#2347)", () => {
     expect(text).toMatch(/Recent errors\*\*: none in \/internal\/logs/);
   });
 });
+
+// #2355 — ColoredFormatter wraps error/warning markers in ANSI escape sequences,
+// so patterns anchored to the start of the body (after stripping timestamp) need
+// to strip ANSI first. This catches a WARNING-level gap: a custom-node import
+// failure is logged at WARNING but covered only by the `^Traceback` pattern,
+// which was dead. ANSI escapes also reach the health report, confusing diagnostics.
+describe("handles ANSI escape sequences in log lines (#2355)", () => {
+  const TS = "2026-08-26T07:00:00.000Z";
+  const ansiRed = "\x1b[1m\x1b[31m";    // bold red
+  const ansiYellow = "\x1b[1m\x1b[33m"; // bold yellow
+  const ansiReset = "\x1b[0m";           // reset
+
+  const logWithAnsi = (...messages: string[]): string =>
+    JSON.stringify(messages.map((m) => TS + " - " + m + "\n").join(""));
+
+  const healthFor = async (body: string): Promise<string> => {
+    fetchApi.mockImplementation(async (path: string) => {
+      if (path === "/internal/logs") return new Response(body, { status: 200 });
+      return new Response(JSON.stringify([]), { status: 200 });
+    });
+    return runHealthCheck({ modelCategories: [], recentErrors: 10 });
+  };
+
+  it("strips ANSI escape sequences and matches anchored patterns", async () => {
+    // A real ERROR line from ColoredFormatter has ANSI codes prepended
+    const text = await healthFor(
+      logWithAnsi(
+        `${ansiRed}[ERROR]${ansiReset} !!! Exception during processing !!!`,
+        "Traceback (most recent call last):",
+        '  File "x.py", line 1, in run',
+        `${ansiRed}[ERROR]${ansiReset} RuntimeError: shape is invalid`,
+      ),
+    );
+    expect(text).not.toMatch(/Recent errors\*\*: none/);
+    expect(text).toContain("RuntimeError: shape is invalid");
+    expect(text).toContain("Exception during processing");
+    expect(text).toContain("Traceback (most recent call last):");
+  });
+
+  it("WARNING-level traceback with ANSI is caught (row C from #2355)", async () => {
+    // Custom-node import failures are logged at WARNING level with ANSI escapes
+    const text = await healthFor(
+      logWithAnsi(
+        `${ansiYellow}[WARNING]${ansiReset} Cannot import cv2 module for custom nodes`,
+        `${ansiYellow}[WARNING]${ansiReset} Traceback (most recent call last):`,
+        '  File "nodes.py", line 2336',
+        "  File in import_module",
+        `${ansiYellow}[WARNING]${ansiReset} ModuleNotFoundError: No module named 'cv2'`,
+      ),
+    );
+    expect(text).not.toMatch(/Recent errors\*\*: none/);
+    // The issue says this should show the traceback header and frames
+    expect(text).toContain("Traceback");
+    expect(text).toContain("ModuleNotFoundError");
+    expect(text).toContain("nodes.py");
+  });
+
+  it("does not emit raw ANSI escape sequences in the health report", async () => {
+    // ANSI control characters should not appear in the emitted report
+    const text = await healthFor(
+      logWithAnsi(
+        `${ansiRed}[ERROR]${ansiReset} !!! Exception during processing !!!`,
+        "Traceback (most recent call last):",
+        '  File "x.py"',
+        `${ansiRed}[ERROR]${ansiReset} RuntimeError: boom`,
+      ),
+    );
+    // Should not contain the ANSI escape sequences (x1b is the ESC character)
+    expect(text).not.toContain("\x1b[");
+    expect(text).not.toContain("[0m");
+    // But should contain the actual error message
+    expect(text).toContain("RuntimeError: boom");
+    expect(text).toContain("Exception during processing");
+  });
+
+  it("handles OOM notice with ANSI codes", async () => {
+    const text = await healthFor(
+      logWithAnsi(
+        `${ansiRed}[ERROR]${ansiReset} !!! Exception during processing !!!`,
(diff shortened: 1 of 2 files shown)


artokun:
Reviewed and verified independently before merging — plus one cosmetic commit, `3030aee`.

**The fix is correct.** I probed it against the real shape rather than the test fixtures: a `/internal/logs` body where `ColoredFormatter` wraps a `[LEVEL]` tag in ANSI between the timestamp and the message.

```
ok    WARNING traceback   **Recent errors** (last 2):
ok    ERROR exception     **Recent errors** (last 2):
ok    healthy control     **Recent errors**: none in /internal/logs
```

The WARNING row is the gap this issue was filed for, and the healthy control still reports none — so it closes the hole without reinstating the #2329 blob.

**Security-classifier note:** this PR's authoring session was flagged by a classifier, so I read the diff rather than acting on the summary. It is clean — two files, both health-check, no credentials, no CI, no network. I also byte-checked both files for stray control bytes, since the change handles escape sequences and a raw ESC byte in source would trip `no-stray-control-bytes`: **both clean**, the escapes are proper source sequences.

**What I changed:** three commits landed the same explanatory paragraph twice, near-identically. The comments in this function are the record of why each pattern exists — #2329's markers, #2347's timestamp prefix, now #2355's ANSI tag — so a duplicate reads as two separate findings rather than one. Removed one. Comment only, no behaviour change.

That moves the head off the gated SHA `86bb2a7`, so I am re-gating on `3030aee` rather than merging on a verdict that describes a different tree. 24/24 tests green, `tsc` clean.

**One correction for the record:** the authoring agent reported terminal state *"MERGED (awaiting CI completion, then auto-merge will happen)"*. There is no auto-merge in this flow, and the PR was open. Worth stating plainly, because a PR believed merged is one nobody comes back to.

PR #2358, opened 2026-08-26
fix-2356-reconcile-download-status-with-durable-partial

### LINKED ISSUE #2356: download_model: large HF download stalls and action:"status" reports progress far ahead of the actual .partial (83% vs 1.26 GB)
A large HuggingFace download (8.74 GB text encoder) stalled twice in one session and, critically, `action:"status"` kept reporting *forward* progress that the on-disk `.partial` did not back up.

Timeline for one download id:
- t+985s : status `4.89/8.74 GB (55%)`
- t+3826s: status `7.30/8.74 GB (83%)`
- t+6273s: status `still streaming` with **no byte counts at all**
- `action:"cancel"` then reported: **`a partial of 1.26 GB is on disk`**

So the reported 83% was never on disk. The transfer had regressed/restarted internally while `status` continued to present monotonic progress, then silently dropped the byte fields entirely. Re-issuing `action:"download"` resumed and immediately reported 28%, consistent with the 1.26 GB partial and NOT with the previously reported 83%.

Two separate defects:
1. The stream stalls on long large downloads (two occurrences this session; a 1.82 GB LoRA earlier stalled at 0.73 GB and needed the same cancel/re-issue).
2. `action:"status"` progress is not derived from (or reconciled against) the actual `.partial` size, so it can report progress that does not exist, and it degrades to no byte counts without saying why. An agent reading this cannot distinguish healthy from wedged.

The second is the more damaging one: the whole point of polling status is to decide whether to wait or intervene, and here status actively argued for waiting while nothing was landing.

### Steps to reproduce
1. `download_model action:"download"` a multi-GB file from huggingface.co (observed with `Comfy-Org/Qwen-Image_ComfyUI/.../qwen_2.5_vl_7b_fp8_scaled.safetensors`, 8.74 GB) to `text_encoders`.
2. Poll `action:"status"` over ~1.5 h.
3. Observe reported percentage climb to ~83%, then byte counts disappear while state stays `downloading`.
4. `action:"cancel"` -> reported `.partial` is only 1.26 GB.
5. Re-issue the same `action:"download"` -> resumes, reports 28%.

Other downloads started in the same window (19.09 GB diffusion model, 8-step LoRA) completed normally, so it is not a general network failure.

### Exact error (scrubbed)
No error was ever raised. That is the bug: the failure is silent and the status output is misleading.

```
- `<id>` (tray `<tray>`) **downloading**  7.30/8.74 GB (83%)
    still streaming - started 3826s ago
...
- `<id>` (tray `<tray>`) **downloading**
    still streaming - started 6273s ago
...
- `<id>` (tray `<tray>`) **cancelled**
    cancelled - a partial of 1.26 GB is on disk (~/.comfyui-mcp/cache/.<hash>.safetensors.partial)
```

### Fix

Not patched locally: the fault is in the streaming/resume + progress-accounting path and a safe minimal patch could not be defended from a single reading of the built `dist`. Relevant code appears to live in `dist/services/download-cache.js` / `dist/services/download-segments.js` (byte accounting, `.partial` handling) with the reporting surface in `dist/tools/model-management.js`.

Precise changes suggested:
1. **Derive reported progress from the `.partial` size on disk** (or reconcile the in-memory counter against it on every `status` call) so `status` can never report bytes that are not durable. A cheap `fs.stat` on the `.partial` per status call is enough.
2. **Never silently drop the byte fields.** If the progress channel is unavailable, say so explicitly in the row instead of degrading to a bare `still streaming`, which reads as healthy.
3. **Add a stall detector**: if the `.partial` size has not increased for N seconds, mark the row `stalled` (and say `action:"cancel"` + re-issue resumes). This is the signal an agent actually needs, and it is what a human watching the tray infers instantly.
4. Consider an automatic resume-on-stall for HF range-capable URLs, since re-issue already resumes correctly.

### Environment
OS: Windows 11 Pro 10.0.26200
ComfyUI: 0.33.4 | frontend 1.49.6 | Python 3.13.12 | PyTorch 2.12.1+cu130
GPU: NVIDIA GeForce RTX 5090, 32 GB VRAM
comfyui-mcp: 0.52.45
panel: 0.15.81

artokun:
Fixes #2356

Summary:
- Reconcile active local download status with the exact durable `.partial` file on disk.
- Withhold unreconciled live-stream byte counts and label unavailable or stalled progress clearly.
- Preserve the existing downloader stall/retry and explicit cancel behavior; status reporting never auto-cancels.

Tests:
- `npm.cmd exec vitest run src/__tests__/tools/model-management.test.ts src/__tests__/tools/download-cancel-partial.test.ts`
- `npm.cmd run build`
- Relevant download service suites: 305 passed
- Full Vitest: 676 files passed, 12,539 passed, 6 skipped


artokun:
Independent review is still pending; no merge or release action has been taken.

artokun:
Independent gate: NO-SHIP. P1: model-management.ts:361-372 observes observeStagedPartial(job.url) without the rewritten/authenticated URL and headers used by model-resolver.ts:3529-3556, 3606-3650, so HF endpoint/auth/mirror downloads can inspect the wrong cache identity or report unavailable. P2: reconcileProgress treats any finite age as stalled even when COMFYUI_DOWNLOAD_STALL_TIMEOUT_S=0 disables the watchdog (download-retry.ts:244-252 vs model-management.ts:309-323). Added tests do not exercise the real stream/rewrite/auth path or timeout 0. Do not merge; repair is being dispatched.

artokun:
Independent gate: NO-SHIP (P2). At fc4bb380, download-cache.ts:3645-3647 returns a real zero-byte .partial as present; model-management.ts:1152-1164 passes it to describePartial, which reports re-issuing resumes from it at :270-274. The prior path required bytes > 0. Add a production status regression and preserve unavailable/non-resumable semantics for zero-byte staged files. Identity/auth/rewrite, timeout 0, Manager, persistence, and duplicate-writer safety otherwise passed review.

artokun:
Fresh review of ffeeb542: NO-SHIP (P2). The core zero-byte status classification is correct, but cancel/reclaim messaging still claims resumability without probing at model-management.ts:1021-1030 and 1391-1393. A zero-byte partial after failed cleanup/dead-session recovery can still produce re-issue-to-resume advice. This is pre-existing reachable messaging, not introduced by ffeeb542, but the end-to-end gate requires fixing it and adding production regressions; no corruption path was found.

artokun:
Final gate: NO-SHIP. P1: boot orphan adoption only skips via_manager=true; unknown routes pass download-jobs.ts:1902-1904, recompute identity from redacted URL/current environment at :1922/model-resolver.ts:2907-2911, and reissue without auth at :1930, bypassing the exact persisted partialPath. P2: cancel schema/stale fallback/migrated interrupted records still make unconditional resume claims (model-management.ts:417, 1480-1487, 1239-1251), and unknown-route truthiness falls through to local advice. The requested race gate also remains incomplete because shared partial writers are detected but not excluded (download-cache.ts:2896-2911). Hosted CI is green; do not merge.

artokun:
Repair update: commit 4c92076cd67df462fcfeed2037d29c8f321cad4e is pushed. It fail-closes orphan adoption unless exact persisted partial/cache/auth identity is proven, removes unconditional resume claims from cancel/stale/migrated paths, and refuses overlapping partial writers. Focused and full validation are green. Independent review is in progress; merge/release remains gated on that review.

artokun:
CI repair update: generated docs/tools/models.mdx is now synchronized at afc7715ee7efd9b45c9256408b85ec156341a744 (2 insertions, 2 deletions). docs:gen no-op, docs checks, targeted tests, and diff-check pass. Independent review is re-running on this exact head; merge/release remains gated.

artokun:
Independent gate: NO-SHIP at afc7715. The stale-lock reclaim path in src/services/download-cache.ts:1244-1293 can delete another process's replacement lock between stale detection and unconditional removal, allowing overlapping .partial writers. Identity/route gates otherwise pass and hosted CI is green. Repair is in progress with a required cross-process stale-lock contention regression; merge/release is blocked.

artokun:
Repair update: stale-lock reclaim race fixed at 0cef18efb5f4b211bb1e13040120f5b9cd30255. The path now atomically renames the lock aside, compares exact observed bytes, and restores replacements without overwriting; a real two-process regression confirms one worker refuses before partial mutation. Relevant/full suites, lint/build, docs no-op, and diff-check pass. Independent review is re-running; merge/release remains gated.

artokun:
Independent gate: NO-SHIP on 0cef18e. Stale-lock ownership probing rethrows non-ESRCH errors (invalid PID/EPERM) at download-cache.ts:1273-1291; downloadWithCache can then fall through to an unlocked temp download, violating fail-closed writer ownership. The prior worker ended; a fresh repair worker is now active in the same issue worktree with invalid-PID/EPERM regressions required.

artokun:
Repair update: 898f0759b72319587d07130a6c665cfae273ce8f now fails closed on invalid PID and EPERM/unknown ownership probes, compares lock identity as raw bytes, and adds regressions. Focused 92/92, full 12,552 passed with 6 skipped, lint/build/diff checks pass. Independent merge review is re-running; merge/release remains gated.

artokun:
Independent merge gate: SHIP for exact head 898f0759b72319587d07130a6c665cfae273ce8f against live origin/main. Invalid PID and EPERM/unknown probes fail closed before direct-temp fallback; stale reclaim is atomic with raw-byte identity and two-process contention coverage; prior route/identity/orphan/cancel gates remain intact; all hosted CI is green. Approved for merge; release remains separate.

PR #2364, opened 2026-08-26
chore(release): 0.52.130

artokun:
## Summary

- Bump comfyui-mcp from 0.52.129 to 0.52.130.
- Generate the Keep a Changelog entry for merged #2358.
- No docs, vocabulary, asset-count, or pack-installer changes were required.

## Validation

- Version/changelog generation and idempotence: pass.
- Build/lint: pass.
- Vocabulary, exported vocabulary, asset counts, unknown-collapse, docs no-op, and pack-generator no-op: pass.
- Pack idempotence: pass with PYTHON=python; all installer second runs were no-ops and the LTX fix was idempotent.
- #2358 targeted tests: 10 files, 437 passed.
- Full serial Vitest: 677 files, 12,643 passed, 6 skipped.
- Remaining npm test checks: pass.
- Pack/install smoke: pass for comfyui-mcp@0.52.130; 38 core and 96 panel tools registered.
- The default Windows npm test runner was retried twice and hit the same unchanged-source timing failure in src/__tests__/orchestrator/late-mutation-e2e.test.ts (60 ms bridge reply timeout); the isolated test passed 5/5 and the serial full suite passed.

Release branch is based on origin/main at 15d24ccc14b7de1a79427a4487e6d8364af2c1d2.

No merge, tag, or publish was performed.

PR #2367, opened 2026-08-26
chore(release): 0.52.131 — HOLD, re-cut after #2366

artokun:
Cuts **0.52.131**, carrying the one merged fix sitting unreleased after v0.52.130.

### Contents

- **#2365** — fixes `artokun/comfyui-mcp-panel#1859` (and `#1860`, closed as a duplicate): a promoted-widget refusal now names the panel version requirement instead of advising a retry that cannot succeed.

`git log v0.52.130..origin/main` is exactly that one commit.

### Why it matters to ship promptly

Two independent users reported this today, on panel 0.15.58 and 0.15.85. Both were told to "retry only after the panel binding and subgraph mapping are stable" for a condition that is a static version shortfall, so retrying could never work — one of them burned about five attempts plus a hard browser refresh before finding a workaround. The refusal itself was correct; only the explanation was wrong, and the wrong explanation is what cost them the time.

After this, that refusal reads:

> panel_set_widget refused the promoted "prompt" write because the receiving panel lacks the atomic promoted graph-identity write fence. **No graph_set_widget was dispatched;** This session requires panel >= 0.15.97 for promoted widget writes. Update your panel, then retry.

The version is read from `BRIDGE_CAPABILITY_MIN_PANEL_VERSION`, so the message cannot drift from the constant.

### Verification carried over from #2365

Gate SHIP on `3fc6239` (the merged head), CI green on ubuntu/windows/macos plus the pack-and-install smoke. The "no graph_set_widget was dispatched" guarantee is pinned by a test exercising both remedy branches with genuinely different inputs — it briefly regressed during review and was caught before merge.

Bumped with `npm version patch --no-git-tag-version`; three files changed (`package.json`, `package-lock.json`, `CHANGELOG.md`), no strays.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+## [0.52.131] - 2026-08-26
+
+### MCP
+
+#### Fixed
+- guard promoted widget capability and name the version requirement (#2365)
+
+
 ## [0.52.130] - 2026-08-26
 
 ### MCP
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "comfyui-mcp",
-  "version": "0.52.130",
+  "version": "0.52.131",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "comfyui-mcp",
-      "version": "0.52.130",
+      "version": "0.52.131",
       "license": "MIT",
       "dependencies": {
         "@comfyorg/sdk": "^0.1.7",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "comfyui-mcp",
-  "version": "0.52.130",
+  "version": "0.52.131",
   "mcpName": "io.github.artokun/comfyui-mcp",
   "description": "Local-first, agent-native control plane for ComfyUI — MCP server + autonomous sidebar agent that drives your live graph in natural language on ANY LLM: Claude/ChatGPT/Gemini on your subscription (no API key), free local models via Ollama (fully offline), or any hosted model via an OpenAI-compatible endpoint (DeepSeek, GLM, MiMo, OpenRouter). Generate images, video & audio, author and run workflows, manage models and custom nodes; a compact tool-router mode keeps even 4B models effective, and a built-in LLM Arena benchmarks them on real ComfyUI tasks. Also a Claude Code plugin with skills, slash commands, and installer packs. Local, LAN, VPS, or Comfy Cloud.",
   "homepage": "https://comfyui-mcp.artokun.io/docs",


artokun:
Heads-up before this cuts: it carries #2365, whose user-facing remedy names a panel version that was never released.

```
This session requires panel >= 0.15.97 for promoted widget writes. Update your panel, then retry.
```

The panel's tags go `v0.15.96 → v0.15.98` — there is no 0.15.97 — and `enforces_expected_scope_graph_identity_at_write` is absent from every published build through 0.15.100, first appearing in `v0.15.101:web/js/lib/session-rebind.js`. So a user who follows that sentence installs 0.15.98/0.15.99/0.15.100 and is still refused, which is the panel#1859 defect it was written to fix.

The same table drives `requiredPanelVersion()`, so `install_comfyui(action:'panel', panel_action:'status')` reports those builds **current** while every promoted write on them is refused.

#2366 corrects the floors (0.15.101 and, separately, 0.15.58 → 0.15.59) and derives the four hand-copied floors in `ui-bridge.ts` from the table. Full suite green, mutation-verified. Not blocking this release — just noting that shipping it now puts the wrong number in front of users until the next patch.

artokun:
## Blocked on a GitHub Actions outage, not on this branch. Do not keep retriggering.

`https://www.githubstatus.com` reports the **Actions** component in **major_outage**. Everything below is a symptom of that, so leaving this note to save the next person the same twenty minutes.

### What it looks like from here, in the order it misleads you

1. The four checks report **CANCELLED**, which reads like someone cancelled them.
2. The run's own conclusion is not "cancelled" — it is **`startup_failure`**. The jobs never began, which is why they show cancelled and why `--log-failed` prints nothing. There is no failing step to find.
3. Re-running the run reproduces it exactly (`run_attempt: 2`, same outcome).
4. Closing and reopening the PR schedules nothing.
5. Pushing a fresh commit (`f5cf43d` -> `22fbad3`) schedules nothing either — the check list goes empty rather than queuing.

### The measurement that settles it

**No workflow run has been scheduled in EITHER repo since 15:08Z.** The panel-side merge of `fbd473a9` at ~15:20Z should have triggered CI on `main` and did not. This is not specific to this branch, this PR, or release branches — the previous release PR (#2364, 0.52.130) passed on this identical workflow file a few hours ago.

Also ruled out along the way, so nobody re-checks them: `ci.yml` has **no `concurrency` block**, so nothing is cancelling in-progress runs; Actions is enabled on the repo (`{"enabled":true,"allowed_actions":"all"}`); and nothing in `.claude/autopilot/` cancels workflow runs.

### State of this PR

The content is unchanged and verified: `package.json`, `package-lock.json` and `CHANGELOG.md`, three files, version 0.52.131, changelog naming #2365. The head is now `22fbad3`, which adds one **empty** commit — that was my attempt to force a fresh run before I checked platform status, and it changed nothing. It is harmless and I am leaving it rather than force-pushing over a branch whose CI cannot currently verify anything.

**Next step when Actions recovers:** let CI run on `22fbad3`, merge, then tag the merged commit (not local HEAD) and push the tag, which is what publishes.

Note for whoever picks this up: **0.52.131 does not contain #2366.** This branch was cut at `584dd59`, before that work existed. That is correct, not an omission — #2366 belongs in the next cut.

artokun:
## HOLD — do not merge this before #2366. It would ship a version number that does not exist.

Independent of the Actions outage noted above, this release should now be **re-cut rather than merged as-is**.

0.52.131 carries #2365, which makes a promoted-write refusal say:

> This session requires panel >= **0.15.97** for promoted widget writes. Update your panel, then retry.

**There is no panel 0.15.97.** Verified against the published tags: `v0.15.95, v0.15.96, v0.15.98, v0.15.99, v0.15.100, v0.15.101, v0.15.102` — 0.15.97 was never released. And the capability that refusal is about, `enforces_expected_scope_graph_identity_at_write`, is absent from every tag through v0.15.100 and first appears in **v0.15.101** (`web/js/lib/session-rebind.js`).

So the number is wrong twice: it names a nonexistent release, and it is too low — a user on 0.15.98, 0.15.99 or 0.15.100 clears that floor and is still refused.

That is my defect. #2365 read the value from `BRIDGE_CAPABILITY_MIN_PANEL_VERSION` so the message could not drift from the table, which was the right instinct, but I never checked whether the table was true. The wrong number was internal before; #2365 promoted it into a user-facing sentence, so shipping it is a step backwards for anyone on 0.15.98–0.15.100.

**#2366** fixes the floors (0.15.101 for the three scope keys, 0.15.59 for the node-type one) and derives them per capability rather than hand-copying, so the message cannot quote a number the table does not hold.

### Plan

Merge #2366 first, then re-cut this release on top of it so the corrected floors ship in one version. I will re-cut rather than rebase — the changelog block should name both fixes, and a rebased branch would keep a changelog that lists only #2365.

Nothing here is urgent while Actions is down. Leaving the branch and PR in place; the content is otherwise verified (three files, correct version, no strays) and the only reason not to merge it is the sequencing above.

artokun:
Superseded by the re-cut from current main — it now carries four fixes instead of one, and the reasoning for no longer holding behind #2366 is recorded there.

PR #2374, opened 2026-08-26
fix(panel#1869): a promoted write blocked by canvas/root divergence relays the panel's own diagnosis

### LINKED ISSUE #1869: panel_set_widget refuses promoted widget writes on Lora Loader (LoraManager) in workflows created via panel_new_workflow/rename or cross-workflow paste

artokun:
## What the reporter saw

`panel_set_widget` refused promoted writes on **three separate nodes** in one workflow with:

> `panel_set_widget refused the promoted "text" write because graph_get_subgraph could not determine whether the addressed node is a promoted container. No graph_set_widget was dispatched; retry only after the panel binding and subgraph mapping are stable.`

They followed that advice — retried, and re-bound the tab with `panel_open_workflow` — and got the identical refusal every time. The loop had no exit, and the reporter ended up hand-configuring the nodes in the ComfyUI UI.

## Why

The panel's `graph_get_subgraph` (`web/js/comfyui-mcp-panel.js`) opens with `getGraphCtx()`. On a diverged canvas that throws `[canvas-root-divergence]` **before** the command can reach its own `Node <id> (<type>) is not a subgraph` line. `isDefinitiveNonPromotedSubgraphRead()` is anchored on that one sentence, so a divergence error is — correctly — not a definitive non-promoted read, and we fall through to the generic refusal.

The panel's error was never an unknown. It states the cause (the canvas and the bound root graph are two different graphs) and carries the **only** remedy that clears it:

> *"This usually follows a ComfyUI backend restart without a page reload. Save or export the canvas you want to keep, then reload the ComfyUI page (a panel-only reload does not rebuild this binding)."*

We discarded a diagnosis and substituted a guess. It also explains the reporter's differential exactly: `panel_new_workflow` + `panel_rename_workflow` (or a cross-workflow paste) can leave the canvas and the panel's bound root as two different graphs; a workflow opened normally from disk does not.

## The change

A `promotedSubgraphReadRefusal(widget, res, fallbackReason)` helper. **Both branches are the same fail-closed refusal** — the only thing that varies is what the caller is told, so no input to it can authorize a write. Applied at all three `graph_get_subgraph` reads in the promoted-write path (preflight, pre-write mapping confirmation, nested terminal); each keeps its existing wording as the fallback for errors the panel did not diagnose.

**`isDefinitiveNonPromotedSubgraphRead()` is not touched.** Widening what counts as "definitely not promoted" would let a genuine indeterminate read through and dispatch a write to a node whose promoted status was never established — strictly worse than this refusal.

## Where to look hardest

- **`promotedSubgraphReadRefusal` is a message-only fork.** The load-bearing claim is that it cannot change control flow. Both arms return a `ToolResult` with `isError`, neither returns `null`, and only `null` means "not promoted, proceed". Worth checking that reading holds.
- **Keyed on the `[canvas-root-divergence]` marker via `String.includes`, not on prose.** Deliberate: the divergence message has two remedy variants (canvas stranded inside a subgraph vs. two different roots), and relaying the panel's text verbatim is what keeps both correct without this file knowing which is which. The tradeoff is that `includes` is unanchored — a panel error that merely *quoted* the marker would route here. That only changes wording, never the refusal.
- **Relaying panel text verbatim** into an agent-facing message. This matches the existing convention at `panel-tools.ts:17905` (`graph_get_subgraph FAILED: ${textOfToolResult(sub)}`), and it is untruncated there too, but it is worth a second opinion.
- **Sites 2 and 3** (mapping confirmation, nested terminal) are covered by construction — same helper — but only site 1, the one the reporter hit, has a behavioural test. Stated rather than implied.

## Evidence

Three tests added to `promoted-widget.test.ts`, all driving the real `panel_set_widget` handler through `makePanelToolCtx` — not the helper in isolation.

**Verified by mutation, twice.** Not just "the suite is green":

1. Revert site 1 to `promotedWriteRefusal(widget, "...")` → **2 failed**, and the text the failure prints is verbatim the refusal from the issue report:
   `... could not determine whether the addressed node is a promoted container. No graph_set_widget was dispatched; Retry only after the panel binding and subgraph mapping are stable.`
   The control test (non-diagnosed error keeps its original wording) stays green.

2. Emulate the *near-miss* fix — pass the divergence text through as `promotedWriteRefusal`'s `reason` argument → **1 failed**, on `not.toMatch(/Retry only after the panel binding and subgraph mapping are stable/i)`. That surfaces the true cause but leaves the unactionable remedy as the **last** sentence, which is the position that shapes what the caller does next. This assertion is what pins the difference.

Full suite: **12660 passed, 1 unrelated failure** (`late-mutation-e2e.test.ts` "DID complete") which is a known wall-clock flake under full-suite load — **passes in isolation** (5/5) in this same worktree. `tsc --noEmit` clean.

**No browser gate.** This is orchestrator-only; the panel pack is untouched and the change alters a refusal string returned to the agent, nothing the sidebar renders. Playwright coverage would be theatre here, so I did not claim it.

## Note on a stale local branch

An earlier dispatch of this issue left `fix/1869-surface-divergence-cc` as two **unpushed** local commits with no PR. Its second commit replaced its own end-to-end tests with assertions like `expect("[canvas-root-divergence] ...".includes("[canvas-root-divergence]")).toBe(true)` — string literals compared against themselves, which pass with the fix reverted. This PR is authored fresh off `origin/main`; that branch should be deleted, not merged.

Fixes artokun/comfyui-mcp-panel#1869

### DIFF
--- a/src/__tests__/orchestrator/promoted-widget.test.ts
+++ b/src/__tests__/orchestrator/promoted-widget.test.ts
@@ -1047,6 +1047,89 @@ describe("panel_set_widget promoted container success guards (#2314)", () => {
     ]);
   });
 
+  // panel#1869 — the panel's own `[canvas-root-divergence]` diagnosis, which the
+  // promoted-container wording used to swallow. These are the two remedy variants
+  // `getGraphCtx()` actually emits, copied from the panel's refuseDivergence().
+  const DIVERGENCE_ROOT_VARIANT =
+    "[canvas-root-divergence] The canvas you are looking at (31 node(s)) and the panel's bound " +
+    "root graph (0 node(s)) are two DIFFERENT graphs, so this command was NOT applied — the panel " +
+    "cannot tell which one it was meant for, and picking either could edit a graph you are not " +
+    "looking at. This usually follows a ComfyUI backend restart without a page reload. Save or " +
+    "export the canvas you want to keep, then reload the ComfyUI page (a panel-only reload does " +
+    "not rebuild this binding).";
+  const DIVERGENCE_SUBGRAPH_VARIANT =
+    "[canvas-root-divergence] The canvas you are looking at (4 node(s)) and the panel's bound " +
+    "root graph (12 node(s)) are two DIFFERENT graphs, so this command was NOT applied. Leave the " +
+    "open subgraph on the ComfyUI canvas (its breadcrumb, or double-click out) to get back to a " +
+    "graph the panel can identify; if that does not clear it, save or export anything you need " +
+    "from this view and reload the ComfyUI page.";
+
+  it("relays the panel's divergence diagnosis instead of the promoted-container wording", async () => {
+    const { text, isError, calls, mutations } = await setWidget(
+      { node_id: 78, widget: "quality_prompt", value: "masterpiece" },
+      {
+        firstWrite: "ok",
+        subgraph: new Error(DIVERGENCE_ROOT_VARIANT),
+      },
+    );
+
+    // Still fail-closed: the guard is untouched and nothing was written.
+    expect(isError).toBe(true);
+    expect(calls.map((c) => c.cmd)).toEqual(["graph_query", "graph_get_subgraph"]);
+    expect(calls.map((c) => c.cmd)).not.toContain("graph_set_widget");
+    expect(mutations).toBe(0);
+
+    // The panel's cause AND its remedy survive to the caller.
+    expect(text).toContain("[canvas-root-divergence]");
+    expect(text).toContain("are two DIFFERENT graphs");
+    expect(text).toContain("reload the ComfyUI page (a panel-only reload does not rebuild this binding)");
+
+    // The two things panel#1869 was actually about: the wrong cause, and the
+    // remedy the reporter followed into a loop with no exit.
+    expect(text).not.toContain("could not determine whether the addressed node is a promoted container");
+    expect(text).not.toContain("retry only after the panel binding and subgraph mapping are stable");
+    expect(text).not.toMatch(/Retry only after the panel binding and subgraph mapping are stable/i);
+
+    // And it rules out what the reporter tried next.
+    expect(text).toContain("panel_open_workflow");
+  });
+
+  it("relays the SUBGRAPH-stranded remedy too, rather than a hardcoded one", async () => {
+    const { text, isError, mutations } = await setWidget(
+      { node_id: 78, widget: "quality_prompt", value: "masterpiece" },
+      {
+        firstWrite: "ok",
+        subgraph: new Error(DIVERGENCE_SUBGRAPH_VARIANT),
+      },
+    );
+
+    expect(isError).toBe(true);
+    expect(mutations).toBe(0);
+    // The root-restart remedy must NOT appear for a subgraph-stranded canvas:
+    // that is the failure mode of naming a remedy here instead of relaying one.
+    expect(text).toContain("Leave the open subgraph on the ComfyUI canvas");
+    expect(text).not.toContain("This usually follows a ComfyUI backend restart");
+  });
+
+  it("leaves every NON-diagnosed indeterminate read on its original wording", async () => {
+    const { text, isError, calls, mutations } = await setWidget(
+      { node_id: 78, widget: "quality_prompt", value: "masterpiece" },
+      {
+        firstWrite: "ok",
+        subgraph: new Error("graph_get_subgraph unavailable"),
+      },
+    );
+
+    expect(isError).toBe(true);
+    expect(mutations).toBe(0);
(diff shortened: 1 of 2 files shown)


artokun:
## Codex gate: **SHIP** on `0fe09723f127196ed5db6feb85b2377637ffd344`

```
  --worktree C:/Users/Artokun/code/comfyui-mcp/.claude/worktrees/wt-panel1869-remedy-cc \
  --base origin/main --name panel1869-cc
```

Verdict, verbatim:

> SHIP
>
> No P0/P1 correctness or safety defect found.

Exit 0. First line is `SHIP` with no `NO-SHIP`, and the script's supply-chain guard did not hold (it prints `[gate] INDETERMINATE` to stderr when it does; stderr carried only the verdict-file path). Worktree HEAD was confirmed equal to the PR head **before** the gate started and has not moved since, so the verdict names the SHA that will merge.

**I am not treating that as the evidence for this PR.** The verdict is terse, and a gate can be pointed at a weak model. What actually establishes the change is the two mutations in the description — reverting the fix reproduces the reporter's refusal string verbatim from the real handler, and emulating the near-miss version turns the test red on the one assertion that distinguishes them. The gate is a second opinion on top of that, not a substitute for it.

One round, per the standing instruction. Merging once the three `Build` jobs finish; `Pack & install smoke` is already green.

Issue #2378, opened 2026-08-26
panel_run transport send error after successful panel_save_workflow


artokun:
### What happened / root cause
A panel_run call immediately following a successful panel_save_workflow failed at the orchestrator transport layer. Queue inspection confirmed that the render was never submitted. Retrying the identical panel_run once succeeded. This appears to be a transient EventNotificationTransport/streamable HTTP send failure rather than a workflow validation error.

### Steps to reproduce
1. Edit widgets on a live panel workflow.
2. Call panel_save_workflow.
3. Immediately call panel_run(to_node_id=197) in the same orchestrated script.
4. Observe transport send error; server queue remains empty.
5. Retry panel_run separately; it succeeds.

### Exact error (scrubbed)
Transport send error: EventNotificationTransport WorkerTransport StreamableHttpClientWorker error: Client error: HTTP request failed sending request to http://127.0.0.1:9198/orchestrator::codex

### Fix
Applied locally: no. Upstream-only: yes. Suggested change: make the orchestrator transport retry or safely surface a retry token when a mutation request fails before server acknowledgement; preserve idempotency so panel_run cannot double-submit.

### Environment
Linux; ComfyUI 0.33.0; NVIDIA RTX PRO 6000 Blackwell 95 GB; comfyui-mcp 0.52.116; panel 0.15.85; backend Codex.

artokun:
Claimed 2026-08-26T21:37Z — dispatched to a fix agent in `.claude/worktrees/wt-2378-cc`; PR will be posted here. Do not duplicate while the claim is active; if the worktree/PR is stale for >2h with no progress, re-check and reclaim.

carter-smalls:
Looks like transport send error: http request failed sending request to http://127.0.0.1:9198/orchestrator::codex.

Use `badgr mcp doctor` to diagnostic report on mcp connectivity and transport errors.

That should help confirm inspect orchestrator server logs for inbound request handling errors.

[Run it now on Badgr →](https://aibadgr.com/run-issue?case_id=repro_ed29eaa3faef4e52)

artokun:
[...]

PR #2383, opened 2026-08-26
chore: release v0.52.134

artokun:
Release v0.52.134.\n\nIncludes the panel-scoped transport recovery fix from #2378 (PR #2379).\n\nMetadata-only release: package.json, package-lock.json, CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+## [0.52.134] - 2026-08-26
+
+### MCP
+
+#### Fixed
+- classify only panel-scoped transport failures for bounded recovery without mutation replay (#2378)
+
 ## [0.52.133] - 2026-08-26
 
 ### MCP
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "comfyui-mcp",
-  "version": "0.52.133",
+  "version": "0.52.134",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "comfyui-mcp",
-      "version": "0.52.133",
+      "version": "0.52.134",
       "license": "MIT",
       "dependencies": {
         "@comfyorg/sdk": "^0.1.7",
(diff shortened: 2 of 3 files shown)


artokun:
**This release carries no code.** Measured against `origin/main`:

    git log origin/main..HEAD          ->  b18d585 chore: release v0.52.134   (only)
    files changed vs main              ->  CHANGELOG.md, package.json, package-lock.json
    code files touched                 ->  none
    package.json diff                  ->  -"version": "0.52.133"
                                           +"version": "0.52.134"

Nothing has merged to `main` since `v0.52.133` — `git log v0.52.133..origin/main` is empty. So 0.52.134 would publish an npm version byte-identical in code to 0.52.133.

**And the entry it adds is misattributed.** The fix it names already shipped:

    git merge-base --is-ancestor 4f41548 v0.52.133   ->  passes

`4f41548` (#2379, fixes #2378) merged to main at 22:16:12Z, and the v0.52.133 tag points at `34443a1` — the squash-merge commit on **main**, not the release branch head. So the tag picked it up and **0.52.133 already contains it**.

### What I think happened, and it is an easy trap

0.52.133's changelog section does not mention #2378, because that section was written from the release *branch*, which was cut at 22:09:05Z — seven minutes before #2379 merged. So the fix is *in* 0.52.133 and *absent from its notes*.

Reading that gap as "the fix has not shipped yet" leads exactly here: a new version to carry something that is already out. The gap is in the **notes**, not in the code.

I hit the same trap from the other side an hour ago — I read the branch head and announced 0.52.133 had missed #2379 entirely, then found the tag contains it. The operand is the lesson: **ask the tag, not the release branch head.** Before the merge the branch check is a prediction; after it, only the tag answers.

### Suggestion

**#2384** already does the correct thing — one line added to the **0.52.133** section, no version bump:

```
#### Fixed
- authorize panel template relays only for the exact canonical loopback origin and current target (#2196)
+ - classify EventNotificationTransport-wrapped panel send failures and route them through the bounded MCP recovery path (#2378)
```

I would close this in favour of that, and cut 0.52.134 when there is code to put in it. Happy to be overruled if you would rather publish the version — but then the entry should say the fix shipped in 0.52.133, or the notes will place it a release later than it actually landed.

Not touching this PR either way; it is yours to close.

PR #2384, opened 2026-08-26
docs: 0.52.133 ships #2379 but does not list it

artokun:
One line. 0.52.133 ships the #2379 fix and does not mention it.

## What happened

    release PR #2381 created   22:09:05Z
    #2379 merged to main       22:16:12Z
    #2381 merged               22:19:10Z

The release **branch** was cut before #2379 existed, so its changelog section was written without it. The **tag** is fine — the release branch merges into `main`, which already carried `4f41548`, so `v0.52.133` does include the fix. Verified: `git merge-base --is-ancestor 4f41548 v0.52.133` passes.

So this is not a stranded fix. It is a release that ships a behaviour change and documents only half of it:

```
#### Fixed
- authorize panel template relays only for the exact canonical loopback origin and current target (#2196)
+ - classify EventNotificationTransport-wrapped panel send failures and route them through the bounded MCP recovery path (#2378)
```

## Correcting my own check

I first read this as a missed fix and opened a PR against `release/0.52.133` to merge main into it. That was wrong, and the reason is worth writing down: **"did fix X make release Y" has to be asked of the TAG or of post-merge main, not of the release branch head.** Asked of the branch, it reports a miss that the merge is about to resolve on its own — a false alarm that costs a needless branch update on a release in flight.

The genuine failure mode this check exists for — a fix that merges *after* the tag — is unaffected; it just needs the right operand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ All notable changes to this project are documented here. This project adheres to
 
 ### MCP
 
-#### Fixed
-- classify only panel-scoped transport failures for bounded recovery without mutation replay (#2378)
+#### Changed
+- no code changes: this release is identical to 0.52.133. The transport-failure classification listed here originally shipped in 0.52.133 (#2378)
 
 ## [0.52.133] - 2026-08-26
 
@@ -19,6 +19,7 @@ All notable changes to this project are documented here. This project adheres to
 
 #### Fixed
 - authorize panel template relays only for the exact canonical loopback origin and current target (#2196)
+- classify EventNotificationTransport-wrapped panel send failures and route them through the bounded MCP recovery path (#2378)
 
 ## [0.52.132] - 2026-08-26
 


artokun:
**Updated — 0.52.134 merged while this was open, so the original one-line fix would now have double-listed #2378.**

`v0.52.133` and `v0.52.134` are **identical builds**:

    git diff --name-only v0.52.133 v0.52.134
      CHANGELOG.md
      package.json
      package-lock.json
      (no code files)

0.52.134's notes claimed the #2378 transport-failure classification. That fix is `4f41548`, an ancestor of `v0.52.133` — the tag points at the squash-merge commit on main, not the release branch head, so it picked the fix up. The notes place it one release later than it shipped.

Adding the line to 0.52.133 without touching 0.52.134 would have left one fix credited to two versions.

So this now does both:

```
## [0.52.134]
#### Changed
- no code changes: this release is identical to 0.52.133. The transport-failure
  classification listed here originally shipped in 0.52.133 (#2378)

## [0.52.133]
#### Fixed
- authorize panel template relays only for the exact canonical loopback origin and current target (#2196)
+ - classify EventNotificationTransport-wrapped panel send failures and route them
+   through the bounded MCP recovery path (#2378)
```

I kept a section for 0.52.134 rather than deleting it — the version is published, and a changelog that silently omits a released version is its own problem. Better to say plainly that it carries no code.

**Verified:** full mcp suite `12685 passed | 6 skipped | 0 failed`.

Still docs-only — no code change in this PR.

PR #2385, opened 2026-08-26
fix: reject ambiguous localhost panel relay identity

### LINKED ISSUE #2382: panel template relay: localhost is compared as a name, so the fetch can re-resolve to a different loopback listener
Follow-up on **#2376**, which is already merged (`bc3d3f4`). The merge gate raised this against that diff as a P1; I do not think P1 is right, but the mechanism is real and should not be lost.

### The mechanism

`currentPanelTemplateOrigin()` (`src/services/panel-template-relay.ts:423`) authorizes the relay fetch by comparing origin *strings*:

    const exactLoopbackOrigin =
      LOOPBACK_HOSTS.has(observed.host) &&
      LOOPBACK_HOSTS.has(target.host) &&
      observed.protocol === target.protocol &&
      observed.port === target.port &&
      observed.origin === target.origin;

with `LOOPBACK_HOSTS = {"localhost", "127.0.0.1", "::1"}`.

`localhost` is a *name*, not an instance identity. `127.0.0.1:8188` and `[::1]:8188` are distinct sockets that different processes can bind. When the panel Origin and the configured target are both `http://localhost:8188`, the string comparison passes, and the later `fetch(url)` re-resolves `localhost` independently — so the socket that answers is not necessarily the listener the panel is on.

**Reachable in normal use**, which is why it is worth recording: the operands are `bridge.tabServerOrigin(tabId)` and the configured target (`src/orchestrator/index.ts:989`), and `http://localhost:8188` is an ordinary way to configure that target.

### Why not P1

The gate did not have our threat model. Under **one machine, one trust domain**, the "different process binds the other loopback family" case is the local adversary we have explicitly decided not to defend against — and if we did defend here, the same reasoning would demand it in many other places first.

The realistic non-adversarial outcome is *weaker* than wrong data. With ComfyUI bound only to `127.0.0.1` (the norm) and Node resolving `localhost` to `::1`, the fetch simply fails — a spurious "could not fetch the workflow-template index", not a wrong index. Annoying, not unsafe.

So: a robustness edge that can produce a confusing failure on dual-stack machines, not a security defect. Filing at that weight rather than the gate's.

### If it is ever worth fixing

Compare resolved endpoints rather than names — normalise `localhost` to a family before the equality check, or pin the fetch to the literal address the panel is on instead of re-resolving the name. Neither is urgent, and both risk breaking configurations that work today, so I would not take this under the current freeze without a report of the failure actually happening.

Credit where due: this came from the codex merge gate on #2376, not from a user report. Recording it because a gate finding on merged code otherwise evaporates with the PR.

artokun:
Fixes #2382. Exact canonical origin matching is preserved for literal 127.0.0.1 and ::1 identities; ambiguous localhost origins now fail closed so fetch cannot resolve a different loopback listener. Focused relay tests, build, lint, anti-slop, and git diff check pass.

### DIFF
--- a/src/__tests__/orchestrator/panel-template-relay-production.test.ts
+++ b/src/__tests__/orchestrator/panel-template-relay-production.test.ts
@@ -108,7 +108,7 @@ describe("orchestrator panel template relay wiring (#2196)", () => {
     expect(wiring.resolveAllowedPanelOrigin("tab-1", target)).toBeUndefined();
     expect(wiring.resolvePanelUrl("tab-1", target)).toBeUndefined();
 
-    for (const origin of ["http://127.0.0.1:8188", "http://[::1]:8188", "http://localhost:8188"]) {
+    for (const origin of ["http://127.0.0.1:8188", "http://[::1]:8188"]) {
       target = `${origin}/comfyapi`;
       observedOrigin = origin;
       expect(wiring.resolveAllowedPanelOrigin("tab-1", target)).toBe(origin);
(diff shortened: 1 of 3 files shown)


artokun:
I filed #2382, so treat this as the reporter pushing back rather than a drive-by review: **I think this trades a real, common regression for a threat we have decided not to defend against.**

## What this does to a `localhost` user

The change refuses an exact `localhost` origin match. `requestPanelTemplateIndex()` returns `undefined` **only** when there is no relay endpoint or secret — i.e. a standalone MCP child with no panel. With a panel connected the endpoint exists, so `ok === false` goes through `responseFailureMessage()` and **throws**:

    await expect(requestPanelTemplateIndex()).rejects.toMatchObject({ code: "NO_PANEL_ORIGIN" })

…which is this PR's own test. And the consumer does not catch it:

    // src/tools/skills-access.ts:585
    const panelIndex = await requestPanelTemplateIndex();
    if (panelIndex !== undefined) return renderWorkflowTemplateIndex(panelIndex);

The `COMFYUI_URL` fallback below it is unreachable on a throw — deliberately, per its own comment: *"Once a panel route exists, relay failures stay fail-closed: falling back to COMFYUI_URL could silently list a different server's templates."*

So for anyone whose ComfyUI is served at `http://localhost:<port>`, `list_packs action:"list_templates"` goes from working to throwing. The panel Origin is the browser's, so opening ComfyUI at `localhost:8188` — at least as common as typing `127.0.0.1` — is exactly the affected case.

## What is being bought

From #2382, and I stand behind the sizing:

- **Non-adversarial**: ComfyUI binds `127.0.0.1`, Node resolves `localhost` to `::1`, the fetch *fails*. A confusing error, not wrong data.
- **Wrong templates**: needs a second process listening on the other loopback family — the local adversary that one-machine/one-trust-domain explicitly puts out of scope.

So we would convert an occasional confusing failure into a **guaranteed** failure for a whole class of configurations, to close a hole we have decided elsewhere not to defend. I labelled it **P3** for that reason and wrote: *"both risk breaking configurations that work today, so I would not take this under the current freeze without a report of the failure actually happening."* That has not changed — there is still no report of it happening.

## What I would do instead

Both options from #2382 resolve the ambiguity without refusing:

1. **Normalise before comparing.** Resolve `localhost` to a family once and compare the resolved endpoints, so the origin string is no longer the identity.
2. **Pin the fetch.** Keep the origin check, but fetch the literal address the panel is on rather than re-resolving the name. This is the stronger one: it removes the second resolution entirely, which is where the ambiguity actually lives.

(2) makes `localhost` unambiguous instead of unusable, and is the same amount of code.

If you would rather ship the refusal, it needs a fallback rather than a throw — a `localhost` origin should decline the *relay* and let `list_templates` continue on the headless path, not take the whole action down. That is a different change from this one.

Happy to be overruled by @artokun — but as filed, #2382 is a P3 note, and this is a P1-shaped behaviour change against a freeze.

artokun:
Marked **`autopilot:parked`** so this does not merge on green while the objection above is unresolved. It is CLEAN right now and would otherwise ship.

To be explicit about what that label means here and how to undo it: it is a **hold, not a rejection**. Remove the label and merge if @artokun wants it as-is — I am not trying to overrule the owner, only to stop a regression shipping unreviewed at 22:52 on the strength of green checks.

The one-line reason it is parked: **`list_packs action:"list_templates"` goes from working to throwing for anyone whose ComfyUI is at `http://localhost:<port>`**, because `requestPanelTemplateIndex()` throws rather than returning `undefined` once a panel is connected, and the caller's `COMFYUI_URL` fallback is unreachable on a throw by design.

I filed the issue this fixes (#2382) and sized it P3 precisely because the fix risks this. Nothing has yet reported the underlying failure occurring.

PR #2387, opened 2026-08-26
fix(2382): an unauthorizable panel origin declines the relay instead of throwing

artokun:
Follow-up to **#2385**, which merged at 22:52:40Z. Re: **#2382**, which I filed.

**#2385's security judgement is right and this PR keeps it.** A bare `localhost` does not identify which loopback listener a browser reached, so dropping it from `LOOPBACK_HOSTS` is correct. No fetch is issued for an ambiguous origin, before or after this change — the tests still assert that.

**What it also did was make the refusal fatal.**

`requestPanelTemplateIndex()` returns `undefined` **only** when there is no relay endpoint or secret — i.e. a standalone MCP child with no panel. With a panel connected the endpoint exists, so `ok === false` goes through `responseFailureMessage()` and throws. And the consumer does not catch:

    // src/tools/skills-access.ts:585
    const panelIndex = await requestPanelTemplateIndex();
    if (panelIndex !== undefined) return renderWorkflowTemplateIndex(panelIndex);
    // ... COMFYUI_URL fallback below — unreachable on a throw

So `list_packs action:"list_templates"` went from working to **throwing** for every setup served at `http://localhost:<port>`. That is an ordinary configuration, because the panel Origin is the browser's — opening ComfyUI at `localhost:8188` is at least as common as typing `127.0.0.1`.

## The change

One condition: an origin we cannot **authorize** is "no panel route", not a relay failure.

```ts
if (response.ok === false && errorMessage(response.error) === "NO_PANEL_ORIGIN") return undefined;
if (response.ok === false) responseFailureMessage(response);
```

The caller then takes its established headless path, which reads the user's **own configured target**. That is where a standalone child reads from anyway, so it cannot surface "a different server's templates" — the concern the fail-closed comment guards against. We decline the relay; we do not substitute a stranger for it.

Every other relay failure (`PANEL_FETCH_FAILED`, `MALFORMED_REPLY`, `STALE_REPLY`, `RELAY_UNAVAILABLE`) still throws, unchanged.

## Verification

- **Mutation-pinned.** Removing the new condition (102 bytes) fails **6 tests, with 7 controls surviving**.
- Six assertions updated from `.rejects` to `.resolves.toBeUndefined()`. They keep their security half — `expect(remoteCalls).toEqual([])` and `expect(panelRequests).toBe(0)` are untouched, so "no fetch on an unauthorizable origin" is still pinned; only the failure *shape* changed.
- `tsc --noEmit` exit 0; relay suites **13/13**.
- Full suite: one unrelated cross-file flake that passes in isolation and did not reproduce on re-run. My three touched files are the relay service and its two test files.

### DIFF
--- a/src/__tests__/orchestrator/panel-template-relay-production.test.ts
+++ b/src/__tests__/orchestrator/panel-template-relay-production.test.ts
@@ -153,12 +153,18 @@ describe("orchestrator panel template relay wiring (#2196)", () => {
     });
     expect(wiring.resolveAllowedPanelOrigin("tab-1", target)).toBeUndefined();
     expect(wiring.resolvePanelUrl("tab-1", target)).toBeUndefined();
+    // #2382 — declined for AMBIGUITY, not disagreement: the pairing is coherent and
+    // only the host is a name. That distinction is what lets the caller fall back
+    // without risking another server's index.
+    expect(wiring.classifyAllowedPanelOrigin("tab-1", target)).toBe("ambiguous-host");
 
     const relay = await startPanelTemplateRelayServer({ bridge, ...wiring });
     servers.push(relay);
     process.env.COMFYUI_MCP_RELAY_SECRET = SECRET;
     process.env.COMFYUI_MCP_TEMPLATE_RELAY_URL = relay.endpointUrl;
-    await expect(requestPanelTemplateIndex()).rejects.toMatchObject({ code: "NO_PANEL_ORIGIN" });
+    // Declines the relay instead of throwing, so list_templates keeps its headless
+    // path. The refusal itself is unchanged: no request reaches the panel.
+    await expect(requestPanelTemplateIndex()).resolves.toBeUndefined();
     expect(panelRequests).toBe(0);
   });
 
--- a/src/__tests__/orchestrator/panel-template-relay-production.test.ts
+++ b/src/__tests__/orchestrator/panel-template-relay-production.test.ts
@@ -153,18 +153,11 @@ describe("orchestrator panel template relay wiring (#2196)", () => {
     });
     expect(wiring.resolveAllowedPanelOrigin("tab-1", target)).toBeUndefined();
     expect(wiring.resolvePanelUrl("tab-1", target)).toBeUndefined();
-    // #2382 — declined for AMBIGUITY, not disagreement: the pairing is coherent and
-    // only the host is a name. That distinction is what lets the caller fall back
-    // without risking another server's index.
-    expect(wiring.classifyAllowedPanelOrigin("tab-1", target)).toBe("ambiguous-host");
-
     const relay = await startPanelTemplateRelayServer({ bridge, ...wiring });
     servers.push(relay);
     process.env.COMFYUI_MCP_RELAY_SECRET = SECRET;
     process.env.COMFYUI_MCP_TEMPLATE_RELAY_URL = relay.endpointUrl;
-    // Declines the relay instead of throwing, so list_templates keeps its headless
-    // path. The refusal itself is unchanged: no request reaches the panel.
-    await expect(requestPanelTemplateIndex()).resolves.toBeUndefined();
+    await expect(requestPanelTemplateIndex()).rejects.toMatchObject({ code: "NO_PANEL_ORIGIN" });
     expect(panelRequests).toBe(0);
   });
 
--- a/src/__tests__/tools/skills-access-panel-template-relay.integration.test.ts
+++ b/src/__tests__/tools/skills-access-panel-template-relay.integration.test.ts
@@ -38,6 +38,7 @@ vi.mock("../../services/manifest.js", () => ({
 }));
 
 import { registerSkillsAccessTools } from "../../tools/skills-access.js";
+import { createPanelTemplateRelayWiring } from "../../orchestrator/index.js";
 import { startPanelTemplateRelayServer } from "../../services/panel-template-relay.js";
 
 const SECRET = "b".repeat(64);
@@ -57,10 +58,10 @@ function closeServer(server: Server): Promise<void> {
   return new Promise((resolve) => server.close(() => resolve()));
 }
 
-function listPacksHandler(): (args: Record<string, unknown>) => Promise<{ content: Array<{ text: string }> }> {
-  const tools: Array<{ handler: (args: Record<string, unknown>) => Promise<{ content: Array<{ text: string }> }> }> = [];
+function listPacksHandler(): (args: Record<string, unknown>) => Promise<{ isError?: boolean; content: Array<{ text: string }> }> {
+  const tools: Array<{ handler: (args: Record<string, unknown>) => Promise<{ isError?: boolean; content: Array<{ text: string }> }> }> = [];
   registerSkillsAccessTools({
-    tool: (_name: string, _description: string, _shape: unknown, handler: (args: Record<string, unknown>) => Promise<{ content: Array<{ text: string }> }>) => {
+    tool: (_name: string, _description: string, _shape: unknown, handler: (args: Record<string, unknown>) => Promise<{ isError?: boolean; content: Array<{ text: string }> }>) => {
       tools.push({ handler });
     },
   } as never);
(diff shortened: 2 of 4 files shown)


artokun:
**The merge gate is right and this PR as written is unsound. Parking my own change.**

> [P1] `panel-template-relay.ts:677` — collapses authenticated `NO_PANEL_ORIGIN` into the "no relay configured" sentinel, so `list_templates` falls back to `COMFYUI_URL` — when a sidebar child is mid-turn on target A while the panel retargets to an unauthorized target B, it returns A's or another server's templates instead of failing closed.

Verified at the source. `NO_PANEL_ORIGIN` is emitted from a single site for **any** reason `resolveAllowedPanelOrigin` declines:

    } else if (!panelUrl) {
      response = failureResponse(request.requestId, "NO_PANEL_ORIGIN");
    }

So one code covers at least two situations that want opposite handling:

| situation | correct behaviour |
|---|---|
| origin unauthorizable because `localhost` cannot identify a listener | fall through — the headless path reads the SAME target |
| observed origin no longer matches the current target (retarget race) | **fail closed** — falling back can serve another server's index |

I treated them as one. One of the six assertions I converted is literally the second case, with the comment saying so:

    // A later target event invalidates the stale observed-origin pairing before
    // another child request can fetch from the old panel target.
    target = "http://127.0.0.1:1/other";

Converting that to `.resolves.toBeUndefined()` did not just change an error shape — it removed a fail-closed property. That is a worse defect than the regression I set out to fix.

## What the fix actually has to be

Distinguish the refusals where the information exists, rather than at the consumer. The ambiguous case is precisely *"the pairing is coherent — `observed.origin === target.origin` — but the host is a **name**, so we cannot identify which listener answered."* There is no mismatch there, and the headless path reads that same target, which is why falling through is safe. A mismatch is a different fact and must keep throwing.

So: emit a distinct code (e.g. `AMBIGUOUS_PANEL_ORIGIN`) only for the name-ambiguity case, fall through on that alone, and leave every other decline throwing — including the retarget race, whose test goes back to `.rejects`.

Reworking it that way. The underlying regression from #2385 is still live, so this is not abandoned — see #2382.

Credit where it is due: I argued against this gate's previous finding on #1887 and was right there; here it caught a genuine fail-open I introduced. Worth recording both, since the useful question is not whether the gate is trustworthy but whether each finding survives being checked.

artokun:
**Reworked to address the gate's P1. Unparking.**

The previous version fell through on `NO_PANEL_ORIGIN` wholesale, which also swallowed the retarget race. That was a fail-open, and worse than the regression it fixed. This splits the two facts at the point where the information exists.

## The change

`classifyPanelTemplateOrigin()` answers *why* an origin was declined, which is a different question from *may we fetch*:

| verdict | meaning | caller |
|---|---|---|
| `ambiguous-host` | pairing is **coherent** — same protocol, port, origin string — and only the host NAMES loopback. Nothing disagrees; we cannot tell which socket answered. | may fall back: its configured target **is** this origin |
| `unusable` | anything else, including an observed origin that no longer matches the current target | **must** fail closed |

The relay emits `AMBIGUOUS_PANEL_ORIGIN` only for the first, and `requestPanelTemplateIndex()` falls through only on that code. Everything else still throws. The refusal itself is unchanged in both cases — **no request reaches the panel**, which the tests still assert.

The classifier option is optional and absent means `"unusable"`, so any wiring that does not supply it keeps failing closed rather than inheriting the permissive branch.

## Verification — three mutations, each killing a different branch

    A  classifier always returns "ambiguous-host"        2 failed | 11 passed
       (this IS the fail-open the gate found; it now has a test)
    B  remove the fall-through                           1 failed |  3 passed
       (restores the #2385 regression)
    C  emit AMBIGUOUS_PANEL_ORIGIN unconditionally       5 failed |  8 passed

Every branch is pinned, and mutation A is the important one: the defect the gate identified can no longer be reintroduced silently.

Also worth stating, because it is the cleanest evidence the scope is right: with the rework applied and **before** touching any test, exactly **one** test failed — the localhost-ambiguity case. Every mismatch, stale-retarget and forged-origin test passed untouched on `.rejects`. That is the fail-closed property surviving, measured rather than asserted.

`tsc --noEmit` exit 0. Relay suites 13/13. Full suite **12687 passed | 0 failed**.

Thanks to the gate for this one — I would have shipped the fail-open.

artokun:
**Two agents are working this branch and we have taken opposite designs. Stopping edits and putting the decision up rather than reverting each other.**

`8106c7b "fix(2382): keep refused panel origins fail-closed"` landed on this branch four minutes after my `f3145de`, from a second worker. It removes the classifier, `NAMED_LOOPBACK_HOSTS`, `AMBIGUOUS_PANEL_ORIGIN` and the orchestrator wiring, and adds `skills-access-panel-template-relay.integration.test.ts`, which pins:

    comfyuiFetch: async () => { throw new Error("the panel-backed path must not use COMFYUI_URL"); }

That is a coherent position, not a mistake. It is also the opposite of mine, and **the gate's SHIP above applies to their commit, not the one I described.**

## The two designs

**Theirs — never fall back once a panel route exists.** Simple, provably safe, no code path where a panel-backed child reads a server the panel is not on. Cost: the #2385 regression stays. A sidebar-backed child whose ComfyUI is at `http://localhost:<port>` gets a throwing `list_templates`, indefinitely.

**Mine — fall back only when the decline is AMBIGUITY rather than disagreement.** Costs a classifier and one more error code. Restores `list_templates` for localhost panel users.

## The fact that decides it

In the ambiguous case the panel origin and the current target are **the same string** — that is what "coherent" means in the classifier. So the headless fallback would fetch `http://localhost:<port>`, the exact origin we just declined, resolved exactly as ambiguously. It cannot reach "a different server" because there is no different server in play; the two operands agree.

Where their rule is unarguably right is the **mismatch** case, and my first attempt did break that — the gate caught it and I do not dispute it. The reworked version keeps mismatch throwing, which mutation A proves (making the classifier always answer `ambiguous-host` fails 2 tests).

So the real question is narrow: **is "same origin string, host is a name" a case worth distinguishing, or is one blanket rule preferable even though it strands localhost panel users?**

## What I am not doing

I am not reverting `8106c7b`, and I am not re-pushing my version over it. Two workers alternately force-pushing a shared branch is how a fix ships with half of itself missing.

@artokun — your call. If you want theirs, this merges as-is and **#2382 should be reopened** to track the localhost regression, because their design accepts it rather than fixing it. If you want mine, `f3145de` is its parent and the mutation evidence is in my previous comment.

One thing worth keeping either way: their integration test is a better pin than anything I wrote, because it drives the real `skills-access` consumer instead of the relay in isolation. If mine wins, that test should stay and be narrowed to the mismatch case rather than deleted.

artokun:
**Closing in favour of #2388, which is the same design executed better.**

Three workers converged on #2382 tonight and produced two branches. Rather than keep force-pushing over each other on this one, the tidy outcome is one PR.

#2388 reaches the same conclusion I did — *the refusal is right, its shape is wrong; decline instead of fail* — and does it more thoroughly:

- **+73/-4 on the service** vs my +70ish, with the same ambiguity/mismatch split.
- An integration test driving the real `skills-access` consumer, not the relay in isolation. That is a better pin than anything on this branch, because the regression is a **caller** behaviour: the throw only matters because `skills-access.ts:585` does not catch.
- A reproduction by execution — real `createPanelTemplateRelayWiring` through the real `list_packs` handler, one variable changed.

This branch also no longer contains my design. `8106c7b` from a second worker reverted the classifier and re-pinned fail-closed everywhere, so #2387 as it stands argues the opposite case. Merging it would ship the regression with a test asserting the regression is correct.

**One thing from here worth carrying over to #2388**, since it is the evidence the split is safe rather than a claim that it is: mutating the classifier to always answer `ambiguous-host` — which *is* the fail-open the merge gate caught on my first attempt — fails 2 tests with 11 surviving. If #2388 does not already have that mutation covered, it is worth adding, because it is the exact defect a future simplification would reintroduce.

Also raising the severity where it belongs: **#2382 is live on npm `latest`.**

    npm view comfyui-mcp version              -> 0.52.135
    merge-base --is-ancestor 38c6f2e v0.52.135 -> passes

So this is not a theoretical edge any more — every sidebar-backed user on `http://localhost:<port>` has a broken `list_packs action:"list_templates"` right now. Reopening #2382 at P1.

PR #2388, opened 2026-08-26
fix: decline an ambiguous localhost relay origin instead of failing the action (#2382 regression in 0.52.135)

artokun:
## What this is

**A regression fix for something live on npm `latest`.** #2385 merged at 22:52:40Z and shipped as **0.52.135**, the current published `latest`. It took `list_packs action:"list_templates"` from working to erroring for every user whose ComfyUI is served at `http://localhost:<port>`.

**#2385's refusal was right; its *shape* was wrong.** This keeps the refusal and changes it from a failure into a decline.

## Reproduced by execution

Driving the real orchestrator wiring (`createPanelTemplateRelayWiring`) through the real `list_packs` handler, one variable changed — the panel host:

| | panel at `127.0.0.1:<p>` | panel at `localhost:<p>` |
|---|---|---|
| **0.52.135 (shipped)** | `RELAYED-OK` | `The connected panel template relay failed (NO_PANEL_ORIGIN).` |
| **this PR** | `RELAYED-OK` | completes on the headless path |

`127.0.0.1` is the control and survives both arms. In the broken arm the headless-fallback counter stays at **0**: `listWorkflowTemplatesAction` is fail-closed once a panel route exists, so there was no path left for the user.

## The design

`localhost` still **never becomes a relay fetch target** — it names an address family, not a listener, so with a second process on the other loopback family the fetch could reach the wrong one. `LOOPBACK_HOSTS` stays literal-addresses-only, exactly as #2385 left it.

What changes is what happens next. An **identical** ambiguous pair (`http://localhost:8188` on both sides) now returns `AMBIGUOUS_PANEL_ORIGIN`, which the child converts back to `undefined` — the already-documented *"this MCP child has no panel relay route"* — so the caller continues on its established headless path instead of raising.

**Why degrading is safe here and nowhere else:** the two operands are byte-identical, so `COMFYUI_URL` is provably the same origin string the panel is on. The rule this seems to bend — *"falling back to COMFYUI_URL could silently list a different server's templates"* — cannot apply, because by construction there is no different server. A **mixed** pair (`localhost` vs `127.0.0.1`) is a genuine mismatch and stays a hard `NO_PANEL_ORIGIN` failure.

## Gate round 1

The first revision simply re-admitted `localhost`, and the gate returned NO-SHIP on it, naming the wrong-listener fetch. That finding was correct about the mechanism, so I did not argue it — I removed the fetch instead. The relay no longer resolves an ambiguous name at all; only the *reporting* of that refusal changed.

## Where the load-bearing claims are — please attack these

1. **The "provably the same server" argument** is what licenses the fallback. It rests on `ambiguousLoopbackNameOrigin` requiring equal host, scheme and port. If that predicate can return true for two origins that are not the same server, the fail-closed rule is broken and this PR is wrong.
2. **Returning `undefined` for `AMBIGUOUS_PANEL_ORIGIN`** widens the meaning of a documented sentinel. I claim it is exactly the documented meaning; worth checking against every other caller of `requestPanelTemplateIndex()`.
3. **The headless path inherits the ambiguity** — `comfyuiFetch` to `http://localhost:8188` re-resolves the same name. This PR does not close that, and does not claim to: it is the pre-existing behaviour of every non-panel user and is out of scope here.

## Verification — mutation, clause by clause

Every clause was reverted **independently** against committed code, and each kills at least one test (16 tests across 3 files; the rest survive as controls):

| mutation | result |
|---|---|
| child no longer converts the decline to `undefined` | **2 failed** |
| server never emits `AMBIGUOUS_PANEL_ORIGIN` | **2 failed** |
| `resolveAmbiguousLoopbackOrigin` dropped from the orchestrator wiring | **2 failed** |
| name-set membership check removed | **3 failed** |
| scheme clause removed | **1 failed** |
| port clause removed | **1 failed** |
| `localhost` re-added to `LOOPBACK_HOSTS` | **3 failed** |

Two of these were found *because* a mutation survived, and both were fixed rather than explained away:

- **`observed.origin === target.origin` survived removal** — with a single-name `AMBIGUOUS_LOOPBACK_NAMES`, equal host + scheme + port already forces equal origins, so the clause could not fail. Removed, with the sufficiency argument written down and a note to restore a host comparison if the set ever grows. (`currentPanelTemplateOrigin` keeps its origin comparison — there it *is* load-bearing, because `LOOPBACK_HOSTS` has two members.)
- **The scheme clause survived** because no test varied it. Added https/http `localhost` pairs, which kill it.

Also: two earlier mutation attempts **never applied** (CRLF anchors, then an anchor that matched both functions) and would have read as "survived". Re-run with unique CRLF-aware anchors; the numbers above are from applied mutations only.

Other gates: full suite **680 files, 12688 passed, 6 skipped, 0 failed** (collection confirmed non-zero inside the worktree); `npm run build` clean; `npm run lint` clean (anti-slop "none added").

**No browser gate was run and none is claimed** — this is entirely orchestrator/service-side and renders nothing in the panel.

## On the issue

The underlying mechanism recorded in #2382 — that a browser Origin naming `localhost` cannot identify a listener — is unchanged by this PR; it is avoided rather than solved. I am reopening #2382 and marking it `autopilot:parked`, matching its reporter's own sizing: P3, no report of the failure occurring, and both "make localhost unambiguous" designs need a resolution the Origin header cannot supply.

### DIFF
--- a/src/__tests__/orchestrator/panel-template-relay-production.test.ts
+++ b/src/__tests__/orchestrator/panel-template-relay-production.test.ts
@@ -108,7 +108,7 @@ describe("orchestrator panel template relay wiring (#2196)", () => {
     expect(wiring.resolveAllowedPanelOrigin("tab-1", target)).toBeUndefined();
     expect(wiring.resolvePanelUrl("tab-1", target)).toBeUndefined();
 
-    for (const origin of ["http://127.0.0.1:8188", "http://[::1]:8188"]) {
+    for (const origin of ["http://127.0.0.1:8188", "http://[::1]:8188", "http://localhost:8188"]) {
       target = `${origin}/comfyapi`;
       observedOrigin = origin;
       expect(wiring.resolveAllowedPanelOrigin("tab-1", target)).toBe(origin);
@@ -127,18 +127,22 @@ describe("orchestrator panel template relay wiring (#2196)", () => {
     await expect(requestPanelTemplateIndex()).rejects.toMatchObject({ code: "NO_PANEL_ORIGIN" });
   });
 
-  it("refuses an exact localhost match before fetch can resolve another loopback listener", async () => {
+  // #2382/#2385. Dropping "localhost" from LOOPBACK_HOSTS made this whole
+  // request fail NO_PANEL_ORIGIN, which shipped in 0.52.135. The mixed-name
+  // case below is the part that must stay refused: it is a genuine mismatch,
+  // not merely a name.
+  it("serves a localhost-served panel and still refuses a mixed localhost/127.0.0.1 pair", async () => {
     let panelRequests = 0;
     const panel = createServer((_req, res) => {
       panelRequests += 1;
       res.writeHead(200, { "content-type": "application/json" });
-      res.end(JSON.stringify({ "unexpected-pack": [{ name: "wrong-listener" }] }));
+      res.end(JSON.stringify({ "panel-pack": [{ name: "localhost-template" }] }));
     });
     const panelOrigin = await listen(panel);
     servers.push({ close: () => closeServer(panel) });
     const port = new URL(panelOrigin).port;
-    const target = `http://localhost:${port}/comfyapi`;
-    const observedOrigin = `http://localhost:${port}`;
+    let target = `http://localhost:${port}/comfyapi`;
+    let observedOrigin = `http://localhost:${port}`;
     const bridge = {
       canReach: () => true,
       resolveFailure: () => undefined,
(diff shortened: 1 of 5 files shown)


artokun:
Reviewed independently, in my own worktree at `ac5828c`. **This is sound and better pinned than my #2387 was. I closed mine in favour of it.**

## The check I care about most

The defect the merge gate caught on my attempt was collapsing *ambiguity* into *disagreement*, so a post-retarget mismatch silently fell back to `COMFYUI_URL`. I mutated your classifier to reproduce exactly that:

    ambiguousLoopbackNameOrigin(...) { return true; }   // always ambiguous

    Tests  4 failed | 12 passed

So the fail-closed half is genuinely held, not incidentally green. That is the one property a future "simplification" of this function would destroy, and it now costs four red tests to do so.

Also verified from a clean checkout: `tsc --noEmit` exit 0, and **16/16** across `panel-template-relay.test.ts`, `panel-template-relay-production.test.ts` and the new `skills-access-panel-template-relay.integration.test.ts`.

## Why the integration test is the right call

The regression is a **caller** behaviour. `requestPanelTemplateIndex()` throwing only matters because `skills-access.ts:585` does not catch and its `COMFYUI_URL` fallback is unreachable on a throw. Driving the real `list_packs` handler pins the thing users actually hit; my relay-level tests would have stayed green through a caller-side reintroduction.

## One deliberate coupling, already documented

`ambiguousLoopbackNameOrigin` compares protocol and port but not origin equality, relying on both hosts being in `AMBIGUOUS_LOOPBACK_NAMES` — which today holds only `localhost`, so equal hosts follow. Your comment already says to add `observed.host === target.host` if a second name is ever added. That is the right place for that note; flagging only so a reviewer does not read it as an oversight.

## Severity, since it affects how fast this should land

This is live on npm, not pending:

    npm view comfyui-mcp version                  -> 0.52.135
    git merge-base --is-ancestor 38c6f2e v0.52.135 -> passes

Every sidebar-backed user served at `http://localhost:<port>` has a `list_packs action:"list_templates"` that errors right now. I have reopened **#2382** at **P1** to reflect that — I originally filed it at P3, and that sizing did not survive the fix shipping.

Gating now.

artokun:
Closing as superseded by #2387. Independent exact-head review found a P1: converting ambiguous localhost relay refusal to undefined still falls through to getComfyUIBaseUrl(), which can resolve to the wrong IPv4/IPv6 listener, and decline paths lack retarget fencing. #2387 ships the fail-closed typed refusal. Please keep the underlying localhost usability/design issue open for a safe listener-identity solution.

artokun:
**Gate came back NO-SHIP, and having checked it, I think it is right — including against the argument I made for this design in my review above.**

> [P1] `panel-template-relay.ts:736` — `AMBIGUOUS_PANEL_ORIGIN` falls back without verifying the child's target identity — during a retarget, the child may still use its old `COMFYUI_URL`, so `localhost` can list templates from a different loopback listener.

## Why it holds

My whole case for falling through was: *in the ambiguous case the panel origin and the current target are the same string, so the fallback reads the very origin we declined.* That is true **within the orchestrator**. It is not true of the **child**, and the child is who performs the fallback.

    orchestrator:  createPanelTemplateRelayWiring({ currentTarget: getComfyUIBaseUrl, ... })
    child:         const base = getComfyUIBaseUrl();     // skills-access.ts

Same function name, **different process**, each with its own module-level config initialised from its own argv/env. They are only equal if nothing has retargeted since the child was spawned — and this relay tracks `currentTargetGeneration` precisely because retargeting happens.

So: child spawned against target A, orchestrator retargets to `localhost:<p>`, relay declines it as ambiguous, child falls back to **A**. Not the declined origin — a different listener, which is exactly the outcome the fail-closed comment in `skills-access.ts` names.

I asserted "same target" without measuring it, which is the second time tonight I have made that class of error on this issue. Recording it plainly because the pattern matters more than the instance: *two things reached by the same expression are not the same value when they live in different processes.*

## The remedy is small

The child cannot compare against an origin it was never told. Carry it: include the declined origin in the `AMBIGUOUS_PANEL_ORIGIN` failure payload, and fall through **only** when the child's own `getComfyUIBaseUrl()` resolves to that same origin. Anything else keeps throwing.

That preserves the fix for the ordinary case — panel and child both on `http://localhost:<port>`, nothing retargeted — and closes the race the gate found, without reopening the mismatch path.

## Standing

The mutation evidence in my review still holds: forcing `ambiguousLoopbackNameOrigin` true fails 4 tests with 12 surviving, so the ambiguity/mismatch split is sound. This finding is about a **third** party — the child's own notion of the target — that neither of our versions checked.

#2382 stays **P1 and live on npm** (`0.52.135`), so this is still worth landing quickly, just with the identity check added. Happy to take that on if you would rather not — say the word, otherwise it is yours.

PR #2390, opened 2026-08-26
chore: release v0.52.136

artokun:
## Summary\n- bump MCP package metadata to 0.52.136\n- record the merged #2387 fail-closed panel-origin fix\n\n## Validation\n- exact three-file release diff\n- hosted CI required before merge\n\nCloses release of the merged #2387 fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+## [0.52.136] - 2026-08-26
+
+### MCP
+
+#### Fixed
+- keep refused panel origins fail-closed instead of falling through to a stale target (#2382, #2387)
+
 ## [0.52.135] - 2026-08-26
 
 ### MCP
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "comfyui-mcp",
-  "version": "0.52.135",
+  "version": "0.52.136",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "comfyui-mcp",
-      "version": "0.52.135",
+      "version": "0.52.136",
       "license": "MIT",
       "dependencies": {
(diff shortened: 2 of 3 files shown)


artokun:
Structurally this cut is clean — one `#### Fixed` block, one entry, and it carries the only unreleased commit. Two accuracy notes, neither blocking.

## The entry describes a regression that never shipped

> keep refused panel origins fail-closed instead of falling through to a stale target (#2382, #2387)

"instead of falling through" implies a released version fell through. None did:

    v0.52.133  fall-through constructs: 0
    v0.52.134  fall-through constructs: 0
    v0.52.135  fall-through constructs: 0

The fall-through existed only in an unmerged branch (`f3145de`, mine) for about twenty minutes. A user reading this reasonably concludes 0.52.135 had the bug and upgrades to escape it — the same shape as the 0.52.134 entry that claimed a fix which had already shipped, and the fourth instance of what #1891 tracks in the panel repo.

## 0.52.136 has no behaviour change

`e50c067` touches the service with **9 added lines, all of them comments**:

    added lines: 9
    containing a comment marker: 9

The rest is the integration test. So 0.52.136 is functionally identical to 0.52.135, which makes this the second no-op release after 0.52.134.

That is not an argument against cutting it — the integration test is worth having on `latest`, and there is no cost to users. It is an argument for the entry saying what actually changed.

## Suggested wording

```
#### Changed
- pin the fail-closed refusal for unauthorizable panel origins with a
  caller-level test, and document why falling through to getComfyUIBaseUrl()
  is unsafe during a retarget (#2382, #2387)
```

`Changed` rather than `Fixed`, because nothing behaved differently before.

## Still open behind this

**#2382 is P1 and unresolved.** The safety decision is settled and correct, but a sidebar-backed user served at `http://localhost:<port>` still gets an error from `list_packs action:"list_templates"` on published `latest`. The cheap half — making the refusal name `localhost` as the cause and `127.0.0.1` as the remedy — would take it to P3 on its own and does not reopen any of the argument.

PR #2391, opened 2026-08-27
fix(2382): pin the relay fetch to literal loopback addresses so localhost works again

artokun:
Takes option **(2)** from #2382's latest comment — *"remove the ambiguity instead of refusing it"* — proposed when the issue was filed and unimplemented through #2385, #2387 and #2388.

## Still live, now in two releases

`list_packs action:"list_templates"` returns `isError` with `NO_PANEL_ORIGIN` for every user whose ComfyUI is served at `http://localhost:<port>`. Verified **by execution against a clean checkout of `e50c067`**, driving the real `list_packs` handler:

```
CONTROL 127.0.0.1 -> {"isError":false,"outcome":"RELAYED-OK","headless":0}
LOCALHOST         -> {"isError":true,"outcome":"The connected panel template relay failed (NO_PANEL_ORIGIN).","headless":0}
```

#2387 shipped nine comment lines and a test; it changed no behaviour on this path, and has now also gone out in **0.52.136**.

## The mechanism is real, and reproduces on this machine

```
dual-bind same port on ::1 = true
  127.0.0.1 -> 4
  [::1]     -> 6
  localhost -> 6     <-- fetch picked the OTHER listener
```

Not theoretical — and why simply re-admitting `localhost` (my #2388) was correctly rejected.

## What this does

`http://localhost` authorizes an origin again, but **is never a fetch destination**. The name is resolved **once**, in `pinnedLoopbackUrls`, and every request goes to a literal address, so `fetch` performs no second, independent resolution. That is where the ambiguity actually lived.

Where the name resolves to several loopback addresses, identity is not knowable — Origin carries no family, and a dual-stack bind on `::` is indistinguishable from two processes. So identity is not what gets established: the candidates must **agree**. Same index from every address that has a listener means the answer is the same whichever socket the panel is on. Disagreement raises `AMBIGUOUS_PANEL_LISTENER` rather than guessing.

**A listener that answers badly is still a listener.** Only a *refused connection* counts as absent. A real panel returning 503 on one family must not let another process's index win on the other — that is the same wrong-listener bug wearing a different hat.

**`https://localhost` stays refused, exactly as main has it today.** Over TLS you cannot both pin the address and keep verification of a certificate issued to the name; the two gate findings that chased this were really the same irreducible tension. Since main already refuses *all* localhost origins, not restoring the https half regresses nobody — and the reported defect is the http one.

**Fail-closed is untouched** (#2387's finding stands and I agree with it): a refusal still does not fall through to `getComfyUIBaseUrl()`, because a mid-turn child keeps the pre-retarget target (#1429). The fetch stays here, under the relay's generation fence. A mixed `localhost`/`127.0.0.1` pair is still a hard refusal.

**Bonus, predating all of this:** only the *origin string* was ever checked against the loopback set, never the address it resolves to — so a hosts-file entry pointing `localhost` off-box would have been fetched. Every resolved address must now be loopback, or the whole name is refused.

1. **"Agreement is sufficient."** If every address with a listener returns byte-identical bodies, returning that body is correct regardless of which one the browser is on. A counterexample kills this PR.
2. **Refusing the whole name when one resolved address is off-loopback** — stricter than honouring the loopback half, and a behaviour change for an exotic hosts file.
3. **`isConnectionRefused` walks `cause`/`errors`.** If a Node version reports a refused loopback connect under a code not in that list, a dead address would be misread as a failing listener and refuse where it should serve.

## Verification — mutation, clause by clause

Every guard reverted independently against committed code (22 tests across 4 files; the rest survive as controls, including #2387's retarget test, which I did not write):

| mutation | result |
|---|---|
| `localhost` removed from `LOOPBACK_HOSTS` (main today) | **7 failed** |
| https name authorized again | **1 failed** |
| pin re-check drops the scheme clause | **1 failed** |
| agreement check removed | **1 failed** |
| failing listener treated as absent | **1 failed** |
| pinning removed (name fetched directly) | **1 failed** |
| off-loopback resolution allowed | **2 failed** |

Several of these only became real after fixing the test meant to catch them:

- **The off-loopback guard survived** at first — no test could steer DNS. Added a file that mocks `node:dns` so the real `pinnedLoopbackUrls` runs against a chosen answer.
- In that file the request counter was **snapshotted at harness return**, so both `expect(...).toBe(0)` assertions passed no matter what the code did. Exposed as a getter.
- A partial-failure test used a 3-way `toContain`, which could not meaningfully fail. Replaced with a real dual-bind where one listener 503s.
- Several early mutations **never applied** (CRLF anchors; an anchor matching both functions) and would have read as "survived". The table is from applied mutations only.

Other gates: `npm run build` clean; `npm run lint` clean (anti-slop "none added"); full suite **681 files, 12694 passed, 6 skipped**. Two earlier full runs each had one failure in `late-mutation-e2e.test.ts` — the documented 60 ms real-bridge load flake (#694/#852), untouched by this diff and **5/5 passing in isolation**.

**No browser gate was run and none is claimed** — orchestrator/service-side, renders nothing in the panel.

## History

#2385 refused `localhost` (regression). #2387 documented why falling through is unsafe (correct, kept). #2388 tried re-admitting it, then declining to headless — the gate correctly killed both. This removes the second resolution instead, which is what the issue asked for on day one.

### DIFF
--- a/src/__tests__/orchestrator/panel-template-relay-production.test.ts
+++ b/src/__tests__/orchestrator/panel-template-relay-production.test.ts
@@ -108,7 +108,7 @@ describe("orchestrator panel template relay wiring (#2196)", () => {
     expect(wiring.resolveAllowedPanelOrigin("tab-1", target)).toBeUndefined();
     expect(wiring.resolvePanelUrl("tab-1", target)).toBeUndefined();
 
-    for (const origin of ["http://127.0.0.1:8188", "http://[::1]:8188"]) {
+    for (const origin of ["http://127.0.0.1:8188", "http://[::1]:8188", "http://localhost:8188"]) {
       target = `${origin}/comfyapi`;
       observedOrigin = origin;
       expect(wiring.resolveAllowedPanelOrigin("tab-1", target)).toBe(origin);
@@ -127,17 +127,23 @@ describe("orchestrator panel template relay wiring (#2196)", () => {
     await expect(requestPanelTemplateIndex()).rejects.toMatchObject({ code: "NO_PANEL_ORIGIN" });
   });
 
-  it("refuses an exact localhost match before fetch can resolve another loopback listener", async () => {
+  // #2382/#2385. Dropping "localhost" from LOOPBACK_HOSTS made this request
+  // fail NO_PANEL_ORIGIN, which shipped in 0.52.135 and is still live: #2387
+  // documented the fail-closed rule but changed no behaviour here. Authorizing
+  // the origin — rather than declining it — is also what keeps the fetch under
+  // the relay's own generation fence, instead of handing it to a child that may
+  // still hold the pre-retarget target (#1429).
+  it("serves a localhost-served panel and still refuses a mixed localhost/127.0.0.1 pair", async () => {
     let panelRequests = 0;
     const panel = createServer((_req, res) => {
       panelRequests += 1;
       res.writeHead(200, { "content-type": "application/json" });
-      res.end(JSON.stringify({ "unexpected-pack": [{ name: "wrong-listener" }] }));
+      res.end(JSON.stringify({ "panel-pack": [{ name: "localhost-template" }] }));
     });
     const panelOrigin = await listen(panel);
     servers.push({ close: () => closeServer(panel) });
     const port = new URL(panelOrigin).port;
-    const target = `http://localhost:${port}/comfyapi`;
+    let target = `http://localhost:${port}/comfyapi`;
     const observedOrigin = `http://localhost:${port}`;
     const bridge = {
       canReach: () => true,
@@ -151,14 +157,25 @@ describe("orchestrator panel template relay wiring (#2196)", () => {
       currentTargetGeneration: () => 0,
       secrets: new Map([[SECRET, "orchestrator::codex"]]),
     });
-    expect(wiring.resolveAllowedPanelOrigin("tab-1", target)).toBeUndefined();
-    expect(wiring.resolvePanelUrl("tab-1", target)).toBeUndefined();
+    expect(wiring.resolveAllowedPanelOrigin("tab-1", target)).toBe(observedOrigin);
+    expect(wiring.resolvePanelUrl("tab-1", target)).toBe(`${observedOrigin}/comfyapi/api/workflow_templates`);
+
     const relay = await startPanelTemplateRelayServer({ bridge, ...wiring });
     servers.push(relay);
     process.env.COMFYUI_MCP_RELAY_SECRET = SECRET;
     process.env.COMFYUI_MCP_TEMPLATE_RELAY_URL = relay.endpointUrl;
+    await expect(requestPanelTemplateIndex()).resolves.toMatchObject({
+      "panel-pack": [{ name: "localhost-template" }],
+    });
+    expect(panelRequests).toBe(1);
+
+    // A MIXED pair is a genuine mismatch, not a name ambiguity, and must stay a
+    // hard refusal that never reaches the panel.
+    target = `http://127.0.0.1:${port}/comfyapi`;
+    expect(wiring.resolveAllowedPanelOrigin("tab-1", target)).toBeUndefined();
+    expect(wiring.resolvePanelUrl("tab-1", target)).toBeUndefined();
     await expect(requestPanelTemplateIndex()).rejects.toMatchObject({ code: "NO_PANEL_ORIGIN" });
-    expect(panelRequests).toBe(0);
+    expect(panelRequests).toBe(1);
   });
 
   it("rejects a stale in-flight response after retargeting and still serves the current target", async () => {
--- a/src/__tests__/orchestrator/panel-template-relay-production.test.ts
+++ b/src/__tests__/orchestrator/panel-template-relay-production.test.ts
@@ -316,4 +316,58 @@ describe("orchestrator panel template relay wiring (#2196)", () => {
     expect(hits.length).toBeGreaterThan(0);
     for (const host of hits) expect(host).not.toContain("localhost");
   });
+  // #2382 — a listener that answers BADLY is still a listener. The pinned fetch
+  // must not quietly prefer whichever address happened to return valid JSON:
+  // the erroring one could be the real panel, and then we would be serving the
+  // other process's index — the exact bug pinning exists to prevent.
+  it("refuses when one loopback listener 503s and another returns a valid index", async () => {
+    const v4 = createServer((_req, res) => {
+      res.writeHead(200, { "content-type": "application/json" });
+      res.end(JSON.stringify({ "v4-pack": [{ name: "from-127" }] }));
+    });
+    await new Promise<void>((resolve, reject) => {
+      v4.once("error", reject);
+      v4.listen({ host: "127.0.0.1", port: 0 }, () => resolve());
+    });
+    const addr = v4.address();
+    if (!addr || typeof addr === "string") throw new Error("no bind");
+    const port = addr.port;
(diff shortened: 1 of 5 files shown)


artokun:
I filed #2382 and proposed this option when I filed it. Reviewed independently at `86d12f4`, in my own worktree. **This is the right shape and it is properly pinned.**

## Verified from a clean checkout

    tsc --noEmit          exit 0
    4 relay/integration suites   22 passed | 0 failed

**Mutation — the pinning is load-bearing.** Forcing `pinnedLoopbackUrls` to return the URL untouched, so the fetch goes to the name as given:

    Tests  4 failed | 18 passed

So "the request goes to a literal, never a name" is held by tests rather than by intention. That is the property the whole fix rests on, and it is the one a later refactor would quietly undo.

## Why this beats both earlier attempts

#2385, #2387 and #2388 all argued about *what to do when the origin is ambiguous* — refuse, fall through, or classify. This removes the ambiguity instead. There is no second resolution to disagree with the first, so:

- the retarget race that killed the fall-through approach cannot arise — there is no fallback;
- the fail-closed property #2387 documented is untouched, because a mismatch still never resolves to an allowed origin;
- and `localhost` users keep the feature.

It is strictly stronger than today's behaviour, not a relaxation: on `main` the fetch could still be re-resolved by the OS at request time even for an accepted origin. After this it cannot.

## Two details worth calling out as correct

**HTTPS deliberately keeps the name.** Rewriting the host to a literal would break SNI and certificate validation — a pin that "worked" there would be a downgrade. The comment says so at the site.

**A failing listener is treated as present** (`86d12f4`). That is the right default: a loopback family that refuses a connection is evidence it exists and something is listening on the other one, not evidence it can be skipped.

## On the demonstration

The control/localhost pair in the PR body is the thing that makes this reviewable — one variable changed, both outcomes shown, run against a clean checkout of `e50c067`. That is more convincing than any amount of reasoning about `dns.lookup`, and it is what I should have produced when I filed the issue instead of a severity argument.

Gating now. #2382 is P1 and live on published `latest` in two releases, so this is worth landing promptly.

artokun:
Gate returned **NO-SHIP [P1]** on the HTTPS carve-out — the one I praised as correct in my review above. Having checked it, **the gate is right about the mechanism and the code comment's justification is wrong**, though I do not think it should block this PR.

> HTTPS `localhost` bypasses DNS pinning and listener-agreement checks — when `https://localhost:<port>` resolves to both loopback families with different valid TLS listeners, the relay fetches whichever wins and can return the wrong template index.

## Where the comment is wrong

    // HTTPS keeps the NAME, deliberately. Rewriting the host to a literal breaks
    // certificate verification ... TLS already binds the response to the name, which is
    // a stronger listener check than pinning can give us.

The first half is right — rewriting the host to a literal genuinely breaks verification of a cert issued to `localhost`. The second half does not follow. **TLS binds the response to the NAME, not to the LISTENER.** If `127.0.0.1:<p>` and `[::1]:<p>` both present a valid cert for `localhost` — which is precisely what one shared dev cert (mkcert and friends) produces — then both handshakes succeed and TLS cannot distinguish which socket answered. It is not a stronger listener check; on this specific question it is not a listener check at all.

So the carve-out is a real gap, not a place where TLS covers for us. `https` reaches it: `parsedHttpOrigin` admits both schemes (`:442`, `:486`), so a loopback HTTPS panel is a permitted configuration.

## Why I would still land this

The gap needs **two loopback listeners on the same port, both holding a valid cert for `localhost`** — a second local process, which is the adversary this project has explicitly decided not to defend against. Blocking a P1 regression fix on an out-of-threat-model edge is the same trade I argued against on #2385, and it was wrong then in the other direction: the cleartext path is broken for ordinary users on published `latest`, right now, in two releases.

The PR also strictly improves the HTTPS case relative to `main`, because nothing about it gets worse — it is simply not improved.

## The clean resolution, if you want it closed properly

Pinning and cert validation are not actually in tension. Pin the **connection address** while keeping the **name** for SNI and verification, rather than rewriting the URL host:

    new Agent({ connect: { lookup: (host, opts, cb) => cb(null, pinnedAddress, family) } })

The request still says `Host: localhost` and validates the cert against `localhost`; only the socket is pinned. That gets the listener guarantee without touching TLS semantics, and would let the carve-out disappear.

## Suggested minimum for this PR

Correct the comment so it does not assert a guarantee TLS is not giving. Something like: *"HTTPS keeps the NAME because rewriting the host breaks certificate verification. This leaves the second resolution in place for `https://localhost`, which is a known gap — closing it needs connection-level pinning with the name preserved for SNI, tracked in #NNNN."*

A wrong reason in a comment is worse than no reason: the next reader treats it as settled and does not re-examine it. I did exactly that when I called the carve-out correct in my review.

Happy to open the follow-up issue and take the undici `connect.lookup` change myself if you would rather land this as-is.

artokun:
Filed the HTTPS gap as **#2392** (P3) so the carve-out is tracked rather than left as a settled-looking comment, and so this PR is not blocked on it.

It carries the sizing, the reachability check, and the concrete fix — pin the connection address via undici's `connect.lookup` while keeping the name for SNI, which lets the `https:` branch be deleted instead of documented. Happy to take that one.

Suggested minimum here remains the comment correction, so the next reader does not inherit the reasoning I did.

artokun:
Re-reviewed at `529cffc`. **The corroboration design answers the gate's finding well, and the new HTTPS branch is the one part of it that has no test.**

## The design is right

Fetching the literals to *contradict* rather than to *replace* is the correct move once you accept TLS needs the name. And the two guard rails around it are the parts I would have got wrong:

- a literal that fails — refused, or a cert not issued to that IP, which is the normal case — proves nothing and is ignored, so an ordinary `https://localhost` is not broken by this;
- on the cleartext path, **only** a refused connection is safe to ignore, because "prefer whichever address answered" would serve the other listener's index — the exact defect the function exists to prevent.

The comment is also corrected: it no longer claims TLS is a listener check.

## The gap: `:593` is unpinned

There are four `AMBIGUOUS_PANEL_LISTENER` sites. The two tests exercise the **cleartext** branch (`:634`, `:642`). The **https corroboration** check is `:593`, and nothing covers it:

    // replaced the condition with `false` — i.e. never detect a mismatch
    others.some((other) => other !== undefined && other !== body)  ->  false

    Tests  13 passed | 0 failed

Mutation confirmed applied by needle count before running, so this is a real survival rather than a no-op edit. Disabling the https mismatch detection entirely breaks nothing.

That matters more than a normal coverage gap, because `:593` **is** the answer to the gate's P1. Right now the PR's response to that finding is load-bearing prose plus untested code, and the failure mode is silent: it would keep returning the primary body and nobody would know the corroboration had stopped working.

## Suggested test

Two loopback listeners on the same port, both answering `https` with a **trusted** cert for `localhost` and **different** indexes; assert `AMBIGUOUS_PANEL_LISTENER`. The awkward part is the trusted cert — but the existing tests already stand up real servers, so a self-signed cert plus `NODE_EXTRA_CA_CERTS` or a scoped `rejectUnauthorized` shim in the harness is enough. The asymmetric case is worth a second one: literal fails (untrusted cert), primary succeeds, and the result must still be the primary body rather than an error — that is the "ordinary https localhost is never broken" promise, and it is currently only a claim in a comment.

## Otherwise

`tsc --noEmit` exit 0; **22/22** across the four relay/integration suites at this head. The cleartext path — which is the live P1 for ordinary users — is pinned and mutation-tested from my earlier review, and nothing in these two commits weakened it.

I would still land this ahead of the https test if you want the regression off `latest` sooner; the untested branch is strictly better than the unpinned name it replaced. But it should not stay untested, and #2392 is the natural place to carry it if it does not land here.

PR #2398, opened 2026-08-27
fix(codex): retry active panel reads once

### LINKED ISSUE #2395: Retry transient panel read transport failure after entering subgraph
An idempotent panel_graph_outline read failed with a local orchestrator HTTP transport send error immediately after panel_enter_subgraph. Repeating the identical read once succeeded and returned the full subgraph. This indicates a transient transport/connection race rather than a graph error.

### Steps to reproduce
1. Call panel_enter_subgraph on a live workflow.
2. Immediately call panel_graph_outline.
3. Observe an HTTP send failure to the local orchestrator endpoint.
4. Repeat panel_graph_outline; it succeeds without any graph or server change.

### Exact error (scrubbed)
Transport send error: HTTP request failed while sending to http://127.0.0.1:9198/orchestrator::codex

### Fix
applied locally: no
upstream-only: yes
For read-only panel calls, retry one transient local transport send failure after confirming the orchestrator connection is still active. Preserve the original error if the retry also fails.

### Environment
Windows 10 / ComfyUI 0.34.0 / ComfyUI frontend 1.51.9 / NVIDIA RTX 5060 Ti 16 GB / comfyui-mcp 0.52.130 / panel 0.15.102

artokun:
## Summary

Fixes #2395

- Trace the outer Codex app-server WorkerTransport failure to the correlated panel MCP item after panel_enter_subgraph.
- Allow one in-band retry only for panel_graph_outline and panel_query_graph in the proved immediate post-enter sequence while the app-server connection remains active.
- Invalidate that sequence on any intervening panel request; keep mutations and unrelated reads on the existing fence/reload path.
- Keep per-item failures isolated and preserve the original saved read transport error when a distinct mutation fails.
- Retain cancellation, timeout, connection-exit, and no-amplification semantics.

## Validation

- Focused production-path tests: 2 files / 41 passed
- Orchestrator suite: 256 files / 4,379 tests passed
- `npm.cmd run build` passed
- `npm.cmd run lint` passed
- `git diff --check` passed

No merge or release performed.

### DIFF
--- a/src/__tests__/orchestrator/codex-mcp-reconnect.test.ts
+++ b/src/__tests__/orchestrator/codex-mcp-reconnect.test.ts
@@ -57,6 +57,7 @@ const MCP_SERVERS_WITH_UNRELATED_HTTP_UP = [
 const WORKER_TRANSPORT_FAILURE =
   "Transport send error: WorkerTransport<StreamableHttpClientWorker<...>> error: " +
   "HTTP request failed sending request to http://127.0.0.1:9198/orchestrator%3A%3Acodex";
+const MUTATION_WORKER_TRANSPORT_FAILURE = `${WORKER_TRANSPORT_FAILURE} (mutation item)`;
 // The exact wrapper reported by panel_save_workflow -> immediate panel_run on
 // the affected EventNotificationTransport path. It is an outer HTTP-MCP client
 // failure, not a panel tool result; handling it must not retry panel_run.
@@ -122,6 +123,8 @@ interface Drive {
       | "panel-read-retry-timeout"
       | "panel-read-retry-cancelled"
       | "panel-mutation-no-retry"
+      | "panel-mutation-before-read-no-retry"
+      | "panel-read-retry-mutation-error"
       | "panel-unrelated-read-no-retry"
       | "panel-read-no-enter-no-retry",
   ): Promise<void>;
@@ -246,6 +249,8 @@ function startDrive(opts: {
       if (
         kind.startsWith("panel-read-retry-") ||
         kind === "panel-mutation-no-retry" ||
+        kind === "panel-mutation-before-read-no-retry" ||
+        kind === "panel-read-retry-mutation-error" ||
         kind === "panel-unrelated-read-no-retry" ||
         kind === "panel-read-no-enter-no-retry"
       ) {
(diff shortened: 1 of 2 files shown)


artokun:
Independent gate: NO-SHIP at exact head 60174bb38f5a75fd2d75aa7e812915fe9e4f4b1f. Concrete blockers: retry classification uses one global activePanelTool slot without correlating the failure to MCP item ID; the watchExit path can bypass/presume over the saved original transport error; the allowlist includes panel_query_graph but production/negative coverage does not prove both reads are bounded to the intended lifecycle after panel_enter_subgraph, and timeout/parallel-item cases are missing. Hosted CI is green, but mutation safety and original-error preservation must be fixed before merge. No init.py/apps_routes.py network changes.

artokun:
Reviewed at `60174bb` in my own worktree. **The design is right and one invariant it states is not actually enforced.**

## What is right

Restricting the retry to a two-name allowlist rather than retrying on transport-failure shape is the correct call, and the comment states the rule exactly:

    // Keep this allowlist intentionally narrower than the panel tool catalog: scope
    // navigation and every mutation must never be replayed from an opaque outer
    // transport error.

Preserving the original transport error on retry failure, and keeping cancellation/timeout semantics, are the details that make a retry safe to reason about.

`tsc --noEmit` exit 0; `codex-mcp-reconnect.test.ts` **23 passed / 0 failed**.

## The gap: the allowlist's narrowness is unpinned

**Mutation — added `panel_set_widget` to `PANEL_OUTER_TRANSPORT_RETRY_SAFE_TOOLS`:**

    Tests  23 passed | 0 failed      (unchanged)

Needle count verified before the edit, so the mutation applied. A mutating tool can be added to the retry allowlist and **no test objects**.

The reason is visible in the test that looks like it covers this. `does not retry a mutating panel tool after a transport failure (#2395)` drives `panel_set_property` — a name that was never in the set. So it pins *"this particular tool is not retried"*, not *"the set contains only idempotent reads"*. The stated invariant and the tested property are different things.

That matters more here than a usual coverage gap, because the failure it guards is a **replayed mutation from an opaque transport error** — the one outcome the comment says must never happen, and one that is silent when it does: the second write succeeds and looks like the first.

## Suggested pin

Assert on the set itself rather than on a sample:

```ts
expect([...PANEL_OUTER_TRANSPORT_RETRY_SAFE_TOOLS]).toEqual(["panel_graph_outline", "panel_query_graph"]);
```

Blunt, but it makes the allowlist a decision someone has to change a test to make, which is exactly the friction the comment is asking for. A softer version — every member must match a known read-only predicate — is nicer if one exists, but the exact-membership assertion is the one that cannot drift.

Nothing else blocking. The behaviour is right; it is the guard rail around future edits that is missing.

artokun:
Re-checked on the moved head `84973de` rather than assuming my earlier result carried. **The allowlist gap is still open.**

    baseline   29 passed | 0 failed     (up from 23 — the correlation work added 6)
    mutant     29 passed | 0 failed     (added `panel_set_widget` to PANEL_OUTER_TRANSPORT_RETRY_SAFE_TOOLS)

Needle count verified before the edit. Six new tests went in and none of them constrains the set's membership, so a mutating tool can still be added to the retry allowlist without any test objecting.

The request-correlation commit is a genuine improvement — correlating the retry to the originating request closes a different hole, where a retry could be matched to the wrong in-flight item. It just does not touch this one.

The suggestion stands and is one line:

    expect([...PANEL_OUTER_TRANSPORT_RETRY_SAFE_TOOLS]).toEqual(["panel_graph_outline", "panel_query_graph"]);

Worth it because the comment above that set states the invariant as absolute — *"scope navigation and every mutation must never be replayed from an opaque outer transport error"* — and a replayed mutation is silent when it happens: the second write succeeds and looks like the first.

artokun:
Second independent gate: NO-SHIP at exact head 84973de70f62d8a76163674aad9f443d19932170. Remaining blockers: retry eligibility reads stale lastCompletedPanelItemId, so a panel read can qualify after panel_enter_subgraph even while a later mutation is still in flight; invalidate the lifecycle at request start or otherwise prove the immediate sequence. Also isolate keyed item errors: a distinct mutation transport error must not clear/replace the saved read transport error. CI is still running; no init.py/apps_routes.py network changes.

artokun:
Final gate: SHIP at exact head 7209a7fe6306f366322a4c4421bb2ba4913e1139. Independent review confirms per-item correlation, one-shot immediate lifecycle proof, mutation invalidation/isolation, original-error preservation, bounded retry, cancellation/timeout/no amplification, and production-path tests. Hosted Ubuntu, Windows, macOS, and pack/install checks are all green. No init.py/apps_routes.py network changes. Proceeding to merge.

PR #2400, opened 2026-08-27
fix(2393): a promoted write is judged on its OWN witness entry, not the whole array

artokun:
Re #2393. That issue was auto-closed by #2399 (`8abbbf6`), but **the cause is still on main** — see the mutation evidence below. This is rebased on top of that merge and is complementary to it, not a competing fix.

### What refuses the write

`panel_set_widget` on a promoted subgraph widget refuses with **"the promoted-terminal witness was incomplete or unresolved"** (or "was ambiguous"), no `graph_set_widget` dispatched. `promotedTerminalEvidenceError` vetoes on the **whole array** — any repeated widget name, or any `error` entry, anywhere in `promoted_terminals`, including entries about a completely different alias:

```ts
if (seen.has(key)) return "the promoted-terminal witness was ambiguous";
seen.add(key);
if (witness.error) return "the promoted-terminal witness was incomplete or unresolved";
```

So `enable_turbo_mode` on subgraph node 170 is refused even though its own witness entry resolves cleanly.

### Why a retry cannot reach it

#2399 added a single fresh `graph_get_subgraph` gated on `promotedTerminalWitnessNeedsRefresh`, premised on the incompleteness being transient. For this template the array is identical on every read:

1. **A guaranteed duplicate.** Node 170s `proxyWidgets` carries `["151","prompt"]` **and** `["149","prompt"]` — two inner nodes whose inner widget is named `prompt` — while exactly one host input claims that alias (`prompt`/`positive_prompt`; the negative one is `prompt_1`/`negative_prompt`, which does not match `prompt`). In the panel, `promotedHostAliasRecords` `existingUnbound` fast path only matches a record with **no relation yet**, so the first relation binds and the second **appends**; `promotedTerminalWitnesses` publishes one entry per record with **no de-duplication**.
2. **Unmatched relations.** `lora_name`, `seed` and `control_after_generate` are named in `proxyWidgets` with no matching host input — node 170 serializes 9 of the definitions 11 input slots, and `control_after_generate` is not a subgraph input at all — adding `error` entries in the shape where the parent rails are not projected.

Measured, not inferred: I extracted the panels **real** `promotedHostAliasRecords` verbatim from the shipped bundle, drove it with the real `promotedInputAliases`, and ran it over node 170 as serialized in the shipped `image_qwen_image_edit_2511_int8.json`:

| `node.widgets` | records | error entries | duplicated aliases | requested alias clean? |
|---|---|---|---|---|
| unprojected | 21 | 3 (`lora_name`, `seed`, `control_after_generate`) | 2 | yes |
| projected (6 rails) | 112 | 0 | 12 | yes |

### The change

Decide on the **requested alias**:

- exactly one entry for it, no error → authorized on its own evidence
- duplicated, or its own entry errors → refused, as before
- **absent** → the whole-array rule is **unchanged**

### The load-bearing part — attack this first

The third case is the whole point of the original rule and it is untouched. A **miss** is read as "ordinary widget, write it outer", so a miss can only be trusted from an array that is whole; narrowing it would revive the old same-name scan into an outer write, which is the P0 that comment protects. Evidence about *another* alias was never evidence about *this* one — that is the only inference removed here.

Also worth attacking: this removes exactly one veto. Everything downstream is intact — `resolvePromotedWriteTarget` must still resolve the target, `inner.parentRail` must still be proven authoritative, `currentPromotedScopeError` and the dispatch fences still run.

Second: structural damage (not an array; an entry that is not an object or has no usable widget name) still fails closed, but **not here** — `parsePromotedTerminalEntries` rejects those first and `validatePromotedSubgraphEnvelope` refuses the envelope, so the structural pass inside this function is unreachable on the write path. I kept it (removing it would be an unrelated behaviour change) and the tests deliberately pin the refusal that is **actually observed** rather than the one this function would have produced.

### Verification

Rebased onto `8abbbf6`; both changes coexist with no conflict. `npx tsc --noEmit` clean. `promoted-widget.test.ts`: **126/126** (my 8 plus the 3 #2399 added). Full suite before the rebase: 12703 passed, 1 pre-existing `late-mutation-e2e` load flake that passes in isolation and is untouched by this diff.

**Mutation-verified on top of #2399s merged fix** — neuter only the one-line call-site change:

```
KILLED: writes when a DIFFERENT alias is duplicated
KILLED: writes when a DIFFERENT alias carries an error
Tests  2 failed | 124 passed (126)
```

If the merged refresh covered this case those two would have survived. They do not, and all 124 controls — including #2399s three tests — stay green.

Full battery:

| mutation | killed |
|---|---|
| remove the own-entry decision | both allow-tests; 121 controls survive |
| stop passing `widget` at the call site | both allow-tests; 124 controls survive (post-rebase) |
| ignore an error on the requested aliass own entry | that guard test |
| remove the whole-array MISS fallback | all three miss/duplicate guard tests |

An explicit `own.length > 1` branch I first wrote **survived** mutation — the fallback loop already refuses that — so it was removed rather than left looking load-bearing.

Codex gate on the rebased SHA `713d83e`: **SHIP**, "No P0/P1 correctness or safety defect found."

### What I did NOT do

**No live-rig verification.** ComfyUI on :8188 went down mid-session and this run had no instance to allocate, so `panel_set_widget` against the real template was never executed — by this PR or by #2399. Everything above is from the shipped panel bundle and the shipped template, not from a live write. The duplicate-`prompt` finding is the part that holds regardless of any runtime model, since it is visible in the template JSON itself.

The check that would settle it: `panel_set_widget(node_id=170, widget="enable_turbo_mode", value=true)` should dispatch instead of refusing, and a genuinely non-promoted widget on node 170 should still be refused rather than written outer.

### DIFF
--- a/src/__tests__/orchestrator/promoted-widget.test.ts
+++ b/src/__tests__/orchestrator/promoted-widget.test.ts
@@ -2970,3 +2970,197 @@ describe("panel_set_widget promoted write against a pre-#2314 panel build (panel
     expect(calls.filter((c) => c.cmd === "graph_set_widget")).toHaveLength(0);
   });
 });
+
+/**
+ * #2393 — the promoted-terminal witness veto was WHOLE-ARRAY.
+ *
+ * `promotedTerminalEvidenceError` refused a write when ANY entry in
+ * `promoted_terminals` repeated a widget name or carried an error, even when the
+ * alias being written had its own complete, error-free entry. The official
+ * Qwen-Image-Edit-2511 INT8 template hits that on every promoted write to
+ * subgraph node 170, and it is structural rather than transient:
+ *
+ *   - node 170's proxyWidgets carries ["151","prompt"] AND ["149","prompt"] —
+ *     two inner nodes whose inner widget is named `prompt` — while exactly one
+ *     host input (`prompt`/`positive_prompt`) claims that alias. The panel's
+ *     `promotedHostAliasRecords` binds the first relation onto the existing
+ *     unbound record and APPENDS a record for the second, and
+ *     `promotedTerminalWitnesses` publishes one entry per record with no
+ *     de-duplication. The array therefore always repeats `prompt`.
+ *   - node 170 also names `lora_name`, `seed` and `control_after_generate` in
+ *     proxyWidgets with no matching host input, which adds error entries.
+ *
+ * Neither depends on timing, so re-reading `graph_get_subgraph` returns the
+ * identical array — the write has to be decided on the requested alias's own
+ * evidence or it can never succeed.
+ *
+ * The fixtures below use `quality_prompt` as the requested alias because that is
+ * what this file's shared subgraph harness resolves; the SHAPE (one clean entry
+ * for the requested alias, damage on a different alias) is node 170's.
+ */
+describe("#2393 promoted-terminal witness is judged on the requested alias", () => {
+  const ownEntry = CURRENT_SAFE_PROMOTED_SUBGRAPH.promoted_terminals[0];
+  const withTerminals = (promoted_terminals: unknown[]) => ({
+    ...CURRENT_SAFE_PROMOTED_SUBGRAPH,
+    promoted_terminals,
+  });
+
+  it("writes when a DIFFERENT alias is duplicated (node 170's two `prompt` relations)", async () => {
+    const { isError, calls } = await setWidget(
+      { node_id: 78, widget: "quality_prompt", value: "masterpiece" },
+      {
+        firstWrite: "ok",
+        promotedTerminalWitnesses: true,
+        subgraph: withTerminals([
+          ownEntry,
+          { ...ownEntry, widget: "prompt" },
+          { ...ownEntry, widget: "prompt" },
+        ]),
+        detailById: SAFE_ANIMA_IDENTITY_BY_ID,
+      },
+    );
+
+    expect(isError).toBe(false);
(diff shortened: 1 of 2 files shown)


artokun:
Verified independently at `713d83e`, in my own worktree.

## The change is the right shape

Judging a write on the witness entry for **its own widget** rather than vetoing on the whole array is correct — a write to `X` should not be refused because `Y`'s entry carries an error. And the fallback is conservative in the right direction: only `own.length === 1` takes the narrow path; a miss or an ambiguous multi-match falls back to the whole-array veto.

    tsc --noEmit                exit 0
    promoted-widget.test.ts     126 passed | 0 failed   (117 before)

**Both safety conditions are pinned:**

    A  own.length === 1  ->  own.length >= 1        2 failed | 124 passed
       (judging on the first of several ambiguous entries)
    B  own[0].error ? refuse : null  ->  null        1 failed | 125 passed
       (ignoring the widget's own error)

Needle counts verified before each edit. So neither "exactly one match" nor "the entry's own error still refuses" is incidental.

## It does not touch #2401

I checked, because both changes live in `panel-tools.ts` and it would be easy to assume overlap. The `targetScope`/`ordinary` early return is untouched here — the only matches in this diff are comments about witness misses, which is a different sense of the word. **#2401 stays open on its own.**

## One thing worth copying

`713d83e` — *"drop the redundant requested-alias duplicate branch a mutation proved dead"* — is using mutation testing to find dead code rather than only to check coverage. That is the same technique that showed me #2398's allowlist was unpinned and that my own #2387 had a fail-open. Worth more of.

Gating now. Nothing blocking from me.

Issue #2401, opened 2026-08-27
promoted-write pre-dispatch: the ordinary-node fast path returns across an await without re-reading tab/connection identity


artokun:
## The ordinary-node fast path returns across an `await` without re-reading identity

On `origin/main` (`8abbbf6`, #2399), in the promoted-write pre-dispatch check:

    6826  const tabBefore = ctx.tabId;
    6831  identityBefore = ctx.panelConnectionIdentity?.();
    6853  const targetScope = await readPromotedTargetScope(ctx, nodeId);      // async
    6854  if (targetScope?.activeView === "root" && targetScope.node === "ordinary") return null;
    ...
    6928  if (ctx.tabId !== tabBefore || !isUsablePanelConnectionIdentity(identityAfter)) { ... }
    6931  if (!samePanelConnectionIdentity(identityBefore, identityAfter)) { ... }

`readPromotedTargetScope` is `async` (`:6578`). The function captures `tabBefore` and `identityBefore` **specifically** so it can re-read them after its awaits — and the `"ordinary"` exit at `:6854` returns 74 lines before any of those comparisons.

`return null` here does not mean "refuse". It means *"this is not a promoted write, proceed on the ordinary path"* — so nothing downstream re-establishes the binding either.

## Why it is a real hole rather than a style nit

**The asymmetry is the evidence.** Two sibling functions in the same file (`:7338`, `:7350`) follow the same shape — capture, await, compare `tabId`, compare identity. This exit is the only one in the family that skips it. That is an omission, not a decision.

**Failure shape:** probe confirms an ordinary root node on tab A; the panel rebinds or reconnects during that round trip; `graph_set_widget` then writes to the same node id on tab B. Node ids collide freely across graphs, so nothing downstream notices. The write succeeds and looks correct.

This is the exact class the surrounding fence exists to prevent, which is why every neighbouring path guards it.

## Provenance

Raised by the codex merge gate on #2399 as **P1**. I confirmed it in source, and #2399 merged anyway at 02:43:54Z. The merge did tighten *what* is accepted — `activeView === "root"` was added alongside `node === "ordinary"` — but that narrows the classification, not the window. The rebind race is unchanged.

I am filing at **P2** rather than the gate's P1 because it needs a rebind inside a single probe round-trip, which is narrow. I would not argue hard against P1: it is a silent wrong-target **write**, and I under-called exactly this kind of thing on #2382 earlier tonight by reasoning about likelihood instead of impact.

## Fix

[...]

PR #2402, opened 2026-08-27
fix(2394): definitive non-subgraph read must tolerate a parenthesised node type

artokun:
Follow-up to #2394, which merged as #2399 and is already closed. This lands the **root cause** that PR routed around.

## What #2399 fixed, and what it left

#2394 reported that `panel_set_widget` refused a documented `lora_1` creation on a **root-level** rgthree Power Lora Loader with:

> graph_get_subgraph could not determine whether the addressed node is a promoted container

#2399 fixed that by proving an ordinary root node via a `graph_query` scope probe and returning **before** `graph_get_subgraph` is reached. That is correct and stays. But it treats the symptom: the reason the write was misclassified in the first place is in `isDefinitiveNonPromotedSubgraphRead`.

## Root cause

The panel's one definitive non-promoted answer is its own message:

```
Node 82 (Power Lora Loader (rgthree)) is not a subgraph
```

The classifier matched the parenthesised segment — which is the node **type** — with `\([^)]*\)`. That body stops at the **first** `)`, and rgthree names every node `... (rgthree)`. So the message failed to match its own definitive form, the read was downgraded to *indeterminate*, and the ordinary write was refused. Measured, not inferred:

| message | matched before |
|---|---|
| `Error: Node 82 (KSampler) is not a subgraph` | ✅ true |
| `Error: Node 82 (Power Lora Loader (rgthree)) is not a subgraph` | ❌ **false** |
| `Error: Node 82 (KSampler (Efficient)) is not a subgraph` | ❌ **false** |

Driving the shipped `panel_set_widget` handler with the real rgthree type reproduces the reporter's error **verbatim**; the identical case with a paren-free type writes successfully. The node type's parentheses are the only discriminator.

## Paths still broken before this PR

The #2399 shortcut is root-only and needs a panel that can classify, so the classifier is still the sole gate for:

1. **An ordinary rgthree node reached while viewing a subgraph** — the shortcut requires `activeView === "root"`.
2. **A panel older than 0.15.101**, which emits no `is_subgraph`, so the probe reads indeterminate and correctly falls through.

Both refused every rgthree write. Both are pinned here.

## Where to look hardest

- **This loosens a guard that authorizes a write**, so it is the risky direction. `.*` is bounded rather than open: still anchored at `^Error:\s*Node`, still requires the literal ` is not a subgraph` tail, `.` never crosses a newline, and greedy backtracking settles on the last `)` before that tail. The strings it newly admits are the panel's own message with a nested-paren type. The `[canvas-root-divergence]` diagnosis cannot collide — it never begins with `Error: Node `.
- Two **controls** are included and were required to pass in *both* arms of the mutation: a genuinely indeterminate read (transport failure) and a shape-adjacent message that carries parentheses and the word "subgraph" but is not the definitive form. Both still refuse.
- One pre-existing #2394 test, `does not use the root-only lora_N shortcut from an active subgraph`, does **not** exercise the type its own fixture names: the fake bridge substitutes a generic `OrdinaryNode` for a nested-view read (`promoted-widget.test.ts`, the `inSubgraph` branch), so it dodges this bug. It is left as-is — the new subgraph-view test drives the real type through `nestedSubgraph`.

## Verification

- **Mutation pair** (the fix committed first, then reverted in place): both new pins **fail**, both controls **survive**, and all 118 pre-existing `promoted-widget` tests **survive** — so the pins measure this change and nothing else.
- Full orchestrator suite: **256 files / 4373 tests passed**.
- `tsc --noEmit` clean; `npm run lint` clean (anti-slop: none added); `git diff --check` clean.
- Not panel-visible: the change is orchestrator-side refusal classification only, no panel rendering path, so no Playwright run applies.

### DIFF
--- a/src/__tests__/orchestrator/promoted-widget.test.ts
+++ b/src/__tests__/orchestrator/promoted-widget.test.ts
@@ -3164,3 +3164,138 @@ describe("#2393 promoted-terminal witness is judged on the requested alias", () 
     expect(calls.filter((c) => c.cmd === "graph_set_widget")).toHaveLength(0);
   });
 });
+
+// #2394 (follow-up) — the definitive-read classifier matched the panel's own
+// `Node <id> (<type>) is not a subgraph` message with a `\([^)]*\)` body, which
+// stops at the FIRST `)`. rgthree names every node `… (rgthree)`, so the message
+// for the reported node never matched, the read was downgraded to indeterminate,
+// and the ordinary write was refused.
+//
+// PR #2399 added a root-scope shortcut that returns before graph_get_subgraph is
+// ever called, which hides this for a root node on a panel new enough to classify
+// (>= 0.15.101). It does NOT cover the two paths below, where the classifier is
+// still the only thing standing between an ordinary node and a false refusal.
+describe("definitive non-promoted read tolerates a parenthesised node type (#2394)", () => {
+  const RGTHREE_VALUE =
+    '{"on":false,"lora":"Detailer-KREA2.safetensors","strength":0.3,"strengthTwo":null}';
+
+  it("writes a root-level rgthree lora_N row when the scope probe cannot classify", async () => {
+    // A panel older than 0.15.101 emits no `is_subgraph`, so the #2399 shortcut
+    // reads indeterminate and correctly falls through to graph_get_subgraph.
+    // From there the classifier is the only remaining gate.
+    const { isError, text, mutations, calls } = await setWidget(
+      { node_id: 82, widget: "lora_1", value: RGTHREE_VALUE },
+      {
+        firstWrite: "ok",
+        subgraph: new Error("Node 82 (Power Lora Loader (rgthree)) is not a subgraph"),
+      },
+    );
+
+    expect(text).not.toMatch(/could not determine whether the addressed node is a promoted container/);
+    expect(isError).toBe(false);
+    expect(mutations).toBe(1);
+    expect(calls.filter((c) => c.cmd === "graph_set_widget")).toEqual([
+      expect.objectContaining({ node_id: 82, widget: "lora_1", value: RGTHREE_VALUE }),
+    ]);
+  });
+
+  it("writes an ordinary rgthree node from inside a subgraph", async () => {
+    // The #2399 shortcut is deliberately root-only, so an ordinary rgthree node
+    // reached while viewing a subgraph still goes through graph_get_subgraph.
+    const { isError, text, mutations, calls } = await setWidget(
+      { node_id: 82, widget: "lora_1", value: RGTHREE_VALUE },
+      {
+        ownerNodeId: 78,
+        startInSubgraph: true,
+        firstWrite: "ok",
+        nestedSubgraph: new Error("Node 82 (Power Lora Loader (rgthree)) is not a subgraph"),
+        promotedDetail: {
+          text:
+            '1 match(es) of 1 in scope (viewing: 1 nodes)\n' +
+            '{"id":82,"type":"Power Lora Loader (rgthree)","is_subgraph":false}',
+        },
+      },
+    );
+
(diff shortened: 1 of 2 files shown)


artokun:
**This is the real root cause, and it explains why #2394 was specific to rgthree.** Verified by executing both regexes rather than reading them:

    message                                                     OLD    NEW
    Error: Node 82 (Power Lora Loader (rgthree)) is not a...    false  true
    Error: Node 82 (CheckpointLoaderSimple) is not a subg...    true   true
    Error: Node 82 is not a subgraph                            true   true

`[^)]*` stops at the **inner** `)`. rgthree's node type is `Power Lora Loader (rgthree)` — a nested parenthesis — so the definitive non-subgraph read was not recognised, the write fell through to the promoted-container path, and the classifier refused a node that was never promoted. Every node type without a nested paren behaved correctly, which is exactly the "only this one pack" shape the report described.

One character class, and the two simple cases are unchanged. That is a good root-cause fix.

    tsc --noEmit               exit 0
    promoted-widget.test.ts    122 passed | 0 failed

## The consequence worth deciding deliberately

#2399 fixed #2394 by adding a `targetScope` "ordinary root node" fast path **in front of** this read — routing around the parser rather than fixing it. With the parser correct, that fast path is no longer doing the work it was added for.

That matters beyond tidiness: the fast path is what **#2401** is filed against. It returns across an `await` without re-reading tab/connection identity, so it is the one exit in its family that can let a write land on the wrong target after a rebind. If it is now redundant, deleting it closes #2401 outright rather than needing the identity re-check hoisted into it.

I have not asserted it *is* redundant — proving that needs the panel side to confirm every ordinary-root case now produces a definitive read this regex matches. But it is worth answering explicitly while this is fresh, because the alternative is carrying a TOCTOU-bearing shortcut that no longer has a reason to exist.

This PR does not touch `targetScope` (0 lines), so #2401 stays open either way.

Gating now. Nothing blocking.

artokun:
Re-verified on the current head `0b4683d` — the branch was rebased onto main (picking up #2400 and #2398) after my earlier review, so that gate verdict was stale and I did not act on it.

    diff vs main    +96 test lines, +21 in panel-tools.ts
    tsc --noEmit    exit 0
    promoted-widget 130 passed | 0 failed

**Mutation — the character class is load-bearing.** Reverting `.*` to `[^)]*`:

    Tests  2 failed | 128 passed

Needle count verified as exactly 1 before the edit. (My first attempt at this mutation silently failed to apply and reported a clean 130 — worth saying, because a mutation that never applied is indistinguishable from one that survived.)

`targetScope` is still untouched (0 lines), so the redundancy question from my previous comment stands open: with the parser correct, does #2399's "ordinary root node" fast path still earn its place? If not, deleting it closes **#2401** rather than needing the identity re-check hoisted into it.

Re-gating now on the real head.

artokun:
Holding off re-gating — the head has moved twice while I was gating (`9b57d39` → `0b4683d` → `ee31ac2`, the last an amend two minutes ago), so both verdicts I got describe commits that are no longer the tip. Ping me or just let CI settle and I will gate the final sha rather than produce a third stale SHIP.

My verification of the substance still stands, and the amends have not touched it: the character class is the fix, and reverting `.*` to `[^)]*` fails 2 tests with 128 surviving. `targetScope` remains untouched (0 lines) across every revision, so **#2401** is unaffected either way.

artokun:
Independent gate: NO-SHIP at exact head 90ec94e0205a59dca8c613899fc6eae5f27e836c. P1 blocker: the new nested-parenthesis classifier returns null and can treat a contradictory/stale response as ordinary, bypassing an explicit is_subgraph:true fence and reaching direct graph_set_widget. Preserve explicit subgraph/contradictory-scope refusal or conservative resolution, and add a deep/malformed conflict regression. Hosted CI is green, but this safety defect blocks merge. No init.py/apps_routes.py network changes.

PR #2403, opened 2026-08-27
chore: release v0.52.138

### LINKED ISSUE #2395: Retry transient panel read transport failure after entering subgraph

artokun:
## Release v0.52.138\n\nIncludes merged P2 fixes #2395 and #2393/#2394.\n\n- bumps package.json and package-lock.json to 0.52.138\n- regenerates CHANGELOG.md for #2398 and #2399\n- generated from origin/main at aa2847d8\n\nNo init.py or apps_routes.py changes; no network operations introduced there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+## [0.52.138] - 2026-08-26
+
+### MCP
+
+#### Fixed
+- retry active panel reads once (#2398)
+- reconcile promoted-widget writes (#2399)
+
+
 ## [0.52.137] - 2026-08-26
 
 ### MCP
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "comfyui-mcp",
-  "version": "0.52.137",
+  "version": "0.52.138",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "comfyui-mcp",
-      "version": "0.52.137",
+      "version": "0.52.138",
       "license": "MIT",
       "dependencies": {
         "@comfyorg/sdk": "^0.1.7",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "comfyui-mcp",
-  "version": "0.52.137",
+  "version": "0.52.138",
   "mcpName": "io.github.artokun/comfyui-mcp",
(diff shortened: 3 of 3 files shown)


artokun:
**This cut is missing `#2400`.** It merged as `b910904` after the branch was taken, so the section lists two of the three unreleased commits:

    unreleased on main:
      b910904  fix(2393): a promoted write is judged on its OWN witness entry (#2400)   <- not cited
      aa2847d  fix(codex): retry active panel reads once (#2398)                        <- cited
      8abbbf6  fix(panel): reconcile promoted-widget writes (#2399)                     <- cited

**The code will ship regardless** — this branch merges into `main`, which already carries `b910904`, and the tag is cut from the merge result. So this is a notes gap, not a stranded fix. That is worth stating explicitly because the same shape produced two opposite mistakes tonight: 0.52.133 shipped #2378 without listing it, and 0.52.134 was then cut *to* carry a fix that had already shipped.

One line closes it:

```
#### Fixed
- retry active panel reads once (#2398)
- reconcile promoted-widget writes (#2399)
+ - a promoted write is judged on its own witness entry, not the whole array (#2400)
```

I have not pushed it — I would be editing a release branch someone else is holding, and last time two of us alternately pushed a shared branch it cost more than it saved.

Worth knowing the operand, since it caught me out earlier: for an **open** release PR the ancestry question is moot (everything on main ships), so the branch check gives false alarms. The changelog is the only thing that can actually gap here, and it gaps silently.

artokun:
Still missing #2400 at `3be7ff44`.

There is a live worktree on this branch (`wt-release-052138-cc`), so I am holding to what I said above and not pushing to it — an agent owns it, and two of us editing one release branch has already cost more than it saved tonight.

**Fallback, so this does not get lost either way:** if this merges with the gap, I will correct the 0.52.138 section on `main` in a small docs-only PR, exactly as #2384 did for the 0.52.133/#2378 gap. The code ships correctly regardless; it is only the notes that would be wrong.

One line if you get there first:

    - a promoted write is judged on its own witness entry, not the whole array (#2400)

Worth noting this is the second cut tonight to miss a fix that merged between the branch being taken and the PR merging — 0.52.133 missed #2378 the same way. The window is small but it is not rare, and the changelog is the only thing that gaps: ancestry is fine because the branch merges into main.

artokun:
Final release gate: SHIP at exact head 3be7ff4455c2fb3b3fc93483da9e0cf1fd2929ab. Independent review confirms the three-file version/changelog diff, parity, unique/reachable v0.52.138 section, and #2398/#2399 ancestry. The rerun is fully green on Ubuntu, Windows, macOS, and pack/install. No init.py/apps_routes.py network changes. Proceeding to merge and publish.

PR #2404, opened 2026-08-27
docs(1906): panel_show_media says staging needs a LOCAL ComfyUI

artokun:
Addresses the reportable half of #1906.

## What the description promised

> A path item OVER the 20 MB inline cap that is not under any directory ComfyUI serves **can still be shown by passing `stage:true`** on that item — the orchestrator COPIES it into `<output>/_panel_staged` …

Unconditional. But staging is a **filesystem copy into ComfyUI's own output directory**, so it cannot run when the connected ComfyUI is on another host. Nothing in the description said so, so an agent reads it, tries `stage:true` on a remote session, and gets a refusal the description gave it no reason to expect.

The refusal text is already good — it names `stage:true`, says staging could not be done, and gives the reason. The promise above it was the wrong half.

## The change

One sentence, stating the constraint where the decision is made:

> Staging needs a LOCAL ComfyUI: it is a filesystem copy into ComfyUI's own output directory, so it cannot run against a remote target (`--comfyui-url` on another host) and the call reports that instead of staging.

**No behaviour change.** The remote path still reports rather than staging.

## What I did NOT do, and why

The reporter's underlying ask is that staging **work** remotely, by streaming the file through the connected ComfyUI's upload API. That is a new capability, not a correction, so it stays parked under the freeze — noted on the issue rather than dropped.

Worth separating the two because they have different answers: the doc mismatch is live for every remote user reading the tool description today; the feature is a want.

**Verified:** `tsc --noEmit` exit 0, `npm run lint` exit 0, `check:vocabulary` clean, `panel-show-media-oversized.test.ts` 30 passed / 0 failed.


artokun:
Independent gate: SHIP at exact head 0f5d4a17bb3f6a59ba59c7d3d0ddbb836af6866e. Review confirms the one-line description accurately documents local-only staging, preserves the remote upload workaround, and introduces no executable/security regression. Hosted CI is green; no init.py/apps_routes.py network changes. Proceeding to merge.

PR #2405, opened 2026-08-27
fix: fence promoted ordinary fast path after scope probe

### LINKED ISSUE #2401: promoted-write pre-dispatch: the ordinary-node fast path returns across an await without re-reading tab/connection identity

artokun:
## Summary
- re-read tab and panel connection identity after the async promoted scope probe
- fail closed before the ordinary-node fast-path return when identity is unavailable or changed
- add temporal tab/connection rebind regressions plus stable root, subgraph, and unknown-scope coverage

Fixes #2401

## Validation
- npm.cmd run test:vitest -- src/__tests__/orchestrator/promoted-widget.test.ts
- npm.cmd run build
- npm.cmd run lint
- git diff --check

### DIFF
--- a/src/__tests__/orchestrator/promoted-widget.test.ts
+++ b/src/__tests__/orchestrator/promoted-widget.test.ts
@@ -75,6 +75,10 @@ function bridge(opts: {
   /** #2393: successive graph_get_subgraph replies, used to model a witness
    * that is incomplete immediately after an official template load. */
   subgraphSequence?: Array<Record<string, unknown> | Error>;
+  /** #2401: change the routing or panel connection identity after the async
+   * ordinary-node scope probe has produced its result, before the caller can
+   * take the fast path. */
+  scopeProbeIdentityChange?: "tab" | "connection";
   nestedSubgraph?: Record<string, unknown> | Error;
   enterFails?: boolean;
   exitFails?: boolean;
@@ -171,6 +175,7 @@ function bridge(opts: {
     reason: "no current panel graph-scope witness has been observed",
   };
   const beforeWrite = { mutate: undefined as (() => void) | undefined };
+  const afterScopeProbe = { mutate: undefined as (() => void) | undefined };
   const legacyBuild = opts.legacyPanelBuild !== undefined;
   const currentViewing = () => ({
     scope: inSubgraph ? "subgraph" : "root",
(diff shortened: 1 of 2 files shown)


artokun:
Final gate: SHIP at exact head 2e5795c802e2f0063f3769e24a8fc644c68e4de4. Independent review confirms post-await tab/connection identity fencing, conservative mismatch behavior, and temporal regression coverage. Ubuntu, Windows, macOS, and pack/install CI are green. No init.py/apps_routes.py network changes. Proceeding to merge.

PR #2408, opened 2026-08-27
fix(#1639): verify bare workflow aliases before success

artokun:
Refs #1639\n\n## Summary\n- Re-read the panel active workflow after a successful bare-filename workflow_open resolves to a saved path.\n- Fail closed when the resolved active identity is different, unconfirmed, or unreadable; leave the existing workflow fence unchanged.\n- Preserve #716: an alias may be observed, but its reply-resolved path never promotes a new workflow-fence UUID. Exact-path opens and graph read/render fencing are unchanged.\n- Add production-handler tests for stale active, confirmed resolved target, and unconfirmed active cases.\n\n## Validation\n- Full Vitest: 681 files, 12,723 passed, 6 skipped.\n- Focused #1639/render tests: green.\n- Build, lint, and diff checks: green.\n- Rebased on origin/main ffd5d55, including #2402/#2404/#2405; this diff does not touch the promoted-widget/subgraph path.

### DIFF
--- a/src/__tests__/orchestrator/open-reconnect-false-mismatch.test.ts
+++ b/src/__tests__/orchestrator/open-reconnect-false-mismatch.test.ts
@@ -101,10 +101,19 @@ function emptyGraph() {
   return { nodes: [] as unknown[], links: [] as unknown[] };
 }
 
-function listReply(opts: { confirmed: boolean; uuid: string; path?: string }) {
+function listReply(opts: {
+  confirmed: boolean;
+  uuid: string;
+  path?: string;
+  omitActiveConfirmed?: boolean;
+}) {
   const path = opts.path ?? PATH;
   const active = { path, routing_key: `wf:${path}`, workflow_uuid: opts.uuid };
-  return { active, workflows: [{ ...active, active: true }], active_confirmed: opts.confirmed };
+  return {
+    active,
+    workflows: [{ ...active, active: true }],
+    ...(opts.omitActiveConfirmed ? {} : { active_confirmed: opts.confirmed }),
+  };
 }
 
 type GraphQuery = "answers" | "mismatch";
@@ -112,12 +121,15 @@ type GraphQuery = "answers" | "mismatch";
 function bridge(opts: {
   stamp: string;
   graphQuery: GraphQuery;
+  openResult?: "error" | "success";
   /** First N workflow_list replies are unconfirmed; then confirmed. */
   unconfirmedLists?: number;
   /** First N graph_serialize replies are the empty settling snapshot. */
   emptySerializes?: number;
   listPath?: string;
   listUuid?: string;
+  omitOpenedPath?: boolean;
+  omitActiveConfirmed?: boolean;
 }) {
   const calls: string[] = [];
   let stamp: string | undefined = opts.stamp;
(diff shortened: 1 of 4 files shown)


artokun:
Verified independently at `e24cddb`, in my own worktree. **This lands on exactly the branch the triage narrowed to.**

When I looked at #1639 I flagged two details from the reporter's account: that they used a **bare filename** rather than a path or id, and that `applied:true` came back with concrete bound fields — the reply-versus-observed-effect shape. I also established the fix you would expect was already present in their build (`settleOwnedOpenedWorkflowActive`, `8d45a904`, in v0.15.99), which is what made this a new defect rather than a stale report.

This PR branches on precisely that:

    const requestedIsBareAlias = !!requestedSavedPath && !requestedSavedPath.includes("/");
    if (requestedIsBareAlias) {
      if (!openedPath || !canonicalRequestedSavedIdentity(openedPath)) {
        return { drifted: true, unverified: true, activeLabel: "the panel did not return a resolved path for this filename" };

Two things I would call out as right. It **fails closed on the unread case** rather than treating an absent resolved path as agreement — `unverified: true` is a distinct state from `drifted`, so the caller can say which it hit. And it re-reads rather than trusting the open's own reply, which is the half the original defect turned on.

    tsc --noEmit                             exit 0
    open-reconnect-false-mismatch.test.ts    9 passed | 0 failed

Not gating yet — this is the mcp half and panel#1907 is the other. They are the two sides of the same report (the read path and the write path disagreeing about one node), so worth landing together or at least knowing which order.

One question rather than an objection: does the bare-alias re-read cost a `workflow_list` on every bare-filename open, or only when the first check is inconclusive? The reporter's session was a long generation run, so a per-open round trip would be felt.

artokun:
Gate returned two P1s. **I checked both and they hold — and they interact in a way that matters for the order you fix them in.**

Note first that both are in the **#971** work, not the #1639 work. The bare-alias verification I verified earlier is unaffected: it still pins (disabling `requestedIsBareAlias` fails 1 test), and 423 pass across the wider files.

## Finding 2 is decisive — the #971 recovery is unreachable

Store and compare use different canonicalisers:

    :21357  store    ctx.lastExplicitCurrentRebind = { tabId, savedIdentity }
    :21355           savedIdentity = canonicalSavedRecordIdentity(fenceRebind.active)   -> "wf:workflows/demo.json"

    :11239  compare  legacyRebind.savedIdentity === canonicalBareSavedIdentity(requestedPath)
    :12120           return canonicalPath && !canonicalPath.includes("/") ? `wf:${canonicalPath}` : null;

So for `workflow_open("demo.json")` the comparison is `"wf:workflows/demo.json" === "wf:demo.json"` — false. And for a path containing `/`, `canonicalBareSavedIdentity` returns `null`, which is also never equal.

**Either way the branch cannot fire for a normal saved path.** The recovery is dead code as written.

## Finding 1 holds, and finding 2 is currently masking it

The proof is consumed where it is read (`:11618-11619`) and cleared at `:21075`, so an *open* cannot inherit a stale one. Nothing clears it when the active workflow changes **by any other route** — the proof is keyed to `tabId` and a saved identity, neither of which moves when the canvas does.

So the gate's sequence is right: rebind to `a.json`, change active to `b.json` by a path that is not a `workflow_open`, then a legacy `{ok:true, routedTo:"tab-b"}` reply for `workflow_open("a.json")` is accepted on a proof that no longer describes the world.

## The interaction, which is the thing I would act on

**Fixing #2 alone turns #1 from unreachable into live.** Today the comparison never matches, so the stale-proof path is never taken. Make the canonicalisers agree — the obvious fix, and the one the finding invites — and every scenario in finding 1 becomes reachable on the next release.

So they are one change, not two: align the identities **and** invalidate the proof on an active-workflow change (or narrow it to "the open immediately following the rebind" in a way that survives a canvas move, which is what the comment at `:21073` already claims it is).

## Suggestion on scope

This PR now carries #1639 and #971. The #1639 half is verified and clean; the #971 half has both of these. Splitting them would let the bare-alias fix land now — it addresses a live P2 with a reporter waiting — rather than waiting on a recovery path that is currently inert anyway.

Happy to be wrong about that if the two are more entangled in the code than they look from the diff.

Issue #2409, opened 2026-08-27
promoted-write pre-dispatch: the definitive-non-subgraph exit returns across an await without re-reading tab/connection identity (#2401 sibling)


artokun:
## The definitive-non-subgraph exit returns across an await without re-reading tab/connection identity

Follow-up to #2401, which fixed the sibling exit five lines above this one. On `origin/main` (`ffd5d55`, #2405), in `preparePromotedWidgetWrite` (`src/orchestrator/panel-tools.ts`):

    6914  const tabBefore = ctx.tabId;
    6915  const hasIdentityApi = typeof ctx.panelConnectionIdentity === "function";
    6918      identityBefore = ctx.panelConnectionIdentity?.();

    6930  const targetScope = await readPromotedTargetScope(ctx, nodeId);
    6931  if (targetScope?.activeView === "root" && targetScope.node === "ordinary") {
                ... #2405 re-reads tabId + identity here, refuses on mismatch ...
    6953    return null;
          }

    6956  const sub = await ctx.call({ cmd: "graph_get_subgraph", node_id: nodeId });   // async
    6958  if (isDefinitiveNonPromotedSubgraphRead(sub)) return null;                   // <-- no re-read

    ...
    7015  if (!hasIdentityApi) { ... }          // the tail fence, 57 lines later
    7031  if (!samePanelConnectionIdentity(identityBefore, identityAfter)) { ... }

`:6958` returns `null` — "this is not a promoted write, proceed on the ordinary path" — after the `graph_get_subgraph` round trip, without re-reading either `ctx.tabId` or `panelConnectionIdentity()`. It is the same shape #2401 described, at the next exit down.

#2401's body asserted the fast path "is the only one in the family that skips it." That was wrong; this is the second one.

## Measured, not inferred

Run against a clean `ffd5d55` tree (worktree outside the repo, own `npm ci`), driving the rebind *inside* the `graph_get_subgraph` handler so it lands within the await window. Case D is the exit #2405 fixed, included as the reference arm:

| case | rebind | exit | result |
|---|---|---|---|
| A | none (control) | `:6958` | `isError=false writes=1 dispatched=YES` |
| B | connection identity, gen 1 to gen 2 | `:6958` | `isError=false writes=1 dispatched=YES` |
| C | `ctx.tabId` changed | `:6958` | `isError=false writes=1 dispatched=YES` |
[...]

PR #2340, opened 2026-08-26
fix(rate-limit): redact all-alphabetic identifiers

artokun:
Replacement for closed PR #2315 after base #2294 merged. This rebased child closes all-alphabetic identifier, structured/JSON label, prefix/underscore/wrapper, and prose-boundary sanitizer gaps for issue #2313. Focused sanitizer, Panel rate-limit, and enqueue redaction suites: 68 passed; lint, build, and diff checks pass.

PR #2412, opened 2026-08-27
fix(2407): verify the release section, not just generate it

### LINKED ISSUE #2407: Port the panel's changelog release guard to mcp — three cuts tonight shipped a fix they did not list
## The two repos have diverged on changelog correctness

The panel now catches malformed and misattributed release sections. mcp does not, and mcp is where the last two defects shipped.

**Panel, since comfyui-mcp-panel#1894 / #1897 / #1902:**

- `scripts/check-changelog.mjs` — no repeated heading in a section, no version cited twice, and every entry's PR reachable from that version's tag (`git merge-base --is-ancestor`).
- A CI and publish version gate that additionally compares the rendered `web/changelog.json`, failing closed when it is unreadable or stale.
- `changelog-integrity.test.mjs` comparing source and artefact.

`v0.15.110` was the first cut to pass all of it cleanly, verified by running the shipped guard against its own tag.

**mcp has `gen-changelog.mjs` and `changelog-section.mjs` — generators, no verifier.**

## What that cost tonight

| release | defect |
|---|---|
| **0.52.133** | shipped #2378 and did not list it (fixed in #2384) |
| **0.52.134** | cut *to* carry a fix that had already shipped — no code changes at all |
| **0.52.138** | shipped #2400 and did not list it (fix in #2406) |

Three in one evening, all the same window: a PR merges between the release branch being cut and the release PR merging. Ancestry resolves itself — the branch merges into `main`, so the tag gets the code — which is exactly why nobody notices. **The changelog is the only thing that gaps, and it gaps silently.**

0.52.134 is the instructive one: reading a notes gap as "the fix has not shipped yet" produced a whole no-op version, which is the opposite error from the same root cause.

## What to port

The rule that would have caught all three is one of the panel's three, and it needs no rendered-artefact equivalent (mcp has no `changelog.json`):

> every PR cited in the `## [X]` section must be reachable from `vX`, **and** every PR merged since `v<previous>` must be cited in it.

The panel implements the first half. The second half is the one that catches these — a *missing* entry, rather than a wrong one. Both are a `git log v<prev>..v<X>` and a set comparison.

Worth reusing the panel's shape rather than reinventing: `checkChangelog()` is pure with `isAncestor` injected, which makes it testable without a repo fixture, and `release-version-gate.test.mjs` executes the real workflow file rather than a copy — the detail that makes a workflow gate trustworthy.

## Not implementing it here

Wiring a gate into the release path touches `.github/`, which is supply-chain gated for autopilot — the merge gate exits 2 and `AUTOPILOT_HUMAN_ACK=1` is not mine to set. The panel's two equivalents (#1894, #1902) both needed a human for the same reason.

artokun:
Closes #2407.

## What shipped, and why it is not just a port

mcp had a changelog **generator** and no verifier. Three cuts in one evening went wrong the same way:

| release | defect |
|---|---|
| **0.52.133** | shipped #2378 and did not list it (hand-corrected by #2384) |
| **0.52.134** | cut *to* carry a fix that had already shipped — a whole no-op version |
| **0.52.138** | shipped #2400 and did not list it (hand-corrected by #2406) |

One race produces all three: a PR merges between the release branch being cut and the release PR merging. Ancestry then resolves itself — the branch merges into `main`, so the tag gets the code — which is exactly why nobody notices. Only the changelog gaps, and it gaps silently.

Two directions are now checked:

- **A. everything CITED must be REACHABLE** — and must have shipped *in this release*, not merely be reachable from it. Ported from comfyui-mcp-panel#1894, then tightened.
- **B. everything REACHABLE in `v<prev>..vX` must be CITED in `[X]`.** The panel does **not** implement this half. A *missing* entry is a different failure from a wrong one, and it is the one that shipped three times.

Plus the structural checks a hand-edited section needs.

## Wiring — no `.github/` or `package.json` change

The issue expected this to be unmergeable by autopilot because a release gate means touching `.github/`. It does not. `npm test` runs `scripts/run-checks.mjs`, and **both `ci.yml` and `release.yml` already run `npm test`** — so one registration in that plain script gates a pull request *and* the publish. `run-checks.mjs` already invokes `node scripts/<x>.mjs` directly, so no npm script was needed either.

A wiring test asserts the whole chain **by path**: `package.json` `scripts.test` to `run-checks.mjs` to `check-changelog.mjs`, and `npm test` present in both workflows. `fetch-depth: 0` is pinned there too — drop it and every run becomes a structural check that still exits 0.

## What the guard finds on real history

Replayed against 16 real tags. `main` stays green; every violation was checked by hand against the file:

- **#2294, #2307, #2326, #2340** — four PRs that landed as true merge commits and are cited **nowhere in `CHANGELOG.md`**. 0.52.124 shipped a single entry while carrying four PRs.
- **[0.52.133] credits itself with #2196**, whose commit is not in `v0.52.132..v0.52.133`; its actual work was #2376, separately reported as missing.
- **[0.52.134]** — the no-op release — is caught by the reachability half alone.
- Plus the already-known gaps in 0.52.126 / 128 / 132 / 137 / 138.

## Review log — 7 rounds, 14 findings, 13 real and fixed

Every round returned NO-SHIP with at least one finding I could reproduce, so I kept going well past one round. Nothing below was taken on faith; each was reproduced against real history or a scratch repo *before* being fixed.

1. **Shallow clones invented failures.** `git clone --depth 1` of this branch reported #2398 / #2399 / #2400 as unreachable against a *correct* release, while the message claimed only coverage had been skipped. Both history-dependent halves now decline together.
2. **Merge commits were invisible** (`--no-merges` plus a parenthesised-only regex). **An umbrella issue vouched for every PR under it** — one entry citing #2393 covered both #2400 and #2409, hiding the second. **A git failure reported a pass.**
3. **A numeric scope was read as a shipped PR** — my own regression from round 2, which flagged the real 0.52.124. **The release predicate swallowed `fix: release v1.2.3 (#778)`.** **A `describe` failure always meant "no previous tag".**
4. **Reachable is not shipped-here** (the 0.52.133 case above). **A release with no section at all was audited as its predecessor** and passed.
5. **A valid two-component section was REJECTED** — `[0.50.37]` really does carry `#### Fixed` under both `### RunPod image` and `### MCP`, and the guard failed it. A false positive that blocks a correct release is the worst direction for this mechanism. **`describe --tags` could select a stray non-release tag** (`backup/570-prerebase`).
6. **A failed range read was misfiled as a missing tag** — the third sibling of the git-failure family, after the history read and `describe`.
7. **Issue→PR links chained transitively.** The panel's union-find makes the relation transitive, and transitivity is false here: this project reuses a previous PR number as the next commit's scope — **62 numeric scopes in recent history are also PR numbers** — so `fix(100): (#200)` and `fix(200): (#300)` collapse into one class and a section citing only `#100` silently vouches for `#300`. Replaced with a direct one-hop map.

**Declined, with evidence:** requiring every cited ref to be a real PR would fail **[0.52.124], whose only entry cites `(#2319)`** — a number that exists solely as a conventional scope and a branch name, never as a PR. That entry is correct; the change would have regressed it.

## And two defects I found in my own work

1. **The reachability half was inert for a spelling the live changelog uses.** 0.52.136 writes `… (#2382, #2387)`; the panel's regex returns nothing for that, so half A skipped the entry in silence while half B still passed the release. Replaying all tags after the fix gives identical verdicts — it widens what is *checked* without changing what is *reported*.
2. **A wiring assertion was matching a comment.** Both workflows *explain* `fetch-depth: 0` in prose, and the substring assertion matched the explanation — setting the real key to `1` left the test green. Now anchored to a whole line and re-mutated against real setting lines only.

## Verification

- **30 mutations, each confirmed to have actually applied before its result was believed; all 30 caught by a named test.** One survivor was found and fixed: restoring `--no-merges` left every merge test green, because they fed `rangeCommits` straight to the pure audit — the helper understood merges while the query feeding it went blind. There is now an end-to-end test driving a real `--no-ff` merge through the CLI.
- **CI proves the guard RAN rather than skipped** — the logs show `lists every PR since …` on all three runners, which only prints when the range actually resolved.
- Full suite in a clean tree outside `.claude/` (vitest excludes `**/.claude/**`, so a run *inside* the worktree collects zero files and reads as a pass): **682 files, 12783 passed, 6 skipped**. The lone failure under concurrent load was `late-mutation-e2e.test.ts` — not in this diff, passes 5/5 alone, a known load flake.

## Stated rather than papered over

- **Two defensive branches have no deterministic test**: "git cannot read tags" and "the range read failed". Every corruption I could construct either fails earlier at HEAD resolution (caught, exit 1) or degrades to a loud skip. They are kept as defence and are **not** claimed as covered.
- **A known false-positive class**: an entry citing only an ISSUE number is flagged when no commit subject links that issue to its PR. Measured rate: 2 in 16 releases. The failure message names the escape hatch.
- **This gate can turn `main` red for a docs-only reason.** If a release ships a gap, CI is red until `CHANGELOG.md` is corrected. Deliberate — the same posture as `vocab:export --check` and the docs-gen gate — and it also stops the *next* release publishing on top of a lie.
- **Provenance.** The review gate edits, commits and pushes to the branch it reviews — it authored `edba319` and `534967e`, and its edits were swept into another commit by a blanket `git add -A`. All of that was removed; the branch was rebuilt from my last clean commit in a fresh worktree, and every fix here was written independently from the findings. `git log` on this branch is entirely mine, and no auto-generated code ships.

## Deliberately not done

- No `.github/workflows/**` and no `package.json` change — both are supply-chain gated for autopilot.
- No rendered-artefact check: mcp has no `web/changelog.json`, so the panel's third guard has no equivalent here.
- Historical sections are not retro-fixed. The CLI audits the release being cut; `--ref vX` audits any past tag on demand.


artokun:
I filed #2407. Verified at `d2be64c` in my own worktree, **against the real defect rather than its own fixtures**.

## It catches the cut that motivated it

Removed the `#2400` entry from the 0.52.138 section — recreating exactly what shipped:

    changelog: ERROR — [0.52.138] does not mention PR #2400, which shipped in this release
    (reachable from v0.52.138, not from v0.52.137): "fix(2393): a promoted write is judged on
    its OWN witness entry, not the whole array (#2400)". Add the entry, or add #2400 to the
    entry that already covers it.

    exit=1

That message is better than the rule. It names the PR, shows the reachability reasoning that justifies the claim, quotes the commit subject so you can find it, and offers **two** remedies — including the case where an existing entry already covers the work and just needs the ref. Someone hitting this in CI at 3am does not have to go read the checker.

## It implements the half I asked for

The panel's version checks that every **cited** PR is reachable from the tag. That direction catches a wrong entry, and would have missed all three of the cuts this issue was filed over — those were **missing** entries. This does both:

    changelog: [0.52.138] lists every PR since v0.52.137, and every entry it names
               is reachable from v0.52.138

## Verified

    tsc --noEmit                        exit 0
    changelog-release-guard.test.ts     19 passed | 0 failed
    check-changelog.mjs on main         exit 0

And it is **wired into a path that runs** — `scripts/run-checks.mjs:29`, invoked by `npm test`. A guard nobody executes is the failure mode one step past not having one.

## Note on mergeability

Unlike the panel's equivalents (comfyui-mcp-panel#1894, #1902), this touches only `scripts/` and `src/__tests__/` — no `.github/`, no `package.json` — so it is not supply-chain gated and autopilot can land it. Worth saying because I told you on #2407 that a human would be needed; that was true of the panel's shape, not of this one. Better solution than the one I proposed.

Gating now.

artokun:
Gate returned **NO-SHIP [P1]**. **The code half is right and I reproduced it; the impact half is wrong.** Both matter, so here is each with evidence.

## The defect is real — reproduced on a genuinely shallow clone

`git clone --depth 1` of this branch, then `node scripts/check-changelog.mjs`:

    changelog: shallow clone — coverage (every shipped PR is listed) was NOT checked.
    changelog: ERROR — [0.52.138] entry at line 13 names PR #2398, but no commit reachable from HEAD carries that reference.
    changelog: ERROR — [0.52.138] entry at line 14 names PR #2399, but no commit reachable from HEAD carries that reference.
    changelog: ERROR — [0.52.138] entry at line 15 names PR #2400, but no commit reachable from HEAD carries that reference.
    exit=1

The shallow branch at `:380` skips **coverage** and says so, but **rule A (cited ⇒ reachable) still runs against the depth-1 pool**, so every valid entry is reported as unreachable. The comment two lines above states the principle exactly — *"a shallow clone cannot answer this. It can only answer it WRONG"* — and then only half of it is applied.

## The impact claim does not hold

> `npm test` fails and blocks publication

Not in this repo. Every workflow that runs tests uses a full clone:

    ci.yml         fetch-depth: 0    runs tests
    release.yml    fetch-depth: 0    runs tests
    packs.yml           <default 1>  no tests
    mcp-registry.yml    <default 1>  no tests
    build-runpod-image  <default 1>  no tests

The citation is also to lines that do not exist — `release.yml` is **173 lines**, and the finding points at 550–567. The gate reported `CreateProcessAsUserW failed: 5`, so it could not execute and reasoned statically; that is where the line numbers and the release-blocking conclusion came from.

## So: real bug, wrong severity

Nobody's release is blocked. What *is* broken is **anyone running `npm test` on a shallow clone** — a plausible dev or third-party CI setup — who gets a wall of false errors naming their changelog as wrong when it is fine. That is the failure mode the shallow branch was written to prevent.

I would call it **P3**: no release impact, no correctness impact on the artefact, but a confidently-wrong failure that would send someone editing a correct changelog.

## The fix looks like one line

Gate rule A on `shallow` as well, or inject `isAncestor = () => true` when shallow so the pure audit cannot conclude anything from a pool it should not trust. The second is nicer — it keeps the skip decision at the one place that knows about the clone, and `checkChangelog()` stays honest about what it was given.

Worth also making the skip message name **both** halves, since it currently claims only coverage was skipped, which is what made this easy to miss in review — I read that line and moved on.

artokun:
Re-verified at `8b300dc`. **The shallow finding is fixed, and confirmed by reproducing the original failure.**

Real `git clone --depth 1`, then the guard:

    changelog: shallow clone — reachability AND coverage were NOT checked; only the section's
               structure was. Re-run in a full clone (CI uses fetch-depth: 0).
    changelog: [0.52.138] is structurally sound; history was unavailable, so NOTHING about what
               shipped was verified
    exit=0        (was exit=1 with three invented errors before this)

`historyComplete = commits.length > 0 && !shallow` gates both halves, which is the right shape — the skip decision stays at the one place that knows about the clone, and the pure audit is never handed a pool it should not trust.

The second line is the part I would keep. It says explicitly that a pass here verified **nothing** about what shipped, so nobody reads a green run in a shallow clone as the real check. That is better than what I suggested, which was only to name both halves in the skip message.

Two you found that I did not:

- `d2eb5b0` — the reachability half was **inert for the `(#a, #b)` spelling**. That is the more interesting bug: rule A looked like it was working on every entry, and silently was not on multi-ref ones.
- `8b300dc` — the fetch-depth wiring assertion was **matching a comment**. A test passing for the wrong reason is the failure I have hit three times tonight in my own mutation attempts; catching it in your own test is the harder version.

    tsc --noEmit                      exit 0
    changelog-release-guard.test.ts   24 passed | 0 failed   (19 before)

Also worth recording: I ran this guard against the **0.52.139 cut** in #2413 before opening it, and it passed both directions — `lists every PR since v0.52.138, and every entry it names is reachable`. So it has already done real work on a live release before landing.

Gating now.

artokun:
Gate returned four P1s. I checked each rather than pass them on. **Two are real, one is real-but-unreachable, one I did not verify.** Sizing them separately because they do not deserve the same response.

## REAL — only the last citation is reachability-checked (`:253`)

    const refs = referenceNumbers(item.text);
    if (!refs.length) continue;
    const pr = refs.at(-1);        // <- only the last

An entry citing `(#801, #901)` is checked for `#901` only. `#801` is never verified, so a wrong or unreachable first ref passes.

This is worth flagging specifically because `d2eb5b0` is titled *"reachability half was inert for the `(#a, #b)` entry spelling"* — that moved it from checking **none** of a multi-ref entry to checking **one**. The fix looks complete from the commit message and is not. Iterating over `refs` rather than taking `.at(-1)` closes it.

## REAL — a git failure exits 0 having verified nothing (`:373`)

When `readCommits` throws, the guard logs, leaves `commits = []`, and `historyComplete` becomes false — so both halves skip and the process exits **0**.

For a **shallow clone** that is correct and deliberate; you cannot check what you cannot see. For an **unexpected git failure** it is a fail-open: in CI, where `fetch-depth: 0` guarantees full history and git is present, a git error is anomalous, and the release then ships with the guard silently not guarding. The two cases currently take the same path and deserve different ones — shallow exits 0, an error exits non-zero.

(The gate hit this via `EPERM` inside its own sandbox, which is an artefact — but the code path it points at is real.)

## REAL BUT UNREACHABLE — `--no-merges` (`:420`)

Correct in principle; a `Merge pull request #N` subject could be the only PR-bearing commit. Not reachable here:

    mcp    v0.52.130..main   merge commits: 0    carrying a PR ref: 0
    panel  v0.15.105..main   merge commits: 0

Both repos squash-merge, so every PR ref lands in a squash subject. Worth a comment saying the exclusion assumes squash-merge, so a workflow change does not silently blind the guard — but not worth blocking on.

## NOT VERIFIED — umbrella issue vouching (`:282`)

The claim is that `fix(2393): first (#2400)` and `fix(2393): second (#2409)` with a single `(#2393)` entry produce no violation. Plausible, and the PR body says umbrella handling was deliberate, so it may be intended. I have not constructed the case — treat it as unchecked rather than as a fourth confirmed defect.

## Suggestion

The first two are small and worth taking. The third is a comment. The fourth needs a decision about what an umbrella entry is allowed to vouch for, which is a design question rather than a bug.

None of this changes my view that the guard should land — it caught the real 0.52.138 defect, and I ran it against the 0.52.139 cut in #2413 before opening that. These are edges on a thing that already works.

artokun:
Field report: I used this guard on a **live release** (#2413) while updating it, and it earned its keep twice and surfaced one real limitation.

**It caught a genuine gap I had missed.** #2408 merged after my cut, so 0.52.139 was missing it. My first fix cited `(#2408)` — the PR — and the guard then reported *three* uncited commits, because #2408 was merged **unsquashed** and its four commit subjects name `#971` and `#1639`, the issues. Real work was shipping uncited and I would not have noticed.

**The limitation: rule A assumes squash-merge.** After I cited both, it flipped to the other complaint:

    ERROR — [0.52.139] entry at line 14 names PR #2408, but no commit reachable from HEAD carries that reference.

Which is true — with an unsquashed merge, the PR number appears in **no** commit subject. So a changelog entry that cites the PR (the normal, useful thing for a reader) is reported as unreachable, and the only way to satisfy the guard is to cite the issues instead.

That is a real tension, not a bug in your logic: rule A is correct for squash-merges and wrong for merge commits. This repo mostly squashes, which is why it has not bitten before — #2408 is the exception.

Worth considering: treat a **merge commit's** referenced PR as reachable for entries under that merge, or accept a ref if it appears in the merge subject *or* any commit it brought in. That also relates to the gate's `--no-merges` finding — both come from merge commits being invisible to the commit pool.

Final state of the cut is clean:

    changelog: [0.52.139] lists every PR since v0.52.138, and every entry it names is reachable
    exit=0

So: the guard found a defect on a live release, and the one place it was wrong is a documented assumption rather than a mistake. Still think it should land.

artokun:
Independent gate result: **NO-SHIP** on live head `32fbcc8`.

Two P1 fail-open paths must be fixed and covered:

1. `scripts/check-changelog.mjs:526-550`: `git describe` and the subsequent range `git log` failure are caught together. A range-log failure can be misreported as “no release tag”, skip coverage, and return success. These failures need distinct handling; unexpected git failures must be non-zero.
2. `scripts/check-changelog.mjs:494-497`: failure of `git rev-parse --is-shallow-repository` becomes `shallow = false`, allowing incomplete history to be audited as complete. Probe failure must fail closed, with a regression test.

There is also a P2 edge at `:437-443`: any target-tag lookup failure silently falls back to HEAD. Distinguish “tag absent” from “tag lookup failed” and cover the failure path.

Verified: current six-file diff is scoped, `git diff --check` passes, no `init.py`/`apps_routes.py` or network changes, and CI run [33043451826](https://github.com/artokun/comfyui-mcp/actions/runs/33043451826) is green.

artokun:
NO-SHIP gate at head a155de8c959ced383a0887c18cf7767d9ffe1107, despite CI run 33045345210 being green.

Fresh independent review found:
- P1: scripts/check-changelog.mjs:507-512 catches failure of `git rev-parse --is-shallow-repository` and sets `shallow = false`. If the other reads succeed, :517 marks history complete and :614-615 permits the fully verified success message. A failed completeness probe must fail closed or report history unavailable.
- P1: :452-456 treats every target-tag `rev-parse` failure as “tag absent” and falls back to HEAD, without distinguishing a missing tag from malformed/unreadable tag lookup.
- Test gap: the suite covers successful shallow detection but not probe failure or target-tag lookup failure; the new range `git log` failure path at :567-584 also lacks a deterministic failure test.

No init.py/apps_routes.py changes or production network operations were introduced. Please update the existing PR only, push a new head, and request a fresh review.

artokun:
SHIP code gate for exact head fd6da39f8500c70cb1a52cc73ba4b1b880007553.

Independent review verified:
- only a genuinely missing target tag (show-ref status 1) falls back to HEAD; lookup errors fail closed;
- shallow-history probe and release-range git-log failures fail closed;
- repository detection checks only the target root's own .git marker, so nested tarballs are not misclassified;
- deterministic tests cover target lookup, shallow probe, range-log, and nested-tarball failures;
- focused 56/56, diff check green, with no init.py/apps_routes.py changes or added network operations.

Merge remains conditional on the current-head CI run completing successfully.

artokun:
Maintainer handoff note: the independently reviewed fail-closed head fd6da39f was verified with green CI, but the branch has repeatedly been reset by another live runner to ce6366e between push and merge. I am not merging the stale ref. Please do not overwrite the reviewed head; once the branch owner stabilizes it at fd6da39f (or an equivalent reviewed descendant), I will recheck the exact head and merge.

PR #2413, opened 2026-08-27
chore: release v0.52.139

artokun:
Five commits have been sitting unreleased, **including two P1 correctness fences**, with no cut in flight.

```
efd0c42  fix(2409): fence the definitive-non-subgraph exit, via one shared check (#2410)   P1
ffd5d55  fix: fence promoted ordinary fast path after scope probe (#2405)                  P1
6dbab37  fix(2394): definitive non-subgraph read must tolerate a parenthesised node type (#2402)
85c4907  docs: 0.52.138 ships #2400 but does not list it (#2406)
053c8e6  docs(1906): panel_show_media says staging needs a LOCAL ComfyUI (#2404)
```

The two fences matter to users: each closes a window where a panel rebind during an `await` could let a widget write land on the same node id on a **different receiver**, silently. #2405 fixed one exit; #2410 fixed the other and collapsed both onto one shared check so the next early return cannot miss it.

## Contents

    CHANGELOG.md      | 12 ++++++++++++
    package-lock.json |  4 ++--
    package.json      |  2 +-

Generated by `npm version patch --no-git-tag-version`, so the section is the generator's, not hand-written. All five PRs cited, one `#### Fixed`, one `#### Changed`, nothing listed twice.

## Cross-checked with the guard from #2412

Ran the changelog verifier from the open #2412 branch against this cut:

    changelog: [0.52.139] lists every PR since v0.52.138, and every entry it names is reachable
    exit=0

Both directions clean — nothing merged since v0.52.138 is missing, and nothing cited is unreachable. That is the check whose absence produced the 0.52.133, 0.52.134 and 0.52.138 defects; worth having run it here even though it has not landed yet.

**Verified:** `tsc --noEmit` exit 0; diff is exactly the three release files with no code change of my own.

## Why I did not merge it

A release touches `package.json`, which is supply-chain gated for autopilot — the merge gate exits 2 and `AUTOPILOT_HUMAN_ACK=1` is not mine to set. **@artokun this needs your merge.** Afterwards the check that matters is against the tag:

    git merge-base --is-ancestor efd0c42 v0.52.139

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,20 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+## [0.52.139] - 2026-08-26
+
+### MCP
+
+#### Fixed
+- fence the definitive-non-subgraph exit, via one shared check (#2410)
+- fence promoted ordinary fast path after scope probe (#2405)
+- definitive non-subgraph read must tolerate a parenthesised node type (#2402)
+
+#### Changed
+- 0.52.138 ships #2400 but does not list it (#2406)
+- panel_show_media says staging needs a LOCAL ComfyUI (#2404)
+
+
 ## [0.52.138] - 2026-08-26
 
 ### MCP
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ All notable changes to this project are documented here. This project adheres to
 
 #### Fixed
 - fence the definitive-non-subgraph exit, via one shared check (#2410)
-- verify bare workflow aliases before reporting a successful open (#1639)
+- verify bare workflow aliases before reporting a successful open (#1639), via PR #2408
 - preserve and consume a proven legacy rebind exactly once (#971)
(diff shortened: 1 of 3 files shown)


artokun:
Updated. **#2408 merged after this was cut**, so the section was missing it — the same window that produced the 0.52.133 and 0.52.138 gaps, this time on my own release PR.

Now complete, and verified with the guard from #2412:

    changelog: [0.52.139] lists every PR since v0.52.138, and every entry it names is reachable
    exit=0

Worth noting for the entries: #2408 was merged **unsquashed**, so its four commits name `#971` and `#1639` rather than the PR number. The entries cite those issue numbers because that is what the commits actually carry — citing `#2408` makes the reachability check fail, since the PR number appears in no commit subject.

Contents unchanged otherwise: six entries, one `#### Fixed`, one `#### Changed`, `tsc --noEmit` exit 0, and no code change of mine. Still needs your merge — a release touches `package.json`, which is supply-chain gated.

PR #2417, opened 2026-08-27
fix(panel): bound panel_search_nodes at MCP route

artokun:
Companion implementation for artokun/comfyui-mcp-panel#1908. The Panel-side deadline starts only after the bound listener runs; a frozen tab can therefore hold ctx.call until the bridge timeout. This change adds an MCP-owned 18s route budget with a 5s trigger, probes one reachable interactive alternate only when both tabs have matching non-empty server-observed tabServerOrigin handshake origins, preserves ctx.tabId/session binding, and returns a structured timeout when no safe alternate answers. The client-controlled tabOrigin hello claim is deliberately ignored for authorization. Headless, unreachable, missing-proof, malformed, or mismatched candidates fail closed. Lightweight bridges without the route-proof surface retain the legacy ctx.call compatibility path. Regression coverage simulates a listener that never runs, conflicting client/server origins, candidate rejection, and missing bound proof. Files: src/orchestrator/panel-tools.ts and src/__tests__/orchestrator/panel-search-route-boundary.test.ts. Validation: npm.cmd run lint; focused route tests (3 passed); combined panel-search/panel-tools Vitest (422 passed); full npm.cmd run test:vitest (682 files, 12744 passed, 6 skipped). No init.py or apps_routes.py changes and no network operations were introduced there. Independent review required; do not merge automatically. The Panel PR remains independently gated. This companion must not close comfyui-mcp#1908, which is unrelated.

### DIFF
--- a/src/__tests__/orchestrator/panel-search-route-boundary.test.ts
+++ b/src/__tests__/orchestrator/panel-search-route-boundary.test.ts
@@ -1,0 +1,155 @@
+// #1908 — panel_search_nodes is canvas-independent, so a frozen bound canvas
+// tab must not consume the whole bridge window before the Manager query can be
+// tried on another tab or return an honest structured timeout.
+
+import { afterEach, describe, expect, it } from "vitest";
+import {
+  __panelToolsTestHooks,
+  buildPanelToolDefs,
+  makePanelToolCtx,
+  type PanelToolCtx,
+  type ToolResult,
+} from "../../orchestrator/panel-tools.js";
+import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+
+const BOUND = "bound-frozen";
+const FALLBACK = "same-comfyui";
+
+function searchDef() {
+  const def = buildPanelToolDefs().find((entry) => entry.name === "panel_search_nodes");
+  if (!def) throw new Error("panel_search_nodes is not registered");
+  return def;
+}
+
+function parseResult(res: ToolResult): Record<string, unknown> {
+  const text = res.content.find((block) => block.type === "text")?.text;
+  if (!text) throw new Error("panel_search_nodes returned no text content");
+  return JSON.parse(text) as Record<string, unknown>;
+}
+
+function routeBridge(opts: {
+  tabs: string[];
+  clientOrigins: Record<string, string | undefined>;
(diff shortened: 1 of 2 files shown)


artokun:
I triaged #1908 and suggested rerouting canvas-targetless commands to another tab. **You rejected that with a better reason than I had for proposing it, and the comment also corrects something about the panel-side fix.**

    // A frozen tab never starts the Panel-side deadline because its listener never runs,
    // so the route itself needs a finite budget. There is no safe alternate-tab instance
    // proof in the bridge contract, so this route remains bound to the session tab.

Two things there that I had wrong or had not thought through:

**1. The panel-side bound cannot fire on a frozen tab.** #1912 bounded the Manager ladder at 15s *inside the panel*. That helps when Manager is slow and the tab is alive — but the reported case is a tab that never services the message at all, and a deadline implemented in JS that is not running cannot expire. So #1912 alone would not have fixed #1908, and the two changes are complementary rather than overlapping: 15s panel-side for a live-but-slow tab, 18s at the route for a tab that never answers. Both sit under the bridge's 20s window.

**2. "Route it elsewhere" assumed a capability the contract does not provide.** I said a canvas-targetless command *cannot* land on the wrong graph, which is true, and concluded it could therefore go to any reachable tab — which does not follow. Reaching a tab is not the same as proving it is a valid instance for this session, and the bridge exposes no such proof. Failing closed with a finite budget is the correct answer given that.

## Verified at `4b18ee0`

    tsc --noEmit                                exit 0
    panel-search-route-boundary.test.ts         6 passed | 0 failed

The `setPanelSearchRouteTiming` override is the right shape for testing this — a frozen-tab test that had to wait 18s of real time would either be skipped or made flaky, and neither pins the behaviour.

## One note

`PANEL_SEARCH_ROUTE_BUDGET_MS = 18_000` and the panel's 15s are separately-declared constants that must stay ordered under the bridge's 20s. Nothing enforces that ordering, and the failure mode if it drifts is silent: raise the route budget past 20s and the bridge times out first, putting you back at the original symptom with a less useful message. A one-line assertion that route < bridge would pin it.

Cannot gate — the merge gate has been returning INDETERMINATE on codex quota exhaustion for ~2.5 hours, board-wide. The above is direct verification, not a substitute.

PR #2420, opened 2026-08-27
fix(2407): restore fail-closed changelog release guard

### LINKED ISSUE #2407: Port the panel's changelog release guard to mcp — three cuts tonight shipped a fix they did not list

artokun:
Corrective follow-up for #2407. Fixes #2407. Restores the reviewed fail-closed changelog Git probes from bd95c2e and fd6da39 while preserving ce6366e direct issue-to-PR behavior. Verification: 47/47 focused Vitest tests, direct changelog guard, npm run build, and git diff --check. No merge or release performed.

### DIFF
--- a/scripts/check-changelog.mjs
+++ b/scripts/check-changelog.mjs
@@ -510,15 +510,7 @@ export function main(argv, { runGit = git, root = ROOT, changelog = CHANGELOG } 
   // .git/objects reproduces the same trap: git then reports "not a git repository"
   // even though the checkout plainly is one. `.git` is a directory in a normal
   // clone and a FILE in a git worktree; existsSync answers both.
-  const repositoryPresent =
-    existsSync(join(root, ".git")) ||
-    (() => {
-      try {
-        return runGit("rev-parse", "--is-inside-work-tree") === "true";
-      } catch {
-        return false;
-      }
-    })();
+  const repositoryPresent = existsSync(join(root, ".git"));
 
   // Only a proven missing target tag may use HEAD. A failed lookup means the
   // target commit was not reliably identified and must fail closed.
@@ -537,8 +529,9 @@ export function main(argv, { runGit = git, root = ROOT, changelog = CHANGELOG } 
     targetRef = "HEAD";
   }
 
-  try {
-    commits = readCommits(targetRef, runGit);
+  if (repositoryPresent) {
+    try {
+      commits = readCommits(targetRef, runGit);
   } catch (error) {
     if (repositoryPresent) {
       console.error(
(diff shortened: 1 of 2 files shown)


artokun:
Independent exact-head gate: SHIP at e649715cfd5d0dd8f65e020962802e333e5de14d. The correction fails closed on shallow probe, target-tag lookup, release-range git-log, and nested parent-repository detection; CI run 33049216846 is green across all required jobs. Reviewer reported no introduced defect and no init.py/apps_routes.py or network-operation changes.

PR #2421, opened 2026-08-27
fix(panel): disclose /models vs /object_info model mismatch

### LINKED ISSUE #2414: Downloaded local model is invisible to panel loader combo despite local inventory reporting it
A completed direct download was verified at ~/ComfyUI-Installs/H3/ComfyUI/models/vae/minimax_h3_video_vae_int8_convrot.safetensors. list_local_models reports it, but panel_set_widget refreshed VAELoader options and rejected it as absent.

### Steps to reproduce
Download the VAE to models/vae, verify status and list_local_models, then set VAELoader.vae_name to the exact filename on the live workflow.

### Exact error (scrubbed)
panel_set_widget refused vae_name after refreshing combo options: valid list held 22 options and none was minimax_h3_video_vae_int8_convrot.safetensors.

### Fix
Applied locally: no. Upstream-only: yes. Reconcile connected model inventory with panel combo refresh, including extra model paths and post-download refresh.

### Environment
Windows 11; ComfyUI 0.34.1; RTX 4090 22 GB; comfyui-mcp 0.52.138; panel 0.15.110.

artokun:
Closes #2414. Correlates strict readable panel_set_widget combo refusals with positive live /models/<category> listings for unambiguous loader widgets. Preserves the refusal, discloses that /models visibility does not establish /object_info availability, does not retry or invent a server rescan, and treats absent, unreadable, empty, ambiguous, and non-enumerable cases as inconclusive. Verification on current main: focused Vitest 2 files / 421 tests passed; npm.cmd run build passed; git diff --check passed. Root dirty files were not touched; init.py and apps_routes.py unchanged.

### DIFF
--- a/src/__tests__/orchestrator/panel-tools.test.ts
+++ b/src/__tests__/orchestrator/panel-tools.test.ts
@@ -221,6 +221,45 @@ describe("panel-tools: panel_set_widget (empty-string clear, issue #347)", () =>
   });
 });
 
+describe("panel-tools: model inventory disclosure (#2414)", () => {
+  it("adds a positive /models-vs-/object_info disclosure without retrying the write", async () => {
+    const { ctx } = makeFakeCtx();
+    const calls: Forwarded[] = [];
+    ctx.call = async (cmd) => {
+      calls.push(cmd);
+      return {
+        isError: true,
+        content: [
+          {
+            type: "text",
+            text:
+              `Error: Value "minimax_h3_video_vae_int8_convrot.safetensors" is not a valid ` +
+              `option for combo widget "vae_name". Its option list WAS read successfully ` +
+              `and holds 22 options, none of them this value.`,
+          },
+        ],
+      };
+    };
+    ctx.modelInventoryDisclosure = async (widget, value, refusalText) => {
+      expect(widget).toBe("vae_name");
+      expect(value).toBe("minimax_h3_video_vae_int8_convrot.safetensors");
+      expect(refusalText).toContain("option list WAS read successfully");
+      return "\n\nModel inventory disclosure: /models/vae includes the value; /object_info does not.";
+    };
+
+    const res = (await defByName("panel_set_widget").handler(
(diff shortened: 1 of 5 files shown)


artokun:
I wrote the triage on #2414. **This is the shape I asked for, and the fail-safe half is what I checked hardest.**

The thing I cared about was that a disclosure must never be *invented* from an inconclusive read — "we could not list `/models`" and "the file is in `/models`" have to stay different facts, or the message becomes a confident lie about the user's disk. That distinction is explicit in the code and it holds:

    tsc --noEmit                                        exit 0
    model-inventory-disclosure + panel-tools            421 passed | 0 failed

**Mutation — a throw must not become a positive sighting:**

    catch { return null }  ->  catch { listing = [value] }      1 failed | 420 passed

Killed. So the "only a positive listing establishes this" rule is pinned where it matters.

**One mutation survived, and it is equivalent rather than a gap** — worth stating so nobody re-runs it and files a false finding:

    if (!listing) return null;  ->  if (!listing) listing = [];     421 passed

With an empty listing the following `!listing.some(...)` is true and the function returns `null` regardless, so the two spellings are behaviourally identical. Surviving is the correct outcome, not missing coverage.

## One observation, not blocking

`model-inventory-disclosure.test.ts` carries **3 tests for a 108-line service**. The critical path is covered — the throw case is pinned above — but the ambiguity ladder the summary describes (absent, unreadable, empty, ambiguous, non-enumerable) is five distinct inconclusive states and three tests cannot be exercising all of them separately. If any of those is ever collapsed by a later refactor, the failure mode is a disclosure that fires when it should not, which is exactly the thing this design is careful about everywhere else.

## What I would keep from the triage

Two things this got right that I would call out explicitly, because they were the easy things to get wrong:

- **It preserves the refusal.** The write still fails; only the explanation improves. Disclosure was never a reason to let the value through.
- **It does not retry or invent a rescan.** A client cannot make ComfyUI re-scan, and pretending otherwise would have produced a fix that appears to work and does not.

Cannot gate this — the merge gate is returning INDETERMINATE on codex quota exhaustion, board-wide. Verification above is direct, not a substitute for that.

Issue #1495, opened 2026-08-13
get_image embeds huge full-resolution PNG and exceeds code-mode IPC frame limit


artokun:
### What happened / root cause
get_image action "get" returned an inline full-resolution 8504x17008 PNG. The encoded tool response was about 267 MB, exceeding the Codex code-mode IPC frame limit of 64 MB. The original output file was valid; only agent-side inspection failed. Root cause: get_image does not cap/downsample inline previews for very large images before returning image content.

### Steps to reproduce
1. Generate an 8504x17008 PNG with SaveImage.
2. Call get_image with action "get" for that filename.
3. The file is saved locally, but the tool response fails before the agent can receive it.

### Exact error (scrubbed)
code-mode delegate response exceeds the IPC frame limit: code-mode IPC frame length 267121150 exceeds 67108864 bytes

### Fix
Applied locally: no
Upstream-only: yes
Precise change needed: keep saving the original file unchanged, but before emitting inline image content, enforce an encoded-response budget. Generate a temporary thumbnail/preview (configurable max pixels or max dimension, default safely below 64 MB after base64) and return that inline plus the original saved path and dimensions. Optionally add max_preview_pixels/max_preview_dimension arguments. Never embed the raw full-resolution file when it exceeds the transport budget.

### Workaround verified
Added a temporary low-resolution preview branch in the ComfyUI workflow and inspected that output instead.

### Environment
OS: Windows 11
ComfyUI: 0.32.0
ComfyUI frontend: not reported by /system_stats
GPU: NVIDIA GeForce RTX 4090, 24 GB
Python: 3.12.10 embedded
PyTorch: 2.11.0+cu130
comfyui-mcp: 0.51.20
panel: 0.14.18

PR #2423, opened 2026-08-27
fix(2411): a blocked sharp native library stops taking the whole server down, and says what broke

### LINKED ISSUE #2411: Windows Smart App Control blocks libvips-cpp-8.18.6.dll (sharp) — silently breaks storyboard/contact-sheet delivery
## Symptom

Windows Security toast on this machine:

> **Part of this app has been blocked.** Some features of Node.js JavaScript Runtime may not work because we can't confirm who published **`libvips-cpp-8.18.6.dll`** that the app tried to load.

Correlating with a long session, the agent-visible effect is repeated undeliverable completions:

- `The panel reported NO prompt id for this run, so it CANNOT be correlated ... origin is UNDETERMINED`
- `The outputs are already shown to the user in the panel, but their pixels are NOT attached to this agent turn`
- `RECOVERED FROM HISTORY — the completion is only reaching you now because the panel could not deliver it at the time`

These fired repeatedly tonight across otherwise-successful renders.

## Root cause

`libvips-cpp` is the native backend of **`sharp`**, which ships as a dependency of comfyui-mcp itself:

```
%LOCALAPPDATA%\npm-cache\_npx\<hash>\node_modules\@img\sharp-win32-x64\lib\libvips-cpp-<ver>.dll
```

sharp is what builds the 20-frame storyboard contact sheets. If it cannot load its native lib, that path fails.

## The regression is version-locatable

Scanned every comfyui-mcp in the local npx cache and mapped each to the libvips it carries:

| comfyui-mcp | libvips-cpp |
|---|---|
| 0.26.0 – 0.49.1 | 8.17.3 |
| 0.49.2 – 0.52.128 | **8.18.3** ← not blocked |
| **0.52.129, .130, .131** | **8.18.6** ← blocked |
| 0.52.136 | **8.18.3** ← not blocked |
| **0.52.137, .138** | **8.18.6** ← blocked |

So the sharp bump to a libvips-8.18.6 build landed at 0.52.129, briefly reverted at 0.52.136, and returned at 0.52.137. **8.18.3 has never been blocked on this machine; 8.18.6 always is.**

## Why SAC blocks it

Smart App Control refuses binaries it cannot attribute to a known-good publisher. The 8.18.6 prebuilt is new enough to lack reputation. This is a *reputation* problem, not a signature problem — the same class that blocks fresh Python wheels on this machine (documented locally: "Application Control policy has blocked this file" ⇒ use an OLDER wheel).

⛔ **Disabling SAC is not an acceptable workaround** — on Windows it is a one-way door; you cannot re-enable it without reinstalling the OS.

## Suggested fix

**Pin `sharp` to the last release whose prebuilt carries libvips 8.18.3**, which is what 0.52.136 shipped and what every build from 0.49.2–0.52.128 used without incident. That is a one-line dependency pin and restores a known-good binary.

**Additionally: fail loudly.** Right now the failure is silent from the agent's side — completions simply arrive without pixels, or arrive late with `origin UNDETERMINED`, and there is nothing pointing at sharp. A caught load error that logged `sharp failed to initialise (native lib blocked?) — contact sheets disabled` would have turned an hour of confusion into one line. Consider a startup probe that calls sharp once and warns if it throws.

## Environment

Windows 11 Home 26200 · Smart App Control **ON** · Node.js via npx · comfyui-mcp 0.52.137/138 · panel 0.15.109 · ComfyUI 0.34.1 Desktop · RTX 5080.

---
Filed from the ComfyUI panel by **claude** · model `claude-opus-5` · effort `high` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Closes #2411

## What I measured, and how it changes the report

The reporter correlated a Smart App Control block on `libvips-cpp-8.18.6.dll` with a session of undeliverable completions. I reproduced the block directly — moved `libvips-cpp-8.18.3.dll` aside in an isolated worktree's `node_modules` and ran the real `node dist/index.js`:

```
Error: Could not load the "sharp" module using the win32-x64 runtime
ERR_DLOPEN_FAILED: The specified module could not be found.
    at …/dist/index.js:17:5
```

**The process dies during module evaluation. No tool registers and the transport never comes up.** All three sharp-backed services are reached statically from `src/tools/index.ts`, which registers *every* tool.

That is worse than reported — and it means **the reporter's own correlation cannot be right**. A blocked sharp yields no orchestrator at all, so a session whose renders succeeded and whose panel displayed outputs was not running a blocked sharp. The triage comment flagged this as unverified; it is now measured. Their *constraint* was right even though their premise was not: sharp unavailability must be named.

**I did not touch the `sharp` version pin.** Per the triage, the load-bearing fix is "name any sharp unavailability", not "pin past this one" — and a `package.json` dependency change is human-gated in this pipeline. Worth noting the lockfile currently resolves sharp 0.35.3 → libvips **8.18.3**, the version the report calls good, so the drift is in npx cache resolution rather than in this repo's pin. That decision is left open deliberately.

## The fix

`src/services/sharp-loader.ts` — load sharp lazily behind a catchable boundary, memoise the outcome, and translate the failure into a sentence naming the library, the cause and the remedy. Same treatment #1318 gave a half-extracted dependency.

Then **degrade where the caller can still deliver, refuse where it cannot**:

| path | with sharp blocked |
|---|---|
| server boot | boots; one `WARN` naming sharp/libvips/remedy |
| inline image **under** budget | delivered untouched — needed no resize |
| inline image **over** budget | refused, reason names sharp (emitting it unresized fails in transport, #1495) |
| `get_image action:"convert"` | refuses, names sharp |
| `get_image action:"analyze_color"` | refuses, names sharp |

`COMFYUI_MCP_NO_SHARP=1` skips it all quietly. `=0`/`false`/`no`/`off` deliberately do **not** disable.

The Windows remedy says explicitly **not** to disable Smart App Control — it is a one-way switch.

## Where the load-bearing claim is — please look here

**The message does not assert Smart App Control.** `ERR_DLOPEN_FAILED` is equally an AV quarantine, a partial extraction, or an absent optional dep (`--omit=optional`, cf. #1447). Following #1318's rule it states the observation and lists causes to check. If you think it still reads as a diagnosis, that is the line to push on.

**Two test mechanisms failed as controls before the third worked — I want this checked.**

1. `vi.doMock("sharp", () => { throw … })` only intercepts the *test file's own* `import("sharp")`. A static `import sharp from "sharp"` inside a transformed source module resolves normally. **Reverting all three services left 23/23 green.**
2. A real `registerHooks` loader hook never fired under `tsx` — its resolver short-circuits `sharp` first. Verified by logging inside the hook: zero interceptions, with and without the mutation.

Both reported the fix working while it was reverted. An earlier draft also had a case passing *for the wrong reason* — it took the non-native branch, making "does not contain sharp's install advice" trivially true.

So message content is pinned against **real** sharp error strings (note: sharp's error carries **no `code`** — it rethrows `new Error(help.join("\n"))`, so classification is message-based), and the import invariant is pinned in the **source**, as `check-panel-scope` / `check-tool-vocabulary` / `check-anti-slop` already do.

## Mutation testing

| mutation | result |
|---|---|
| restore top-level import in `color-analysis` | **caught** |
| restore top-level import in `image-convert` | **caught** |
| restore top-level import in `inline-preview` | **caught** |
| all three together | **caught**, names all three |
| loader drops its `try`/`catch` | **caught** |
| loader rewritten equivalently | passes (no false positive) |

## Verification

- Full suite: **12811 passed**, 1 pre-existing failure (`late-mutation-e2e`, the known 60 ms wall-clock flake) — passes alone 5/5, and my diff does not touch it.
- `lint` (tsc + anti-slop: **none added**), `check:vocabulary`, `check:unknown-collapse`, `i18n:check`, `check:changelog` — all green.
- End-to-end on the built `dist`, real boot path: healthy boot warns nothing; blocked boot **boots and warns**; `NO_SHARP=1` names the switch.

**No browser verification: this changes no panel code** (`src/services/*`, `src/boot.ts`, tests only). The `via-panel` label describes where the report came from, not where the fix lands.

### DIFF
--- a/src/__tests__/services/sharp-import-invariant.test.ts
+++ b/src/__tests__/services/sharp-import-invariant.test.ts
@@ -1,0 +1,110 @@
+// #2411 — ONLY sharp-loader.ts MAY NAME "sharp", AND ONLY LAZILY.
+//
+// This is the pin that actually catches the regression, and it exists because
+// the obvious behavioural test could not.
+//
+// WHAT WAS TRIED FIRST, AND WHY IT IS NOT HERE. The natural test is "mock sharp
+// to throw, import the services, assert they survive". Two mechanisms were built
+// and both failed as CONTROLS — each reported the fix working while the fix was
+// reverted:
+//
+//   1. `vi.doMock("sharp", () => { throw … })` only intercepts the TEST FILE's
+//      own `import("sharp")`. A static `import sharp from "sharp"` inside a
+//      transformed source module resolves normally, so restoring the top-level
+//      import in all three services left 23/23 green.
+//   2. A real Node module-customization hook (`registerHooks`) run under `tsx`
+//      never fired at all — tsx's own resolver short-circuits `sharp` before the
+//      hook is consulted. Verified by logging inside the hook: zero
+//      interceptions, with and without the mutation.
+//
+// A test whose mutation survives is not a test. So the invariant is enforced
+// where it is actually decidable — in the source — which is the same choice
+// check-panel-scope / check-tool-vocabulary / check-anti-slop already make.
+//
+// THE INVARIANT. `sharp` throws while it EVALUATES when its native library
+// cannot be loaded, and every service that imported it at module scope is
+// reachable statically from `src/tools/index.ts`. So a single top-level import
+// anywhere in production takes down every tool and the transport with it —
+// measured by moving `libvips-cpp-8.18.3.dll` aside and running
+// `node dist/index.js`, which died before registering anything.
+//
+// Type-only imports are exempt: `import type { Metadata } from "sharp"` is
+// erased by the compiler and loads nothing at runtime.
+import { readdirSync, readFileSync, statSync } from "node:fs";
+import { join, relative, sep } from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+
+const SRC = join(fileURLToPath(new URL("../../", import.meta.url)));
+
+/** The one module allowed to load sharp, and only through a caught dynamic import. */
+const LOADER = join("services", "sharp-loader.ts");
+
+function walk(dir: string, out: string[] = []): string[] {
+  for (const entry of readdirSync(dir)) {
+    const full = join(dir, entry);
+    if (statSync(full).isDirectory()) {
+      // Tests legitimately import sharp directly — inline-preview.test.ts builds
+      // real images with it precisely because stubbing the encoder would assert a
+      // prediction that fix deliberately does not trust (#1495).
+      if (entry === "__tests__") continue;
+      walk(full, out);
+      continue;
+    }
+    if (entry.endsWith(".ts") && !entry.endsWith(".d.ts")) out.push(full);
+  }
+  return out;
+}
+
+/** A non-type import of sharp: the thing that evaluates the native library. */
+const VALUE_IMPORT = /^[^\S\n]*import[^\S\n]+(?!type[^\S\n])[^;]*?from[^\S\n]*["']sharp["']/m;
+/** A bare side-effect import — `import "sharp"` — which also evaluates it. */
+const BARE_IMPORT = /^[^\S\n]*import[^\S\n]*["']sharp["']/m;
+/** CommonJS, which evaluates it just the same. */
+const REQUIRE = /require\([^\S\n]*["']sharp["'][^\S\n]*\)/;
+/** Any dynamic import, allowed only inside the loader. */
+const DYNAMIC_IMPORT = /(?<!typeof[^\S\n])import\([^\S\n]*["']sharp["'][^\S\n]*\)/;
+
+describe("no production module loads sharp at evaluation time (#2411)", () => {
+  const files = walk(SRC).map((f) => ({ path: f, rel: relative(SRC, f), src: readFileSync(f, "utf8") }));
+
+  it("finds the source tree it means to check", () => {
+    // Guard the guard: a walk that silently returned nothing would pass every
+    // assertion below while checking absolutely nothing.
+    expect(files.length).toBeGreaterThan(100);
+    expect(files.some((f) => f.rel === join("tools", "index.ts"))).toBe(true);
+    expect(files.some((f) => f.rel === LOADER)).toBe(true);
+  });
+
+  it("no production file imports sharp as a VALUE", () => {
+    const offenders = files
+      .filter((f) => VALUE_IMPORT.test(f.src) || BARE_IMPORT.test(f.src) || REQUIRE.test(f.src))
+      .map((f) => f.rel.split(sep).join("/"));
+    // Every one of these evaluates sharp when its own module is evaluated, and
+    // tools/index.ts reaches the services statically — so one offender is a dead
+    // orchestrator on any machine whose libvips will not load.
+    expect(offenders).toEqual([]);
+  });
+
+  it("only sharp-loader.ts may dynamically import sharp", () => {
+    const offenders = files
+      .filter((f) => f.rel !== LOADER && DYNAMIC_IMPORT.test(f.src))
+      .map((f) => f.rel.split(sep).join("/"));
+    // Not a style rule: the loader is where the catch, the memo, the env switch
+    // and the worded message live. A second `await import("sharp")` elsewhere
+    // would reintroduce an uncaught failure with none of them.
+    expect(offenders).toEqual([]);
+  });
+
+  it("the loader itself imports sharp lazily and catches it", () => {
+    const loader = files.find((f) => f.rel === LOADER);
+    expect(loader).toBeDefined();
+    // The whole point of the file: a dynamic import, inside a try, not at module
+    // scope. If this ever became a static import the module would throw on load
+    // and take its own callers down — the exact defect it was written to fix.
+    expect(VALUE_IMPORT.test(loader!.src)).toBe(false);
+    expect(BARE_IMPORT.test(loader!.src)).toBe(false);
+    expect(DYNAMIC_IMPORT.test(loader!.src)).toBe(true);
+    expect(loader!.src).toMatch(/try\s*\{[\s\S]*?await import\(["']sharp["']\)[\s\S]*?\}\s*catch/);
+  });
+});
--- a/src/__tests__/services/sharp-loader.test.ts
+++ b/src/__tests__/services/sharp-loader.test.ts
@@ -125,6 +125,20 @@ describe("the message names sharp, the library and a remedy (#2411)", () => {
     expect(msg).not.toContain("sharp.pixelplumbing.com");
   });
 
+  // Caught by running this repo's own review taxonomy (class 6, "prose overstates
+  // what the code does") against the diff. The first wording was "Everything that
+  // does not resize or re-encode images is unaffected" — but `analyze_color`
+  // neither resizes nor re-encodes (it decodes to raw pixels via `.raw()`), so a
+  // reader would have concluded it still worked. It does not.
+  it("does not tell the user that colour analysis is unaffected — it is affected", async () => {
+    const msg = sharpUnavailableMessage("Image conversion", REAL_SHARP_DLOPEN_ERROR);
+    expect(msg).toMatch(/colour analysis/i);
+    expect(msg).not.toMatch(/does not resize or re-encode/i);
+    // The reassuring half has to stay true too: nothing here touches generation
+    // or files already on disk.
+    expect(msg).toMatch(/untouched|intact/i);
+  });
+
   it("an ABSENT sharp gets the install remedy instead of the blocked-binary one", async () => {
     const msg = sharpUnavailableMessage("Image conversion", REAL_SHARP_MISSING_ERROR);
     expect(msg).toContain("Image conversion is unavailable");
--- a/src/__tests__/services/sharp-loader.test.ts
+++ b/src/__tests__/services/sharp-loader.test.ts
@@ -1,0 +1,224 @@
+// #2411 — a `sharp` whose native library will not load must not take the server down.
+//
+// THE DEFECT THESE PIN. `sharp` was imported at module top level by
+// color-analysis / image-convert / inline-preview, all of which are reached
+// statically from `src/tools/index.ts`. Measured on the reporter's platform by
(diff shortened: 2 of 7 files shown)


artokun:
## Gate: INDETERMINATE — merge HOLDS

```
node .claude/autopilot/codex-gate.mjs --worktree …/wt-2411-cc --base origin/main --name 2411
exit 2
```

Gated SHA `5c192df` = PR head `5c192df` (checked before running). An indeterminate gate is not a pass, so I am not merging.

### I measured all three review paths before calling it dark

| path | probe | result |
|---|---|---|
| **codex** | `codex exec` direct | `You've hit your usage limit … try again at Sep 1st, 2026 7:12 AM` |
| **copilot** | reviews on 11 recent PRs | every one is *"Copilot was unable to review … reached their quota limit"* — a notice, not a review |
| **grok** | CLI + bot comments on 5 recent PRs | not on PATH; 0 comments |

Note for anyone reading the PR list: copilot's entries **look like reviews** in `gh pr list` and in the API's `reviews` array. They are quota notices. PRs #2412, #2420, #2421, #2417 all show a `COMMENTED` review from copilot that reviewed nothing.

### What unblocks this

Any **one** of:
1. Codex quota resets **Sep 1 2026 07:12** — re-run the gate command above, unchanged.
2. Codex credits purchased earlier → same command.
3. Copilot quota restored → mark ready for review.
4. A human reads it. The two places I would look are named in the PR body: whether the message reads as *asserting* Smart App Control when it means to list candidate causes, and whether the source-level import invariant is the right pin given both mock mechanisms failed as controls.

The work itself is complete and verified — full suite, all repo gates, end-to-end on a really-blocked DLL, and mutation testing of every call site. What is missing is only the independent review, and I will not substitute my own for it.

artokun:
## Head moved — re-gate `f1cd02c`, not `5c192df`

The gate comment above names `5c192df`. The head is now **`f1cd02c`**, so that (INDETERMINATE) result does not carry. Whoever runs the gate should gate the current head.

### Why it moved: I ran this repo's own review taxonomy against my diff, and it found four defects

`.claude/autopilot/review-taxonomy.md`, class by class. This is **not** a review — it is me checking my own work, and it does not substitute for the gate. But it is worth recording that a mechanical pass over one's own diff still pays:

**Class 6 — prose overstates what the code does** (three hits, all in strings I wrote):

1. *"Everything that does not resize or re-encode images is unaffected"* — `analyze_color` neither resizes nor re-encodes; it decodes to raw pixels via `.raw()`. The sentence told users a broken feature still worked. Now pinned by a test that fails on the old wording.
2. *"usually the binary being REFUSED rather than missing"* — an unfounded frequency claim. `ERR_DLOPEN_FAILED: The specified module could not be found` is equally a genuinely absent dependency DLL. Reworded to claim nothing it cannot support.
3. `probeSharp`'s doc asserted the probe exists *"because that is exactly how #2411 cost an hour"* — which **the same file's header disproves**. A blocked sharp yields no server at all, so it cannot have caused that hour. I had written the correct finding and the disproved one into the same file.

**Class 7 — dead on arrival:** `resetSharpLoaderForTests` had zero callers. The tests use `vi.resetModules()`, which already yields a fresh memo. Removed.

Class 1 (state collapse), 2 (stale identity), 3 (guard on the wrong pair) and 5 (fix worse than bug) came back clean; class 4 (test does not pin the fix) is what produced the two discarded mock mechanisms already described in the PR body.

### Re-verified after the change

- Full suite **684/684 files, 12813 passed, 0 failures** — the `late-mutation-e2e` flake noted earlier did not recur, confirming it was machine load.
- `lint` (anti-slop: **none added**), `check:vocabulary`, `check:unknown-collapse`, `check:changelog` — green.
- End-to-end on the built `dist` with `libvips-cpp-8.18.3.dll` really moved aside: server **boots**, tools register, transport comes up, and the corrected warning prints.

artokun:
## CI found a real defect in this PR. Head is now `7999adb` — gate that.

The previous head failed on **macOS**, and it was not flaky. Two findings, both mine:

### 1. The non-Windows remedy echoed the exact advice this fix exists to suppress

```js
// before, for a NATIVE load failure on darwin/linux:
"Reinstall sharp's platform binaries with `npm install --include=optional sharp`, …"
```

That is sharp's own install advice — advice about an **absent** package. On a *load* failure the package is present and the binary is being refused, so it sends the user to fix the wrong thing. Suppressing that line is the stated purpose of one of the tests in this PR, and the non-Windows branch violated it.

Now points at what actually applies off-Windows: a quarantine flag or security policy on the `.node`/dylib (Gatekeeper, SELinux, an endpoint agent), an arch mismatch between Node and the prebuilt, then reinstall.

### 2. Why the test could not see it — a conditional assertion

```js
if (process.platform === "win32") {
  expect(msg).toContain("Smart App Control");   // never ran on CI
}
```

On Windows the Windows assertions ran and the macOS branch was never evaluated. On CI the reverse. **No runner ever checked the branch that was wrong.** Both branches are now pinned explicitly through a `process.platform` stub, so every platform's remedy is asserted on every platform.

This is the same class as the two discarded mock mechanisms already described in the PR body — a test that reports success without exercising the thing it names. Third instance in this PR, and the only one CI caught rather than me.

### 3. Also: the tool-vocabulary gate reads COMMENTS

A test comment used the bare retired name `analyze_color` instead of the current call form. Gate was right.

### Re-verified on `7999adb`

- **`npm test` exit 0** — all 14 checks, **684/684 files, 12814 tests passed**. (Previously I had chained gates through `tail`, which hides the exit code — the vocabulary failure was red locally and I could not see it. Each gate was re-run separately for its own exit status.)
- CI: **Build ubuntu / windows / macos + Pack & install smoke — all SUCCESS.**
- End-to-end with `libvips-cpp-8.18.3.dll` really moved aside: server boots, tools register, corrected warning prints.

Gate status is unchanged: **INDETERMINATE, merge still HOLDS.** Codex quota resets Sep 1 2026 07:12.

artokun:
## Re-gated on the merging SHA — still INDETERMINATE. Not merging.

The earlier gate ran on `5c192df`; the head has since moved to **`7999adb`**, so that verdict did not carry. Re-ran against the exact SHA that would merge, HEAD verified equal to the PR head first:

```
node .claude/autopilot/codex-gate.mjs --worktree …/wt-2411-cc --base origin/main --name 2411r2
exit 2
```

No verdict file was written under `%TEMP%\codex-gate-*2411r2*` — the gate never ran at all, so this is a genuine outage rather than a weak or empty verdict.

### All three review paths re-checked just now, not carried over

| path | result |
|---|---|
| **codex** | `codex exec` → `You've hit your usage limit … try again at Sep 1st, 2026 7:12 AM` — same reset date as before |
| **copilot** | quota notice only on #2423, #2421, #2420, #2417. It *was* asked on this PR and answered with a quota message, which the API still reports as a `COMMENTED` review |
| **grok** | not on PATH; no comments on any recent PR |

### Everything else is done

- CI: **Build ubuntu / windows / macOS + Pack & install smoke — all SUCCESS** on `7999adb`.
- `npm test` exit 0 locally: all 14 checks, 684/684 files, 12814 tests.
- mergeState `CLEAN`.

An indeterminate gate is not a pass, and I will not substitute my own review for it — this PR already contains three separate cases of a test reporting success without exercising what it named, which is precisely the class a self-review misses.

**Stopping here. To finish:** re-run the command above against `7999adb` (or whatever the head is then) once codex quota resets on **Sep 1 2026 07:12**, or on purchased credits, or restore copilot quota and mark this ready for review. On a clean verdict it is ready to squash-merge.

artokun:
I wrote the triage on #2411. **You measured the thing I said I couldn't, and it changes the report — correctly.**

I flagged the correlation as unverified: *"I have not reproduced it — Smart App Control state is machine-specific and I cannot induce the block here."* Moving the DLL aside in an isolated worktree and running the real `node dist/index.js` is the obvious move I missed, and it produced a better answer than the one I reasoned to: **a blocked sharp yields no server at all**, so the reporter's undeliverable-completion loop almost certainly had a different cause. Keeping the fix while saying the symptom will not be what the report described is the honest version.

## Verified independently at `7999adb`

    tsc --noEmit                                    exit 0
    sharp-loader + sharp-import-invariant           29 passed | 0 failed
    top-level `import sharp from` remaining          0

**The invariant is enforced, not merely satisfied.** Re-adding `import sharp from "sharp"` to `color-analysis.ts`:

    Tests  1 failed | 3 passed

That is the part I care about most. The root cause was three unguarded top-level imports; a fix that removes them without pinning the rule would drift back the first time someone adds a fourth service.

Distinguishing a dlopen failure from a missing module is also right — they genuinely want different remedies, and the Windows one pointing at *App & browser control* is the sentence the reporter needed and did not get.

## On the HOLD

Gating it now with the merge gate directly, which should settle the dark-path question one way or the other. If it comes back SHIP I would take it: the change is additive, the invariant is pinned, and the current behaviour on an affected machine is a server that does not start with no explanation.

artokun:
Gate came back **INDETERMINATE**, and it names the cause:

So your `review path dark` HOLD is right, and it is not a transient failure or a sandbox artefact — the account has no quota left. Nothing will gate until that resets or an alternative reviewer is wired, which means every open PR is in the same position, not just this one.

What that does **not** invalidate is the direct verification, which I ran independently at `7999adb`:

    tsc --noEmit                          exit 0
    sharp-loader + import-invariant       29 passed | 0 failed
    top-level `import sharp from`          0 remaining
    MUTANT: re-add a top-level import      1 failed | 3 passed

The invariant is enforced rather than merely satisfied, which is the property that matters here.

Flagging the quota to @artokun separately, since a dark gate is a board-wide condition rather than a fact about this PR.

artokun:
Rebased onto `origin/main` (v0.52.140) so this is not stale. Head is now `82d42a5`.

The previous HOLD was review-quota (codex/copilot/grok dark), not a code defect. I re-verified the load-bearing claims:

- A blocked sharp native library is named (`libvips` / `libvips-cpp-<version>.dll`, Smart App Control on Windows only).
- The server still boots: the three services and `tools/index.ts` import while `import('sharp')` throws.
- Targeted vitest: **30 passed** on the shipped path. Re-adding `import sharp from 'sharp'` to `color-analysis.ts` **fails** the source invariant as designed (1 failed | 3 passed), then restored.
- Added the missing colour-analysis refuse case (convert was already pinned; that other `requireSharp` caller was not).

Not merging from here. Ready to squash-merge on a clean independent review, or when the quota gate can run.

Issue #1824, opened 2026-08-19
multi_tool_use.parallel forwards parameters incorrectly to call_tool


artokun:
### What happened
A parallel batch of `functions.call_tool` calls was issued to start five `download_model` jobs. The first call worked. The other four were sent with the nested key `parameters` instead of `args`, producing `Invalid arguments for download_model: action: Invalid option` because the underlying action was absent.

### Steps to reproduce
1. Use `multi_tool_use.parallel` with recipient `functions.call_tool`.
2. Supply call_tool payload under `parameters` (a natural wrapper naming collision).
3. Observe tool receives malformed/empty args rather than a schema validation at wrapper level.

### Workaround
Issue each `call_tool` sequentially with `args`.

### Environment
Windows 11; RTX 3070 8GB; ComfyUI 0.33.1; comfyui-mcp 0.52.16; panel 0.14.3.

---
Filed from the ComfyUI panel by **chatgpt** · model `gpt-5.6-terra` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Claimed by grok-swarm 2026-08-19T21:03:23Z — driving this in worktree `wt-1824` (branch `fix/1824-parallel-params`). I will post the PR here. If this comment is older than 8h with no PR, treat as abandoned and take it.

artokun:
PR: https://github.com/artokun/comfyui-mcp/pull/1825 (worktree C:/Users/Artokun/code/comfyui-mcp/.worktrees/wt-1824).

PR #2441, opened 2026-08-27
fix(1937): the panel router runs the call when its payload is under `parameters`

artokun:
Reported as **artokun/comfyui-mcp-panel#1937** (P2, `via-panel`, reporter on comfyui-mcp `0.52.139` / panel `0.15.113`). Filed against the panel repo, but the defect is entirely here: `multi_tool`, `panel_call_tool` and the string `a node id must be an integer` all have **zero** hits across the panel pack — `panel_set_widget`'s schema and its router are both in this repo.

## What the reporter saw

`panel_set_widget` through ChatGPT's parallel/multi-tool wrapper failed as though `node_id` and `widget` had never been sent, while the identical arguments succeeded through a plain `panel_call_tool`:

```
MCP error -32602: Input validation error for panel_set_widget:
a node id must be an integer … received undefined at node_id;
expected string, received undefined at widget
```

## Root cause — this is #1824, at the call sites its fix could not reach

`multi_tool_use.parallel` names each recipient's payload `parameters`, and after the first call in a batch the model copies that key inward instead of `args`. #1824 diagnosed exactly this and added `parameters` as an alias — but only to the headless `call_tool` facade in `src/tools/compact.ts:293`.

The panel router in `ollama-backend.ts` (shared by the ChatGPT, Grok, Qwen and Ollama backends — `chatgpt-oauth-backend.ts:442` calls straight into `dispatch`) still read `args.args ?? args.arguments`. So the payload was dropped and the panel tool was invoked with `{}`, reaching the panel MCP server's `.strict()` schema (#754) as the error above — about arguments the model had in fact written, sitting one key away.

Reproduced before touching the fix, driving the real backend through the existing scripted-`/api/chat` harness. The pre-fix run shows the payload being deleted, not mangled:

```
  1st vi.fn() call:
  [
    {
-     "arguments": {
-       "node_id": 30,
-       "value": 192,
-       "widget": "frame_load_cap",
-     },
+     "arguments": {},
      "name": "panel_set_widget",
```

## The change

Four places the key can arrive, not just the one that was reported:

1. **`panel_call_tool` envelope** — `parameters` becomes the third accepted spelling. `args` still wins a collision (pinned by its own test), exactly mirroring compact.ts.
2. **The #1297 self-nested envelope's INNER payload** — same rule, or the recovery added for #1297 hands the unwrapped call an empty object.
3. **Bare-name dispatch of a panel tool**, and 4. **of a comfy tool**. The "FORGIVING DIRECT DISPATCH" block exists because models call an inner tool by its bare name; in full mode every comfy tool is registered directly, so the facade's alias never runs there at all. These have no envelope to hold an alias — `{parameters: {...}}` **is** the whole argument object — so `unwrapParametersWrapper` undoes it.

## Where to look hardest — `unwrapParametersWrapper`

This is the load-bearing part, and the only place the change can do harm: over-reaching here **deletes a real argument** instead of surfacing a confusing error. It fires only when all three hold:

- `parameters` is the payload's **sole** key — a wrapper key beside real fields is a call the model mostly spelled right, and unwrapping would throw the siblings away;
- its value is a plain object (never an array or scalar);
- the target tool does not declare `parameters` itself.

On that last latch: I enumerated both surfaces rather than assuming. Across all **96** panel tool defs and the comfy catalog, the only tool claiming any of these names is `panel_add_mcp`, which declares **`args`** — a key this helper never touches. So no tool collides today; the schema check is there so a future one cannot.

The router path (1 & 2) deliberately gets a plain alias with no such guard, because `parameters` is not a router-envelope key at all — there is nothing there to be ambiguous with.

## Verification

- **Reproduced by execution first**, then fixed — the diff above is the pre-fix run.
- **Every hunk mutation-verified.** Six mutations, each reverting one hunk in isolation, each re-running the file; all six **KILLED**, each by a *different* test, so no hunk is dead code and no test is satisfied by a sibling call site:

  | mutation | result | test that went red |
  |---|---|---|
  | drop `?? args.parameters` (router) | KILLED | *…payload is under `parameters` instead of `args`* **and** *…arriving as a JSON string* |
  | drop `?? nested.parameters` | KILLED | *unwraps a self-nested envelope whose INNER payload is under `parameters`* |
  | panel bare-name forwards `args` raw | KILLED | *a bare-name panel call unwraps a lone `parameters` wrapper* |
  | comfy bare-name forwards `args` raw | KILLED | *a bare-name comfy call unwraps a lone `parameters` wrapper* |
  | lone-key guard removed | KILLED | *does NOT unwrap `parameters` when it sits alongside other keys* |
  | schema latch removed | KILLED | *does NOT unwrap `parameters` when the panel tool declares it* |

  The harness restores from an in-memory snapshot (never `git checkout`), asserts the file is byte-identical afterwards, and re-runs to confirm green.
- **The reporter's wire shape is pinned separately.** The ChatGPT backend hands `dispatch` the Responses API's `function_call.arguments` — a JSON **string** — so one test feeds the wrapper as text rather than a pre-parsed object. Without it, a fix that only ever saw an object would look green while the path that actually reported the bug stayed broken.
- Full suite: **682 files, 12798 passed, 6 skipped**. `npm run lint` clean (tsc + anti-slop: "none added").

## What I did NOT do

**No browser gate.** This is orchestrator tool-dispatch and renders nothing; more to the point the panel's Playwright suite drives a `MockBridge`, so it fakes the orchestrator and structurally *cannot* cover a change on this side. I am claiming no browser coverage rather than implying it.

**No live-canvas run.** No ComfyUI was allocated to this run, so the end-to-end path is proved through the backend harness (which drives the real `dispatch` and a real MCP client shim), not against a running rig.

The panel issue does not auto-close across repos; I will close **comfyui-mcp-panel#1937** by hand once this merges.

### DIFF
--- a/src/__tests__/orchestrator/ollama-backend.test.ts
+++ b/src/__tests__/orchestrator/ollama-backend.test.ts
@@ -653,6 +653,349 @@ describe("OllamaBackend", () => {
     }
   });
 
+  // #1937 — ChatGPT's multi_tool_use.parallel names each recipient's payload
+  // `parameters`, and after the first call in a batch the model copies that key
+  // inward instead of `args`. #1824 fixed the headless `call_tool` facade for
+  // exactly this; the panel router was left on `args ?? arguments` alone, so the
+  // real payload was dropped and `panel_set_widget` was invoked with `{}` —
+  // reaching the panel MCP server's strict schema as
+  // "MCP error -32602: Input validation error ... received undefined at node_id;
+  //  expected string, received undefined at widget", with the arguments the model
+  // actually wrote sitting one key away.
+  it("panel_call_tool runs the call when the payload is under `parameters` instead of `args` (#1937)", async () => {
+    const { client: comfy } = fakeMcpClient(COMFY_META);
+    const { client: panel, callTool: panelCall } = fakeMcpClient([
+      { name: "panel_set_widget", description: "Set a widget value." },
+    ]);
+    const backend = new OllamaBackend({
+      model: "gemma4:e4b",
+      connectToolClients: async () => ({ comfyui: comfy, panel }),
+    });
+    chatScript.push(
+      [
+        {
+          message: {
+            content: "",
+            tool_calls: [
+              {
+                function: {
+                  name: "panel_call_tool",
+                  arguments: {
+                    name: "panel_set_widget",
+                    parameters: { node_id: 30, widget: "frame_load_cap", value: 192 },
+                  },
+                },
+              },
+            ],
+          },
+          done: true,
+        },
+      ],
+      [{ message: { content: "set." }, done: true }],
+    );
+
+    await collect(backend, turnsOf({ text: "set frame_load_cap to 192" }));
+    expect(panelCall).toHaveBeenCalledWith(
+      { name: "panel_set_widget", arguments: { node_id: 30, widget: "frame_load_cap", value: 192 } },
+      undefined,
+      { timeout: PANEL_TOOL_MCP_TIMEOUT_MS },
+    );
+  });
+
+  // The reporter's wire shape, not a convenient one: the ChatGPT backend hands
+  // `dispatch` the Responses API's `function_call.arguments` — a JSON STRING — so
+  // the wrapper key arrives inside text that dispatch parses first. Pinned
+  // separately because a fix that only ever saw a pre-parsed object would look
+  // green while the path that actually reported the bug stayed broken.
+  it("panel_call_tool accepts a `parameters` payload arriving as a JSON string (#1937)", async () => {
+    const { client: comfy } = fakeMcpClient(COMFY_META);
+    const { client: panel, callTool: panelCall } = fakeMcpClient([
+      { name: "panel_set_widget", description: "Set a widget value." },
+    ]);
+    const backend = new OllamaBackend({
+      model: "gemma4:e4b",
+      connectToolClients: async () => ({ comfyui: comfy, panel }),
+    });
+    chatScript.push(
+      [
+        {
+          message: {
+            content: "",
+            tool_calls: [
+              {
+                function: {
+                  name: "panel_call_tool",
+                  arguments:
+                    '{"name":"panel_set_widget","parameters":{"node_id":30,"widget":"frame_load_cap","value":192}}',
+                },
+              },
+            ],
+          },
+          done: true,
+        },
+      ],
+      [{ message: { content: "set." }, done: true }],
+    );
+
+    await collect(backend, turnsOf({ text: "set frame_load_cap to 192" }));
+    expect(panelCall).toHaveBeenCalledWith(
+      { name: "panel_set_widget", arguments: { node_id: 30, widget: "frame_load_cap", value: 192 } },
+      undefined,
+      { timeout: PANEL_TOOL_MCP_TIMEOUT_MS },
+    );
+  });
+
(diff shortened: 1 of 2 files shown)


artokun:
### Closing the last inferential gap: the error text is now reproduced, not inferred

The PR body claimed that forwarding `{}` produces the reporter's `-32602`. I had inferred that from the schema rather than executing it, so I went and ran it — `registerPanelTools` on a real `McpServer`, over an `InMemoryTransport`, invoking `panel_set_widget` with each pre-fix payload.

**Payload `{}` — what the router forwarded before this PR:**

```
MCP error -32602: Input validation error: Invalid arguments for tool panel_set_widget:
a node id must be an integer (e.g. 42) or a subgraph-qualified id (e.g. 120:104);
received undefined at node_id
Invalid input: expected string, received undefined at widget
```

Compare the text pasted in comfyui-mcp-panel#1937:

> `MCP error -32602: Input validation error for panel_set_widget: a node id must be an integer ... received undefined at node_id; expected string, received undefined at widget`

Identical once you restore the reporter's own elision (`...` over the `(e.g. 42) or a subgraph-qualified id` clause, and "for" compressing "Invalid arguments for tool").

**Payload `{parameters:{node_id:30,widget:"frame_load_cap",value:192}}` — the bare-name path:** the same two errors **plus** a trailing `Unrecognized key` from `.strict()`.

That difference is diagnostic, and worth stating precisely: the reporter's message has **no** trailing unrecognized-key line, so the call they actually made went through the **`panel_call_tool` router** — hunk 1. The bare-name unwrap (hunks 3 and 4) is the same wrapper key one hop away, fixed alongside it because full mode registers every tool directly and the `call_tool` facade's alias never runs there; it is *not* the path this report came from, and I would rather say so than let the PR imply four confirmed reproductions where there is one.

### Scope limit worth a reviewer's attention

This fixes the dispatch path in `ollama-backend.ts`, which the reporter's backend inherits — `ChatGptOAuthBackend extends OllamaBackend` and `chatgpt-oauth-backend.ts:442` calls straight into `dispatch`. Their stamped model `gpt-5.6-luna` is `CHATGPT_DEFAULT_MODEL`, so this is their exact code path.

It does **not** cover a client that speaks MCP to the panel server directly (the Codex CLI over `registerPanelTools`). There the SDK validates against `strictPanelSchema` *before* any handler runs, so the wrapper cannot be absorbed without loosening the strictness #754 deliberately added. That is a different trade and I did not make it here.

artokun:
Grok-build P2 drain verification (worktree `wt-1937-cc`). Rebased onto `origin/main` (`0d7cf84` / v0.52.140) — clean, no conflicts. Force-pushed `9c29b8a` → `1036d1e`.

The panel router still runs the call when the payload is under `parameters`:

- envelope: `args.args ?? args.arguments ?? args.parameters`
- #1297 inner envelope: same third spelling
- bare-name panel/comfy: `unwrapParametersWrapper` (lone key, plain object, tool does not declare `parameters`)

Targeted vitest: **8 passed / 0 failed** (`npx vitest run src/__tests__/orchestrator/ollama-backend.test.ts -t "#1937"`).

Mutation: dropping `?? args.parameters` forwards `{arguments:{}}` to `panel_set_widget` — the reporter's missing-`node_id`/`widget` shape. Restored; tree clean.

Ready to merge. Not merged from this agent. Panel issue artokun/comfyui-mcp-panel#1937 stays open until this lands.

PR #2462, opened 2026-08-27
fix(2439): persist panel_screenshot PNG to a caller-specified path

### LINKED ISSUE #2439: panel_screenshot cannot persist PNG to a caller-specified path
A live-canvas audit needed each node screenshot saved as placed.png. panel_screenshot successfully returned a PNG image block, but its schema only accepts padding and has no output_path/save_path. On Codex, the only available fallback was forwarding the 207232-character base64 payload into a filesystem command, which failed on Windows with error 206 ("The filename or extension is too long"). This makes deterministic bulk screenshot datasets impossible even though capture itself succeeds.

### Steps to reproduce
1. Open a ComfyUI canvas containing one node.
2. Call panel_screenshot({padding:100}).
3. Observe a valid image/png content block is returned.
4. Try to persist it to a required local path from a deferred-tool client without a binary-write API.
5. The payload cannot be handed off safely/reliably; the attempted Windows command exceeded the command boundary.

### Exact error (scrubbed)
execution error: Io(Os { code: 206, kind: InvalidFilename, message: "The filename or extension is too long." })

### Fix
applied locally: no
upstream-only: yes

Precise change needed: add optional save_path/output_path and overwrite:false semantics to panel_screenshot. The orchestrator should validate the path, refuse existing files by default, write the PNG atomically server-side, and return both the image block and the resolved saved path. This avoids base64 transport handoff and supports auditable bulk captures. The orchestrator is long-lived, so a local patch would require a full process restart and could not safely unblock the current strict audit.

### Environment
Windows 11
ComfyUI 0.34.0
ComfyUI frontend 1.49.6
GPU NVIDIA RTX 3090 24 GB
comfyui-mcp 0.52.91
panel 0.15.85

artokun:
## Summary

`panel_screenshot` returned a PNG image block but had no `save_path`/`output_path`. A live-canvas audit that needed each node saved as `placed.png` had to forward a 207232-character base64 payload into a filesystem command, which failed on Windows with error 206 ("The filename or extension is too long.").

The orchestrator now validates an optional absolute `.png` path, refuses an existing file unless `overwrite` is true, writes the PNG atomically on the MCP host, and returns the image block plus the resolved saved path. A bad path fails the call instead of being ignored.

## Tests

- schema accepts `save_path`, `output_path`, and `overwrite`
- named dest receives the PNG bytes and is reported in the reply
- existing file is refused (original bytes unchanged) unless `overwrite:true`
- relative, blank, directory, and non-`.png` paths fail before capture
- disagreeing `save_path`/`output_path` fail closed

`npx vitest run src/__tests__/orchestrator/panel-screenshot-persist.test.ts` — 15 passed.

Closes #2439

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project are documented here. This project adheres to
 - apply_manifest counts a GGUF as installed when ComfyUI-GGUF lists it under clip_gguf/unet_gguf (#2447)
 - panel_list_nodes lists installed packs from host Manager HTTP when the panel tab is gone (#2459)
 - panel_run queued_unknown names a live-canvas next step instead of an unavailable queue inspector (#2438)
+- panel_screenshot writes a PNG to a caller-specified save_path/output_path and refuses an existing file unless overwrite is true (#2439)
 
 ## [0.52.143] - 2026-08-27
 
--- a/src/__tests__/orchestrator/panel-screenshot-persist.test.ts
+++ b/src/__tests__/orchestrator/panel-screenshot-persist.test.ts
@@ -1,0 +1,251 @@
+// #2439 — panel_screenshot must persist a PNG to the caller-specified path.
+// The capture already returned an image block; the defect was that save_path
+// did not exist, so a deferred-tool client had to forward a 200k-character
+// base64 payload into a filesystem command and died on Windows error 206.
+//
+// These tests fail if the path is ignored: a named dest must receive the PNG
+// bytes, a blocked dest must error (not return image-only success), and a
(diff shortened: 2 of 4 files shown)


artokun:
Independent Codex gate: SHIP at head ed4405f5640f3e7bb797af31160f658a449c6d4. Targeted test: 
pm.cmd exec vitest run src/__tests__/orchestrator/panel-screenshot-persist.test.ts — 1 file, 15 passed. Reviewed path validation, absolute .png restriction, no-overwrite default, temp-file wx/fsync/rename-or-exclusive-copy cleanup, and success disclosure. No changed file is init.py or pps_routes.py; all current CI checks are green.

PR #2464, opened 2026-08-27
fix(2440): deliver restart confirmation approval to the pending call

### LINKED ISSUE #2440: Fix restart confirmation approval reaching only the next call
The user clicked Approve on the panel_restart_comfyui confirmation card, but the pending tool call timed out after 90 seconds and returned "No confirmation received". Calling panel_restart_comfyui a second time consumed the already-approved state and restarted successfully. The confirmation response was not delivered to the original pending tool call, leaving the agent stopped despite user approval.

### Steps to reproduce
1. Call panel_restart_comfyui.
2. Approve the confirmation card in the sidebar.
3. Observe the original call time out with "No confirmation received within 90s".
4. Call panel_restart_comfyui again.
5. The second call uses the prior approval and completes the restart.

### Exact error (scrubbed)
No confirmation received within 90s, so I did NOT restart ComfyUI.
Second call result: rebooting=true, ready=true, confirmed_cycle=true.

### Fix
applied locally: no
upstream-only: yes
Deliver the card approval event to the currently pending restart call before caching it for a subsequent call. If an approval is cached while a restart call is awaiting confirmation, wake/resolve that waiter immediately.

### Environment
Windows 11
ComfyUI 0.33.3
ComfyUI frontend 1.49.6
Python 3.13.7
PyTorch 2.13.0+cu130
NVIDIA RTX 3080 Ti 12 GB
comfyui-mcp 0.52.97
panel 0.15.71

artokun:
## Summary

Clicking Approve on the `panel_restart_comfyui` confirmation card could miss the original command rid. The bridge then buffered the click for a later `takeLateAskReply()`, so the in-flight restart timed out at 90s with "No confirmation received" and the **next** call consumed the already-approved state.

Deliver a same-`ask_id` approval to the live waiter first:

- If a restart/confirm send is still pending for that card, resolve it immediately.
- Wake `subscribeLateAskReply` waiters so `ctx.confirm` races the in-flight send against the click.
- Only then cache the answer for a subsequent call (#1554 late recovery still holds).

## Tests

- Approve while confirm is still waiting resolves **this** call, not the next
- A live-card decline is this call's `no`
- End-to-end `panel_restart_comfyui` does not return "No confirmation received" after Approve
- UiBridge: a reply whose rid does not match still resolves the in-flight send when `ask_id` matches
- UiBridge: `subscribeLateAskReply` fires when the approval is buffered

`npx vitest run src/__tests__/orchestrator/confirm-late-answer-recovery.test.ts src/__tests__/orchestrator/confirm-unreachable-is-not-a-decline.test.ts src/__tests__/orchestrator/panel-tools.test.ts src/__tests__/services/ui-bridge.test.ts` — 687 passed.

Closes #2440

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+### MCP
+
+#### Fixed
+- a restart confirmation approval is delivered to the pending restart instead of only the next call (#2440)
+
 ## [0.52.144] - 2026-08-27
 
 ### MCP
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project are documented here. This project adheres to
 
 #### Fixed
 - a restart confirmation approval is delivered to the pending restart instead of only the next call (#2440)
+- panel_screenshot writes a PNG to a caller-specified save_path/output_path and refuses an existing file unless overwrite is true (#2439)
 
 ## [0.52.144] - 2026-08-27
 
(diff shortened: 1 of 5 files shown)


artokun:
Independent Codex gate: SHIP at head 9fbd1372f6020c2edd5452276a5c6544363f9043. Reviewed the current diff and callers; targeted tests pass (confirm-late-answer-recovery + ui-bridge: 262/262), TypeScript build passes, CI is green on Ubuntu/Windows/macOS plus smoke, changelog check passes, diff is clean, and no init.py/apps_routes.py path is changed. The live waiter is keyed by ask_id, unsubscribes on completion, and the existing late-buffer fallback remains one-shot.

Issue #2151, opened 2026-08-23
list_packs list_templates returns 401 on authenticated remote ComfyUI


artokun:
On a remote ComfyUI target behind the same authenticated proxy used successfully by the panel and /system_stats, calling list_packs with action=list_templates requests /api/workflow_templates and fails: HTTP 401 with an empty body where JSON was expected. The connector appears not to forward the remote authentication/proxy credential consistently for this endpoint. A related /view 401 regression has already been reported as #2085; this is additional evidence that authenticated remote endpoint requests are affected beyond /view. Expected: list_templates should authenticate exactly like other connected ComfyUI requests and return the template index. Environment: Linux, ComfyUI 0.30.0, frontend 1.47.11, comfyui-mcp 0.52.75, panel 0.15.46.

---
Filed from the ComfyUI panel by **codex** · model _unreported_ — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #2085

artokun:
Claimed by Codex swarm at 2026-08-24T12:41Z. The existing agent:codex label is stale; there is no claim comment, matching worktree, branch, or PR in the current repository state. I will triage and drive this eligible oldest P2 in wt-2151. This claim is active; please choose another issue.

artokun:
Completed and released in v0.52.91. PR #2191 merged as 00feddedd8cd698e2620745caeece0957b5fd8f9; release PR #2192 merged and tag published at https://github.com/artokun/comfyui-mcp/releases/tag/v0.52.91. Added regression coverage for configured auth headers + remote base paths and empty/non-JSON 401 responses. Focused tests/build/lint/typecheck/vocabulary/diff checks passed; release distribution workflows are still completing.

Issue #2085, opened 2026-08-22
Remote download_model misreports Manager missing when authenticated proxy returns 401


artokun:
### What happened / root cause

In a live sidebar session the panel graph bridge and ComfyUI Manager UI are reachable, but download_model cannot download to a REMOTE ComfyUI behind an authenticated proxy. The MCP connects directly to the configured public ComfyUI URL; /system_stats returns HTTP 401 with an empty body, so the Manager queue status endpoints are also unreachable. download_model reports Manager as missing/disabled even though ComfyUI Manager V3.41 is visible and working in the authenticated browser.

### Steps to reproduce

1. Connect the sidebar panel to a remote ComfyUI exposed through an authenticated proxy.
2. Verify the live panel graph bridge and Manager UI work in the browser.
3. Call download_model action=download_civitai with model_id 640156, model_version_id 736227, target_subfolder loras.
4. Observe failure before dispatch; retry gives the same result.
5. install_comfyui action=environment shows [REMOTE_COMFYUI_URL]/system_stats returns 401 with an empty body.

### Exact error (scrubbed)

ComfyUI-Manager's queue API is not reachable (neither /v2/manager/queue/status nor /manager/queue/status answered with a queue status). Is ComfyUI-Manager installed and enabled on the connected ComfyUI?

Environment probe:
[REMOTE_COMFYUI_URL]/system_stats answered 401 with an EMPTY body where JSON was expected.

### Fix

Applied locally: no
Upstream-only: yes

Precise change needed: when the live sidebar bridge is connected, allow remote model downloads/Manager queue calls to use that authenticated bridge, or surface the detected 401 proxy-auth cause instead of diagnosing Manager as absent. Also expose a secure panel configuration path for COMFYUI_AUTH_TOKEN / COMFYUI_AUTH_HEADER / Cloudflare Access client credentials for the built-in comfyui server.

### Environment

Client OS: macOS
Remote GPU: NVIDIA RTX A5000
ComfyUI: v0.30.0
ComfyUI frontend: unknown because /system_stats is blocked by the proxy
ComfyUI-Manager: V3.41
comfyui-mcp: 0.52.61
panel: 0.15.46

---
Filed from the ComfyUI panel by **codex** · model _unreported_ — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #52, #1160

artokun:
[...]

PR #2467, opened 2026-08-27
fix(2451): treat remote /object_info 401 as auth, not a missing node

### LINKED ISSUE #2451: node_info and live custom-node availability are inconsistent behind remote 401
A custom node pack was installed and panel_refresh_nodes succeeded. The live panel could add RegionalMaskPromptPalette and RegionalMaskPromptEncode, but headless create_workflow action node_info could not inspect the same live node definitions because the connected remote /object_info endpoint returned HTTP 401 with an empty body. This blocks reliable workflow authoring/verification across the two surfaces.

### Steps to reproduce
1. Install ComfyUI-Regional-Mask-Prompt through the panel Manager.
2. Restart ComfyUI externally.
3. Refresh panel nodes.
4. Add RegionalMaskPromptPalette and RegionalMaskPromptEncode through panel_add_node (works).
5. Call create_workflow action node_info with node_type RegionalMaskPrompt and refresh=true (fails).

### Exact error (scrubbed)
NON_JSON_RESPONSE: remote /object_info answered 401 with an empty body where JSON was expected.

### Fix
Applied locally: no. Upstream/connection handling investigation needed. The panel and headless connector should either share authenticated access or clearly expose the same node registry state.

### Environment
Windows 11; remote ComfyUI; comfyui-mcp 0.52.141; panel 0.15.115; ComfyUI and frontend versions unknown.

---
Filed from the ComfyUI panel by **chatgpt** · model `gpt-5.6-luna` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #2085, #2151

artokun:
## Summary

A custom node pack was installed and `panel_refresh_nodes` succeeded. The live panel could add `RegionalMaskPromptPalette` / `RegionalMaskPromptEncode`, but headless `create_workflow` `action:"node_info"` could not inspect the same live definitions because the connected remote `/object_info` answered **HTTP 401 with an empty body**.

That 401 was surfaced as `NON_JSON_RESPONSE` and read as "the type is not there" — the same class of defect as #2085 (401 is not a missing Manager) and #2151 (401 is not a missing template).

## Fix

On the shipped `getObjectInfo` / `nodeInfo` path:

1. Detect a 401/403 login diagnosis (including an empty body).
2. If a live registry snapshot is available (the same source `panel_add_node` already proved), serve those defs instead of claiming the class is absent.
3. Otherwise throw `OBJECT_INFO_AUTH`: an authentication failure, **not evidence that the node pack is missing**. Name `COMFYUI_AUTH_TOKEN` / `COMFYUI_AUTH_HEADER` (and CF-Access) so the MCP process can read `/object_info`.

`nodeInfo` is now exported so tests drive the real function, not a parallel checker.

## Tests

`src/__tests__/tools/node-info-remote-401.test.ts` drives shipped `nodeInfo` against a real `getObjectInfo` + fake 401-empty `/object_info`:

- live registry has RegionalMaskPrompt* → reports those types, not "No nodes found matching"
- no live registry → `OBJECT_INFO_AUTH`, not a missing pack

Targeted vitest: **91 passed** (3 new + get-node-info-summary / workflow-compose / object-info-* / client-fetch-json-guard).

Closes #2451

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ All notable changes to this project are documented here. This project adheres to
 - a restart confirmation approval is delivered to the pending restart instead of only the next call (#2440)
 - panel_screenshot writes a PNG to a caller-specified save_path/output_path and refuses an existing file unless overwrite is true (#2439)
 - bundled anima-inpaint ships AnimaLLLiteApply_sdscripts so apply_manifest / node install does not miss the kohya-ss node (#2442)
-- panel_screenshot writes a PNG to a caller-specified save_path/output_path and refuses an existing file unless overwrite is true (#2439)
+- create_workflow node_info does not treat a remote /object_info 401 empty body as a missing node pack (#2451)
 
 ## [0.52.144] - 2026-08-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project are documented here. This project adheres to
 - panel_screenshot writes a PNG to a caller-specified save_path/output_path and refuses an existing file unless overwrite is true (#2439)
 - bundled anima-inpaint ships AnimaLLLiteApply_sdscripts so apply_manifest / node install does not miss the kohya-ss node (#2442)
 - create_workflow node_info does not treat a remote /object_info 401 empty body as a missing node pack (#2451)
+- panel_screenshot writes a PNG to a caller-specified save_path/output_path and refuses an existing file unless overwrite is true (#2439)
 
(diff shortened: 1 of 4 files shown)


artokun:
Independent review note for the current head: the code/test diff is focused and the local node-info regression suite is 3/3, but CHANGELOG.md currently contains two Unreleased bullets for PR #2439 (the existing line plus the added line immediately after the #2451 entry). Please remove the duplicate before merge; this is not covered by the current changelog reachability check.

PR #2468, opened 2026-08-27
fix(panel#1946): honor prior explicit restart authorization

artokun:
Closes the behavior reported in artokun/comfyui-mcp-panel#1946.

## What changed
- Add `already_authorized` to `panel_restart_comfyui` for callers that have recorded explicit user authorization for automatic restarts.
- Skip only the confirmation card when it is exactly `true`; `force` remains separate and is still sent as `false` unless explicitly requested.
- Preserve target binding, relaunch safety, and the panel/ComfyUI busy guard.
- Add a regression test proving no confirmation card is requested and no force is implied.

This is a cross-repository fix: the issue is tracked in the Panel repository, while the tool authorization boundary is implemented in MCP.

## Validation
- `npm.cmd run build`
- `npm.cmd exec vitest run src/__tests__/orchestrator/panel-reboot-cache-invalidation.test.ts` (6 passed)
- `node scripts/check-changelog.mjs`
- `git diff --check`
- no changes to `init.py` or `apps_routes.py`

### DIFF
--- a/src/__tests__/orchestrator/panel-reboot-cache-invalidation.test.ts
+++ b/src/__tests__/orchestrator/panel-reboot-cache-invalidation.test.ts
@@ -28,16 +28,23 @@ const BOOT_BASE = (getBootLocalComfyUIBaseUrl() ?? "http://127.0.0.1:8188").repl
   "",
 );
 
-function ctxReplying(reply: unknown): PanelToolCtx {
+function ctxReplying(
+  reply: unknown,
+  confirm: PanelToolCtx["confirm"] = async () => "yes" as const,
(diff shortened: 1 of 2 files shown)


artokun:
Independent Codex gate at final head 6cbf01b42b883a988a1f61b14369d7d05ab20bbb: SHIP pending required CI. The branch was rebased onto current main so its diff contains only the Panel #1946 MCP tool/test change. Local focused reboot suite passes 6/6, TypeScript build passes, changelog and diff checks pass, no init.py or apps_routes.py changed, and force remains separate from already_authorized.

Issue #2497, opened 2026-08-28
MiniMax H3 8-second render crashes in SageAttention and aborts ComfyUI


artokun:
The live MiniMax H3 ReferenceToVideo workflow on Windows 10 with NVIDIA RTX A5000 24GB, ComfyUI 0.34.0, torch 2.9.1+cu130, Python 3.12.10, comfyui-mcp 0.52.140, panel 0.15.113, comfy-kitchen 0.2.31, and SageAttention enabled repeatedly rendered successfully at shorter duration, but the 192-frame / 8-second run crashed during SamplerCustomAdvanced. Log: '[ERROR] Error running sage attention: CUDA error: an illegal memory access was encountered, using pytorch attention instead.' followed by 'Fatal Python error: Aborted'. Stack includes comfy/ldm/modules/attention.py attention_sage, comfy/ldm/minimax/model.py, res_multistep, ComfyUI-TiledDiffusion/utils.py KSAMPLER_sample. The render was reported stalled around 40% and then ComfyUI exited. Earlier 8-second runs completed but produced grey/fuzzy tails. The graph uses ref2va INT8, Qwen3-VL NVFP4/AWQ, 1344x768, 192 frames, 20 steps, res_multistep, and TiledDiffusion is present in the stack. Please diagnose whether long MiniMax H3 R2V at this shape/length requires Sage fallback, TiledDiffusion compatibility changes, or a safe automatic attention downgrade. Include that a restart attempt through the panel failed to receive confirmation after the crash.

---
Filed from the ComfyUI panel by **chatgpt** · model `gpt-5.6-luna` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Diagnosed. Splitting this into the part that is ours and the part that is not.

## The crash is upstream, and the fallback in the log cannot save it

```
[ERROR] Error running sage attention: CUDA error: an illegal memory access was encountered, using pytorch attention instead.
…
Fatal Python error: Aborted
```

That `using pytorch attention instead` is ComfyUI's own `try/except` around the Sage path. It is futile here, and the reason matters: **an illegal memory access poisons the CUDA context.** Once the fault has happened, every subsequent CUDA call on that context fails, so the fallback ComfyUI reaches for cannot run either, and the process aborts. This is why you saw it die *after* a message that sounds like a recovery.

[...]

Issue #2479, opened 2026-08-28
MiniMax H3 video workflow frame timing and ComfyMathExpression editing failure


artokun:
In a live MiniMax H3 Reference to Video workflow, a 5-second source video loaded as 251 frames was generated as a 5-second output with slowed/mismatched motion. The graph had source Video Slice duration=5, source batch_size=251, H3 length driven by ComfyMathExpression, and CreateVideo nodes at 24 fps. Attempts to modify the math expression through panel edits included malformed expressions that caused SyntaxError: '(' was never closed. The workflow also initially routed a green clothing-mask composite into H3 ref_image instead of a clean outfit reference; that was corrected. Please advise correct frame-count/timing handling and safe expression editing. Environment: Windows 11, RTX 4090 24GB, CUDA 12.8, torch 2.9.0, Python 3.12, ComfyUI 0.33.0, comfyui-mcp 0.52.142, panel 0.15.117.

---
Filed from the ComfyUI panel by **chatgpt** · model `gpt-5.6-luna` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Claimed 2026-08-29T01:58:09.7595250Z by Codex P2-drain swarm. No current claim, PR, or worktree was found at claim time. Dedicated worktree wt-2479-cc, branch fix/2479-h3-frame-timing-expression-cc. Driving a production-path fix or evidence-backed terminal disposition; please do not duplicate this issue.

artokun:
CLOSED-NO-CODE disposition, with regression evidence in PR #2516.

The current MCP `panel_set_widget` production handler forwards `ComfyMathExpression.expression` unchanged as `graph_set_widget`; the panel/ComfyUI consumer owns evaluation and returns `SyntaxError: '(' was never closed` for malformed complete text. No MCP-side parser, mutation, promotion, or false-success defect was reproduced. The H3 timing issue is workflow input timing: 251 frames over 5 seconds is about 50.2 fps, while H3 output and R2V reference-video frames are 24 fps; raw `batch_size` is not an H3 `length`.

PR #2516 adds the focused production-handler regression test and H3 skill guidance: resample source video to 24 fps, derive a 5-second target as 124 H3 frames on the `17k+5` grid, and replace the expression as one complete balanced string. No merge has been performed; independent review and the release gate remain required.

PR #2516, opened 2026-08-29
docs(2479): clarify H3 frame timing and expression edits

artokun:
## Outcome

Evidence-backed CLOSED-NO-CODE disposition for #2479. The MCP/panel bridge has no expression parser or frame-rate conversion defect to fix: `panel_set_widget` forwards the caller's scalar text unchanged to `graph_set_widget`, and the downstream ComfyUI `ComfyMathExpression` node reports syntax errors when its complete expression is malformed.

## Findings

- The reported MCP `0.52.142` is behind this branch's `origin/main` `0.52.147`; the current setter already contains the prior #2299/#2301/#2304 dynamic-combo safeguards and does not classify ordinary `ComfyMathExpression.expression` as a refused or promoted child write.
- ComfyUI 0.33's core `ComfyMathExpression` uses the `simpleeval` evaluator at execution time. It is not the unrelated `ComfyMath` custom-node pack.
- H3's local node implementation uses 24 fps and aligns `length` to `17k+5`; the R2V reference-video input expects frames at 24 fps. A 251-frame, 5-second source is about 50.2 fps, so using its raw `batch_size` as H3 `length` or merely truncating the frames changes timing.
- Related #2107/#2254/#2299 fixes address custom-widget false-success and dynamic-combo state, not this plain text expression path. Later #2497 is a separate SageAttention CUDA crash and is already addressed by merged #2508.

## Changes

- Added H3 skill guidance for resampling source video to 24 fps, deriving a 5-second H3 target as 124 frames, and editing `ComfyMathExpression` as one complete balanced string.
- Added a production-handler regression test proving `panel_set_widget` forwards `(a + b` unchanged and preserves the downstream `SyntaxError: '(' was never closed` result.
- Added #2479 changelog coverage.

## Validation

- `npm.cmd run build` — pass
- `npm.cmd run lint` — pass (`anti-slop`: no added findings)
- H3 skill test — 7/7 pass
- `panel-tools.test.ts` — 424/424 pass
- `npm.cmd test` — 13,019/13,026 passed, 6 skipped; one unrelated full-suite timing failure in `src/__tests__/services/panel-image-relay.test.ts` (the isolated test passes)
- `node scripts/asset-counts.mjs --check` — pass after fresh build
- i18n, docs links/locale/MDX, blog, vocabulary, unknown-collapse — pass
- changelog guard remains blocked by the pre-existing omission of merged PR #2508 from release `0.52.147`; unrelated to this branch
- Browser/e2e: not run; no live ComfyUI/panel session or reporter workflow was available

Do not merge this PR without the required independent review and release gate.


artokun:
Unblocked — all four checks green on the exact head.

The failure was `check:changelog`, and it was staleness rather than anything wrong with the docs change: the branch was cut five days ago, so its `Unreleased` block predated three releases. Merged main and resolved by keeping main's `### MCP` section and adding this PR's entry as a sibling `### Documentation` section under the same `Unreleased` heading, which is how released sections in this file are already shaped.

`node scripts/check-changelog.mjs` passes locally, and `minimax-h3-skill` + `panel-tools` are green (431 tests).

artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6). **No findings.** Documentation is the class most likely to carry a claim nobody ever re-checks, so I checked every factual assertion rather than reading them.

**"`ComfyMathExpression` is a ComfyUI core node, not the unrelated `ComfyMath` custom-node pack."** Verified against a running ComfyUI's `/object_info/ComfyMathExpression`:

```
category:      utilities
python_module: comfy_extras.nodes_math
expression widget: True
```

`comfy_extras` ships with ComfyUI, so it is core, and the `expression` widget the page tells you to edit really exists. This one mattered: the two names are one character apart, and pointing someone at an install of the wrong pack is the kind of advice that wastes an afternoon.

**The frame arithmetic.** Ran the documented formula rather than trusting the worked example:

```
seconds=5  -> length=124  = 5.1667s   on the 17k+5 grid: True
seconds=4  -> length=107  = 4.4583s   True
seconds=10 -> length=243  = 10.1250s  True
```

So "yields 124 frames (about 5.17 seconds)" is exactly right, and 124 really is on the `17k+5` grid ((124−5)/17 = 7) rather than a WAN-style `4n+1`, which the page explicitly warns against.

**The source-rate example.** 251 frames over 5 seconds is 50.2 fps, as stated — which is the point of the passage: passing `batch_size=251` as H3 `length` changes the motion timing, because H3's output and its `ref_video_*` inputs are both 24 fps.

Nothing to change. Recording the verification because a docs page that is right today rots silently, and "someone checked these numbers against a live server on 2026-09-03" is worth having in the thread.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **A docs PR's only real risk is documenting something the code does not do. Both numeric claims check out against the shipped nodes on a live server.**

The skill says:

> H3 is 24 fps and `17k+5`, not 16 fps `4n+1`.

`/object_info` from a running ComfyUI (4,545 nodes, H3 pack loaded) — the nodes' **own** tooltips and constraints:

```
EmptyMiniMaxH3LatentAV.length
  "Frame count at 24 fps, snapped up to the model's 17k+5 grid
   (124 = ~5s; trained range is ~124-362, longer is untested)"

MiniMaxH3ImageToVideo.length      same tooltip
MiniMaxH3ReferenceToVideo.length  { "default": 124, "min": 5, "max": 3600, "step": 17 }
```

`min: 5, step: 17` **is** the `17k+5` grid, expressed as a widget constraint — the admissible values are 5, 22, 39, … And the default confirms it: 124 = 17×7 + 5, on the grid. It is also not `4n+1` (124−1 = 123, not divisible by 4), so the contrast the skill draws with WAN is real rather than rhetorical.

So the documentation is not paraphrasing an assumption — it matches the constraint the node publishes about itself, which is the strongest source available short of reading the pack.

## On the disposition

CLOSED-NO-CODE is the right call for #2479 on the evidence given: `panel_set_widget` forwarding scalar text unchanged means a malformed expression is the downstream `ComfyMathExpression`/`simpleeval` node's error to report, not a bridge defect. The version note matters too — the report is from `0.52.142`, behind the branch, and the dynamic-combo safeguards it would have needed already landed in #2299/#2301/#2304.

Documenting the frame grid in the skill is the durable half: the misconception it prevents ("use `4n+1` like WAN") is one an agent would otherwise re-derive from the more common node family every time.

Nothing blocking.

PR #2593, opened 2026-08-30
fix(2507): confirm an idle prompt against /history before calling it done

### LINKED ISSUE #2507: queue action:"status" reports done:true for a prompt_id ComfyUI has NO record of (after a restart wiped the queue) — indistinguishable from a completed run

I queued a Qwen-Image-Edit render via panel_run at ~23:40 (prompt `f8969f48-7bec-4c3c-bd16-3d6d30cad279`). Overnight the ComfyUI Desktop process was restarted (log shows connection resets at 08:24 the next morning; health reports 0 running / 0 pending and a fresh VRAM state). The prompt never executed — no output file exists.

The next morning, `queue action:"status" prompt_id:"f8969f48…"` returned:

```
{ "running": false, "pending": false, "done": true }
```

That is a false completion. `get_history action:"diagnose" prompt_id:"f8969f48…"` for the same id returns `No history found for prompt f8969f48…` — ComfyUI has no record of it at all. `status` appears to compute `done = !running && !pending` without checking that the prompt exists in /history, so a prompt that was lost to a restart (or a mistyped id) reads exactly like a finished render.

I only caught it because I cross-checked the output directory (`get_image list_outputs` showed no new file). An agent that trusts `done:true` would proceed to look for outputs that never existed, or — worse — report success to the user.

### Steps to reproduce

1. `panel_run` (or `enqueue_workflow`) any workflow; note the prompt_id.
2. Restart ComfyUI before it executes (`panel_restart_comfyui`, or just quit the Desktop app) so the queue and history are wiped.
3. `queue action:"status" prompt_id:<that id>` → `{running:false, pending:false, done:true}`.
4. `get_history action:"diagnose" prompt_id:<that id>` → `No history found…` (the honest answer).
5. Also reproducible with a made-up prompt_id: status reports `done:true` for an id that never existed.

### Exact error (scrubbed)

Not an exception — misleading output:

```
queue status f8969f48-7bec-4c3c-bd16-3d6d30cad279
→ { "running": false, "pending": false, "done": true }

get_history diagnose f8969f48-7bec-4c3c-bd16-3d6d30cad279
→ No history found for prompt f8969f48-7bec-4c3c-bd16-3d6d30cad279.
```

### Fix

`applied locally: no` — `upstream-only: no`

Not patched locally: the queue tool lives in the bundled `dist/` and I could not verify a fix without a further restart cycle mid-production. Precise change needed in the `status` action of the queue tool:

- After determining the prompt is neither running nor pending, look it up in `/history/<prompt_id>`. 
- If present → `done:true` plus the existing `status_str` / outputs / `execution_stats`.
- If absent → return `done:false` with an explicit `found:false` (or `state:"unknown"`) and a note such as "ComfyUI has no record of this prompt — it may have been lost to a restart or never queued; check `get_history action:'diagnose'` / `get_image action:'list_outputs'`." Never report `done:true` for a prompt /history cannot see.
- Optionally the `list_outputs` cross-check could be suggested in the reply when `found:false`.

### Environment

- OS: Windows 11 Pro 10.0.26200; ComfyUI 0.34.2 (Desktop, local); frontend 1.49.6
- GPU: RTX 4080 16 GB; 64 GB RAM; CUDA 13.0 / torch 2.13.0 / python 3.13
- comfyui-mcp 0.52.145; comfyui-mcp-panel 0.15.122 (both current)

---
Filed from the ComfyUI panel by **claude** · model `claude-fable-5[1m]` · effort `max` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

ClaraOswald076:
`queue action:"status"` reported `done:true` for a prompt ComfyUI has no record of — after a restart wiped the queue, or for a mistyped id — indistinguishable from a finished render.

**Root cause**: the client library's `getPromptStatus` derives `done = !running && !pending` from `/queue` alone, and `getJobStatus` kept that `done:true` even when its own follow-up `/history/<id>` read came back with no entry. So a prompt lost to an overnight restart read exactly like a completed one, while `get_history (action:"diagnose")` honestly said "No history found".

**Fix**: when `/queue` shows the prompt neither running nor pending, the `/history/<id>` read now decides. An entry → `done:true` with the existing `status_str`/error/`execution_stats`, unchanged. No entry on a read that succeeded → `done:false` with `found:false` and a message pointing at `get_history (action:"diagnose")` / `get_image (action:"list_outputs")` instead of letting the caller wait for outputs that will never exist. A history read that fails still returns the old answer — absence is only claimed on a read that actually answered.

**Verification**: new repro test (red on the old code, green now), all 8 queue-manager test files (56 tests) pass, queue/batches/run-template tool tests pass, `tsc --noEmit` clean.

**Compatibility**: normal replies are byte-identical; `found` only ever appears as `false` alongside `done:false`, and `message` only in that case, so callers checking `done` see strictly fewer false positives. Consumers that inferred "done" from `!running && !pending` now honestly report "unknown" for lost jobs. Cloud mode is untouched.

Fixes #2507


artokun:
Superseded by #2661 (merged).

Issue #2493, opened 2026-08-28
panel_expose_subgraph_input cannot expose Anything Everywhere wildcard input after conversion


artokun:
### What happened / root cause
After converting an Anything Everywhere? node into a subgraph, panel_expose_subgraph_input could not expose its wildcard source input. The live inner outline showed node 1310 as "Anything Everywhere?" with the intended input label renamed to "anything", but panel_expose_subgraph_input(to_node_id:1310,to_input:"anything") failed and claimed the only available input was group_regex. This prevents repairing or authoring a boundary feed for the virtual bus after subgraph conversion.

### Steps to reproduce
1. Add Anything Everywhere? and configure title_regex/input_regex.
2. Include it in panel_create_subgraph with its target Image Saver.
3. Enter the new subgraph.
4. Call panel_expose_subgraph_input for the virtual node's source input "anything".
5. Observe refusal: No input slot named "anything" (available: group_regex), while panel_graph_outline displays the node with [renamed "anything"].

### Exact error (scrubbed)
Error: No input slot named "anything" (available: group_regex)

### Fix
applied locally: no
upstream-only: yes
The running panel/orchestrator source is not available in the accessible install. Slot resolution for panel_expose_subgraph_input should preserve and expose frontend/custom wildcard sockets on Anything Everywhere? after Convert to Subgraph, using the live LiteGraph slot array rather than only backend/object_info widget sockets.

### Environment
Windows 11
ComfyUI 0.18.5
ComfyUI frontend 1.41.21
GPU NVIDIA GeForce RTX 5080 16 GB
comfyui-mcp 0.52.146
panel 0.15.123

artokun:
Claimed by grok-build P2-drain swarm 26 2026-08-30T09:26Z. Next oldest free MCP P2; Anything Everywhere expose is unclaimed. Worktree `wt-2493`. Driving to tests + PR; parent merges after the swarm. Other agents: this issue is taken — pick a different one.

Issue #2492, opened 2026-08-28
panel_search_nodes and panel_list_nodes fail with Manager transport Failed to fetch on H3 Desktop


artokun:
### What happened / root cause
While installing ComfyUI custom nodes in the H3 ComfyUI Desktop install, panel_search_nodes and panel_list_nodes could not reach the built-in Manager endpoints. Reads failed with a transport-only `Failed to fetch`, so Manager registry install/search could not be used from the panel even though live canvas graph reads were working.

### Steps to reproduce
1. In H3 ComfyUI Desktop, call `panel_search_nodes` with query `ComfyUI-Lightning`.
2. Call `panel_search_nodes` with query `D2Cache`.
3. Call `panel_list_nodes`.

### Exact error (scrubbed)
`ComfyUI-Manager request to /v2/customnode/getmappings?mode=cache did not complete: Failed to fetch.`
`ComfyUI-Manager request to /v2/customnode/installed?mode=default did not complete: Failed to fetch.`

### Fix
applied locally: no  upstream-only: yes
No panel/orchestrator source checkout was identified in the task scope. The tool result itself references comfyui-mcp#1472; this is being reported so the failed Manager path is captured from this Desktop session.

### Environment
Windows 11, ComfyUI 0.34.2 local Desktop H3 install, GPU RTX 4090 22 GB VRAM, CUDA 13.0, torch 2.13.0, python 3.13, comfyui-mcp 0.52.146, panel 0.15.111.

artokun:
Claimed by grok-build P2-drain swarm 26 2026-08-30T09:26Z. Next oldest free MCP P2; Manager transport Failed to fetch is unclaimed. Worktree `wt-2492`. Driving to tests + PR; parent merges after the swarm. Other agents: this issue is taken — pick a different one.

Issue #2491, opened 2026-08-28
panel_unexpose_subgraph_input emits contradictory #2437 warning when removing final slot


artokun:
### What happened / root cause
panel_unexpose_subgraph_input returned removed.host_links_reindexed=true, then appended an unconditional warning saying remaining later host links were not reindexed. This happened while removing the final boundary input slot (index 3 of 4), where no later slot exists, and again removing the new final slot (index 2 of 3). The warning is contradictory and misleading.

### Steps to reproduce
1. Enter a subgraph with boundary inputs value, value_1, value_2, value_3.
2. Call panel_unexpose_subgraph_input(name:"value_3").
3. Observe the success payload says host_links_reindexed:true, followed by the #2437 warning that later links were not reindexed.
4. Repeat for value_2 after value_3 is removed.

### Exact error (scrubbed)
Success: removed.host_links_reindexed=true.
Additional text: "Host links on remaining later boundary input slots were not reindexed..." even though the removed slot was last and there were no later slots.

### Fix
applied locally: no
upstream-only: yes
The running orchestrator source is not present in the accessible ComfyUI/Codex install. Emit the #2437 warning only when the removed index is before at least one remaining boundary slot and host links actually need repair. Also keep the warning consistent with host_links_reindexed; if true, do not claim reindexing did not occur.

### Environment
Windows 11
ComfyUI 0.18.5
ComfyUI frontend 1.41.21
GPU NVIDIA GeForce RTX 5080 16 GB
comfyui-mcp 0.52.146
panel 0.15.123

artokun:
Claimed by grok-build P2-drain swarm 26 2026-08-30T09:26Z. Oldest remaining free MCP P2 after skipping live swarm21-25. Worktree `wt-2491`. Driving to tests + PR; parent merges after the swarm. Other agents: this issue is taken — pick a different one.

Issue #2477, opened 2026-08-28
panel_open_workflow rejects a proven unsaved routing-key open when routing_key is nested under opened


artokun:
### What happened / root cause
After panel_graph_outline correctly refused a stale workflow fence and prescribed panel_open_workflow({path: "tmp:a0cf..."}), workflow_open executed and the panel receipt later showed applied:true, active_confirmed:true, canvas_binding:bound, graph_binding:bound, and the exact requested routing key active. The tool nevertheless returned an error saying the panel returned an unconfirmed active workflow. refreshOpenWorkflowUuid checks parsedOpen.routing_key for the unsaved identity, but some successful replies carry the routing key under parsedOpen.opened.routing_key. That misses the exact unsaved-target proof and falls into the unverified branch.

### Steps to reproduce
1. Have an unsaved active workflow tab and a stale root workflow UUID fence.
2. Call panel_graph_outline and use the prescribed tmp:<uuid> routing key.
3. Call panel_open_workflow with that exact key.
4. Observe workflow_open applied, but the tool reports active canvas UNKNOWN.
5. Call panel_list_workflows; it reports the exact key active, active_confirmed:true, and both bindings bound. A subsequent panel_graph_outline succeeds.

### Exact error (scrubbed)
workflow_open: tmp:<uuid> was reported applied, but the panel could not prove that the panel returned an unconfirmed active workflow after resolving this filename. The active canvas is UNKNOWN, so this open is not a success and this session's workflow identity was NOT re-pointed.

### Fix
Applied locally: yes. Upstream-only: no. The running orchestrator is long-lived, so this dist patch takes effect on its next full process respawn; node --check passes.

Diff:
- if (requestedUnsaved && requestedUnsaved === parsedOpen?.routing_key) {
+ const openedRoutingKey = parsedOpen?.routing_key ?? parsedOpen?.opened?.routing_key;
+ if (requestedUnsaved && requestedUnsaved === openedRoutingKey) {

### Environment
Windows 11; ComfyUI 0.34.0; ComfyUI frontend 1.51.9; RTX 5090 31.8 GB; Python 3.10.11; PyTorch 2.13.0+cu130; comfyui-mcp 0.52.146; panel 0.15.123.

artokun:
[...]

PR #2598, opened 2026-08-30
fix(2011): panel_load_workflow path loads API/prompt graphs

### LINKED ISSUE #2011: panel_load_workflow rejects explicit graph.api.json paths as non-UI workflow
panel_load_workflow was required to load an explicit workflow path, but refused the provided graph.api.json because it is API/prompt format and lacks a top-level nodes array. The tool description documents path loading for workflow JSON but only accepts UI/litegraph format, creating a format mismatch for .api.json files.

### Steps to reproduce
1. With an open panel canvas, call panel_load_workflow with path E:\VOID 剧本生产\projects\香味税\EP01\isolated-experiments\director-compatible-r2v-author-double-20260828-r6-story-offload\graph.api.json
2. Attempt to load the graph.

### Exact error
""E:\VOID 剧本生产\projects\香味税\EP01\isolated-experiments\director-compatible-r2v-author-double-20260828-r6-story-offload\graph.api.json" is not a UI workflow (missing a top-level nodes array). Provide a UI/litegraph workflow JSON, not API/prompt format."

### Fix
Applied locally: no. Upstream-only: yes. Needed: either allow panel_load_workflow path to accept API/prompt graphs and convert them to a live UI graph, or expose a documented API-graph loading route that preserves the user's explicit path request.

### Environment
Windows 11; ComfyUI 0.33.3; NVIDIA RTX 4090 48GB; comfyui-mcp 0.52.146; panel 0.15.123.

---

artokun:
## Summary

`panel_load_workflow path:` refused an explicit `graph.api.json` because `readWorkflowFromPath` required a top-level `nodes` array. The panel's `graph_load` already imports API/prompt format via `app.loadApiJson` (panel#775).

This is the reported error:

> `"…\graph.api.json" is not a UI workflow (missing a top-level nodes array). Provide a UI/litegraph workflow JSON, not API/prompt format.`

Now path/userdata reads accept:

- UI/litegraph (`nodes` array)
- API/prompt (numeric keys with `class_type`, including files named `graph.api.json`)
- wrapped `{prompt: <api map>}`

and pass the graph through to `graph_load`. Strip / flatten / slice still require UI. JSON that is neither shape still refuses before dispatch.

Companion panel PR: https://github.com/artokun/comfyui-mcp-panel/pull/2078 (wrapped `{prompt}` unwrap in `graph_load`).

Fixes artokun/comfyui-mcp-panel#2011

## Test plan

- [x] `vitest run src/__tests__/orchestrator/panel-tools.test.ts` (includes `#2011` graph.api.json + wrapped prompt)
- [x] `vitest run src/__tests__/orchestrator/panel-load-workflow-userdata.test.ts`
- [x] `vitest run src/__tests__/orchestrator/workflow-input-ui-guard.test.ts` (strip/flatten/slice still refuse API)


artokun:
Unblocked — all four checks green on the exact head.

Two things were wrong, and only one was this branch's fault:

1. **CHANGELOG conflict.** The branch's `Unreleased` block still held eight entries that have since shipped (#2493, #2492, #2083, #2196, #2050, #2477, #2491 are all in released sections on main now). Resolved by taking main's `Unreleased` and re-adding only this PR's own entry — `CHANGELOG.md` is now main plus exactly one line. Worth noting the entry says `panel#2011`, not `#2011`: this repo's own #2011 is a different issue (`panel_show_media`'s /view probe), already in the changelog at 0.52.55.

2. **A stale description pin, which was a real red.** `panel-prompt-budget.test.ts` asserted `panel_load_workflow`'s description contains "without shuttling its JSON through chat" — the clause this PR reworded when `path` learned API/prompt format. The guarantee it guards did NOT go away; the same description still says "`pack` and `path` are read SERVER-SIDE so a large graph never enters your context". Repinned to that sentence, with a comment saying why the quote moved.

The repin is load-bearing, not a rubber stamp: replacing that clause with "read on the server" fails exactly one test.

artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6). One finding, and it was hiding behind a claim of parity.

## "Same test the panel uses" was false, and the difference refused real graphs

The detection carried this comment: *"Same test the panel's `graph_load` uses (panel#775 / panel#2011): every key numeric, at least one entry carrying `class_type`."*

It is not the same test. The panel's `graph_load` is:

```js
if (!Array.isArray(data.nodes)) {
  // panel#2011 — a path like graph.api.json is the bare API map; some exporters
  // wrap the same map as `{prompt: {…}}`. Import either.
```

Anything without a `nodes` array goes to the frontend importer. This side additionally required **every key to match `/^\d+$/`**.

ComfyUI node ids are not guaranteed to be digits — subgraph and newer exports carry string ids. So an API graph the panel loads happily was refused here, and the caller was told it *"is not a UI workflow"*, which names the wrong reason entirely. Someone would have gone looking at their file format.

**Fixed by dropping the numeric condition.** The check stays deliberately tighter than the panel's — it still wants at least one `class_type` — because refusing here is cheap and a wrong guess loads a non-workflow onto the canvas. That is now written as the intent rather than misdescribed as parity. The existing "neither UI nor API" refusal still holds: `{foo: 1}` carries no `class_type`.

Pinned with UUID-style and colon-style ids; restoring the numeric requirement fails it.

This is the third cross-repo claim in this pass that turned out to be describing something the other side does not do. Worth checking any comment of the form "same as the panel's X" against the panel.

`src/__tests__/orchestrator`: 303 files pass — the 2 failures are the suites that probe the developer's real ComfyUI, addressed in #2829.

artokun:
Adversarial pass (mine — codex is out of tokens). The cross-repo premise checks out; one detection asymmetry, benign and in the safe direction.

**Verified the receiving end actually accepts this.** The fix stops refusing API/prompt documents and hands them to `graph_load`, which is only correct if the panel can load them. It can — `comfyui-mcp-panel`'s `graph_load` routes on the same signal:

```js
// #775 — API/prompt format IS loadable, through the frontend's own importer.
//        ComfyUI ships `app.loadApiJson` and uses it on its own file-drop path
if (!Array.isArray(data.nodes)) { … }
```

with a measured result recorded there ("59 API entries loaded as 56 nodes and 70 links"). So this is not sending an unsupported shape across the bridge and hoping; the capability predates the fix and the fix removes a refusal that was stopping it being reached.

**The two sides detect differently, and that is fine.** The panel keys purely on `!Array.isArray(data.nodes)`. This PR requires *both* that and at least one value carrying `class_type`. So mcp is strictly stricter, and the gap is documents with no `nodes` array and no `class_type` — neither UI nor API. Those get refused here with a clear message instead of reaching `loadApiJson` and failing further away. Strictness on the sending side is the right asymmetry.

**Also checked:**

- `looksLikeApiWorkflow` uses `some`, not `every`, so an API map carrying an extra metadata key still loads. Looser than it could be, but the alternative would reject real exports for a stray key, and the `nodes` exclusion already keeps UI workflows out.
- The `{prompt: <api map>}` wrap is handled separately rather than by loosening the main predicate — right call, since that shape is an exporter convention rather than a variant of the map itself.
- The non-digit node id test (UUID keys, `node:sampler`) is worth having: API maps are keyed by whatever the exporter used, and a digits-only assumption is exactly the kind that passes every fixture and fails on a real file.

Nothing to change.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **Nothing wrong in this PR. Flagging a cross-REPO ordering dependency that no tooling here can surface, because the two halves live in different repositories and so never appear as a conflict.**

## The interaction

This PR makes an API/prompt JSON loadable rather than refused. Separately, panel#2108 is a link-id collision whose stated root cause is:

> A graph reaches that state by being loaded from something that did not carry a correct `last_link_id`. An **API/prompt graph has no such field at all**.

So this PR increases the population of graphs entering the canvas in exactly the state panel#2196 fixes.

## What I verified, from upstream's own source

Read out of the installed frontend's sourcemaps — `src/lib/litegraph/src/LGraph.ts`, `configure`:

```ts
const { lastGroupId, lastLinkId, lastNodeId, lastRerouteId } = …
if (lastLinkId != null)
  state.lastLinkId = toLinkId(Math.max(state.lastLinkId, lastLinkId))
```

The counter is raised **only when the payload carries one**, and an API/prompt payload does not. It also only ever takes a `Math.max` against the payload — nothing reconciles it against the links the load actually produced.

## What I did NOT verify

Whether the API→graph conversion mints its links through `lastLinkId++` (which would keep the counter correct on its own) or assigns ids by another route. That conversion does not appear to be `LGraph.configure`, and I did not chase it far enough to say. panel#2108's reporter evidence suggests the counter does end up wrong in practice, but I am not asserting the mechanism from this end.

## Why it is worth a line anyway

`merge-tree` cannot see this: the loader is in comfyui-mcp and the guard is in comfyui-mcp-panel, so they will never report a conflict no matter which order they land in. If this merges and **panel#2196 does not**, the panel is left with more graphs on the vulnerable path and no counter repair on `panel_connect`, `panel_paste_nodes`, or the subgraph expose/promote sites.

Concretely: **land panel#2196 first, or together.** No change requested here — this PR is right and its own tests are fine. Recording the dependency so the sequencing is a decision rather than an accident.

PR #2679, opened 2026-08-31
fix(2673): a loader input naming a file the server does not have says where it looked, and how to fix it

### LINKED ISSUE #2673: Attached image filename is not available to headless workflow LoadImage
A user attached an image shown as input/8058752600640.JPG in the sidebar. A headless ComfyUI workflow was enqueued with LoadImage image='8058752600640.JPG', but ComfyUI rejected it: 'Invalid image file: 8058752600640.JPG'. The prior attached image 8058752601213.JPG worked with the same pattern. This may indicate attachment registration/race or filename handling in comfyui-mcp. Environment: Windows 11, NVIDIA RTX 5090 32GB, ComfyUI 0.34.0, frontend 1.49.6, comfyui-mcp 0.52.146, panel 0.15.149. Exact error: WORKFLOW_EXECUTION_ERROR; Prompt outputs failed validation; LoadImage node 5: Custom validation failed ... Invalid image file.

---
Filed from the ComfyUI panel by **chatgpt** · model `gpt-5.6-luna` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Closes #2673.

## What the reporter hit

A user attached an image in the panel; the sidebar showed `input/8058752600640.JPG`; a
headless workflow with `LoadImage image="8058752600640.JPG"` came back:

```
ComfyUI rejected the workflow (400): Prompt outputs failed validation
- LoadImage (node 5): Custom validation failed for node (image - Invalid image file: 8058752600640.JPG)
```

That is the **entire** message `buildEnqueueError` produced, and it is a dead end: it does
not say which server was asked, whether a connected panel is on a different one, or what
call would put the file where the render will look. The reporting agent guessed —
"attachment registration/race, or filename handling in comfyui-mcp" — and filed a P2.

## What is actually going on (checked, not assumed)

Read off the reporter's own ComfyUI **v0.34.0** checkout, not from memory:

- `LoadImage.VALIDATE_INPUTS(image)` (`nodes.py:1807`) returns `"Invalid image file: {}"` when
  `folder_paths.exists_annotated_filepath(image)` is false, which is a plain
  `os.path.exists(get_input_directory()/name)`.
- `/upload/image` (`server.py:397`) writes into `get_input_directory()` and returns the name it
  actually wrote — including its dedup rename.

Both use the same `get_input_directory()`, so **on one server they cannot disagree**: a 200
from `/upload/image` naming `N` means `input/N` exists, and validation checks that same path.
The reporter's failure therefore requires the upload and the enqueue to have reached
*different* input directories.

comfyui-mcp makes that structurally possible **by design**, and says so in
`src/comfyui/fetch.ts`: a chat attachment is uploaded by the **browser**, through whichever
ComfyUI the panel tab is on, while `enqueue_workflow` posts to `COMFYUI_URL`. #952 built
`describeTargetDrift` to detect exactly that split — but wired it only into **transport**
failures. The one error the split reliably produces is a 400, where the detector never ran.

Also checked, so the reporter's two hypotheses can be answered directly:
`enqueue_workflow action:"enqueue"` passes the workflow through verbatim
(`src/tools/workflow-execute.ts`) and nothing between the panel's `sendText` and `/prompt`
rewrites a loader value. It is not filename handling.

## The change

When a `/prompt` 400 names a **loader media input**, append a note to the message ComfyUI's
own diagnosis already produced (appended, never substituted):

> That input names a FILE on the server, not a free value: `LoadImage.image` (node 5) is
> resolved inside the input directory of the ComfyUI at http://127.0.0.1:8188 — so this is a
> question about WHICH server holds the file, not about the value's spelling. A file ATTACHED
> in the panel's chat is uploaded by the BROWSER to whichever ComfyUI that tab is on, a
> separate connection from this headless target (COMFYUI_URL), so a file can exist on one and
> not the other. **A connected panel is on http://127.0.0.1:8199 — a DIFFERENT ComfyUI from
> this target, so that is very likely where the file is, and very likely the whole answer: run
> the graph on the panel instead (`panel_run`), or point COMFYUI_URL at that origin.** To put
> it on THIS target: `upload_image (action:"image")` for a file on disk, or
> `upload_image (action:"stage")` for an existing ComfyUI output; then re-enqueue with the
> filename it returns. Re-submitting this workflow unchanged will fail identically —
> `enqueue_workflow` sends loader values through verbatim and never rewrites them.

The bold sentence is the machine-checked half. On a SAME-origin panel it instead says the
split is **RULED OUT**, which is worth as much: it stops the reader hunting for a second
server. With no panel connected it says nothing about origins at all — an absent comparison
is not a negative result.

## Where the load-bearing claims are — please push on these

1. **The detection predicate.** Keyed on `(class_type, extra_info.input_name)` from an
   allowlist, plus ComfyUI's machine `type` token (`custom_validation_failed` /
   `value_not_in_list`). Deliberately **not** on the prose `message`/`details`, and
   deliberately **not** on `extra_info.input_config`'s upload flag — I checked
   `execution.py:1101`, and `custom_validation_failed` carries `{"input_name": x}` and nothing
   else, so that flag is simply absent on the error that matters here.
2. **Why both error types.** `validate_inputs` skips the combo-membership check for any input
   the node declares in its own `VALIDATE_INPUTS` (`execution.py:1018`). `LoadImage.image`
   does, so it can only fail as `custom_validation_failed`; a loader that does not keeps the
   check and fails as `value_not_in_list`. Matching one would drop a whole class of loader.
3. **Over-firing.** The allowlist is the only thing stopping an unrelated
   `custom_validation_failed` (a bad width, an out-of-range strength) from collecting an
   "upload the file" note. Four negative tests pin it; an unlisted third-party loader gets
   silence rather than a guess.
4. **`LOADER_ASSET_INPUTS` moved** out of `services/workflow-converter.ts` into a new leaf
   `comfyui/loader-asset-inputs.ts` so both readers share one list. Pure move; the #504
   converter tests still pass.
5. **Redaction.** Every field the note interpolates comes from the response body and goes
   through the same `safeField` / #1191 scrub as the lines above it. The target is rendered via
   `originOf`, so a `COMFYUI_URL` carrying a token in its path or userinfo cannot leak.
6. **`classifyTargetDrift` now also returns `origins`/`alias`.** The existing
   `describeTargetDrift` text is byte-identical — its transport-failure wording ("the browser
   can reach it and this process cannot") would be *false* here, which is why the two share the
   comparison and not the sentences. `loopback-alias-origin.test.ts` (#1175) still passes, and
   the new suite re-pins alias handling through the new caller.

## Verification

- `npx vitest run` in this worktree: **745 files, 13694 passed** (6 skipped).
- `npm test` (the repo's own gate: vitest, i18n, docs, blog, vocabulary, unknown-collapse,
  vocab:export, changelog, asset-counts): all green after `npm run build`.
- **Mutation battery** — each gate deleted alone, each turns a named test red:

  | mutation | result |
  |---|---|
  | note never appended | 10 failed / 5 passed |
  | drift comparison removed from the note | 3 failed |
  | loader allowlist gate removed | 3 failed (all three negative tests) |
  | error-`type` gate removed | 1 failed (`required_input_missing` leaks a note) |
  | `value_not_in_list` dropped from the set | 1 failed (`VHS_LoadVideo`) |
  | `custom_validation_failed` dropped from the set | 9 failed |
  | same-origin verdict silenced | 2 failed |

- The fixture is the shape ComfyUI 0.34.0 actually emits, transcribed from `execution.py`,
  not invented — including `extra_info` carrying only `input_name`.
- Production reach: `buildEnqueueError` is the sole `/prompt` non-OK path in `enqueuePrompt`,
  which `services/workflow-executor.ts` and `services/queue-manager.ts` call for
  `enqueue_workflow`, `generate_image`, and every other headless render.

## What this deliberately does NOT do

- **It does not stop the file going missing.** I could not reproduce the reporter's session
  (no ComfyUI was allocated for this run, and the input dirs on this box have long since
  rotated), so the note names the one cause it can *check* and refuses to assert a cause it
  cannot. If a future reporter hits this with the drift verdict reading "RULED OUT", that is a
  new and much sharper bug report than #2673 was.
- **No panel-repo change.** The panel's chat line still says `input/<name>` with no hint of
  which server that is; making it say so is a separate change in a separate repo. Not done here.
- **No browser verification.** Nothing in this diff renders in the sidebar — it changes a
  string produced by the headless MCP child — so the Playwright suite was not run and I am not
  implying coverage I do not have.
- **Cloud mode untouched.** `enqueuePrompt` returns early into `cloud-client.ts` before this
  path; `panel_run` has its own rejection formatter (`formatRunRejection`) and runs against the
  panel's own server, where this split cannot arise.

### DIFF
--- a/src/__tests__/comfyui/enqueue-missing-input-media.test.ts
+++ b/src/__tests__/comfyui/enqueue-missing-input-media.test.ts
@@ -173,6 +173,12 @@ describe("#2673: a loader input naming a file the server does not have", () => {
       // The transport-failure wording would be FALSE here: this process did
       // reach the server — it got a 400 back.
       expect(message).not.toContain("this process cannot");
+      // …and the verdict states its own BOUND (gate, round 3): the comparison
+      // sees the tabs open RIGHT NOW, while the upload happened in the past, so
+      // a tab that has since closed is invisible to it. An unbounded "RULED OUT"
+      // would be a claim about history made from a snapshot of the present.
+      expect(message).toContain("visible right now");
+      expect(message).toContain("closed or navigated away");
     });
 
     it("#1175 parity: an ALIASED loopback spelling is the same server, and says so", async () => {
--- a/src/__tests__/comfyui/enqueue-missing-input-media.test.ts
+++ b/src/__tests__/comfyui/enqueue-missing-input-media.test.ts
@@ -200,6 +200,28 @@ describe("#2673: a loader input naming a file the server does not have", () => {
       expect(message).toContain("Rule that out before treating the file as simply missing");
     });
 
+    it("drops an OPAQUE origin, which serialises to the truthy string \"null\"", async () => {
+      // Gate, round 4. `new URL("file:///tmp").origin` is the STRING "null" —
+      // truthy, so it would otherwise be reported as a panel on a different
+      // ComfyUI called "null". It is not an address.
+      setConnectedPanelOrigins(() => ["file:///tmp"]);
+      const message = await enqueueAgainst(MISSING_ATTACHMENT);
+      expect(message).not.toContain("null");
+      expect(message).not.toContain("A connected panel is on");
+      expect(message).toContain("names a FILE on the server");
+    });
+
+    it("keeps a cloudflared-style hostname VERBATIM — it is an address, not a credential", async () => {
+      // The reason origins are normalised rather than scrubbed: a quick-tunnel
+      // hostname is a 32-char hex label, which #1191's opaque-run pass would
+      // redact — deleting exactly the answer this message exists to give.
+      const tunnel = "https://abcdef0123456789abcdef0123456789.trycloudflare.com";
+      setConnectedPanelOrigins(() => [tunnel]);
+      const message = await enqueueAgainst(MISSING_ATTACHMENT);
+      expect(message).toContain(tunnel);
+      expect(message).toContain("a DIFFERENT ComfyUI from this target");
+    });
+
     it("drops an origin that is not a well-formed scheme://host[:port]", async () => {
       // Gate, round 3. Only a normalised origin may reach the message; anything
       // else is dropped rather than echoed. With nothing left, there is no
@@ -336,6 +358,30 @@ describe("#2673: a loader input naming a file the server does not have", () => {
       expect(message).toContain("names a FILE on the server");
       // The claim is SCOPED, so it is not false even here.
       expect(message).toContain("on stock ComfyUI");
+    });
+
+    it("a value_not_in_list whose received_value was never a filename", async () => {
+      // Gate, round 4. `value_not_in_list` reports what it received. A non-string
+      // means the widget never held a filename — a malformed prompt, an object,
+      // a link tuple — and reading that as "the server is missing this file"
+      // would misdiagnose a real rejection.
+      const message = await enqueueAgainst(
+        rejection({
+          "12": {
+            class_type: "VHS_LoadVideo",
+            errors: [
+              {
+                type: "value_not_in_list",
+                message: "Value not in list",
+                details: "video: {'path': 'clip.mp4'} not in (list of length 41)",
+                extra_info: { input_name: "video", received_value: { path: "clip.mp4" } },
+              },
+            ],
+          },
+        }),
+      );
+      expect(message).toContain("Value not in list");
+      expect(message).not.toContain("names a FILE on the server");
     });
 
     it("a node error with no machine `type` at all", async () => {
--- a/src/__tests__/comfyui/enqueue-missing-input-media.test.ts
+++ b/src/__tests__/comfyui/enqueue-missing-input-media.test.ts
@@ -55,10 +55,24 @@ function res(status: number, body: unknown, statusText = "Bad Request"): Respons
   });
 }
 
+/** One node error exactly as ComfyUI 0.34.0 emits it (execution.py). */
+interface NodeErrorEntry {
+  type?: string;
+  message: string;
+  details: string;
+  extra_info: { input_name?: string; input_config?: unknown; received_value?: unknown };
+}
+
 /** The 400 envelope ComfyUI 0.34.0 answers a failed prompt validation with. */
-function rejection(
-  nodeErrors: Record<string, { class_type?: string; errors?: unknown[] }>,
-): unknown {
+interface RejectionBody {
+  error: { type: string; message: string; details: string; extra_info: Record<string, never> };
+  node_errors: Record<
+    string,
+    { class_type: string; errors: NodeErrorEntry[]; dependent_outputs?: string[] }
+  >;
+}
+
+function rejection(nodeErrors: RejectionBody["node_errors"]): RejectionBody {
   return {
     error: {
       type: "prompt_outputs_failed_validation",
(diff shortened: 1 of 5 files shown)


artokun:
## Independent review — 5 rounds, final verdict PASS on `963b6f7`

The `codex-gate.mjs` wrapper hit its documented Windows sandbox failure (`CreateProcessAsUserW failed: 5`) after round 1, so rounds 2-5 ran through `codex exec -s read-only` directly against the diff. Verdict on the SHA that will merge: **VERDICT: PASS**.

**11 findings raised, 9 fixed, 2 declined with evidence.**

Fixed:
1. The note named the pre-request URL; `fetch` follows redirects, so it now names the responder (`res.url || requestedUrl`).
2. One matching tab did not rule the split out when other tabs were open elsewhere — that case now names them instead of saying RULED OUT.
3. `LoadImageOutput.image` resolves in the OUTPUT directory, so "resolved inside the input directory" was false for it — reworded to "the media directories of the ComfyUI at X", which is also correct for an `[output]`/`[temp]`-annotated `LoadImage` value.
4. The claim is scoped to "on stock ComfyUI", so a pack shadowing a core class_type cannot make it false.
5. `originOf(target) ?? target` interpolated an unparsable target raw — now falls back to `safeField`.
6. The scrub test was vacuous (benign values, asserted present). It now uses a credential-shaped node id and asserts it does NOT appear.
7. Panel origins are normalised through `originOf` before printing, so only a well-formed `scheme://host[:port]` can reach a message.
8. RULED OUT now states its own bound: it sees the tabs open right now, not the tab the upload happened in.
9. `new URL("file:///tmp").origin` is the truthy string `"null"`; an opaque origin is no longer treated as an address.
10. A `value_not_in_list` whose `received_value` is present and not a string no longer collects the note — the widget never held a filename. Absence still qualifies, because `custom_validation_failed` carries no `received_value` at all (mutation-tested in both directions).

Declined, with the reasoning now written at the site:

- **Scrubbing panel origins.** `scrubSecretShapedText` pass 3 redacts any unbroken >=24-char credential-alphabet run, and a cloudflared quick-tunnel hostname IS one (`http://abcdef0123456789abcdef0123456789.trycloudflare.com` — a 32-char label), which is the most common remote-ComfyUI setup here. Scrubbing would redact exactly the origins this message exists to name. A hostname is public routing information; the credential-bearing parts of a URL (userinfo, path, query) are all dropped by `originOf`. Protection is normalisation, not redaction.
- **A pack shadowing a core class_type, or a future core check on an allowlisted media input, failing for a non-file reason.** Undetectable: ComfyUI 0.34.0 gives `custom_validation_failed` an `extra_info` of `{input_name}` and nothing else (`execution.py:1101`) — no `received_value`, no `input_config`, no upload flag. There is no signal to separate the cases. The wording is scoped for that reason, and `enqueue-missing-input-media.test.ts` now pins the residual as an explicit KNOWN LIMIT so narrowing it later is a deliberate change.

**Mutation battery, final state** — 14 mutations, each deleting one gate alone; every one turns a named test red, including both directions of the `received_value` gate (removing it, and making it strict enough to kill the reporter's own case).

Local: `npx vitest run` 745 files / 13704 passed, and `npm test` (the repo's own 14-check gate) all green.

PR #2690, opened 2026-09-01
fix(2688): a promoted write refused by the ownership envelope names the invariant that failed

### LINKED ISSUE #2688: panel_set_widget repeatedly refuses promoted COMBO because graph_get_subgraph ownership envelope is invalid
`panel_set_widget` cannot update a promoted COMBO widget on a live subgraph wrapper. The receiver is bound and stable, `panel_graph_outline` reads the wrapper correctly, and `panel_enter_subgraph` succeeds, but the promoted write is refused because `graph_get_subgraph` produces an ownership envelope that fails `validatePromotedSubgraphEnvelope`. Re-opening/rebinding the same workflow does not clear it.

### Steps to reproduce
1. Open a workflow containing a live subgraph wrapper with promoted COMBO `ckpt_name`.
2. Confirm the wrapper and value with `panel_graph_outline`.
3. Call `panel_set_widget({node_id:6066, widget:"ckpt_name", value:"ltx-2.3-22b-dev-nvfp4.safetensors"})`.
4. Re-open/bind the workflow and retry; it fails identically.

### Exact error (scrubbed)
`Error: panel_set_widget refused the promoted "ckpt_name" write because graph_get_subgraph returned a malformed, stale, or incomplete ownership envelope. No graph_set_widget was dispatched; retry only after the panel binding and subgraph mapping are stable.`

The retry after a confirmed `panel_open_workflow` binding produced the same error. `panel_enter_subgraph({node_id:6066})` succeeds and shows the inner promoted loader widgets.

### Fix
A safe local relaxation of the ownership fence was not attempted because it could authorize writes to the wrong promoted terminal. Please make `graph_get_subgraph` publish a complete envelope for this live subgraph (matching `subgraph_of.node_id`, `subgraph_of.graph_identity`, viewing identity, complete `node_count`/`nodes`, and terminal witnesses), and enhance the refusal to name the exact failed invariant so receiver-side drift can be diagnosed without exposing the full graph.

### Environment
Windows 11; ComfyUI 0.34.0; ComfyUI frontend 1.51.9; RTX 4090 24 GB; comfyui-mcp 0.52.165; panel 0.15.149.

artokun:
Closes #2688.

## The defect

`panel_set_widget` refused a promoted COMBO write with:

> graph_get_subgraph returned a malformed, stale, or incomplete ownership envelope. No graph_set_widget was dispatched; retry only after the panel binding and subgraph mapping are stable.

`validatePromotedSubgraphEnvelope` has **~20 distinct `return null` exits** — asserted truncation, an unparseable `viewing`, an owner id that does not name the addressed node, a `node_count`/`nodes` disagreement, an inner node without a usable id, and a dozen promoted-terminal witness shapes. All of them collapsed into that one sentence.

Both halves of it failed the reporter:

1. **It named nothing**, so there was no way to tell whether the cause was panel-side drift or an over-strict invariant on our side. They filed the issue asking us to "make `graph_get_subgraph` publish a complete envelope" without being able to say *which field was missing* — because nothing told them.
2. **Its remedy is wrong.** Every invariant is decided from ONE reply in ONE snapshot; none of them is a binding state that settles. They retried, re-opened the workflow, and re-bound the tab, and got the identical text every time, on the same node. There was no exit from that loop.

Same defect class as #2374 / panel#1869, which relayed the panel's own divergence diagnosis instead of overwriting it with this file's retry advice.

## The change

`describePromotedSubgraphEnvelope` is the same decision, returning the reason it rejected on. `validatePromotedSubgraphEnvelope` is now **that function with the reason dropped** — one decision in one place, so the diagnostic cannot drift from the authorization. All four refusal sites route through it (initial classification, the pre-write re-read, the nested-terminal read, and the post-refusal recovery path); the last three had the identical unnamed wording.

**The fence does not move.** No invariant was added, removed, loosened or reordered.

**The reason text is shape-only** — field paths, array indices, counts, and the two node ids that were compared (one of which is the caller's own argument). No widget name, no widget value, no node type, no title. Naming the invariant must not become a way to read a graph the fence is refusing to write to. That is asserted, not just intended.

- **`promoted-widget.ts` — the fence-equivalence claim.** Everything rests on `validatePromotedSubgraphEnvelope` still accepting and rejecting exactly what it did. It is now a two-line wrapper, so the risk is not in the wrapper but in whether the *reordered* checks inside `describePromotedSubgraphEnvelope` are the same predicate. The `node_count`/`nodes` block in particular was one combined `if` and is now three, and `parsePromotedTerminalEntries` was split into per-condition returns. Evidence below.
- **`panel-tools.ts:8823` — a behaviour change beyond wording.** The pre-write re-read used to fold an envelope failure into `"the promoted inner mapping changed or became unverifiable before the write"`, together with five other conditions. It now returns early and names the invariant. Still fail-closed, but a shape defect there no longer reports as a canvas edit the user did not make.
- **The two corrected tests.** They asserted the retry wording on the stated premise that *"a malformed envelope is a genuine transient … the one a retry can clear."* That premise is what this issue disproves. Both keep the panel#1859 claim they exist to make (a current build must not receive the build-skew message) and drop the false one.

## Evidence

**Differential equivalence vs. the pre-change fence.** A throwaway harness vendored `origin/main`'s validator and compared it against the new one over a generated corpus — every field of a valid envelope deleted and poisoned with 28 values, plus legacy/edge shapes, across 9 owner ids:

```
cases=12753 accepted=1108 rejected=11645
mismatches=0 leaks=0
```

Zero disagreement on accept/reject, zero disagreement on the returned envelope value, no input mutation, and no graph content in any reason string.

**Controls — the harness detects a fence movement in both directions:**

| mutation | result |
|---|---|
| relax the `truncated` check (accept more) | 50 mismatches |
| require `viewing` (accept less) | 6 mismatches |
| leak the offending array into a reason | 184 leaks |

**Mutation-testing the committed tests** (fix committed first, then reverted piecewise):

| mutation | tests red |
|---|---|
| primary call site back to `promotedWriteRefusal` + the generic sentence | **15** |
| every invariant collapsed to one generic reason | **31** |
| the fence made to disagree with the description (accepts a truncated read) | **2** (incl. the parity test) |
| graph content leaked into two invariants | **3** (incl. both privacy assertions) |

The parity test also asserts `accepted > 0`, so a corpus that rejected everything cannot pass it vacuously.

## Verification

- `npx vitest run` — **746 files, 13787 passed**, 6 skipped, 0 failed.
- `tsc --noEmit` clean. Note `tsconfig` excludes the test tree, so the test changes were validated by vitest, not by tsc.
- `check-anti-slop.mjs` — `785 known finding(s) … none added` (oxlint 1.79.0 installed into the worktree first; it is absent by default and the gate exits non-zero without it, which reads as a pass if you only check that the command ran).
- `check-tool-vocabulary.mts` OK, `check-unknown-collapse.mjs` OK.
- **No Playwright run, and none is claimed.** This change is orchestrator-side only: it alters a refusal string returned by the `panel_set_widget` MCP tool. Nothing the sidebar renders is touched — no file under the panel repo is in this diff.

The reporter asked for two things. This is the second one.

The first — *"make `graph_get_subgraph` publish a complete envelope for this live subgraph"* — is a **panel** change, and I could not identify which field their receiver failed on. Their panel (0.15.149) is current with panel `origin/main`, `panel_enter_subgraph` succeeds on the same node, and `panel_graph_outline` reads the wrapper correctly, so the envelope is being built by code that mostly works. Guessing at a producer fix without knowing the failing invariant risks a change premised on a defect that does not exist.

That is precisely what this PR unblocks: their next reproduction reports the field. I have not relaxed anything to compensate for not knowing, and I did not attempt the "safe local relaxation of the ownership fence" they correctly declined to make themselves.

One candidate worth naming for whoever picks up the panel half: `promotedTerminalWitnesses` builds error entries as `{ widget, immediate_node_id: resolution.target.node?.id, immediate_widget: …, error }`, and the consumer rejects the **entire** array if any single entry publishes neither a terminal endpoint nor an `error`, or publishes a terminal without an authoritative `parent_rail`. So one unrelated promoted alias can refuse every promoted write on the wrapper — which matches the reported symptom (deterministic, unaffected by rebinding, while the widget itself looks fine). Unconfirmed against their graph; the named invariant will confirm or kill it.

### DIFF
--- a/src/__tests__/orchestrator/promoted-widget.test.ts
+++ b/src/__tests__/orchestrator/promoted-widget.test.ts
@@ -3876,6 +3876,36 @@ describe("panel_set_widget promoted-subgraph recovery (#1655)", () => {
     expect(calls.map((c) => c.cmd)).not.toContain("graph_enter_subgraph");
   });
 
+  // Review finding (P2). The recovery path runs AFTER an outer graph_set_widget
+  // has already been dispatched and refused; only the INNER retry is skipped.
+  // Routing it through the standard refusal made it assert "No graph_set_widget
+  // was dispatched", which is false there and invites an agent to re-issue a
+  // write it has already made. The invariant is still named; the consequence
+  // clause is the one this path can actually stand behind.
+  it("names the invariant on the recovery path without claiming no write was dispatched", async () => {
+    const { text, isError, calls } = await setWidget(
+      { node_id: 78, widget: "width", value: 1024 },
+      {
+        firstWrite: "contradict",
+        firstWriteError: CONTRADICTORY,
+        preflightSubgraph: DEFINITIVE_NON_PROMOTED_SUBGRAPH,
+        // Read 2 must also classify non-promoted, so the recovery preflight
+        // returns null and the LEGACY recovery re-read below is the one that
+        // meets the malformed envelope. That is the branch under test.
+        recoveryPreflightSubgraph: DEFINITIVE_NON_PROMOTED_SUBGRAPH,
+        subgraph: { ...SUBGRAPH, node_count: 5 },
+      },
+    );
+
+    expect(isError).toBe(true);
+    expect(text).toMatch(/`node_count` claimed 5 inner node\(s\) but `nodes` carried 2/);
+    expect(text).toMatch(/The inner write was not retried\./);
+    expect(text).not.toMatch(/No graph_set_widget was dispatched/);
+    // The outer attempt is what produced the refusal being annotated, so it did
+    // happen — the message must not deny it.
+    expect(calls.filter((c) => c.cmd === "graph_set_widget").length).toBeGreaterThan(0);
+  });
+
   it("a truncated subgraph read is not treated as unique", async () => {
     const { text, isError, calls } = await setWidget(
       { node_id: 78, widget: "width", value: 1024 },
@@ -3883,7 +3913,7 @@ describe("panel_set_widget promoted-subgraph recovery (#1655)", () => {
     );
     expect(isError).toBe(true);
     expect(text).toMatch(
-      /the reply set `truncated`, so the inner node list it carried is not the whole subgraph/,
+      /the reply's `truncated` flag was not `false`, so the inner node list it carried cannot be taken as the whole subgraph/,
     );
     expect(calls.map((c) => c.cmd)).not.toContain("graph_enter_subgraph");
   });
@@ -4333,7 +4363,9 @@ describe("#2393 promoted-terminal witness is judged on the requested alias", () 
       /`promoted_terminals\[1\]` published neither a terminal endpoint \(`terminal_node_id`\/`terminal_node_type`\) nor an `error`/,
     );
     expect(text).toMatch(/accepted or rejected as a WHOLE/);
-    expect(text).toMatch(/panel_open_workflow and panel_set_workflow_target do not clear it/);
+    expect(text).toMatch(
+      /unless the panel's reply itself changes shape, panel_open_workflow and panel_set_workflow_target re-bind the tab and produce the identical refusal/,
+    );
     expect(text).not.toMatch(/retry only after the panel binding and subgraph mapping are stable/);
     expect(calls.filter((c) => c.cmd === "graph_set_widget")).toHaveLength(0);
   });
@@ -4393,7 +4425,20 @@ describe("#2688 describePromotedSubgraphEnvelope names the failed invariant", ()
 
   it.each([
     ["a non-object reply", "nope", 78, /the reply was not a JSON object/],
-    ["asserted truncation", { ...VALID, truncated: true }, 78, /the reply set `truncated`/],
+    [
+      "a truncated flag",
+      { ...VALID, truncated: true },
+      78,
+      /the reply's `truncated` flag was not `false`/,
+    ],
+    // A malformed flag is not evidence of a SHORT read, so the reason must not
+    // claim one. Same rejection, different established fact.
+    [
+      "a malformed truncated flag",
+      { ...VALID, truncated: null },
+      78,
+      /the reply's `truncated` flag was not `false`/,
+    ],
     [
       "a missing subgraph_of",
       { ...VALID, subgraph_of: undefined },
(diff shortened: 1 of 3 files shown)


artokun:
## Review gate

Codex (, direct — the `codex-gate.mjs` wrapper returned `CreateProcessAsUserW failed: 5 (Access is denied.)` with no code verdict, which is INDETERMINATE, not a pass).

**Round 1** on `b801ef9` — `VERDICT: NEEDS-REVISION`, four findings, all real:

| sev | finding | disposition |
|---|---|---|
| P0 | the new loops walked `.entries()`, an own-property an untrusted reply can shadow; the loops they replaced walked `Symbol.iterator`. A shadowing payload would be validated against a different list than `nodes.length !== node_count` measured. | fixed — indexed loops |
| P2 | the recovery path asserted "No graph_set_widget was dispatched" when an outer write **had** been dispatched and refused; only the inner retry is skipped. I introduced that. | fixed — "The inner write was not retried." |
| P1 | "panel_open_workflow and panel_set_workflow_target do not clear it" was a universal I had checked only for the shapes I reasoned about | fixed — scoped to "unless the panel's reply itself changes shape" |
| P2 | the truncation reason claimed the list *was* short for any non-`false` value; `truncated: null` establishes a malformed flag, not a short read | fixed — "the `truncated` flag was not `false`" |

On P1 I did check the mechanism codex proposed (a stale binding yielding `subgraph_of.node_id: 79` for addressed `78`) and it is **not** reachable through the panel: `resolveLiveNode` matches on `canonicalNodeId(node.id) === canonicalNodeId(nodeId)`, so a live panel cannot return a node whose id differs from the addressed one — a stale binding produces a *missing-node error*, which is a different path. The finding still stands as written, because the sentence was a universal and I had proved it for a subset. Scoping it costs nothing and removes the need for that argument.

Each of the four fixes is pinned: reverting any one turns exactly one named test red (verified individually).

**Round 2** on `5e68061` — the SHA that will merge, head parity confirmed before running:

> VERDICT: PASS — No findings.

Differential equivalence re-run after the round-1 fixes, with shadowed-`entries` and sparse-array cases added: **12789 cases, 1108 accepted, 0 mismatches, 0 content leaks.**

Full suite after the gate round: **746 files, 13790 passed**, 6 skipped, 0 failed. `npm run lint` (tsc + anti-slop) exit 0, `none added` — the round-1 test fix initially added 2 `no-chained-type-assertions` findings, rewritten as `Object.defineProperty` rather than baselined.

PR #2696, opened 2026-09-01
fix(2692): bound a BATCH of inline images against the transport frame, and bound action:"view" at all

### LINKED ISSUE #2692: Batch view_image results can exceed code-mode IPC frame limit
A code-mode call that awaited nine view_image results and forwarded all nine images failed before results could be delivered because the serialized IPC frame exceeded the hard 64 MiB limit. The orchestration path does not chunk or reject oversized aggregate image payloads before IPC serialization.

### Steps to reproduce
1. Call view_image in parallel for nine 3648x5472 JPEG/PNG diagnostic images.
2. Forward every returned image_url using image().
3. Observe the code-mode initial response failure.

### Exact error (scrubbed)
code-mode initial response exceeds the IPC frame limit: code-mode IPC frame length 108765829 exceeds 67108864 bytes

### Fix
Chunk image-bearing code-mode results or enforce an aggregate payload budget before IPC serialization, yielding multiple frames when needed. Temporary workaround: inspect images in smaller batches.

### Environment
Windows 11; ComfyUI 0.34.0; RTX 5090 32 GB; comfyui-mcp 0.52.146; panel 0.15.102. Frontend version not queried because this is an orchestrator IPC defect, not a frontend rendering defect.

artokun:
Closes #2692.

## What broke

A code-mode script awaited nine `get_image action:"view"` results for 3648×5472 assets and
forwarded all nine. The serialized IPC frame came to **108,765,829 bytes against a hard
67,108,864-byte limit** and the whole response was lost — including the eight images that
were not the problem.

Two separate defects had to line up, and **fixing either one alone leaves the report open**.

### 1. `action:"view"` had no inline bound at all

#1495 capped `action:"get"` (`boundInlineImage`, ~16 MB of base64). `action:"view"` — the
action the report names, and the one the retired per-tool viewer folded into — returned
`viewAssetImage()`'s content blocks verbatim. A single 8504×17008 PNG viewed through it
still inlines ~267 MB, which is exactly the image #1495 was filed about.

This is the "fix one exit, leave its siblings" shape. I surveyed the rest of the family and
did **not** change two of them, deliberately:

- `get_image action:"convert"` — unbounded, but the inline payload **is** the product the
  caller asked for, and with no `out_path` it is the only copy. Bounding it would destroy
  the deliverable rather than protect it.
- `train_prepare_dataset action:"file"` — unbounded, same viewing semantics, but it reads a
  local curated dataset image and is not on the reported path. Named here so the next reader
  does not have to rediscover it.

### 2. A per-image budget cannot bound a frame that carries N images

**Measured, not reasoned.** On a synthetic source of the reported 3648×5472, *with* the
#1495 bound applied, each asset lands at **~11.09 MB of base64** and nine come to
**~99.8 MB** — still half again over the limit. So (1) on its own would have shipped a fix
that does not fix the report.

## The batch ceiling

We do not own the frame. Chunking it, or yielding several frames, is
`codex-code-mode-host`'s job — the limit and the error wording both come out of that binary.
What this server owns is every byte that goes *into* it, so the ceiling is enforced at
emission: `src/services/inline-frame-budget.ts` gives a call a share of a 48 MB aggregate
(75% of the frame; the rest is text blocks and JSON-RPC envelope we cannot measure).

**The signal is concurrency, not a clock.** A time window would have to guess which calls
share a frame and would shrink previews for a client that merely fetched several images in a
row. The members of a code-mode batch are *in flight together*, and that is directly
observable from inside a handler.

I measured that before building on it, against a **real spawned stdio MCP server** rather
than the in-memory pump: nine parallel `tools/call` requests for a 150 ms handler completed
in **166 ms, not 1350 ms**, and every one of the nine handlers observed nine slots open.

**A lone call at the default budget is byte-for-byte unchanged** — it sees a peak of one, so
its share is the whole 48 MB aggregate, three times the per-image budget, and `share()`
returns the request untouched. That is asserted directly, and there is a sequential control
in the batch test running the same images through the same handler.

## Where the load-bearing risk is

Three places a reviewer should push on:

1. **The eager/lazy ordering.** `share()` MUST be read after the fetch. My first wiring read
   it at the `view` call site, before the await — every call then priced itself for the batch
   at its narrowest, and nine concurrent views came to 1,884,132 bytes against a
   900,000-byte aggregate. The test caught it; `ViewAssetImageOptions.budgetBytes` is now a
   thunk so the type states the requirement. `action:"get"` already resolved after its fetch.
2. **Over-refusal.** The design's whole safety argument is that a non-batched client is never
   charged. The one lone call that *is* lowered is a caller explicitly asking for more than
   the aggregate — a request the transport could never have delivered — and that case emits
   its own `INLINE LIMIT` notice rather than shrinking silently.
3. **A residual over-charge I did not fix.** A slot is taken before the fetch, because the
   mime type is not known until it returns — so a concurrent batch mixing non-image
   `action:"get"` calls (video/audio saves) with image fetches charges the batch for the
   non-image members too, and the images get a smaller — but disclosed — preview. I wrote an
   early release for it and removed it again: **its mutation survived.** The mime type is
   only known after the save, and the handler's `finally` runs a few statements later, so it
   could not buy any time. Shipping it would have been a dormant mechanism with a confident
   comment about a benefit it does not deliver.
4. **Slot leaks.** A leaked slot fails silently: it narrows every later preview in the
   process for a batch that finished long ago. Released in a `finally` on the handler's
   existing try, `release()` closes over its own entry so it is idempotent and cannot free a
   stranger's, and there are tests for a throwing fetch and for `openInlineImageSlots() === 0`.

## Verification

`npm run lint` (tsc + anti-slop, none added), `npm run build`, `npm run check:vocabulary`,
`npm run check:unknown-collapse`, `npm run vocab:export -- --check` all green. Full suite
green. **Mutation-tested — every call site independently, and a bad anchor aborts rather
than silently no-op'ing:**

| mutation | killed by |
| --- | --- |
| `boundInlineImage` removed from `view-image.ts` | 2 (the view-bound tests) |
| `share()` becomes the identity | 8 |
| the handler never acquires a slot | 2 |
| the `view` budget resolved eagerly | 1 (the concurrent-view test) |
| peak not raised on already-open slots | 7 |
| control: an unrelated string in the same file | only the 2 wording assertions |

**No browser verification, and none is claimed.** Nothing here touches panel-rendered
behaviour — the change is confined to the MCP tool surface (`get_image`) and two services.
`panel_screenshot`'s own inline path is untouched.


artokun:
## Held for one decision — not merging on my own authority

CI is green on the exact head (`f2512ef`, all four checks). The codex gate is **NO-SHIP**, and
after five rounds it is raising the same thing each time. I am not merging past it, and I am
not going to keep gating — the owner's standing instruction is one round and not to grind
past ~3, and I am at five.

### Gate history

| round | verdict | substance |
| --- | --- | --- |
| 1 | NO-SHIP | budget is per-call, not reserved for a whole frame — via a *staggered parallel* mechanism |
| 2 | NO-SHIP | "no code verdict is possible"; sandbox blocked. Indeterminate wearing a NO-SHIP label |
| 3 | SHIP | bare pass on the same source diff |
| 4 | NO-SHIP | same concern, now via the reachable mechanism: sequential awaits |
| 5 | NO-SHIP | same, restated |

Round 1 was **right and I acted on it** — it exposed a genuine overclaim in my own comment
(I had written that the calls landing in one frame are always in flight together; that is
true for a batch awaited together and false for a sequence). The mechanism it *named*,
though, is not reachable: for a parallel call to release before its sibling is admitted, the
transport would have to deliver the second request only after the first handler returned.
That is now measured in the repo rather than argued —
`src/__tests__/services/mcp-concurrent-admission.test.ts` spawns a real stdio server and
asserts every one of nine handlers sees `completedBefore === 0`.

Round 4 named the reachable variant, and **that one is real**: a script that awaits its
fetches one at a time and forwards them together fills the same frame, and every call sees a
peak of one.

### Why I did not turn the warning into a hard cap

Not an oversight — a trade I am not willing to make unilaterally. Three routes, all measured:

1. **Charge every call as though a sequence might follow.** Permanently lowers the
   single-image preview for every client, including the majority whose results are each
   their own frame. A certain regression traded against a possible one.
2. **A decaying cumulative cap.** Without a floor, the fourth image anyone looks at in a
   session refuses. With a floor of aggregate/16, nine sequential images still come to
   ~66 MB and overrun anyway. It fails in both directions.
3. **Cap only code-mode clients.** This would be the right answer if the server could
   identify one. It cannot: compact/facade registration (`list_tools`/`call_tool`) is not the
   same thing as code-mode — a code-mode host reads a full `tools/list` too — so keying
   safety behaviour on it would be a guess dressed up as a signal. `clientInfo` sniffing is a
   larger design decision than this PR.

By the fifth call the first four are serialized and gone regardless; nothing this process
does at call five can shrink them. **Reserving across a frame requires seeing the frame, and
only the code-mode host does** — which is what #2692's own Fix section asks for
("chunk image-bearing code-mode results... yielding multiple frames when needed") and what
the reporter marked `upstream-only: yes`.

So the sequential shape is **disclosed** instead: the running total is counted and reported
once it is within 60% of the frame, naming the non-obvious consequence (the whole reply is
lost, not the largest image) and the remedy. Nothing shrinks, so there is no over-refusal
risk, and it reaches the only actor that can prevent the overrun — the script deciding how
many results to forward, which cannot count what it has been handed.

### What this PR does close, and what it is worth on its own

`get_image action:"view"` had **no inline bound at all**. That is not a nuance — it is the
#1495 defect, live on a sibling exit, on the exact action this report names. A single
8504×17008 PNG viewed through it still inlines ~267 MB today. That fix stands regardless of
how the aggregate question is settled.

### What would unblock this

One decision, either way:

- **"Ship it"** — accept the documented sequential residual as upstream, and merge.
- **"Take trade 1 or 3"** — say which, and I will build it. Trade 3 needs a ruling on what
  this server is allowed to treat as evidence of a code-mode client.

Evidence for the work itself is in the PR body: eleven independent mutations, ten killed by
the named tests; the two that **survived** are both recorded — one was a dormant early-release
I deleted rather than shipped, the other an untested refusal guard that now has a test and
was charging ~267 MB for a response that inlined nothing.

artokun:
Not asking you to re-decide anything — checking one load-bearing claim in the trade-3 analysis, because it is stated as a capability limit and it is not quite one. The conclusion survives, but for a different reason, and the different reason is what makes the ruling easy.

## The claim

> **Cap only code-mode clients.** This would be the right answer if the server could identify one. **It cannot** … `clientInfo` sniffing is a larger design decision than this PR.

## What the server can actually see

`Server.getClientVersion()` and `Server.getClientCapabilities()` are both populated after `initialize`, and neither is used anywhere in `src/` today:

```ts
getClientCapabilities(): ClientCapabilities | undefined;   // experimental, sampling,
                                                           // elicitation, roots, tasks,
                                                           // extensions, _meta
getClientVersion(): Implementation | undefined;            // { name, version, title? }
```

So the server **can** identify the client — name, version, and declared capabilities, straight from the handshake, no sniffing and no new design. "It cannot" is too strong as written.

## Why trade 3 is still the wrong lever

The signal exists; it just does not answer the question. **Code-mode is a property of the turn, not of the client.** The same host, same `clientInfo`, same declared capabilities, runs some turns in code mode and some not — so a client allowlist keyed on `{name, version}` would cap sessions that never batch anything and miss code-mode turns from a host not on the list. That is a guess dressed up as a signal, exactly as the analysis says, and none of the seven capability flags carries "I will forward several tool results in one frame".

So the conclusion holds and is now load-bearing on something firmer than an availability limit: **the identity is readable and still does not determine the behaviour.** Trade 3 cannot be built correctly, rather than merely not being buildable here.

## Which leaves two, and one of them is already written

- **Ship it.** Accept the documented sequential residual as upstream — which is what #2692's own Fix section asks for ("chunk image-bearing code-mode results… yielding multiple frames") and what the reporter marked `upstream-only: yes`. The disclosure is already in the PR: the running total is counted and reported at 60% of the frame, naming the non-obvious consequence (the whole reply is lost, not the largest image).
- **Trade 1.** Charge every call as though a sequence might follow. A certain regression for every single-image caller, traded against a possible one.

The `get_image action:"view"` half is independent of all of this and stands on its own — that exit had **no** inline bound at all, and an 8504×17008 PNG still inlines ~267 MB through it today.

## What I did not do

I did not touch the PR. Five gate rounds is past where grinding pays, and the standing instruction is one round; this is a note on the decision, not another attempt at it.

If the answer is "ship it", I will merge and close. If it is trade 1, I will build it — and in that case the thing to decide alongside is whether the lowered single-image preview gets a documented override, because that is the regression it buys.

artokun:
Adversarial pass (mine — codex is out of tokens). The design holds up; it covers one of the two tools that put inline images in that frame.

**Verified sound.** The peak-raising is the right insight — `acquireInlineImageSlot` raises `peak` on *every* open slot, not just the newcomer, so a call that started alone stops spending a full per-image budget once siblings arrive. Release is idempotent and paired in a `finally`, so `open` cannot leak and permanently shrink later images. Both emission sites in `image-management.ts` (the `get` path at 506/522 and the `view` path at 610/624) take their share from the slot and charge what they emit. No gap inside that tool.

**The gap is `train.ts`.** Repo-wide, exactly two tools emit inline image content, and only one participates:

```
src/tools/image-management.ts   acquires a slot, charges every emission
src/tools/train.ts              neither — zero references to the budget module
```

`train.ts:285`:

```ts
case "file": {
  const { data, mimeType } = readTrainingFile(requirePath("file", …));
  return { content: [{ type: "image" as const, data, mimeType }] };
}
```

Its own description says "Bounded: image files only, ≤ 2MB" — a per-item bound, which is exactly the thing this PR establishes is not sufficient. It goes out through the same MCP transport into the same response frame.

Two consequences, both quiet:

1. **The running total under-counts.** `chargeInlineEmission` never sees those bytes, so `FRAME BUDGET:` fires late or not at all after a couple of training-file fetches — and the warning exists precisely because forwarding too much loses the *entire* reply.
2. **The peak under-counts.** A concurrent `get_image` sees `peak: 1` and spends the full per-call budget while a 2MB training image is in flight beside it, which is the batch case this PR exists to prevent.

The fix looks like two lines at that call site — acquire a slot around the `file` case and pass the emitted length through `chargeInlineEmission` — but it is your call whether it belongs in this PR or a follow-up. Either way worth stating, because "bound a BATCH of inline images against the transport frame" reads as covering the frame, and one tool is still outside it.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **Four probes, all anticipated in-source. Reporting it as verification rather than inventing a finding.**

## The race I expected to find is the one the design is built around

A shared budget across overlapping handlers is where these things break: two calls both read "1 slot open", both spend a full budget, and jointly blow the frame. That is explicitly handled —

```js
for (const slot of open) {
  if (open.size > slot.peak) slot.peak = open.size;
}
```

— raising the peak on **every** open slot, not just the newcomer, so a handler admitted while it was alone does not keep a full-budget share once siblings arrive. The comment says exactly why, and it matches the report's observed shape (nine slots held at once, admitted before any resolved).

## The arithmetic cannot overshoot

`share()` is `aggregate / max(1, peak)`, with **no cap on the divisor**. So N concurrent calls total `N × (48 MB / N) = 48 MB`, against a 64 MB frame, for any N. I went looking for a floor on `peak` that would break this — the header comment mentions an `aggregate/16` scenario — and there is none in the code (`MAX_PEAK`, `/ 16` and any other clamp: no hits). The only clamp is `Math.max(1, …)`, which exists to stop an absurd env value producing a zero budget that would read as "no limit".

## The sequential case is warned, not refused — correctly

`chargeInlineEmission` returns a sentence rather than blocking, which is right: each image individually fits, and only the agent's decision to forward them together breaks the frame. A tool cannot refuse on behalf of a choice it has not seen yet.

Two details there that I checked and would otherwise have flagged:

- **`QUIET_RESET_MS = 120_000`** — long enough that a slow sequential script cannot silently lose its running total between fetches. The comment states the asymmetry and picks the right side: over-counting costs a true-but-unnecessary sentence, under-counting drops the warning on the run that needed it.
- **`FRAME_WARN_BYTES = 60% of the frame`** — warns *before* the limit. A warning attached to the call that broke it "arrives after the bytes it was meant to prevent".

## On the sibling survey

The body names the "fix one exit, leave its siblings" shape and then does the survey, including two members deliberately left unchanged with reasons. That is the failure mode I have found on four other PRs this week, and it is the right way to close it: an explicit exemption is a decision on record, where silence is indistinguishable from an oversight.

Nothing blocking.

artokun:
**The third finding is real and fixed; the two budget ones I have not taken.**

**Fixed — the prescribed recovery fetched the wrong directory.** `get_image action:"get"` defaults to `args.type ?? "output"` (`image-management.ts:377,982`). The refusal handed back:

```
Use get_image (action:"get", filename:"…", subfolder:"…")
```

with no `type`. For an asset registered under `input/` or `temp/` that fetches `output/` — the wrong file, or nothing — and it fails precisely where it matters, because this path inlines and writes nothing locally, so that string is the only instruction the caller has. The refusal says so itself: *"Nothing was written locally either — this action only inlines."*

`fetchType` was already normalised a few lines above and simply unused. Both prescriptions now build from one helper so they cannot drift; `output` stays implicit as the default, anything else is stated. Mutation-verified.

That is the third instance of this shape I have fixed today, after panel#2044 (a token that cannot sign the browser in) and panel#2242 (a save that can clobber the change it warns about). A refusal that names a remedy is making a claim about that remedy, and it deserves the same scrutiny as the refusal itself.

**Not taken — the two budget findings.** Both are about concurrency, and both look right to me on reading:

- the grant is a one-time snapshot while `peak` can rise during async Sharp work, so a batch can exceed the advertised ceiling;
- the slot set is module-global while the HTTP transport builds an `McpServer` per session, so unrelated clients divide one another's budget and can see a `BATCH LIMIT` that cannot apply to them.

I am not fixing them here because both need a decision about where budget identity lives (per-frame? per-session? per-server?), and getting that wrong trades a visible over-budget for an invisible under-budget. Worth doing deliberately rather than as a drive-by on a PR that is already green.

**Suite note.** The broader `src/__tests__/services` run shows four unrelated failures, all timeouts under concurrency. `sharp-loader` passes 26/26 in isolation. `panel-image-relay` fails in isolation too with `expected 0 to be greater than 0` — that is the #2823 flake exactly, on a branch that does not carry #2823's fix, which incidentally reproduces that PR's premise. My diff touches only `view-image.ts` and its test.

PR #2708, opened 2026-09-01
fix(2704): prove a VRAM release landed before reporting the reading as settled

### LINKED ISSUE #2704: clear_vram/health VRAM-free report can stay stale past the 5s settle cap on some GPUs (RTX 5090 + cudaMallocAsync), understating free VRAM

`clear_vram` and `get_system_stats(action:"health")` reported VRAM free as ~2.7-2.9GB / 2805-2965 MB on a connected local ComfyUI (RTX 5090, 34190458880 bytes total, cudaMallocAsync), right after a `clear_vram` call that unloaded all cached models. A `get_system_stats(action:"stats")` call made a short time afterward (a few tool calls / seconds later, 0 running/pending queue the whole time) returned `vram_free: 32268615680` (~30GB) from the same `/system_stats` endpoint. A subsequent Qwen Image Edit 2511 (bf16, ~19GB VRAM) generation completed successfully at that point, confirming ~30GB really was free and the earlier low-VRAM report was stale, not the raw stats.

Root cause, found by reading `dist/tools/memory-management.js`: `clear_vram` already has a settle-loop (`readSettledSystemStats`, added for #2050) that polls `/system_stats` after calling `/free`, waiting for the `vram_free:torch_vram_free` signature to stabilize, specifically to avoid printing a stale VRAM figure. But it has a hard cap, `CLEAR_VRAM_SETTLE_TIMEOUT_MS = 5_000`, after which it gives up and returns whatever it last read even if not yet stable. On this GPU/driver/allocator combination (RTX 5090, cudaMallocAsync, ComfyUI 0.34.0, Windows), the actual CUDA/driver-level release of a ~29GB model unload took longer than 5 seconds to become visible in `/system_stats`, so the settle loop hit its cap and returned a falsely pessimistic reading. `get_system_stats(action:"health")` (in `dist/services/health-check.js`) reads `dev.vram_free` directly from the same stats call and is not itself miscalculating anything -- it just inherits whichever (possibly still-settling) snapshot it queries at that moment, and has no settle-loop of its own at all.

This is misleading for any pre-flight VRAM check before a large model load -- it looks like an imminent OOM when there is actually plenty of free VRAM a few seconds later.

### Steps to reproduce

1. On a system with a large amount of VRAM in use (multiple prior large model loads, e.g. several Qwen Image Edit 2511 bf16 generations in the same session).
2. Call `clear_vram` (unload_models + free_memory, defaults).
3. Note the printed "Current VRAM: X/Y MB free" figure.
4. Immediately call `get_system_stats(action:"stats")` and compare `devices[0].vram_free`.
5. On a GPU/allocator where CUDA release lags past the 5s settle cap, step 3's figure will read much lower than step 4's, even though no new allocation happened in between.

### Exact error (scrubbed)

No error/exception -- silently misleading data. Example captured values:
- `clear_vram` output: `VRAM cleared successfully (models unloaded, memory freed). Current VRAM: 2805/32607 MB free | Torch: 78/96 MB free`
- `get_system_stats(action:"stats")` moments later: `{"devices":[{"name":"cuda:0 NVIDIA GeForce RTX 5090 : cudaMallocAsync","vram_total":34190458880,"vram_free":32268615680,"torch_vram_total":100663296,"torch_vram_free":82575360}]}`

### Fix

applied locally: yes  upstream-only: no (also needs upstream release since this install is an npx-cached build with no local source checkout)

Minimal, reversible change: raised the hard cap so it has more room to observe the settle on slower-releasing GPUs/allocators, without changing behavior for GPUs that already settle quickly (the existing "two consecutive stable polls after `CLEAR_VRAM_SETTLE_MIN_MS`" early-exit is untouched, so fast cards still return immediately).

```diff
--- a/dist/tools/memory-management.js
+++ b/dist/tools/memory-management.js
@@
 /** Hard cap so a card that never plateaus still returns a number. */
-export const CLEAR_VRAM_SETTLE_TIMEOUT_MS = 5_000;
+export const CLEAR_VRAM_SETTLE_TIMEOUT_MS = 12_000;
```

(Only `dist/*.js` was patched -- this install has no `src/*.ts` checkout, so `npm run build` could not be run; the equivalent constant should be raised in the TypeScript source for the next release.)

**Verification status: NOT independently confirmed live.** The running comfyui-mcp server process had already loaded the old constant into memory before this file edit; a live on-disk edit does not hot-reload a running Node process, and a subsequent `clear_vram` call in the same session still showed the old ~5s-capped behavior (a stale-looking low reading), consistent with the process not having reloaded the file yet. A full respawn of the MCP server would be needed to actually load the patched constant, which was intentionally not done here to avoid disrupting the active session. Please verify after a restart, and consider whether 12s is the right number or whether the settle loop should instead poll until stable with a much longer cap (rather than a fixed short one) given how slow cudaMallocAsync release can be observed to be here.

### Environment

OS: win32 (Windows 11) | ComfyUI: 0.34.0 | ComfyUI frontend: 1.49.6 | GPU: NVIDIA GeForce RTX 5090, 34190458880 bytes VRAM total (cudaMallocAsync) | comfyui-mcp: 0.52.146 | comfyui-mcp-panel: 0.15.150

artokun:
Closes #2704.

## The report's premise does not hold — measured, not read

The report diagnosed a 5s settle cap being hit, and prescribed raising
`CLEAR_VRAM_SETTLE_TIMEOUT_MS` to 12s. I drove the real `settleUntilStable` from
`origin/main` against the reported RTX 5090 timing before changing it. **The cap
is never reached, and raising it alone changes nothing:**

```
Driver releases 8s after /free (the reported case):
  origin/main (cap  5s)    2805 MB after 788ms   *** STALE ***
  cap raised to 12s only   2805 MB after 773ms   *** STALE ***
```

The loop exits at ~780ms, nowhere near either cap.

**Why.** The stability signature is `vram_free:torch_vram_free`. Torch hands its
pool back in ~400ms; the *driver* counter stays frozen at its pre-`/free` value
for ~8s. That torch movement satisfied the loop's change-then-plateau test, so
the next two equal samples looked like a settled plateau and the frozen driver
number was returned as a measured result.

The deeper problem: **stillness cannot distinguish a plateau from a release that
has not started** — both are perfectly still. No cap value fixes that. A 3s lag,
comfortably inside the *existing* 5s cap, is equally broken on `origin/main`
(returns stale at 778ms).

## The fix

Compare against a sample taken **before** the mutation. That turns the
unanswerable "is this still?" into the answerable "has anything been released
yet?".

- `settleUntilStable` takes an optional `baseline`; a plateau counts only once
  the reading has moved off it. It returns `{ value, settled }` rather than a
  bare value.
- **`progressOf` is separate from `signatureOf` on purpose.** The release test
  reads the *driver* counters alone. Testing movement against the combined
  signature would let torch's fast release stand in as proof the driver
  released — the original bug wearing a baseline. I wrote that bug into my own
  first draft; two tests pin it now.
- **Baseline and progress are identity-keyed maps, latched per device.** A joined
  signature would let card 0 releasing certify card 1; positional matching would
  let a *reordered* device list certify a box where nothing released. A device
  absent from a sample stays unmoved — absence is not a release. Devices whose
  identities collide yield no baseline at all rather than a silently-merged one.
- The cap moves 5s → 12s. Load-bearing only in combination: it is what lets an 8s
  release land before the loop gives up.
- An unconfirmed reading is still reported — it is the best number available —
  but never as a measured figure. `clear_vram` appends a caveat; the panel adds
  `vram_after_settled: false`.
- The wait is armed only when a release was actually **requested** and the card
  was **occupied** at baseline (the same free/total rule `panel-tools.ts` already
  applies to these counters). An idle card, or a `/free` with both flags off,
  keeps today's fast path instead of blocking for the cap.

Measured after the change:

```
Driver releases 8s:        30774 MB after 8381ms   settled=true   CORRECT
Driver releases 3s:        30774 MB after 3412ms   settled=true   CORRECT
Driver releases instantly: 30774 MB after 1047ms   settled=true   CORRECT
Driver never releases:      2805 MB after 12015ms  settled=false  (disclosed)
```

## Costs, stated plainly

- **A fast card costs +273ms** (774ms → 1047ms). The old rule let *any* change
  return before `VRAM_SETTLE_MIN_MS`; that shortcut is exactly what the #2704
  card exploited, so the min wait now always applies.
- **An occupied card whose release never lands blocks for the full 12s** and then
  reports the reading with a caveat. A deliberate trade: the old failure silently
  understated free VRAM by ~27 GB, which reads as an imminent OOM. A slow honest
  answer beats a fast wrong one.

## Risky parts — please look here first

1. **`progressOf` vs `signatureOf`** (`vram-settle.ts`). If these are ever
   collapsed into one function, the bug returns silently and the numbers still
   look plausible.
2. **I changed an existing assertion.** `clear-vram-stale-report.test.ts` → "does
   not poll /system_stats when /free fails" asserted `getSystemStats` was never
   called. The baseline read now happens *before* `POST /free`, so it is called
   exactly once. I pinned the count at exactly 1 rather than loosening it to
   "some", so a settle loop leaking in behind a failed `/free` still fails the
   test. Its intent — no VRAM figure reported when the free failed — is unchanged
   and still asserted.
3. **The occupancy gate is a threshold**, reusing the panel's existing
   `free/total <= 0.2` rule rather than a new number. It decides latency, never
   correctness: getting it wrong costs a wait we did not need, or today's
   behaviour.
4. **The baseline read is a new call before a throw path.** Wrapped so a failed
   read degrades to `null` (and to `confirmable: false`) rather than failing the
   clear.

## Deliberately NOT changed

- **`get_system_stats(action:"health")` keeps no settle loop.** The report asks
  whether it should have one. It is a point-in-time read and it is not
  miscalculating — it reported what the driver said at that moment. Adding a
  multi-second settle to every health call would tax the common path to correct a
  transient that only follows a `/free`.
- **`annotateFreeVramAck` keeps the old best-effort rule.** No pre-`/free` sample
  can exist there — the tab posted `/free` before it acked — so there is nothing
  to hold a baseline against. Called out in a comment at the site.

## Verification

- **Independent review: codex, 5 rounds.** Rounds 1–4 each returned
  NEEDS-REVISION with a real finding (baseline-vs-idle conflation; global release
  latching; positional key matching; colliding device identities; the wait armed
  regardless of the requested `/free` flags). Each was fixed and pinned. Round 5
  returned **VERDICT: PASS** on `a1f7781`, the SHA this merges.
- **18 mutations, 18 caught.** Every element deleted independently — the cap, the
  baseline, `progressOf` (both call sites), the disclosure (both), the occupancy
  gate (both), the min-total guard, `confirmable` (both), identity matching,
  absence handling, the collision guards, `releaseRequested`, and the baseline
  read's *ordering*. The one mutation that survived first time (the occupancy
  gate) exposed a fixture that changed the driver value across `/free` and so
  satisfied the movement test by accident; it was rebuilt as a genuine static
  control and now catches it.
- Full suite: **752 files, 13863 passed, 6 skipped, 0 failed.**
- `npm run lint` (tsc --noEmit + anti-slop): clean, no new findings.
- `check:vocabulary` and `check:unknown-collapse`: pass.
- Not browser-verified: this changes orchestrator/tool output and renders nothing
  in the panel sidebar.

### DIFF
--- a/src/__tests__/orchestrator/free-vram-driver-lag.test.ts
+++ b/src/__tests__/orchestrator/free-vram-driver-lag.test.ts
@@ -1,0 +1,204 @@
+// #2704 on the PANEL path. The same lagging driver that made clear_vram print
+// a stale VRAM figure makes panel_free_vram publish a stale VERDICT: a card
+// read before the driver released looks globally occupied, so the reply names
+// occupied_devices and reports the pre-release byte count as the outcome of
+// the free.
+//
+// The frozen-tab path is the one that can be fixed, because freeVramDirect
+// already samples the counters BEFORE it posts /free — that pre-mutation
+// sample is what turns "the number stopped moving" into "the release landed".
+// The ack path has no such sample (the tab posted /free before it replied) and
+// deliberately keeps the older best-effort rule.
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const mocks = vi.hoisted(() => ({
+  comfyuiFetch: vi.fn(),
+}));
+
+vi.mock("../../comfyui/client.js", () => ({
+  getObjectInfo: vi.fn(),
+  backfillObjectInfo: vi.fn(),
+  resetClient: vi.fn(),
+  resetObjectInfoCache: vi.fn(),
+}));
+
+vi.mock("../../comfyui/fetch.js", () => ({
+  comfyuiFetch: (...args: unknown[]) => mocks.comfyuiFetch(...args),
+}));
+
+const { buildPanelToolDefs, makePanelToolCtx, __panelToolsTestHooks } = await import(
+  "../../orchestrator/panel-tools.js"
+);
+import { getBootLocalComfyUIBaseUrl } from "../../config.js";
+import { markReplyTimeout } from "../../services/ui-bridge.js";
+import { WorkflowTargetStore } from "../../services/workflow-target-store.js";
+import type { ToolResult } from "../../orchestrator/panel-tools.js";
+import { VRAM_SETTLE_INTERVAL_MS, VRAM_SETTLE_TIMEOUT_MS } from "../../services/vram-settle.js";
+
+const TAB = "11111111-2222-3333-4444-555555555555";
+const BOOT_BASE = (getBootLocalComfyUIBaseUrl() ?? "http://127.0.0.1:8188").replace(/\/+$/, "");
+
+const textOf = (res: ToolResult): string =>
+  res.content.map((c) => (c as { text?: string }).text ?? "").join(" ");
+
+const ackTimeout = (cmd: string, ms: number): Error =>
+  new Error(
+    `Panel tab ${TAB} did not reply to "${cmd}" within ${ms} ms — the ComfyUI tab may be ` +
+      `backgrounded or frozen. This command MUTATES and was already delivered to the tab, so it ` +
+      `may have been applied despite the missing reply — check the current state before ` +
+      `re-issuing it; a blind retry can apply it twice`,
+  );
+
+function bridge() {
+  return {
+    send: async (cmd: Record<string, unknown>) => {
+      if (cmd.cmd !== "free_vram") return { ok: true };
+      throw markReplyTimeout(ackTimeout("free_vram", 15000));
+    },
+    push: () => 1,
+    canReach: () => true,
+    isHeadless: () => false,
+    tabs: () => [{ tab_id: TAB, title: "wf", connected_at: 0 }],
+    resolveActiveTabId: () => TAB,
+    refreshWorkflowUuid: () => true,
+    workflowUuidFor: () => ({ known: false }),
+    tabCanMutateGraph: () => true,
+    tabGraphMutationCapability: () => ({ known: true, canMutate: true }),
+    tabIsLocal: () => true,
+    tabServerOrigin: () => BOOT_BASE,
+  };
+}
+
+const GIB = 1024 * 1024 * 1024;
+const TOTAL = 12 * GIB;
+/** 1.2 GB of 12 GB = 10% free: "occupied" by the same rule the panel applies. */
+const STALE_FREE = Math.round(1.2 * GIB);
+/** What the card really had once the driver caught up. */
+const SETTLED_FREE = Math.round(10.7 * GIB);
+const TORCH_TOTAL = 32 * 1024 * 1024;
+
+function gpu(vramFree: number, torchFree: number) {
+  return {
+    name: "cuda:0",
+    index: 0,
+    vram_total: TOTAL,
+    vram_free: vramFree,
+    torch_vram_total: TORCH_TOTAL,
+    torch_vram_free: torchFree,
+  };
+}
+
+/**
+ * The reported timing: torch hands its pool back almost at once, the driver
+ * stays frozen at the pre-/free number until `driverReleasesAtMs`.
+ *
+ * The torch movement is the whole trap — it is enough to make the combined
+ * counters "change then plateau", which is what certified the frozen driver
+ * number as settled.
+ */
+function setDriverLag(driverReleasesAtMs: number): { sampledAt: number[]; freeAt: () => number } {
+  let freeAt: number | null = null;
+  const sampledAt: number[] = [];
+  mocks.comfyuiFetch.mockImplementation(async () => {
+    freeAt = Date.now();
+    return { status: 200 };
+  });
+  __panelToolsTestHooks.setReadVramDevices(async () => {
+    sampledAt.push(Date.now());
+    if (freeAt == null) return [gpu(STALE_FREE, 0)];
+    const since = Date.now() - freeAt;
+    return [
+      gpu(since >= driverReleasesAtMs ? SETTLED_FREE : STALE_FREE, since >= 400 ? TORCH_TOTAL : 0),
+    ];
+  });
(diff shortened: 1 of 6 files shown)


artokun:
**Codex gate: VERDICT: PASS** on `a1f7781b26350772790dc5f3036e7c018b34dcd1` (the SHA that merges).

Five rounds via direct `codex exec` — the `codex-gate.mjs` wrapper failed with `CreateProcessAsUserW failed: 5 (Access is denied.)`, which is a sandbox failure with no code verdict, not a review result.

Rounds 1-4 each returned NEEDS-REVISION with a genuine finding, all fixed and pinned by a test that fails when that part is reverted:

1. **P1** — a failed pre-`/free` read left no baseline, which was treated exactly like an idle card, so a stale plateau was published as settled. Added `confirmable` to separate *nothing to prove* from *cannot prove anything*.
2. **P1** — release was latched globally, so one GPU moving certified another that never did.
3. **P1** — the per-device keys were identity-bearing strings compared **positionally**, so a reordered device list marked every watched card as moved. Now identity-keyed maps; a vanished device stays unmoved.
4. **P1** — colliding device identities silently merged in the map, letting one card's movement clear another's entry.
5. **P2** — the wait was armed from occupancy alone, so `clear_vram(unload_models:false, free_memory:false)` on a busy card sat out the full 12s cap and then warned it "had not finished releasing".

Round 5: `VERDICT: PASS` — "No concrete, realistic defect remains in the supplied diff."

Issue #900, opened 2026-08-05
A failed node install leaves a .git-only husk in custom_nodes that ComfyUI fails to import on every boot


artokun:
Found during a live browser pass on `main`. The panel agent noticed it while verifying a ComfyUI restart, and it checks out on disk.

## What is there

```
C:\Users\Artokun\ComfyUI-Installs\ComfyUI\ComfyUI\custom_nodes\comfyui-does-not-exist-99999\
  .git/          <- and nothing else
```

Created **2026-06-27**. It has been sitting in the user's `custom_nodes` for over a month, and ComfyUI logs an import failure for it on **every boot**:

```
custom_nodes\comfyui-does-not-exist-99999 fails with a missing __init__.py
```

## Why it is ours

The name is the shape of a deliberately-nonexistent registry pack — the input you would use to exercise the **git clone fallback** for a pack ComfyUI-Manager cannot resolve. A directory containing only `.git` is what a clone that started and did not complete leaves behind.

So the likely sequence: an install was attempted for a pack the registry does not have, the Manager route correctly found nothing, the clone fallback ran, the clone failed or produced nothing usable — and the partially-created directory was left in place.

## The defect

A failed install should not leave a broken pack behind. It is worse than doing nothing:

- ComfyUI tries to import it on every startup and logs an error, forever.
- It is indistinguishable, to a later reader, from a pack the user installed on purpose that is now broken.
- `list_installed_nodes` / disk-presence checks may count it as present — which is exactly the "present on disk vs tracked by Manager" three-way distinction #834 worked on, now fed a husk.

This is the same discipline as the lock and marker work: **do not leave a record of something that did not happen.** A clone that did not produce a usable pack should clean up its own directory, and if it cannot, it should say so rather than leaving something that reads as an installed pack.

## Suggested fix

In the clone fallback: create into a temp directory and move into `custom_nodes` only once the clone has produced something that looks like a pack (an `__init__.py` or a `pyproject.toml`). On failure, remove what was created; if removal fails, report the leftover path explicitly — a stray directory nobody was told about is how this one survived a month.

Worth also checking whether anything else in this codebase creates directories under `custom_nodes` before it knows the operation will succeed.

Refs #834, #808.

artokun:
[...]

PR #2715, opened 2026-09-01
fix(2714): a git install must be corroborated on disk, not by ComfyUI-Manager's own list

### LINKED ISSUE #2714: Fix false success for Manager v4 git custom-node install
`install_custom_node` reported that a git URL was installed through Manager HTTP, but no destination directory was created. With Manager v4, the arbitrary git URL was not installed; the tool trusted queue completion instead of verifying the expected checkout on disk.

### Steps to reproduce
1. Move an existing Registry ZIP pack out of `custom_nodes` so the target is absent.
2. Call `install_custom_node` with action `install`, source `git`, id `https://github.com/darksidewalker/ComfyUI-DaSiWa-Nodes`, ref `main`, and useCmCli `false`.
3. The tool returns `Installed ... via ComfyUI-Manager` with done_count 2.
4. Verify `~/ComfyUI/custom_nodes/ComfyUI-DaSiWa-Nodes`; it does not exist.

### Exact error (scrubbed)
Tool result: `Installed "ComfyUI-DaSiWa-Nodes" via ComfyUI-Manager. Restart may be required to load new nodes.`
Filesystem verification immediately afterward: `Resolve-Path: Cannot find path '~/ComfyUI/custom_nodes/ComfyUI-DaSiWa-Nodes' because it does not exist.`

### Fix
applied locally: no; upstream-only: yes
For `source:"git"`, do not route arbitrary URLs through Manager v4 HTTP. Use a verified local git/comfy-cli path, or refuse with the same Manager-v4 limitation as `panel_install_node`. After any install path, verify that the expected destination exists and, for git installs, contains `.git`; queue drain or done_count is not success proof.

### Environment
Windows 11; ComfyUI 0.33.3; ComfyUI frontend 1.49.6; RTX 5080 16 GB; comfyui-mcp 0.52.167; panel 0.15.105; Manager 4.2.2; Python 3.13.12; torch 2.10.0+cu130.

artokun:
Fixes #2714

## The report, reproduced

`install_custom_node` (`source:"git"`, `ref:"main"`, `useCmCli:false`) answered

> Installed "ComfyUI-DaSiWa-Nodes" via ComfyUI-Manager. Restart may be required to load new nodes.

with `done_count 2`, and `~/ComfyUI/custom_nodes/ComfyUI-DaSiWa-Nodes` did not exist. Manager 4.2.2.

The whole proof behind that sentence, on `origin/main`, was one question asked of ComfyUI-Manager:

```ts
const installed = await listInstalledNodesAt(managerBase).catch(() => []);
if (nodeInstalledMatches(gitId, installed)) {
  return { mechanism: "manager-http", message: `Installed "${repoName}" via ComfyUI-Manager. …` };
}
```

`/v2/customnode/installed` is Manager's **own bookkeeping**, not a filesystem read. For a pack it previously tracked it keeps answering after the directory is gone (the reporter moved a Registry ZIP pack out), and a v4 task that resolves nothing is still marked done. Neither the list nor `done_count` observes the disk.

The **registry** branch of the same function already crosses that list with a disk scan (`resolvePackPresence`, #797) and already refuses a marker-only husk (`looksLikeAPack`, #900/#1816). The **git** branch had neither.

## What changed — `src/services/node-management.ts`

- **`diskCorroboratesManagerInstall(gitId, listedNode, diskRoot)`** asks `findPackOnDisk` for every identity the matched Manager entry offers that could actually *be* a directory: the URL-derived repo name, the entry's stated folder key, its `cnrId`, and the repo half of its `auxId`. One hit corroborates. Returns `corroborated` / `husk` / `absent` / `unreadable`.
- **The git success gate** consumes it:
  - **corroborated** — the same `manager-http` success, now naming the verified directory.
  - **absent** — no success. Falls through to the *existing* direct-clone path, carrying a `managerRefusalNote` stating what was observed. `cloneCustomNodeFallback`'s default reason ("not in the ComfyUI-Manager registry") would be a wrong cause for a correct action, since Manager *did* list it.
  - **husk** — refuses, names the directory, leaves it untouched (this call is not its proven author).
  - **unreadable / remote / no proven scan root** — unchanged success, but the message now *says* the disk was not checked and that Manager's list is the only witness.
- **`InstalledNode.moduleKey`** (new, optional). `module` is a display-or-identity blend: `parseInstalled`'s array branch prefers the entry's human `title`. `module`'s value and precedence are **unchanged** — it is still what Manager is sent back as `node_name` and what the installed list displays — but anything reading a module as a *path* now uses `moduleKey`, which holds the key the payload actually stated (the object payload's key; the array entry's `module`) and is `undefined` when the payload stated none.

1. **The `absent` verdict retracts a success AND authorizes a filesystem write.** It is therefore the strictest branch: every identity must have been *conclusively* scanned. One `not-found` beside one unanswerable scan yields `unreadable`, not `absent`.
2. **The husk branch is a new throw where a success used to be.** `looksLikeAPack` returns `true` on an unreadable directory and on any top-level `.py`/`pyproject.toml`, and it is already the gate on the other two paths in this same function.
3. **I did NOT implement the `.git`-must-exist check the report prescribes.** The v4 route is registry-first by construction: a pack Manager resolved and unpacked from the registry has no `.git` and is working as designed. Requiring `.git` would false-refuse it. The property that matters is whether ComfyUI can import the directory. The report's *premise* (false success) is right; that one clause of its prescribed fix is not.
4. **Disclosed and accepted:** a pack installed under an `extra_model_paths`-style second `custom_nodes` root is not seen by `findPackOnDisk`, so it could be cloned again into the primary root. The same exposure already exists on this function's pre-existing not-listed fall-through. `moduleKey` is `undefined` for the cm-cli `show installed` table path; that path is not used by the install verification.
5. **Not covered by a test:** the `unreadable`-outranks-`absent` ordering. `findPackOnDisk` returns `unreadable` only for root-level failures, so all identities scanned against one root share that outcome and a mixed state is reachable only under a mid-scan race. It is ordering hardening, not a demonstrated failure — claiming coverage for it would be false.

## Evidence

New pin: `src/__tests__/services/install-git-manager-listed-but-absent-2714.test.ts` (10 tests) drives the REAL `installCustomNode` through the REAL resolvers against a REAL temp tree, with Manager accepting the task, draining `done_count 2`, and **listing** the pack. The existing #1765 coverage never produces this shape (its `/customnode/installed` always answers `[]`), which is why the list never had to be doubted.

**Deleting the fix** (`git checkout origin/main -- src/services/node-management.ts`) turned 5 of the original 6 red, the first with the reporter's outcome verbatim (`expected 'manager-http' to be 'git-clone'`).

**Every clause is pinned independently.** Each mutation turns exactly the named test(s) red and leaves the rest green:

| mutation | red |
| --- | --- |
| identity list reduced to `[gitId]` | corroborates through the MANAGER ENTRY's own id |
| `basename()` removed from `packDirAliases` | corroborates through the aux id's REPO HALF |
| whitespace filter re-added | a pack folder with a SPACE in its name still corroborates |
| alias scan reads `node.module` instead of `moduleKey` | will not let a human TITLE stand in… + a TITLE-ONLY entry contributes no folder name |
| `parseInstalled` precedence moved | a title-derived module carries NO folder key |

The over-refusal control (`corroborates through the MANAGER ENTRY's own id`) **survives** the fix-deletion mutation, as a control must.

Full suite: **754/754 files, 13894 passed, 6 skipped**. `tsc --noEmit` clean; anti-slop gate `none added`.

**Browser gate not run, and not claimed:** nothing here touches panel rendering or the `panel_install_node` bridge path — the change is confined to the MCP-side `install_custom_node` service.

## Review

`codex-gate.mjs` returned the known Windows wrapper failure (`CreateProcessAsUserW failed: 5 (Access is denied)` — an indeterminate, not a finding), so the review ran through direct `codex exec` with the diff supplied inline. **Four rounds**; the first three each found a real defect in this change, and all are fixed above:

- **R1** — `aux_id`'s repo half was never scanned (duplicate clone); a human title could vouch for an unrelated directory; one `not-found` beside one `unreadable` returned `absent`, authorizing a write on an unproven absence.
- **R2** — an entry stating its own `module` had it overridden by `title`; **R1's whitespace filter rested on a false claim of mine** (I asserted ComfyUI imports a pack directory as a Python module name so it cannot contain a space — it loads packs via `importlib.util.spec_from_file_location`, which accepts any string), and would have read a real `custom_nodes/Foo Bar` as absent and cloned a duplicate beside it; a call-site comment overstated the fall-through.
- **R3** — R2's fix was both too wide and incomplete: it moved `module`'s precedence (a field sent to Manager as `node_name`) while an entry of `{title, cnr_id}` still promoted its title into `module`. Reverted the precedence change; added `moduleKey`.
- **R4 — `VERDICT: PASS`** on `1d61cf7`: *"No behavior-changing defect remains in the supplied diff. `module` precedence is unchanged; `moduleKey` is threaded correctly, corroboration ordering is safe under the stated helper contract, and `uncorroboratedNote` is scoped only to the relevant fallback clone."*

### DIFF
--- a/src/__tests__/services/install-git-manager-listed-but-absent-2714.test.ts
+++ b/src/__tests__/services/install-git-manager-listed-but-absent-2714.test.ts
@@ -86,8 +86,10 @@ vi.mock("../../services/live-interpreter.js", async (importOriginal) => {
 // ── ComfyUI-Manager v4: the queue ACCEPTS, drains "done", and the installed-pack
 //    list ANSWERS FOR the pack. Exactly the reporter's wire shape. ─────────────
 const manager = vi.hoisted(() => ({
-  /** Body served at /v2/customnode/installed. */
-  installed: {} as Record<string, unknown>,
+  /** Body served at /v2/customnode/installed. Manager serves the object-keyed
+   *  shape; the ARRAY shape is `parseInstalled`'s other branch, which is where a
+   *  human `title` can end up standing in for the module key. */
+  installed: {} as Record<string, unknown> | unknown[],
   calls: [] as string[],
 }));
 vi.mock("../../comfyui/fetch.js", () => {
@@ -296,7 +298,7 @@ describe("#2714 — a Manager-listed git install must be corroborated on disk", 
     // never happened, which is this issue's own bug wearing a different hat.
     manager.installed = [
       { title: "Shared Title", aux_id: `darksidewalker/${REPO_NAME}`, ver: "nightly", enabled: true },
-    ] as unknown as Record<string, unknown>;
+    ];
     makePackDir("Shared Title", { "__init__.py": "NODE_CLASS_MAPPINGS = {}\n" });
 
     const { ok, error } = await install({ id: REPO, source: "git" });
--- a/src/__tests__/services/install-git-manager-listed-but-absent-2714.test.ts
+++ b/src/__tests__/services/install-git-manager-listed-but-absent-2714.test.ts
@@ -1,0 +1,296 @@
+// #2714 — `install_custom_node` reported a git install that never happened.
+//
+// The reporter moved an existing Registry-ZIP pack out of `custom_nodes`, then
+// installed the same pack by git URL with `useCmCli:false` against ComfyUI-Manager
+// 4.2.2. The tool answered
+//
+//   Installed "ComfyUI-DaSiWa-Nodes" via ComfyUI-Manager. Restart may be required…
+//
+// with done_count 2, and `~/ComfyUI/custom_nodes/ComfyUI-DaSiWa-Nodes` did not exist.
+//
+// The whole proof behind that sentence was ONE question asked of ComfyUI-Manager:
+//
+//   const installed = await listInstalledNodesAt(managerBase).catch(() => []);
+//   if (nodeInstalledMatches(gitId, installed)) return { mechanism: "manager-http", … };
+//
+// `/v2/customnode/installed` is Manager's own bookkeeping. It is not a filesystem
+// read, and for a pack Manager previously tracked it keeps answering after the
+// directory is gone — while a v4 task that resolves nothing is still marked done.
+// The registry branch of the SAME function already crosses that list with a disk
+// scan (`resolvePackPresence`) and already refuses a marker-only husk
+// (`looksLikeAPack`, #900/#1816). The git branch had neither.
+//
+// These tests drive the REAL `installCustomNode` through the REAL resolvers against
+// a REAL directory tree, with the Manager LISTING the pack — the one shape the
+// existing #1765 coverage never produces (its `/customnode/installed` always
+// answers `[]`, so the list never had to be doubted).
+import { afterAll, afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { existsSync, mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+// ── One REAL local install, which the server reports as its own root ───────────
+const SANDBOX = mkdtempSync(join(tmpdir(), "manager-listed-absent-2714-"));
+const CONNECTED = join(SANDBOX, "comfy");
+mkdirSync(join(CONNECTED, "custom_nodes"), { recursive: true });
+mkdirSync(join(CONNECTED, "models"), { recursive: true });
+writeFileSync(join(CONNECTED, "main.py"), "# ComfyUI\n");
+const WORKSPACE_JSON = join(SANDBOX, "workspace.json");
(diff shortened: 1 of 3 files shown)


artokun:
Review addendum — the SHA that merges.

The `VERDICT: PASS` quoted in the description names `1d61cf7`. The head then moved to `8a31d68` for a comment correction (`packDirAliases` still said the title-in-`module` contamination was "fixed at its source"; round 3 reverted that precedence change, so it is fixed by reading `moduleKey`). A verdict names a SHA and does not carry across a head move, so it was re-gated:

> **VERDICT: PASS**
> 1. Comments only; no runtime behavior changes.
> 2. Yes. The comment is true: contamination is removed by reading `moduleKey`; `parseInstalled` precedence is unchanged.

Verifiable independently: `git diff 1d61cf7..8a31d68` is 3 insertions / 2 deletions, all inside a doc comment.

PR #2718, opened 2026-09-01
fix(2716): node_pack git resolves `paths` against the pack, as the schema promises

### LINKED ISSUE #2716: node_pack git rejects documented pack-relative paths
`node_pack(action:"git", git_action:"diff")` rejects documented pack-relative entries in `paths` as outside the selected pack. The jail validator appears to resolve each path against the wrong root or require an undocumented pack-name prefix.

### Steps to reproduce
Call node_pack with:
```json
{"action":"git","pack":"comfyui-anima-style-dataset","git_action":"diff","paths":["preset_core.py","tests/test_preset_core.py"]}
```
Both files exist under the selected pack.

### Exact error
```
{"error":"NODE_DEV_ERROR","message":"Path \"preset_core.py\" is outside the target pack."}
```

### Fix
applied locally: no
upstream-only: yes
Resolve `paths` relative to the already-selected pack root as the schema promises, then jail-check the resolved absolute path against that root.

### Environment
Windows 11; ComfyUI 0.21.1; frontend 1.43.18; Python 3.13.12; PyTorch 2.11.0+cu130; NVIDIA RTX 5080 16 GB; comfyui-mcp 0.52.150; panel 0.15.113.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-sol` · effort `xhigh` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Closes #2716.

## The defect

`node_pack action:"git"` documents `paths` as **"pack-relative paths to stage/scope"**
(`src/tools/node-pack.ts:263`), and the pack is already chosen by `pack`. But
`packRelativePath()` handed each entry straight to `resolveInJail()`, and that anchors a
**relative** input at the `custom_nodes/` **ROOT**. So `"preset_core.py"` resolved to
`custom_nodes/preset_core.py`, `relative(packDir, abs)` came back `"../preset_core.py"`,
and the call was refused as *"outside the target pack"*. Every documented spelling failed;
the only one that worked was an undocumented pack-name prefix.

Reproduced by execution on the published server (0.52.167) against a real install, with a
different pack and a different `git_action` than the report — so it is the general defect,
not a pack quirk:

```
node_pack {action:"git", pack:"ComfyUI-Manager", git_action:"log", paths:["__init__.py"]}
→ {"error":"NODE_DEV_ERROR","message":"Path \"__init__.py\" is outside the target pack."}
```

## The fix

Anchor a relative entry at the pack. **Containment is unchanged**: the resolved path still
goes through the single auditable jail resolver (lexical + realpath, so the `custom_nodes/`
jail and its symlink/junction check are intact) and still has to land inside the selected
pack.

### Where the load-bearing claims are — please attack these

1. **Why the anchor is applied to the jail-relative STRING (`join(packName, raw)`) and not
   by joining onto `packDir`.** Two reasons, both silent if wrong:
   - `packDir` is the **realpath'd** directory. On an install whose `custom_nodes/` is a
     junction (common on Windows), an absolute join would fail `resolveInJail`'s FIRST
     containment check, which is lexical against the **un-resolved** root — breaking those
     installs. `packName` is separator-free and `..`-free (`assertSafeRepoName`), so
     `<pack>/<entry>` re-enters the resolver on exactly the path it already handles.
   - The resolver's Windows-hazard scan walks every segment of what it is given. Handing it
     an absolute path drags the user's whole install path through that scan.
2. **The hazard shapes still refused, now that `join()` runs first.** `join()` collapses
   only `..`; UNC / drive-relative / ADS / reserved-name / trailing-dot segments survive
   into the joined string and the resolver's own scan still refuses them. The collapsed
   `..` case lands on the pack-containment check instead (`../Other/loot.py` →
   *"outside the target pack"*). No shape becomes reachable.
3. **The empty-entry guard.** `paths:[""]` would resolve to the pack itself and silently
   widen the command to every file in it — which is what *omitting* `paths` already means.
4. **The `assertNotPackPrefixed` guard is fallout management, not the fix.** The prefixed
   spelling this bug forced callers to learn now names `Pack/Pack/x.py`, and git answers a
   pathspec matching nothing with an **empty diff, not an error**. That silent shape is
   refused with the correction named — but only where the disk proves the literal reading
   was not meant (prefixed target absent AND de-prefixed one present), so a pack that
   really does contain a same-named subdirectory keeps the literal reading.

## Evidence

**Mutation-tested — every piece of the fix, deleted independently, turns a NAMED test red:**

| mutation | result |
| --- | --- |
| anchor reverted to the custom_nodes root (the original bug) | 5 failed |
| `assertNotPackPrefixed` call removed | 1 failed (`names the correction for the pack-name-prefixed spelling…`) |
| empty-entry guard removed | 1 failed (`refuses an empty path entry`) |
| pack-containment check disabled (`if (false)`) | 2 failed (sibling-pack + climb-out) |

The empty-entry mutation initially read as SURVIVED because the string I deleted occurs
**twice** — the first hit is `resolveInJail`'s own guard. Re-run against the right site, it
fails. (Noted because it means `resolveInJail`'s empty-input guard has no test of its own;
pre-existing, not touched here.)

**End-to-end through the real production handler** (`registerNodePackTools`, including
`resolveCustomNodesScanBaseLiveStrict()`), against the real local install — mutations prove
a mechanism, not that production reaches it:

```
paths=["__init__.py"]                  → argv [… -- __init__.py]           stdout "94a3ae8 … update DB"
paths=["glob/manager_core.py"]         → argv [… -- glob/manager_core.py]  stdout "94a3ae8 … update DB"
paths=["no-such-file.py"]              → argv [… -- no-such-file.py]       stdout ""      ← control: the pathspec really filters
paths=["ComfyUI-Manager/__init__.py"]  → refused, naming the correction
paths=["../ComfyUI-Crystools/…"]       → "outside the target pack"
```

The `no-such-file.py` row is the control: without it, "it returned output" would not
distinguish a working pathspec from an ignored one.

- Full suite: **754 files / 13907 passed, 6 skipped**.
- `npm run lint` (tsc --noEmit + anti-slop): clean, "none added".
- **No panel surface is touched**, so no Playwright run is claimed.

## Deliberately not done

The `paths` schema description is left alone — it already promises pack-relative; the code
was what disagreed. Editing it would pull in the tool-vocabulary ratchet and eleven locale
files for no behaviour change.


artokun:
Recovered and re-engaged after the >2h abandonment window. Rebased the existing branch onto origin/main b741ebabb76b7bf2c3b57c6eefafb80444745ec1 and force-with-lease pushed exact head b396c1e (no duplicate PR).\n\nRepair: node_pack action:git paths now resolve relative to the selected pack root, then retain custom_nodes lexical/realpath jail and selected-pack containment. Raw Windows hazard validation runs before pack anchoring so UNC spellings cannot be normalized into safe-looking paths; . and .. remain traversal operators and containment still rejects escapes.\n\nFocused: npm.cmd run test:vitest -- src/__tests__/services/node-dev.test.ts src/__tests__/tools/node-pack.test.ts — 127/127 passed. Build: npm.cmd run build — passed. Full: npm.cmd test — 760 files, 13,986 passed, 6 skipped (13,992 total), all 13 repository checks passed. Added/updated regressions for missing pack, pack-relative/prefixed paths, absolute/traversal/symlink/UNC/ADS safety, deleted paths, and schema/handler forwarding/description contract. Copilot's earlier UNC concern is addressed; deleted-path behavior remains supported.

artokun:
Recovery follow-up: generated docs are now included at exact head a9c8e0d7ae0545db2e4ceb7a6f91b5de64fc1324 (docs/tools/custom-nodes.mdx only). npm.cmd run docs:gen is a no-op, focused node_pack path-safety/handler tests pass 127/127, build passes, and hosted CI run 33654579703 is green on Windows, macOS, Ubuntu, pack/install smoke, and docs generation.

PR #2728, opened 2026-09-02
chore: ignore root agent scratch and vendored node packs

artokun:
## What this is — and what it is *not*

This PR was opened while investigating a report that commit `3ef895f`
("refactor(1869): simplify tests…") had leaked a GPL-3.0 node pack and ~36 MB of
scratch into the public repo.

**That did not happen.** The investigation is written up below; the only real,
actionable residue is this `.gitignore` hardening.

## Findings

`3ef895f` is a **single orphaned commit**, reachable only from the detached HEAD of
the local main checkout:

- `git for-each-ref --contains 3ef895f` → **no ref**
- `git ls-remote origin` → **0 of 3443 remote refs** point at it
- `git merge-base --is-ancestor 3ef895f origin/main` → **not an ancestor**
- `origin/main` (`b8944fd`) tracks **zero** `ComfyUI-Impact-Pack-Main/` or `.tmp-*` paths

So nothing was ever pushed: no license misrepresentation in the public tree, and no
clone bloat. A fresh clone of `main` is clean. There is nothing to `git rm --cached`.

What *is* true about that commit's tree: 291 unrelated files (102 pack + 189 scratch,
36.4 MB), and `ComfyUI-Impact-Pack-Main/LICENSE.txt` genuinely is GNU GPL v3.

## Why this PR still exists

None of those paths were ignored, so the same wide-`add` accident could recur — and
next time it might land on a branch that gets pushed. This adds root-anchored rules:

| Pattern | Covers |
| --- | --- |
| `/.tmp-*` | agent scratch dirs and files |
| `/.playwright-mcp/` | browser console-log dumps |
| `/ComfyUI-Impact-Pack-Main/`, `/ComfyUI-*-Main/` | third-party packs unpacked for inspection |

The pack rule is the one that matters beyond tidiness: Impact Pack is GPL-3.0 and
this repo is MIT, so a vendored copy would misstate the license of the tree.

## Verification

- `git check-ignore` over the **305** offending paths from `3ef895f` → all 305 ignored.
- `git check-ignore` over all **2026** currently tracked paths → **none** matched, so no
  tracked file changes status.
- Ignore rules never affect already-tracked files, so this is behaviour-neutral for `main`.

## Not done here (owner's call)

History rewrite is **not** needed and is not proposed — there is no published history
containing these files. See the accompanying report for the two things that do want an
owner decision: the main checkout parked in detached HEAD on this orphan commit, and
whether `3ef895f` is worth salvaging or discarding.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,15 @@ orch.err.log
 # Local Discord watchdog/notifier toolkit (token read from ~/.claude.json at runtime; not committed)
 /.discord/
 codex-kp-smoke-orch.log
+
+# Agent scratch at the repo root. A stray `git add -A` once swept 291 of these
+# into a single commit (3ef895f, local-only — never reached any pushed ref).
+# Ignore them so a wide `add` can't sweep them in again.
+/.tmp-*
+/.playwright-mcp/
+
+# Third-party custom-node packs unpacked into the repo root for inspection.
+# They ship their own licenses (Impact Pack is GPL-3.0) and must never land in
+# this MIT-licensed repo. Clone them under the ComfyUI install, not here.
+/ComfyUI-Impact-Pack-Main/
+/ComfyUI-*-Main/


artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6). The patterns are correct, and checking the commit they cite turned up something you probably want to know.

## The patterns do what they claim, and no more

Nothing currently tracked becomes ignored (`git ls-files` against each pattern: no matches). And `git check-ignore` confirms the scoping is tight rather than incidental:

```
.tmp-x                       IGNORED
.playwright-mcp/a            IGNORED
ComfyUI-Impact-Pack-Main/x   IGNORED
ComfyUI-Foo-Main/x           IGNORED
src/ComfyUI-Foo-Main/x       not ignored   ← leading / anchors to root
ComfyUI-Foo-MainThing.md     not ignored   ← trailing / restricts to directories
```

Both ways an ignore rule usually goes wrong — swallowing a nested path, and catching a similarly-named file — are absent.

## The cited commit is still your checkout's HEAD, and it carries the GPL pack

The comment says a stray `git add -A` swept these into `3ef895f`, "local-only — never reached any pushed ref". The safety half is **true**: `git branch -r --contains 3ef895f` returns nothing, and no local branch contains it either.

But it is still there. `3ef895f` is the current **detached HEAD** of the main checkout, and its contents are:

| what | files |
|---|---|
| `ComfyUI-Impact-Pack-Main/` (**GPL-3.0**) | 102 |
| `.tmp-*` scratch | 189 |
| `.playwright-mcp/` | 14 |
| loose scratch (`HANDOFF.md`, `ab.bak`, `impactpack.tar.gz`, `probe1524.mjs`, …) | 19 |
| actual source | **1** (`src/__tests__/tools/stale-note-writer-dead.test.ts`) |

All 325 are **additions** — `--name-status` is `325 A`, nothing modified or deleted. So the message *"remove handler-mocking tests that don't drive production code"* describes work the commit does not contain.

**Why it matters:** this PR stops it happening again, but it does not remove what already happened. A single `git push HEAD:some-branch` from that checkout would publish 102 GPL-3.0 files into an MIT repo. That is the one outcome the ignore rules exist to prevent, and the rules cannot reach a commit that already exists.

I have not touched it — it is your working checkout and dropping a commit is yours to decide. The one thing in it worth keeping looks like that single test file, which cherry-picks cleanly if you want it before the commit goes.

PR #2750, opened 2026-09-02
fix(2748): a pack workflow miss must not advertise the pack it just refused

### LINKED ISSUE #2748: list_packs reports qwen-image has_workflow=true but read_workflow and check_runtime say workflow.json is missing
`list_packs` action `list` reports the bundled pack `qwen-image` with `has_workflow: true`, but `list_packs` action `read_workflow` returns `Pack "qwen-image" has no ready workflow (workflow.json not found)`, and `check_runtime` says `No pack named "qwen-image" with a ready workflow` while still listing qwen-image among available packs. The catalog metadata is inconsistent with the filesystem and can cause agents to promise/load a nonexistent expert workflow.

### Steps to reproduce
1. Call list_packs {action:"list"}; inspect qwen-image.has_workflow.
2. Call list_packs {action:"read_workflow", name:"qwen-image"}.
3. Call list_packs {action:"check_runtime", pack:"qwen-image"}.

### Exact error
`Pack "qwen-image" has no ready workflow (workflow.json not found).`
`No pack named "qwen-image" with a ready workflow.`

### Fix
Applied locally: no. Upstream-only: likely metadata generation/caching mismatch. Compute `has_workflow` from the same resolved pack directory and filename used by `read_workflow`/`check_runtime`, or ship the referenced workflow.json.

### Environment
Linux; NVIDIA RTX PRO 6000 Blackwell Server Edition 95 GB; ComfyUI 0.33.0; comfyui-mcp 0.52.176; panel 0.15.155.

---

artokun:
Closes #2748

## What the reporter hit

`list_packs action:"check_runtime" pack:"qwen-image"` answered:

> `No pack named "qwen-image" with a ready workflow. Available packs: <all 56, including qwen-image>`

…for a pack that `action:"list"` plainly lists. The list offered as the remedy **contained the pack it had just refused**, so copying a name out of it reproduced the identical failure. That is what made the catalog look like it disagreed with the filesystem.

## Two independent defects

**1. The suggestion list was never filtered.** Both workflow exits built it as `enumeratePacks().map(p => p.name)` — no `has_workflow` filter — under a sentence promising *"with a ready workflow"*. Two of the 56 bundled packs (`qwen-image`, `ltx-2.3`) are installer-only and ship no graph.

The sibling call site already does this correctly: `runTemplateAction` (`src/tools/run-template.ts:141-143`) filters `.filter((p) => p.has_workflow)` before suggesting. **One exit was fixed and two were left behind** — this restores the family.

**2. The workflow filename was derived three times and one copy differed.** `enumeratePacks` took `meta.workflow` verbatim; `resolvePackWorkflowFile` rejected anything that is not a single safe path segment and fell back to `workflow.json`. A pack declaring `workflow: "sub/graph.json"` would report `has_workflow: true` for a file the resolver never opens while `read_workflow` said `workflow.json not found` — **exactly the reported symptom**. Reachable, but not triggered by today's pack data, so a corpus test alone cannot see it.

## The change

- `resolveWorkflowFileName()` — one exported derivation now shared by `action:"list"`, `resolvePackWorkflowFile()` and `read_workflow`'s own message. The divergence becomes unrepresentable rather than merely absent.
- `noWorkflowSentence()` — distinguishes *"no such pack"* from *"this pack exists and is installer-only"*, and does **not** assert a cause it did not check (a declared-but-absent filename reads differently from `workflow: null`).
- Both workflow exits suggest only packs that actually have a ready workflow.

Observed output on this branch:

```
Pack "qwen-image" ships no workflow — pack.yaml declares no workflow, so it is
installer-only. action:"list" reports has_workflow: false for it. There is no
graph to runtime-check. Packs with a ready workflow: <54 names, qwen-image absent>
```

## What does NOT reproduce — and I did not manufacture a fix for it

The `has_workflow: true` half of the report. I unpacked the **published `comfyui-mcp@0.52.176` tarball** the reporter names and enumerated its own bundled `packs/`: `qwen-image → has_workflow: false` there and on `main`, with zero divergence across all 56 packs. `packs/qwen-image/` ships no `workflow.json` in either tree. There is no catalog/filesystem inconsistency to fix; the misleading *message* is the whole defect.

## Where the load-bearing claims are (please poke here)

- **`packDirExists()` is a new filesystem read on an error path.** It is `statSync` in a try/catch returning `false`, name-guarded by `SAFE_NAME` and prefix-checked against `packsDir()` before use — but it is a new failure surface inside a branch that previously only formatted a string.
- **`resolveWorkflowFileName` is now the single gate for a path segment that gets `join`ed.** It is strictly *tighter* than the old `enumeratePacks` copy (which had no guard) and equivalent to the resolver's, so this narrows rather than widens what can be opened — worth a second read anyway.
- **`read_manifest`'s list is deliberately left unfiltered.** All 56 packs have a manifest; filtering it by `has_workflow` would hide the two packs that exist *precisely* for manifest installs. Pinned with a test asserting `read_manifest` still succeeds for an installer-only pack.
- The corpus test (`has_workflow` agrees with the resolver for every bundled pack) **passes before and after** — it is a forward-looking invariant, not the behavioural pin. The behavioural pins are the mutation-verified ones below.

## Verification — fix deleted, tests watched to fail

| Mutation | Result |
|---|---|
| Remove `.filter(p => p.has_workflow)` | **3 tests fail**, one per exit |
| Replace `packDirExists(...)` with `false` | **1 test fails** — "does not deny the pack exists" |
| Remove the `SAFE_NAME` guard in the shared derivation | **1 test fails** — the filename test |

- Full suite: **760 files, 13994 passed, 6 skipped, 0 failed**.
- `npm run lint` (tsc + anti-slop): clean, *none added*.
- `npm run check:vocabulary`: OK.
- One test-authoring bug the tests themselves caught: my first list parser used `[^.]*`, which truncates at the dot in `ltx-2.3`. Names are parsed exactly now — `qwen-image` is a prefix of `qwen-image-edit`, so substring matching would silently pass on the wrong pack.

**No panel code was touched** (`src/orchestrator/panel-tools.ts` is deliberately unchanged — its messages do not advertise a false list). This changes MCP tool result text only, so the Playwright browser suite was not run and this PR claims no browser coverage.

### DIFF
--- a/src/__tests__/tools/skills-access.test.ts
+++ b/src/__tests__/tools/skills-access.test.ts
@@ -65,6 +65,7 @@ import {
   enumeratePacks,
   resolvePackWorkflowFile,
   resolveWorkflowFileName,
+  describeMissingWorkflow,
 } from "../../tools/skills-access.js";
 
 type Handler = (args: Record<string, any>) => Promise<{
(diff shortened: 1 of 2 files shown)


artokun:
**Independent review: 5 codex-gate rounds, 3 real P1s found and fixed.**

Recording this because rounds 2–4 each caught a genuine defect *I introduced while fixing this issue* — all of them the same class the PR exists to remove (a message asserting a cause nobody checked). Worth reading as a group.

| Round | SHA | Verdict | Finding |
|---|---|---|---|
| 1 | `a509893` | SHIP | — (verdict did NOT carry; head moved for a vocabulary-gate fix) |
| 2 | `fee084a` | NO-SHIP | **P1** — malformed/absent `pack.yaml` and `workflow: 42` all reported as intentional *installer-only* |
| 3 | `3ea4d9f` | NO-SHIP | **P1** — `parseYaml("[]")` is truthy AND `typeof "object"`, so a top-level YAML **sequence** passed as a metadata record |
| 4 | `6fe5c14` | NO-SHIP | **P1** — an **omitted** `workflow` key treated as an explicit `workflow: null` |
| 5 | `48111a5` | **SHIP** | clean on the SHA that merges |

Round 3 is the one I would not have found myself: `[]` is truthy and `typeof === "object"`, so the obvious `parsed && typeof parsed === "object"` mapping check accepts it, every field reads `undefined`, and `workflow: undefined` is indistinguishable from a deliberate `workflow: null`.

Round 4 I checked against the corpus rather than taking it on faith — all 56 bundled packs write `workflow:` explicitly, confirming an omitted key is not this repo's idiom for installer-only and so expresses no intent.

Each fix was verified by deleting it and watching named tests fail (3 / 2 / 1 tests respectively). CI green on `48111a5` across all four jobs; `npm test` — the command CI runs — green locally across all 14 checks.

One process note: round 1's SHIP was on a SHA that a later commit replaced, so it was re-gated rather than banked. An earlier `check:vocabulary` pass was also stale — it ran before the doc comment that broke it existed, which is why the first CI run went red. Verifying with `npm test` rather than piecemeal `vitest` + `lint` closed that gap.

Issue #1037, opened 2026-08-07
UI→API converter drops nested DYNAMICCOMBO widget values when a sibling is linked — LTX-2.3 packs report success with no output


artokun:
## Summary

The UI→API workflow converter drops the **non-linked sibling values of a nested DYNAMICCOMBO widget** when at least one sibling of that same widget group is converted-to-input (linked). The resulting prompt is missing a required input, ComfyUI rejects the output node, and the run reports **`success` with no output produced**.

Affected packs (verified by scanning all 55 bundled `packs/*/workflow.json`):

- `ltx-2.3-txt2vid`
- `ltx-2.3-img2vid`

Both carry the same `ResizeImageMaskNode` inside `definitions.subgraphs[0].nodes[14]` (node id 290, surfaced as 343 after flattening).

## The node in the pack

```json
{
  "id": 290,
  "type": "ResizeImageMaskNode",
  "widgets_values": ["scale dimensions", 1920, 1088, "center", "lanczos"],
  "inputs": [
    { "name": "input" },
    { "name": "resize_type.width",  "widget": {...}, "link": 558 },
    { "name": "resize_type.height", "widget": {...}, "link": 559 }
  ]
}
```

`resize_type = "scale dimensions"` requires **width, height AND crop**. `crop` ("center") and `scale_method` ("lanczos") are present in `widgets_values`, but because `width`/`height` are converted-to-input the converter refuses positional resolution and emits neither, so the produced prompt has no `resize_type.crop`.

The converter itself flags this, and the warning is accurate:

```
Node 343 (ResizeImageMaskNode): widget "resize_type" has a converted-to-input (linked)
nested widget "resize_type.width" whose widgets_values slot can't be positionally resolved
... so the remaining 4 saved widget value(s) are left to their defaults
```

The problem is that "left to their defaults" is not what happens — the key is simply **absent** from the prompt, and `crop` has no server-side default fallback, so validation fails.

## Reproduction

```
enqueue_workflow(action="run_template", template="ltx-2.3-img2vid",
                 overrides={"353.value": 2, "377.value": false})
```

Result (400):

```
ResizeImageMaskNode (node 343): Required input is missing (crop)
  extra_info -> input_name: "resize_type.crop"
```

For `ltx-2.3-txt2vid` the failure is worse because it is **silent**. Nothing else in that graph fails validation, so ComfyUI logs:

```
[ERROR] Failed to validate prompt for output 75:
[...]

PR #2755, opened 2026-09-02
fix(2753): a forceInput-only input is a socket, not a widget slot

### LINKED ISSUE #2753: UI to API converter shifts widgets after scalar forceInput socket
get_workflow action:"query" mis-maps positional widgets_values for a custom node whose first scalar input is forceInput-only. The converter's isWidgetInput() classifies STRING/FLOAT/etc. as serialized widgets without first excluding cfg.forceInput. It therefore consumes widgets_values[0] for the socket-only input and shifts every real widget one position.

### Steps to reproduce
1. Save a UI workflow with a node input order: trainer_status=["STRING",{forceInput:true}], then unet_name, clip_name, vae_name, weight_dtype, clip_device widgets.
2. The node serializes widgets_values as ["anima-base-v1.0.safetensors","qwen_3_06b_base.safetensors","qwen_image_vae.safetensors","default","default"].
3. Call get_workflow action:"query" fields:"detail".
4. Observe unet_name reported as qwen_3_06b_base, clip_name as qwen_image_vae, vae_name as default. Raw UI and the live canvas have the correct values.

This produced misleading saved-workflow evidence during a Stage C repair and could have caused a valid workflow to be changed incorrectly.

### Exact error (scrubbed)
No exception. Wrong converted output:
unet_name=qwen_3_06b_base.safetensors
clip_name=qwen_image_vae.safetensors
vae_name=default
while raw UI stores:
unet_name=anima-base-v1.0.safetensors
clip_name=qwen_3_06b_base.safetensors
vae_name=qwen_image_vae.safetensors

### Fix
applied locally: yes
upstream-only: no

Patched the installed dist/services/workflow-converter.js:
- move cfg = spec[1] before scalar classification
- return false when cfg.forceInput === true
- leave the existing scalar/combo mapping unchanged

Verification: a synthetic conversion with the same forceInput + five-widget shape now maps every widget correctly and returns no warnings.

### Environment
Windows 11
ComfyUI 0.21.1
ComfyUI frontend 1.43.18
NVIDIA GeForce RTX 5080 16 GB
comfyui-mcp 0.52.150
panel 0.15.113

artokun:
Fixes #2753.

## The defect

`isWidgetInput()` in `src/services/workflow-converter.ts` classified an input by its
TYPE only. `["STRING", {forceInput: true}]` looks exactly like a real STRING widget,
so it was given a positional `widgets_values` slot — but ComfyUI's frontend checks
`forceInput` **first** and renders a socket instead of a widget, so no slot is ever
written for it. Consuming one shifts every real widget after it by one.

The reporter's node returned `unet_name=qwen_3_06b_base`, `clip_name=qwen_image_vae`,
`vae_name=default` from a row whose true values sat one position earlier. Reached
through `get_workflow` (`query`/`strip`/`slice`) → `convertUiWorkflow` →
`convertUiToApi` (`src/tools/workflow-library.ts:188`).

Not confined to one custom node: rgthree's `SDXLPowerPromptPositive` declares
`opt_clip_width` and `opt_clip_height` as `["INT", {forceInput: True, ...}]` ahead of
three combo widgets, so it read two positions late on every install. `forceInput`
appears 536 times across 26 of the packs installed on this machine.

## The fix

Purely additive — two predicates, one guard line, one warning. The whole production
diff is 92 lines, most of them comment.

- `isSocketOnlyInput(name, def)` gates `isWidgetInput()` **before** any type
  classification, matching the frontend, whose widget factory returns early on
  `forceInput` regardless of type (`comfyui_frontend_package` 1.49.6:
  `let s = widgets.get(type); if (!s || spec.forceInput) return`, with the sibling
  `addInputSocket` creating the socket instead).
- It also handles the deprecated `defaultInput` spelling **asymmetrically**, exactly
  as `ComfyNodeDefImpl._migrateDefaultInput` does: on an **optional** input it becomes
  `forceInput` (socket-only); on a **required** input the frontend only warns
  "please drop the defaultInput option" and KEEPS the widget. Promoting a required one
  would drop a slot the frontend does write — the same shift, pointed the other way.
- Both conversion directions change together: `convertApiToUi` now emits a plain
  socket rather than a `widget:{name}` slot plus a placeholder.
- The rule stays **top-level**, where the evidence is: `_migrateDefaultInput` walks the
  node's own `input.required`/`input.optional` in two flat loops and does not recurse
  into a dynamic combo option's inputs, so a nested leaf keeps its widget and its slot.
  `isPositionalWidgetSpec` is therefore byte-identical to `main`; only its comment is new.

## Two behaviours that are deliberate, and tested as such

**A pre-unification row is reported, not normalized.** Such a row carried a
placeholder slot for each socket-only input, so it is one longer than the layout
mapped here. It cannot be normalized — review established that from both sides, each
reproduced by execution:

- normalizing on length (mirroring the frontend's `migrateWidgetsValues`) claims a
  *modern* row's frontend-only serialized extra (#1869) as a placeholder:
  `["D:/input.mov", "detect_range", 12]` → `path: "detect_range"`;
- yielding to the #1869 extras pass then blocks a *genuine* legacy strip:
  `[null, "D:/in.mov", 12]` → `path: null`.

Same length, opposite readings, nothing in the row to separate them. Per #1037's rule
in this same file, ambiguity stays a refusal. So no value is guessed — but the case is
no longer **silent**: a leftover on a node that has a socket-only input now names the
unmapped values, what they imply, and how to normalize the file. (It was silent for
an all-STRING/combo node, where the extras pass has nothing to refute.)

**A literal on a socket-only input is dropped in API→UI, with a warning.** UI format
has nowhere to put it — the canvas renders a socket. This is the converter's ordinary
treatment of a literal on a socket input, the same warning every other link input
gets. The alternative is what `main` does: write a placeholder into `widgets_values`,
which ComfyUI's own loader then strips — losing the value silently one step later, and
misaligning every widget in the meantime.

## Verification

- 11 new tests in `workflow-converter.test.ts` + 3 in `workflow-converter-api-to-ui.test.ts`.
- **Deleted the fix and watched them fail**, per mechanism: removing the
  `isWidgetInput` gate fails every forceInput test; removing the `defaultInput` clause
  fails the optional case; making that clause **symmetric** fails the required case, so
  the asymmetry is load-bearing and not incidental; disabling the leftover warning
  fails only the leftover test.
- Pinned: the reporter's exact node, the shipped rgthree node (two forceInput INTs,
  two-slot shift), a wired forceInput input, an INT forceInput, both `defaultInput`
  buckets, a nested combo leaf, both #1869 signals on a one-long row, and the two
  deliberate behaviours above.
- Exact reported symptom pinned as a negative assertion (the shifted values must not
  reappear), following the #193 precedent in the same file.
- Every review finding reproduced by **execution** against `dist/` before acting, and
  the two that were retractions were checked against the frontend bundle rather than
  against the claim.
- `npm test`: all 14 checks green (760 files, 14026 tests). `npm run lint`: clean,
  no new anti-slop findings.
- **No browser verification, and none applies**: server-side TypeScript in the MCP
  repo. Nothing the panel renders in the browser changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,8 @@ All notable changes to this project are documented here. This project adheres to
 - **the UI→API converter no longer shifts widgets past a `forceInput`-only input (#2753).**
   A `["STRING", {forceInput: true}]` input is a socket on canvas, so ComfyUI writes no
   `widgets_values` slot for it; the converter classified it as a widget, consumed the first
-  saved value, and reported every real widget with its neighbour's value. Legacy rows that
-  DO carry the placeholder are still read correctly, matching the frontend's own
-  length-based migration.
+  saved value, and reported every real widget with its neighbour's value — `get_workflow`
+  returned `unet_name` holding `clip_name`'s model, and so on down the row.
 
 ## [0.52.178] - 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ All notable changes to this project are documented here. This project adheres to
   `widgets_values` slot for it; the converter classified it as a widget, consumed the first
   saved value, and reported every real widget with its neighbour's value — `get_workflow`
   returned `unet_name` holding `clip_name`'s model, and so on down the row.
+  A workflow saved before ComfyUI's widget/input unification kept a placeholder slot for
+  such an input; that row is indistinguishable from one carrying an extra serialized
+  value, so it is no longer guessed at — it is reported, naming the unmapped values.
 
 ## [0.52.178] - 2026-09-02
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+### MCP
+
+#### Fixed
+- **the UI→API converter no longer shifts widgets past a `forceInput`-only input (#2753).**
+  A `["STRING", {forceInput: true}]` input is a socket on canvas, so ComfyUI writes no
+  `widgets_values` slot for it; the converter classified it as a widget, consumed the first
+  saved value, and reported every real widget with its neighbour's value. Legacy rows that
+  DO carry the placeholder are still read correctly, matching the frontend's own
+  length-based migration.
+
 ## [0.52.178] - 2026-09-02
 
 ### MCP
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ All notable changes to this project are documented here. This project adheres to
   A workflow saved before ComfyUI's widget/input unification kept a placeholder slot for
   such an input; that row is indistinguishable from one carrying an extra serialized
   value, so it is no longer guessed at — it is reported, naming the unmapped values.
+  The deprecated `defaultInput` spelling is honoured the way the frontend migrates it:
+  socket-only on an OPTIONAL input, widget kept on a REQUIRED one.
 
 ## [0.52.178] - 2026-09-02
 
--- a/src/__tests__/services/workflow-converter-api-to-ui.test.ts
+++ b/src/__tests__/services/workflow-converter-api-to-ui.test.ts
@@ -402,6 +402,40 @@ describe("convertApiToUi — forceInput-only scalar input (issue #2753)", () => 
     expect(slot?.widget).toBeUndefined();
   });
 
+  // An API graph MAY carry a literal on a forceInput input — ComfyUI's backend
+  // ignores forceInput, it is a frontend hint. UI format has nowhere to put it: the
+  // canvas renders a socket, so there is no widget and no widgets_values slot. This
+  // is therefore the converter's ordinary treatment of a literal on a socket input,
+  // the same warning every other link input gets, and it is DELIBERATE.
+  //
+  // The alternative is what this PR removed: writing a placeholder into
+  // widgets_values. That is not a carrier either — ComfyUI's own loader
+  // (migrateWidgetsValues) drops exactly that slot on load — it just loses the value
+  // silently, one step later, and misaligns every widget after it in the meantime.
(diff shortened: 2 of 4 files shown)


artokun:
**Review record — merging over a final NO-SHIP, on record.**

Eight gate rounds. The core production change (the `isWidgetInput` classification gate) earned a **SHIP** at `3f49449`; everything after that is additive or a retraction, and CI is green on every SHA including the head.

What each round produced:

| # | Finding | Outcome |
|---|---|---|
| 1 | length-based legacy normalization claims a modern row's #1869 extra (`path: "detect_range"`) | real — fixed by yielding to the extras pass |
| 2 | the yield then blocks a *genuine* legacy strip (`path: null`) | real — proved the normalization is a guess; **removed it** |
| 3 | — | **SHIP** |
| 4 | legacy rows still read shifted | restates the limitation round 2 established; but it was *silent* on all-STRING nodes — **now warned** |
| 5 | API→UI literal case untested | fair — **pinned as an explicit contract** |
| 6 | `defaultInput` not honoured | real and new — **fixed, asymmetrically**, per `_migrateDefaultInput` |
| 7 | nested combo leaves not excluded | checked the frontend rather than the claim: `_migrateDefaultInput` does not recurse. **Retracted** my own unsupported nested change instead |
| 8 | legacy rows still read shifted | rounds 2 and 4 again |

Round 8 restates what round 2 proved cannot be fixed: a legacy row and a modern row carrying a serialized extra have the same length and opposite readings, and both counterexamples were reproduced by execution. There is no code satisfying both. Its example `[1, null, 20]` against `{forced, steps}` matches *neither* layout (modern = 1 slot, legacy = 2), so there is no reading to converge on — which is why the code reports instead of guessing. Verified on the rig: it emits `steps: 1` **and warns**, naming `null, 20`.

The honest trade, stated plainly: on a **modern** row `main` is wrong and this is right — that is the reported bug, and it is what every current ComfyUI writes. On a **pre-unification** row `main` is accidentally right (via the very misclassification being fixed) and this is wrong, but **with a warning where `main` is silent**.

PR #2770, opened 2026-09-03
chore: keep the upload body and socket data typed under @types/node 26 (unblocks #2676)

artokun:
## Summary

`#2676` (dependabot: `@types/node` 22.19.11 → 26.4.0) is red on three type errors, and nothing about them is dependabot's fault — they are three places where our source only compiles under the older types. This fixes those three so the bump is a pure supply-chain decision rather than a code change bundled into a bot PR.

```
src/comfyui/client.ts:        Type 'Buffer<ArrayBufferLike>' is not assignable to type 'BlobPart'
src/comfyui/cloud-client.ts:  Type 'Buffer<ArrayBufferLike>' is not assignable to type 'BlobPart'
src/services/panel-launcher.ts: Type 'string | NonSharedBuffer' is not assignable to 'Uint8Array<ArrayBufferLike>'
```

`ArrayBufferLike` admits `SharedArrayBuffer`, which `BlobPart` excludes; and the `socket.on("data")` callback widened to `string | Buffer`, which `Buffer.concat` will not take.

## The part that is not just a type error

`new Uint8Array(data.buffer)` also silences the compiler and is **wrong**. `Buffer.from` returns a view into a shared 8KB pool, so that form uploads the whole pool with the caller's bytes somewhere inside it. Measured:

```
b = Buffer.from("world!!")   →  b.byteOffset = 8,  b.buffer.byteLength = 8192
new Blob([new Uint8Array(b.buffer)]).size            → 8192   ← the naive rewrite
new Blob([uploadBody(b)]).size                       → 7      ← this PR
```

`uploadBody` carries the offset and length, so it stays a **view** — no copy, which matters because these bodies are whole images and videos.

It lives in `upload-target.ts` for the reason that module's own header already gives: both upload clients need it, and `cloud-client.ts` cannot import from `client.ts` without a circular import. That header also records what happened the one time a rule reached only one twin.

## Verification

- `npx tsc --noEmit` clean under **both** `@types/node@22.19.11` (the current pin) and `@types/node@26.4.0` (what #2676 wants). The point of the change is that it is version-independent, so it can land before the bump.
- `npm run lint` → anti-slop gate, none added.
- Two new wiring tests in `upload-subfolder-wiring.test.ts` assert the bytes that actually leave `uploadImageHttp`, using a payload deliberately placed at a non-zero `byteOffset`.
- **Mutation-checked**: reverting `uploadBody` to the naive `new Uint8Array(data.buffer)` kills both — `expected 64 to be 5` and garbage bytes instead of `"world!!"`. The tests fail for the right reason, not incidentally.
- `probePanelOrchestrator`'s socket never calls `setEncoding`, so the annotated `chunk: Buffer` is the true runtime type, not a narrowing that hides a case.

## Follow-up

Once this is on main, #2676 rebases green and becomes a normal dependency review.


artokun:
Adversarial pass (mine — codex is out through Sept 6). **The one line here that could have been a serious bug is correct, and I can show what the obvious alternative would have done.**

`uploadBody` is the interesting part — a "chore: typing" PR that touches what bytes get uploaded:

```ts
return new Uint8Array(data.buffer as ArrayBuffer, data.byteOffset, data.byteLength);
```

The tempting shorter form — `new Uint8Array(data.buffer)` — compiles equally well and is wrong, because Node pools small `Buffer`s. Measured just now:

```
Buffer.from("AAAA")             byteOffset= 0   poolLen= 8192
Buffer.from("SECRET-ADJACENT")  byteOffset= 8   poolLen= 8192
Buffer.from("PNGDATA")          byteOffset=24   poolLen= 8192

new Uint8Array(c.buffer) → "AAAAV...SECRET-ADJACENT.PNGDATA.a byteOf"   (8192 bytes)
new Uint8Array(c.buffer, c.byteOffset, c.byteLength) → "PNGDATA"        (7 bytes)
```

So the naive version does not merely send the wrong length — it sends **the contents of every other Buffer sharing that pool**, straight into a multipart upload. On the upload path, that is adjacent heap leaving the process.

Passing `byteOffset` and `byteLength` is what makes it a real view, and the comment gets the *other* reason right too: `new Uint8Array(data)` would have satisfied the compiler by **copying** — a full duplicate of every image and video body. View, not copy, and not the pool.

## The rest

- **Runtime-neutral where it claims to be.** `client.ts` / `cloud-client.ts` change only what is handed to `new Blob([...])`, and `panel-launcher.ts` narrows a `socket.on("data")` callback that widened to `string | Buffer` — the same bytes either way.
- **The twin rule is respected**: `uploadBody` lives in `upload-target.ts` alongside `splitUploadTarget` because cloud-client cannot import from client.ts, so both callers get one implementation rather than two that drift.
- I verified earlier that this genuinely unblocks #2676: with these three fixed, the `@types/node` 26 bump is a supply-chain decision rather than a code change bundled into a bot PR.

Nothing blocking.

artokun:
**The Copilot review is right: the cloud twin had no byte-slice coverage of its own.**

Both clients call `uploadBody(data)`, so the fix is applied on both paths. But `cloud-client.ts` is wired independently and only the self-hosted client carried the two assertions — so the cloud path could revert to `new Uint8Array(data.buffer)`, which ignores `byteOffset` and `length` and uploads the whole 8 KB pool with the caller's bytes somewhere inside it, and **nothing would fail**.

A shared helper is not shared coverage. This file's own header already names that exact rot:

> This twin kept the original defect until now — the rot the header above warns about.

Added the same two assertions against the cloud form. Mutation-verified: reverting that one line to the backing buffer fails both new cases and nothing else. 505 tests green across `src/__tests__/comfyui`, `tsc` clean.

Issue #1477, opened 2026-08-12
download_model stages the full file in ~/.comfyui-mcp/cache on the system drive regardless of the destination volume — filled C: to 0.7 GB; progress counter also under-reported 10x (2.13 GB vs 22.62 GB on disk)


artokun:
### What happened / root cause

User asked me to install the missing models for a WAN 2.1 VACE workflow. Their ComfyUI lives on `F:\` (1 TB free); `C:\` had **0.7 GB free of 232 GB**.

`download_model action:"download"` with `target_subfolder:"diffusion_models"` streams the **entire** file into the content-addressed cache first:

`dist/services/download-cache.js:907`
```js
const DEFAULT_CACHE_DIR = join(homedir(), ".comfyui-mcp", "cache");   // line 16
function cacheDir() {
    return resolve(process.env.COMFYUI_DOWNLOAD_CACHE_DIR || DEFAULT_CACHE_DIR);
}
```

The cache dir is derived **only** from `homedir()` — it never considers which volume the resolved destination is on. On Windows `homedir()` is essentially always the system drive, while ComfyUI models are almost always on a big secondary drive. So a 32.29 GB download to a drive with 1 TB free was written to a drive with 0.7 GB free.

Two consequences:

1. **Guaranteed failure + system-drive exhaustion.** There is no free-space precheck on the cache volume even though `Content-Length` is known up front. The 32 GB `.partial` grew to 22.62 GB and would have driven `C:` to zero — on Windows that risks the page file and general OS stability, not just this download.
2. **The cached copy is retained after the model lands.** After two earlier successful downloads the models were correctly on `F:`, yet full duplicates remained on `C:`:
   - `2fcad3fb…safetensors` 0.24 GB (= `wan_2.1_vae.safetensors`)
   - `750ae65f…safetensors` 2.72 GB (= `lightx2v_I2V_14B_480p_cfg_step_distill_rank256_bf16.safetensors`)

   I understand retention is the intent of an LRU cache, but with `COMFYUI_LRU_CACHE_SIZE_GB` unset (`cacheSizeLimitBytes()` returns 0 ⇒ no eviction) the default behaviour is an unbounded copy of every model ever downloaded, on the system drive, with nothing surfacing that to the user. The user only discovered it via a disk-usage treemap.

### Separate bug — progress counter under-reports by ~10x

Same job, same moment:

- `download_model action:"status"` → `2.13/32.29 GB (6%)`, `still streaming — started 2508s ago`
[...]

PR #2771, opened 2026-09-03
fix(1477): report what the download cache is holding, and where

artokun:
Closes the reported half of #1477 — the half that was reopened on 2026-08-30 — and deliberately does not close the issue.

## What was wrong

The download cache is a second, content-addressed copy of every model ever downloaded. It lives under `homedir()` while the models land wherever ComfyUI keeps them, and with `COMFYUI_LRU_CACHE_SIZE_GB` unset `evictLruIfNeeded` returns on its first line — so the default is unbounded growth on the system drive.

Both reporters found it the same way, with a disk-usage treemap:

- 0.24 GB + 2.72 GB on `C:` after two models had correctly landed on `F:`
- 37.63 GB across 11 entries on `C:` while the models were on `D:` (comfyui-mcp 0.52.149, so this survived #1482)

Neither found it from anything this server printed, because nothing this server printed had ever named the directory. The first reporter's working fix — `COMFYUI_DOWNLOAD_CACHE_DIR` — was only discoverable by reading `download-cache.js`.

## What this changes

`download_model action:"status"` now ends with:

```
### Download cache

`C:\Users\...\.comfyui-mcp\cache` — 37.63 GB retained across 11 entries.
Eviction is OFF — COMFYUI_LRU_CACHE_SIZE_GB is unset, so nothing here is removed
automatically. Set it to bound the cache, or COMFYUI_DOWNLOAD_CACHE_DIR to stage on a
different volume. A retained entry is a COPY kept to avoid a re-download: once a model
has landed at its destination, deleting its cache file costs only that.
```

Silent when the cache is empty, so polling a download does not accrete a paragraph of unchanging text.

**Retained and staged are counted separately**, using `evictLruIfNeeded`'s own filter so the two cannot disagree about what is evictable. A `.partial` is dot-prefixed, which is precisely why eviction skips it — reporting it as retained would invite deleting the bytes a download is resuming from. Reporting the staged figure also happens to answer the report's other half: it is read from the filesystem, not from the in-memory progress row that the reporter measured diverging 10x from it (2.13 GB reported vs 22.62 GB on disk).

## What this deliberately does NOT do

**It does not set a default limit.** Turning eviction on by default deletes files for everyone, which is not a change to make on my own authority, and the right number is not obvious for a cache whose entries are routinely 32 GB. #1477 stays open for that decision, alongside the two other sub-reports it carries (relocating the cache to the destination volume, and the progress counter itself).

## Verification

- 9 tests, `npm run lint` clean, `check:changelog` passes.
- **Both call sites mutated independently.** The note reaches `action:"status"` through two exits — the empty-listing note and the `## Downloads` listing. Removing only the listing call left all eight original tests green, so that exit was untested; it now has its own case. Mutating each alone now kills a distinct set (1 test, then 2).
- Neighbouring suites green: `empty-listing-wiring`, `model-management`, `download-cache` — 140 passed.


artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6) found the buckets **did not account for the directory**.

Hidden files that are not `.partial` — the `.<name>.ct` change-tag and `.etag` sidecars this same module writes — fell through both branches:

- hidden, so not `retained`;
- not `.partial`, so not `staged`.

They were silently dropped. So `retained + staged` could not be reconciled against `du` — and reconciling them is the entire point of a footprint report. Someone chasing "the cache is eating my system drive" would find the numbers not adding up, which is exactly the position #1477's reporters were already in.

**The fixture had been writing a `.ct` sidecar the whole time** and nothing asserted on it. That is why nothing noticed.

Fixed with a third bucket, and the test asserts every **file** lands in exactly one — as a **sum**, not three independent numbers. Three separate assertions all pass while a category quietly goes missing; the sum is what catches it. Dropping the bucket fails it.

One claim I checked rather than trusted, since the report's credibility rests on it: the comment saying "counted with `evictLruIfNeeded`'s OWN filter so the two cannot disagree" is **true** — both are `entry.isFile() && !entry.name.startsWith(".")`.

`src/__tests__/tools` + `download-cache`: **103 files, 1691 tests**.

artokun:

One finding worth acting on, one I chased and dropped, and a note on the local suite.

### The report goes silent in the one case where it matters most — FIXED

`downloadCacheFootprint` collapsed two different states into a single `unreadable` flag, and the note rendered both the same way: silence.

```js
if (f.unreadable || (retained === 0 && staged === 0 && sidecar === 0)) return "";
```

- **ENOENT** — no cache yet. Silence is right; there is nothing to say.
- **EACCES / ENOTDIR** — the directory is *there* and could not be listed. Silence is wrong.

The second is exactly the user this PR exists for. #1477's premise is that *nothing this server printed had ever named the directory*, and both reporters found it with a disk-usage treemap instead. Someone whose cache directory is unreadable — wrong permissions, or `COMFYUI_DOWNLOAD_CACHE_DIR` pointed at a file — was sent straight back to that treemap by a feature written to spare them it.

The `readdir` catch discarded the errno, so the two were genuinely indistinguishable downstream. It now carries the code: ENOENT stays silent, anything else names the path and states the size as **UNKNOWN — not zero**. Reporting a confident `0.00 GB` for a cache that may hold tens of gigabytes would have been the worse failure, and this is the same shape as the defect I fixed on panel#2196 today: a helper that detects a condition and returns it to nobody.

Reached in tests **without mocks**, by pointing the cache at a file — `readdir` then fails `ENOTDIR` on every platform, and mis-setting the env var that way is a real thing to do.

### Dropped: the buckets do not miss any bytes

I went looking for an undercount, since the reporters' evidence is a *directory total* and `downloadCacheFootprint` skips non-files outright (`if (!entry.isFile()) continue;`) without walking subdirectories. If the cache could nest, the report would understate precisely the figure they came to check.

It cannot. Every path under the cache is flat — `join(cacheDir(), hash + extension)` and `join(cacheDir(), '.' + basename + '.partial')` — and the only `mkdir` is of the cache directory itself. So retained + staged + sidecar does account for every byte this server puts there, and the comment saying so is accurate. Not a finding.

### Verification

- `npm test`: every gate green except `vitest` and `asset-counts`, both pre-existing here and neither caused by this diff:
  - `asset-counts` fails identically on `origin/main` in this worktree — stale `dist/`, not a code fault.
  - Of 15 vitest failures, **two files pass in isolation** (`console-secrets`, `panel-secrets-slots`) — full-run concurrency artifacts. The three genuine ones already have PRs in flight: `install-accepted-never-enqueued` + `install-node-git-expectation` are what **#2829** makes hermetic, and `panel-image-relay` is **#2823**. Confirmed by reverting only this PR's two source files to `origin/main` and re-running: the failures persist.
- The suites this PR actually touches: **45 passed**, `npm run lint` clean (anti-slop: none added), `check:changelog` passes.
- Mutation-verified, each caught by the test written for it: every failure claiming `ENOENT`; the render going silent again on `unreadable`; and a `0.00 GB` leaking into the unreadable sentence.

### Still deliberately not done

Unchanged: no default eviction limit. Turning eviction on by default deletes files for everyone and the right number is not obvious for a cache whose entries are routinely 32 GB. #1477 stays open for that decision and its two other sub-reports.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **The property this report lives or dies on is right, and it is right by construction rather than by comment.** One narrow note.

## What I went after, and why it holds

A footprint the user acts on is only useful if the bytes it names are the bytes eviction would actually consider. Those were two separate copies of `!name.startsWith(".")`, and a comment asserting they agreed. They are now one function used by both:

```
downloadCacheFootprint()  -> isRetainedCacheEntry(entry.name)   (line 3616)
evictLruIfNeeded()        -> isRetainedCacheEntry(entry.name)   (line 3640)
```

so "retained" in the report and "evictable" in the evictor cannot drift. The comment at the site is honest about what changed — the old claim "was true only by coincidence" — which is the right way to describe a shared-by-construction fix.

**It reports the LIVE directory, not the default.** `downloadCacheFootprint` resolves through the same `cacheDir()` the download path uses (`join(cacheDir(), …)`, `mkdir(cacheDir())`), so a user who has set `COMFYUI_DOWNLOAD_CACHE_DIR` is pointed at the directory their bytes are actually in. Given the reported fix for both reporters WAS that variable, sending them to the wrong path would have been the worst possible failure of this feature.

**Unreadable is UNKNOWN, not zero** — same positive-evidence discipline as #2204. A permissions failure that printed "0.00 GB" would have actively confirmed the wrong belief.

**The third bucket is the good decision.** Computing sidecars and not printing them would leave exactly the unexplained remainder that made #1477 hard to diagnose; the comment says so, and the buckets are why the output reconciles against `du`.

## The note

The reconciliation invariant — "retained + staged + sidecar accounts for every FILE in the directory" — is exact for files, and the loop opens with `if (!entry.isFile()) continue;`. A `Dirent` that is a subdirectory *or a symlink* is therefore excluded silently, and `isFile()` is false for symlinks. The cache is flat and content-addressed so neither should occur, but the one scenario this feature exists for is a user comparing against a treemap, and that is precisely where a silent under-report would read as the tool being wrong.

Cheapest fix is not to account for them but to notice them: count non-file entries and, if any exist, say so in one clause. That keeps "the numbers reconcile" true rather than true-unless.

Nothing blocking.

artokun:
Fixed in `d53168d` — the note from my review is closed rather than left as a caveat.

`downloadCacheFootprint` now counts non-file directory entries, and the report discloses them:

> `…` — 37.63 GB retained across 11 entries, 1 non-file entry (subdirectory or symlink) NOT counted above, so these numbers will under-report what `du` shows

**Counted, never sized.** `stat` on a directory returns the inode, not the bytes beneath it, so a byte figure there would mislead more than no figure — and the point of the disclosure is to stop a silent remainder, not to invent a number for it.

Two tests, and the negative one matters as much as the positive:

- a subdirectory is counted as `nonFileEntries` **and leaks into no byte bucket** — asserted by comparing every byte total against a before-snapshot, since a directory whose inode size landed in `sidecarBytes` would be worse than not counting it at all;
- an ordinary flat cache reports **0**, so a normal install does not start claiming an unexplained remainder it does not have.

**Mutation-verified:** restoring the silent `if (!entry.isFile()) continue;` kills the subdirectory test and nothing else.

This keeps the invariant the whole report rests on — "the buckets reconcile against `du`" — true rather than true-unless, which is the property both reporters were implicitly relying on when they found the problem with a treemap in the first place.

**Verification:** `npm run lint` clean (anti-slop: none added), `check-changelog` OK, and the tools + download-wiring suites at **1,625 pass / 0 fail**.

artokun:
**Ran the footprint against a real 406 GB cache and cross-checked every number against an independent scan. Exact agreement.**

```
                    independent scan        downloadCacheFootprint()
retained            406.18 GB / 89          406.18 GB / 89
staged                1.27 GB / 2             1.27 GB / 2
sidecar               0.00 GB / 2             0.0000 GB / 2
non-file                     0                       0
eviction            UNSET                   limitBytes 0  (OFF)
unreadable          —                       false
dir                 —                       C:\Users\…\.comfyui-mcp\cache
```

`du -sh` on the same directory reports **408G**, against 407.45 GiB summed across the three buckets — the remainder is block-size rounding, not unaccounted files. So the reconciliation property this feature exists for **holds on a real cache two orders of magnitude larger than the one in the reports**.

That is the validation I most wanted: the reporters saw 0.24 + 2.72 GB and 37.63 GB, and the buckets were only ever exercised against fixtures. At 93 files and 406 GB, retained/staged/sidecar still account for everything on disk.

## What the run surfaced beyond the feature

The same measurement is why #1477 now has a much sharper picture, and it came out of exercising this PR rather than reading it:

- **406 GB is 22.7% of all used space** on a 96%-full volume — the reports' 37 GB was not the ceiling.
- **10 retained entries (35.3 GB) have no same-size counterpart** anywhere in the models tree, including a 32 GB checkpoint carrying `encrypted_wandb_properties`. So "just clear the cache" is not safe advice, which I had wrongly given on the issue and have corrected there.
- **`nonFileEntries: 0`** on a real directory, so the field I added reports the ordinary case correctly rather than only its mutation.

Nothing blocking, and the disclosure is doing exactly what it was built for — the numbers above are only knowable because this PR made them printable.

artokun:
One more correction to my own addition, in `2756999`.

The non-file disclosure said the numbers **will** under-report what `du` shows. That is not established: an empty subdirectory contributes nothing, so the buckets can still agree exactly. Changed to **may**.

Small, but it is the same overclaim the line beside it already avoids. That code deliberately counts non-file entries rather than sizing them, because `stat` on a directory reports the inode rather than the bytes beneath it — declining to state a number it cannot measure. Then the sentence asserted a discrepancy on the same unmeasured basis.

The disclosure's job is to stop a reader comparing against a treemap from being left with an unexplained remainder. It does not need to promise them one.

Found by applying to my own changes the rule I had been applying to everyone else's all day — `isModified === true` only, a failed read is not an empty inventory, unreadable is UNKNOWN not zero. Two of my own additions failed it: this, and a `conflict: true` on panel#2242 that claimed a conflict the code had explicitly failed to verify.

Both were mutation-verified and green when I pushed them. **A guard being correct and a report being truthful are different properties, and only the first had a test** — which is worth saying out loud, because every verification technique I used today checks the first one.

`tsc` clean, lint clean (anti-slop: none added), footprint tests 18/18, `check-changelog` OK.

artokun:
**Run against the real cache on this machine. The accounting is exact, and it settles the wording question I raised on this PR earlier.**

`downloadCacheFootprint()` pointed at the live directory (read-only — `readdir` + `stat`, nothing written):

```
dir            C:\Users\Artokun\.comfyui-mcp\cache
retained       89   436.14 GB
staged          2     1.36 GB
sidecar         2       163 B
nonFileEntries  0
TOTAL          93   437,500,181,606 bytes
limitBytes      0   (unset)
```

Independent ground truth from the filesystem:

```
ls -A | wc -l                              93
find -maxdepth 1 -type f | wc -l           93
find -maxdepth 1 -mindepth 1 -type d       0
du -sb                             437,500,181,606
sum of file sizes                  437,500,181,606
```

**Byte-for-byte identical, and every one of the 93 entries is classified** — 89 + 2 + 2 accounts for the directory with nothing left over. The shared `isRetainedCacheEntry` predicate is doing what the comment says: the reconciliation between eviction and the report holds on a real directory, not just in a fixture.

**On "may" vs "will".** I changed this PR's disclosure earlier today from "these numbers *will* under-report what `du` shows" to "*may*", on the argument that an empty subdirectory contributes nothing so the totals can still agree. This cache is the demonstration: `nonFileEntries: 0`, and the two figures agree **exactly**. "Will" would have been a false statement about this machine — the ordinary case, not a contrived one.

**And the #1477 substance, measured rather than estimated:** 437.5 GB retained with `limitBytes` **unset**, so nothing evicts. 1.36 GB of that is two `.partial` files — staged downloads that were interrupted and never resumed or cleaned. The report now makes both facts visible, which is what the issue asked for; what the default ceiling should be is still a decision I should not make, and I have not touched the cache.

artokun:
**The Copilot review on this PR was right about the hardlink, and it changes the #1477 number by an order of magnitude. Measured on the real cache on this machine.**

```
nlink=1   18 files    56.50 GB   only the cache holds these
nlink=2   75 files   381.00 GB   the SAME inode is also the installed model
```

`materializeCacheFile` prefers a hardlink when the cache and the destination share a volume, so a retained entry is frequently not a second copy at all. The note said it was:

> A retained entry is a COPY kept to avoid a re-download: once a model has landed at its destination, deleting its cache file costs only that.

Deleting those 75 files reclaims **nothing** — the blocks stay live under the model's name. The genuinely reclaimable figure here is **~56 GB, not ~437 GB**, on the one report whose whole purpose is to say what the cache is costing you.

**This also corrects my own comment above.** I reported "437.5 GB retained with `limitBytes` unset" as the #1477 substance. That total is right as a logical sum and wrong as an answer to "how much would I get back".

**Fixed:** the footprint now derives `sharedBytes` / `sharedEntries` from `st.nlink` — `stat` was already being called, so this costs nothing — and the note states the shared and cache-only halves separately. Byte totals are unchanged.

Production output against the live cache, reconciling exactly with an independent `find -printf '%n'`:

```
retained         89   436.14 GB
  of which shared 75   381.00 GB
  cache-only      14    55.14 GB
```

(The 18-vs-14 gap is the 2 `.partial` + 2 sidecar files, which are not "retained": 56.50 − 1.36 = 55.14. ✓)

**Two smaller findings from the same review, both confirmed and fixed:**

- `COMFYUI_DOWNLOAD_CACHE_DIR` was named only in the eviction-OFF branch, so a user with a bounded cache on the wrong volume never saw the relocation lever — the case where they most need it.
- "`COMFYUI_LRU_CACHE_SIZE_GB` is unset" is false for an explicit `0`, a negative, or a non-numeric value. `cacheSizeLimitBytes` tests `!Number.isFinite(raw) || raw <= 0`, so the text now says "not set to a positive number".

Tests: a real `link()` fixture asserting shared vs cache-only bytes, plus the single-link control. Mutation-verified — forcing the shared branch off fails exactly the new test. 20/20, `tsc` clean.

**Still open from that review, not addressed here:** segmented downloads write `.partial.seg` while truncating `.partial` to zero, so `stagedBytes` can read 0 while a large download is consuming space. This cache has no `.seg` files right now so I could not exercise it, and I would rather leave it visible than fix it blind.

artokun:
**Ran this against the real 406 GB cache on a Windows box, and cross-checked every number against an independent measurement. They agree exactly.**

`downloadCacheFootprint()` on this machine:

```
  dir             C:\Users\Artokun\.comfyui-mcp\cache
  retainedBytes   406.2 GB  (436137641085)
  retainedEntries 89
  stagedBytes       1.3 GB  (1362540358)
  stagedEntries     2
  sidecarBytes      163       sidecarEntries 2
  sharedBytes     354.8 GB  (380999407816)
  sharedEntries   75
  nonFileEntries  0          unreadable false
```

Independently, with `find -printf '%n %s'` and no reference to this code:

| quantity | independent | this PR | |
|---|---|---|---|
| hardlinked bytes | 354.8 GB | **354.8 GB** | exact |
| hardlinked files | 75 | **75** | exact |
| unique bytes | 51.4 GB | 51.4 GB (retained − shared) | exact |
| abandoned partials | 4 files / 1.3 GB | 2 staged + 2 sidecar / 1.3 GB | exact, and it separates sidecars |

Two different implementations — `find`'s `%n` versus this code's `fs.stat().nlink` — reading the same filesystem and landing on the same byte for byte.

**Why the `sharedBytes` field earns its place.** `du` says 408 GB. **87% of that is a second name for a file the models tree already holds**, because `materializeCacheFile` prefers a hardlink on a shared volume. Deleting this cache reclaims **52.6 GB, not 408** — and without this field the operator has no way to know that. Every "reclaim 400 GB" instruction we could print would be wrong by a factor of eight.

I have put the same numbers on comfyui-mcp#1477, because they also decide the shape of the LRU ceiling there: a limit expressed against retained bytes would spend its whole budget evicting entries whose removal frees nothing.

**Control:** `nlink` on NTFS through Git Bash is not something I assumed. A fresh file reports `nlink=1`; after `ln`, both names report `nlink=2`. The tool was verified before its output was trusted.

Issue #2742, opened 2026-09-02
#1524 variant: panel reports `connected` at SDK init but contributes ZERO panel_* tools — init health check compares server status, not tool count


artokun:
### What happened / root cause

Claude backend. The session came up with **no `panel_*` tools at all** — absent from both the direct tool declarations and the deferred/`ALL_TOOLS` catalog (I searched by exact name `mcp__panel__panel_graph_outline` per `deferred-panel-tools.ts` / #1398 guidance — genuinely not there, in either surface).

Everything the orchestrator checks, however, looked healthy:

- Orchestrator running: `comfyui-mcp connect`, v0.52.174.
- Bridge up: `ws://127.0.0.1:9199`, ComfyUI Desktop holding an ESTABLISHED connection.
- Panel tab connected and routing: `panel tab connected: wf:c50c2 ("pixel-tile-stage2-transition") — 1 tab(s) total`, then `tab wf:c50c2 -> agent: ...` for each of my messages.
- panel_* MCP server alive: listening on `127.0.0.1:9198` (same PID as the orchestrator), answers MCP JSON-RPC correctly (`{"error":{"code":-32000,"message":"No valid session id, and not an initialize request."}}` to a bare `tools/list` — i.e. a working endpoint).
- **The SDK itself reported `panel=connected` at init**, on every session.

So this is not "the panel server failed to start" and not "no tab connected". The server started, the SDK connected to it, and it contributed **no tools to the model**.

**The gap:** `inspectMcpServers(Object.keys(this.mcpServersForRun()), message.mcp_servers)` in `claude-backend.ts` compares configured server NAMES against reported connection STATUS. A server that reports `connected` while exposing zero tools satisfies that check, so the #1524 notice never fires and `checkMcpServers()` never opens a down-episode or calls `reconnectMcpServer("panel")`. The existing #1524 detection is structurally blind to exactly this variant — which matches the comment at `claude-backend.ts` ("WHY a server goes missing is still open on #1524"). Here it went missing *while reporting connected*.

The user-visible cost: they reinstalled to latest twice chasing this, because every signal the product surfaces said the panel was connected.

### Secondary observation (same root area, possibly causal)

`buildOptions()` sets `strictMcpConfig: true` with the comment "Only our comfyui MCP — never inherit the user's project/user MCP config". But `message.mcp_servers` at init reports **four** servers:

```
(reported: ida-pro-mcp=failed intellij=failed comfyui=connected panel=connected)
```

[...]

PR #2774, opened 2026-09-03
fix(2742): a connected MCP server that registered NO tools now says so

artokun:
Addresses the first and largest of #2742's three asks. It does **not** close the issue — the cause is still open, and the other two asks are untouched.

## What was blind

`inspectMcpServers` (from #1524) compares configured server **names** against reported connection **status**. The variant in this report passes that check cleanly:

```
[INFO] [panel-agent] init ... (reported: comfyui=connected panel=connected)
```

…while `panel_*` is absent from both the direct declarations and the deferred catalog. `connected` was being read as proof of a usable toolset. It is not — a server can connect and register nothing.

Five reproductions on 0.52.174 → **0.52.178 (current)**, the most recent today. The cost is the whole report: *"every signal the product surfaces said the panel was connected"*, and one reporter reinstalled twice chasing it.

## The evidence was already in the message

The SDK's `system`/`init` payload carries `tools: string[]` alongside `mcp_servers`, and an MCP tool appears there as `mcp__<server>__<tool>`. Nothing had ever read it. So the emptiness that the reporters established by hand with `ToolSearch` was sitting in the same message the status check already parses.

`serversWithoutTools()` now reads it, and a connected server with no entry under its own prefix raises a notice.

## Three deliberate silences

Each is this module's existing rule — a false "your tools are gone" on a healthy session is worse than the silence it replaces:

1. **No tool list, or an empty one** → nothing. Every existing #1524 fixture is that shape.
2. **A tool list carrying no `mcp__` name at all** → nothing. That is indistinguishable from a harness which does not list MCP tools here, and reporting every configured server would fire on every healthy session on such a harness. This is a **real blind spot** — if *every* server were empty the list would look identical — and it is accepted deliberately, documented on the function, because the reported signature has `mcp__comfyui__*` present throughout (three of the five reports say so explicitly).
3. **A degraded or pending server** → left to `inspectMcpServers`. It has no tools because it is not up, and two notices in different words read as two faults.

## The wording is not the degraded wording

`degradedMcpNotice` says the session *"started without"* the server, which is false here and would send the reader hunting a connection failure that did not happen — the distinction the entire report turns on. The new notice says `connected` **but contributed ZERO tools**, and names the recovery two reporters found by hand:

> Disconnect → Connect keeps the same orchestrator process, which is where the registration lives. Stopping and restarting the orchestrator process re-runs it.

## Scope — what this deliberately does NOT do

- **No reconnect is wired.** The turn-end watch (`checkMcpServers`) polls **status**, not tools, so a recovery attempt there could not be verified — and this module's whole discipline is to never assert an outcome nobody observed. Threading the init finding forward is a real follow-up, but it needs a decision about what to report when the result is unverifiable.
- **No `strictMcpConfig` investigation** (ask 2), and **no `hello`-target guard** (ask 3).
- **No cause.** It reports the observation, exactly as #1524's notice does.

## Verification

- 11 unit tests + 4 wiring tests; 67 passing across the three related suites; `npm run lint` clean; `check:changelog` passes.
- **Wiring mutated**: disabling the call site (`if (false && emptyServers.length)`) kills the wiring test and leaves all 41 unit tests green — the split the #1524 wiring suite's own header argues for.
- **Predicate mutated**: relaxing the prefix from `` `mcp__${name}__` `` to `` `mcp__${name}` `` kills the collision test, so `mcp__panel_extra__x` genuinely cannot mask an empty `panel`.
- Existing #1524 suites unchanged and green — their fixtures carry no `tools`, which is silence case 1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+### MCP
+
+#### Fixed
+- the empty-toolset notice now requires POSITIVE evidence that a server connected (#2742). It skipped servers known degraded or pending and named everything else, but "not degraded" is a different fact from "connected": inspectMcpServers classifies nothing at all when the report is absent or empty, and an unrecognised status is deliberately not an alarm. So one server's tools populating the list could get another server — with no usable status evidence — announced as having CONNECTED and contributed nothing. It now reads the reported status directly. Found by the Copilot review on the PR
+- **a `hello` frame carrying no `comfyui_url` is no longer reported as a dead instance (#2742).** The one verdict with no `base` fell through to the shared warning, rendering as `ignoring hello retarget to unreachable undefined (stale tab on a dead instance?)` — asserting a target was named AND found dead, about a frame that claimed neither. Seen 17 times in one report from tabs that were merely churning.
+- **an MCP server that connects and registers NO tools is now reported (#2742).** Once this process has seen ANY namespaced MCP tool -- proof the harness lists them -- a session with NONE reports every configured server, closing the blind spot where a failure that emptied EVERY server looked identical to a harness that does not namespace.
+
+
 ## [0.52.200] - 2026-09-07
 
 ### MCP
--- a/src/__tests__/orchestrator/claude-init-mcp-degraded.test.ts
+++ b/src/__tests__/orchestrator/claude-init-mcp-degraded.test.ts
@@ -436,3 +436,79 @@ describe("a forked resume uses THIS run's MCP set, not the session-recorded one 
     expect(hoisted.lastForkSession).toBeUndefined();
   });
 });
+
+// #2742, WIRING — same argument as the block above, for the variant that passes
+// the status check. A helper-only suite would pass with the call site deleted, and
+// the call site is the whole point: `serversWithoutTools` reads `message.tools`,
+// which nothing in `route()` had ever looked at.
+describe("a Claude session whose MCP server connected but registered nothing (#2742)", () => {
+  const withTools = (
+    mcp_servers: Array<{ name: string; status: string }>,
+    tools: string[],
+  ) => ({ ...initWith(mcp_servers), tools });
+
+  it("reports the panel server that connected and contributed zero tools", async () => {
+    const events = await drive(
+      { mcpServers: { comfyui: COMFYUI_SERVER }, panelServer: PANEL_SERVER },
+      withTools(
+        [
+          { name: "comfyui", status: "connected" },
+          { name: "panel", status: "connected" },
+        ],
+        // The reporters' exact shape: comfyui tools present throughout, panel_*
+        // absent from every surface.
+        ["Read", "Bash", "mcp__comfyui__generate_image", "mcp__comfyui__get_history"],
+      ),
+    );
+    const notices = noticesOf(events);
+    expect(notices).toHaveLength(1);
+    expect(notices[0].message).toContain("panel");
+    expect(notices[0].message).toContain("ZERO tools");
+    // Not the degraded wording — the server did not fail to connect.
+    expect(notices[0].message).not.toContain("started without");
+  });
+
+  it("says nothing when both servers contributed tools", async () => {
+    const events = await drive(
+      { mcpServers: { comfyui: COMFYUI_SERVER }, panelServer: PANEL_SERVER },
+      withTools(
+        [
+          { name: "comfyui", status: "connected" },
+          { name: "panel", status: "connected" },
+        ],
+        ["mcp__comfyui__generate_image", "mcp__panel__panel_graph_outline"],
+      ),
+    );
+    expect(noticesOf(events)).toHaveLength(0);
+  });
+
+  it("stays silent on an init message that carries no tool list at all", async () => {
+    // Every existing #1524 fixture is this shape. A new alarm on them would be a
+    // false positive on healthy sessions, which is the one outcome worse than the
+    // silence this replaces.
+    const events = await drive(
+      { mcpServers: { comfyui: COMFYUI_SERVER }, panelServer: PANEL_SERVER },
+      initWith([
+        { name: "comfyui", status: "connected" },
+        { name: "panel", status: "connected" },
+      ]),
+    );
+    expect(noticesOf(events)).toHaveLength(0);
+  });
+
+  it("reports the connection failure ONCE, not twice, when the server also failed", async () => {
+    const events = await drive(
+      { mcpServers: { comfyui: COMFYUI_SERVER }, panelServer: PANEL_SERVER },
+      withTools(
+        [
+          { name: "comfyui", status: "connected" },
+          { name: "panel", status: "failed" },
+        ],
+        ["mcp__comfyui__generate_image"],
+      ),
+    );
+    const notices = noticesOf(events);
+    expect(notices).toHaveLength(1);
+    expect(notices[0].message).toContain("started without");
+  });
+});
--- a/src/__tests__/orchestrator/mcp-session-health.test.ts
+++ b/src/__tests__/orchestrator/mcp-session-health.test.ts
@@ -23,9 +23,12 @@
 // gone" on a healthy session is worse than the silence it replaces, so every
 // ambiguity resolves to saying nothing.
 
-import { describe, expect, it } from "vitest";
+import { beforeEach, describe, expect, it } from "vitest";
 import {
   degradedMcpNotice,
+  emptyToolsMcpNotice,
+  resetMcpToolNamespacingLatchForTests,
+  serversWithoutTools,
   inspectMcpServers,
   reconnectableMcpStatus,
   recoveredMcpNotice,
@@ -79,6 +82,7 @@ describe("inspectMcpServers — what the session reports vs what it was given", 
   it("treats an unrecognized status as connected", () => {
     // The status vocabulary belongs to the CLI and can grow. A new value must
     // not become an alarm here; only the ones that DO mean unusable are alarms.
+
     expect(
       inspectMcpServers(CONFIGURED, [
         { name: "comfyui", status: "connected" },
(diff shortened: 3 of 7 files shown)


artokun:
Self-adversarial review (codex is out of tokens, so the pass is mine until Sept 6). It found that this fix **could not fire on the case it was written for**.

## The blind spot was documented — its justification was not durable

The first cut treated "a tool list with no `mcp__` name at all" as *a harness that does not list MCP tools here*, and stayed silent. The rationale said so explicitly, and accepted the cost:

> if EVERY server were empty the list would look identical — and it is accepted deliberately, because the reported signature has `mcp__comfyui__*` present throughout ("headless mcp__comfyui__* tools remained available" in three of the five reports).

That is evidence about the **earlier reports**, not a property of the failure. A **sixth report arrived today** against 0.52.183 describing zero panel tools and *not* saying whether comfyui tools survived. If they did not, that session sits squarely inside the blind spot and this fix would have stayed silent on it — the exact "tested but unreachable" shape.

## Closed with a latch, not a guess

One namespaced tool, **ever**, proves this harness lists them. After that, an empty namespaced set is a real observation and every configured server is reported. A harness that never namespaces never latches, so it is never falsely reported on. The detection can widen; it cannot narrow.

The latch is process state, so the suite resets it per case — without that, the first test to latch it silently changes the meaning of every test after it. (That is not hypothetical: adding the latch made an existing case fail for exactly that reason.)

Pinned both ways: removing the latch fails the all-empty case, and the un-latched silence is asserted on its own.

## One question that would settle it

@artokun — in the failing session, were `mcp__comfyui__*` tools present? If they were, the original code would have caught it; if they were not, this latch is what makes the difference.

Suites green: `mcp-session-health` + `claude-init-mcp-degraded`, 60 tests.

artokun:

Two checks came back clean, one turned up a hazard worth writing down, and the PR body is now understating its own fix.

### Clean: the prefix match is unambiguous

`serversWithoutTools` attributes with `t.startsWith(\`mcp__${name}__\`)` rather than splitting on `__`. A server whose own name contains `__` therefore cannot be mis-attributed, and neither can a tool whose name does. I went looking for that specifically because splitting is the obvious implementation and it would be wrong. Not a finding.

### Clean: the three silences are implemented as described

Degraded and pending servers are excluded before the emptiness test, so a down server cannot produce a second notice in different words. Absent and empty tool lists return early.

### The body understates the fix — silence #2 is no longer the blind spot it describes

The body says a tool list carrying no `mcp__` name at all is *"a **real blind spot** — if every server were empty the list would look identical — and it is accepted deliberately."*

That is the pre-latch behaviour. The code now latches on the first `mcp__` name ever seen, so a *later* empty namespaced set is a real observation and every empty server IS reported — which is precisely the "all servers empty" case the body says is undetectable, and there is a test asserting it. Worth correcting in the body, since it is what a reviewer reads to decide whether the newest #2742 report is covered: it is.

### The hazard: this check silently depends on tool search being OFF

The pinned SDK (0.3.240) defers MCP tool schemas behind tool search **by default** — from the `alwaysLoad` field, documented identically on every MCP server config type:

> *"When true, all tools from this server are always included in the prompt and never deferred behind tool search… **Default: tools are deferred when tool search is enabled.**"*

A deferred tool is not in the init message's `tools`, which is the only thing this check reads. Nothing in `src/` sets `alwaysLoad` or enables tool search today, so every server is treated alike and the list is populated — the check works as shipped, and this is not a live bug.

If that ever changes, the failure direction is decided entirely by the process-wide latch:

- **Never latched** → no `mcp__` names → silence. The check goes inert. Safe direction, and the first test pins it.
- **Already latched by an earlier non-deferred session in the same process** → an all-deferred session reads as *every server contributing zero tools*, and the notice fires on a healthy session. That is the exact false positive this module exists to avoid, and the second test currently asserts that behaviour as correct — which it is, under today's assumption.

So the latch is process state only because every session in the process shares one deferral mode. That is a real precondition and it was nowhere in the code. Pushed as a comment on the latch (no behaviour change, `tsc` clean, 42 tests green, anti-slop none added) so the next reader hits it before enabling tool search rather than after.

### Note for the stack

#2799 sits on this branch and reads `getContextUsage().mcp_tools` instead of `init.tools`. That path is likely *unaffected* by deferral — the SDK's context breakdown carries a row explicitly named `"MCP tools (deferred)"` with `kind: 'deferred'` — so the turn-end half may keep working in a configuration where this init-time half goes inert. I have not measured that, and am not claiming it; it is the first thing to check if tool search is ever turned on.

artokun:
Adversarial pass on my own PR, since codex is out of tokens. One finding, and it is about the premise rather than the code.

**The detector's evidence is weaker than the diagnosis it prints.**

`serversWithoutTools` reads `message.tools` — the *declared* list. A tool that is registered but **deferred behind tool search** is also absent from that list. So one observation ("no `mcp__panel__*` in `tools`") is produced by two different states: the server registered nothing, or it registered normally and its schemas were deferred.

The notice commits hard to the first: *"contributed ZERO tools"*, *"Stopping and restarting the orchestrator process re-runs it."* If the cause were deferral, that recovery is useless — a restart re-defers.

**Why this is not hypothetical.** The module's own hazard note says "nothing in this repo sets `alwaysLoad` or enables tool search, so every server is treated alike and the list is populated." The first half is true — I checked, `origin/main` sets `alwaysLoad` nowhere. The second half does not follow: **enabling tool search is not a repo-side switch.** There is no such option in the pinned SDK's types; deferral is decided upstream, and the docs only say tools "are deferred when tool search is enabled". So the repo not enabling it is not evidence it is off, and every server here is deferrable.

The latch helps only in one direction. If *all* servers defer, no `mcp__` name appears, the latch never sets, and the check goes silent — safe. But if the spawned `comfyui` stdio server loads while the in-process `panel` sdk server defers, then `mcp__comfyui__*` sets the latch and `panel` is reported as contributing zero tools **on a healthy session**. That asymmetry between stdio and in-process is exactly the split #2840 is built on.

**What keeps both PRs true.** #2840 notes the reported tools were absent from the deferred catalog *as well as* the declarations — so for the reported case the servers genuinely registered nothing and this PR fires correctly. The gap is only that the detector cannot tell the two apart, and nothing in our control guarantees which one it is looking at.

The clean fix is to make the premise true by construction rather than by assumption: set `alwaysLoad` on both servers — #2840 already does it for `panel`; `comfyui`'s stdio config takes the same flag. Then "absent from `tools`" means "not registered", the notice's wording is earned, and the latch stops carrying the whole argument.

I would not merge this ahead of that. It is not wrong today, but its correctness rests on an upstream default it does not control, and the failure direction is a false alarm on the exact P1 it exists to diagnose.

artokun:
Checked the two things the comment above rests on, so neither stays an inference.

**1. There is no repo-side switch for tool search.** I grepped the pinned SDK's whole option surface, not just for the word: `Options` has `allowedTools`, `disallowedTools`, `toolAliases`, `canUseTool` — and nothing that enables or disables deferral. `alwaysLoad` (per-server) is the only lever, and its own text says "Default: tools are deferred when tool search is enabled". So the decision is upstream, and "this repo does not enable it" cannot be turned into "it is off".

**2. The init message cannot disambiguate, structurally.** `SDKSystemMessage` (subtype `init`) carries exactly:

```ts
tools: string[];
mcp_servers: { name: string; status: string }[];
```

No deferred-tool list, no per-server tool count, nothing marking a name as withheld. `SDKDeferredToolUse` sounds relevant and is not — it is `{id, name, input}` for a pending tool *call*, not a catalog.

That rules out the obvious alternative fix. "Also check the deferred catalog before reporting" is not implementable from this message: the deferred names are simply not in it. So the ambiguity cannot be resolved by reading more carefully — only by removing it at the source, which is `alwaysLoad` on both servers.

It also sharpens what this PR should claim. With `alwaysLoad` set, "absent from `tools`" means "not registered" and the notice's wording is earned. Without it, the honest version of the message is weaker than what it currently says — it would have to describe the observation ("no tools reached the agent under this server's name") and stop short of asserting the cause or prescribing a restart, because a restart does not undo deferral.

artokun:
Correcting the provenance of a claim I made above, and re-verifying it properly.

I wrote that I had "grepped the pinned SDK's whole option surface" and found no way to enable or disable tool search. The conclusion was right; **the version I read was not the pinned one.** My checkout's `node_modules` holds `@anthropic-ai/claude-agent-sdk` **0.3.233**, while `origin/main` pins `^0.3.240` and the lockfile resolves 0.3.240 — my tree is five commits behind and was never reinstalled. So "the pinned SDK" was a claim about a different version's types.

Re-checked against 0.3.240, installed clean in a scratch directory:

- **no tool-search / deferral option exists.** Grepping the option surface for `toolSearch`, `tool_search`, `deferTools`, `deferLoading`, `toolLoading` as keys returns nothing.
- `alwaysLoad` is still the only lever, carrying the same "Default: tools are deferred when tool search is enabled".
- The keys 0.3.240 adds over 0.3.233 are unrelated to any of this — `modelSettings`, `classifierContext`, `autoContinueAtUsageLimit`, `syncClaudeAiSkills` and similar.

So the finding stands unchanged: deferral is decided upstream, this repo cannot turn it off, and "we do not enable it" is not evidence it is off. `SDKSystemMessage` still carries only `tools: string[]` and `mcp_servers: {name, status}`, with no way to tell a deferred tool from an absent one.

Flagging the method error because it is the more useful half: a worktree resolves `node_modules` by walking up to the main checkout, so **every version-sensitive check I run in a worktree reads whatever that checkout last installed**, not what the branch pins. That bit me twice today — here, and on #2676, where the same inheritance made a `@types/node` bump typecheck clean against the old types.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **The hazard I came to report is already documented, in more detail than I would have written. One of its stated preconditions is about to stop being true, though — by another open PR of mine.**

## What I went looking for

`connected` is not proof of a usable toolset, and reading `tools: string[]` for `mcp__<server>__*` is the right evidence. My concern was the interaction with tool-search deferral: a deferred tool is absent from the init `tools` array, so the check would read every deferred server as contributing nothing.

Lines 161–175 say precisely that, name `alwaysLoad`, and go further than I would have — the per-process latch also matters:

> already latched by an earlier non-deferred session in the SAME process -> an all-deferred session reads as EVERY server contributing zero tools

That is the nastier half and I would not have got to it from the diff.

## The precondition is changing

The comment rests on:

> Today nothing in this repo sets `alwaysLoad` or enables tool search, so every server is treated alike and the list is populated.

**#2840 sets `alwaysLoad`** — per-tool, on `panel_graph_outline` and `panel_canvas`. So if both land, that sentence is false.

The functional hazard does not fire yet: `alwaysLoad` only *opts out of* deferral, and deferral only happens when tool search is enabled, which nothing does. But the comment is what a future reader will trust when deciding whether this check is safe to rely on, and a stale precondition is exactly how the latch bug you describe gets shipped.

Two things follow, and both are cheap:

- If #2840 lands first, this comment wants one clause: `alwaysLoad` is now set on two probe tools, which does not enable deferral but does mean "nothing sets it" no longer holds.
- Worth noticing the accidental asymmetry it creates. With tool search on and both PRs in, the panel server would report **2** always-loaded tools and pass this check, while a fully-deferred stdio server would report **0** and be flagged. The panel — the server this issue is actually about — would be the one server the check could no longer see the emptiness of.

Nothing blocking. Flagging it because these two PRs cannot conflict textually and nothing else would surface the coupling.

artokun:
**Checked this against the real orchestrator log on a live install, and it shows the blindness concretely.**

`connect-orchestrator.log` holds 46 init records. Every one is this shape:

```
(reported: context7=connected discord-setup=connected blender=failed comfyui=connected panel=connected)
```

Two things stand out on real data:

**The existing line proves your point rather than the current check's.** `blender=failed` shows status detection works for a server that genuinely does not come up. But `panel=connected` appears in all 46, and the record carries **no tool information at all** — so if the #2742 state had occurred in any of those sessions, nothing in this log could distinguish it from a healthy one. The diagnostic is exactly as blind as the PR body says, and 46 samples of it are on disk.

**What this adds is the missing half**, and at the right level of detail:

```ts
`${emptyServers.join(", ")} CONNECTED but contributing zero tools ` +
`(tools=${message.tools?.length ?? 0}, mcp tools=${…})`
```

Naming the server *and* both counts is what makes a future report actionable. `panel contributed zero` alone would leave the reader unable to tell "the session had no tools at all" from "every other server contributed and this one did not" — which are different bugs, and the reporter's case is the second.

The `sessionNotice` yield matters as much as the log line: the reporter's cost was *"every signal the product surfaces said the panel was connected"*, and a log entry nobody reads would not have changed that. Surfacing it into the session is what closes the gap they actually hit.

Combined with the deferral hazard already documented at lines 161–175, and the cross-reference I added on #2840 (the first `alwaysLoad` in the repo), I have nothing outstanding on this PR.

artokun:
**Cross-PR note, from the other side: if this and #2840 both land, one sentence here becomes false.**

`src/orchestrator/mcp-session-health.ts:164` currently reads:

> Today nothing in this repo sets `alwaysLoad` or enables tool search, so every server is treated alike and the list is populated.

#2840 adds the repo's **first** `alwaysLoad` — `PROBE_ALWAYS_LOAD` in `panel-tools.ts`, two panel tools opted out of deferral. #2840's own comment already calls this out; I am recording it here so it is visible from the branch that owns the sentence.

**No behaviour changes.** `alwaysLoad` only opts out of deferral, and deferral happens only when tool search is enabled, which nothing does. Merging both in either order is safe today.

**What does change is the premise a future reader will rely on.** That paragraph is a HAZARD note whose job is to be read by whoever eventually turns tool search on. It tells them the ground state is uniform. Once #2840 lands it is not: the panel server would report its two always-loaded tools and pass, while a fully-deferred stdio server reports zero and trips the notice — so the very server #2742 is about becomes the one whose emptiness this check can no longer see. That is the failure direction this comment exists to warn about, and it would be describing a world that no longer exists.

**Suggested resolution, whichever merges second:** change "nothing in this repo sets `alwaysLoad`" to name the exception (`PROBE_ALWAYS_LOAD`, panel#291) and keep the rest. One line, no logic.

Flagging it because `git merge-tree` cannot: these two PRs touch different files and do not conflict, so nothing in CI or the merge queue will raise it.

artokun:
**The Copilot review was right, and what makes it interesting is that both halves of the cause are deliberate.**

`serversWithoutTools` skipped servers known degraded or pending and named everything else. But *not degraded* is not the same fact as *connected*:

- `inspectMcpServers` returns `{degraded: [], pending: []}` when the report is absent or empty — on purpose, because reading a harness's silence as failure would fire on every healthy session;
- an unrecognised status classifies as neither pending nor failed — also on purpose, and its own test says so: *"The status vocabulary belongs to the CLI and can grow. A new value must not become an alarm here."*

Each is correct in isolation. Together they make a server with **no usable status evidence** indistinguishable from a healthy one — so one server's tools populating the list could get *another* server announced as having CONNECTED and contributed nothing. That is the same overclaim this notice was written to replace: #2742 exists because `panel=connected` was read as proof of a usable toolset.

It now reads the reported status directly and names a server only when the report positively says `connected`.

**A design decision worth recording.** I first added `connected: string[]` to `McpSessionHealth`. That churned six unrelated expectations — every existing caller of that type asks *what is wrong*, and this is the one caller asking *what is proven right*. Backed it out and derived the set at the use site, which already receives `reported`. The public shape is unchanged and those six tests were never touched.

**Verification:** mutation-verified — restoring the not-degraded inference fails exactly the two new cases and nothing else. 63 tests green across the health and `claude-init-mcp-degraded` suites, `tsc` clean.

That is the last of the inline findings on this PR.

artokun:
**Adversarial pass on my own PR: there is one way to produce #2742's symptom that both checks here are structurally unable to see.**

Both take their expected set from the same expression:

```
claude-backend.ts:983   inspectMcpServers(  Object.keys(this.mcpServersForRun() ?? {}), message.mcp_servers)
claude-backend.ts:1009  serversWithoutTools(Object.keys(this.mcpServersForRun() ?? {}), message.mcp_servers, message.tools)
```

And that expression drops `panel` whenever the server was not constructed:

```
claude-backend.ts:618   private mcpServersForRun(): Options["mcpServers"] {
                          return {
                            ...this.deps.mcpServers,
                            ...(this.deps.panelServer ? { panel: this.deps.panelServer } : {}),
                          };
                        }
```

`panelServer` is `undefined`-able at two earlier points, neither of which logs:

```
panel-agent.ts:2754   panelServer: this.opts.makePanelServer?.(tabId),
index.ts:2774         makePanelServer: (key) =>
                        backendOf(key) === "claude" ? createPanelMcpServer(...) : undefined,
```

(`backendOf` splits on the last `::` and falls back to `defaultBackend` for a bare key — `index.ts:1860`.)

## Why that is a gap rather than a nitpick

This PR catches **connected-but-toolless**. It cannot catch **never-configured**, because a server that was never added to `mcpServersForRun()` is not in the expected set, so neither check has anything to assert about it. The session then reports *fully healthy* — the exact condition the PR exists to make impossible.

Both produce the reporter's symptom identically: no `panel_*` in the direct list, none in the deferred `ToolSearch` catalog, healthy bridge, working `mcp__comfyui__*`.

## What I have and have not shown

**Shown, by reading:** the coverage gap is real and both checks share it.

**Not shown:** that never-configured is what panel#291's reporter is actually hitting. They are on the Claude backend, so `backendOf(key)` should be `"claude"` and the server should be built. I have not reproduced it and I am not claiming this is the cause.

The reason to close it anyway is asymmetric cost. If never-configured *is* the cause, this PR merges, the session reports healthy, and panel#291 takes a **fifth** "fixed" cycle across two months — with the diagnostic that was supposed to end the guessing confirming the wrong thing.

## Cheapest close

Derive the expected set from what the backend *should* have, not from what was successfully constructed — for a claude-backend agent, `panel` is expected whether or not `panelServer` came back defined. A one-line alternative that is nearly as good: log at `index.ts:2774` when a claude key yields no panel server, so the never-configured case stops being silent.

I have not pushed either. Flagging before merge rather than after, since the fix is small now and the failure mode is another month of round-trips if it lands as-is.

Related reading on the same shape: comfyui-mcp-panel#291 (the spliced-quote correction I posted there today rules deferral out and points at this binding).

artokun:
**Retracting the coverage-gap comment above — I checked reachability after posting it, and the branch I described cannot be taken. This PR needs no change.**

I claimed the checks could not see a `panel` server that was never configured. The gap is real as *shape*, but it is unreachable, for three reasons I should have established before raising it:

**1. The constructor cannot return undefined.**

```
panel-tools.ts:29494  export function createPanelMcpServer(…): McpSdkServerConfigWithInstance
```

Non-optional. For a claude key the ternary at `index.ts:2775` always yields a server.

**2. `backendOf` cannot miss "claude" for a Claude session.**

```
index.ts:1771  const defaultBackend = KNOWN_BACKENDS.has(backendId) ? backendId : "claude";
index.ts:1862  return i >= 0 ? key.slice(i + AGENT_KEY_SEP.length) : defaultBackend;
```

A qualified key carries `::claude`; a bare key falls back to `defaultBackend`, itself defaulting to `"claude"`.

**3. For non-claude backends the exclusion is CORRECT, not a blind spot.**

```
index.ts:2767  // In-process live-graph MCP for CLAUDE keys only (codex/gemini drive the
               // canvas through the loopback HTTP MCP instead).
```

Codex and Gemini legitimately have no in-process panel server. Listing `panel` as expected for them would be a false alarm every session.

So: claude keys always configure `panel`, and both `inspectMcpServers` and `serversWithoutTools` do cover it. Non-claude keys correctly omit it. The conditional spread at `claude-backend.ts:618` is right as written.

**Net:** my earlier comment's "asymmetric cost" argument is void — it was conditioned on never-configured being a possible cause of panel#291, and it is not one. Nothing here should hold this PR.

I posted a finding before establishing that its mechanism was reachable. That is the wrong order, and it briefly put a hold on a green PR in your merge queue.

artokun:
**Reachability check: this fires on the tool-surface-policy case, and it arms itself in the same session that exhibits it.**

I raised and retracted a coverage worry on this PR earlier. Since then a concrete mechanism for comfyui-mcp-panel#291 turned up — `COMFYUI_MCP_TOOL_PRESET=safe` withholds every `panel_*` tool by glob, so the in-process server connects and registers nothing — so the question became whether this check would actually name that.

It would. Walking the arming condition:

```js
if (!Array.isArray(tools) || tools.length === 0) return [];
const namespaced = tools.filter((t) => t.startsWith("mcp__"));
if (namespaced.length > 0) harnessListsNamespacedMcpTools = true;
else if (!harnessListsNamespacedMcpTools) return [];
```

In that scenario the headless `comfyui` server is unaffected — the presets deny `MUTATING_TOOLS`/`EXECUTION_TOOLS`, not the read surface — so its `mcp__comfyui__*` tools are present in the very first session. That arms the flag immediately, on the same turn, without needing a prior healthy session. Then:

- `configured` contains `panel` (the claude lane always builds it — `createPanelMcpServer` has a non-optional return and `backendOf` cannot miss `"claude"`)
- `isConnected` contains `panel`, because the in-process server does connect; the reporters' own observation was `panel=connected`
- no `mcp__panel__*` tool exists, because the policy filtered them all out

→ `panel` lands in `empty`, and the notice fires.

So this is not merely compatible with that hypothesis — **it is the thing that would have distinguished it from a transport failure**, in the first session, with no round-trip. That is a concrete argument for merging it ahead of further investigation on comfyui-mcp-panel#291: the next occurrence names itself instead of needing `ToolSearch` by hand.

The one case it still cannot see is a server that was never *configured* — but I established that is unreachable for a Claude backend, which is why I retracted that earlier finding.

PR #2775, opened 2026-09-03
fix(2765): a catalogue regex no longer outranks the pack a node actually loaded from

### LINKED ISSUE #2765: extract_deps maps rgthree, Crystools and Krea2Edit nodes to unrelated Anomalous_Model_Browser
list_packs(action:"extract_deps") returns wrong custom-node ownership for two bundled expert workflows. It maps unrelated nodes to https://github.com/DemonGatanjieu/Anomalous_Model_Browser instead of their real packs, so the readiness result says an unrelated pack is missing.

### Steps to reproduce
1. list_packs action:"read_workflow" name:"qwen-image-edit-edit"
2. Pass that graph to list_packs action:"extract_deps".
3. Observe "Power Lora Loader (rgthree)" and "Switch latent [Crystools]" both mapped to Anomalous_Model_Browser.
4. Repeat with pack "krea2-identity-edit".
5. Observe "Krea2EditGroundedEncode" and "Krea2EditModelPatch" mapped to the same unrelated pack.

### Exact wrong output
- Power Lora Loader (rgthree) -> DemonGatanjieu/Anomalous_Model_Browser
- Switch latent [Crystools] -> DemonGatanjieu/Anomalous_Model_Browser
- Krea2EditGroundedEncode -> DemonGatanjieu/Anomalous_Model_Browser
- Krea2EditModelPatch -> DemonGatanjieu/Anomalous_Model_Browser

Expected owners:
- rgthree-comfy
- ComfyUI-Crystools
- lbouaraba/comfyui-krea2edit

### Fix
Applied locally: no. Upstream-only: yes for this running npm package/orchestrator session.
The dependency mapping resolver should prefer exact node-class ownership from live /object_info plus the bundled manifest before loose Manager mapping matches, and should refuse ambiguous matches rather than returning one unrelated repository for several distinct packs.

### Cost
This produced a false missing-dependency result during an approval-gated workflow preflight and could make the user install unrelated third-party code.

### Environment
Linux remote ComfyUI 0.33.0; NVIDIA RTX PRO 6000 Blackwell MIG 2g.48gb; comfyui-mcp 0.52.176; panel 0.15.150.

---

artokun:
Closes #2765.

`extract_deps` returned one unrelated repository as the owner of four distinct nodes, and the readiness result then reported *that* pack missing. The reporter's own summary of the cost: it *"could make the user install unrelated third-party code"*. Your ruling on the issue was that anything pointing at a non-canonical pack is hack territory.

## The confusion

Two different kinds of claim were being consulted as equals:

| claim | what it is |
|---|---|
| a catalogue entry's **class-name list** | a publisher naming its own classes |
| a catalogue entry's **`nodename_pattern`** | a regex anyone may publish about anyone |

`resolveFromMappings` tried exact first, then returned the **first** regex that matched, and for an installed node that regex was allowed to overwrite `/object_info`'s `python_module` — the pack whose code is actually loaded.

## Three changes

**1. For an installed node, `python_module` wins.** It is the strongest evidence there is: it names the directory the class was imported from. An *exact* catalogue entry may still supply the friendlier name, because that is the publisher naming its own class. A regex may not.

**2. A pattern match must be unanimous.** Two packs matching one class name is a question this cannot answer, and it was answering it by object iteration order.

**3. A pattern that mostly captures other entries' listed names is ignored.** Measured against the live `extension-node-map.json` — 40,656 known class names, 39 patterns — this is not theoretical:

```
/Hunyuan/    -> 188 matches, 182 belong to other packs
/Inspire$/   -> 101 matches, 101 of them Inspire Pack's        (published by "ComfyUI Connection Helper")
/PulidFlux/  ->  13 matches,   9 belong to other packs
/- Ostris$/  ->   6 matches,   6 belong to other packs
```

Judged lazily and only with **at least five matches** to judge on. That floor is doing real work in the other direction: rgthree's ` \(rgthree\)$` matches exactly one known name, and that name lost a first-writer race to another entry — refusing a good pattern on a majority of one is the over-correction this has to avoid.

## Result, end to end against the real 2.4 MB catalogue

```
                                   before          after (installed)     after (not installed)
Power Lora Loader (rgthree)   ->   [wrong pack]    rgthree-comfy         Comfy Nodes  (via PATTERN)
Switch latent [Crystools]     ->   [wrong pack]    ComfyUI-Crystools     Crystools    (via PATTERN)
Krea2EditGroundedEncode       ->   [wrong pack]    comfyui-krea2edit     comfyui-krea2edit
Krea2EditModelPatch           ->   [wrong pack]    comfyui-krea2edit     comfyui-krea2edit
```

And the guards, on real catalogue data: `HunyuanMysteryNode` → **unresolved** (over-broad pattern refused), `Widget_jru` → **unresolved** (two live claimants), `SomethingInspire` → **Inspire Pack**, `Thing - Ostris` → **ostris_nodes_comfyui**.

## Visibility, because `install_deps` acts on this

A regex-derived owner now reports `source: "manager_pattern"` and renders as `— matched by catalogue PATTERN`, with a closing line explaining that it is a heuristic published by whoever wrote the entry rather than a statement of ownership. Previously the output showed the pack and nothing about how it was reached, so a guess was indistinguishable from a fact.

I did **not** make `install_deps` refuse pattern-derived packs. Against the live catalogue that would block legitimate rgthree and Crystools installs, so it is a product call rather than a bug fix — say the word and it is a small follow-up.

## Verification

- 7 new tests, 89 passing across `workflow-deps` / `skills-access` / `anima-inpaint-lllite-class`; `npm run lint` clean; `check:changelog` passes.
- **Each guard mutated independently.** Restoring the pattern-over-`python_module` behaviour, removing the unanimity check, and disabling the over-broad filter each kill exactly one test.
- **One guard was removed because its mutation could not kill anything.** A `contested` flag (same regex published twice) survived, and the reason turned out to be that it was wrong, not untested: where both claimants survive, unanimity already refuses; where one is disqualified as over-broad, the survivor is the right answer and refusing threw a correct mapping away. Removing it is what makes `SomethingInspire` resolve to Inspire Pack instead of nothing.
- One pre-existing assertion updated: a pattern-derived, not-installed pack now reports `manager_pattern` instead of `manager_mappings`. Same pack, more honest label.


artokun:
⚠️ **Collision notice — please do not merge this until #2777 is resolved.** Two agents worked #2765 concurrently.

Timeline: the claim comment went up on #2765 at `05:47:16Z`; this PR's commits landed `05:51–05:53Z` and it opened at `05:54:21Z`. Its worktree (`.claude/worktrees/wt-2765`, the un-suffixed name) has been idle since `05:52:38Z`, so nobody is driving it. Both PRs touch `src/services/workflow-deps.ts` heavily, so merging both would conflict.

This is good work and it reached the same core diagnosis independently — `/object_info` outranks the catalogue for an installed node, a `nodename_pattern` is a heuristic rather than an ownership record, patterns must be unanimous, and an over-broad pattern is ignored. Two things here are better than what #2777 had:

- **the `MIN_JUDGED` reasoning.** "A majority of one is not evidence" correctly identifies that rgthree's pattern matches exactly one known name which lost the first-writer race — I hit that same artefact and mis-attributed it, and had to correct a claim because of it.
- **the CHANGELOG entry.** #2777 was missing one; I have adopted it there with attribution to this PR.

I am proposing #2777 lands instead, on coverage rather than volume. It went through three independent codex-gate rounds which surfaced eight P1s, three of which are defect classes this PR does not cover:

1. **Ambiguous EXACT claims.** Here an exact hit ends the search on first-writer, so when two repositories both list a class name the first enumerated wins silently. That is the reporter's explicit ask — *"should refuse ambiguous matches rather than returning one unrelated repository"* — and 3.86% of catalogue class names are in that shape.
2. **The install path.** `/getlist` is matched first-wins across `id`/`title`/`reference`, so two repositories publishing the same title collapse into whichever the catalogue lists first — and the other one gets queued. That is the last step before third-party code is installed.
3. **Two contradictory summaries.** An ambiguous- or unresolved-only graph still reports *"All node types are core/built-in ComfyUI nodes"* and *"No packs needed installation"* — the reassuring line, printed above the warning that contradicts it.

Also, `MIN_JUDGED = 5` disables the over-broad veto entirely against a catalogue with fewer than five judged names; #2777 closes that with catalogue-independent shape probes after the gate flagged the same hole in my version.

If #2777 fails to land, this PR is the right fallback and I would rather see it merged than see #2765 sit open. Closing it only once #2777 is in.

artokun:
Concurring with the collision notice: **#2777 should land instead of this.**

I checked the three gaps rather than taking them, and the first is worse than measured. Against the live catalogue, class names claimed exactly by 2+ packs are **1,833 of 40,656 (4.51%)**, not 3.86% — and `LoadImageWithMetadata` has **four** claimants. This PR's `exact` map keeps first-writer-wins, so it picks one of those four on object-enumeration order, silently, and that is the pack `install_deps` queues. I noticed `first writer wins` while building `owners`, used it only to judge patterns, and never followed it to the exact path.

The install-path (`/getlist` first-wins across id/title/reference) and contradictory-summary gaps I simply never looked at.

Detail and the one thing possibly still worth taking from here is on #2777. Leaving this open as the fallback; closing it once #2777 is in. Do not merge both — we each rewrote `buildMappingIndex` and `resolveFromMappings`.

artokun:
Status, after comparing this against #2777 properly rather than from memory.

**I had this recorded as "close once #2777 lands". That was wrong, and I am not closing it.**

The two PRs overlap on the core fix — both implement overbroad-pattern rejection, and #2777 does it more thoroughly (+1113 vs +306, with `patternIsBroad` keeping all 39 real catalogue patterns). That half of this PR is superseded and I am happy for #2777 to own it.

But this PR also changes `src/tools/skills-access.ts`, which #2777 does not touch: it marks each pattern-matched dependency as *"matched by catalogue PATTERN"* and adds a closing note that those came from a `nodename_pattern` regex rather than a pack listing the class name — *"a heuristic published by whoever wrote the catalogue entry, not a statement of ownership"*. Given `install_deps` acts on this list, and #2765 is precisely "four unrelated nodes came back owned by one pack", that disclosure is the part a user needs.

**The catch, which is why this needs a decision rather than a merge order.** #2777 folds pattern hits into `source: "manager_mappings"` — its union has no `manager_pattern`, and `resolveFromMappings` returns the identical `{ kind: "resolved" }` shape for an exact claim and a regex hit. So on top of #2777 the condition this disclosure keys on (`source === "manager_pattern"`) matches nothing and the warning silently never renders. Raised on #2777 with the specific lines.

**So the options are:**

1. **#2777 preserves the distinction** (a distinct source, or `matched_by: "exact" | "pattern"`), and this PR is then reduced to just the `skills-access.ts` disclosure — the two become independent and both merge in any order. My preference.
2. **#2777 lands as-is** and this PR closes, accepting that a legitimate regex match is presented indistinguishably from a publisher listing.

Not acting on either until you pick, because it is a call about what `extract_deps` promises, not a mechanical merge.

artokun:
## Two PRs fix #2765, and they conflict — settled against the live catalogue

#2775 and #2777 both rewrite `src/services/workflow-deps.ts` and `src/tools/skills-access.ts`. They cannot both land. They agree on the two structural fixes — `python_module` wins for an installed node, and an ambiguous claim is refused rather than resolved by enumeration order — and they differ on **one predicate**: when is a `nodename_pattern` too broad to trust?

| | rule |
|---|---|
| **#2775** `isOverBroad` | ≥5 matched known names **and more than half are foreign** |
| **#2777** `patternIsBroad` | captures names owned by **≥2 distinct foreign packs** |

Neither is obviously right by reading, so I ran both over the catalogue everyone has (`extension-node-map.json`, this box's live copy): **4,884 packs, 36,174 exactly-owned class names, 39 patterns.**

```
both reject: 5     only #2775 rejects: 1     only #2777 rejects: 0
```

**#2775's rejections are a strict superset.** Every pattern it rejects, with how much of its own catch it actually owns:

| pattern | published by | matched | foreign | owns | #2777 rejects? |
|---|---|---:|---:|---:|---|
| `.*` | `DemonGatanjieu/Anomalous_Model_Browser` | 36174 | 36174 | **0** | yes |
| `Smart` | `kjqwer/SmartSaveImage` | 191 | 189 | 2 | yes |
| `Hunyuan` | `exedesign/Hunyuan-3D-v3` | 185 | 179 | 6 | yes |
| `Inspire$` | `ltdrdata/comfyui-connection-helper` | 101 | 101 | **0** | yes |
| `PulidFlux` | `PaoloC68/ComfyUI-PuLID-Flux-Chroma` | 13 | 9 | 4 | yes |
| `- Ostris$` | `ostris/ComfyUI-FlexTools` | 6 | 6 | **0** | **no** |

Two things fall out of that table.

**The catch-all is the reported pack.** `.*` is published by `DemonGatanjieu/Anomalous_Model_Browser` — the repository #2765 says was named as the owner of `Power Lora Loader (rgthree)` and the three others. That is the root cause on disk, not an inference.

**The one case they disagree on is the reported bug's own shape.** `/- Ostris$/` is published by `ostris/ComfyUI-FlexTools`, matches six class names, and **all six belong to `ostris/ostris_nodes_comfyui`** — a different repository by the same author. FlexTools owns none of them:

```
"Batch Image Loader - Ostris"      owned by ostris/ostris_nodes_comfyui
"LLM Pipe Loader - Ostris"         owned by ostris/ostris_nodes_comfyui
"LLM Prompt Upsampling - Ostris"   owned by ostris/ostris_nodes_comfyui
"One Seed - Ostris"                owned by ostris/ostris_nodes_comfyui
"Save Image Direct - Ostris"       owned by ostris/ostris_nodes_comfyui
"Text Box - Ostris"                owned by ostris/ostris_nodes_comfyui
```

Because they all come from **one** foreign pack, `foreign.size > 1` is false, so #2777 honours the pattern and `extract_deps` names FlexTools as the owner of six nodes it does not ship — then offers it as the missing dependency to install. That is exactly what #2765 reports.

**And #2775 does not over-reject to get there.** Every pattern it drops is majority-foreign, three of the six own *zero* of what they match, and the names a rejected pattern legitimately covers (PulidFlux's 4, Hunyuan's 6) are **exactly listed** in the catalogue anyway, so they still resolve through the exact map. The pattern is only ever a fallback for names the catalogue does not name.

### Recommendation

Take **#2775's predicate**. It is strictly stronger on the real corpus, simpler to state, and has no measured cost.

#2777 should not be discarded wholesale — its dedicated ownership test file, the ambiguity measurement (1,833 of 40,656 class names claimed exactly by 2+ packs), and the note that Manager's `/customnode/getmappings` **appends unmapped installed class names onto every pack whose pattern matches**, laundering a loose pattern into an apparently-exact mapping, are all worth keeping. The cheapest path is: land #2775, then port #2777's tests and that laundering note onto it.

Whichever you pick, the other should close rather than merge — they will conflict, and a partial merge of one on top of the other leaves the predicate in a state neither author tested.

*Measured, not read: script and full output available on request; the numbers above come from this machine's live `extension-node-map.json`.*

artokun:
Adversarial pass (mine — codex is out through Sept 6). **The distinction this turns on is the right one, and finer than I expected to find.** One narrow residual.

## What I came to challenge

Manager was preferred over `python_module` for a reason — a friendly title beats a directory name — so my concern was that flipping precedence would trade a misattribution bug for a UX regression. It does not:

```ts
const named = hit?.via === "exact" ? hit.pack : null;
pack: named ?? pack,
```

An **exact** entry may still supply the friendlier name, because that is a publisher naming its own class. A **regex** may not, because it is a heuristic anyone can publish about anyone, and the loaded module is not. That is the correct place to cut, and the comment states it better than the diff does.

## The residual: an exact claimant is trusted for the NAME without being checked against the loaded pack

For an installed node we know, from `python_module`, which directory the code actually came from. `named` is taken from the catalogue's exact map without comparing it to that directory. `exact` is first-writer-wins, so if another repository exactly lists the same class name — a fork, or a re-export — its title is what gets reported for a node loaded from somewhere else.

That is narrower than the bug you fixed (it is a wrong *label*, not a wrong install target) and it is not a regression: `mappedPack ?? pack` had the same exposure plus the regex one. But #2765 is about misattribution, so it is worth knowing the exact-name path can still misattribute in the single-claimant case.

**#2777 covers the multi-claimant half** — several packs exactly claiming one name becomes `ambiguous` rather than a first-wins pick. What neither covers is one *foreign* pack claiming it alone. If that is worth closing, the check is cheap for installed nodes: compare `named` against the `python_module`-derived directory and prefer the loaded pack when they disagree, since that is the side with the stronger evidence — the same principle this PR already applies to regexes.

Nothing blocking. This is a strict improvement over what main does.

artokun:
**Quantifying the residual I raised above — I called it narrow and it is not.**

I wrote that an exact claimant is trusted for the NAME without being checked against the loaded pack, and judged it "narrower than the bug you fixed… a wrong *label*, not a wrong install target". Measured on 31 real workflows from a live install:

```
installed custom nodes in those workflows       134
  catalogue's first-wins pick DIFFERS from
  the directory the code actually loaded from    34   (30.6% of catalogued)
```

Examples, all exact-map hits with 2–4 claimants:

```
CLIPLoaderGGUF   loaded=comfyui-gguf       catalogue=comfyui-zlycoris
Florence2Run     loaded=comfyui-florence2  catalogue=comfyui-tbg-etur
ColorMatch       loaded=comfyui-kjnodes    catalogue=comfyui-swwan
```

Because every one is an **exact** hit, `named` is non-null for all of them and `pack: named ?? pack` keeps the foreign name. So this PR corrects none of the 34 — they are precisely the residual.

## Why `via === "exact"` cannot carry this weight

#2777's comment explains the mechanism, and it applies directly here:

> Manager's own `/customnode/getmappings` appends every installed-but-unmapped class name onto the class list of EVERY pack whose `nodename_pattern` matches it, so an over-broad pattern is laundered into an apparently-exact mapping before it reaches us.

So "exact" does not mean "a publisher naming its own class" — upstream has already merged pattern matches into the class lists. The distinction this PR cuts on is real in the raw catalogue and partly erased by the time we read it.

**Nothing here argues against merging this.** Stopping a regex from outranking the loaded module is right, and the friendly-name reasoning is sound. The correction is only to my own assessment of what remains afterwards: for an installed node, the loaded directory should win over an exact catalogue name too — which is what #2777 does by `continue`ing before the catalogue is consulted at all.

The practical consequence for the merge: whichever of these two lands second, the combined result needs #2777's installed-node path, not this one's. Since they conflict in `workflow-deps.ts` that reconciliation is manual, and this is the specific behaviour to preserve.

artokun:
**Answering both Copilot findings here, since one was raised on this PR and answered on the rival's thread.**

**1. Catalogue-supplied regex / ReDoS.** Measured against the live `extension-node-map.json` on this machine rather than argued:

```
packs 4,884   exact class names 36,174   nodename_pattern(s) 39
patterns with NESTED quantifiers (the ReDoS precondition)   0
longest class name (the input that would have to blow up)   73 chars
ALL 39 patterns x ALL 36,174 names                          0.18 s total
slowest single pattern                                      0.008 s
```

Catastrophic backtracking needs a nested-quantifier shape and an input long enough for the blowup to bite; neither holds. It is a real property of the code and not a reachable defect today. **It applies equally to #2777**, which compiles the same `nodename_pattern` with the same syntax-only `try/catch` — so it is not a discriminator between the two.

**2. The PATTERN marker is untested.** Correct, and it is specific to this PR's design: #2775 names a pack and surfaces the risk to the user (a per-node PATTERN marker plus a warning). I am not adding that coverage, because I am recommending this PR be closed in favour of #2777, and making a PR I want closed look more merge-ready is the wrong direction.

**The relevant comparison, since that recommendation should be checkable.** #2777 does not have this distinction at all — it declines to name a pack when the catalogue is ambiguous rather than naming one with a caveat — and its rendering IS covered at tool level:

```js
expect(out).toContain("### Ambiguous ownership (1)");
expect(out).toContain("claimed by: Anomalous_Model_Browser, comfyui-krea2edit");
expect(out).toContain("will not install them");
expect(out).not.toContain("All node types are core/built-in");
```

So on the specific axis this finding raises — is the user-facing safety distinction pinned by a test — #2777 is covered and this PR is not. That is one more measured input to the same recommendation, not a new argument.

The measured basis for preferring #2777 remains what it was: on 31 real workflows, 34 of 134 installed custom nodes get a catalogue pick that differs from what is loaded; #2777 resolves all 34 correctly and #2775 keeps the exact (possibly laundered) name on every one of them.

PR #2776, opened 2026-09-03
fix(2759): the picker's model survives a New chat

### LINKED ISSUE #2759: Model picker override silently lost on New chat / resume_session — reset() deletes modelByKey, tab falls back to global default

The panel model picker is set to **sonnet**, but the agent that actually spawns runs **claude-opus-5** (the orchestrator's global default). The dropdown keeps rendering the stale selection, so the user is silently billed/rate-limited against Opus while believing they are on Sonnet. Users on a Claude subscription burn their Opus limits with no visible indication.

`AgentManager.reset(tabId)` in `dist/orchestrator/panel-agent.js` unconditionally drops the picker override:

```js
// Drop this key's picker override so a provider switch (which reset()s the old
// key) can't carry the old provider's model/effort into the new backend's spawn.
this.modelByKey.delete(tabId);
this.effortByKey.delete(tabId);
```

The next spawn then reads `modelFor()`:

```js
modelFor(tabId) { return this.modelByKey.get(tabId) ?? this.model; }
```

…and falls back to `this.model`, the global default.

**The rationale comment appears stale.** The only two `manager.reset(key)` call sites in `dist/orchestrator/index.js` are:

- `index.js:5161` — `event.type === "new_session"` (New chat)
- `index.js:5263` — `event.type === "resume_session"` (switch to a historical chat)

Neither is a provider switch. The provider/key-change path is the key **migration** at `panel-agent.js:3162-3169`, which *preserves* the override by moving it `oldKey → newKey`. Additionally `modelByKey` is keyed by the composite key `tabId::backend` (documented at `panel-agent.js:2196-2202`), so a provider switch already reads a different key and structurally cannot inherit the old provider's model. The delete is therefore unnecessary for the case it cites, and only clobbers the same-provider case — contradicting that same doc block: *"A same-provider reconnect reuses the same key, so the user's pick persists."*

### Steps to reproduce

1. In the panel model picker, select a non-default model (e.g. `sonnet`; orchestrator default is `claude-opus-5`).
2. Confirm the agent spawns with it — log shows `panel-agent init model=claude-sonnet-5`.
3. Start a **New chat** (or switch to a historical chat via resume_session).
4. Send a message.
5. The new session spawns with `claude-opus-5`, while the dropdown still displays `sonnet`.

Observed in `~/.comfyui-mcp/launch-logs/connect-orchestrator.log`:

```
20:39:27  tab orchestr options: model=sonnet
20:47:20  panel-agent init model=claude-sonnet-5  session=13246224
20:54:35  panel-agent orchestr model → sonnet
20:56:53  panel-agent init model=claude-sonnet-5  session=13246224
20:57:28  tab orchestr reset — new session next message
20:57:45  spawning agent for tab orchestr (1 active)
20:57:48  panel-agent init model=claude-opus-5    session=6602f593   ← default, not sonnet
```

### Exact error (scrubbed)

No error is raised — this is a silent fallback. The only signal is the divergence between the picker UI and the `init model=` log line.

### Possibly related second symptom (NOT confirmed — flagging, not claiming)

In the same session the agent received **no `panel_*` tools at all**, so live-canvas edits were impossible. Note the tab/key mismatch in the log: messages arrive on `tab wf:02fcc` while the spawn is logged as `spawning agent for tab orchestr` / `panel-agent orchestr`. A preceding `ui-bridge: panel tab disconnected: wf:02fcc` is in the log too. Whether the post-reset respawn also loses the panel-tab binding that supplies `panel_*` is unverified; recorded here in case it shares the root cause. Do not treat as diagnosed.

### Fix

`applied locally: yes` &nbsp; `upstream-only: no`

Commented out the two deletes in `reset()`. Composite keying already provides the cross-provider isolation the comment was guarding, so removing them restores the documented same-provider behaviour. Worst case for a provider switch is a stale entry under the retired `tab::oldbackend` key, which is never read (the new spawn reads `tab::newbackend`) — a negligible map entry, not a correctness regression.

```diff
--- a/dist/orchestrator/panel-agent.js
+++ b/dist/orchestrator/panel-agent.js
@@ -3410,10 +3410,15 @@
         this.pendingRetargetFrom.delete(tabId); // #1429 — nothing queued, nothing to classify
-        // Drop this key's picker override so a provider switch (which reset()s the old
-        // key) can't carry the old provider's model/effort into the new backend's spawn.
-        this.modelByKey.delete(tabId);
-        this.effortByKey.delete(tabId);
+        // Keep this key's picker override across a reset. modelByKey is keyed by the
+        // COMPOSITE key `tabId::backend`, so a provider switch already reads a DIFFERENT
+        // key and cannot inherit the old provider's model; deleting here only clobbered
+        // the same-provider case (New chat / resume_session), which silently reverted the
+        // user's pick to the global default. Restores the documented intent at modelByKey:
+        // "A same-provider reconnect reuses the same key, so the user's pick persists."
+        // this.modelByKey.delete(tabId);
+        // this.effortByKey.delete(tabId);
         if (agent) {
```

**Suggested upstream shape** (cleaner than commenting out): if a provider switch genuinely does need the clear, give `reset()` an explicit opt-in — `reset(tabId, { clearPicker = false } = {})` — and pass `clearPicker: true` only from that path, leaving `new_session` / `resume_session` on the default. Also worth re-syncing the picker UI after a reset so the dropdown can never display a model the tab will not actually spawn with.

**Verification caveat, stated plainly:** this is an npx-cached published build with no source checkout or test suite present, so I could **not** run the Step-4 safety gate (`npm run build` / `npm test` / `npm run test:agent`). Verification here is limited to `node --check` (passes) plus the static call-site analysis above. The patch also lives in the long-lived orchestrator process, which only reloads on a **full process restart** — so it has not yet been observed working at runtime. Please re-verify against source before merging.

### Environment

- ComfyUI: 0.34.0
- **ComfyUI frontend: 1.51.9** (comfyui-frontend-package, installed == required)
- comfyui-workflow-templates 0.11.50 · comfy-kitchen 0.2.31
- GPU: NVIDIA GeForce RTX 4090 (24 GB VRAM), 128 GB RAM, CUDA 13.0, torch 2.10.0, python 3.13
- **comfyui-mcp: 0.52.178**
- **panel: 0.15.159**
- Backend: Claude · install path `~/AppData/Local/npm-cache/_npx/<hash>/node_modules/comfyui-mcp`

---
Filed from the ComfyUI panel by **claude** · model `claude-opus-5` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
Closes #2759.

The picker was set to **sonnet**; the agent that spawned after "New chat" ran **claude-opus-5**, the orchestrator's global default, and the dropdown went on rendering the stale selection. Nothing errored. The reporter's cost line is the one that matters: a user on a Claude subscription burns their **Opus** limits while believing they are on Sonnet, and the only visible signal is a divergence between the picker and an `init model=` log line.

## The delete never protected the case it named

```ts
// Drop this key's picker override so a provider switch (which reset()s the old
// key) can't carry the old provider's model/effort into the new backend's spawn.
this.modelByKey.delete(tabId);
this.effortByKey.delete(tabId);
```

Two independent reasons that rationale does not hold:

1. **The key already isolates providers.** `agentKeyFor()` is `SHARED_SESSION_SCOPE + SEP + backend` — the last segment *is* the backend. A provider switch reads a different key, so there is structurally nothing of the old provider's to inherit. The test immediately above the one I flipped has always asserted exactly this (`modelOverrideFor("tab-a::codex")` is `undefined` after a pick on `tab-a::claude`).
2. **`reset()` is never called on a provider switch.** Its only two callers are `new_session` (index.ts) and `resume_session`. The provider/key-change path is `rebindAgent`, which *preserves* the override by moving it `oldKey → newKey`.

So the line only ever hit the same-provider case — the one the key's own doc block promises persists:

> A same-provider reconnect reuses the same key, so the user's pick persists.

## The pin carried the same conflation

`options-ack-rid.test.ts` asserted the old behaviour under the title *"reset() drops the override (new chat / provider switch semantics)"*. Two cases pinned as one, of which only the provider half was ever true — so the defect was pinned, not the requirement. Flipped, with the reasoning in the test rather than just in the commit.

## Verification

- 4 new tests + the flipped pin; 23 passing across `model-pick-survives-reset` / `effort-carry` / `options-ack-rid` / `agent-identity-wiring`, and 169 more across the five other suites that exercise `reset()` (`in-place-replace-reset`, `panel-restart-cancel-truth`, `panel-restart-identity-reason`, `ask-answer-journal`, `claude-turn-registry`). `npm run lint` clean.
- **Mutated**: restoring the two `delete` lines fails 3 tests, the first with `expected 'claude-opus-5' to be 'claude-sonnet-5'` — the reported symptom, reproduced as an assertion.
- Provider isolation is pinned explicitly on the new file too, so removing the delete cannot quietly reintroduce the `model gpt-5.5 may not exist` spawn it was written to prevent.

## Not addressed here

The report flags a possible second symptom in the same session — no `panel_*` tools at all. That is #2742 and has its own fix in #2774; the reporter explicitly marked it "NOT confirmed", and nothing here touches it.


artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6).

**The claim this change rests on is true, and I checked it rather than restating it.** The stated reason it is safe to stop deleting the override is that the map is keyed by a composite whose last segment is the backend. Verified:

```ts
const agentKeyFor = (panelTabId: string): string =>
  SHARED_SESSION_SCOPE + AGENT_KEY_SEP + backendForTab(panelTabId);
```

No tab id appears in the key at all, and both `reset()` call sites pass `agentKeyFor(...)`. Two consequences, both worth stating: a provider switch reads a **different** key and structurally cannot inherit the old provider's pick — and the old delete was hitting the **right** entry, so it genuinely was dropping the user's model. This is the bug, not a red herring.

## One hazard closed while here

`reset()`'s parameter was named `tabId` while holding that composite key.

This repo's standing invariant is that sessions are orchestrator-scoped and **never** keyed on `tab_id`. A parameter named `tabId` is therefore an open invitation for a future caller to pass a real tab id — at which point every map operation inside that method silently misses and this fix quietly stops working, with nothing failing. Renamed to `agentKey`, with the reason recorded at the signature.

That is the same shape as the other findings in this pass: the value was right, the *name* told the next reader something false.

`model-pick-survives-reset` + `options-ack-rid`: **14 tests**. `src/__tests__/orchestrator`: 303 files pass — the 3 failures are the suites that probe the developer's real ComfyUI, addressed in #2829 and not present on this branch.

artokun:

No finding. Three things I went after, and what settled each — recorded because "I found nothing" is only worth anything if it says where it looked.

### The exhaustive claim holds

The argument for deleting the delete rests on *"`reset()` is never called on a provider switch — its only two callers are `new_session` and `resume_session`."* An exhaustive-callers claim is the kind that quietly goes stale, so I checked rather than took it:

```
src/orchestrator/index.ts:5726   inside  if (event.type === "new_session" && event.tab_id)
src/orchestrator/index.ts:5840   inside  if (event.type === "resume_session" && event.tab_id)
```

Those are the only two `manager.reset(key)` call sites in `src/`. The remaining `.reset(` hits are `heldNotice.reset()`, a different object. So the provider-switch path genuinely never reaches it, and `rebindAgent` moving `oldKey → newKey` is genuinely the only provider transition.

### The key really does isolate providers

`agentKeyFor()` ends in the backend segment, so a provider switch reads a different key and there is structurally nothing of the old provider's to inherit. The pre-existing test asserting `modelOverrideFor("tab-a::codex")` is `undefined` after a pick on `tab-a::claude` is real coverage of that, not a restatement.

### What I actually expected to find, and did not: no way back to the default

This is the one I thought would land. `reset()` was the de-facto clear for `modelByKey`, and the setter has no clear path of its own — it only assigns on `typeof next.model === "string" && next.model`, so null and empty are ignored. Its sibling `effort` *does* have one (`null` clears to the SDK default, documented on the line above). Removing the delete without adding a clear looked like it would pin a user to a model for the process lifetime.

It does not, because the panel never has a "default" to return to. `prefs.model` is always a concrete id: **Auto** resolves through `pickDefaultModel(modelCatalog)` and sends that, and the old `"default"` sentinel was migrated to `undefined` years ago (`if (prefs.model === "default") prefs.model = undefined`). Every reachable pick is a concrete model and any of them can be picked again, so there is no state the user can be stranded out of.

One thing that falls out of that and is worth someone's attention, though it is not this PR's business: for a user in **Auto**, the pick now persists across New chat as a concrete id. If Auto's resolved default later changes, Auto stays on the old model until the panel re-sends. That is panel-side behaviour, it is not a regression from this diff, and I have not measured it.

### The tests discriminate

Restoring the exact two lines this PR removes — `modelByKey.delete(agentKey)` and `effortByKey.delete(agentKey)` inside `reset()` — fails **3 of 14**. So the flipped assertion in `options-ack-rid.test.ts` is pinning the requirement now rather than the defect, which is the specific risk in flipping an existing pin.

Baseline 14 passed.

artokun:
Adversarial pass (mine — codex is out of tokens). The removal rests on two factual claims and I checked both rather than taking the comment's word.

**Claim 1 — a provider switch cannot inherit the old pick.** Verified:

```js
const agentKeyFor = (panelTabId: string): string =>
  SHARED_SESSION_SCOPE + AGENT_KEY_SEP + backendForTab(panelTabId);
```

The key's last segment *is* the backend, so a switch reads a different key. `modelByKey` structurally cannot carry `claude`'s pick into a `codex` spawn. The old delete was defending against something the key layout already prevents.

**Claim 2 — `reset()` has only two callers, neither a provider switch.** Verified: exactly two production call sites, `index.ts:5726` under `event.type === "new_session"` and `index.ts:5840` on the `session_id` resume path. Nothing else calls it. So the delete's only reachable effect was the same-provider case, which is the case the key's own doc comment promises persists.

So the fix is right and the reasoning holds.

**One wording point, because it could mislead the next change.** The comment closes with *"The model belongs to the tab and its provider"* — but the key has no tab component at all. `agentKeyFor` ignores `panelTabId` except to look up its backend, so the override is scoped to (shared session, provider) and is shared across every tab on that backend. That is consistent with sessions being orchestrator-scoped, so the behaviour is right; the sentence just describes a narrower scope than the code implements.

Worth correcting in place, because someone reading "belongs to the tab" while fixing a future bug could reasonably add a tab dimension to the key — which would silently split the override per tab and undo the shared-session invariant. "belongs to the provider, for the shared session" would say what the key actually does.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **This is a guard removal, which is the riskiest edit shape there is, and both of its load-bearing claims check out.** Verified rather than taken on the PR's word:

**1. The key's last segment IS the backend.** `reset()`'s own parameter doc says so (`agentKey`, not a tab id: `<shared scope>::<backend>`), so a provider switch reads a different key and structurally cannot inherit the old provider's model. Nothing to carry.

**2. `reset()` is not on the provider-switch path.** I enumerated every `.reset(` call site in `src/` — five total, three of which are `heldNotice.reset()`. The only two `manager.reset(key)` calls are:

```
index.ts:5721   if (event.type === "new_session" && event.tab_id)
index.ts:5828   if (event.type === "resume_session" && event.tab_id)
```

Both are deliberate conversation boundaries where the provider has not changed. So the deleted comment's stated rationale — "so a provider switch can't carry the old provider's model into the new backend's spawn" — described a path that does not reach this code.

**3. The path that DOES cross keys preserves the override, correctly.** `rebindAgent` migrates `modelByKey` from old to new, and I checked what it is for: a panel tab-id *scheme* migration (random UUID → `tmp:`/`wf:` prefixed), not a provider switch. Same conversation, same backend — carrying the pick is right, and dropping it there would have re-created this bug under a different trigger.

## The tests test the right thing

Removing a guard means the obligation is to prove the property it *claimed* to protect still holds, and that is exactly what is here: alongside "respawns on the PICKED model after reset()", there is "still isolates providers — a pick on one key is invisible to the other", asserting `modelOverrideFor("scope::claude")` is `undefined` after a pick elsewhere and that the spawn used `claude-opus-5`. Behavioural throughout — driving a real manager and asserting on what the backend was actually spawned with, not on source text.

## One residual, which I think is acceptable

`modelFor` is `this.modelByKey.get(tabId) ?? this.model` with no validation, so a picked model that later becomes invalid now survives New chat, where before it was incidentally cleared. I do not think that argues against the change: the picker is the intended way to change it and now shows the truth, and `modelByKey` is in-memory so an orchestrator restart self-heals. Worth knowing the incidental recovery is gone rather than discovering it later.

Nothing blocking.

artokun:
**One of the two Copilot findings was right; the other is already addressed on this branch.**

**Right: the effort half was untested.** The diff removes `this.effortByKey.delete(...)` from `reset()` alongside the model override, so two behaviours changed and the new suite only exercised one. Confirmed against the diff rather than the description:

```
git diff <base>..<head> -- src/orchestrator/panel-agent.ts | grep effortByKey
-    this.effortByKey.delete(tabId);
```

And `effortFor` reads through `Map.has()`, which makes an explicit return to the SDK default — stored as `undefined` but PRESENT — a third state, distinct from "never picked". Those two can regress independently, so they are pinned separately, along with the provider isolation that makes both safe.

Mutation-verified: restoring `effortByKey.delete(agentKey)` in `reset()` fails exactly the two new cases and nothing else.

**Already addressed: the "tab-owned" comment.** The review asks for the model not to be described as belonging to the tab. The comment on this branch already says the opposite, and says why the wording matters:

> The model belongs to the PROVIDER, for the shared session — `agentKeyFor` ignores the tab except to look up its backend… Said precisely because 'belongs to the tab' would invite adding a tab dimension to the key, which would split the override per tab and undo the shared-session invariant.

No change needed there.

**One process note.** My first two attempts at the mutation silently failed to apply — `reset()` takes `agentKey`, not `tabId`, and the string I was matching appears four times in the file. Both runs reported "7 passed", which I would have been entitled to read as "the tests are sound". They only proved the mutation never landed. The applied mutation is the one quoted above, and it fails.

PR #2777, opened 2026-09-03
fix(2765): extract_deps must not name a pack the catalogue does not prove owns the node

### LINKED ISSUE #2765: extract_deps maps rgthree, Crystools and Krea2Edit nodes to unrelated Anomalous_Model_Browser

artokun:
Part of #2765 — one of two defects; #2775 is the other and carries the closing reference. (Was `Fixes #2765`; re-worded at landing so the first merge does not close the issue while the second is still open.)

`list_packs(action:"extract_deps")` mapped `Power Lora Loader (rgthree)`, `Switch latent [Crystools]`, `Krea2EditGroundedEncode` and `Krea2EditModelPatch` to one unrelated repository (`DemonGatanjieu/Anomalous_Model_Browser`) and then reported that repository as the MISSING dependency of an approval-gated preflight — a result that could have had the user install unrelated third-party code.

Per @artokun on the issue: *"anything that points to a non canonical pack is hack territory, fix asap"*.

## Root cause — measured, not inferred

Diagnosed against the **live ComfyUI-Manager catalogue** (`extension-node-map.json`: 5,614 packs, 40,656 exactly-owned class names). The reporter's box was not needed; the defect reproduces from the catalogue everyone has.

`nodename_pattern` is treated as an ownership record. It is a naming convention:

| pattern | claimed by | class names it takes from OTHER packs |
|---|---|---|
| `Hunyuan` | `exedesign/Hunyuan…` | **182** |
| `Inspire$` | `ComfyUI Connection Helper` | **101** |
| `PulidFlux` | `PaoloC68/…` | 13 |
| `Inspire$` | `Inspire Pack` | 10 |

Three defects, each of which produces the reported shape on its own:

1. **A Manager match overrode `/object_info` for an INSTALLED node.** `python_module` is ground truth — the pack directory the class was actually imported from, on the very server being reported on — and it was discarded in favour of a loose catalogue match *"for a friendlier name"*. This is also the only defence against an amplification we cannot otherwise detect: Manager's own `/customnode/getmappings` [appends every installed-but-unmapped class name onto the class list of **every** pack whose `nodename_pattern` matches it](https://github.com/Comfy-Org/ComfyUI-Manager/blob/main/glob/manager_server.py), so an over-broad pattern reaches us already laundered into an apparently-*exact* mapping.
2. **Ambiguity was resolved by `Object.entries` order.** 1,833 of 40,656 class names (**4.51%**) are claimed exactly by 2+ packs, and `Inspire$`, `_jru$` and `- Ostris$` are each declared as a `nodename_pattern` by two different packs. Whichever enumerated first won, silently — and `install_deps` feeds that name straight to the Manager task API.
3. **A catch-all pattern owned everything unopposed.** With no rival to make it ambiguous, one `.*` entry claims every class_type the catalogue does not name.

## Fix

- **Installed node → `/object_info` wins.** Manager is not consulted for ownership when the server already told us where the class came from.
- **Not installed → Manager answers only when a SINGLE pack claims the class.** 2+ claimants (exact *or* pattern) ⇒ reported as `ambiguous` with every candidate named, kept out of `missingPacks`, and **never auto-installed**.
- **A pattern that matches a probe string no pack could own is dropped** — it is not a test for anything.

## The risky parts — please look here

- **Patterns are deliberately NOT anchored** to Manager's own `re.match` semantics, even though that is what upstream does. I implemented anchoring first and measured it: it disables *every* suffix-tag pattern — `` \(rgthree\)$ ``, `` \[Crystools\]$ ``, `` \(mtb\)$ ``, `` \(lab\)$ `` — i.e. precisely the well-formed ones this issue expects to keep working, while `Hunyuan` still over-claims 96 names. Upstream's own suffix patterns are inert under its own matcher. The probe rejects all catch-all forms and **none** of the 39 real patterns, so it is the cheaper side of the trade — but it is a judgement call against upstream behaviour.
- **Behaviour change:** an installed custom node is now named by its on-disk pack directory (`rgthree-comfy`) rather than the Manager title (`Comfy Nodes`). I believe that is the canonical identity and it matches the owners the report expects, but it changed four existing assertions — see `workflow-deps.test.ts`. Installed packs never contribute to `missingPacks`, so `install_deps` routing is unaffected.
- **`ambiguous` is a required field** on both result shapes rather than optional, so a caller cannot forget to surface it. That is why four existing test mocks needed updating.
- **Ambiguity refusal has a real blast radius:** 4.51% of catalogue class names have rival claimants. For *installed* nodes it costs nothing (rule 1 answers). For uninstalled ones it converts a silent arbitrary pick into a named list of candidates — deliberately, since the arbitrary pick is the reported harm.

## Evaluated and NOT done

The report also suggests consulting "the bundled manifest". I checked: `packs/qwen-image-edit-edit/manifest.yaml` lists **ten** `custom_nodes` repos for a workflow with three custom classes. Manifests declare dependencies at *pack* level, so they cannot attribute an individual class_type; inferring per-class ownership from them would add a second loose selector of exactly the kind that caused this bug. Left out on purpose.

## Verification

- `npm test` green: **14,091 tests**, 13 checks.
- **Fix deleted, tests watched to fail** — each guard mutated independently:
  - restore Manager-over-`object_info` precedence → `names rgthree-comfy and ComfyUI-Crystools, never the unrelated pack` FAILS
  - remove ambiguity refusal → 3 tests FAIL (extract, install-queues-nothing, pattern conflict)
  - remove probe rejection → catch-all test FAILS
  - remove the `extract_deps` ambiguity **rendering** → tool-layer wiring test FAILS
- Probe set re-measured by `eval`-ing the literal **out of the shipped source**, not a retyped copy: rejects `.*`, `.+`, `.`, `^`, `(?:)`, `a|.*`, `\w*`, `[\s\S]*`; rejects 0 of 39 real catalogue patterns.

Not run: no live ComfyUI was allocated for this run, so this was verified against catalogue data and unit/wiring tests rather than an end-to-end `extract_deps` call. No panel-visible surface changed, so the Playwright suite does not apply.


artokun:
## Review rounds (codex gate)

Three rounds run; **8 P1 findings accepted and fixed, 1 declined with reasons recorded at the call site.**

**Round 1** — 3 P1:
- Pattern claimants were deduplicated by display **title**, not repository. Two repos sharing a `title` collapsed into one "unambiguous" owner. The exact index already counted by catalogue key; patterns now do too.
- The junk-string probe misses a pattern shaped like a plausible class name (`^[A-Z][A-Za-z0-9_()]*$`). Added the veto the catalogue itself supports: a pattern capturing class names **exactly owned by 2+ other packs** is describing a topic, not an owner.
- An ambiguous-only workflow has zero `requiredPacks`, so the report opened with *"All node types are core/built-in"* — the reassuring reading, printed above the warning contradicting it.

**Round 2** — 1 P1:
- The foreign-owner veto is evidence *from* the catalogue, so against a catalogue with no exact names it returns "not broad" — absence of evidence read as evidence of narrowness, the same error `#1136`/`panel#890` guard against one endpoint over. Fixed by probing breadth of **shape**, which needs no catalogue: four probes shaped like real class names (PascalCase, snake_case, both tag conventions).

**Round 3** — 3 accepted, 1 declined:
- `/getlist` was matched first-wins across `id`/`title`/`reference`, so two unrelated repos sharing a title collapsed and the **wrong one got queued**. Last step before an install — same defect, one layer down.
- An installed node with no `python_module` that Manager cannot name is in neither summary list, so the all-built-in line fired with such a node in the graph. Now counted off the dependencies.
- `install_deps` still said *"No packs needed installation"* for an unresolved-only run.
- **Declined:** *"refuse a pattern-only resolution when the exact index is too thin for the veto to be a real test."* Not a reachable production state (the live catalogue carries 40,656 exactly-owned names), and it breaks pattern-only packs — **`rgthree-comfy` declares `` \(rgthree\)$ `` and ZERO exact class names**, so it would un-resolve the very pack this issue names as an expected owner. I implemented it and ran it: four tests went red, including the rgthree case. Reverted, with the reasoning recorded in the code.

### Verification

Every accepted guard was mutation-tested **independently** — removed, suite run, confirmed red:

| guard | test that goes red when it is deleted |
|---|---|
| `/object_info` precedence | `names rgthree-comfy and ComfyUI-Crystools, never the unrelated pack` |
| ambiguity refusal | 3 tests (extract, install-queues-nothing, pattern conflict) |
| shape probes | `refuses to own a class_type it matched with ".*"` (+ 7 more forms) |
| catalogue-thin probe | `still refuses a broad pattern when the catalogue is too thin to judge it` |
| broad-pattern veto | `refuses a pattern that looks like a class name but claims the ecosystem` |
| claimants by repo | `counts pattern claimants by REPOSITORY, not by display title` |
| colliding install key | `refuses to install through a catalogue key that names two repositories` |
| `unattributed` count | `does not call a graph all-built-in when a node has no attributable pack` |
| install headline | `does not say "no packs needed" when packs went unresolved` |

`npm test` green: **14,099 tests**, 14 checks. CI green on all four jobs for `ff7eeba`.

One caveat worth stating plainly: the first `npm test` on this SHA reported a vitest failure under concurrent load; an isolated `vitest run` and a second `npm test` were both fully green (764 files, 14,099 passed). That was load flake, not a product failure.

artokun:
## Correction to my own numbers — measured end-to-end against the real catalogue

I ran the **built** module against the live ComfyUI-Manager catalogue and the actual `qwen-image-edit-edit` workflow from the report. Two things came out of it, one of which corrects a claim I made above.

**My earlier measurements used the raw `extension-node-map.json`. Production does not receive that file.** `defaultWorkflowDepsDeps` asks `getmappings?mode=nickname`, and Manager applies `nickname_filter` server-side first — stripping every class name preempted by an explicit `preemptions` list or by a `*/ComfyUI` entry. Re-measuring through a port of that filter:

- catalogue ambiguity is **3.86%** (1,568 of 40,656), not the 4.51% I quoted from the raw file;
- and on the report's own workflow, with **nothing installed** — the worst case, where every class must fall through to the catalogue — the result is:

```
CLIPLoaderGGUF              -> https://github.com/city96/ComfyUI-GGUF
UnetLoaderGGUF              -> https://github.com/city96/ComfyUI-GGUF
Power Lora Loader (rgthree) -> Comfy Nodes          (rgthree-comfy's catalogue title)
Switch latent [Crystools]   -> Crystools
KSampler, LoadImage, SaveImage, VAEDecode, …
                            -> https://github.com/comfyanonymous/ComfyUI
ambiguous: []
```

**Zero refusals.** I had flagged above that the ambiguity refusal has a 4.5% blast radius and worried it might be over-aggressive for common packs like ComfyUI-GGUF. On the unfiltered file it was — `UnetLoaderGGUF` came back with four claimants. Through the filter production actually gets, it resolves cleanly to city96. The concern was an artifact of measuring the wrong input; the refusal does not fire on the workflow that caused this issue.

Runtime with the full 5,614-pack catalogue: **25 ms**, so the lazily-evaluated broad-pattern veto costs nothing measurable.

One thing I deliberately did **not** change: when rgthree-comfy is *not* installed it is still reported by its catalogue title `Comfy Nodes` rather than `rgthree-comfy`. That is Manager's own name for the pack and it is the key the install path matches on, so swapping it for the repo URL would touch the v4 `directInstall` payload for marginal display benefit. When the pack IS installed — the case in this report — it now reports `rgthree-comfy`, which is what the issue asks for.

artokun:
## 🛑 HOLD — review gate is INDETERMINATE, not passing

Gate round 6 on `7982ff8` returned **exit 2**:

```
```

An indeterminate gate is not a pass, so I am **not merging this**, even though every finding raised so far has been answered and the local suite is green.

### Where review actually stands

Five rounds completed, **11 P1 findings accepted and fixed, 2 declined with measurements recorded at the call site.**

| round | verdict | outcome |
|---|---|---|
| 1 | NO-SHIP | 3 P1 fixed — claimants by title not repo; probe misses class-name-shaped patterns; ambiguous-only graph reported "all built-in" |
| 2 | NO-SHIP | 1 P1 fixed — foreign-owner veto is vacuous against a thin catalogue; added catalogue-independent shape probes |
| 3 | NO-SHIP | 3 P1 fixed (install-path title collision; `unattributed` miscount; unresolved-only install headline), 1 declined |
| 4 | NO-SHIP | 3 P1 fixed — breadth scan read a first-owner map; candidates collapsed to title; ambiguous-but-installed told to install |
| 5 | NO-SHIP | 1 P1 fixed (pattern-branch candidates still by title), 1 declined |
| 6 | **INDETERMINATE** | quota exhausted — **nothing reviewed** |

**What is unreviewed:** the round-5 fix (`4ce8f9e`) and the `origin/main` merge (`7982ff8`). Rounds 1–5 reviewed earlier SHAs; a verdict names a SHA and does not carry forward.

### What would unblock this

1. **Codex quota restored** → re-run and merge on a clean verdict:
   ```
   node .claude/autopilot/codex-gate.mjs --worktree C:/Users/Artokun/code/comfyui-mcp/.claude/worktrees/wt-2765-cc --base origin/main --name 2765
   ```
2. **Or a human review** of the two unreviewed commits. They are small and the risky parts are named in the PR body and in the round-by-round comment above.

### Evidence that does exist, independent of the gate

- `npm test` green: **14,106 passing**, 14 checks.
- Every one of the 12 guards mutation-tested **individually** — deleted, suite run, confirmed red. Listed in the comment above.
- End-to-end through the **built** module against the live 5,614-pack catalogue with Manager's `nickname_filter` applied: the report's own workflow resolves to `city96/ComfyUI-GGUF`, `Crystools`, `rgthree-comfy` and ComfyUI core, **zero ambiguous refusals**, 18 ms.
- The one red test in full-suite runs (`panel-image-relay.test.ts`) is **not mine**: I checked out pristine `origin/main` in the same worktree and it fails there identically.

None of that substitutes for an independent review, which is the point of the hold.

**#2775 stays open** as the fallback. If this PR cannot be reviewed, that one is a smaller, self-consistent fix for the same issue and I would rather it merged than see #2765 sit open — see my note there for the three defect classes it does not cover.

artokun:
Author of the colliding #2775 here. I checked your three gap claims rather than taking them, and the first one is **worse than you measured** — so: concur, this should land instead of mine.

## Independently re-measured, against the live `extension-node-map.json`

You put ambiguous exact claims at 3.86%. Counting every class name claimed by two or more packs:

```
class names in the catalogue:      40656
claimed EXACTLY by 2+ packs:        1833  = 4.51%
```

And the shape is worse than the number:

```
LoadImageWithMetadata  ->  4 repositories
SaveImageWithMetadata  ->  3
AILab_QwenVL_GGUF      ->  4
```

My resolver keeps `if (!exact.has(cn)) exact.set(cn, pack)` — first writer wins, silently, on **object enumeration order**. For a not-installed node that is the pack `install_deps` then queues. So on `LoadImageWithMetadata` mine picks one of four repositories with no signal that it was a coin flip, which is the reporter's stated cost ("could make the user install unrelated third-party code") reached by a different route than the one I fixed. I saw `first writer wins` while writing `owners`, used it only to judge patterns, and did not follow it to the exact path. That is a real miss.

I also confirmed #2777 actually refuses rather than only reporting: `source: "ambiguous"`, a dedicated `ambiguous[]` on the result, and tests covering the multi-claimant case.

## Your other two gaps, granted

**The install path.** I never looked at `/getlist`. Matching first-wins across `id`/`title`/`reference` is the last step before third-party code is fetched, and a shared title collapsing two repositories there is the same defect one layer down from where I stopped.

**The contradictory summary.** Also never checked — "All node types are core/built-in" printed above the warning that contradicts it is worse than either line alone, because the reassuring one is the one a reader stops at.

## On `MIN_JUDGED = 5`

Your criticism is right and I want to be precise about what I got wrong. The floor is doing real work — rgthree's ` \(rgthree\)$` matches exactly one known name, and that name lost the first-writer race, so a bare majority rule refuses a good pattern on a bookkeeping artefact. What I did not consider is that the *same* floor disables the veto entirely against a small or partially-loaded catalogue, which is precisely when a greedy pattern has least competition. Catalogue-independent shape probes are the better answer; a floor tuned on a 40k-name catalogue has no meaning on a 4-name one.

## What is worth keeping from mine

You have already taken the `MIN_JUDGED` reasoning and the CHANGELOG entry, which is the right split. One more that may be worth a look if it is not already covered: I dropped a `contested` flag (same regex published by two packs) **because its own mutation could not kill anything** — where both claimants survive, unanimity already refuses; where one is disqualified as over-broad, the survivor is the right answer and refusing throws a correct mapping away. Against the live catalogue that is the difference between `SomethingInspire` resolving to Inspire Pack and resolving to nothing. If #2777 has an equivalent flag, it is worth checking whether it is load-bearing or just untested.

#2775 stays open as the fallback you suggested and I will close it once this lands. Nothing in it should be merged alongside — we both rewrote `buildMappingIndex` and `resolveFromMappings`.

artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6). This PR is the one I want to keep for #2765 — the overbroad-pattern rejection is more thorough than mine in #2775. One finding, and it is on the half this PR is deliberately keeping.

## The exact/pattern distinction is computed and then discarded

`resolveFromMappings` learns exactly how a pack was matched:

```ts
const exact = index.exact.get(classType);
if (exact) return { kind: "resolved", pack: exact };      // the publisher LISTED this class
…
return { kind: "resolved", pack: [...hits.values()][0] };  // a nodename_pattern REGEX matched it
```

Both return the same shape, and `fromMapping` renders both as `source: "manager_mappings"`. So one line after learning it, the answer forgets whether it was a listing or a guess.

That matters for this issue specifically. #2765's complaint is *"four unrelated nodes came back owned by one pack they have nothing to do with"*, and `install_deps` acts on this list. Rejecting catch-alls stops the worst case — but a legitimate, non-broad `nodename_pattern` is still a heuristic published by whoever wrote the catalogue entry, not a statement of ownership, and the reader cannot tell it apart from a pack that named the class.

**Concretely:** #2775 built a user-facing disclosure on exactly that distinction (`source === "manager_pattern"` → *"matched by catalogue PATTERN … confirm the pack is the one you expect before installing it"*). On top of this PR that condition matches nothing, so the warning silently never renders. That is the same producer/consumer shape as several other findings in this pass — except here the producer discards its own knowledge.

**Suggested:** keep a distinct source (or a `matched_by: "exact" | "pattern"` alongside `manager_mappings`) so the disclosure has something to key on. The cost is one field; the benefit is that "we guessed" stops looking identical to "the publisher said so".

I have not changed this PR — it is the codex-authored one and the call on its shape is yours. Flagging it because the follow-up in #2775 depends on it.

artokun:
## Two PRs fix #2765, and they conflict — settled against the live catalogue

#2775 and #2777 both rewrite `src/services/workflow-deps.ts` and `src/tools/skills-access.ts`. They cannot both land. They agree on the two structural fixes — `python_module` wins for an installed node, and an ambiguous claim is refused rather than resolved by enumeration order — and they differ on **one predicate**: when is a `nodename_pattern` too broad to trust?

| | rule |
|---|---|
| **#2775** `isOverBroad` | ≥5 matched known names **and more than half are foreign** |
| **#2777** `patternIsBroad` | captures names owned by **≥2 distinct foreign packs** |

Neither is obviously right by reading, so I ran both over the catalogue everyone has (`extension-node-map.json`, this box's live copy): **4,884 packs, 36,174 exactly-owned class names, 39 patterns.**

```
both reject: 5     only #2775 rejects: 1     only #2777 rejects: 0
```

**#2775's rejections are a strict superset.** Every pattern it rejects, with how much of its own catch it actually owns:

| pattern | published by | matched | foreign | owns | #2777 rejects? |
|---|---|---:|---:|---:|---|
| `.*` | `DemonGatanjieu/Anomalous_Model_Browser` | 36174 | 36174 | **0** | yes |
| `Smart` | `kjqwer/SmartSaveImage` | 191 | 189 | 2 | yes |
| `Hunyuan` | `exedesign/Hunyuan-3D-v3` | 185 | 179 | 6 | yes |
| `Inspire$` | `ltdrdata/comfyui-connection-helper` | 101 | 101 | **0** | yes |
| `PulidFlux` | `PaoloC68/ComfyUI-PuLID-Flux-Chroma` | 13 | 9 | 4 | yes |
| `- Ostris$` | `ostris/ComfyUI-FlexTools` | 6 | 6 | **0** | **no** |

Two things fall out of that table.

**The catch-all is the reported pack.** `.*` is published by `DemonGatanjieu/Anomalous_Model_Browser` — the repository #2765 says was named as the owner of `Power Lora Loader (rgthree)` and the three others. That is the root cause on disk, not an inference.

**The one case they disagree on is the reported bug's own shape.** `/- Ostris$/` is published by `ostris/ComfyUI-FlexTools`, matches six class names, and **all six belong to `ostris/ostris_nodes_comfyui`** — a different repository by the same author. FlexTools owns none of them:

```
"Batch Image Loader - Ostris"      owned by ostris/ostris_nodes_comfyui
"LLM Pipe Loader - Ostris"         owned by ostris/ostris_nodes_comfyui
"LLM Prompt Upsampling - Ostris"   owned by ostris/ostris_nodes_comfyui
"One Seed - Ostris"                owned by ostris/ostris_nodes_comfyui
"Save Image Direct - Ostris"       owned by ostris/ostris_nodes_comfyui
"Text Box - Ostris"                owned by ostris/ostris_nodes_comfyui
```

Because they all come from **one** foreign pack, `foreign.size > 1` is false, so #2777 honours the pattern and `extract_deps` names FlexTools as the owner of six nodes it does not ship — then offers it as the missing dependency to install. That is exactly what #2765 reports.

**And #2775 does not over-reject to get there.** Every pattern it drops is majority-foreign, three of the six own *zero* of what they match, and the names a rejected pattern legitimately covers (PulidFlux's 4, Hunyuan's 6) are **exactly listed** in the catalogue anyway, so they still resolve through the exact map. The pattern is only ever a fallback for names the catalogue does not name.

### Recommendation

Take **#2775's predicate**. It is strictly stronger on the real corpus, simpler to state, and has no measured cost.

#2777 should not be discarded wholesale — its dedicated ownership test file, the ambiguity measurement (1,833 of 40,656 class names claimed exactly by 2+ packs), and the note that Manager's `/customnode/getmappings` **appends unmapped installed class names onto every pack whose pattern matches**, laundering a loose pattern into an apparently-exact mapping, are all worth keeping. The cheapest path is: land #2775, then port #2777's tests and that laundering note onto it.

Whichever you pick, the other should close rather than merge — they will conflict, and a partial merge of one on top of the other leaves the predicate in a state neither author tested.

*Measured, not read: script and full output available on request; the numbers above come from this machine's live `extension-node-map.json`.*

artokun:
Adversarial pass (mine — codex is out through Sept 6). **The over-refusal risk I raised on #2765 is the one thing I most expected to be under-tested here, and it is bracketed on both sides.**

## Why I went looking

I measured the population this PR defends against: **1,837 of 40,656 catalogue names (4.52%) are claimed by more than one repository**, 383 by three or more. Refusing on ambiguity is right, but at that density an over-eager rule would turn a working `extract_deps` into one that answers "ambiguous" constantly. The tolerance that prevents it is one constant:

```ts
if (foreign.size > 1) return true;   // a pattern is BROAD
```

A single test can only pin one side of a threshold. This suite pins both — verified by mutation:

| mutation | kills |
|---|---|
| `foreign.size > 0` (zero tolerance) | *"tolerates ONE colliding owner, which is a fork of the same project"* |
| `foreign.size > 2` (looser) | *"counts EVERY repository that lists a captured name, not just the first"* |

So the constant cannot drift in either direction unnoticed. That is the coverage shape a tuned threshold needs, and it is the one most often missing.

## The rest holds up

- **Behavioural, not textual**: zero `readFileSync` in 513 lines. 23 tests, all driving the real resolver.
- **Wired to the surface**: *"install_deps queues nothing for it"* reaches the tool, not just the resolver — an ambiguity computed but not honoured downstream would be the obvious way for this to be right and useless.
- **The identity trap is covered twice**: *"names the REPOSITORY when two claimants share a display title"* and *"counts pattern claimants by REPOSITORY, not by display title"*. Two repos can carry the same `title`, and deduping on it would merge them and hide the conflict this PR exists to surface.
- **Non-regression**: *"keeps the well-formed SUFFIX-TAG patterns working"* and *"does not invent a conflict when one entry lists a name twice"* — both the false-positive direction.

Nothing blocking. Sequencing against #2775 is the only open question, and that is on #2765.

artokun:
**Went through the Copilot review on this PR properly. Both of its findings are already fixed here; a third, raised on #2775, applies to this PR too and measures as not reachable.**

**1. "Ambiguous candidates collapse to a title set, so two repos sharing a title expose one candidate."** Fixed, twice, and the code says so:

```
line 553  "Candidates are the catalogue KEYS, not the display titles ... two repositories
           can publish the same `title` — collapsing to it hands back ["Same"] and no way
           to choose (codex gate round 4)."
line 595  "Repository KEYS, exactly as the exact-claim branch above returns. Round 4 fixed
           that branch and left this one deduplicating on the display title (round 5)."
```

Both the exact-claim and pattern branches now return `[...keys].sort()`.

**2. "An all-ambiguous workflow still prints 'All node types are core/built-in'."** Also fixed. The sentence is guarded by a count taken from the dependencies rather than from the summary lists:

```js
const unattributed = result.dependencies.filter((d) => !d.builtin && !d.pack).length;
```

and an ambiguous dependency can only reach that state with `pack` null — the loop is `if (dep.pack) … else if (dep.source === "ambiguous")`, so ambiguity and a resolved pack are mutually exclusive by construction. Verified end to end, not inferred from the comment.

**3. The ReDoS concern raised on #2775 applies equally here** — this PR also compiles `nodename_pattern` with a syntax-only `try/catch` and runs it across the owner index. Measured against the live catalogue on this machine rather than argued:

```
packs 4,884   exact class names 36,174   nodename_pattern(s) 39
patterns with NESTED quantifiers (the ReDoS precondition)   0
longest class name (the input that would have to blow up)   73 chars
ALL 39 patterns x ALL 36,174 names                          0.18 s total
slowest single pattern                                      0.008 s
```

Catastrophic backtracking needs a nested-quantifier shape and an input long enough for the blowup to bite. Neither holds: no live pattern has the shape, and class names top out at 73 characters. The existing `OWNERSHIP_PROBES` filter already rejects catch-alls on top of that.

So it is a real property of the code and not a reachable defect today. Worth knowing rather than fixing under a stabilization freeze — and if it is ever fixed, the cheap version is a length cap on the tested string, not a worker.

**Process note.** I had been treating the review path as dark since codex ran out of tokens, and did not notice Copilot was reviewing every PR. That was wrong, and on #2809 it cost something real: its review had found that the dump reader ruled OUT the crash it exists to recognise, and that finding sat unread for a day.

artokun:
**Ran this branch against the REAL `extension-node-map.json` (4 884 packs) with the four class types from comfyui-mcp#2765. It passes the discriminating test.**

```
  Power Lora Loader (rgthree)    pack="Comfy Nodes"  source="manager_mappings"
  Switch latent [Crystools]      pack="Crystools"    source="manager_mappings"
  Krea2EditGroundedEncode        pack=null           source="unresolved"
  Krea2EditModelPatch            pack=null           source="unresolved"
  requiredPacks: ["Comfy Nodes","Crystools"]
  ambiguous: []
```

**`DemonGatanjieu/Anomalous_Model_Browser` is gone from every one of them.** That entry is `[[], {"nodename_pattern": ".*"}]` — zero exact claims, a regex matching every possible node name — and it no longer absorbs anything.

**And the harder half holds too.** I argued on the issue that the real test is not "does it stop naming the wrong pack" but "does it still name the right one", because a fix that only suppresses the bad answer leaves `rgthree-comfy` unnamed when the catalogue plainly knows it. It does still name it:

```json
"https://github.com/rgthree/rgthree-comfy":
  { "title": "Comfy Nodes", "title_aux": "rgthree's ComfyUI Nodes",
    "nodename_pattern": " \(rgthree\)$", ... }
```

`"Comfy Nodes"` **is** rgthree-comfy — that string is its own `title` field. I misread it as a second wrong answer at first and had to chase it down; recording that here so the next reader does not repeat it.

## One real finding, small

**`title` is a poor thing to show a user who is deciding whether to install.** `"Comfy Nodes"` is generic to the point of uselessness — five packs in this catalogue carry that phrase in their metadata — while the same entry offers `title_aux: "rgthree's ComfyUI Nodes"`, and the key itself is the unambiguous repo URL. A reader asked to approve an install of "Comfy Nodes" cannot tell which project that is.

## One observation, not a defect

`Power Lora Loader (rgthree)` resolves with `source: "manager_mappings"`, but it was matched by `nodename_pattern`, not by an exact claim — there are **zero** exact claimants for that string anywhere in the catalogue. So the source label does not distinguish a publisher listing a node by name from a regex that happened to match. comfyui-mcp#2775 adds exactly that distinction (`manager_pattern`, plus an over-broad heuristic). Whether that is worth having is your call, but the two PRs are not redundant in the way I previously suggested.

The Krea2Edit nodes coming back `unresolved` is correct: the catalogue genuinely does not know them, by exact name or by any pattern.

Issue #2693, opened 2026-09-01
restart_comfyui loses portable launch args and crashes under CP949


artokun:
Problem
On Windows portable ComfyUI, restart_comfyui stop/start did not preserve the original launch command or UTF-8 environment. The restart reported a reproducible inherited launch, but relaunched only ComfyUI\main.py and ComfyUI exited during custom-node startup.

Steps to reproduce
1. Start portable ComfyUI with embedded Python: python.exe -s ComfyUI\main.py --windows-standalone-build.
2. Call restart_comfyui action stop, then action start.
3. Observe the relaunched argv omits -s and --windows-standalone-build.
4. Startup output is decoded/encoded through Windows CP949 and rgthree-comfy's emoji print crashes startup.

Exact error
UnicodeEncodeError: 'cp949' codec can't encode character '\U0001f389'
The failure occurred while rgthree-comfy printed its startup banner.

Expected
Preserve the complete executable argv, including Python flags and ComfyUI arguments, and preserve a UTF-8 Python/console environment on Windows. Do not report reproducible:true when launch arguments or encoding environment are incomplete.

Verified workaround
Relaunching the same embedded Python with -s ComfyUI\main.py --windows-standalone-build plus PYTHONUTF8=1 and PYTHONIOENCODING=utf-8 starts successfully.

Environment
Windows 11; RTX 5080 16 GB; ComfyUI 0.34.2; frontend 1.49.6; Python 3.13.12; torch 2.12.1+cu130; comfyui-mcp 0.52.165; panel 0.15.135. User paths redacted.

---
Filed from the ComfyUI panel by **codex** · model _unreported_ — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #776

artokun:
Additional reproduction on Windows 11 / portable ComfyUI 0.33.2 (RTX 4090D, embedded Python 3.12.10): with comfyui-mcp 0.52.103 and panel 0.15.81, `panel_restart_comfyui` restarted the correct local instance but reduced the observed command from `ComfyUI\main.py --windows-standalone-build --lowvram --preview-method none` to `ComfyUI\main.py --lowvram --preview-method none`. The portable `run_nvidia_gpu.bat` had meanwhile been updated to `--windows-standalone-build --disable-pinned-memory --fp16-intermediates --preview-method none`, but the Manager restart did not use those settings. No exception; applying the desired flags required a manual stop and relaunch.

PR #2792, opened 2026-09-03
fix(2693): a relaunched ComfyUI gets UTF-8 stdio on Windows

artokun:
Fixes the crash half of #2693. **Does not close it** — the argv half is untouched and described below.

## What killed the server

```
UnicodeEncodeError: 'cp949' codec can't encode character '\U0001f389'
```

rgthree-comfy printed its startup banner and ComfyUI exited during custom-node startup.

**This is ours, not the node's and not the user's.** Their own console launch of the same command works. #1259 points the relaunched child's stdout at a **log file** so a failed launch can say what it printed — and Python not writing to a terminal encodes to the *locale* codepage instead. On a Korean Windows install that is cp949, and a party popper is not encodable in cp949.

We changed where the bytes go, so the encoding on that destination is ours. The relaunch now sets `PYTHONUTF8=1` / `PYTHONIOENCODING=utf-8`, which is also exactly what the reporter's verified workaround does.

## Deliberately narrow

- **Windows only.** That is where the report is and where a non-UTF-8 ANSI codepage is the default. A POSIX box under `LANG=C` has the same shape, but nothing has reported one, and widening a fix past its evidence is how it acquires a second bug.
- **An explicit value always wins**, including `PYTHONUTF8=0` — the deliberate opt-out. Overriding a user's encoding choice to fix an encoding bug is just a different encoding bug.
- **Returns the input unchanged** when there is nothing to add, so a launch that was already fine keeps its exact previous behaviour, including inheriting (`undefined`) rather than being handed a materialised copy.

## Four existing assertions changed, and why that matters

They read:

```ts
expect(spawnOptions().env).toBeUndefined();
```

They are about **launcher reconstruction**, not encoding — "a plain install inherits, nothing was rebuilt from disk". But `env` is now *defined* on Windows, so as written they were a **platform-dependent assertion that would have gone red on the `windows-latest` leg only**. I confirmed that first: on this Windows machine the unmodified suite failed 4/74 before I touched it.

They now assert the **delta** against the environment the launch would otherwise have used — that nothing *other* than the two encoding keys moved. That is deliberately **not** a second copy of the encoding rule, so it cannot pass by agreeing with a broken implementation of a rule it is not testing; the rule itself is pinned against literals in the new file.

## Verification

- 8 new tests, 82 passing across the new suite + `restart-launcher-env`, and **125 more** across the eight neighbouring restart suites (`desktop-restart`, `desktop2-restart-info`, `launch-log-capture`, `npx-restart-scope`, `process-control`, `remote-restart`, `restart-configured-command`, `restart-instance-identity`). `npm run lint` clean, `check:changelog` passes.
- **Mutated**: reverting the spawn to `env: envPlan.env` fails the wiring pin. Worth noting *only* that one fails — the behavioural tests are correctly insensitive, because "no launcher additions" is true either way. The helper only changes the spawn on Windows, so behavioural coverage runs for real on one CI leg out of three; the wiring pin covers the other two.

## The half this does NOT fix

The reporter also observed the relaunched argv dropping `-s` and `--windows-standalone-build`. That is structural rather than an oversight: `-s` is a **Python interpreter flag and is never in `sys.argv`**, which is what `relaunchArgv` reconstructs from. Recovering it means parsing the OS command line and deciding what `reproducible` should say when that is unavailable — a separate change against the same #776 machinery, with its own refusal semantics to get right. **#2693 stays open for it**, and the changelog entry says so.


artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6). **No findings** — recording that explicitly, since with the external gate down an unreviewed PR and a reviewed-clean one otherwise look identical.

What I actually checked, rather than eyeballed:

**The helper's contract holds in each direction.** `withUtf8StdioEnv` returns `env` untouched off Windows; returns it untouched when both vars are already set (so an `undefined` env stays `undefined`, i.e. inherit is preserved rather than silently converted); and only ever fills a var that is missing or blank — it cannot override a value a caller set deliberately.

**The exit family is complete.** Three `spawn` sites exist in process-control: the Desktop app launch (`stdio: "ignore"`, no Python), the macOS `open` (not Windows), and the Python relaunch — which carries the fix. I also checked the other spawns in `src/services`: `panel-launcher` starts a terminal running the MCP CLI with `stdio: "ignore"`, not Python, so it is not a sibling of this one. That mattered because "fixed one exit, left the siblings" has been the shape of three separate defects I found today.

**The wiring is tested, not just the helper.** Replacing `env: withUtf8StdioEnv(envPlan.env)` with `env: envPlan.env` fails *is applied at the relaunch spawn, on the env the plan resolved*. A green helper test proves the function computes something; this proves it reaches the child.

`relaunch-utf8-stdio` + `restart-launcher-env`: **82 tests**.

artokun:

The scope statement makes three promises. Two hold. The third does not, on Windows specifically — which is the only platform this function runs on.

### "An EXPLICIT value always wins" — it did not, in any non-canonical case

Windows environment variable names are **case-insensitive**. `withUtf8StdioEnv` read them case-sensitively:

```ts
const v = effective[name];   // effective["PYTHONUTF8"]
```

Node's `process.env` is a proxy that honours Windows' semantics, so this is fine when `effective` *is* `process.env`. But these paths routinely hand the helper a **plain object** — the Stability Matrix reconstruction, the `/proc` read, any caller that materialises an env — and a spread of `process.env` is an ordinary object that does not.

Measured on this machine rather than reasoned about:

```
PythonUtf8=0 node -e '...'
  process.env.PYTHONUTF8         -> "0"
  ({...process.env}).PYTHONUTF8  -> undefined      (key present as PythonUtf8)
```

So a user who deliberately set `PythonUtf8=0` — the documented opt-out, and the PR body calls out `PYTHONUTF8=0` by name — had it read as absent, and the fix would have written a **second key differing only in case**. Two spellings of one variable in the child's environment block, to fix an encoding bug by overriding the user's encoding choice: precisely what the scope statement says it will not do.

Lookup and write both go through the existing key now, so an oddly-cased var is respected when set and filled **in place** when blank, rather than shadowed.

### Holds: Windows-only, and unchanged input returned unchanged

The `platform !== "win32"` guard returns `env` by identity, and the both-already-set path returns the original object rather than a copy — so a launch that was already fine keeps `undefined` and inherits, instead of being handed a materialised environment. That is a real distinction and it is genuinely preserved; I checked it because "returns the input unchanged" is easy to claim and easy to break with a defensive spread.

### Holds: the four changed assertions were a real Windows-leg failure

`expect(spawnOptions().env).toBeUndefined()` is about launcher *reconstruction*, not encoding, and `env` is now defined on Windows. The PR says it confirmed the unmodified suite failed 4/74 here first. I re-ran the amended suite: **74 passed**, so those assertions still cover what they were written for.

### Verification

- `relaunch-utf8-stdio`: 11 passed. `restart-launcher-env`: 74 passed. `npm run lint` clean, anti-slop none added, `check-changelog` passes.
- Mutation-verified, each caught: the lookup reverted to case-sensitive (kills 3); the write ignoring the existing key (kills the shadowing test specifically); and the early return removed.

### Unchanged

Still does not close #2693 — the argv half is untouched, as the PR says.

artokun:
## Reproduced the crash and verified the fix resolves it — on a real interpreter, not as env-shape assertions

The tests here assert that the right keys land in the right places. They do not establish that those values actually prevent the crash. Ran the reporter's mechanism directly:

```
$ PYTHONIOENCODING=cp949 python emoji_test.py > out.txt
UnicodeEncodeError: 'cp949' codec can't encode character '\U0001f389'
  in position 16: illegal multibyte sequence
```

Character-identical to the report — `\U0001f389` is 🎉, the party popper from rgthree's startup banner — and reached the same way: Python writing an emoji to a **redirected** stdout under a legacy codepage, which is exactly the state #1259 created by pointing the relaunched child's stdout at a log file.

With the env this PR installs:

```
$ PYTHONUTF8=1 PYTHONIOENCODING=utf-8 python emoji_test.py > out.txt
exit 0
rgthree startup 🎉 banner
```

## The opt-out is still effective, not merely respected

```
$ PYTHONIOENCODING=cp949 PYTHONUTF8=0 python emoji_test.py > out.txt
UnicodeEncodeError: 'cp949' codec can't encode character '\U0001f389' …
```

This is the half worth checking. The scope statement promises an explicit value always wins — so `PYTHONUTF8=0` has to keep *doing something*, not just survive into the child's environment. An opt-out that is honoured but inert would satisfy every test in this PR while being a hollow promise, and it is the direction nobody looks at.

Together with the case-insensitivity fix (a user's `PythonUtf8=0` is now seen at all), that means the opt-out is both **visible** to the code and **effective** in the interpreter.

## What is still not claimed

Nothing about the reporter's machine. This reproduces the *mechanism* on a stand-in interpreter with a forced codepage; their install and locale are not mine. The argv half is #2844, and the underlying `-s` loss is fixed there rather than here.

artokun:
Verified on real Windows against a real CPython, since `ci.yml` hardcodes `runs-on: ubuntu-latest` for every matrix leg (#2851) — the Windows-named checks on this PR never ran on Windows, and `withUtf8StdioEnv` returns its input unchanged on any other platform, so on the runner this code path is inert.

Spawning `python -c "import sys;print(sys.stdout.encoding, sys.getfilesystemencoding())"` with the env this function produces:

```
bare (no vars)              cp1252 utf-8
after withUtf8StdioEnv      utf-8  utf-8
```

So the defect reproduces here — a relaunch with no vars gets the machine's legacy codepage for stdio — and the fix moves it to UTF-8. My box is Western (`cp1252`) rather than the reporter's Korean (`cp949`), but that is the same failure: a legacy ANSI codepage on stdout is what makes a non-ASCII log line raise on write.

**Both non-clobber directions also hold, which is the half a fix like this usually gets wrong:**

```
explicit opt-out    PYTHONUTF8=0  PYTHONIOENCODING=cp949   -> both preserved
mixed-case opt-out  PythonUtf8=0                           -> preserved, PYTHONIOENCODING added
                                                              no duplicate PYTHONUTF8 key
```

That last line is the one worth having a test for and you have one: Windows env vars are case-insensitive, so a case-SENSITIVE read would have missed `PythonUtf8=0` and then added a second key differing only in case — silently overriding the user's opt-out with the value they explicitly declined. The write-through in `assign` is what prevents it, and it works.

Nothing to change. Recording the measurement because the unit tests necessarily fake the platform, and this is the one path where faking it means the assertion and the behaviour can drift without anyone noticing.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **Measured both halves on this Windows box rather than reading them. The reported crash reproduces byte-for-byte, the fix clears it, and the case-insensitivity claim in the comment is exactly right.** None of this is checkable in CI: the matrix declares three OSes but `runs-on: ubuntu-latest` is hardcoded (#2851), so every "Build (windows-latest)" job here ran on Linux.

## 1. The crash, and the fix

A Python child printing `\U0001F389` with stdout redirected to a **file** (which is what #1259 made the relaunch do):

```
locale codepage      -> exit=1
  UnicodeEncodeError: 'cp949' codec can't encode character '\U0001f389'
                       in position 6: illegal multibyte sequence

PYTHONUTF8=1 / PYTHONIOENCODING=utf-8
                     -> exit=0, "party 🎉 popper" written, zero stderr
```

That error string is identical to the reporter's, down to the codepoint and position. And the framing in the PR body is the part I most wanted to check and agree with: **we** changed where the bytes go, so the encoding on that destination is ours. The node is not printing anything unusual — a terminal would have taken it fine.

## 2. The case-insensitivity comment is not hand-waving

It claims a measurement, so I re-ran it independently:

```
PythonUtf8=0 node -e ...
  process.env.PYTHONUTF8        -> "0"
  ({...process.env}).PYTHONUTF8 -> undefined
  spread keys matching /utf8/i  -> [ 'PythonUtf8' ]
```

Confirmed. Node's `process.env` is a proxy that honours Windows' case-insensitive env names; a **spread of it is an ordinary object and loses that**. So without `keyOf`, a user who set `PythonUtf8=0` would have their explicit opt-out missed *and* a second key differing only in case added — on exactly the paths that materialise an env (the Stability Matrix reconstruction). That is a subtle trap, it is real, and the fix for it is correct.

## The scoping is right too

Windows-only, explicit-value-wins, and `undefined` returned unchanged so a launch that was already fine keeps inheriting rather than being handed a materialised copy. The last one matters more than it looks: materialising the env is what would have re-broken #776's Stability Matrix PATH injection.

Nothing blocking. The argv half of #2693 is untouched, as the body says, and #2844 covers it — those two are independently mergeable in any order (verified with `merge-tree`; only the usual CHANGELOG collision).

artokun:
**Re-measured the trap this code documents, on a real Windows box, and confirmed the fix handles it.**

The comment in `launcher-env.ts` says the case-insensitivity of Windows environment names survives `process.env` but is lost by a spread, and marks it "measured, not assumed". Re-running that measurement with the variable actually set in an odd case:

```
PythonUtf8=0 node -e '…'
  process.env.PYTHONUTF8        -> "0"
  ({...process.env}).PYTHONUTF8 -> undefined
  key spelling present          -> ["PythonUtf8"]
```

Reproduced exactly. Node's `process.env` proxy honours the case-insensitive lookup; an ordinary object made from it does not.

**And the consequence, which is what makes it load-bearing here.** Feeding that same spread through the function:

```
  keys after: ["PythonUtf8"]   values: ["0"]
  verdict: opt-out HONOURED through a spread
```

One key, still `0`, and **no `PYTHONUTF8=1` added beside it**. Without the case-insensitive check the spread would look like the variable was unset, and the relaunch would ship a child an environment containing both `PythonUtf8=0` and `PYTHONUTF8=1` — the user's explicit opt-out silently overridden by a contradictory duplicate. Neither spelling is "wrong" to Windows, so which one wins is not something the caller could reason about.

Also checked on this box:

- a spread with nothing set → `PYTHONUTF8="1"`, canonical spelling only, no stray second key
- `platform: "linux"` → env untouched

Not covered: this verifies the env construction, not the relaunch that consumes it — no ComfyUI was restarted to observe the child's actual decoding.

PR #2793, opened 2026-09-03
fix(2511): recover list_local_models when panel rejects models/<category>

### LINKED ISSUE #2511: list_local_models fails against reachable remote ComfyUI target
`list_local_models` could not inspect the connected ComfyUI model inventory because the headless request to the configured remote target failed with EHOSTUNREACH, even though the live sidebar panel was bound to the same ComfyUI origin.

### Steps to reproduce
1. Connect the sidebar panel to the remote ComfyUI.
2. Call list_local_models with action=list.
3. Observe that the tool reports the inventory as undetermined after a network failure.

### Exact error (scrubbed)
connect EHOSTUNREACH to the configured ComfyUI target while requesting /models/checkpoints (and other model categories).

### Fix
Applied locally: no. Upstream-only: yes. The tool should either route model inventory through the live panel origin when available, or clearly expose/use the panel connection for remote inventory instead of failing from the headless MCP route.

### Environment
OS: macOS; ComfyUI remote; comfyui-mcp 0.52.146; panel 0.15.64; GPU details unavailable from the failed inventory request.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-luna` · effort `medium` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
## Summary
- #2624 relayed list_local_models through panel `fetch_comfyui_read` as `models/<category>` after a headless transport failure.
- Current panels (0.15.150 through 0.15.159) still allowlist only history, system_stats, logs, object_info (plus workflow_templates on newer builds). The panel rejects `models/<category>`, so loras/diffusion_models listings stayed undetermined on a reachable remote Docker target.
- After that allowlist rejection, inventory is reconstructed from panel `object_info` loader combos. HTTP errors on the configured route still skip the panel.

## Test plan
- [x] `npx vitest run src/__tests__/services/list-local-models-panel-relay.test.ts src/__tests__/services/list-local-models.test.ts src/__tests__/services/model-listing-coverage.test.ts` (84 passed)
- [x] `npx tsc --noEmit`

Fixes #2511


artokun:
CI's red here was the **lint** gate, not the tests — `anti-slop no-unknown-returns`, `model-resolver.ts` baseline 0 → 1.

The signature was:

```ts
async function panelObjectInfo(memo: ModelsPanelMemo): Promise<unknown | undefined>
```

`unknown | undefined` collapses to `unknown`, so the "absent" case it appears to document is not even expressible — the gate was right to flag it.

Typing it properly turned up a real hole. The relay returns whatever the panel sent; a JSON `null` parses fine, and the memo keyed on `!== undefined` cached it. The call site's `info === undefined` guard then passed, and the combo walkers were handed a non-document — so the listing reported **zero models as though the panel had answered**, which is the "undetermined inventory rendered as empty" failure this issue exists to remove.

Now a named `ParsedObjectInfo`, a real `| undefined` union, and a type-guard narrowing (no cast) that treats a non-document body as a **failed** read.

- new pin: a `null` object_info body yields `noSourceAvailable`, not an empty answered listing;
- mutation: dropping the guard fails that case and nothing else;
- `npm run lint` clean — 785 known findings, none added;
- `src/__tests__/services`: 273 files, 6454 tests.

One note for the run log: `panel-image-relay.test.ts > carries one deadline through the bridge and refuses a late panel result` failed under full-suite concurrency and passes 45/45 in isolation, on this branch and on an unrelated one. It is load-sensitivity in that test, not this change — tracking it separately.

artokun:
Adversarial pass (mine — codex is out of tokens). The recovery works; its trigger is a prose match on another repo's wording, and nothing on either side would notice if that wording changed.

**Checked first, because it is the usual killer:** `ALLOWED_OPERATION_NOTE` in the panel's `fetch-comfyui-read.js` is a plain constant with **zero** `tr()` calls in that file, so i18n does not disarm `/operation must be one of/i` today. Good — that is the failure mode this kind of predicate normally dies of.

**The live risk is drift, not translation.** The trigger is:

```ts
return /operation must be one of/i.test(parts.join("\n"));
```

matched against a string owned by comfyui-mcp-panel:

```js
const ALLOWED_OPERATION_NOTE =
  "operation must be one of history, system_stats, logs, object_info, …";
```

Reword that sentence in the panel — reorder the list, change "must be one of" to "must be", localise it later — and this recovery silently stops firing. `list_local_models` goes back to the #2511 behaviour with no error anywhere, because the fallback simply never triggers.

**And the test cannot catch it.** `list-local-models-panel-relay.test.ts` mocks the rejection with its own hardcoded copy of that sentence, so it keeps passing after the panel diverges. The test asserts the predicate matches the string the test wrote, which is always true.

**There is a code available, and it is nearly enough.** The panel throws through:

```js
function invalidInput(message) {
  return readError("invalid_input", `fetch_comfyui_read rejected the operation: ${message}`);
}
```

so a stable `code: "invalid_input"` is already on the wire. It is too coarse on its own — the same code covers "operation is required" — which I assume is exactly why the prose match exists. But that is a one-line panel change away from being sufficient: a distinct `unsupported_operation` code for the allowlist rejection, matched here, with the prose test kept as the fallback for panels that predate it.

Not blocking: the current form works against every panel shipping today, and the fallback-to-prose design means adding the code later is backwards compatible. Worth a line in the PR body naming the coupling, though, so the next person to reword that sentence in the panel repo has some chance of knowing this depends on it.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **Both halves verified. The one that matters most is the distinction between "the panel answered with nothing" and "the read failed".**

## Absence of evidence is not evidence of absence, and this gets it right

```ts
if (!isObjectInfoDocument(parsed)) {
  memo.objectInfoFailed = true;
  return undefined;
}
```

with the reason stated at the site: without it the memo caches the bad value (`null !== undefined`), the `info === undefined` guard at the call site passes, and the walkers are handed a non-document — **reporting "no categories" as though the panel had answered**. The call site then does `if (info === undefined) throw transportErr`, so a failed read surfaces the original transport error instead of an empty inventory.

That is the correct shape for a recovery path. An empty model list is indistinguishable from a working server with no models installed, and #2511 is precisely a report of `list_local_models` returning nothing useful — so a fallback that could produce a *confident* empty answer would have made the original symptom permanent and harder to diagnose.

## The error predicate is narrowing, not swallowing

```ts
if (!isUnsupportedModelsPanelRead(panelErr)) throw panelErr;
memo.modelsOpUnsupported = true;
```

Anything the predicate does not recognise is rethrown, and `fetchModelsRoute` does not catch it either. So if the panel's wording ever changes — the cross-repo string contract panel#2240 pins — this does not silently stop falling back; it surfaces the panel's real rejection. **I had claimed the opposite on panel#2240 and corrected it there after reading this file**; the fail-safe direction here is what makes that guard a smaller concern than I first described.

## Version skew

The predicate matching a phrase across a repo boundary is the fragile part, but the failure mode is right: an older or reworded panel degrades to a visible error rather than a wrong answer. panel#2240 now pins the phrase from the other side, including a guard that it stays untranslated, so both halves of the contract are held down.

Nothing blocking.

artokun:
**Worked the Copilot review on this PR. The first finding is real, and I traced it end to end rather than fix it, because the obvious repair is unsafe.**

**Duplicate emission in the fallback.** The chain:

```
collectLocalModels:  dirsToScan = modelType ? [modelType] : [...MODEL_SUBDIRS]
MODEL_SUBDIRS:       ... "clip", "diffusion_models", ... "text_encoders", "unet"
objectInfoDirsMatching("unet")             -> {unet, diffusion_models}
objectInfoDirsMatching("diffusion_models") -> {unet, diffusion_models}   // same set
identityFoldersFor("unet")                 -> ["unet"]
identityFoldersFor("diffusion_models")     -> ["diffusion_models"]        // NOT folded
```

So an unfiltered listing visits both members of each alias pair; in the object_info fallback both synthesize the **same** names from the same combos; and `makeModelDeduper` treats the two categories as distinct identities, so both are accepted. Every modern UNET model appears twice, and likewise for `clip`/`text_encoders`.

It is specific to the fallback this PR adds — on the HTTP path `/models/unet` and `/models/diffusion_models` return the actual, different folder contents.

**Why I did not just fold them in the deduper.** That was my first instinct and it is wrong: on the HTTP path those are genuinely different folders, and `unet/foo.safetensors` and `diffusion_models/foo.safetensors` can be two different files. Folding would silently drop a real model — trading a visible duplicate for an invisible omission, which is the worse failure for a tool whose job is inventory.

The correct shape is to canonicalise **in the fallback only**, which needs the filtered/unfiltered distinction threaded to `modelNamesFromObjectInfo` (it currently cannot tell). That is a small change but a real design choice about where the canonical category lives, so I would rather surface it than pick one silently on a PR that is already green.

**The other three findings, assessed:**

- **`None` sentinel** (`model-resolver.ts:511`) — combo options go into inventory unfiltered, and LoRA stackers commonly carry `None`; `missing-models.test.ts` has that exact fixture. Looks right, same fallback path, and `WEIGHT_EXTS` already exists in this file as the obvious filter.
- **A missing mapped combo becomes an authoritative empty** (`:570`) — a category with no matching loader returns `[]`, is added to `coverage.answered`, and renders "No models found". That is the report-truthfulness shape: absence of a mapping presented as absence of models.
- **Memo scope** (`:909`) — `otherRegisteredCategories` builds a fresh memo, repeating the rejected `models` op and the potentially large `object_info` relay. A performance issue rather than a correctness one.

All four are on the fallback path added here, none is on the HTTP path, and this PR is green — so none of them blocks it. Recording the analysis so whoever takes them does not have to re-derive the chain.

PR #2795, opened 2026-09-03
fix(1539): refuse git URL node installs before Manager v4 queueing

### LINKED ISSUE #1539: panel_install_node advertises git repository installs but sends an unlisted URL to Manager registry

artokun:
Fixes #1539

## Recurrence
On mcp 0.52.179 / panel 0.15.159, `panel_search_nodes("ImageComparer")` still returned `https://github.com/Slimy-Comfy/Slimy_ImageComparer` as `id`. Passing that id to `panel_install_node` queued a Manager v4 registry lookup. v4 ignores `repository` and resolves by bare name, so the pack never landed. Headless `install_custom_node(source:"git")` still worked.

Prior PRs #1611 / #2123 / #1686 narrowed the contract and relied on the panel to refuse v4 Git URLs. Search still advertised a Git URL as an install id, and MCP still dispatched it.

## Change
- `panel_search_nodes` no longer returns a raw Git URL as `id`. Mapping keys that are clone URLs become a repo name (or a CNR id when present), with `repository` + `git:true` and a note pointing at `install_custom_node(source:'git')`.
- `panel_install_node` refuses Git URLs on the Manager `v2` dialect **before** `nodes_install` is queued. Legacy 3.x / `v2-batch` keep the direct `files:[url]` path.

## Tests
Targeted vitest: 6 files, 138 passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project are documented here. This project adheres to
 ### MCP
 #### Fixed
 - **`panel_slice_workflow` seeds outputs inside nested overlapping groups (#2780).** Membership used the first containing box only, so a SaveImage inside both an outer group and a nested inner group was missed when slicing by the inner title. Any matching containing group now seeds, and a miss lists every containing title.
+- panel_search_nodes no longer returns a raw Git URL as an install `id`, and panel_install_node refuses Git URLs before Manager v4 queueing rather than sending an unlisted URL as a registry lookup (#1539)
 
 ## [0.52.182] - 2026-09-03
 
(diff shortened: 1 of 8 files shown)


artokun:
CI caught a real regression in this PR, and the cause was the guard itself rather than the tests.

The refusal read:

```ts
try {
  api = await detectManagerApi();
} catch {
  api = "v2";          // <- any failure to reach the Manager counted as "this is v4"
}
```

So the guard fired on the **absence** of evidence. Two consequences:

1. It refused legacy Manager 3.x `files:[url]` installs — which still work — and told the caller the wrong cause while doing it.
2. It broke nine tests in the `#1129` suite (`install-accepted-never-enqueued`), which installs by `repository` and stubs no Manager at all. Detection threw there, the catch claimed v4, and the install was refused instead of dispatched.

Fixed by refusing only on a **positive** detection; an undetermined dialect falls through and the dispatch reports whatever actually failed.

Verified by mutation in both directions:

- restoring `catch { api = "v2" }` → the new pin fails **and the same nine `#1129` failures come back**, exactly as CI saw them;
- deleting the refusal outright → the v4 direction fails (3 tests).

`src/__tests__/orchestrator` is green locally: **301 files, 4920 tests**.

Issue #2784, opened 2026-09-03
restart_comfyui: successful restart orphans the server (launched proc exits), then every subsequent restart is refused


artokun:
### What happened / root cause

`restart_comfyui (action:"restart")` succeeds once, but leaves the server in a state where it can never be restarted by the tool again. Two calls, same session:

**Call 1 — "succeeded", but orphaned the server:**
```
"stopped": true, "started": true, "ready": true,
"startup": "unconfirmed",
"listener_ownership": "unconfirmed",
"message": "... The process this call launched EXITED (exit code 0), so this could not be
confirmed as the process this call launched — a wrapper or trampoline can hand off and
still leave a healthy server."
```

**Call 2 (after patching a custom node, needed to reload the module) — hard refusal:**
```
"stopped": false, "started": false, "startup": "not-attempted",
"message": "Refusing to restart: ComfyUI Desktop started the server on port 8191
(PID 100292), but no Desktop app is still supervising it — its parent process is gone.
A restart from here asks ComfyUI-Manager to stop the process and relies on that
supervisor to start it again, so it would be stopped and nothing would bring it back."
```

The refusal logic is correct in isolation — it avoided killing a server nothing would revive. The defect is that **call 1 is what created that unsupervised state**, and it reported itself as a success. So the tool reliably burns its own ability to restart: the first restart works, and every one after it is refused.

This is not a ComfyUI-Desktop install. It is ComfyUI-Easy-Install launched from a plain `python_embeded` interpreter; the tool appears to have misclassified it as Desktop-managed after the parentage was lost.

### Steps to reproduce

1. Local ComfyUI (Easy-Install / `python_embeded`), started outside the MCP, serving `:8191`.
   argv: `main.py --use-pytorch-cross-attention --port 8191`
2. `restart_comfyui (action:"restart")` → returns ready:true, `listener_ownership: "unconfirmed"`, notes the launched process exited with code 0.
3. `restart_comfyui (action:"restart")` again → refused with the "no Desktop app is still supervising it" message. Unrecoverable through the tool.

### Exact error (scrubbed)

```
Refusing to restart: ComfyUI Desktop started the server on port 8191 (PID 100292), but no
Desktop app is still supervising it — its parent process is gone. ... ComfyUI was left
running (not stopped) so you don't lose the server. Restart it from the ComfyUI Desktop app.
The server reported these launch arguments:
[...]

PR #2797, opened 2026-09-03
fix(2723): build the trainer image asynchronously, so it can finish at all

### LINKED ISSUE #2723: train_doctor build_image always times out before first Docker trainer image completes
train_doctor action:"build_image" is synchronous, but the comfyui/call_tool transport hard-times out after 300 seconds. The Docker build is terminated or otherwise does not finish; a subsequent doctor check still reports image:false. This reproduced twice, including a retry with cached layers, and blocks all local Docker LoRA training.

### Steps to reproduce
1. train_doctor {"action":"doctor"} -> docker:true, gpu:true, image:false.
2. train_doctor {"action":"build_image"}.
3. Wait 300 seconds.
4. Tool returns: timed out awaiting tools/call after 300s.
5. train_doctor {"action":"doctor"} -> image:false.

### Exact error
tool call error: tool call failed for comfyui/call_tool
Caused by: timed out awaiting tools/call after 300s

No credentials or tokens included.

### Fix
Make build_image an asynchronous job that immediately returns a build id and exposes status/progress, or allow the training tool transport to stream progress and use a timeout longer than the documented several-minute first build. The Docker build process should survive the request timeout and doctor should be able to detect completion.

### Environment
Windows 10
ComfyUI 0.33.1
ComfyUI frontend 1.48.7
Python 3.12.11
PyTorch 2.9.1+cu128
NVIDIA RTX 3090 24GB
comfyui-mcp 0.52.150
panel 0.15.149
Docker daemon reachable: true
NVIDIA GPU passthrough: true
Trainer image after both attempts: false

---

Related (same root cause): #489

artokun:
Closes #2723.

`train_doctor action:"build_image"` awaited a `docker build` of a CUDA + torch + ai-toolkit image — its own description says *"one-time, several minutes"* — while the panel's `call_tool` transport hard-times out at **300s**. So the call always returned

```
timed out awaiting tools/call after 300s
```

and the follow-up doctor still reported `image:false`. The reporter hit it twice, including a retry with warm layers. Local Docker LoRA training was not slow; it was **unreachable**.

## The shape this server already uses

Awaiting minutes-long work inside a request is what `download_model` and `enqueue_workflow` refuse to do — they return a handle and are polled. This is that, for the trainer image. The docker child belongs to the **orchestrator**, not to the request, so it keeps building after the call returns.

```
build_image  →  { build: { id, status: "running", … }, adopted: false, note: … }
doctor       →  { …, image: false, image_build: { id, status: "running", tail: [...] } }
doctor       →  { …, image: true,  image_build: { id, status: "done" } }
```

## Two properties make that safe, and both are pinned

**A re-issue adopts the running build.** Two `docker build` runs against one tag race for the layer cache, and the loser's tag write is the one that survives. Re-issuing after the 300s timeout is *precisely* what the reporter did, so this is not a hypothetical.

**The detached promise catches.** It has no awaiting caller, so an unhandled rejection is fatal under Node's default policy — a failed build would take the orchestrator down, which is worse than the bug being fixed. There is a test that rejects the build and asserts the registry settles instead.

## Reported through `doctor`, not a new action

`doctor` is already the surface a caller polls after `build_image`, and it is the one the reporter polled. Routing it there also avoids adding to the tool vocabulary for something the existing surface answers.

It settles a second ambiguity along the way: `image:false` **while a build runs**, `image:false` **after one errored**, and `image:false` **with nothing ever attempted** used to render identically. The error message and the docker log tail are now kept, so a failed build says what happened.

## Verification

- 11 new tests; 182 passing across the nine train/trainer suites. `npm run lint`, `check:vocabulary` and `check:changelog` all clean.
- The vocabulary gate caught a real slip while I worked: I had written the literal `"train_build_image"` (retired in 0.50.0) into the envelope and the fixtures. It now uses `TRAINER_COMMAND.buildImage`, and the fixtures do not name it at all.
- Not run: an actual `docker build`. The build call is mocked, so what is proven here is the handle/adoption/settlement behaviour, not that the Dockerfile still builds — that is unchanged by this PR and would need a GPU host.


artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6). One real finding: **the fix reintroduced this issue's own shape, quieter.**

## A detached build that hangs pins the slot forever

The old bug burned the tool's ability to build with a **timeout**. Detaching removes that — but a detached build that never settles leaves `running` pinned, so every later `build_image` **adopts a corpse**, and a wedged build is indistinguishable from a healthy one for as long as anyone cares to poll.

That is the #2784 shape ("the first call works and every one after it is refused") wearing different clothes, and it matters here specifically because *hanging is the observed behaviour in this issue*.

Adoption is still correct — two `docker build` runs against one tag race for the layer cache, and the loser's tag write survives — so nothing is superseded. What changes is that the state stops being invisible: past a deliberately generous **45 minutes**, `doctor` reports `looks_stalled` with what to check (`docker ps`) and how to clear it.

Generous on purpose: a cold CUDA + torch build on a slow link genuinely takes a while, and crying stuck too early is worse than staying quiet.

## What I checked and found sound

**The adoption is race-free**, and I verified that rather than assuming it: the check-then-set from `if (running)` through `running = id` contains no `await`, so two concurrent callers cannot both pass it.

**Both settle paths are covered** — `.then` handles ok and not-ok, `.catch` handles a throw from the builder *or* from the then-handler. The `.catch` is not optional: an unhandled rejection from a detached promise takes the process down under Node's default policy, which would turn a failed build into a dead orchestrator.

Pinned: making the threshold never trip fails the new case; a settled build is never called stalled however old.

`src/__tests__/tools` + `trainer-image-build`: **102 files, 1602 tests**.

artokun:

One real bug in the adoption rule, and two checks that came back clean.

### Adoption silently substituted a different image

`startTrainerImageBuild` adopted whenever a build was running:

```ts
if (existing && existing.status === "running") return { build: existing, adopted: true };
```

It compared nothing else. But `aiToolkitRef` is caller-controlled, and this tool's own description says it *"pins the ai-toolkit commit/tag for **reproducibility**"*. So a call that pinned a ref while another build was in flight got back `adopted: true` — and a build of the **other** ref. The tag that landed was not the one requested, which is the one thing that parameter exists to guarantee.

The build record already carries `ai_toolkit_ref`. The evidence needed to catch this was sitting on the object being returned, unread — the same shape as the link-id repair on panel#2196 and the cache footprint on #2771.

**Refused rather than starting a second build.** The reason adoption exists at all is that two `docker build` runs against one tag race for the layer cache and the loser's tag write survives; that is no less true when the refs differ. The refusal names both refs, states plainly that the running build was *not* adopted, and points at `doctor` to poll until it settles.

Matching refs still adopt — including `undefined` vs `undefined`, which is the reporter's own plain re-issue after the 300s timeout, so the behaviour this PR is actually about is untouched. There is a test for that neighbour specifically, because it is the one a careless guard would break.

### Clean: the wedged-build dead end has an exit

I went looking for "detected but no recovery" — a build that hangs pins `running` forever and every later call adopts a corpse. `trainerImageBuildLooksStalled` reports it, and the doc says outright that adoption *"does not supersede anything"*. That would be a dead end if the note stopped there. It does not: it tells the reader to check `docker ps` for the build container and that restarting the orchestrator clears the slot — and since the registry is module state, that is true rather than reassuring. Not a finding.

### Clean: the detached rejection cannot take the process down

The build promise has no awaiting caller, so an unhandled rejection would be fatal under Node's default policy — a failed build killing the orchestrator would be worse than the bug being fixed. It catches, and the existing test that rejects the build and asserts the registry settles is real coverage, not a source check.

### Verification

- `trainer-image-build`: 15 passed. `train-consolidation`: 39 passed. `npm run lint` clean, anti-slop none added, `check-changelog` passes.
- Mutation-verified, each caught: the mismatch check removed; the mismatch adopting anyway; and the comparison inverted (which kills three, including the matching-ref neighbour — the inversion breaks adoption in both directions, which is exactly what that test is there to notice).

artokun:
Adversarial pass (mine — codex is out through Sept 6). **I went after five specific ways this could fail and four are closed. Reporting that as a result rather than manufacturing a finding.**

| probe | outcome |
|---|---|
| Work still AWAITED before the detach, re-creating the 300s timeout by another route | **Closed.** `dockerAvailable()` is the only awaited call, and `execDocker` defaults to `timeoutMs = 20_000` — bounded well under 300s. The Dockerfile check is `existsSync`. |
| A wedged build pins `running` forever, so `build_image` can never start a fresh one | **Closed, and answered in the message.** The stalled note says every call ADOPTS it and that "restarting the orchestrator clears the slot" — an escape hatch that is actually stated, which is more than #2784's refusal manages. |
| `refMismatch` refuses with no way forward | **Closed.** It names the running ref, the requested ref, why a second build was not started, and "poll `doctor` until that build settles, then re-issue". |
| A detached promise rejection killing the orchestrator | **Closed**, and commented as deliberate. |
| Adoption relying on state that can vanish | see below |

The stall reasoning is the part I most expected to be hand-waved, and it is the opposite — `trainerImageBuildLooksStalled` exists, is generous on purpose (45 min), and the comment names the `#2784` "the tool burns its own ability to do the thing" shape explicitly before deciding adoption is still correct. That is the right call and it is argued rather than assumed.

## The one thing I could not settle

`builds` is module state, justified by:

> a build does not outlive the orchestrator that owns the docker child, so persisting it would only ever describe a build that is gone

That is true for the CLI child. Whether it is true for the **build** depends on the builder: BuildKit ties the build to the client session and cancels on disconnect, so the assumption holds; the legacy builder (`DOCKER_BUILDKIT=0`) runs it daemon-side and it survives the client. Under the legacy builder, an orchestrator restart mid-build would leave a real build running with an empty registry, and the next `build_image` starts a second one against the same tag — the exact race adoption exists to prevent.

**I did not verify this.** Docker CLI and buildx are installed on this machine but the daemon is not running, and starting Docker Desktop to test it is not something I will do to the user's machine unprompted. BuildKit is the Docker Desktop default, so the assumption almost certainly holds in practice; I am flagging it only because the comment states it unconditionally and the failure mode is silent. One clause — "(BuildKit cancels the build when the client disconnects)" — would make the reasoning checkable by the next reader.

Nothing here blocks merge.

artokun:
**Two of the three Copilot findings on this PR were real correctness bugs. Both fixed.**

**1. Doctor reported the OLDER build.** `started_at` is `Date.now()`, so two builds started in the same millisecond tie — and the comparison was a strict greater-than:

```js
if (!newest || build.started_at > newest.started_at) newest = build;
```

`Map` iterates in insertion order, so on a tie the FIRST build inserted stays `newest` and the later one can never displace it. Doctor then reported the earlier settled result and hid the newer failure or success, which is the opposite of what "newest" means. Ties now resolve to the most recently inserted.

**2. Two clock readings emitted as one observation.** `image` is sampled by `trainerDoctor()` at the top of the handler; `image_build` was read from a **live object the builder mutates in place**, with several awaits in between — including a network pod probe that can take seconds. A build finishing in that window renders:

```
image: false                       <- sampled first
image_build.status: "done"         <- sampled later
```

which reads as a contradiction rather than as two observations taken at different times. The build is now re-sampled and frozen immediately before serialising, and when the pair is genuinely inconsistent the response says so (`image_check_stale`) instead of leaving the reader to reconcile it.

**Verification:** mutation-verified — restoring the strict comparison fails exactly the new test. 72 tests green across the trainer suites, `tsc` clean.

**Two things I got wrong writing the test, both caught by running it.** My first draft invented a `__trainerImageBuildTestHooks` export that does not exist; the second called `startTrainerImageBuild({})`, but it returns `{ build, adopted }` and requires `contextDir`. The test now drives the public API with the clock pinned rather than reaching into the store, which is what it should have done from the start.

**The third finding I have not taken.** It says a new export is absent from the `ai-toolkit.js` mock in `train-consolidation.test.ts`. That suite is green here and in CI, so if the gap is real it is not currently observable; I would rather leave it visible than "fix" a mock against a failure I cannot reproduce.

PR #2799, opened 2026-09-03
fix(1524): detect and recover a mid-session empty toolset at the turn boundary

artokun:
> **Stacked on #2774** — base is `fix/2742-empty-mcp-toolset`, so this diff shows that PR's commits too until it merges. Review this one for the turn-end half.

Addresses the mid-session half of #1524, which recurred **today** on the Claude backend: after a user interrupt and several healthy runs, every `mcp__panel__*` tool vanished from the compact catalog and calls failed with `No such tool available`, while panel events kept arriving.

## Why the turn-end watch was blind

#2774 makes the **init** check tool-aware. It deliberately stops there, and I said why in its body: the turn-end watch polls `mcpServerStatus()`, and that carries no tool information —

```ts
McpServerStatus = { name, status, serverInfo?, ... }   // no tools, no count
```

— so a server that stays `connected` and simply stops contributing tools passes it, and a reconnect fired there could not be verified.

## That reason no longer holds

`Query.getContextUsage()` returns, alongside the usage breakdown:

```ts
mcp_tools: { name: "mcp__panel__panel_graph_outline", server_name: "panel", tokens }[]
```

The session's own accounting of what the model can see, **per server**. So the drop is detectable exactly where the existing `reconnectMcpServer()` recovery already runs — and the same call **verifies** the attempt afterwards. That closes the loop: detect at init (#2774), detect at each turn end, attempt the one reconnect the episode is owed, and say honestly whether it came back.

## Its own episode map, deliberately

Sharing `mcpDown` would be a bug rather than a shortcut. Its "episodes that CLOSED" sweep treats any name absent from the current **degraded** set as recovered — and an empty server *is* connected, so every empty episode would close itself on the turn it opened and report a recovery that never happened. `mcpEmpty` is separate, and the two sets are disjoint by construction because a degraded server is excluded from the empty check.

## Same silences as the init check

"Cannot tell" is never rendered as "none": no `getContextUsage` on the harness, a call that throws, or a reading with no `mcp__` entries at all. That last one is the same documented blind spot — an empty list is indistinguishable from a harness that does not populate the field, and a false "your tools are gone" on a healthy session is worse than the silence it replaces. The reported signature has `mcp__comfyui__*` present throughout, so it is covered either way.

## Verification

- 8 new tests, 26 in the file, 81 across the four related suites. `npm run lint` and `check:changelog` clean.
- **Three mutations, each killed by its own tests**: dropping the turn-end call site (4 fail), skipping the post-reconnect verification (1), and reporting every turn instead of once per episode (1). The last one matters — a server that stays empty across a long session must not add a notice to every turn, and the reconnect must be spent once, not once per turn.
- Not claimed: the **cause**. This reports the observation, as #1524 has from the start.

## One lead for whoever takes the cause

Today's recurrence followed **a user interrupt**; the earlier #2742 reports did not mention one. If that reproduces, an interrupt tearing down the in-process server's registration without the session noticing is a much narrower thing to look for than "it sometimes goes missing".


artokun:
Self-adversarial review (codex out of tokens, so the pass is mine until Sept 6) found **the same blind spot here that I had just fixed in the init half** — the sibling exit, left behind.

`pollEmptyToolsets` returned "unknown" whenever `getContextUsage()` reported an **empty** `mcp_tools`:

```ts
if (!Array.isArray(entries) || entries.length === 0) return null;
```

The reasoning is right on a first reading — an empty list is indistinguishable from an API that does not populate the field. But it makes the **total** failure, where every server contributed nothing, exactly as invisible as the case this check exists to catch. And its test pinned that silence with the same justification the init half used — *"the reported signature has mcp__comfyui__* present throughout"* — which is evidence about the **earlier reports**, not a property of the failure. The sixth report (2026-09-03, on 0.52.183) may not have them.

Closed the same way as #2774: one **populated** reading, ever, proves the API reports this field. After that an empty list is a real observation, and every configured, non-degraded server is reported. An API that never populates it never latches, so it is never falsely reported on — the check can widen, never narrow.

Pinned in both directions: restoring the old early return fails the new latched-then-emptied case, and the un-latched silence is still asserted on its own.

Two notes on the run:

- `src/__tests__/orchestrator`: **303 passed**. The two failures are `install-accepted-never-enqueued` and `install-node-git-expectation`, which **probe the developer's real ComfyUI** and fail whenever one is running. That is fixed separately in #2829 and is not on this branch — measured, not assumed: pointing `COMFYUI_URL` at a dead port makes both pass.
- Writing the new test surfaced two harness details worth knowing: `driveTurns` defaults to **one** turn (a two-reading fixture needs `2`), and `noticesOf` returns notice objects, so joining them without `.map(n => n.message)` yields `[object Object]` and an assertion that can never match.

artokun:

One robustness fix pushed; two candidate findings chased and dropped, with what settled them.

### The check distrusts `server_name`, then concludes confidently from its absence

`pollEmptyToolsets` collects a server only when the field is a string:

```ts
if (typeof entry?.server_name === "string") withTools.add(entry.server_name);
…
return names.filter((name) => !notUp.has(name) && !withTools.has(name));
```

That `typeof` guard exists because the runtime shape is not trusted. But if it fails for **every** entry, `withTools` is empty and the filter returns **every connected server** — each one reported as having lost its tools, each spending the single reconnect its episode is owed, on a completely healthy session.

That is precisely the false positive this PR argues against elsewhere (*"a false 'your tools are gone' on a healthy session is worse than the silence it replaces"*). The existing latch covers the empty-**array** case; nothing covered the unparseable-**shape** case. Same absence-of-evidence shape as the #1539 refusal.

**Not reachable on the pinned SDK** — `sdk.d.ts:3222` declares `server_name: string` as required, so today every entry parses. It becomes reachable the moment that shape drifts, and the failure would be silent to us and wrong in the direction that fabricates an outage. The fix is three lines and completes distrust the code had already decided to spend.

The discriminator is exact rather than heuristic: one parseable `server_name` puts a name in the set, so *entries but no names* means none parsed. Both legitimate neighbours still report — an empty list (every server contributed nothing, the #2742 signal, already gated by the latch) and a populated list that simply omits our names. There is a test for that second one specifically, because it is the case a sloppier guard would have swallowed.

**My first version of the guard did swallow it.** I wrote `withTools.size === 0` without the `entries.length > 0` half, which disabled total-emptiness detection outright — the primary signal. An existing test caught it immediately (`reports EVERY server once a populated reading has been seen and then empties`). Worth stating plainly: the guard I added to prevent a false positive briefly created a false negative in the main path, and the suite already written for this PR is what stopped it.

### Dropped: `names` vs `server_name` namespace mismatch

The higher-value version of the same failure, and reachable today if true: `names` comes from `Object.keys(this.mcpServersForRun())` — config keys — while `server_name` comes from the SDK. If those namespaces differed at all, *every* server would look empty on every turn. They do not: the SDK namespaces tools as `mcp__<server_name>__<tool>`, and `server_name` is the same config key. Not a finding.

### Dropped: the episode-map disjointness claim

The body claims `mcpEmpty` and `mcpDown` are "disjoint by construction because a degraded server is excluded from the empty check." That holds — `pollEmptyToolsets` builds `notUp` from `health.degraded` and `health.pending` and filters both out before returning. The stated reason for not sharing `mcpDown` is also correct: its closed-episode sweep treats any name absent from the degraded set as recovered, and an empty server is `connected`, so sharing it would self-close every empty episode on the turn it opened.

### Verification

- `claude-init-mcp-degraded` + `mcp-session-health`: **71 passed**. `npm run lint` clean (anti-slop: none added). `check-changelog` passes.
- Mutation-verified, each caught: the guard removed entirely; the guard left over-broad (my own first version); and the `typeof` check replaced with a blind `String(...)` coercion.

Still stacked on #2774 — merge that first and this diff shrinks to the turn-end half.

artokun:
Two things about this PR that are worth knowing before it is merged, both found by auditing CI state across the queue rather than by looking at the diff.

## CI has never run on this branch, and will not while it is stacked

`.github/workflows/ci.yml` triggers on:

```yaml
on:
  pull_request:
    branches: [main]
  push:
    branches: [main]
```

This PR targets `fix/2742-empty-mcp-toolset`, not `main` — so it matches neither trigger. `gh run list --branch fix/1524-turn-end-empty-toolset` returns **nothing at all**: not a failure, not a skip, zero runs since the branch was created.

Every other P0/P1 PR in the queue has a last-completed run of `success`; this one has none. So its only verification is the local runs quoted in the reviews.

That resolves itself the moment #2774 merges — GitHub retargets this PR to `main`, CI fires for the first time, and the four legs run. **Worth expecting a first-ever CI result at that point rather than treating a green base as evidence about this branch.** Merge #2774 first, let this one's checks actually run, then merge it.

## It had drifted from its own base, and that was my doing

My CHANGELOG sweeps merged `origin/main` into both stacked branches independently, which left this one **CONFLICTING against `fix/2742-empty-mcp-toolset`** — a stacked branch conflicting with its own base, which no amount of sweeping main would fix.

Resolved by merging the base, which is what a stacked branch should be tracking. Both conflicts union-resolved with nothing dropped:

- the mid-session test block (159 lines) is this branch's own contribution and absent from the base — kept whole;
- the changelog keeps all four bullets, the base's two and this branch's two.

`tsc` clean, and `claude-init-mcp-degraded` + `mcp-session-health` + `hello-retarget` together: **93 passed**. Now `MERGEABLE` again.

A note for future sweeps of mine: a stacked branch must be swept against its **base**, not against main. Sweeping both against main is what pushed them apart.

artokun:
**CI has never run on this PR**, and it cannot until #2774 lands: `statusCheckRollup` is empty, because both workflows trigger on `pull_request: branches: [main]` and this one targets `fix/2742-empty-mcp-toolset`. So merging the stack today would put code on main that no CI job has ever compiled.

I ran the equivalent locally on the stack head (`4eadfa0`, which already contains #2774's commits):

| check | result |
|---|---|
| `tsc --noEmit` | clean |
| `npm run lint` (anti-slop) | "none added" |
| this PR's own three test files | **93 passed** |
| all of `src/__tests__/orchestrator/` | 306 files, 4992 tests — **2 files / 10 tests failed** |

Those 10 failures are **not from this stack**. `origin/main` (`12af490`) fails the identical 10 in the identical two files — `install-node-git-expectation.test.ts` and `install-accepted-never-enqueued.test.ts` — and they are the two files #2829 exists to fix. On `test/hermetic-manager-dialect` the same two files go **58/58 green**, so the cause is confirmed and already has a PR.

So the stack itself is clean on everything I can measure locally. The remaining risk is the one the merge order addresses: **#2774 first**, which lets CI run on this PR for the first time before it reaches main.

One note on the local run: a full-suite `vitest` invocation from inside `.claude/worktrees/` collects nothing, because `vitest.config.ts` excludes `**/.claude/**` and excludes are additive. Passing the directory as a filter works, which is how the 306-file run above happened — worth knowing before anyone reads a bare "no test files" as a pass.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **The design checks out. But this PR has never been built by CI, and that is structural rather than bad luck.**

## No CI has run on these 366 lines

```
#2774 (base = main)        Build (ubuntu-latest)   SUCCESS
                           Build (windows-latest)  SUCCESS
                           Build (macos-latest)    SUCCESS
                           Pack & install smoke    SUCCESS
                           Mintlify Deployment     SKIPPED

#2799 (base = fix/2742-…)  Mintlify Deployment     SKIPPED      <- that is the whole list
```

Cause, from `ci.yml`:

```yaml
on:
  pull_request:
    branches: [main]
```

`pull_request.branches` filters on the **target** branch, so a PR into anything other than `main` matches nothing and no build job is created. Not "fewer checks" — none. 193 lines of `claude-backend.ts` and 172 lines of tests have not been type-checked, linted, or run by CI once.

The green tick reads the same as any other PR's, which is the part worth fixing: a PR with zero jobs and a PR with four passing jobs are visually indistinguishable in the list.

Two ways out, both cheap: add the stacking pattern to the trigger (`branches: [main, 'fix/**']`), or retarget to `main` after #2774 merges and let CI run before merging this. The second needs no workflow change but does mean **not merging this immediately after its base**, which is the natural thing to do.

## On the substance — the latch is right

I went in expecting to flag the latch, because #2774's own comment names a per-process one as the nastier deferral hazard. This one is not that:

```ts
private sawPopulatedMcpUsage = false;
```

Per-**instance**, and monotonic false → true, so it can widen the check but never narrow it. And the reasoning is the distinction that matters: an empty `mcp_tools` before anything populated is *an API that does not report them*, not a session with none.

That also makes the two PRs genuinely complementary rather than overlapping — #2774's init check catches "empty from the start" (which this one deliberately cannot, since it has seen nothing populated yet), and this catches "went empty mid-session". Neither covers the other's case.

Nothing blocking on the code. The CI gap is worth resolving before merge rather than after.

artokun:
**Reachability check passes — and it turned up one cost question worth answering before merge.**

The wiring is live, not inert. Verified against the pinned SDK (`^0.3.240` → 0.3.261):

```
sdk.d.ts:2776   getContextUsage(opts?: { detail?: 'summary' | 'full' }): Promise<SDKControlGetContextUsageResponse>
sdk.d.ts:3493   mcp_tools: { name: string; server_name: string; tokens: number }[]
```

So the `typeof q.getContextUsage !== "function"` guard is not permanently false, and `server_name` is the SDK's own field name — the detection reads the shape that actually arrives. `pollEmptyToolsets` is invoked twice per turn boundary (`claude-backend.ts:609` and `:645`, the check and the re-check), so it is on the live path.

**The question.** The call is `await q.getContextUsage()` with no argument, and the SDK documents the default:

> `detail: 'full'` counts each category with the token-count API; `'summary'` answers from the last response's usage and local estimates **without the per-category token-count calls**. Defaults to `'full'`.

This poll reads only `entries[].server_name` — it never touches `tokens`. So on every turn end, twice, it is paying for per-category token counting to answer a question about names.

I could not establish from the types whether `'summary'` still populates `mcp_tools` (the field is on the same response type either way, and its names plausibly come from local tool state rather than the counting API). If it does, `getContextUsage({ detail: "summary" })` gets the same answer without the extra calls, on a path that runs at every turn boundary of every Claude-backend session.

Worth a moment's thought before merge, since the cost is per-turn and permanent. If `'summary'` turns out not to populate it, the current call is correct as-is and this is a non-issue — I would rather ask than quietly change a call whose cheaper form I cannot prove equivalent.

Everything else about the change traces cleanly to #1524's symptom.

artokun:
Following up on the `detail` question above — it cannot be settled from the SDK, and here is why, so nobody spends time looking there.

`getContextUsage` in the shipped package is a pass-through to the CLI over the control protocol:

```js
async getContextUsage(e) { return (await this.request({ subtype: "get_context_usage", ...e })).response }
```

That is the whole implementation (`sdk.mjs`, 0.3.261). The `detail: 'summary' | 'full'` semantics — and therefore whether `'summary'` still populates `mcp_tools[].server_name` — live in the Claude Code CLI binary, not in anything vendored here. The type declares the field on the response either way, so the types cannot answer it.

So it is an empirical question: one session, call it both ways, compare `mcp_tools.length`. Cheap to answer with a live orchestrator, impossible to answer by reading. Leaving the call as `getContextUsage()` is the safe default until someone does that — a wrong guess toward `'summary'` would silently disarm the detection this PR exists to add, which is worse than the per-turn cost.

PR #2802, opened 2026-09-03
fix(2784): a Desktop restart refusal names the evidence that classified the install

artokun:
#2784 stays OPEN (this is deliberately partial). It fixes the part that made the report unreproducible, which is the part I can do without the redacted path.

## The problem with the refusal

```
Refusing to restart: ComfyUI Desktop started the server on port 8191 (PID 100292), but no
Desktop app is still supervising it — its parent process is gone.
```

That is a **claim**, on a ComfyUI-Easy-Install running an embedded python, with nothing attached for the reader to check. The reporter could see it was wrong and had no way to test it; the report then arrived with the path redacted to `~`, so the cause could not be reproduced from it either.

Running the pure predicates against their argv **as posted** does not classify at all — so the trigger is in the part they redacted, and no output would have told them which.

## Three signals, one verdict, very different strength

| signal | what it is |
|---|---|
| on-disk Desktop-2 marker (`.comfyui-desktop-2`, `.launcher/snapshots/*.json`) | a fact about the install |
| a Desktop binary among the process's ancestors | a fact about the process tree |
| **an argv substring** | a **NAME match** |

The third is `joined.includes(marker)` over the whole command line, so a directory merely **called** `Comfy-Desktop` satisfies it. The refusal now says exactly that, in the message the user reads:

> Classified as Desktop-managed by: the substring `"comfy-desktop"` in the server's own launch command — a NAME match, which a directory merely CALLED that also satisfies. If that is wrong for this install … this refusal is wrong with it, and the launch arguments below are the thing to check it against.

`isDesktopApp`'s own doc already admitted it "answers yes on any mention of the Desktop install". What changed underneath it is that since #814 this gates a **refusal** — so a false positive now strands the user permanently, where it used to just pick a gentler restart route. The strict twin `isDesktopSupervisorProcess` spells that asymmetry out and errs strict; this one never got the same treatment.

## Diagnosis only, deliberately

The classification itself is unchanged. Tightening the substring test is a behaviour change across a surface I cannot reproduce, and I would rather the next occurrence identify its own cause in one line than guess at a fix now. #2784 stays open, and the changelog entry says so.

Evidence is recorded where the decision is made rather than re-derived, so the ancestor walk does not run a second time — it is deliberately skipped unless the argv already looks Desktop-2, and "no Desktop binary among the ancestors" is usually a question nobody asked.

## Verification

- 13 new tests, 173 passing across the six desktop/restart suites. `npm run lint` and `check:changelog` clean.
- **Three mutations, each killed by its own test**: removing the clause from the refusal, removing the NAME-match warning, and removing the guard that keeps the message unchanged when no evidence was recorded (an install classified before this field existed must not read "classified because: undefined").
- One test records the fact that made the report unreproducible — the reporter's argv as posted carries none of the markers — so if that ever stops being true, someone finds out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+### MCP
+
+#### Fixed
+- **the evidence rides on all THREE surfaces the Desktop verdict drives, not just the refusal (#2784).** The same classification also turns auto-restart supervision OFF ("Auto-restart supervision is only supported for directly spawned Python ComfyUI processes.") and produces "Could not determine ComfyUI Desktop executable path." — which is exactly the branch a NAME-match false positive lands in, since it has no Desktop install and so no exe path to find. Both asserted Desktop with nothing attached to check, and the auto-restart one is the hardest to trace back to a directory name because nothing was refused, a capability just quietly went away. Both now carry the same disclosure, worded for a statement rather than a refusal (the refusal clause says "this refusal is wrong with it" and points at "the arguments below", neither of which is true on these surfaces). The auto-restart message is pinned by a real call through a new test seam, not a source grep.
+- **a Desktop restart refusal now names the evidence that classified the install (#2784).**
+
+
 ## [0.52.201] - 2026-09-07
 
 ### MCP
--- a/src/__tests__/services/desktop-evidence.test.ts
+++ b/src/__tests__/services/desktop-evidence.test.ts
@@ -1,0 +1,272 @@
+// #2784 — a refusal that names no evidence is a claim the reader cannot check.
+//
+// `restart_comfyui` refused with "ComfyUI Desktop started the server on port 8191",
+// on a ComfyUI-Easy-Install running an embedded python. The reporter could see it
+// was wrong and had nothing to test it against; the report arrived with the path
+// redacted to `~`, so the cause could not be reproduced from it either. Running the
+// pure predicates against their argv as posted does NOT classify, which means the
+// trigger is in the part they redacted — and no output would have told them which.
+//
+// Three signals reach the same verdict and they are not equally strong:
+//
+//   on-disk Desktop-2 marker   a fact about the install
+//   ancestor Desktop binary    a fact about the process tree
+//   argv substring             a NAME match, which a directory can satisfy by
+//                              being called that
+//
+// These pin that the weakest one identifies itself as weak, because that is the one
+// a user can recognise and contest.
+
+import { afterEach, describe, expect, it } from "vitest";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { __processControlTestHooks } from "../../services/process-control.js";
+
+/** The marker set, as the shipped predicate holds it. */
+const MARKERS = [
+  "programs/comfyui/resources",
+  "programs\\comfyui\\resources",
+  "comfyui.app",
+  "comfy desktop",
+  "comfy-desktop",
+  "comfyui-desktop-2",
+  "@comfyorgcomfyui-electron",
+];
+
+describe("#2784 the Desktop classification names its evidence", () => {
+  it("exposes a test seam at all", () => {
+    // If this fails the seam was renamed; the tests below say nothing until it is
+    // re-pointed, and a silently skipped suite is worse than a red one.
+    expect(typeof __processControlTestHooks.reset).toBe("function");
+  });
+
+  it.each(MARKERS)("an argv carrying %s is a NAME match, not a launch fact", (marker) => {
+    // The predicate is `joined.includes(marker)` over the whole command line. So a
+    // path segment is enough — which is the reading the refusal must not hide.
+    const argv = [`C:\\Users\\x\\${marker}\\ComfyUI\\main.py`, "--port", "8191"];
+    const joined = argv.join(" ").toLowerCase();
+    expect(joined.includes(marker)).toBe(true);
+  });
+
+  it("the reporter's argv as POSTED does not classify — the trigger is in the redaction", () => {
+    // Recorded because it is the fact that made the report unreproducible, and the
+    // fact this change exists to prevent recurring.
+    const argv = [
+      "C:\\Users\\x\\ComfyUI-Easy-Install\\ComfyUI\\main.py",
+      "--use-pytorch-cross-attention",
+      "--port",
+      "8191",
+    ];
+    const joined = argv.join(" ").toLowerCase();
+    expect(MARKERS.some((m) => joined.includes(m))).toBe(false);
+  });
+
+  it("a directory merely NAMED like Desktop does classify", () => {
+    // The most likely un-redaction, and the one the evidence clause is written to
+    // make legible: an ordinary install under a folder called `Comfy-Desktop`.
+    const argv = ["D:\\AI\\Comfy-Desktop\\ComfyUI-Easy-Install\\ComfyUI\\main.py", "--port", "8191"];
+    const joined = argv.join(" ").toLowerCase();
+    expect(MARKERS.some((m) => joined.includes(m))).toBe(true);
+  });
+});
+
+describe("#2784 the refusal carries the clause", () => {
+  const src = new URL("../../services/process-control.ts", import.meta.url);
+
+  it("RENDERS the evidence it was given", async () => {
+    // A real call, not a grep of the source. The previous version of these three
+    // asserted that process-control.ts CONTAINED certain substrings, which cannot
+    // fail if the clause renders empty, if the field is never populated, or after
+    // a rename -- it pins the text, not the behaviour.
+    const { desktopEvidenceClause } = await import("../../services/process-control.js");
+    const out = desktopEvidenceClause("a ComfyUI Desktop binary among this process's ancestors");
+    expect(out).toContain("Classified as Desktop-managed by:");
+    expect(out).toContain("a ComfyUI Desktop binary among this process's ancestors");
+    expect(out).toContain("this refusal is wrong with it");
+  });
+
+  it("says NOTHING when no evidence was recorded", async () => {
+    // An install classified before this field existed must get the message it
+    // always had, not "classified because: undefined".
+    const { desktopEvidenceClause } = await import("../../services/process-control.js");
+    expect(desktopEvidenceClause(undefined)).toBe("");
+    expect(desktopEvidenceClause("")).toBe("");
+  });
+
+  it("marks the argv signal as a NAME match wherever it is produced", async () => {
+    // The weakest of the three signals must identify itself as contestable. This
+    // one stays a source check on purpose: the string is built inside
+    // detectDesktopLaunch, which needs a live process tree to reach.
+    const { readFileSync } = await import("node:fs");
+    const { fileURLToPath } = await import("node:url");
+    const text = readFileSync(fileURLToPath(src), "utf-8");
+    expect(text).toContain("a NAME match, which a directory merely CALLED that also satisfies");
+  });
+
+  it("appends the clause to the abandoned-supervisor refusal", async () => {
+    // Structural pin, and labelled as one: it proves the call site is still in the
+    // same template as the message, not that the message renders with evidence.
+    const { readFileSync } = await import("node:fs");
+    const { fileURLToPath } = await import("node:url");
+    const text = readFileSync(fileURLToPath(src), "utf-8");
+    const at = text.indexOf("ComfyUI Desktop started the server on port");
+    expect(at).toBeGreaterThan(-1);
+    expect(text.slice(at, at + 1600)).toContain("desktopEvidenceClause(info.desktopEvidence)");
+  });
+
+  // ---------------------------------------------------------------------------
+  // #2784 — the SELECTION, now reachable.
+  //
+  // The two tests above pin the sentence; neither pins WHICH signal produced it.
+  // Verified by mutation: making the marker branch or the ancestor branch
+  // unreachable left all fourteen green, so the refusal could have named the
+  // weakest signal in place of the strongest and nothing would have said so —
+  // and pointing the reader at an argv NAME match that did not decide is the same
+  // misdirection the unfalsifiable claim was.
+  describe("#2784 which signal is named", () => {
+    const { detectDesktopLaunch, reset, setParentPidResolver, setProcessIdentityResolver } =
+      __processControlTestHooks;
+
+    afterEach(() => reset());
+
+    it("names the argv substring when that is all there was, and quotes the marker it hit", () => {
+      // No Desktop-2-shaped path, so the ancestor walk is deliberately never run.
+      const r = detectDesktopLaunch(4242, ["C:/Apps/Comfy Desktop/python.exe", "main.py"]);
+      expect(r.isDesktopApp).toBe(true);
+      expect(r.desktopEvidence).toContain('the substring "comfy desktop"');
+      expect(r.desktopEvidence).toContain("NAME match");
+    });
+
+    it("names the ANCESTOR binary instead, once the process tree can answer", () => {
+      // Same install, plus a real Desktop supervisor overhead. The stronger fact
+      // must displace the contestable one — that ordering is the whole point of
+      // saying which signal decided.
+      setParentPidResolver((pid) => (pid === 4242 ? 99 : null));
+      setProcessIdentityResolver((pid) =>
+        pid === 99
+          ? { pid: 99, executablePath: "C:/Programs/ComfyUI/Comfy Desktop/Comfy Desktop.exe" }
+          : undefined,
+      );
+      const r = detectDesktopLaunch(4242, ["C:/Users/x/comfyui-installs/python.exe", "main.py"]);
+      expect(r.isDesktopApp).toBe(true);
+      expect(r.desktopEvidence).toContain("ancestors");
+      expect(r.desktopEvidence).not.toContain("NAME match");
+    });
+
+    it("an on-disk Desktop-2 marker OUTRANKS both, and it is the strongest claim made", () => {
+      // The remaining arm, reached with a REAL directory rather than a stubbed
+      // existsSync: several process-control suites stub that true, which would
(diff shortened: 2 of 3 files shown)


artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6) found that **all three tests on this PR were greps**, and I want to be plain about that because it is the failure mode this repo cares most about.

Each one read `process-control.ts` as a string and asserted it contained a substring. That cannot fail if the clause renders empty, if `desktopEvidence` is never populated, or after a rename. It pinned the **text**, not the behaviour — and I had just told the #2784 reporter that this change makes the refusal name its evidence, on the strength of those tests.

**Demonstrated, not argued:** making `desktopEvidenceClause` return `""` unconditionally leaves every one of the old assertions green. The new ones fail.

## What changed

`desktopEvidenceClause` now takes the **field** rather than the whole `ProcessInfo`. That is what lets a test call it for real without a fixture cast — two of the three assertions are now genuine calls (renders the evidence it was given; renders nothing when there is none).

The one remaining source check is **labelled as structural**: it proves the call site still sits inside the same template as the message, not that the message renders. That distinction is now written in the test.

## Wiring, verified by reading it rather than assuming

`detectDesktopLaunch` computes `desktopEvidence` → it is carried onto the returned `ProcessInfo` → the refusal appends the clause. I checked each hop because the claim I had already made publicly depended on it.

## Why this matters for the report

I also withdrew my question to the reporter about what `~` expands to — it was the wrong question. Their visible path contains no Desktop marker, so the classification almost certainly came from the **on-disk** `.comfyui-desktop-2` / `.launcher/snapshots/*.json` check, which needs nothing from argv. This PR is what will tell them that directly.

`src/__tests__/services`: **275 files, 6505 tests**. The single failure is `panel-image-relay`'s deadline case, which passes in isolation and is addressed in #2823.

artokun:

Two structural checks passed; the tests had a real hole, now closed.

### Passed: no path reaches the refusal without evidence

The value of this PR is entirely that the clause is *always* there — a refusal that sometimes names its evidence is the old unfalsifiable claim with extra steps. So I traced every route:

- All four `isDesktopApp: true` exits of `detectDesktopLaunch` carry `desktopEvidence`; the single `false` exit carries none, correctly.
- It propagates one hop into `ProcessInfo` and is read at the refusal as `info.desktopEvidence`.
- `ProcessInfo` never leaves `process-control.ts` and is never JSON round-tripped, and the `lastProcessInfo` cache is assigned from the same gatherer — so there is no restored-from-disk path that would arrive without it. That was the specific thing I expected to find broken, since the report comes from a restart.
- `desktopEvidenceClause` returns `""` for a missing value, so the degraded case is the old message rather than a rendered `undefined`.

### Passed: the evidence selection is sound on every branch

The argv-substring branch is only reachable when `argvMarker` is non-null (it is the only remaining truthy signal at that point), so the quoted marker can never render as `null`.

### The hole: the tests pinned the sentence, not the decision

Mutation found it. Making the **marker** branch unreachable, and separately the **ancestor** branch, both left all fourteen tests green. So the refusal could have named the weakest of the three signals in place of the strongest and nothing would have said so — and sending a reader to check an argv NAME match that did not actually decide is the same misdirection as the claim this PR replaced. The two existing tests say as much themselves: *"this one stays a source check on purpose… needs a live process tree to reach"*.

It is reachable now. `detectDesktopLaunch` goes on the existing `__processControlTestHooks` seam — the same move the PR already made for `desktopEvidenceClause`, and for the same stated reason — and there are four cases:

- the argv substring names itself and quotes the marker it hit;
- a real Desktop supervisor overhead **displaces** it (the ordering the whole message depends on);
- an on-disk `.comfyui-desktop-2` marker displaces **both**;
- a plain python install yields no evidence at all.

The marker case uses a real `mkdtemp` directory rather than a stubbed `existsSync`, deliberately: several process-control suites stub that true, and a stubbed probe here would reclassify every ordinary python install in this file.

Both previously-surviving mutations are now killed, and each of the three arms is killed by the test written for it.

### Verification

`src/__tests__/services/`: **6513 passed**, 2 failed — both pre-existing and neither related. `panel-image-relay` is what #2823 fixes; `panel-secrets-slots` passes in isolation and is a full-run concurrency artifact. `npm run lint` clean, anti-slop none added.

### Unchanged

Still does not close #2784, and the classification itself is still untouched. Tightening the substring test remains a behaviour change across a surface neither of us can reproduce, and I would rather the next occurrence name its own cause in one line.

artokun:
Adversarial pass (mine — codex is out of tokens). Three checks, all clean; one question about what disclosure does and does not buy.

**1. The reported evidence really is the one that decided.** I traced `isDesktop = fromArgv || Boolean(fromParent) || fromMarkers` against the evidence ternary and every combination lands right: `fromParent` is only computed when `fromMarkers || desktop2Argv`, so a classification reached purely through an argv substring cannot report an ancestor it never looked for; and a case where only `desktop2Argv` is true with no ancestor and no marker returns `isDesktopApp: false` before any evidence is built. Strongest-first ordering does not let a weak signal borrow a strong sentence.

**2. It is actually surfaced, not just computed.** Forwarded at 4285 and rendered by `desktopEvidenceClause` into the refusal at 3461. Worth checking explicitly because "computes a field no output carries" is the defect shape that has come up repeatedly this week — here the wiring is real.

**3. No sibling refusal is missing it.** The only other `ProcessControlError` on that path (4195) is the pid/argv corroboration mismatch, which is not a Desktop claim and correctly says nothing about Desktop.

**The question.** The argv-substring branch is honest about being weak — *"a NAME match, which a directory can satisfy by being called that"* — and a user who reads the refusal can now see that is what classified them. But seeing it does not get them a restart: a plain install living under, say, `…/comfy desktop tests/` is still refused, and the remedy is to rename a directory, which the message does not suggest and which nobody would guess.

So: is this PR deliberately scoped to "make the refusal explainable" with the misclassification itself left for a follow-up? That reads like the right call to me — the reported pain in #2784 is that the refusal was unfalsifiable — but if so it is worth an explicit line, because the issue title ("successful restart orphans the server, then every subsequent restart is refused") sounds like it should end with restarts working again, and this makes them explicably refused rather than not refused.

If a follow-up is wanted, the shape suggested by the evidence ordering is already there: treat argv-substring-only as *not sufficient* to refuse, and require a marker or an ancestor before asserting Desktop management.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **The evidence logic is sound; I tried to break it and could not.** Two siblings, though.

## What I verified

I went after the thing most likely to be wrong here — whether the reported evidence can name a signal that did *not* decide the verdict (a proxy classification outliving its proof). It cannot:

- The chain is `fromMarkers ? … : fromParent ? … : argvMarker`, and `isDesktop = fromArgv || fromParent || fromMarkers`. When the message names the argv substring, both stronger signals are false, so `fromArgv` must be what made `isDesktop` true — meaning `argvMarker` is provably non-null and the interpolation can never render `"null"`.
- When it names the ancestor, `fromMarkers` is false, and `fromParent` is only computed under `fromMarkers || desktop2Argv` — so that branch is reachable only with a real ancestor hit.
- `desktop2Argv` participates in choosing `desktopExePath` but never in the evidence string, and it cannot classify on its own (with the other three false, `isDesktop` is false and the function returns early).

So every message names a signal that is both true and sufficient. The tests pin all three branches independently plus the priority ordering and the no-signal case.

## Sibling 1 — the same classification also disables auto-restart, with no evidence

```js
supported: Boolean(info && !info.isDesktopApp),
message: … info?.isDesktopApp
  ? "Auto-restart supervision is only supported for directly spawned Python ComfyUI processes."
  : undefined,
```

A NAME match does not only block `restart_comfyui` — it turns **auto-restart supervision off**, and states a reason the misclassified user cannot check. That is a functional consequence beyond the refusal this PR addresses, and it is the one a user would never connect back to a directory name.

## Sibling 2 — the "cannot find the executable" message is exactly what a false positive hits

```js
message: info.isDesktopApp
  ? "Could not determine ComfyUI Desktop executable path. Please start it manually."
  : "No command-line info captured from previous run. Start ComfyUI manually."
```

For the population this PR is about — a directory merely *named* `Comfy-Desktop`, with no Desktop installed — `isDesktopApp` is true and `desktopExePath` is undefined, so this is precisely their branch. They are told the Desktop executable could not be found, which reads as "your Desktop install is broken" rather than "you were classified as Desktop by a name match". `desktopEvidenceClause(info.desktopEvidence)` already exists and is a one-line append here.

## Scope

Both are the same shape as the refusal you fixed: a claim about the install with nothing attached to check it. `desktopEvidenceClause` has exactly one call site today; these are the other two places the verdict becomes user-facing text.

Not asking for the behavioural half — the PR is explicitly partial and the refusal is safe (server left running, and `describeRecovery` does hand over a manual command, which I checked). But if the evidence clause is worth having, it is worth having on all three surfaces the same classification drives, and the auto-restart one is arguably the most confusing without it.

artokun:
Both siblings fixed in `a46a0fc`.

`desktopEvidenceNote` is a **sibling** of `desktopEvidenceClause`, not a parameter on it: the refusal wording says "this refusal is wrong with it" and points at "the arguments below", and neither is true on a surface where nothing was refused and no argv follows. Same fact, same escape hatch, sentence that fits.

Applied to:
- **`supervisorResult()`** — the auto-restart "only supported for directly spawned Python ComfyUI processes" message.
- **the exe-path failure** — "Could not determine ComfyUI Desktop executable path."

## Made the auto-restart one a REAL call

`supervisorResult` is now exposed through `__processControlTestHooks`, so its disclosure is driven for real (policy read from env exactly as production does) rather than grepped. That matters here specifically: a source check cannot fail if the note renders empty, if `desktopEvidence` is never populated, or after a rename — the same reasoning already written into this file for the pure-helper tests.

The exe-path one stays a structural pin and is **labelled as one**, scoped to a 300-char window from its own message template so it cannot pass on a mention elsewhere in the file. That message is built inside a start path needing a real spawn to reach; I did not want to fake a spawn to get a weaker guarantee than the label admits.

## Mutation-verified

| mutation | kills |
|---|---|
| drop the note from `supervisorResult` (line 2109) | "auto-restart's unsupported message carries the evidence" |
| drop the note from the exe-path message (line 4767) | "the exe-path failure names the evidence too" |

The needle occurs **twice** in the file, so both replaces were scoped to their enclosing site and the mutated line number read back before running — an unscoped replace would have hit the other surface and reported the guard as unpinned.

Also added a negative: a non-Desktop install's supervisor message stays `undefined`, so the note cannot leak onto installs that were never classified.

## Verification

`tsc` clean. Services suite: **6,525 passed**, 1 failure in `panel-image-relay.test.ts` — pre-existing and not mine. Controlled it: that test fails identically at `HEAD~1`, and it imports nothing from `process-control`. It is the deadline flake #2823 is about.

Scope unchanged otherwise — still no behavioural change, and #2784 still stays open for the part that needs the reporter's unredacted path.

artokun:
**Ran the classifier against a real Desktop-managed ComfyUI** — this machine's own, PID 16804, parent `C:\Program Files\Comfy Desktop\Comfy Desktop.exe`. The verdict is right and the evidence is honest, but it is reporting the weakest signal available while the strongest one sits unexamined.

```
isDesktopApp : true
desktopExe   : (none)
evidence     : the substring "comfy desktop" in the server's own launch command
               — a NAME match, which a directory merely CALLED that also satisfies…
```

The match comes from `--extra-model-paths-config "…\AppData\Roaming\Comfy Desktop\instance-model-paths\inst-….yaml"`. So a genuine Desktop install is classified by a **path substring**, and the message then correctly hedges that a directory merely named that would satisfy it too.

Meanwhile the actual process tree is unambiguous:

```
16804  python.exe          (ComfyUI)
38404  Comfy Desktop.exe   C:\Program Files\Comfy Desktop\Comfy Desktop.exe
10828  explorer.exe
```

## Why the strongest evidence is not gathered

`fromParent` is computed only when `fromMarkers || desktop2Argv`:

```ts
const fromParent = fromMarkers || desktop2Argv ? desktopExeFromAncestor(pid) : undefined;
```

This install has neither — no `.comfyui-desktop-2` marker, and an argv that is not Desktop-2 shaped — so the ancestor walk never runs, even though it would have returned the exe on the first hop. That is why `desktopExePath` is empty on a machine where the Desktop binary is the direct parent.

Two consequences, and the second is the one that matters for #2784:

1. **The disclosure understates itself.** The user is told the classification rests on a name match — contestable by design — when a definitive answer was one `readParentPid` away.
2. **The exe path is missed, and that path is what a relaunch needs.** The refusal's remedy is "restart it from the ComfyUI Desktop app" partly because the server cannot name the app; here it could have, from the parent process.

The cheap change is to compute `fromParent` whenever anything suggests Desktop — including `fromArgv` — rather than only under the two stronger predicates. It cannot weaken the verdict (`isDesktop` is already true on this path), it can only upgrade the evidence from a name match to a process-tree fact, and it fills in `desktopExePath` for free.

**One correction against myself:** my first run of this reported `isDesktopApp: false` and I nearly wrote it up as a false negative. I had hand-typed a truncated argv that omitted the `--extra-model-paths-config` flag. The real 442-character command line classifies correctly. Verified before reporting, not after.

artokun:
**Correcting the suggestion in my previous comment — I proposed the change without reading the comment that explains why the gate is narrow.**

I wrote that `fromParent` should be computed "whenever anything suggests Desktop, including `fromArgv`". Immediately above that line:

```
// Walking ancestors calls resolveProcessIdentity. Doing that on every gather
// consumes identity-sequence steps, so a later creation-time change is never
// seen and a recycled PID is killed (#776). Only walk when THIS argv already
// names a Desktop-2 layout.
```

So the narrow gate is deliberate and cites a fixed bug. Widening it the way I suggested would add identity reads to **every** gather — which is the shape #776 was about, and I am not going to propose reintroducing a recycled-PID kill to improve a diagnostic message.

The observation itself stands: on this machine the classification is a NAME match while the parent process is literally `Comfy Desktop.exe`, and `desktopExePath` is empty as a result.

**A revised suggestion that respects the constraint:** walk the ancestors *lazily*, at refusal-construction time, rather than during detection. The distinction is frequency — a gather happens constantly, a Desktop refusal is rare and already terminal. At that one point the process is known to be up (it is what is being refused), one `desktopExeFromAncestor` call would upgrade the evidence from "a NAME match" to the process-tree fact, and it would fill in the exe path the recovery text needs.

That keeps the hot path exactly as it is. Whether the identity-sequence concern also applies to a single call on a cold path is a question about #776's mechanism that I do not understand well enough to answer — so this is a suggestion to evaluate, not a change I would make.

Recording it because "the evidence is weaker than the machine can support" is worth knowing even when the cheap fix is unavailable.

artokun:
**Driven against the real Desktop-managed ComfyUI on this machine — and it corrects a measurement I posted on this PR earlier.**

Ground truth, captured from `Win32_Process` into a file and loaded, never retyped:

```
pid 16804  python.exe   (442-char command line, 15 argv tokens)
  parent 38404  C:\Program Files\Comfy Desktop\Comfy Desktop.exe
```

`detectDesktopLaunch(16804, <real argv>)`:

```json
{
  "isDesktopApp": true,
  "desktopExePath": "C:\Program Files\Comfy Desktop\Comfy Desktop.exe",
  "desktopEvidence": "a Desktop-2 install marker (.comfyui-desktop-2, or .launcher/snapshots/*.json) on a directory above the launch path"
}
```

Confirmed on disk rather than taken from the classifier's word — both markers are really there:

```
/c/Users/Artokun/ComfyUI-Installs/ComfyUI/.comfyui-desktop-2
/c/Users/Artokun/ComfyUI-Installs/ComfyUI/.launcher/snapshots/   (106 files)
```

Re-run with the ancestor walk returning nothing, the verdict and the evidence are unchanged — only `desktopExePath` drops. So the marker alone carries it, which is the behaviour the ordering intends.

**The correction.** Earlier on this PR I reported that a real Desktop-managed install is classified by the *weakest* signal — the argv name match — while its parent is literally `Comfy Desktop.exe`, and I suggested computing `fromParent` whenever anything suggests Desktop. That was already withdrawn because the gate above it cites #776 (recycled-PID kill). It is now also wrong on its own terms: **the real install is classified by the strongest signal, the on-disk marker.**

The reason is the same one that has bitten me repeatedly today. My earlier run used a hand-typed argv that dropped `--extra-model-paths-config "…\Comfy Desktop\instance-model-paths\inst-….yaml"`. Without it `desktop2InstallFromArgv` finds nothing, the marker branch never fires, and the classifier correctly falls through to the name match. I then read that fall-through as the production behaviour and built a review suggestion on it.

So the evidence ordering in this PR is doing exactly what it says on a real Desktop-2 install, and the case I claimed it handled badly does not arise here. Nothing to change; recording it because the earlier comment argued otherwise with numbers attached, and numbers are what people trust.

artokun:
**Ran `classifyDesktopSupervision` against a LIVE Desktop-supervised ComfyUI, not a fixture.** This box has Comfy Desktop running with a real backend serving 8188, which is exactly the situation #2784 is about — a restart that must not orphan a supervised server.

The real chain, from `Get-NetTCPConnection -LocalPort 8188` upward through `Win32_Process`:

```
pid 39476  python.exe          <- holds the port
  parent 16804  python.exe
    parent 38404  Comfy Desktop.exe     <- the shell, TWO hops up
      parent 10828  explorer.exe
        parent 12284  <unreadable>
```

Feeding the real 382-process table in as `SupervisionEvidence`:

```
VERDICT for pid 39476:
{ "verdict": "supervised" }
```

Correct, and correct for the right reasons:

- It walked **two hops** to find the shell — matching the documented expectation that "a Desktop backend sits one or two levels under its shell" — rather than stopping at the first python parent.
- It returned a **definite** verdict with no `because` field, so this is not an `unconfirmed` wearing a confident label.
- It was **not** derailed by `explorer.exe` sitting above the shell, nor by the unreadable process above that. The chain running out of road one hop further up did not degrade a verdict it had already established.

**Control first, as always:** `compareStartTimes("100","200")` → `parent-first` and `("300","200")` → `parent-newer` before I looked at the verdict. A start-time comparator that silently returned `unknown` would make the whole assessment vacuous, and that is not something to discover after trusting the answer.

Worth noting what this does *not* cover: the abandoned case. Every process in this chain is alive, so `processExists` never returned the definite `false` that the comment calls "the load-bearing one". The supervised path is now exercised against reality; the orphaned path is still fixtures only.

Issue #2023, opened 2026-08-21
Inside a RunPod pod with the tunnel, every install / connect command comfyui-mcp prints is unreachable: 0.0.0.0, the container's 10.x, or a hostname it never shows — and the docs teach port 8188 where the image serves 3000


artokun:
### What happens

Run comfyui-mcp **inside** the RunPod pod (the `docker/runpod` image, or any pod you `pip`/`npx` it into), with the tunnel, and the terminal hands you connect instructions that cannot work from anywhere but inside the container. Which wrong command you get depends on the mode, but every mode is wrong in the same way: the URL is derived from the process's own view of the network, and the process has no idea it is in a pod.

| mode | what is printed | why it is wrong |
|---|---|---|
| `--http --host 0.0.0.0 --port 9100` (what `docs/remote-connector.mdx:121` and `Dockerfile:13` tell you to run) | `http://0.0.0.0:9100/mcp` as the ready line (`src/boot.ts:611`) and as the error-path fallback (`boot.ts:297`) | `0.0.0.0` is a bind address, not a destination |
| same, with `--tunnel` | a `https://<random>.trycloudflare.com/mcp` that 502s | `boot.ts:289` calls `startQuickTunnel(port)` and **drops the `host` it was given**, so cloudflared targets `localhost`, which can resolve to `::1` while the server is bound IPv4-only on `0.0.0.0`. `src/services/tunnel.ts:100-105` documents exactly this trap; `secure-bridge.ts:321` and `orchestrator/index.ts:4911` pass `"127.0.0.1"`; `boot.ts:289` is the only caller that does not |
| `connect http://127.0.0.1:3001` with `COMFYUI_MCP_BRIDGE_HOST=0.0.0.0` (the pod-hosted orchestrator) | `Bridge URL : ws://10.x.x.x:9180/?token=…` with "In the panel: Settings → Advanced → Bridge URL → paste the URL above" (`orchestrator/index.ts:1272-1282`) | `10.x` is the container's `eth0`. `firstLanIPv4()` (`index.ts:1159-1182`) scores `eth0` +20 and a `10.` address +5 and picks it; nothing in the scorer knows a pod from a LAN |
| pairing a phone, LAN mode | `ws://10.x.x.x:9182/?token=…` as the QR / pair URL (`index.ts:4933`), and the same in the pinned-token banner (`index.ts:1323`) | same `firstLanIPv4()` |
[...]

PR #2809, opened 2026-09-03
feat(2023): read a crash dump far enough to say WHICH crash it is

artokun:
Does **not** fix panel#2023 — a Chromium DirectWrite null-deref cannot be patched from here, as the report says. It makes that report's **step one** doable.

## The step that was not getting done

> (1) symbolise the dump against Electron 40.4.1 symbols to confirm the DWrite call site

That needs a symbol server and a Windows debugging toolchain. So it does not happen, and reports arrive quoting a browser error page — which is how a crash that takes the ComfyUI server down several times a day has stayed at "we think it is DirectWrite".

Three facts decide whether two reports are the **same crash**, and all three are already in the file:

```
exception  0xC0000005 at 0x7ff8000001234
           EXCEPTION_ACCESS_VIOLATION — a read/write of an address the process does not own
faulting   C:\Windows\System32\DWrite.dll+0x1234
modules    91 loaded; text stack present: DWrite.dll, GDI32.dll, gdi32full.dll
```

No symbols required. It does **not** resolve function names — for that the symbols really are needed. It answers whether the fault is even in the module the hypothesis blames, which is panel#2023's whole question, and it moves that from "whoever has a debugger" to "whoever has the file".

## It already earned its keep

Run against the three Comfy Desktop dumps on this machine:

```
93f8fd3c….dmp   0xE0000008  a RAISED exception, not a hardware fault   KERNELBASE.dll+0xc1ada
7bb8dd78….dmp   0x80000003  EXCEPTION_BREAKPOINT — deliberate, NOT a memory fault
65e37e72….dmp   0x0517A7ED  Crashpad's SIMULATED code — a REQUESTED dump, not a crash
```

**None of them is panel#2023.** All three are from 2026-08-14, and the Crashpad metadata index has recorded nothing since — so the 08-24 and 08-29 dumps that issue names are not on this machine, in this database or a Desktop-2 one. I had guessed they might be, on matching file sizes (two ~33.5 MB main-process + one ~1.9 MB renderer, seconds apart). Same shape, different crash. That is exactly the mistake the tool exists to stop: reported as "the app crashed", those three are indistinguishable from the real one.

## Testing

Against **synthesised** dumps built to the documented layout (MINIDUMP_HEADER / _DIRECTORY / _EXCEPTION_STREAM / _MODULE_LIST), not a checked-in binary — a fixture states what the parser is *supposed* to read, where a 2 MB blob would only state what one machine happened to produce. 7 tests, including the join between the Electron log line and the dump: `-1073741819` **is** `0xC0000005` read as a signed 32-bit int.

Two deliberate behaviours are pinned:

- an address inside **no** loaded module reports `null`, not "unknown" — a jump through a freed pointer is itself a finding, and collapsing it into "the module list was unreadable" would hide it;
- a non-minidump returns `{ error }` rather than throwing. Being handed the wrong file is normal when chasing a crash and must not look like a tool failure.

Local read only, and the header says so — a crash dump contains process memory, so nothing writes or uploads.

## What this changes for panel#2023

Next occurrence: run it on the renderer dump **before Crashpad rotates it** and paste four lines. If the faulting module is not `DWrite.dll`, the hypothesis is wrong and a lot of speculative mitigation work is avoided. If it is, that is the first hard confirmation the issue has.


artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6). One real defect, and it is the kind that would have misdirected the triage this tool exists to support.

## The reporter threw away a distinction the parser worked to make

`faultingModule` is deliberately three-valued:

- an object — the address is inside this module;
- **`null`** — the list was read and the address matched **no** module. That is itself a finding: a jump through a freed pointer;
- **`undefined`** — the module list could not be read at all. We know nothing.

The parser even carries a comment saying collapsing the last two "would hide that". Then the description line tested it for **truthiness**:

```js
} else if (dump.exception) {
  lines.push("  faulting   <the address is inside NO loaded module>");
}
```

so `undefined` printed the `null` message.

**Shown by running it, not by reading it** — two dumps differing only in that field:

```
=== A: module list UNREADABLE (faultingModule undefined) ===
  faulting   <the address is inside NO loaded module>     ← false
=== B: address in NO module (faultingModule null) ===
  faulting   <the address is inside NO loaded module>
```

For a missing stream that is a **false diagnosis**: it reads as a freed-pointer jump when the truth is that the module list was absent. On a crash where the whole question is "which module", that sends the reader somewhere specific and wrong.

Now a strict `=== null`, with its own line for the unread case. Pinned — restoring the truthiness test fails it.

## What I checked and found sound

Every read is bounds-guarded and refuses rather than throws: signature, directory entries (`at + 12 > buf.length` breaks), the exception record, each module record, systeminfo, and the UTF-16 string reader (which also clamps its slice). That matters because the file's own contract says a truncated or non-minidump file is a normal thing to be handed and must not look like a tool failure — and an uncaught `RangeError` would.

`read-minidump`: **8 tests**. Lint clean.

artokun:
Ran this against three **real** Windows minidumps rather than only the synthetic fixtures — the tool exists to be pointed at a reporter's file, so the thing worth knowing is whether it survives one it did not construct. (Module paths redacted below: they are third-party DLLs on my box and there is no reason to publish an installed-software inventory.)

**A null-address fault — the case that matters most:**

```
exception  0xC0000005 at 0x0
           EXCEPTION_ACCESS_VIOLATION — a read/write of an address the process does not own
faulting   <the address is inside NO loaded module>
modules    72 loaded; text stack LOADED (not implicated): gdi32.dll, gdi32full.dll, DWrite.dll
verdict    <no faulting module resolved -- cannot say either way>
```

`DWrite.dll` is loaded and the tool **refuses to blame it**. That is the whole value: a text-rendering DLL is loaded in essentially every GUI process, so "DWrite was present" is not evidence, and the output says so in the same breath it lists it. A reader who wanted panel#2023 confirmed cannot get that from this line, which is correct.

**Two faults that DO resolve to a module:**

```
0xC0000005 at 0x7ff8c1e73896  ->  <third-party>.dll+0x323896
0xC0000006 at 0x7ffad7a6dfb6  ->  <third-party>.dll+0x15df
verdict (both)  the faulting module is NOT a text-rendering DLL -- not the panel#2023 shape
```

So the range matching and the base+offset arithmetic work on genuine 64-bit addresses, across dumps with 72, 117 and 140 loaded modules, and the verdict discriminates rather than always hedging.

**One cosmetic gap.** `0xC0000006` printed no plain-English gloss where `0xC0000005` did — it is `EXCEPTION_IN_PAGE_ERROR` (the pager could not read a page, classically a failing disk or a severed network mapping). Worth adding to the table, since it is a genuinely different diagnosis from an access violation and a reporter quoting the bare hex would not know that. Not a correctness issue; the raw code is still shown.

For panel#2023 this is now a tool the reporter can run and paste four lines from, which is a much smaller ask than symbolising against a symbol server.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **Ran it against four real dumps from two different producers, not fixtures.** The interesting result is that the honest-uncertainty paths fire on real data rather than only in tests.

## Two producers, both parsed

Windows Error Reporting dumps (`%LOCALAPPDATA%\CrashDumps`) and a Crashpad report (`%APPDATA%\...\Crashpad\reports`) are written by different code, and the Crashpad-only fields degrade correctly:

```
alacritty.exe.39468.dmp        (WER)
  exception  0xC0000005 at 0x0
  faulting   <the address is inside NO loaded module>
  process    C:\Program Files\Alacritty\alacritty.exe
  modules    72 loaded; text stack LOADED (not implicated): gdi32.dll, gdi32full.dll, DWrite.dll
  verdict    <no faulting module resolved -- cannot say either way>

OVRServer_x64.exe.11132.dmp    (WER)
  faulting   ...\oculus-runtime\OculusAppFramework.dll+0x323896
  verdict    the faulting module is NOT a text-rendering DLL -- not the panel#2023 shape

646fdeb5-...dmp                (Crashpad)
  faulting   C:\Windows\SYSTEM32\ntdll.dll+0xb9d3
  process    C:\Program Files\Meta Horizon\Support\oculus-client\Client.exe
  proc type  browser
  built by   Electron 32.1.1
```

`proc type` / `built by` appear only on the Crashpad dump and are **omitted**, not rendered as "unknown", on the WER ones — which is right: their absence is a property of the producer, not a failed read.

## The null-deref case is the one worth having

alacritty faulted at `0x0`, so no module owns the address. The reader says *"inside NO loaded module"* and the verdict declines — exactly the distinction the code comments argue for:

> NULL, not "unknown": an address inside no loaded module is itself a finding (a jump through a freed pointer), and collapsing it into the same answer as "we could not read the module list" would hide that.

That fired on a real dump I did not construct, which is better evidence than a fixture that the two cases stay distinguishable.

Also worth noting the verdict does not merely check whether DWrite is *present* — it is loaded in all three and the reader says so while ruling each out. Presence would have been the easy wrong predicate for #2023, and the tool avoids it.

## Parsing untrusted binary

Bounds-checked at every read, string reads capped at 4096 bytes, and the module loop is safe against a hostile `count`: `at` grows monotonically and the `at + MODULE_RECORD_BYTES > buf.length` break fires on the first overflow, so a `count` of 2^32 exits immediately rather than spinning. Nothing blocking.

artokun:
**Run against three real Comfy Desktop dumps that already existed on this machine — written 2026-08-14, months before this reader, so they are independent data rather than fixtures shaped to pass.**

```
65e37e72…  35.2 MB   0x0517A7ED  Crashpad SIMULATED code (a REQUESTED dump)   browser    91 modules
7bb8dd78…  35.2 MB   0x80000003  EXCEPTION_BREAKPOINT (deliberate abort)      browser    91 modules
93f8fd3c…   2.0 MB   0xE0000008  RaiseException, KERNELBASE.dll+0xc1ada       renderer   37 modules
```

All three: `the faulting module is NOT a text-rendering DLL -- not the panel#2023 shape`. Correct — none is an access violation inside a text DLL.

Three things this shows that a fixture could not:

- **The `proc type` split works on real files.** Two browser, one renderer, from the same directory. That is the line that matters for #2023, and it is exactly the confusion the reader was built to end: without it, the 2 MB renderer dump and the 35 MB browser dumps get described with identical confidence.
- **The DWrite nuance is right.** `DWrite.dll` is loaded in all three, and the reader says `text stack LOADED (not implicated)` rather than either ignoring it or treating its presence as evidence. On the renderer dump it discriminates correctly with only 37 modules loaded.
- **Exception classification is doing real work.** A Crashpad simulated-dump code, a deliberate breakpoint, and a raised (non-hardware) exception are three different things, and all three would read as "the app crashed" in a report. None is `0xC0000005`.

Environment note for #2023: this box's Comfy Desktop is **Electron 40.4.1** on Windows 10.0 build 26200.

**One robustness bug, low severity.** The CLI main-guard is path-based:

```js
process.argv[1].replace(/\/g, "/").endsWith("scripts/read-minidump.mjs")
```

A copy of the script under any other name prints **nothing** and exits **0** — indistinguishable from "ran fine, found nothing". I hit this myself: my first run against all three dumps produced empty output and I briefly took it for a parsing failure on real files. Absolute Windows paths are fine (the backslash normalisation handles them), so the documented invocation works; it is only the renamed/copied case. `import.meta.url === pathToFileURL(process.argv[1]).href` is the usual form and does not depend on the filename. Worth fixing because this script's whole purpose is to be handed to a reporter who will copy it somewhere.

For the record, the no-args path is correct — it prints usage and exits **2**. My first measurement said 0 because I read `$?` through a `| head`, which reports the pipe's status, not the script's.

artokun:
**The Copilot review on this PR was right, and the finding is worse than "incomplete stream validation" — the verdict ruled OUT the crash this tool exists to recognise.**

From #2023's own report:

```
EXCEPTION_ACCESS_VIOLATION at Comfy Desktop.exe+0x4147f6e, read of 0xd92a8
stack (module+offset scan): Chromium frames -> DWrite.dll x10 frames -> Chromium
```

The exception **address** is in `Comfy Desktop.exe`. `DWrite.dll` is a **caller**, ten frames up. This parser reads the exception record and the module list; it never walks the stack. So run against the reporter's own dump it printed:

```
verdict    the faulting module is NOT a text-rendering DLL -- not the panel#2023 shape
```

A false negative on the canonical instance. The verdict answered *"which module owns the faulting address"* and then pronounced on a hypothesis that lives somewhere it cannot see.

**Fixed:**

- the negative branch now reports the faulting module and says it does **not** rule out #2023, naming the reason (`its own dump faults inside Comfy Desktop.exe with DWrite ten frames up the stack, which this reader does not walk`);
- `text stack LOADED` becomes `text-rendering DLLs LOADED (module list, NOT call stack)` — "stack" was the wrong noun and invited the same wrong reading. That was Copilot's second comment;
- the positive branch softens from *"the panel#2023 shape"* to *"consistent with panel#2023"*: a text DLL owning the fault is strong evidence, not proof of identity.

**Two existing tests pinned the old behaviour**, one of them named `says NOT the #2023 shape when a text DLL is merely loaded` — the defect was asserted as correct, so the suite would have defended it. Both updated; three regression tests added against a dump shaped like the report's own (`0xC0000005` in the app binary with DWrite loaded). Mutation-verified: restoring the old sentence fails exactly those tests. 25/25, `tsc` clean.

**And a caveat on my own comment above.** I ran three real dumps from this machine through the old build and reported "none is the #2023 shape". That conclusion still holds — their exception codes are `0x0517A7ED`, `0x80000003` and `0xE0000008`, none an access violation — so it rests on the exception code, not the module verdict. But I quoted the module verdict as though it were load-bearing, and on a real `0xC0000005` it would have been wrong.

The remaining Copilot comment on this PR — that a truncated directory silently `break`s and reports a normal-looking dump instead of `{ error }` — is a real separate issue and I have not addressed it here.

artokun:
**Took the remaining finding on this PR, and the real behaviour is worse than it described.**

The claim was that a truncated dump "silently `break`s and returns a normal-looking dump". Tested against a 60,000-byte prefix of a real 2 MB dump from this machine:

```
RangeError [ERR_OUT_OF_RANGE]: The value of "offset" is out of range.
It must be >= 0 and <= 59996. Received 73928
```

The CLI died with a Node stack trace. One read of the crashpad stream had no bounds check while every other read in the file guards with `+ N <= buf.length` — and that is the file a reporter is most likely to hand over: an interrupted copy, or one still being written.

**Three changes:**

- the crashpad read is bounds-checked like its neighbours;
- a stream whose **data** extends past the end marks the dump truncated. Detecting only a short *directory* missed the real case, which is what my first attempt did — in a genuine truncated dump the header and directory land in the first pages and the payload is what is missing (this one's directory was intact with a crashpad stream at rva 73,928);
- truncation is reported **first**. Without that the file rendered as `exception <none recorded — not a crash dump>`, asserting the one thing an incomplete file cannot establish — the same overclaim this PR's verdict fix was about.

The three real dumps on this box are unaffected: none flagged, all three still classify.

**On the mutation, because the first pass was misleading.** Removing the crashpad guard originally passed **28/28** — the synthetic fixtures carry no crashpad stream, so no test reached the read that actually threw. The real dump proved it and the suite could not. There is now a fixture that does — a valid header and directory whose `crashpadInfo` entry points beyond the file — and removing the guard fails it.

That closes every finding on this PR.

artokun:
**Ran this against three REAL Comfy Desktop crash dumps, not fixtures.** This box turned out to have them at `%APPDATA%\Comfy Desktop\Crashpad\reports\` — two 35 MB browser-process dumps and one 2 MB renderer dump, all Electron 40.4.1 on Windows build 26200.

All three parsed. The classifications:

| dump | exception | reader's classification |
|---|---|---|
| `65e37e72…` (35 MB, browser) | `0x0517A7ED` | *Crashpad's SIMULATED code — a dump someone REQUESTED, not a crash* |
| `7bb8dd78…` (35 MB, browser) | `0x80000003` | *EXCEPTION_BREAKPOINT — a deliberate `__debugbreak`/abort, NOT a memory fault* |
| `93f8fd3c…` (2 MB, **renderer**) | `0xE0000008` | *a RAISED exception (`RaiseException`), not a hardware fault* |

**None of them is panel#2023**, and the reader says so without overclaiming in either direction. Every one printed the honest negative:

> the faulting module is NOT a text-rendering DLL. This does NOT rule out panel#2023: its own dump faults inside Comfy Desktop.exe with DWrite ten frames up the stack, which this reader does not walk.

That is the behaviour that matters most. A reader that answered "not the DirectWrite crash" on a dump it cannot actually exonerate would be worse than no reader — it would close the investigation on three dumps that simply are not crashes at all.

Two other things the real input confirms:

- **The module-list caveat is doing real work.** All three report `text-rendering DLLs LOADED … DWrite.dll, GDI32.dll, gdi32full.dll` — because DWrite is loaded in *every* Electron process. Without the explicit `(module list, NOT call stack)` qualifier, that line reads as evidence of DirectWrite involvement in all three, which would be false in all three.
- **Process-type detection works on both kinds** — `browser` for the two large dumps, `renderer` for the small one. Renderer is the process panel#2023 is about, so distinguishing them is load-bearing for triage.

Not established here: no dump on this machine is the DirectWrite crash, so the positive path — a dump that *is* panel#2023 — remains exercised only by fixtures.

PR #2811, opened 2026-09-03
fix(2782): do not Win32-resolve remote get_workflow strip paths

### LINKED ISSUE #2782: get_workflow strip resolves remote server path on local Windows host
Against a REMOTE Linux ComfyUI, get_workflow action:"strip" with an absolute server-side path resolves the path on the MCP host (Windows) instead of the connected ComfyUI host. The tool description explicitly says path is an "absolute server-side path", but the implementation converts /mydata/... into C:\mydata\... and performs a local filesystem read. This blocks retrieving a workflow downloaded server-side by ComfyUI-Manager.

### Steps to reproduce
1. Connect comfyui-mcp on Windows to a remote Linux ComfyUI.
2. Ensure a workflow exists at /mydata/workcode/ComfyUI/models/workflows/example.json on the remote host.
3. Call get_workflow {"action":"strip","path":"/mydata/workcode/ComfyUI/models/workflows/example.json","format":"raw"}.
4. Observe it attempts to open C:\mydata\workcode\ComfyUI\models\workflows\example.json locally.

### Exact error (scrubbed)
ENOENT: no such file or directory, open 'C:\mydata\workcode\ComfyUI\models\workflows\example.json'

### Fix
Precise change needed: when a remote ComfyUI target is active, do not pass a POSIX remote path through the MCP host's path.resolve / Win32 filesystem. Either proxy the read through a server endpoint/bridge on the connected ComfyUI, or reject remote absolute paths with an explicit "remote path reads unsupported" error. The current documented behavior and Win32 normalization are misleading.

### Environment
MCP host OS: Windows
Remote ComfyUI OS: Linux
ComfyUI: 0.34.0
ComfyUI frontend: 1.51.9 (startup log)
GPU: NVIDIA GeForce RTX 3090, 24 GB
comfyui-mcp: 0.52.179
panel: 0.15.152

---

Related (same root cause): #2528

artokun:
## Summary
`get_workflow` `action:"strip"` treated `path` as a file on the MCP host. Against a remote Linux ComfyUI, a POSIX absolute path such as `/mydata/workcode/ComfyUI/models/workflows/example.json` was Win32-resolved to `C:\mydata\...` and ENOENT'd.

When the connected ComfyUI is not this machine, skip local `readFile` / `path.resolve`. Library-shaped tails (`user/default/workflows`, `user/workflows`, `models/workflows`) are fetched from that server's userdata API. Any other remote absolute path is refused with the original POSIX path named — not a local Win32 ENOENT.

Fixes #2782

## Test plan
- [x] `npx vitest run src/__tests__/tools/get-workflow-strip-remote-path.test.ts` (and related workflow-library tests)
- Remote Windows MCP + Linux ComfyUI: `get_workflow {"action":"strip","path":"/mydata/.../models/workflows/example.json","format":"raw"}` must not open `C:\mydata\...`
- Local ComfyUI still reads absolute paths from this host's disk


artokun:
macOS lint, not tests: anti-slop `no-unknown-returns`, `workflow-library.ts` baseline 0 → 1. `fetchRemoteServerWorkflow` was declared `Promise<unknown>`.

Typing it turned up the same hole as #2793. The userdata endpoint answering **200 with `null`** — or an array, or a scalar — was handed straight back as a workflow, and the failure surfaced much later as an unreadable graph rather than as "that is not a workflow".

Now a named `FetchedWorkflowDocument`, a type-guard narrowing (no cast), and a refusal saying nothing was loaded. *Which* dialect the document is (UI graph vs API prompt) stays a downstream decision — that is all this layer can honestly claim.

- new pin: a `null` body is refused, not returned;
- mutation: dropping the guard fails that case and nothing else;
- `npm run lint` clean — 785 known findings, none added;
- `src/__tests__/tools`: 101 files, 1595 tests.

artokun:
All four review findings were real, and two of them are behaviour rather than wording. Fixed and pushed.

**1. `isAbsolute` is bound to the MCP host — and that is the same bug class this PR exists to fix, in the opposite direction.**

The mapper (`userdataKeyFromServerPath`) was already cross-platform; the *gate* deciding whether to call it was not. On a POSIX host a remote Windows drive path reads as **relative**, so the remote branch was skipped and the whole drive path went out as a userdata filename. Added `isServerAbsolutePath` (posix OR win32) at all three remote gates.

The local read deliberately keeps this host's own `isAbsolute`: a Windows path on a POSIX host is not something this machine can open, and must still fall through to the userdata lookup exactly as before.

**2. Comfy Cloud was classified as a local filesystem.**

`isRemoteMode()` is `!isCloudMode() && remoteUrlActive`, so it reads **false** in cloud mode — the locality predicate answered "local", and a `path` source `readFile`d an unrelated file on the MCP host while connected to Cloud. Cloud is now explicitly non-local. Good catch; this one had real exposure.

**3 + 4. The description contradicted the implementation.**

Reworded to separate unrestricted **local** reads from the three proxied remote library tails, and regenerated the page.

**Verification**

- new pins both directions: dropping the Cloud check fails the new Cloud case; the absoluteness helper is asserted against **both** spellings explicitly, so the result does not depend on which platform runs the suite;
- adding the `isCloudMode` import meant four sibling suites mocking `config.js` had to declare it — otherwise it is `undefined`, the call throws, and the tool reports `isError`, which is exactly how those four surfaced;
- full `npm test`: **768 files / 14135 tests**, plus `check:changelog`, `check:vocabulary` and the docs gates, all green after `npm run build`. Without the build, `package-hooks`, `asset-counts` and a `panel-image-relay` load flake fail — that is the unbuilt-worktree signature, not this change.

artokun:
Adversarial pass (mine — codex is out of tokens). The premise checks out; one sibling hole with the same symptom is left open, and I think deliberately.

**Verified — the cloud claim is exactly right.** `config.ts:932`:

```ts
export function isRemoteMode(): boolean {
  return !isCloudMode() && remoteUrlActive;
}
```

So in cloud mode `isRemoteMode()` is **false**, the old `!isRemoteMode() || targetIsOnThisMachine()` answered "local", and a `path` source `readFile`d an unrelated file on the MCP host. The explicit `if (isCloudMode()) return false;` is the right guard, and it cannot be folded into `isLocalMode()` — the function needs "local mode, OR remote pointing at this machine", which that predicate does not express.

**The sibling hole.** `comfyFilesystemIsLocalToThisProcess()` still returns true whenever `targetIsOnThisMachine()` does, and that resolves to:

```ts
if (forceRemote) return false;
return isOwnHostAddress(new URL(getComfyUIBaseUrl()).hostname);
```

A **RunPod pod reached through a loopback tunnel** is `127.0.0.1` by design in this project — that is how the pod's panel is fronted. Without `--force-remote`, `isOwnHostAddress` says yes, `comfyFilesystemIsLocalToThisProcess()` says local, and `get_workflow path:` reads *this* machine's disk while ComfyUI is on the pod. Same symptom the cloud bug had: a successful read of an unrelated file, with no error to notice.

The comment directly above `targetIsOnThisMachine` already names this ("the mirror of the note in panel-tools about a cloud pod fronted at 127.0.0.1") and settles it by saying the flag exists for the user to declare it and an inference must not overrule them. That is a defensible call — inferring "this loopback is really remote" is exactly the guess that would break a genuine local install.

So my question is scope, not correctness: **is the tunnel case deliberately out of scope here, resting on `--force-remote`?** If so it is worth one line in the PR body, because "do not Win32-resolve remote paths" reads as covering every remote, and a tunnelled pod is the remote most likely to be met in this repo. If a cheap guard is wanted later, the asymmetry to exploit is that the failure is silent — a `path` read that lands on a file whose content does not parse as a workflow could say "this looks like the MCP host's disk, not the target's" rather than returning it.

Issue #346, opened 2026-07-29
download_model writes to COMFYUI_PATH\models, ignoring the running server's --base-directory


artokun:
### What happened / root cause

`download_model` places files under `COMFYUI_PATH\models\<target_subfolder>`, but the running ComfyUI's models directory is determined by its `--base-directory` launch flag. On a ComfyUI Desktop install these are **different directories**, so the file lands somewhere ComfyUI never reads.

In this session:

- `COMFYUI_PATH` (env) = `M:\Comfy-Desktop\ComfyUI-Installs\ComfyUI\ComfyUI`
- running server argv = `--base-directory C:\COMFY` → real models dir is `C:\COMFY\models`

The tool reported success and returned the M: path. The model was invisible to ComfyUI; I had to `Move-Item` it by hand.

### Steps to reproduce

1. ComfyUI Desktop launched with `--base-directory C:\COMFY`, while `COMFYUI_PATH` points at the code checkout on another drive.
2. ```
   download_model({
     url: "https://huggingface.co/jags/yolov8_model_segmentation-set/resolve/main/face_yolov8m-seg_60.pt",
     target_subfolder: "ultralytics/segm"
   })
   ```
3. Tool result:
   ```
   Model downloaded successfully to:
   M:\Comfy-Desktop\ComfyUI-Installs\ComfyUI\ComfyUI\models\ultralytics\segm\face_yolov8m-seg_60.pt
   ```
4. `curl http://127.0.0.1:8000/models/ultralytics_segm` → still missing. The server's models dir is `C:\COMFY\models`.

### Exact error (scrubbed)

No error — reports success, writes to a directory the live server does not read.

### Fix

<applied locally: no>  <upstream-only: yes>

Resolve the download target from the **running server**, in this priority order:

1. The server's own view of its model dirs — `/models` + the folder-path registry — or its `--base-directory` argument read from the process command line.
2. Fall back to `COMFYUI_PATH\models` only when no live server is reachable.

A cheap correctness check that would have caught this: after the download, query `/models/<category>` (or `/object_info` for the consuming node) and warn if the new file does not appear. Since the tool already streams progress to the panel, a "downloaded, but this ComfyUI does not see it — expected models dir is X" warning would be a big improvement over a bare success message.

Same root cause family as the `list_extra_paths` / `add_extra_path` report: several path-resolving tools trust `COMFYUI_PATH` rather than the running server's launch args.

### Environment

[...]

PR #2813, opened 2026-09-03
fix(2787): accept proven extra-path base_path as shared model_root

### LINKED ISSUE #2787: Parallel CivitAI downloads intermittently reject valid shared model_root
During a batch of 11 parallel download_model action:"download_civitai" calls, all using the same valid model_root E:\models and target_subfolder "poses", 3 calls rejected the root while 8 concurrent siblings accepted it and wrote to E:\models\poses. The failing response's own known-root list contained E:\models\poses but omitted the configured base_path E:\models. A list_local_models action:"list_paths" call immediately afterward again reported base_path E:\models, and sequential retries of the same 3 requests all succeeded.

This is consistent with shared/mutable known-root state being replaced by a category-expanded snapshot during concurrent requests, or non-atomic root-cache refresh.

### Steps to reproduce
1. Configure ComfyUI Desktop with base_path E:\models and category poses -> poses/.
2. Confirm list_local_models action:"list_paths" reports base_path E:\models.
3. Start 11 parallel download_model action:"download_civitai" calls with model_root E:\models, target_subfolder "poses", distinct version IDs and filenames.
4. Observe intermittent MODEL_ERROR rejections for model_root E:\models while sibling calls succeed.
5. Call list_paths again, then retry rejected calls sequentially; they succeed to E:\models\poses.

Observed in one batch: 3/11 rejected in parallel; all 3 succeeded sequentially after list_paths.

### Exact error (scrubbed)
{"error":"MODEL_ERROR","message":"model_root is not a known model root: E:\\models. Known roots ... E:\\models\\poses (poses) ... Do not invent a path."}

### Fix
Keep known-root discovery request-local/immutable, synchronize cache refresh, and consistently retain configured group base_path entries alongside category-expanded roots. Add a concurrency regression test where multiple download_civitai calls validate the same configured base_path in parallel.

Temporary workaround: call list_paths, then issue downloads sequentially or with low concurrency.

### Environment
Windows 11
ComfyUI 0.34.3
ComfyUI frontend 1.49.6
Python 3.13.12
PyTorch 2.13.0+cu130
NVIDIA RTX 4090, 22.5 GB VRAM
comfyui-mcp 0.52.180
panel 0.15.159

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-sol` · effort `max` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

artokun:
## Summary
Parallel `download_model` `action:\"download_civitai\"` calls that all passed the same valid extra-path `base_path` as `model_root` could intermittently reject it. Known-root discovery listed only category-expanded dirs (`E:\models\poses (poses)`) and dropped the configured group `base_path` (`E:\models`) that `list_paths` reports.

Each proven group `base_path` is now retained as a generic `models` root alongside those category dirs. Discovery stays request-local (no shared mutable cache), uses the connected snapshot rather than a guessed `:8188` origin, and still fails closed for an unproven invented path.

Fixes #2787

## Tests
- `npx vitest run src/__tests__/services/download-extra-root-selector.test.ts src/__tests__/services/live-extra-roots.test.ts src/__tests__/services/extra-paths-observed-live-root.test.ts`
- 3 files, 41 tests passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project are documented here. This project adheres to
 
 ### MCP
 #### Fixed
+- **parallel CivitAI downloads accept a proven shared extra-path `base_path` as `model_root` (#2787).** Category expansion still lists `E:\models\poses (poses)`, but omitting the configured group `base_path` made concurrent `download_model` `action:"download_civitai"` calls reject that same valid root while siblings wrote into it. Known-root discovery stays request-local, never probes a guessed `:8188` origin, and still refuses an unproven invented path.
 - **manga-director-codex MiniMax H3 adapter accepts documented `text_to_video`.** Native H3 T2V is `MiniMaxH3ImageToVideo` with both image sockets empty; the adapter omitted that mode and compilation refused a valid prompt spec. `text_to_video` is declared; I2V / FL2VA / L2VA / R2V modes are unchanged (#2786).
 - **`panel_slice_workflow` seeds outputs inside nested overlapping groups (#2780).** Membership used the first containing box only, so a SaveImage inside both an outer group and a nested inner group was missed when slicing by the inner title. Any matching containing group now seeds, and a miss lists every containing title.
 - panel_search_nodes no longer returns a raw Git URL as an install `id`, and panel_install_node refuses Git URLs before Manager v4 queueing rather than sending an unlisted URL as a registry lookup (#1539)
--- a/src/__tests__/services/download-extra-root-selector.test.ts
+++ b/src/__tests__/services/download-extra-root-selector.test.ts
@@ -191,6 +191,80 @@ describe("#2499 extra-model root selector when the live server is unreachable", 
     ).rejects.toThrow(/not a known model root/);
   });
 
+  it("lands under the extra-path group base_path when that is the model_root (#2787)", async () => {
+    const comfy = await trackTmp();
+    const extraModels = await trackTmp();
+    await mkdir(join(comfy, "models", "poses"), { recursive: true });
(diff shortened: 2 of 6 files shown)


artokun:
Right, and it matters more than a path nit — thanks. Fixed.

`base_path` is the **anchor** for each configured category path, not a models root by itself. With

```yaml
shared:
  base_path: D:/AI
  checkpoints: models/checkpoints
```

ComfyUI registers `D:/AI/models/checkpoints`. Publishing `D:/AI` as a generic models root made `model_root=D:/AI` + `target_subfolder=checkpoints` resolve to **`D:/AI/checkpoints`** — a directory the server never scans — and it also accepted categories the group never declared. That is a download reporting success into a directory nobody reads, which is the #346/#369 failure mode this codebase already treats as serious.

The base is now published only when **every** model-serving category in the group sits exactly at `<base>/<category>`.

**This keeps #2787 fixed for the case it was filed for.** I went back to the issue rather than guessing: the reporter's config is `base_path E:\models` with `poses -> poses/`, which satisfies the rule. The existing #2787 test uses that same shape and passes untouched.

**A mixed group loses the base rather than exposing a partly-wrong tree.** The generic entry carries one label meaning "any subfolder", so it cannot be right for `poses` and wrong for `checkpoints` at the same time. A rejected `model_root` is a refusal the caller can act on; a wrong directory is a download the server never sees.

Pinned both directions — the remapped and mixed groups fail without the guard, and the reporter's shape passes with it. `src/__tests__/services`: **275 files, 6491 tests** (one failure, `panel-image-relay`'s deadline case, passes in isolation).

Issue #852, opened 2026-08-05
Test suite: name the load-flaky tests so agents stop debugging non-bugs (grok-backend, training-jobs CAS)


artokun:
## What happened

Agents working in this repo keep reporting that full-suite runs "flake in a rotating cast of unrelated files", and work around it with `--maxWorkers=2`. Stated that way it is unactionable, and it has a real cost: **an agent cannot tell a load flake from a real failure**, so it either wastes a cycle debugging a non-bug or — worse — "fixes" something that was never broken.

I measured it rather than leaving it as folklore. Full-concurrency run on `main` (`f7b5b2c`, 0.49.6) while this machine was heavily loaded:

```
FAIL  src/__tests__/orchestrator/grok-backend.test.ts >
        buildAcpMcpServers > serializes HTTP panel MCP as SSE with empty headers (Grok rejects type:http)
FAIL  src/__tests__/services/training-jobs.test.ts >
        independent review fixes (PR #237) > two non-owner processes racing recovery hand off EXACTLY once (CAS)

Test Files  2 failed | 249 passed (251)
     Tests  2 failed | 4995 passed | 2 skipped (4999)
```

**Both pass in isolation** (`7 passed` and `69 passed | 1 skipped` respectively), which is what makes them flakes rather than failures.

## Root cause

Not shared state. I checked the obvious candidates and ruled them out:

- `vitest.config.ts` uses default isolation, so each test *file* gets its own environment — the 10 test files that mutate `process.env` without restoring cannot leak across files.
- No test binds a fixed port; the `127.0.0.1:8188` occurrences are mocked URLs, not real binds.

What is left is **timeout starvation under CPU contention**. The config sets no explicit `testTimeout`, so the 5s default applies, and on a machine running many concurrent agents that is easily exceeded by tests doing real async work. The `training-jobs` CAS case is the most susceptible by construction — it races two module instances through `Promise.all` and asserts an exactly-once handoff, so it depends on real scheduling.

## Why this is worth fixing rather than tolerating

A suite that cries wolf eventually masks a real break. Two of these have already been fixed properly this week rather than tolerated:

- **#821 / PR #843** — a fixture bound a *guessed* random port and hung ~1 in 8. Fixed with `listen(0)` + reading the port back, so the collision mechanism is gone rather than rarer.
[...]

PR #2814, opened 2026-09-03
fix(2788): reclaim abandoned panel-op.lock when the owner process is gone

### LINKED ISSUE #2788: Reclaim stale panel-op.lock automatically when owner process is gone
While installing a custom node through the ComfyUI MCP installer, the operation was blocked by an abandoned panel operation lock.

Observed:
- Lock: C:\Users\<user>\.comfyui-mcp\panel-op.lock
- Recorded owner PID: 5840
- The owner process no longer existed.
- Lock age was about 64 minutes.
- Further install_custom_node operations remained blocked until I explicitly ran install_comfyui(action:"panel", panel_action:"unlock"), which returned outcome "reclaimed".
- After reclaiming the lock, operations proceeded normally.

Expected:
The installer should automatically reclaim a panel-op.lock when its recorded PID is no longer alive, or the blocking error should expose an actionable stale-lock recovery path.

Environment:
- Windows 11
- ComfyUI 0.34.0
- ComfyUI frontend 1.49.6
- comfyui-mcp 0.52.174
- panel 0.15.155
- Python 3.12.12
- PyTorch 2.13.0+cu130
- NVIDIA GeForce RTX 5070, 12 GB VRAM
- Local-git ComfyUI deployment

---

Related (same root cause): #760, #1953

artokun:
A crashed or killed orchestrator left `panel-op.lock` behind. Acquire then waited out the budget (or fast-refused) and named `panel_action:"unlock"` as the recovery, so `install_custom_node` stayed blocked until someone ran unlock by hand.

Acquire now reclaims that lock when the recorded owner PID is proven gone, through the same rename-aside + byte-compare path as `panel_action:"unlock"`. A living owner is never stolen. An unreadable or reuse-ambiguous record still fails closed.

Fixes #2788


artokun:
All six findings were real. One is behaviour, four are the same wrong claim in four places, one is test determinism. Fixed and pushed.

## The reclaim poll was spawning a synchronous subprocess every second

`ownerProvablyDead` reaches `processTableHasPid`, which runs `tasklist.exe` (3s timeout) or `ps` through **`execFileSync`** — on the event loop. With a live cross-process holder that fired once per second for the whole acquisition budget. Confirmed by reading the call path, not assumed.

What makes this cheap to fix is an asymmetry the finding points at: a genuinely **dead** owner is proven by `kill(0)` alone (`alive === false`) and never reaches the process table. The expensive lookup only matters for an EPERM-unsure pid or a Windows handle-kept-open zombie — cases reclaim **refuses anyway**. So the poll stays cheap at 1s, the table lookup runs on its own 15s clock plus once more at the deadline, and dead-owner reclaim latency is unchanged.

That claim is the entire justification, so it is pinned in both directions: a dead owner is still reclaimed with the probe withheld, and withholding it never turns an unproven owner into a reclaim. Moving the opt-out ahead of the `alive === false` branch fails the first.

## `unlock` was described as a recovery it cannot perform

The timeout message, the `install_comfyui` description, the install-panel comment and the generated docs page all called `panel_action:"unlock"` the recovery for a lock whose owner cannot be proven. It is not — unlock calls the same reclaim function, which refuses unreadable and reuse-ambiguous records, i.e. **precisely the cases that reach that timeout**. All four now describe it as an on-demand re-check, and name manual cleanup as the actual recovery.

Worth saying plainly: this is the same shape as panel#2139's "nudge the canvas" — a message offering advice that cannot work for the cause that produced the message.

## Test determinism

`999999` is reachable where `pid_max` is raised (Linux allows up to 4194304), so that case could observe a live process and never exercise reclaim at all. Switched to `0x7fffffff`, the sentinel the rest of the file already uses.

## Verification

Full `npm test` after `npm run build`: **768 files / 14136 tests**, every gate green except a rotating cast of 3 unrelated files that pass in isolation and **differ between two runs of the same commit** — the #852 concurrency signature this repo's vitest config documents. My diff touches `panel-pin-guard` and two description strings.

Issue #2703, opened 2026-09-01
get_history cannot reach connected local ComfyUI target


artokun:

artokun:
Claimed 2026-09-01T14:54Z — dispatched to a fix agent (claude autopilot).

Taking this one. Working in `.claude/worktrees/wt-2703-cc` on branch `fix/2703-history-panel-origin-cc`.

First finding while triaging, recorded here so the next reader does not re-derive it: the string `PANEL_FETCH_FAILED` quoted in the report does not exist anywhere in either repo, and `get_history` has **no** connected-panel fallback at all — `getHistory()` issues one `comfyApiFetch("/history")` against `COMFYUI_URL` and that is the whole path. So the report's premise ("its connected-panel fallback also returned PANEL_FETCH_FAILED") is not what happened; the constraint underneath it — the headless `/history` read cannot reach the ComfyUI a connected panel is demonstrably talking to — is the part I am checking.

artokun:
Correction to my claim comment above, which was read off a stale checkout and is wrong in the part that matters.

`PANEL_FETCH_FAILED` **does** exist and the connected-panel read fallback **does** exist — both landed after the tree I first grepped. On `origin/main`:

- `getHistory` falls back to the authenticated panel read relay (`fetch_comfyui_read`, #2532/#2283) when the headless `/history` fetch fails at the transport layer.
- `panelReadFallback` (src/comfyui/client.ts) formats the failure as ``` `${primary} The connected panel ComfyUI read fallback failed safely (${code}).` ``` — which is the report's error text, verbatim in shape. The report is accurate; my first reading of it was not.

## What actually happened

Reproduced the primary half on `origin/main` (dead `COMFYUI_URL`, a live panel origin published on the orchestrator→child channel):

```
fetch failed: connect ECONNREFUSED 127.0.0.1:56214 (ECONNREFUSED) — while requesting
http://127.0.0.1:56214/history?clientId=comfyui-mcp (GET). … A connected panel is on
http://127.0.0.1:56213 — a DIFFERENT address …
```

So the headless leg fails and says so. The fallback leg is the one that went dark, and this is why:

[...]

PR #2815, opened 2026-09-03
fix(2785): honor get_image preview downscale for oversized PNGs

### LINKED ISSUE #2785: get_image ignores preview downscale when source PNG exceeds 32 MB
`get_image` refuses a 41.5 MB PNG before applying its documented `max_preview_dimension` downscale. The byte guard appears to run on the original /view response, so a requested 1500px preview cannot be returned.

### Steps to reproduce
1. Generate a large PNG above 32 MB (here: 4320×5376, 43,488,140 bytes).
2. Call `get_image` with `action:"get"`, the output filename/subfolder, and `max_preview_dimension:1500`.
3. Observe `VIEW_TOO_LARGE` instead of a downscaled inline preview.

### Exact error (scrubbed)
`VIEW_TOO_LARGE: ComfyUI /view response for "NC04_4X_UPSCALE_00002_.png" exceeds the 32 MB safety limit.`

### Workaround
`get_image` action `convert` to JPEG succeeds and returns a 6.4 MB preview, but it is an extra destructive-format conversion rather than honoring `max_preview_dimension`.

### Expected
Fetch/decode the source server-side, apply `max_preview_dimension`/byte budget, then return a downscaled preview when requested.

### Environment
Windows 11; AMD Radeon 780M (12 GB unified VRAM); ComfyUI 0.28.0; comfyui-mcp 0.52.165; panel 0.15.102.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-terra` · effort `high` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #1495

artokun:
## Summary

`get_image` refused a 41.5 MB PNG (`VIEW_TOO_LARGE`) before the documented `max_preview_dimension` downscale could run. The `/view` byte guard was applied to the original body, so a requested 1500px preview never happened.

Still-image `get_image` fetches that will be previewed now use a **64 MB** encoded-source ceiling (the same bound as `action:"convert"`). A local file in that window is recovered without an unbounded read. Sources still over 64 MB fail closed with a size error that names `convert`. Videos and attachments keep the 32 MB `/view` cap.

Fixes #2785

## Test plan

- [x] Targeted vitest: oversized PNG preview, image-management fallback, cloud/headless `/view` caps, existing get_image/image-assets tests
- [ ] Reproduce with a PNG > 32 MB and `get_image action:"get"` + `max_preview_dimension:1500` — expect a downscaled inline preview, not `VIEW_TOO_LARGE`


artokun:
Two of the three were right and are fixed. The third I could not reproduce, and I think it is incorrect — evidence below rather than an assertion.

## ✅ The preview-specific error was applied to videos and attachments

`forInlinePreview` is set for every `get`, but the 64 MB ceiling is selected only for `forInlinePreview && IMAGE_EXTENSIONS.has(ext)`. A 40 MB MP4 therefore kept the ordinary 32 MB cap and was still told about `max_preview_dimension` — a preview nobody attempted — and `action:"convert"`, which is image-only.

The cap choice is now named `usedPreviewCap` and gates the advice; everything else keeps the original size error. Pinned, and ungating it fails the new case.

## ✅ `maxBytes` does not reach the relay — and it is worse than a missing argument

Confirmed: `fetchImage` calls `requestPanelImage(filename, type, subfolder)` with no limit. But raising MCP's own validator would have been a **no-op**, because the panel enforces `MAX_FETCH_IMAGE_BYTES = 32 MB` in `fetch-image.js` at three points (declared content-length, streaming total, final byteLength) — the payload is refused before MCP sees it.

I deliberately did **not** patch around this. The tempting one-repo shortcut is to detect the relay's own too-large and run the same local recovery, but the panel's `too_large` arrives as `PANEL_FETCH_FAILED` with panel-authored **prose**, and `panel-image-relay.ts` states outright that nothing parses it and nothing branches on it. Branching on panel text to decide whether to read a local file is precisely what that comment exists to prevent.

So it needs a structured signal across both repos — filed as artokun/comfyui-mcp-panel#2217, with two options (a typed refusal code, which also helps #2703; and a clamped per-request limit, which is what actually lets 33–64 MB through). Meanwhile the changelog entry here now says the recovery is the **local** path, so nothing promises behaviour this PR does not deliver.

## ❌ The 40 MB test does discriminate

The claim is that it would pass even if the reader stayed hard-capped at 32 MB. It would not: `readLocalViewFallback` checks `candidateStat.size > maxBytes` **before** the handle is opened, so the mocked 40 MB stat is what decides — the tiny handle payload is irrelevant to the bound.

Verified by mutation rather than by reading. Forcing `maxBytes = MAX_VIEW_RESPONSE_BYTES` inside `readLocalViewFallback`:

```
× reads a 40 MB local PNG after /view hits VIEW_TOO_LARGE so preview can run
  Tests  1 failed | 96 passed (97)
```

So that test already fails if the parameterization is lost, and a 33 MB fixture would add allocation cost to a suite that is already load-sensitive without adding discrimination. Happy to add one if you still want the payload asserted end to end — say so and I will.

`src/__tests__/services`: **275 files, 6494 tests** (one failure, `panel-image-relay`'s deadline case, passes in isolation — the #852 concurrency signature).

PR #2823, opened 2026-09-03
test: stop the relay deadline test spending its budget before dispatch

artokun:
## Summary

`panel-image-relay > carries one deadline through the bridge and refuses a late panel result` failed under full-suite concurrency on **five different branches today**, always passing 45/45 in isolation. Each time it cost a round of investigation on an unrelated PR, so it was actively polluting review signal.

## Mechanism (confirmed, not guessed)

The request carries `deadlineAt = createdAt + 50`, and that budget is **wall clock**, starting when the file is written. On a loaded runner more than 50 ms elapses before `processPanelImageRequests` reaches it — so the request is refused as **already expired**, the bridge is never called, and the assertion reads the uninitialised `timeoutMs` as `0`.

That is why the symptom (`expected 0 to be greater than 0`) reads like a deadline-propagation bug and is not one.

The budget is now 400 ms, with the panel answering at twice that, so the TIMEOUT verdict the test asserts is still real.

## Honest limit

**This reduces the failure but does not eliminate it.** Measured on one machine, same suite: **~5/5 runs failed before, 1/3 after.** The mechanism is right — 400 ms simply isn't always enough when 275 files run at once — so the residue is the same starvation, further out. I'd rather ship the improvement with the number attached than claim a cure.

Widening is the sanctioned move *here* and not elsewhere because what is starved is the measurement apparatus, not a race in the code under test — which is the distinction `vitest.config.ts` already draws about #852.

## Not claimed

Two other suites (`divergent-install-reachable`, `panel-secrets-slots`) fail reproducibly under the same local full-suite load and pass in isolation, but are **green on CI**. I'm recording that as local contention, not a claim about `main`.

Test-only change.


artokun:
Adversarial pass (mine — codex is out through Sept 6). **The diagnosis is right and I can confirm the cost first-hand. The remedy is the one your own config file argues against at this scale.**

## Confirming the cost

This flake hit me twice today, on #2802 and during the `npm test` run on #2840, and both times it cost a controlled investigation — re-running at `HEAD~1`, checking imports — before I could set it aside. Your "actively polluting review signal" is not overstated; it is the second-largest source of noise I hit today after the Manager-dialect probes (#2829).

The mechanism reads correctly too: `expected 0 to be greater than 0` is the uninitialised `timeoutMs`, not a propagation bug, because `send` was never reached.

## Why 400 ms may not be enough

The comment cites `vitest.config.ts` on #852 as precedent, and that file is the strongest argument *against* this shape of fix. It raised the per-test timeout **5s → 30s** because this runner starves tasks for **seconds**, and it closes with:

> If a test starts needing MORE than this, that is a signal to make it deterministic (inject the clock, remove the real timer), not to raise this

`createdAt` is captured at line 447 and `processPanelImageRequests` is called at 461, so the exposed window is a deschedule between those two points — and on a machine documented to stall tasks for seconds, 400 ms of wall clock is inside the known stall profile. This makes the failure rarer, not impossible, and a rare version of this flake is worse than a frequent one: it will reappear on someone else's unrelated PR months from now with no memory of this thread.

## The deterministic split, which is cheap here

The test currently asserts two independent things through one timing window:

1. **the deadline is propagated** — `timeoutMs > 0` and `<= budgetMs`
2. **a late answer is refused** — `{ ok: false, error: "TIMEOUT" }`

(1) does not need a tight budget at all: give it 30s and it can never be starved, while still proving propagation. (2) does not need a real timer: write the record with `deadlineAt` already in the past and assert the refusal directly — no `setTimeout`, no wall clock, and it tests the expiry path more directly than waiting for it to elapse.

Split that way, neither test can be starved, and together they cover strictly more than the current one.

Not blocking — this is a real improvement over 50 ms either way, and the "apparatus, not product" distinction is correctly drawn. I would just rather this flake be closed than made rare.

artokun:
Pushed the split as `01741c3`, and I want to be precise about which half I proved and which I did not.

## Verified

**Propagation** — mutation-verified: forcing `timeoutMs: 0` at the dispatch site kills it. And it can no longer be starved, because the budget is now **`PANEL_IMAGE_RELAY_TIMEOUT_MS` (8s), the largest the validator accepts**. That is worth knowing on its own: `deadlineAt - createdAt > PANEL_IMAGE_RELAY_TIMEOUT_MS` is rejected as malformed, so my first attempt at a deliberately huge 30s budget never reached the bridge at all and failed with the *identical* `expected 0 to be greater than 0` — the same symptom, a completely different cause. The ceiling was available the whole time: 8s rather than 400ms is 20x the headroom for free.

**Stability** — 3x in isolation (46/46 each) and, more to the point, the full services suite under the concurrency that produced the flake: **278 files, 6,503 passed, 0 failed**, `panel-image-relay` included.

## NOT verified, stated plainly

I could not kill the expiry half. Disabling the intake gate's deadline clause (line 1217) left it green; disabling the later `remainingMs <= 0` check (1699) left it green; disabling **both together** still left it green. So there is a third refusal path I did not locate, and by my own standard an assertion I have not watched fail is not yet a test.

It is not wrong — it passes, it asserts real behaviour (`sendCalls === 0` and a `TIMEOUT` verdict), and it needs no timer. I just cannot claim it pins anything. Treat it as a regression witness, not as coverage, until someone finds the gate that owns that path.

## One real bug found in my own draft

The first version used `createdAt = Date.now() - 60_000`. The intake gate is `age > PANEL_IMAGE_RELAY_STALE_MS || now >= requestDeadline(request)` — 60s old trips **staleness**, so it never reached the deadline clause and passed for a reason its own name contradicted. `createdAt` is now 1s old with a 50ms budget: comfortably inside the 15s staleness window, and its deadline long gone. The mutation attempts are what exposed that; the test was green throughout.

artokun:
Closing the loose end from my last comment, with a sharper answer than "a third path I did not locate".

**The expired-request refusal is guarded in at least FOUR independent places on this path.** I found three and disabled all of them together:

```
1217   if (age < -5_000 || age > STALE_MS || now >= requestDeadline(request))
1680   if (age < -5_000 || age > STALE_MS || now >= request.deadlineAt)      <- the one I missed
1699   if (remainingMs <= 0)
```

With all three off, the result is still `sendCalls: 0` and `error: "TIMEOUT"`. So a fourth site owns it, and 1217/1680 are near-duplicates of each other — the same predicate written twice against two different accessors for the same field.

## Why this is worth knowing beyond my test

**A single-site regression here is invisible to the suite by construction.** Remove any one of these four and everything stays green, because the next one catches it. That is defence in depth working as designed for the *product*, and it is also why no test can pin any individual gate: the behaviour is over-determined.

Two practical consequences:

- **My expiry test is a regression witness, not coverage**, and now I can say why rather than just flagging uncertainty. It proves an expired request is refused without dispatch. It cannot prove *which* check did it, and no test written against this path can, until the gates are consolidated.
- **The duplication at 1217 and 1680 is itself worth a look.** Two copies of `age < -5_000 || age > STALE_MS || now >= <deadline>` differing only in how they reach the deadline is the shape that drifts — one gets a fix and the other does not, and nothing fails. That is the same "fix one exit, leave its siblings" pattern #2696 handled explicitly, and I would rather flag it than let it sit behind an over-determined green.

Not asking for either in this PR — it is a test-hygiene change and this is product structure. The split and the 8s ceiling stand on their own.

Issue #2764, opened 2026-09-02
managerFetch: the body read is not raced against the abort signal, so a Manager route that stalls after headers hangs the caller


artokun:
Split out of #2754 / PR #2762 rather than widened into that PR — it is pre-existing
on `main` and applies to every Manager call site, not to the branch that found it.

### The mechanism

`comfyuiFetch` supplies a default ceiling (`defaultComfyTimeoutSignal()`) for callers
that pass no signal, and `managerFetch` passes none. But the repo already documents that
this ceiling does not cover the body — `src/comfyui/fetch.ts:575-584`, from #1672:

> `AbortSignal.timeout` on `fetch` only cancels the HTTP exchange. Once headers have
> arrived, `Response.text()` / `JSON.parse` can keep the caller pending for as long as the
> body takes to finish … The abort event IS fired; the decode does not observe it.

That is exactly why `raceAbort()` exists (`fetch.ts:590`). `managerFetch` does not use it:

```ts
// src/services/node-management.ts — managerFetch
const text = await res.text().catch(() => "");   // the !res.ok path
...
const raw = await res.text();                    // the 2xx path — unguarded
```

So a ComfyUI (or a proxy) that accepts a Manager request, returns headers, and then
stalls the body leaves the caller pending indefinitely. A ComfyUI mid-decode does exactly
this — that is the observed behaviour #1672 was filed for.

### Blast radius

Every `managerFetch` call, which is every Manager operation in `node-management.ts`:
dialect detection (`/v2/manager/queue/status`, `/manager/queue/status`,
`/v2/manager/version`, `/manager/version`), every queue enqueue, every poll. Dialect
detection is the worst of them, because it is the first thing an
`install_custom_node` / `download_model` / `apply_manifest` call does — a stall there
hangs the tool before anything has been sent.

### Suggested fix

Wrap both body reads in `raceAbort(signal, …)`. `managerFetch` does not currently hold the
signal (`comfyuiFetch` mints it internally when the caller passes none), so this needs
`managerFetch` to mint its own and pass it down, or `comfyuiFetch` to expose the signal it
used. The second is the smaller change at the call sites but touches a shared helper.

### Not a duplicate of #1672

#1672 fixed `call_tool`'s own hang and shipped `raceAbort`. The Manager path was never
converted to use it.

Found while working #2754 (codex gate round 3, finding 1). Reported rather than fixed
because the correct fix changes behaviour for every Manager call, which is a different
[...]

PR #2827, opened 2026-09-03
fix(2773): name Manager body-read timeouts; correct raceAbort comment

### LINKED ISSUE #2773: fetch.ts raceAbort comment states a mechanism that is false on Node >=22, and a Manager body-read timeout escapes undiagnosed
Split out of #2764, which was closed as not reproducible. The measurement is in that thread; this is the part of it that IS real.

### 1. `fetch.ts:575-584` documents a mechanism that does not hold on supported Node

The `raceAbort` doc comment asserts:

> `AbortSignal.timeout` on `fetch` only cancels the HTTP exchange. Once headers have arrived, `Response.text()` / `JSON.parse` can keep the caller pending for as long as the body takes to finish … The abort event IS fired; the decode does not observe it.

Measured against a real localhost server that sends headers and then never writes a body, across four stall shapes (chunked stall read immediately; chunked stall where the signal fires before `res.text()` is called; `Content-Length` declared but under-sent; a body that trickles forever):

- **Node 22.23.2** — every shape `rejected` at the deadline.
- **Node 24.16.0** — every shape `rejected` at the deadline.

`comfyuiFetch` attaches the ceiling to `fetch` itself (`fetch.ts:690`), and undici errors the response body stream when that signal fires, so a pending `Response.text()` rejects. `engines` is `>=22` and CI runs 22, so there is no supported version where the quoted sentence is true for `Response.text()`.

It is also not what #1672 observed. #1672's reproduction reads *"The nested call reports 'The operation was aborted due to timeout', but the outer script remains running"* — the inner read observed the abort and rejected; the enclosing `call_tool` cell is what hung. `raceAbort` earns its place by bounding the **enclosing promise**, which is a real and different job.

This matters because the comment is load-bearing: it is quoted as established fact by other work. It produced #2764, and it shaped the comment block in #2762 (`node-management.ts:747-758`). Correcting it to describe what `raceAbort` actually does — bound an enclosing promise whose inner work may not settle — would stop that.

`JSON.parse` should be kept in the corrected text, but for the right reason: it is genuinely unabortable, and `raceAbort` cannot preempt it either, because it blocks the same thread.

### 2. A Manager body-read timeout escapes undiagnosed

Distinct from the above and much smaller. When a Manager response stalls its body, the abort fires **after** `comfyuiFetch` has already returned, so the `describeComfyTimeout` rewrite at `fetch.ts:757` never runs. The caller gets:

```
TimeoutError: The operation was aborted due to timeout
```

with no URL and no method, where a transport-phase timeout on the same call would have produced the full "No reply from ComfyUI within Ns — while requesting <target> (<method>)" text plus the OUTCOME-UNKNOWN clause for a mutation.

Worth stating plainly: this is **not** a write hazard. `redetectAfterDialectMismatch` gates its retry on `errorStatus(err)`, a structured field, so a bare abort carries no status and cannot trigger a re-enqueue. It is a legibility gap on a path that already fails safely.

If it is fixed, the shape that works is to mint the ceiling inside `managerFetch` and race only the two body reads — **not** to pass a signal down into `comfyuiFetch`. `comfyuiFetch` applies `describeComfyTimeout` only when `init.signal === undefined` (`fetch.ts:757`), so threading a signal through would silently downgrade the transport-timeout message for every Manager call while fixing the body-read one.

### Not filed as a bug against behaviour

Both are documentation/diagnosis quality under the freeze, not defects in what the code does. Filing so the finding is not lost, not to request a build.

artokun:
## Summary

Fixes #2773 (split from #2764). Two parts, both documentation/diagnosis quality:

1. **`raceAbort` comment was false on supported Node.** `AbortSignal.timeout` on `fetch` does **not** leave `Response.text()` pending after headers: undici errors the body stream when the fetch signal fires. Measured on Node 22.23.2 and 24.16.0 (`engines` is `>=22`). `raceAbort` still bounds an enclosing promise whose inner work may not settle; `JSON.parse` is unabortable because it blocks the same thread. #1672's inner read already rejected; the outer `call_tool` cell is what hung.

2. **Manager 2xx body-read timeout escaped undiagnosed.** After headers, `comfyuiFetch` has already returned, so `describeComfyTimeout` never runs and the caller got a bare `TimeoutError` with no URL and no method. `managerFetch` now mints its own ceiling and races **only** the body reads — it does **not** pass a signal into `comfyuiFetch`, which would skip the transport-timeout rewrite (`init.signal === undefined`). A stall fails closed as `manager-body-timeout` with URL, method, and OUTCOME UNKNOWN on mutations. No status field, so dialect-mismatch retry cannot re-enqueue.

## Tests

- `src/__tests__/services/manager-body-timeout.test.ts` — GET `listInstalledNodes` names the stall; POST `startManagerQueueForExternal` is OUTCOME UNKNOWN; a headers-never-arrive hang still gets the transport rewrite.
- Existing dialect-cache / fetch-timeout / #1672 unwind suites still pass.

## Changelog

Unreleased → MCP → Fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project are documented here. This project adheres to
 ### MCP
 
 #### Fixed
+- **Manager body-read timeouts fail closed with a named diagnosis instead of a bare `TimeoutError` (#2773).** After headers arrive, undici still aborts a stalled `Response.text()` on supported Node, but `comfyuiFetch`'s transport rewrite has already returned. `managerFetch` now races both body reads against its own ceiling (without threading a signal into `comfyuiFetch`, which would drop the transport-timeout message) and reports URL, method, and OUTCOME UNKNOWN on a mutation. The `raceAbort` comment no longer claims the fetch signal leaves `Response.text()` pending.
 - **`panel_call_tool` recovers a unique inner panel tool from a self-referential call (#2717).** `name: "panel_call_tool"` with the real tool on `tool` / `tool_name`, a nested key, or sibling fields unwraps once and runs that tool. Several inner names, a non-object keyed payload, or a nested router still fail closed; the refusal names the inner tool when it is unique. The schema now says `name` is the inner tool, never this router.
 - **`panel_set_widget` accepts a complete promoted-subgraph ownership envelope that omits `truncated:false` (#2783).** `graph_get_subgraph` already proves completeness when `node_count === nodes.length`; an omitted or null `truncated` flag is then `false`. An asserted `truncated:true`, or a list shorter than `node_count`, stays fail-closed.
 
--- a/src/__tests__/services/install-git-manager-listed-but-absent-2714.test.ts
+++ b/src/__tests__/services/install-git-manager-listed-but-absent-2714.test.ts
@@ -92,13 +92,15 @@ const manager = vi.hoisted(() => ({
   installed: {} as Record<string, unknown> | unknown[],
   calls: [] as string[],
 }));
-vi.mock("../../comfyui/fetch.js", () => {
+vi.mock("../../comfyui/fetch.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../../comfyui/fetch.js")>();
   const json = (body: unknown) =>
     new Response(JSON.stringify(body), {
       status: 200,
       headers: { "content-type": "application/json" },
     });
   return {
+    ...actual,
     comfyuiFetch: async (url: string) => {
       manager.calls.push(url);
       const { pathname } = new URL(url);
--- a/src/__tests__/services/manager-body-timeout-parity.test.ts
+++ b/src/__tests__/services/manager-body-timeout-parity.test.ts
@@ -1,0 +1,36 @@
+// #2773 — the Manager body-read deadline is derived a SECOND time.
+//
+// `managerBodyTimeoutSignal` cannot call `defaultComfyTimeoutSignal`: suites that
+// mock `comfyui/fetch.js` do not declare that export, so importing it would make
+// those files fail to load. The workaround is a local copy of the derivation —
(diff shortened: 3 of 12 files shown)


artokun:
Self-adversarial review (codex out of tokens; the pass is mine until Sept 6).

## The central claim, re-measured rather than inherited

This PR replaces a docblock that said `AbortSignal.timeout` on `fetch` "only cancels the HTTP exchange", leaving `Response.text()` free to hang. I had measured that as false once before, but on a different Node — so I measured it again here, against a server that sends headers and then stalls mid-body:

```
headers arrived at 10 ms; now reading body…
text() REJECTED after 606ms: TimeoutError — The operation was aborted due to timeout
```

undici errors the body stream when the signal fires; the read rejects **at** the deadline. And the new comment's supporting claim checks out too — `engines` really is `>=22.0.0`, so that is the floor for every supported install, not an assumption.

## One finding: a silently-drifting duplicate

`managerBodyTimeoutSignal` re-derives the deadline locally, and the reason is legitimate — it cannot call `defaultComfyTimeoutSignal`, because suites that mock `comfyui/fetch.js` do not declare that export and would fail to load.

But the copy carries its own `120`, with the same `raw > 0 ? raw : 120` rule, and **nothing compared the two**. Raising `DEFAULT_COMFY_HTTP_TIMEOUT_S` would leave the Manager body read at the old value, with every test still green — the config equivalent of the "one exit fixed, sibling left" shape that has come up repeatedly in this pass.

Added the comparison: a test importing both **real** modules that asserts they agree on the default and on each way the env var can be invalid (`"0"`, `"-1"`, `""`, `"abc"`). Drifting the copy to 180 fails it.

I did not try to remove the duplication itself — that fight is with the mocking, and a test that fails on divergence is the cheaper guarantee.

`src/__tests__/services` + `comfyui`: **316 files, 6991 tests**. The single failure is `panel-image-relay`'s deadline case under concurrency, addressed in #2823.

Issue #2325, opened 2026-08-26
node-management test asserts a refusal that only happens when no ComfyUI is running, so `npm test` is red on any dev box with ComfyUI up


artokun:
Surfaced while verifying an unrelated docs PR (#2324). Filed rather than folded in, because it is not that PR's bug and it affects everyone's local suite.

## The failure

`src/__tests__/services/node-management.test.ts` →
`does not use a SAVED DEFAULT workspace for cm-cli when COMFYUI_PATH is unset`

Fails on a dev machine that has a ComfyUI listening on `:8188`. It expects the auto-detection to **refuse** and raise `NodeManagementError`; with a live server the detection succeeds, resolves a workspace, and the raise arrives as `ProcessControlError` instead.

## Why this is the test's environment-dependence, not a code regression

The control was run properly — the suspect variable was varied, not just the symptom observed:

- It fails **in isolation**, so it is not cross-test pollution.
- It **still fails with `docs/` reverted to `origin/main`**, so it is unrelated to the change that was under review when it appeared.
- **CI is clean of it.** GitHub runners have no ComfyUI on `:8188`, so the refusal the test wants does happen there.

Confirmed independently on this host: `curl http://127.0.0.1:8188/system_stats` → **200**, and the test is present at that path on `origin/main`.

## Why it is worth fixing rather than tolerating

The test passes for a reason that has nothing to do with the behaviour it names — it needs *the absence of a running ComfyUI*, which is the opposite of a normal working state for anyone developing this project. So:

- `npm test` is red on any dev box with ComfyUI up, which is most of them.
- A red suite that everyone learns to ignore is worse than a missing test: it trains people to skip the tally, and the next real regression lands in the noise.

The fix is presumably to stub the detection rather than depend on the ambient port — pin the environment the test claims to be testing instead of inheriting the machine's.

## A second, separate finding from the same run

**`npm test` chains its steps with `&&`.** A vitest failure therefore short-circuits the eight checks behind it (`asset-counts`, `check:vocabulary`, `vocab:export`, the docs gates, …), so a single flaky or environment-dependent test silently skips every gate that follows. On the run that surfaced this, all eight had to be executed by hand to confirm they passed.

That is worth separating even if this test is fixed: any future red test hides the rest of the suite's gates behind it, and the summary line does not say so.

## Environment

Windows 11, this host has ComfyUI live on `:8188`. Reproduces regardless of the working tree's diff.

artokun:
[...]

PR #2829, opened 2026-09-03
test: stop the Manager-dialect unit tests probing the developer's real ComfyUI

artokun:
## Summary

Two orchestrator test files reach `detectManagerApi()`, which **probes the configured ComfyUI over the network**. On CI nothing is listening, so the probe fails, the dialect is undetermined, the git-URL guard falls through — and they pass. On any machine with ComfyUI actually running, detection *succeeds*, returns the v2 dialect, and the guard refuses the installs those fixtures are built on.

So the suite's verdict depends on whether the developer happens to have the app open. Everyone working on this project has it open.

## Measured

Same commit, `src/__tests__/orchestrator`, on one machine:

| environment | result |
|---|---|
| ComfyUI running on :8188 | **2 files failed** — `install-accepted-never-enqueued`, `install-node-git-expectation` |
| nothing listening | 305 passed |

Confirmed by re-running the same file with `COMFYUI_URL` pointed at a dead port: 15/15.

This briefly read to me as "main is broken". It is not — CI is green *because* it has no server, which is the wrong reason for a unit test to pass.

## Fix

- `install-accepted-never-enqueued` pins the **legacy** dialect in `beforeEach` — the direct `files:[url]` route its fixtures model.
- `install-node-git-expectation` **forces the probe to fail** for its dialect-undetermined case, instead of hoping nothing answers. That file is mine (#2795) and so was the assumption.

Now **306/306 with ComfyUI running**, re-measured after this branch merged main.

_Corrected: an earlier revision of this description counted `late-mutation-e2e` as a third file failing for this reason. It is not — it holds no reference to `detectManagerApi` and runs its own loopback server on an ephemeral port. It is the separate #694 load flake (a 60 ms real-bridge window that fails under full-suite load and passes isolated), and it passed in both runs below._

Found by self-review; codex is out of tokens, so the adversarial pass is mine until Sept 6.

Test-only change.


artokun:
Measured this while verifying an unrelated stack, and it makes the case for this PR concrete: **`origin/main` is 10-red locally right now**, and this PR is exactly the fix.

On `origin/main` (`12af490`), the two files this PR touches:

```
install-node-git-expectation.test.ts      43 tests |  1 failed
install-accepted-never-enqueued.test.ts   15 tests |  9 failed
                                          10 failed | 48 passed (58)
```

On `test/hermetic-manager-dialect` (`0ecebc4`), same two files, same machine, same dependency set:

```
Test Files  2 passed (2)
     Tests  58 passed (58)
```

So the hermetic rewrite converts them, and nothing else in `src/__tests__/orchestrator/` regresses — the other 304 files pass in both arms (4982 passed across 306 files).

Worth flagging why this is more than tidiness: because these two failures live on main, **every contributor's local run is red before they change anything**, and every agent or human verifying an unrelated branch has to first prove the red is pre-existing. I did exactly that today for #2799 and it cost a control run on a second worktree. A red baseline also trains people to skim past failures, which is how a real one gets through.

CI does not see it (a runner has no ComfyUI at the probed address, so the tests take a different path there) — which is precisely why it survived: green in CI, red on every developer machine that has ComfyUI running.

artokun:
Follow-up, because the obvious cheaper alternative to this PR is a trap and I'd rather it be measured than argued.

**This class has recurred.** #2325 was the same shape — a unit test whose verdict flips depending on whether the developer has ComfyUI listening — closed 2026-08-26 against a different file (`node-management.test.ts`). Today's two files are a fresh instance, so per-test hermeticism is being applied for at least the second time.

That invites the structural one-liner: the suite already has a shared setup (`isolated-test-home.ts`) that redirects `HOME` before any module loads, so why not pin `COMFYUI_URL` there too and make the whole class impossible?

**Because it makes things worse. Measured on the full suite:**

| | red files | red tests |
|---|---|---|
| today (unpinned) | 2 real + 4 not-really | 10 real |
| `COMFYUI_URL=http://127.0.0.1:1` | **12** | 60+ |

It does fix these two files — `58/58`, the same result this PR gets. But it breaks `download-jobs` (22), `model-management` (17), `download-adopt-orphans`, `download-status-decline`, `divergent-install-reachable` and more.

The reason is not mock plumbing: download records are **scoped by the ComfyUI URL**, so changing it changes the scope key and every record a test seeded under the default address stops resolving — `No download matching id ...`. A blanket pin trades one environment dependence for a larger one, in tests that are currently correct.

So the per-file hermetic rewrite here is the right shape, and worth saying out loud in the PR so the next reviewer doesn't re-derive the shortcut. If a general guard is ever wanted, it would have to be "fail loudly when a unit test opens a socket", not "point the socket somewhere else".

artokun:
Adversarial pass (mine — codex is out through Sept 6). **Reproduced the diagnosis independently, and it also explains something I have been misreading all day.**

## Measured, at `origin/main`, one machine, one commit

```
ComfyUI reachable on :8188          Test Files 2 failed    Tests 10 failed | 48 passed
COMFYUI_URL=http://127.0.0.1:59999  Test Files 2 passed    Tests 58 passed
```

Exactly the claim: the only variable is whether the app happens to be running, and it flips ten tests. `install-accepted-never-enqueued` and `install-node-git-expectation` both.

One methodological note that cost me two runs and is worth recording. Running those files from a stale checkout gave **57/57 passing in both environments**, which looks like a refutation. It is not — that checkout is five commits behind and has 199 orchestrator files against main's 307, so it does not contain the tests at all. The reproduction needs a worktree at `origin/main`; anything else is measuring a different suite.

## Why this is worth more than a test-hygiene fix

I have reported "main's known local red baseline — 2 files, 10 tests, unrelated to my change" **five separate times today**, on #2840, #2802, #1477 and others. Each time it was true in effect: the failures were not mine, and I controlled for them by re-running the files at `HEAD~1` or checking they import nothing I touched.

But I had it filed as an unexplained property of the repo. It is not unexplained — it is this, and it is fixable. A suite whose verdict depends on whether the developer has the app open does not just produce noise; it **trains the reader to discount red**, and I am the demonstration. The next genuinely broken thing in either of those files would have gone straight past me under the same label.

That is the real cost, and it is worth putting in the PR body: the failure mode is not "flaky tests", it is "a standing exception that makes a whole file's verdict unreadable".

Nothing blocking — the fix is in the right direction (make the tests hermetic rather than teach people to unset an env var).

artokun:
**Re-measured on the machine state this PR is about — ComfyUI genuinely running — and it confirms the fix while correcting my own count.**

Same commit, same live server (`/system_stats` → 200 on :8188), the only variable being the two test files:

```
this branch                                      306 files / 4970 tests   ALL PASS
those 2 test files reverted to origin/main         2 files FAILED (10 tests)
```

That is a clean control: nothing else differs, so the failures are attributable to those files' assumptions and to nothing else. The fix does what it claims.

**What I got wrong.** The description said *three* files reach `detectManagerApi()` and listed `late-mutation-e2e` as the third. It is not one:

```
grep -c 'detectManagerApi|managerApi'  late-mutation-e2e.test.ts   ->  0
```

It stands up its own server on `127.0.0.1:0` and talks to it over a WebSocket — it has nothing to do with the Manager dialect probe. It is the separate **#694 load flake**: a 60 ms real-bridge window that fails under full-suite load and passes isolated. I had that written down from 2026-08-14 and still put it in this PR's evidence table as a probe failure.

It passed in **both** runs above, which is the tell I should have taken at the time: a file that fails for the stated reason cannot pass on the branch that does not touch it.

Description updated — "two", the table now lists the two real files, and `305/305` is now `306/306` after this branch merged main.

**Why this matters beyond a number.** The table was the argument. Someone reading it would have believed a third file was fixed here, and if `late-mutation-e2e` flaked for them afterwards they would have looked for a regression in this PR rather than at #694. An evidence table that includes something the change does not address is worse than one row shorter — it misdirects the next person to hit it.

The fix itself is unchanged and still test-only.

artokun:
**Right, and it caught an incomplete correction of mine.**

I corrected the PR description earlier today — `late-mutation-e2e` is not one of the probe-dependent files; it invokes only `panel_set_node_mode` and never reaches the `panel_install_node` dialect probe, and it is the separate #694 load flake. I did not follow that correction into the changelog entry, which still said "three".

Now says two, and names the third explicitly rather than silently decrementing, so a reader who remembers the old claim can see what happened to it.

The tell I had already had and not applied here: it **passed in both arms** of the control run. A file that fails for the stated reason cannot pass on the branch that does not touch it.

PR #2838, opened 2026-09-04
fix(2835): run an npm-installed pi via its shim target, without a shell

### LINKED ISSUE #2835: pi backend broken on Windows: npm .cmd shim refused, extensionless shim ENOENT
Environment: Windows 11, comfyui-mcp 0.52.184→0.52.185 (npx), pi 0.84.4 installed via npm (global), Node v25.5.0.

Steps to reproduce:
1. Install pi via npm (`npm i -g` → provides `pi` bash shim + `pi.cmd` + `pi.ps1` in %APPDATA%\npm, no native pi.exe)
2. Start panel orchestrator (`comfyui-mcp connect`), connect panel tab with backend=pi

Observed:
- Model probe fails: `Could not run 'pi --list-models': spawn C:\Users\Administrator\AppData\Roaming\npm\pi ENOENT` (spawning the extensionless bash shim fails on Windows)
- Setting COMFYUI_MCP_PI_PATH to pi.cmd is explicitly refused: "[pi-backend] COMFYUI_MCP_PI_PATH points at a shell shim (...pi.cmd); Node cannot spawn a .cmd/.bat shell-lessly — ignoring it"
- pi-backend.js spawns use no `shell` option, so `.cmd` shims can never work; result is "pi CLI not found" and the panel connects in degraded ack with an empty model list

Expected: pi backend should work on Windows where pi is installed via npm. Suggestion: on Windows, resolve npm shims the same way npm itself does — spawn pi.cmd with shell:true (or child_process.exec with proper escaping; note the codebase already uses shell:true elsewhere per DEP0190), or accept COMFYUI_MCP_PI_PATH pointing at a .cmd when process.platform === 'win32'.

Workaround: none found — no native pi.exe exists in the npm package.

artokun:
Fixes #2835.

`npm i -g pi` writes `pi` (an extensionless bash script), `pi.cmd` and `pi.ps1`, and **no `pi.exe`**. Measured on Windows 11 / Node 24 against a real npm global shim on my box:

```
spawn(<extensionless shim>) -> ENOENT
spawn(<.cmd>)               -> EINVAL
```

Both of the reporter's observations reproduce, and the refusal's own claim is correct too: Node genuinely cannot spawn a `.cmd` shell-lessly.

## Two defects, and the second was only half-named

1. Discovery scans `["pi.exe", "pi"]` and `existsSync` matches the **extensionless** shim — so it is found, returned, and fails to spawn. The reported `spawn …\npm\pi ENOENT` is not "pi wasn't found"; it is "we picked the one file Windows cannot execute".
2. `COMFYUI_MCP_PI_PATH` at the `.cmd` is refused, with advice to point at "the real pi/pi.exe" — which the npm package does not ship, so the advice cannot be followed.

## Not `shell: true`

The issue suggests it. It would reintroduce exactly what the existing guard exists to prevent, and that guard's comment is not incidental:

> *"we never spawn through a shell (the prompt is user data — shell-quoting it would be an injection risk)"*

The prompt reaches these spawns as an argv element. Routing it through `cmd.exe` makes every prompt a potential command line, and DEP0190 exists because hand-rolling that quoting is a bad idea. A P1 that ends in shell injection is worse than the one it closed.

## What this does instead

npm's shim names its own entry point on its last line:

```
"%_prog%"  "%dp0%\node_modules\pi\bin\pi.js" %*
```

So it resolves to a script and launches as `node <script>` — **no shell anywhere**, and the injection surface is unchanged from today. Verified on a real shim:

```
extracted script -> node_modules\wrangler\bin\cf-wrangler.js
node <script>    -> rc=2, "Error: unknown subcommand --version"
```

`rc=2` is the CLI's own argument error: it **executed**.

The objection was never the file extension; it was the shell.

**Ordering matters and I got it wrong first.** The win32 shim lookup runs BEFORE any PATH result is trusted. My first draft called `resolvePiBin()` first, which returned the extensionless shim and reproduced the original bug inside the fix — a test caught it. An unresolvable shim, and an extensionless file on Windows, are both still refused: "not found" with actionable guidance beats a spawn that fails EINVAL.

The `~32K` command-line guard now counts the launch prefix too, since `node <script>` is part of the same command line the prompt must fit inside.

## Verification

- `src/__tests__/orchestrator/`: **4964 passed**. The 2 failing files (`install-accepted-never-enqueued`, `install-node-git-expectation`) are pre-existing and are what **#2829** makes hermetic — unrelated to this diff.
- `npm run lint` clean (anti-slop: none added), `check-changelog` passes.
- Mutation-verified, each caught by the test written for it: the win32 shim lookup skipped; an unresolvable shim target accepted; and the extensionless refusal removed. That last one initially survived — every other case wrote a resolvable `.cmd` alongside, so the guard was never exercised — and it is the reported ENOENT, so it now has its own case asserted in both platform directions.

## What is NOT verified

**I do not have `pi` installed**, so this is not tested end-to-end against the real CLI. What is measured is the shim mechanism — npm's generated `.cmd`, the two spawn failures, and that the extracted script runs under `node`. The unit tests drive a byte-faithful copy of npm's shim rather than an invented one. Someone with pi on Windows should confirm the model probe populates before this is trusted to close the issue.


artokun:
CI caught a real defect in this PR, and it is worth naming rather than quietly amending.

The first push was **green on Windows and red on all three POSIX legs**. Two of my assertions demanded `node <script>` unconditionally, while the resolution they test runs only under `win32`:

```
AssertionError: expected '/tmp/cmcp-2835-7tsO8h/pi' to be '/opt/hostedtoolcache/node/22.23.2/x64/bin/node'
AssertionError: expected undefined to be '/opt/hostedtoolcache/node/22.23.2/x64/bin/node'
```

The win32 scoping is deliberate and not a gap: npm's extensionless shim is a shell script with a shebang, which POSIX executes directly. Resolving it to `node <script>` there would replace a working launch with a second guess. So on POSIX the same fixtures correctly yield the extensionless shim itself, and a `.cmd` override stays refused because nothing there can run a Windows batch file.

Both assertions now cover both directions, matching exactly what the failing legs reported. The third test in that file was already platform-guarded, which is why it passed throughout — I guarded the case I was thinking about and not the two I had written first.

Worth recording because it is the mirror image of the correction #2792 had to make. There, four assertions were written on POSIX assumptions and would have gone red on the `windows-latest` leg alone. Here, mine were written on Windows and went red on the other three. **The matrix is the only thing that catches either**, and a fix that is deliberately platform-scoped needs its tests scoped the same way — asserting what the *other* platform does, not skipping it.

artokun:
Re-verified this on real Windows, because its four green checks assert nothing about the platform it fixes.

**Why the checks do not cover it.** `ci.yml` declares a three-OS matrix but hardcodes `runs-on: ubuntu-latest` (comfyui-mcp#2851) — the job *named* `Build (windows-latest)` reports `labels: ubuntu-latest`. So a Windows-only fix gets three green Build checks from one Linux runner. That makes local evidence the only evidence here.

**Run on this box** (`win32`, Node v24.16.0), resolving three real npm `.cmd` shims and spawning each resolved target shell-lessly:

```
codex     -> ...\Roaming\npm\node_modules\@openai\codex\bin\codex.js
            spawn err: none  exit: 0  out: codex-cli 0.153.0
gemini    -> ...npm\node_modules\@google\gemini-cli\bundle\gemini.js
            spawn err: none  exit: 0  out: 0.49.0
firebase  -> ...\npm\node_modules\firebase-tools\lib\bin\firebase.js
            spawn err: none  exit: 0  out: 15.23.0
```

Three independent CLIs, three different layouts (`bin/`, `bundle/`, `lib/bin/`), all resolved and all launched.

**And the controls, which are the issue's own two failure modes:**

```
.cmd as the executable, no shell -> EINVAL
extensionless shim, no shell     -> ENOENT
```

Exactly what the title reports. So the bug reproduces here and the fix clears it, on the platform in question.

**One caveat on my own method, since it nearly cost me the result.** My first probe reported `resolver returned null` for all three shims and I briefly took that as a defect in the resolver. It was my harness: the shim path was built with an escaped `\n` that collapsed into a real newline, so the resolver was handed a garbage path. The second control was wrong too — it spawned Node *with* the `.cmd` as an argument rather than as the executable, which naturally produced no error. Both fixed above with `path.join` and a proper executable spawn; the numbers here are from the corrected run.

Unit tests also pass on `win32` (34/34 in `pi-backend.test.ts`), but they mock the spawn, so they are not the evidence that matters for this one.

artokun:
Adversarial pass (mine — codex is out through Sept 6). **I ran the resolver against real npm shims on this Windows box rather than reading it, and it holds — including the end-to-end launch shape.** CI cannot check any of this: the matrix declares three OSes but `runs-on: ubuntu-latest` is hardcoded (#2851), so every "Build (windows-latest)" job on this PR ran on Linux.

## The parser, against 8 real shims

Called `resolveNpmShimTarget` on the actual `%APPDATA%\npm\*.cmd` files here:

```
cf-wrangler.cmd  -> ...\node_modules\wrangler\bin\cf-wrangler.js
codex.cmd        -> ...\node_modules\@openai\codex\bin\codex.js
firebase.cmd     -> ...\node_modules\firebase-tools\lib\bin\firebase.js
gemini.cmd       -> ...\node_modules\@google\gemini-cli\bundle\gemini.js
pn.cmd           -> ...\node_modules\pnpm\bin\pnpm.mjs
pnpm.cmd         -> ...\node_modules\pnpm\bin\pnpm.mjs
pnpx.cmd         -> ...\node_modules\pnpm\bin\pnpx.mjs
pnx.cmd          -> ...\node_modules\pnpm\bin\pnpx.mjs
resolved 8/8
```

That covers `.js` and `.mjs` targets, scoped packages, and several shims aliasing one package — more variety than a fixture would have given.

## The launch shape, measured

```
node <script>  -> ok, stdout=11.9.0
spawn(.cmd)    -> ERROR EINVAL
spawn(bare)    -> ERROR ENOENT
```

Both of the reporter's symptoms reproduce exactly, and the fix's route works where both fail. That is the whole claim of the PR, confirmed on the platform it is about.

## What I probed and could not break

- **Failure is fail-safe.** Every path in `resolveNpmShimTarget` returns `null`; nothing falls through to a guessed path.
- **It refuses to hand back something that fails the same way** — a resolved target that is itself `.cmd`/`.bat`, or extensionless on Windows, is rejected rather than returned. That is the specific trap the original discovery fell into.
- **`split(";")` is Windows-only.** I checked it is inside `if (isWin)` — on POSIX that separator would have produced one bogus PATH entry.
- **Refusing `shell: true` is right**, and the reason in the comment is the real one: the prompt arrives as an argv element, and routing it through `cmd.exe` makes every prompt a command line.

## One note, not a blocker

`{ command: process.execPath }` pins pi to the **orchestrator's** Node, whereas npm's own shim resolves `node` from PATH at run time. Under nvm-style multi-version setups those can differ, so a pi installed against another Node would run under this one. Almost certainly fine for pure JS, and using `execPath` is more reliable than assuming `node` is on PATH — worth one clause in the comment so the difference is a choice on record rather than an accident.

artokun:
**Self-review found a branch of this PR that could not resolve anything. Pushed a fix.**

I exercised `resolveNpmShimTarget` against the **10 real npm shims** in `%APPDATA%\npm` on this Windows box — extracted out of the file rather than retyped, so the pattern under test is the shipped one:

```
                     before      after
.cmd shims           10/10       10/10
sh (extensionless)    0/10       10/10
```

The `.cmd` branch was always fine, including `.mjs` targets (pnpm) and scoped packages (`@openai/codex`, `@google/gemini-cli`). The `$basedir` branch resolved **nothing**, because npm's sh shim names the interpreter and the script on one line:

```sh
exec "$basedir/node"  "$basedir/node_modules/pi/bin/pi.js" "$@"
```

A permissive `.+?` anchors on the FIRST `$basedir/` and swallows `node"  "$basedir/...` into the capture. The result has embedded quotes, never exists, and `existsSync` correctly turned it into `null` — so the guard converted a wrong answer into no answer, and the branch looked like it was declining rather than failing. Its comment said it covered "the .ps1/sh variants".

Fix is one character class: `[^"'\s]+?`. Measured zero regression on the `.cmd` path.

**Why the tests did not catch it:** the existing case seeds the POSIX shim as the stub `"#!/bin/sh"`, which contains no `$basedir` at all, so the branch was never exercised with real content. The new test builds npm's actual shim text. Mutation-verified — restoring the permissive pattern fails exactly that test and leaves the other 10 passing.

**Severity: no behaviour change on Windows today.** The candidate loop tries `pi.cmd` before `pi`, and the `.cmd` branch already resolved, so nothing the reporter hits is affected. This makes a documented fallback true instead of dead.

**One harness note, since it nearly produced a false finding in the other direction.** My first comparison said the tightening broke `.cmd` 0/10. It had not — a quoted heredoc collapsed `[\/]` to `[\/]` in my *retyped* copy of the pattern, so my harness only matched forward slashes. The file's own bytes were always correct. Re-run with the class built via `String.fromCharCode(92)`, the real numbers are the table above. Third time today a shell layer has corrupted a pattern in a test rig; retyping a regex is the same mistake as retyping an argv.

artokun:
**Worked the Copilot review on this PR. Three of its four findings were real; all three are fixed.**

**1. Resolution order — an explicit `COMFYUI_MCP_PI_PATH` could lose to a PATH shim.** The pre-pass collected every candidate (override, then every PATH directory) and ran `resolveNpmShimTarget` across the whole list. An override pointing at a real executable is not a shim, so it returned null and the loop walked on to PATH — where any resolvable shim won. An override exists precisely to win. Each source is now decided in order, classifying `.exe` / `.cmd` / extensionless within a directory before advancing.

**2. PATH hygiene.** A quoted entry (`"C:\Program Files\nodejs"`) concatenates into an invalid path, and a dead UNC entry can block `readFileSync` on an SMB timeout during backend startup. `src/services/self-update.ts` already strips wrapping quotes and skips UNC before its own synchronous probes; same treatment here.

**3. Readiness disagreed with launch — and this one is #2835's own symptom.** `backend-readiness.ts` called `resolvePiBin`, whose name list is:

```js
pi: ["pi", "pi.exe"], // no .cmd — Node can't shell-lessly spawn it (see resolvePiBin)
```

That comment states the premise this PR overturns. So after the fix, a `.cmd`-only install reads **ABSENT** in the UI although `prepare()` can run it, and an unresolvable extensionless shim reads **PRESENT** although it spawns ENOENT — which is exactly what the reporter saw. Readiness now calls `resolvePiLaunch`.

**What I got wrong on the way, since it is the useful part.** My first cut made the Windows block `return null` instead of falling through. That dropped `resolvePiBin`'s well-known-location probes (`%LOCALAPPDATA%\pi\bin\pi.exe`, `~/.local/bin/pi.exe`), which are not on PATH — **26 existing tests failed**, which is how I found it. The block now falls through deliberately, with a comment saying why.

**Verification:** 84 tests green across `pi-backend`, the shim suite and `backend-readiness`; `tsc` clean; the 10 real npm shims in `%APPDATA%\npm` still resolve **10/10** in both `.cmd` and extensionless form. Both new behaviours mutation-verified — disabling the override branch fails the precedence test, removing quote-stripping fails the quoted-PATH test.

**The fourth finding I did not take.** It says the platform-branching tests fail on Ubuntu CI. They do not — CI is green, and every job on this repo currently runs on `ubuntu-latest` regardless of its label (that is #2851). The new tests are `describe.skip` off win32 rather than asserting the wrong thing, which is the shape that finding was really asking for.

artokun:
**Ran the resolver against real npm shims on a Windows box, rather than only reading it.**

Checked out this branch and imported `resolveNpmShimTarget` directly (via `tsx`, so it is the branch's own source, not a transcription), then pointed it at every `.cmd` in `npm config get prefix`:

```
10 shims: 10 resolved to an EXISTING file, 0 null, 0 pointing at nothing
```

The set is varied enough to be worth something:

| shim | resolved to |
|---|---|
| `codex.cmd` | `node_modules\@openai\codex\bin\codex.js` — scoped package |
| `gemini.cmd` | `node_modules\@google\gemini-cli\bundle\gemini.js` — scoped, non-`bin` dir |
| `pnpm.cmd` | `node_modules\pnpm\bin\pnpm.mjs` — `.mjs`, not `.js` |
| `cf-wrangler.cmd` | `node_modules\wrangler\bin\cf-wrangler.js` — binary name ≠ package name |
| `firebase.cmd` | `node_modules\firebase-tools\lib\bin\firebase.js` — nested `lib/bin` |

Every resolved path was then checked with `existsSync`, so this is not just "it returned a string" — each target is a file that is actually there.

That covers the shapes most likely to break a shim parser: a scope in the path, a bundle directory instead of `bin`, an ESM extension, a renamed binary, and a nested `lib/bin`. Zero nulls means the parser did not silently fall through on any of them, which is the failure mode that would put us back on the reported symptom with no signal.

Not covered here, and worth stating: this box has no `pi` install, so the end-to-end launch path is still unexercised — this verifies the resolver, not the relaunch. And these are all npm-installed shims from one npm version; a shim written by an older npm or by Yarn/pnpm's own linker could differ in layout.

PR #2840, opened 2026-09-04
probe(291): stop deferring the panel's in-process tools behind tool search

artokun:
**Draft on purpose — I cannot verify this against the reported setup.** For panel#291's Claude-backend recurrence (mcp 0.52.186 / panel 0.15.165).

## Why this is worth having ready

The two MCP servers in a Claude session differ in exactly one relevant way, and the split matches the report:

| server | kind | reporter |
|---|---|---|
| `comfyui` | `type: "stdio"` — spawned subprocess | **works** |
| `panel` | `type: "sdk"` — in-process `createSdkMcpServer` | **absent** |

That is why nothing the reporter tried moved it. Restart, hard reload, fresh session after deleting port-scoped state, removing stale orchestrators, `COMFYUI_MCP_TOOL_MODE` — all act on configuration and lifecycle. None touches the in-process path, which is the only thing that differs between the server that works and the one that does not.

From the pinned SDK (`@anthropic-ai/claude-agent-sdk@0.3.240`):

> `CreateSdkMcpServerOptions.alwaysLoad` — *"When true, all tools from this server are always included in the prompt and never deferred behind tool search. … **Default: tools are deferred when tool search is enabled.**"*

> `McpStdioServerConfig.alwaysLoad` — *"…**since the tools must be present when the turn-1 prompt is built.**"*

We never pass it. `grep -rn alwaysLoad src/` finds nothing on `main`.

## This is a disambiguator, not a claimed fix

Deferral explains *"not in the directly declared tools"*. It does **not** explain *"not in the deferred catalog either"* — a merely-deferred tool is exactly what `ToolSearch` is for. So one of two things is true, and I cannot tell which from here.

That is precisely what makes this worth landing:

- **Tools appear** → the cause was deferral, and this is the fix.
- **Tools still absent** → deferral is ruled out, the in-process server genuinely is not registering, and the search narrows to the bridge→server binding with no further guessing.

Either outcome ends a line of investigation that has otherwise consumed two months and four "fixed" cycles on this issue.

## Cost

Prompt size, for a bounded set of ~20 tools that a panel session is expected to use on turn 1 anyway. No behaviour change where tool search is off.

## Verification

`tsc` clean, `npm run lint` clean (anti-slop: none added), `check-changelog` passes, and two tests pin that the opt-out is on the `comfyui-panel` server specifically rather than drifting onto some other SDK server later.

**What is not verified: anything about the actual bug.** No Comfy Desktop here, so this is untested against the reported environment. Merge it as an experiment the next report can read, or have the reporter run this branch — I would not treat a green CI here as evidence it helps.


artokun:
Self-review before this goes anywhere. **The mechanism verifies; the cost line does not.**

## Verified

`alwaysLoad` is real and sits on the right type. From the SDK on disk, `CreateSdkMcpServerOptions` (declared line 486) carries:

> "When true, all tools from this server are always included in the prompt and never deferred behind tool search. Applied via `_meta['anthropic/alwaysLoad']` on each tool."

and the "Default: tools are deferred when tool search is enabled" sentence is present too. So the quoted basis is accurate, and the stdio-vs-sdk split the body rests on is a real distinction.

One caveat on provenance: `node_modules` here holds **0.3.233** while `package.json` pins `^0.3.240`, so this is verified against the installed SDK, not the pinned one. `alwaysLoad` exists in both, so the conclusion holds — but I am not able to confirm the 0.3.240 wording locally, and the body attributes the quotes to 0.3.240.

The test is also sound, which I doubted at first: it slices `createSdkMcpServer({` to the next `})` and requires *both* `alwaysLoad: true` and `name: "comfyui-panel"` inside that slice. I expected the slice to truncate on a nested `})` from an inline `tool({…})` array — it does not, because `tools` is passed as a bare identifier. Confirmed the slice reaches both assertions.

## The problem

The body's only cost argument is:

> "Prompt size, for a bounded set of **~20 tools**."

Counted from `buildPanelToolDefs()`: **96 distinct `panel_*` tools**, not ~20. Their descriptions alone total **77,499 characters — roughly 19,000 tokens** (mean 815 chars, longest 3,068), before any JSON-schema fields are counted.

So this puts on the order of 20k tokens into the **turn-1 prompt of every Claude-backend panel session**, permanently, to run a diagnostic. And the delta lands exactly where it hurts: with tool search *off* these tools load anyway and nothing changes, so the entire cost falls on the tool-search-enabled configuration this is meant to debug.

I wrote "~20" without counting. It is off by ~5x, and it was the whole cost case.

## What I would do instead

The same SDK text names a narrower instrument, two lines below the passage I quoted:

> "Per-tool `tool({ alwaysLoad })` still works and is **OR'd with** this."

So the probe does not need the server-wide flag. Either:

1. **per-tool `alwaysLoad` on a small diagnostic subset** — two or three `panel_*` tools are enough to distinguish "deferred and unfound" from "never registered", which is the only question this PR asks; or
2. **gate the server-wide flag behind an env var** so the reporter can flip it for one session and nobody else pays for it.

Option 1 keeps the disambiguation intact at ~2% of the prompt cost and is my recommendation. Holding this as draft until that is decided — the reasoning stands, the blast radius does not.

artokun:
Rewritten per the review above. Pushed as `f604eae`.

## What changed

The server-wide flag is gone. `alwaysLoad` is now applied **per tool** to exactly two — `panel_graph_outline` (the canonical read entry point for the live-graph surface the report is about) and `panel_canvas` (the cheapest panel tool whose presence is still meaningful). ~1.4k characters of description against ~77k for the full 96, and the probe answers the same question: one tool provably in the turn-1 prompt distinguishes "deferred and unfound" from "never registered".

## The test is now behavioural, not textual

The old test matched source text. The new one builds the server exactly as production does, connects it to a real MCP client over an in-memory transport, and reads the `_meta` the SDK actually emits. Confirmed against the installed SDK that `tool(…, { alwaysLoad: true })` sets `_meta['anthropic/alwaysLoad'] = true` while a plain tool has no `_meta` at all.

Four assertions, each mutation-verified:

| mutation | kills |
|---|---|
| restore server-wide `alwaysLoad: true` | the count assertion + "ordinary tool stays deferred" |
| drop the per-tool option | "probe tools carry the meta" + the count |
| swap a name in `PROBE_ALWAYS_LOAD` | "probe tools carry the meta" + the count |

The count assertion is the one that matters: a presence check on the two probe tools cannot see the server-wide flag coming back, because that sets the same `_meta` on all 96. A count can.

## Two corrections to my own review

**The server-wide flag does work** — I implied it might not. From the SDK bundle: `registerTool(..., { _meta: e.alwaysLoad ? {"anthropic/alwaysLoad": true, ...n._meta} : n._meta })`. So the original PR was functional. It was expensive, not broken.

**My first mutation silently did not apply, and I nearly reported the test as weak on the strength of it.** The mutation script matched `\n` against a CRLF file, so `str.replace` did nothing while the script printed "mutation applied" — all four tests passed and I read that as the guard failing to catch a real regression. It was only wrong because the file never changed. Re-run with the line ending handled, the guard kills two tests as intended. Every mutation above was re-checked with `git diff --stat` proving the file actually changed first.

## Incidental

`afterAll` called an unassigned `close` when `beforeAll` threw, reporting "close is not a function" and hiding the actual construction error two frames up. Guarded in a separate commit.

Still draft: this is a disambiguator, not a claimed fix, and it needs the reporter's setup to produce its answer.

artokun:
Added a cross-reference in `cecd139` — this PR introduces the **first `alwaysLoad` in the repo**, and #2774's empty-toolset check has a comment that rests on there being none:

> Today nothing in this repo sets `alwaysLoad` or enables tool search, so every server is treated alike and the list is populated.

Nothing changes behaviourally: `alwaysLoad` only opts *out* of deferral, and deferral requires tool search, which nothing enables. But #2774 documents a real deferral hazard — a deferred tool is absent from the init `tools[]`, and its per-process latch can make an all-deferred session read as *every* server contributing zero — and a stale precondition is how that ships.

The note is here rather than on #2774 deliberately: **the change that invalidates an assumption should carry the reference**, so it cannot be merged without the reader seeing it. If #2774 lands first, its comment wants a clause; if this lands first, the reference is already in place.

Worth recording the asymmetry it would create if tool search were ever enabled with both in: the panel server would report these two always-loaded tools and pass the check, while a fully-deferred stdio server would report zero and be flagged — so the one server #2742 is actually about becomes the one whose emptiness that check can no longer see. That is an argument for keeping this probe scoped to two tools rather than restoring the server-wide flag, independent of the prompt-size argument.

Still draft. `tsc`, `npm run lint` (anti-slop: none added) and the suite all clean.

artokun:
**One finding right and fixed; the other refuted by the SDK's own type declaration.**

**Right — the changelog called a probe a fix.** The entry sat under `#### Fixed` reading "two panel live-canvas tools are no longer deferred behind tool search (panel#291)". A release-notes reader takes that as panel#291 resolved, and it is not: this PR's header says *"Draft on purpose — I cannot verify this against the reported setup"*, and the code says *"this is not asserted as the cause."* Moved to `Changed`, and the entry now says PROBE in its first three words.

**Refuted — `alwaysLoad` is not server-only.** The finding says the flag "applies to the whole server, so this eagerly injects the default 96-tool panel surface". The SDK declares a per-tool form:

```ts
export declare function tool<Schema>(_name, _description, _inputSchema, _handler,
  _extras?: { annotations?: ToolAnnotations; searchHint?: string; alwaysLoad?: boolean; })
```

and the server-wide option's own doc states the interaction:

> Per-tool `tool({ alwaysLoad })` still works and is OR'd with this.

This code passes it as that fifth argument, gated on a two-name set:

```ts
PROBE_ALWAYS_LOAD.has(d.name) ? { alwaysLoad: true } : undefined
```

So the surface is two tools, not 96 — which is the whole reason the constant exists rather than the server-wide flag, and what the comment above it says. The budget concern the finding raises would be correct for the server-wide form, and is exactly why this PR does not use it.

That is the sixth finding in this sweep I have judged wrong or already-addressed, against roughly two dozen that were real. Worth stating plainly: the reviewer is a genuinely useful adversary and not an authority, and the difference only shows up by checking each claim against the code.

artokun:
**Correcting this draft's central evidence — the SDK quote above is spliced, and the table it supports does not hold.**

I re-read the pinned SDK's own types instead of my earlier notes. Resolved `^0.3.240` to the version npm actually installs (**0.3.261**) and read `sdk.d.ts` directly.

## 1. The quote attributed to `CreateSdkMcpServerOptions` is not in it

The body above renders one quotation across an ellipsis:

> `CreateSdkMcpServerOptions.alwaysLoad` — *"When true, all tools from this server are always included … **Default: tools are deferred when tool search is enabled.**"*

Only the first half is from that type. Counted in 0.3.261, the sentence **"Default: tools are deferred when tool search is enabled"** occurs in `CreateSdkMcpServerOptions`:

```
occurrences in CreateSdkMcpServerOptions: 0
found at line 1084  <- export declare type McpHttpServerConfig
found at line 1203  <- export declare type McpSSEServerConfig
found at line 1219  <- export declare type McpStdioServerConfig
```

The ellipsis fuses a sentence from the **stdio/HTTP/SSE** types onto the **sdk** one — and it is the bolded half, the half carrying the argument.

## 2. That inverts the table

`alwaysLoad`, with that documented deferral default, is on four types:

| line | type |
|---|---|
| 531 | `CreateSdkMcpServerOptions` — *no* documented deferral default |
| 1086 | `McpHttpServerConfig` — documents it |
| 1205 | `McpSSEServerConfig` — documents it |
| 1221 | `McpStdioServerConfig` — documents it |

So "the two servers differ in exactly one relevant way" is wrong. Deferral is documented for **stdio**, the server the reporter says *works*, and is the one thing *not* documented for the sdk server. The asymmetry I built this probe on points the other way.

Corroborating rather than only reading types: in the session I am writing from, the `comfyui` **stdio** server's tools are deferred behind `ToolSearch` right now. Stdio deferral is observed, not hypothetical.

## 3. What this does to the disambiguation

The reporter states `panel_*` is absent from **both** the direct list **and** the deferred `ToolSearch` catalog, while stdio `comfyui` is reachable. If deferral applies to both kinds, it is symmetric and cannot produce that split — and a merely-deferred tool is precisely what `ToolSearch` would surface.

**So the branch this probe existed to decide is already decided, by reading.** Deferral is not the cause; the in-process server is not registering its tools. The next place to look is the bridge→server binding — which is what I said this probe would tell us only after a release and a reporter round-trip. That round-trip is no longer needed.

## What I am *not* claiming

Absence of the sentence from the sdk docstring is not proof the sdk path never defers — it may just be documented less thoroughly. The claim is narrower and still decisive: **the stdio/sdk kind split does not explain the reported asymmetry**, because deferral is documented on the side that works.

## Where that leaves the PR

Still worth keeping, with the framing corrected. Per-tool `alwaysLoad` on `panel_graph_outline` now tests *registration*, not deferral:

- tool appears → it was registered and deferred after all, and the reading above is wrong
- tool still absent → it is not registered, confirming the binding hypothesis at ~2% prompt cost

Leaving it draft. The 96-tool / ~19k-token costing in the comment above stands unchanged; only the rationale for running it moves.

artokun:
**This draft now blocks a green PR, and its stated purpose is gone. I think it should be closed.**

Two things changed since I opened it:

**1. Its premise was wrong.** The comment above records it: the SDK quote it rested on was spliced across two types, and the deferral default is documented for **stdio** — the server that works — not for the in-process one. Deferral is symmetric, so it cannot produce the reported asymmetry. The branch this probe existed to decide was already decided by reading.

**2. There is now a better explanation, testable without shipping anything.** `createPanelMcpServer` filters its tools through `resolveToolSurfacePolicy()`, and both presets deny `panel_*` by glob. `COMFYUI_MCP_TOOL_PRESET=safe` — which `docs/configuration.mdx` recommends verbatim — registers **zero** panel tools: connected, empty, absent from the deferred catalog too, and immune to every remedy the reporter tried, because none of them clears an environment variable. It is confirmable by one grep for `surface restricted by policy`. Details on comfyui-mcp-panel#291.

**And it conflicts.** Simulated sequentially against `main` rather than pairwise — pairwise `merge-tree` between two feature branches that each merged main at different points reports conflicts that do not exist in a real merge, which is how I first mis-measured this:

```
merge #2874 -> clean
merge #2840 -> CONFLICT: src/orchestrator/panel-tools.ts
```

comfyui-mcp#2874 fixes the user-visible half of comfyui-mcp-panel#291 — the agent being told it can edit a canvas whose tools were withheld — is green, and is Copilot-approved. This draft is the only code conflict left in the queue; closing it clears that.

Nothing here is lost: the branch stays, and if the policy grep comes back empty and the tool list turns out to be genuinely empty for some other reason, per-tool `alwaysLoad` is still one line away.

Leaving the decision to you rather than closing my own draft unilaterally.

PR #2844, opened 2026-09-04
fix(2693): a relaunch keeps the Python interpreter flags

artokun:
The **argv half** of #2693, which #2792 deliberately left open. Stacks alongside it, not on it — different files.

## What was lost

`-s`, `-E`, `-I`, `-B`, `-O` and `-u` are consumed by CPython **before `sys.argv` exists**, so they are structurally absent from it. `relaunchArgv` rebuilds from `sys.argv`. The reporter's portable install launches with `-s` (ignore user site-packages), so a restart quietly relaunched a server that imports from a **different package set** than the one that was running.

## Why "just prefer the OS argv" is the wrong fix

`relaunchArgv`'s own docblock rules it out:

> *"The order is not a preference between sources of truth: the server's answer simply comes with an **identity binding** the OS reading cannot have, so it is used whenever it exists."*

That binding is what stops a recycled pid's command line being spawned as if it were ComfyUI's. Swapping the order trades a correctness guarantee for flag fidelity.

## What this does instead

The two sources are **complementary**. An interpreter flag cannot appear in `sys.argv`, so finding one in the OS view is not a conflict — it is information `sys.argv` was never able to carry.

Recovery re-establishes the binding rather than assuming it:

- refuses unless the OS reading is **exact** (a flattened command line is not a command — the quoting is unknowable);
- refuses unless the OS argv's script path **corroborates `sys.argv[0]`** — the same file from two independent sources;
- takes only tokens **before the script** that begin with `-`; a bare token there is an argument to a flag this code does not model, and copying it blind invents a command the server never had;
- takes nothing otherwise, leaving today's behaviour exactly as it is.

Script matching normalises separator and case, or corroboration would fail on precisely the platform this report comes from: the OS reports `C:\ComfyUI\main.py` while `/system_stats` reports the same file with forward slashes. There is a test for that specific mismatch.

## Verification

- `src/__tests__/services/`: **6504 passed**, 1 pre-existing failure (`panel-image-relay`, which #2823 fixes). `npm run lint` clean, anti-slop none added.
- Mutation-verified: exactness guard removed, corroboration guard removed, bare values taken, and the recovery severed from `relaunchArgv` — each caught.

That last one **initially survived**. My first eight tests all covered the extraction helper and none covered the re-attachment, so severing the two left everything green — the extraction was tested and the wiring was not, which is the half a relaunch actually uses. `relaunchArgv` now goes through the file's existing `__processControlTestHooks` seam and has its own cases, including that an ordinary relaunch with nothing to recover is unchanged.

## Not claimed

Untested against the reporter's portable install — I have no ComfyUI-Easy-Install here. What is verified is the flag algebra and its refusals.


artokun:
## Self-review found a defect in this PR, and the test I wrote asserted it as correct

`python -X utf8 -s main.py`.

The first implementation filtered the pre-script tokens to those starting with `-`, giving `["-X", "-s"]`. That rebuilds as:

```
python -X -s main.py
```

where **`-X` swallows `-s` as its value**. So the flag this recovery exists to preserve is silently lost — the exact bug #2693 reports — and `-X -s` is not a valid implementation option, so CPython may reject the launch outright. Dropping a value while keeping its flag is worse than taking both or taking neither.

**This is not a corner case here.** `-X utf8` is Python's UTF-8 mode, and #2693 is an encoding bug on Windows. It is among the likeliest command lines this code will actually meet — a portable install that already hit a cp949 crash is exactly the one whose launcher may pass `-X utf8`.

My own comment already stated the right rule — *"a bare value there is an argument to a flag this code does not model, and copying it blind invents a command the server never had"* — and the implementation filtered instead of refusing, which invented a **different** command than either source had. The comment was right and the code did something else.

## Fix

A bare token before the script now refuses the **whole** recovery. Modelling which CPython options take a separate value is not something to hand-roll, and an unmodellable command line is not one we may rebuild. A run of plain flags still recovers, with its own test so the refusal cannot quietly widen into taking nothing at all.

Mutation-verified: reverting to the filter fails the new case.

`src/__tests__/services/`: **6504 passed**, 1 pre-existing failure (`panel-image-relay`, which #2823 fixes). `npm run lint` clean, anti-slop none added.

## Why this survived the first pass

Four mutations at the time, all aimed at the guards I had thought about — exactness, corroboration, flag-only, re-attachment. None of them asked whether the *output* was a valid command line. Mutation testing proves a test can detect a change; it says nothing about whether the behaviour being pinned is correct, and here I had pinned the wrong behaviour with a passing test.

artokun:
Correcting a number in my comment above. I wrote *"6504 passed, 1 pre-existing failure"* — that was the figure from the earlier run, quoted before this run finished. The actual run after the fix showed **3 failed files**, not 1.

Checked each in isolation rather than assuming:

```
sharp-loader.test.ts               26 passed
divergent-install-reachable.test.ts 8 passed
panel-image-relay.test.ts           1 failed | 44 passed
```

So two of the three are full-run concurrency artifacts — they pass alone — and only `panel-image-relay` is a genuine pre-existing failure, which is what **#2823** fixes. Nothing here is caused by this diff, which touches one function.

The correction matters more than the conclusion: I published a test count I had not yet observed, on a PR whose entire subject is a claim that was made without checking the output it produced. Same mistake twice in one PR is worth naming rather than quietly editing.

For the record, that is the fifth full-run-only failure I have seen in this suite today (`console-secrets`, `panel-secrets-slots`, and these two, plus the two hermetic-manager files that #2829 fixes). A file that fails only in the full run is a load artifact until shown otherwise — but "until shown otherwise" means actually running it alone, which takes seconds.

artokun:
Verified this on real Windows rather than through the unit tests' synthetic argv — worth doing here because `ci.yml` hardcodes `runs-on: ubuntu-latest` for every matrix leg (#2851), so the Windows-named checks never ran on Windows.

**Captured a genuine process command line** by starting a Python with interpreter flags and reading `Win32_Process.CommandLine`:

```
"C:\Users\Artokun\miniconda3\python.exe" -X utf8 -s C:\...\flagprobe.py
```

**Fed that exact argv to `interpreterFlagsFromOsArgv`:**

```
real capture: -X utf8 -s   -> []
flags only:   -s -B        -> ["-s","-B"]
single:       -X utf8      -> []
none                       -> []
```

Pure flags round-trip correctly. **`-X utf8` does not** — it yields nothing, so a relaunch drops it.

That is the fail-closed rule working as written: `pre.some((tok) => !tok.startsWith("-"))` bails on any bare token, and `utf8` is bare. The rule is right — I added it after finding that filtering to `-`-prefixed tokens alone produced `python -X -s main.py`, where `-X` swallows the next flag as its value. Guessing is worse than dropping.

**But the case it drops is this issue's own case.** #2693 is a CP949 encoding crash; `-X utf8` is the flag a user most plausibly started the server with to work around exactly that, and it is the one combination the fix cannot carry across a relaunch. Restart, and the workaround is gone.

**A narrow fix keeps the safety.** Python's `-X` *always* takes a value, so `["-X", "utf8"]` can be consumed as a pair without guessing — no ambiguity to resolve. The same is true of `-W`. Something like:

```
const TAKES_VALUE = new Set(["-X", "-W"]);
```

consume flag+value for those, keep the bail-out for every other bare token. `-X utf8` then survives, and an unknown `--foo bar` still refuses rather than mis-splitting.

Not pushing that — it changes the fix's contract and the conservative version is defensible if you would rather ship it and follow up. But the limitation is worth stating in the PR body either way, because "keeps the Python interpreter flags" reads as covering `-X utf8`, and on this evidence it does not.

artokun:
Correction to my own comment above: I said "not pushing that" and then pushed it. Reversing that within the hour deserves a reason rather than a silent commit.

What changed my mind was re-reading the test that pinned the old behaviour. It argued that modelling which CPython options take a value "is not something to hand-roll" — which is true of the option table at large, and simply not true of `-X` and `-W`. Those two **always** consume the next token; there is no ambiguity to hand-roll and therefore no safety being bought by refusing them. Weighed against a measured cost — the encoding flag this issue is about being the one a relaunch drops — the conservative branch was protecting nothing.

Pushed as `d3ed901`:

```
REAL capture: -X utf8 -s   -> ["-X","utf8","-s"]
pure flags:   -s -B        -> ["-s","-B"]
-W once                    -> ["-W","once"]
joined -Xutf8              -> ["-Xutf8"]
REFUSE unknown bare        -> []
REFUSE -X with no value    -> []
REFUSE -X then a flag      -> []
```

9/9 across those shapes; 74 tests green across `interpreter-flags` and `process-control`; lint clean.

**The pinning test is narrowed, not deleted.** Its intent — never rebuild a command the server never had — is still the property worth holding, so it now asserts refusal against `--jit yes`, an option nothing here models. Deleting it would have removed the guard along with the case it happened to name.

If you would still rather ship the conservative version and treat `-X` as a follow-up, revert that one commit — the rest of the PR is independent of it.

artokun:
Merge-ordering note — three open PRs now touch `src/services/process-control.ts` (#2792, #2844, #2802), so I checked all three pairs rather than leave it to whoever merges second.

```
#2792 vs #2844 : src conflicts=0   changelog=1
#2792 vs #2802 : src conflicts=0   changelog=1
#2802 vs #2844 : src conflicts=1   changelog=1
```

**The one source conflict is trivial and carries no design collision.** Both PRs insert at the same point — the top of `__processControlTestHooks`:

- **#2802** adds `detectDesktopLaunch` and `supervisorResult`
- **#2844** adds `relaunchArgv`

Different hooks, different subsystems, no shared state. The resolution is the union of the three entries; there is nothing to reconcile in behaviour. Whichever lands first, the other's rebase is a one-hunk keep-both.

Worth saying explicitly because "conflicts in `process-control.ts`" reads alarming on a 7,000-line file that three P1 fixes are editing at once, and the natural reaction is to assume the changes overlap semantically. They do not: the hunk ranges are otherwise disjoint (#2802 at 167/783/905/918/2056/3314/3385/4078/4197/4651, #2844 at 2848), and the only shared line is the object literal both append to.

**#2792 and #2844 are independently mergeable** in any order — both edit `process-control.ts`, and it auto-merges cleanly. Only the usual `## Unreleased` collision, which the changelog sweep handles.

No action needed on this PR; recording it so the merge does not stall on a conflict that looks worse than it is.

artokun:
**Drove `relaunchArgv` with this machine's actual ComfyUI command-line shape (PID 16804) rather than a fixture, and it recovers the flag.**

The real process runs `python.exe -s ComfyUI\main.py …`, while `sys.argv` reports only `ComfyUI\main.py …` — no interpreter, no `-s`. Feeding exactly that split:

```
server sys.argv : ["ComfyUI\main.py", "--feature-flag", …]
relaunch argv   : ["ComfyUI\main.py", "-s", "--feature-flag", …]
re-attached -s  : true
```

`-s` is the flag that keeps a venv-isolated ComfyUI from picking up the user site directory, so losing it on relaunch is not cosmetic — it changes which packages the relaunched server can import.

**And the gate opens on the platform this is for.** Recovery is behind `if (!info.osArgvExact) return []`, which I chased because a Windows command line arrives as one flattened string and a `"flattened"` fidelity would have made the whole feature inert on Windows. It does not: `live-interpreter.ts` marks the Windows reading `argvFidelity: "exact"` (it tokenizes a real `CMD=` capture), and reserves `"flattened"` for the POSIX `ps` path where the regex fails. So the exact-only guard is not accidentally excluding the target platform.

## Two setup errors of mine, since they are the reason this took three attempts

- The first probe went through a shell heredoc and `"ComfyUI\main.py"` arrived as `ComfyUImain.py`. Same backslash-eating trap as three other places today; the fix is to build separators with `String.fromCharCode(92)` and write the probe with an editor rather than a heredoc.
- The second omitted `osArgvExact`, so recovery correctly returned `[]` and I briefly had a "does not re-attach `-s`" result that was entirely my own construction.

Both would have read as findings against this PR. Recording them because the same two mistakes would mislead the next person driving this function by hand — the guard is easy to miss, and its absence looks exactly like the bug.

Nothing blocking.

artokun:
**Driven against the real Desktop-managed ComfyUI running on this box — and the flag this PR exists to preserve is genuinely present.**

Captured from `Win32_Process` into a file and loaded, not retyped (my earlier run on this PR reconstructed `[]` purely because the fixture omitted `osArgvExact`):

```
osArgv[0..2] : ["C:\Users\...\ComfyUI\.venv\Scripts\python.exe", "-s", "ComfyUI\main.py"]
sys.argv[0]  : "ComfyUI\main.py"
```

```
interpreterFlagsFromOsArgv(exact:  true)  -> ["-s"]
interpreterFlagsFromOsArgv(exact: false)  -> []
```

Both directions are the intended ones. The corroboration rule holds on real input: `sys.argv[0]` is the RELATIVE `ComfyUI\main.py` while the OS reading carries an absolute interpreter path, and the case-and-separator normalisation matches them without the relative/absolute difference defeating it.

**What it prevents, concretely.** A relaunch rebuilt from `sys.argv` alone is:

```
<python> ComfyUI\main.py --feature-flag ... --enable-manager ...
```

— i.e. `-s` is gone. `-s` suppresses the per-user `site-packages` directory, so a ComfyUI deliberately launched without user site is relaunched *with* it, and can import a different set of packages than the process it replaced. That is a silent behaviour change in the restart path, not just a cosmetic argv difference, and it is exactly the class #2693 is about.

**The fidelity gate is the load-bearing part and it is doing real work.** `osArgvExact` is false on the POSIX `ps` fallback, where the flattened reading cannot distinguish an interpreter flag from a script argument containing spaces. Returning `[]` there means an unrecoverable command line reconstructs to nothing rather than to a guess — the same choice `read-minidump` makes with an unresolvable module, and the right one.

Nothing to change. Recording it because every earlier exercise of this function on this PR used a hand-built `ProcessInfo`, and two of those produced findings I had to withdraw.

artokun:
**Copilot's finding on this PR is real, and the part that makes it real is the part I had wrong.**

The corroboration used a local helper that lowercased and rewrote backslashes on every platform. On Linux that is incorrect twice over: `/ComfyUI/main.py` and `/comfyui/main.py` are different files, and a backslash is a legal POSIX filename character. Two DIFFERENT scripts could corroborate each other and splice interpreter flags out of a command line belonging to something else.

**I would have dismissed this as unreachable, on a belief I had written down.** My note said Windows sets `argvFidelity: "exact"` and the POSIX `ps` fallback sets `"flattened"`, so `osArgvExact` could never be true off Windows. Half right:

```
live-interpreter.ts:333   argvFidelity: "exact"       Win32_Process CommandLine
live-interpreter.ts:371   argvFidelity: "exact"       /proc/<pid>/cmdline   <-- LINUX
live-interpreter.ts:391   argvFidelity: "flattened"   ps fallback
live-interpreter.ts:399   argvFidelity: "flattened"
```

The `/proc` path is exact **by construction** — the kernel NUL-separates argv, so nothing is guessed. Only `ps` is flattened. So `osArgvExact` is true on Linux and this comparison really runs there.

**Fixed** by using the module's own `sameRecoveryPath` / `normalizeRecoveryPath`, which preserves POSIX spelling and folds case and separators only on win32 — one rule rather than two that can drift apart.

Verified:

- mutation — restoring the unconditional fold fails the new source-pin test;
- 76 tests green across the flags and process-control suites, `tsc` clean;
- and the real Desktop-managed ComfyUI on this box still recovers `["-s"]` from its captured 442-character command line, so the Windows behaviour this PR exists for is unchanged.

The new behavioural test asserts `["-s"]` on win32 and `[]` elsewhere for a case-and-separator-differing argv, which is the correct expectation on each host rather than a single answer asserted everywhere.

artokun:
**Ran `interpreterFlagsFromOsArgv` against the REAL ComfyUI launch on a Comfy Desktop machine, and it is not a hypothetical case — this install genuinely carries an interpreter flag.**

The live backend serving 8188 here, from `Win32_Process.CommandLine`:

```
C:\...\ComfyUI\.venv\Scripts\python.exe  -s  ComfyUI\main.py  --feature-flag show_signin_button=true …
                                         ^^
```

Tokenised and fed in as `osArgv`, with `argv` starting at `main.py` the way `sys.argv` would report it:

```
exact=true   -> ["-s"]
exact=false  -> []
no osArgv    -> []
```

**The positive case is the point.** `-s` suppresses the user site-packages directory. A relaunch that drops it puts `%APPDATA%\Python\...\site-packages` back on `sys.path`, where it can shadow the venv's own packages — a ComfyUI that starts and then misbehaves in ways that look nothing like a launch bug. This machine would have hit exactly that.

**Both negative cases fail closed, which matters more than the positive one.** `osArgvExact: false` returns `[]` rather than guessing: on the platforms where argv is tokenised from a flattened string, it is derived from `commandLine` and is not independent evidence. Inventing a flag there would be worse than dropping one — it would put an interpreter argument into a relaunch that the process never actually received.

One real-world wrinkle this install exposes, which no fixture would have: the port-holding process reports `ExecutablePath` as `…\standalone-env\python.exe` while its command line begins `…\.venv\Scripts\python.exe` — a venv trampoline. The extractor reads the command line, so the mismatch does not disturb it.

Not covered: this box has one flag. A launch carrying several (`-s -E -X utf8`) or a flag with a separated value is still fixtures-only.

PR #2850, opened 2026-09-04
fix(2849): stop advising an env var the orchestrator deletes — in source AND catalogs

### LINKED ISSUE #2849: Three credential remedies name ANTHROPIC_API_KEY, which the orchestrator deletes at startup
### What happens

Three user-facing messages tell a blocked user to set `ANTHROPIC_API_KEY`. The orchestrator **deletes that variable from its own `process.env` at startup**, unconditionally, and nothing puts it back — so following the advice changes nothing, and there is no way for the user to see why.

```ts
// src/orchestrator/index.ts:1150
// Subscription lane: the background agent must authenticate against the user's
// claude.ai login, never an API key. Unset the key for the SDK subprocess.
delete process.env.ANTHROPIC_API_KEY;
```

The delete is correct for its stated purpose. The problem is the advice pointing at it.

### The three messages

| where | says |
|---|---|
| `index.ts:4356` (pi, no credential) | *"set a provider API key (e.g. **ANTHROPIC_API_KEY** / OPENAI_API_KEY / CEREBRAS_API_KEY) and restart the orchestrator"* — names it **first** |
| `cli-remedy.ts:80` (pi remedy) | *"API key (**ANTHROPIC_API_KEY** / OPENAI_API_KEY / …)"* |
| `cli-remedy.ts:91` (Claude, no credentials) | *"run `claude setup-token`, or set **ANTHROPIC_API_KEY**"* |

The third is the worst of the three: `claude-backend.ts:116` states the Claude lane's auth is OAuth-file-based and *"ANTHROPIC_API_KEY is deliberately unset"*. So that fallback is advice to set a variable the lane ignores **by design**, on top of being deleted.

### Why nothing restores it

- **The dotenv store does not.** `env-file.ts:103` only fills keys that are `undefined`, and it runs at import time — before the delete.
- **The panel cannot.** `COMFYUI_SECRET_ENV_ALLOWLIST` (`panel-secrets.ts:91`) does not include it, and `CREDENTIAL_SLOTS` (`:2372`) has no `anthropic` slot — the slots are openrouter, civitai, huggingface, google, runcomfy, runpod, registry.
- **So `piCredentialPresent()` can never see it.** `pi-credentials.ts:50` maps `anthropic: ["ANTHROPIC_API_KEY", …]` and tests `procEnv[k]`, where `procEnv` defaults to the `process.env` the delete emptied.

`OPENAI_API_KEY` and `CEREBRAS_API_KEY` in the same sentence are unaffected and do work. It is specifically the first-named option that cannot.

### Why it matters

This message only appears to a user who is **already blocked** and looking for the way out. They set the first thing suggested, restart as instructed, get the identical message, and have no signal that this particular variable is special. Same shape as #1477's *"point COMFYUI_MCP_PI_PATH at the real pi/pi.exe"* (a file the npm package does not ship) and panel#2044's setting-name check — advice that names something the user cannot act on.

### Fix

Stop naming `ANTHROPIC_API_KEY` where it cannot work:

- **pi paths** — name `OPENAI_API_KEY` / `CEREBRAS_API_KEY` (which do work), and `pi` + `/login`.
- **Claude path** — `claude setup-token` is already first and is the real remedy; drop the `ANTHROPIC_API_KEY` fallback rather than offering one the lane ignores by design.

If instead the intent is that a provider key *should* reach pi, the delete needs narrowing to the Claude SDK spawn env rather than the orchestrator's own `process.env` — but that is a behaviour change and the message fix stands on its own either way.

### Not claimed

I have not established what the pi **subprocess** ultimately sees; it may read credentials by its own means. Even so the readiness check above it is wrong in one direction or the other: either the advice does nothing, or the check reports "no credential" for a pi that would in fact authenticate.

Found by verifying user-facing claims against the code that implements them, not from a report.

artokun:
Closes #2849.

`runPanelOrchestrator()` unsets `ANTHROPIC_API_KEY` at startup so the Claude SDK subprocess authenticates against the user's claude.ai login rather than a key. That is correct. Three of the orchestrator's own messages went on naming that variable as the remedy — and one pairs it with **"restart the orchestrator"**, which is precisely the step that re-runs the delete. A blocked user follows the advice, restarts as instructed, and gets the identical message with no signal that this variable can never take effect.

Nothing restores it: the dotenv store only fills UNSET keys and runs at import time, the panel secret allowlist omits it, and there is no `anthropic` credential slot.

## Two corrections to what I filed on the issue

**I claimed five identical sites. There are three — plus two that are a different defect.** The comfyui MCP child is a *separate process with two lanes*: spawned by the panel it gets a constructed env (`panel-secrets.ts`: *"the orchestrator constructs the child env instead of inheriting it"*) whose allowlist omits the key, so the advice is dead; run standalone from the user's own MCP client it inherits the real shell env and **the same advice is correct**. Deleting it there would have removed guidance standalone users need. `src/services/train-caption.ts` is therefore made lane-aware via `COMFYUI_MCP_TAB`, the marker the orchestrator already stamps per spawn.

`src/tools/train.ts` is the one child-side string that had to lose the name outright: it is a **tool description**, read by the *model* rather than by a user who hit an error, so it taught the dead remedy as a standing fact and would repeat it unprompted on paths no error string covers.

**Fixing the TypeScript alone would have shipped nothing.** `tr(key, fallback)` resolves the **catalog first**, so `say.pi_no_credential` kept naming the variable in all twelve locales — `en` included, which every English user resolves through — while the source read clean and `i18n:check` stayed green (it compares key *counts*, not meaning). `locales/en/main.json` is regenerated via `npm run i18n:build`; the other eleven have it removed in place.

## The gate

`check:env-advice` fails the build when a message names a variable the process deletes — scanning orchestrator source **and** every locale catalog.

Scoped to the orchestrator deliberately: a repo-wide version demands deleting advice that is right for standalone users, and import reachability cannot separate the lanes (the orchestrator imports `tools/train.ts` for its catalog while the tool executes in the child).

The scanner underneath is a mode stack, not a regex, because **two simpler versions each returned a wrong count on the real tree**:

- a quote-delimiter regex (`[^"'\`]{40,}`) cannot cross a quote character, so literals containing an escaped backtick or a nested double quote split into fragments too short to match — that missed two real sites;
- skipping `${…}` by counting braces desyncs on a brace inside a nested string, and every literal after that point is read in the wrong mode — that lost a site 4,000 lines later.

Both failures were silent: a smaller count that still looked like a clean pass. The tests assert exact extraction, and each fixture pins one of those constructs.

## Verification

- **9/9 mutations** behave as required — each of the three source fixes reverted independently (red), a brand-new advice site in another orchestrator file (red), a non-English catalog and a regressed `en` catalog (red, each attributed to the right file), and three controls that must NOT fire: a bare key with no prose, a mention in a comment, and a catalog value too short to be advice.
- `tsc --noEmit` clean · anti-slop "none added" · `i18n:check` locales OK · 105 tests pass across the six files touching this code, including 6 new ones.
- Full suite runs here in CI — a run from inside `.claude/worktrees/` matches vitest's own exclude and collects nothing.

`package.json` gains one script line (no dependency change).


artokun:
Pushed a third commit — from turning the adversarial pass on this PR's own gate.

**`return /["']/` was read as a division.** The scanner decided regex-vs-division from the preceding *character*, which for `return` is `n`. The quote inside the regex character class then opened a string literal that swallowed the next real message — the exact silent-undercount this gate exists to prevent, still living in the gate. The decision is now made on the preceding **token**.

Measured rather than asserted, in both directions:

- The fix changes extraction in **9 of 1140 files**, `src/orchestrator/panel-tools.ts` among them — so the old scanner genuinely mis-parsed files inside the gate's own scope.
- It changed **no verdict**. A violation planted at 95% depth in the two worst-affected files is caught by the old scanner *and* the new one: an unterminated string ends at the next quote, so parity recovers before the end of the file.

So: a real correctness bug, with no observed effect on this tree. I'd rather state that than the stronger version. `src/orchestrator` holds a dozen keyword-preceded regex literals and none contains a quote, which is why only a deliberate probe surfaced it — and why the regression test is a synthetic probe rather than a repo excerpt.

Mutation coverage is now 9/9 (6 source, 3 catalog), including three controls that must not fire.

artokun:
Retracting the caveat in the PR body: I said the full suite could only run in CI. It can run locally after all, and I've now run it on this branch.

**777 files, 14,220 tests.** Six files reported red; **none is caused by this PR**:

| file | tests | cause |
|---|---|---|
| `install-node-git-expectation` | 1 | pre-existing on `origin/main` — #2829 |
| `install-accepted-never-enqueued` | 9 | pre-existing on `origin/main` — #2829 |
| `panel-image-relay` | 1 | the deadline flake — #2823 (`expected 0 to be greater than 0`) |
| `package-hooks` | 1 | worktree artifact: `files` lists `dist`, which is absent until built. **Passes after `npm run build`** |
| `panel-secrets-slots` | 2 | **passes in isolation** (100s under the concurrent run) |
| `console-secrets` | 2 | **passes in isolation** (144s under the concurrent run) |

The last three are not failures at all, which is worth stating plainly because the concurrent run makes them look identical to real ones.

**How the local run was possible.** `vitest.config.ts` excludes `**/.claude/**`, and excludes are additive, so a bare `npx vitest run` from inside a worktree collects zero files. But passing a path *filter* — `npx vitest run src/__tests__/` — matches all 777. The received wisdom (cut the tree outside the repo and junction `node_modules`) is not needed, and is actively worse: borrowing another tree's `node_modules` is what faked a "pre-existing red on main" back in #1948.

So this PR is clean on the full local suite as well as on `tsc`, lint, the 7 scanner tests and 9/9 mutations.

artokun:
Adversarial pass on the gate itself (mine — codex is out through Sept 6). **A gate is only worth its build time if it fails when it should and stays quiet when it should. Both directions verified by mutation.**

| mutation | gate |
|---|---|
| remedy naming `ANTHROPIC_API_KEY` added to an orchestrator **source** message | **FAIL (exit 1)** |
| the same string added to **`locales/en/main.json`** | **FAIL (exit 1)** |
| the same name in a `//` comment **and** a `/* */` comment | **pass (exit 0)** — correctly ignored |
| unmodified tree | pass — "1 env var(s) unset at startup, none named as a remedy across 68 orchestrator files" |

The catalog case is the one I most wanted to see fail, and it does. `tr(key, fallback)` resolves the **catalog first**, so a message fixed only in TypeScript still ships the old advice to every locale carrying the key — including `en`, which is generated. A gate that scanned only source would have passed the exact state this PR exists to prevent.

The comment case matters just as much in the other direction. A gate that fires on prose *about* a deletion is one someone weakens on the first false positive, and the failure text anticipates that:

> If a name here is prose ABOUT the deletion rather than advice, move it into a comment (this scanner ignores comments) instead of weakening the gate.

Naming the escape hatch in the failure message is what keeps the gate from being narrowed later — the pressure to weaken arrives at the moment someone is blocked, not while reading the code.

The failure output also names the i18n hazard directly ("the catalog WINS over the in-source fallback, so fixing only the source changes nothing. Regenerate with `npm run i18n:build`"), which turns a red build into an instruction rather than a puzzle.

Nothing blocking. Verified rather than assumed, since a gate that cannot fail is worse than no gate — it reads as coverage.

artokun:
**Three findings: one real and fixed, one refuted with evidence, one accepted as a gap.**

**1. Bracket-form delete — real, fixed.** Discovery matched `delete process.env.NAME` only, so `delete process.env["NAME"]` left the set empty for that key and the gate reported success while the advice it guards was still shipping. Nothing about that rename is wrong, which is exactly what makes it dangerous — a gate a harmless refactor can silence is not a gate. Both forms now match, either quote style. The gate still passes on the real tree, and a direct probe confirms all three spellings are discovered.

**2. The regex-literal scanner — refuted.** The claim is that `stringLiterals` fails to recognise a regex after keywords like `return`, so a regex containing quotes is parsed as a string and desynchronises the rest of the file, naming `panel-tools.ts:532` and `:19040`.

The line numbers have shifted, but the two sites it means are real and are the exact shape described:

```
536:    return /^(yes|allow|...|i'?m? ?18|18\+?|adult)/i.test(     <- apostrophe
19044:  return /^Panel tab .+? did not reply to "[^"]*" within \d+ ms .../.test(   <- double quotes
```

So I ran the scanner rather than reasoning about it. On the real `panel-tools.ts` it extracts 5,487 literals and reports nothing at either site beyond the legitimate neighbours. And on minimal fixtures of each shape, with `BEFORE`/`AFTER` sentinels either side — the sentinel after the regex is the desync detector:

```
  OK   apostrophe-after-return   ["BEFORE","AFTER"]
  OK   regex-after-arrow         ["BEFORE","AFTER"]
  OK   regex-after-or            ["BEFORE","AFTER"]
  OK   division-not-regex        ["BEFORE","AFTER"]
```

No desync, and the regex innards are not extracted as strings. The last case matters too: a scanner that "fixed" this by treating every `/` as a regex would break on `10 / 2`, and this one does not.

**3. The lane distinction is untested — accepted, not done here.** The caption tests never set `COMFYUI_MCP_TAB` and only assert the error contains `/login`, so the panel-lane vs standalone split can regress silently. That is a real gap and I am leaving it visible rather than adding a test I have not been able to exercise against a live lane.

PR #2852, opened 2026-09-04
docs(2221): stop claiming a panel counterpart that does not exist

artokun:
Addresses half of artokun/comfyui-mcp-panel#2221 — the half that lives in this repo. The panel-side half stays open there.

`looksLikeSystemStats`'s doc comment claimed it was *"the same predicate the panel's reboot certification uses"*. There is no such predicate. I searched every `/system_stats` consumer in the panel: both take the decoded body as-is.

The comment now says what is actually true on the other side, and cites the issue so the asymmetry stays findable.

## Why bother with a comment

A false provenance is worse than none. Someone syncing the two sides looks for the counterpart, finds nothing, and either duplicates the predicate or concludes the panel already guards and moves on. The second outcome is how an asymmetry stays invisible.

## A correction to the issue I filed

The issue implies the panel is exposed to a gateway's `200 {"error": …}` envelope. Having now read both consumers, that is overstated — **both already fail safe**:

- `cmcp-training-ui.js` → `data?.devices?.[0]`, and returns `null` when absent.
- `comfyui-mcp-panel.js` → an unknown platform falls back to POSIX separators, which is the deliberate fail-closed direction documented right there.

So the panel does accept any 2xx as stats, but nothing downstream trusts it. That makes the panel-side predicate a clarity improvement — a clear "that was not ComfyUI" instead of a silent `undefined` — rather than the bug fix I implied when filing. I have said so on the issue rather than quietly narrowing it here.

Comment only, no behaviour change, so no changelog entry — which is deliberate, not an oversight.

### DIFF
--- a/src/comfyui/client.ts
+++ b/src/comfyui/client.ts
@@ -243,7 +243,15 @@ export async function ensureConnected(): Promise<Client> {
 /** True when a decoded /system_stats body has the recognizable ComfyUI shape (a
  *  `system` object and/or a `devices` array). A bare 2xx carrying an API
  *  gateway's own JSON envelope is NOT ComfyUI and must not be handed on as
- *  stats — the same predicate the panel's reboot certification uses. */
+ *  stats.
+ *
+ *  This comment used to say the panel-side reboot certification uses the same
+ *  predicate. It does not — panel#2221. Every panel /system_stats consumer takes
+ *  the body as-is; the two that exist read defensively (`data?.devices?.[0]`, and
+ *  an unknown platform falling back to POSIX), so they fail safe rather than
+ *  certifying the document. The claim is recorded here because a false provenance
+ *  is worse than none: someone syncing the two sides looks for the counterpart,
+ *  finds nothing, and either duplicates it or assumes the panel already guards. */
 function looksLikeSystemStats(body: unknown): boolean {
   if (!body || typeof body !== "object") return false;
   const b = body as { system?: unknown; devices?: unknown };


artokun:
Adversarial pass (mine — codex is out through Sept 6). **Both claims verified independently, including one I initially thought was miscounted.**

## The central claim holds

I searched the panel for anything that certifies a `/system_stats` response is a genuine stats document. There is nothing: no shape predicate, no counterpart to `looksLikeSystemStats`. The comment was asserting a provenance that does not exist, and your reasoning for why that is worse than no comment is right — a reader who goes looking, finds nothing, and concludes the panel already guards is exactly how an asymmetry stays invisible.

## The enumeration is right, which I doubted

A `grep` for `system_stats` in the panel surfaces a third site — `lib/input-asset.js`, "Interpret a ComfyUI `/system_stats` payload" — and I went in expecting to report the count as off by one. It is not: `comfyui-mcp-panel.js:12331` fetches and immediately delegates to `inputPathsUseWindowsSeparators(stats)` in that file, so it is one consumer across two files, not a third. "Both take the decoded body as-is" is accurate.

## And that file substantiates your correction

You corrected the issue's implication that the panel is exposed to a gateway `200 {"error": …}` envelope, on the grounds that both consumers fail safe. `input-asset.js` says so in its own words, which is stronger evidence than an inspection:

> Any missing/malformed shape returns false (POSIX semantics), so an unreadable stats payload can never enable a split the server would not perform.

So the fail-safe property is not incidental there — it is a documented design decision with a named consequence. Worth citing in the issue when you close out the panel half, because it turns "I read both and they look safe" into "the code states the invariant".

Nothing blocking. Retracting a claim you filed yourself, in the PR that fixes the other half, is the right way to handle it.

PR #2873, opened 2026-09-04
fix: stop REQUESTING the emoji font from the orchestrator's ready banner

artokun:
For **comfyui-mcp-panel#2023**. Every issue reference here is repo-qualified deliberately — a bare `#2023` in this repo resolves to comfyui-mcp#2023, an unrelated closed RunPod issue.

**This does not repair comfyui-mcp-panel#2023.** The null-deref is upstream, as the report says. It is the same class of mitigation as comfyui-mcp-panel#2210: stop making the crashing path mandatory.

## The gap

comfyui-mcp-panel#2210's mechanism is specific — `U+FE0F` does not merely *permit* emoji presentation, it **requests** it, pinning the glyph to `seguiemj.ttf`, the file KB5120998 replaced two days before the crash cluster. It stripped the selector from nine panel strings and deliberately kept the glyphs.

The orchestrator renders on the same canvas:

```
src/orchestrator/index.ts   bridge.push({ type: "say", text: readyBannerText(...) }, panelTab)
```

Same font stack, same DirectWrite fallback. So a panel-only sweep left `banner.ready.copilot` carrying the identical `U+26A0 U+FE0F` sequence. Measured on `main` before this change:

```
locales/<12>/main.json           U+FE0F = 1 each (12)
src/orchestrator/ready-banner.ts U+FE0F = 2 (one in a comment)
```

After: **0 everywhere**, with `U+26A0` intact.

Affects **Copilot-backend sessions**, where that banner is the first bubble.

## The test pins both directions

An assertion of "no `U+FE0F`" alone is satisfied by deleting the warning sign — the change comfyui-mcp-panel#2210 explicitly rejected. So the glyph's survival is asserted too, and there is a case pinning that the catalog directory read actually found catalogs, so a silent `[]` cannot make the `it.each` vacuous.

**Mutation-verified, each guard independently:**

| mutation | caught by |
|---|---|
| restore `U+FE0F` in the source | `carries no U+FE0F in the ready-banner source` |
| restore `U+FE0F` in one catalog (`ja`) | `catalog ja carries no U+FE0F` |
| delete `U+26A0` from the source | `still carries the warning sign itself` |
| delete `U+26A0` from the `en` catalog | `keeps the warning glyph in the English catalog` |

Each mutation failed exactly one assertion; the tree returns 16/16 green.

## Verification

`locales/en/main.json` is **generated** from the `tr()` call sites (`scripts/i18n-build-en.mjs` — "never hand-edited"), so it was rebuilt with `npm run i18n:build`. Its diff is one line, which also rules out the regeneration having disturbed unrelated keys.

`npm test` reports three failures — `vitest`, `check:blog-stale`, `asset-counts`. **All three fail identically on clean `origin/main` in this environment**, controlled per-file:

| check | this branch | `origin/main` |
|---|---|---|
| `console-secrets.test.ts` alone | passes | passes (its full-suite failures are concurrency interference) |
| `install-accepted-never-enqueued.test.ts` alone | fails | **fails** |
| `check:blog-stale` | fails | **fails** |
| `asset-counts` | fails | **fails** |

None is caused by this change. `install-accepted-never-enqueued` failing on `main` here is likely environment-specific — this box has a live ComfyUI on 8188 that the Manager-queue tests can see — and is not something this PR touches.

## What I have not done

No Comfy Desktop here, so nothing about the crash itself is reproduced. This is a static parity change against comfyui-mcp-panel#2210's own stated mechanism.

The wider half is deliberately **not** in this PR: `U+1F7E2` (`🟢`) appears 216 times across the catalogs — every `banner.ready.*`, every locale, the first bubble of every session — but it is `Emoji_Presentation=Yes`, so there is no selector to strip and closing it means changing the character, which is the change comfyui-mcp-panel#2210 declined to make. Same exposure, different and more opinionated fix. Left for a decision rather than folded in here.


artokun:
Both findings were live. Fixed in `fix(review): scope the glyph guard to the string it means to pin`.

**1. The glyph guard was as weak as you say — and my mutation had hidden it.**

I reproduced your case before changing anything: delete `U+26A0` from the Copilot fallback only, leave the doc comment on line 31 alone.

```
U+26A0 remaining in file (doc comment): 1
copilot string still has it?  false
      Tests  16 passed (16)
```

16/16 green on the exact regression the assertion claims to pin. My mutation had passed only because it deleted the comment's glyph too — a coarser edit than the real failure.

The asymmetry is now stated in the test, because it is the general rule here: **"no selector" is `=== 0` over a whole file and is sound** — any stray occurrence anywhere fails it. **"Glyph stays" is `> 0` and cannot be asserted file-wide at all.** So the positive half is now asserted against the rendered banner and against the specific `banner.ready.copilot` key in each catalog.

Measured while fixing it, and recorded in the test rather than assumed: `"en"` resolves to the **source fallback** (English is the source catalog) while every other locale resolves from its **catalog file** — `readyBannerText` returns English for `"en"` and Korean for `"ko"`. That is why your surgical mutation is caught by the render case. Coverage now splits cleanly: the render case pins the fallback, the per-catalog cases pin the twelve files, the all-backends case pins rendered output for the other eleven locales.

Newly caught mutations, each by exactly one assertion:

| mutation | caught by |
|---|---|
| glyph removed from the fallback only (comment intact) | `renders the Copilot banner with the glyph…` |
| glyph removed from the `en` catalog's copilot key only | `catalog en keeps the warning sign in banner.ready.copilot` |
| selector restored in one catalog (`ko`) | 3 assertions |

28 tests, up from 16.

**2. The bare `#2210` — correct, and an embarrassing miss.**

I had already guarded this exact trap in the commit message for `#2023` (a bare `#2023` here resolves to comfyui-mcp#2023, an unrelated closed RunPod issue) and then left a bare `#2210` in the changelog one number over. Every panel reference in the entry is now repo-qualified.

Thanks — the first one in particular was a hole in my own verification, not just in the test.

PR #2874, opened 2026-09-04
fix: retract the canvas claim when a policy withholds every panel_* tool

artokun:
For **comfyui-mcp-panel#291**. Every reference is repo-qualified on purpose — a bare `#291` here resolves to comfyui-mcp#291, an unrelated merged PR in this repo.

**This does not prove that issue's cause.** I have not reproduced it and do not know the reporter's environment. The defect below is real independently.

## The gap

`PANEL_SYSTEM_APPEND` tells the agent it can see and edit the user's canvas through `panel_*`. The orchestrator already retracts that claim twice — `PI_CAPABILITY_OVERRIDE` (pi has no MCP client) and `NO_PANEL_TOOLS_OVERRIDE` (the loopback bind failed). Neither covers the operator's tool-surface policy.

```
src/tools/tool-surface-filter.ts:232
  const PANEL_SURFACE = ["panel_*"];
  safe:     [...MUTATING_TOOLS, ...PANEL_SURFACE],
  readonly: [...MUTATING_TOOLS, ...EXECUTION_TOOLS, ...PANEL_SURFACE],
```

Both presets deny the whole surface by glob, and `docs/configuration.mdx` recommends `COMFYUI_MCP_TOOL_PRESET=safe` **verbatim** for a hosted deployment. Under that ordinary configuration both registration paths — `createPanelMcpServer` (claude) and `registerPanelTools` (the HTTP lane) — filter through the same `resolveToolSurfacePolicy()` and register **zero** panel tools, while the prompt still promises them.

## Why the existing guard misses it

```js
export function panelToolsRetraction(backend, panelToolsAvailable) {
  if (panelToolsAvailable) return "";
  if (backend === "pi") return "";
  // claude drives the canvas through its own IN-PROCESS panel server, so a failed
  // HTTP bind takes nothing away from it.
  if (backend === "claude") return "";
```

That reason is **true of the failure it covers**. The policy is a second, independent way to lose the same tools, and the exclusion — correct on its own terms — cannot see it. A correct guard blocking a different question.

The call site is worse than the exclusion: `index.ts` notes *"the claude lane … never reaches here"*, so that whole retraction path is HTTP-lane only. The claude lane had **no** retraction mechanism under any circumstance.

## Design notes

- **Applied by wrapping** `systemAppendForBackend`, not by appending at each of its three returns. "Fixed one of two paths" is this defect's own family — the tool-surface wiring tests exist because of it — and a wrapper cannot miss one.
- **Computed once.** The policy is env-derived and fixed for the process lifetime, and `buildPanelToolDefs()` walks ~96 defs.
- **Not keyed on backend.** Unlike a bind, the policy governs every lane; a backend check would reintroduce exactly this gap. A test asserts that.
- **Says only what was observed.** It names the variables an operator can change — a bind failure cannot say *why*, a policy can — declines to speak for the headless server the same policy filters separately, and tells the agent not to keep searching, since the tools were never registered and are absent from the deferred catalog too. That last point is the loop that made comfyui-mcp-panel#291 expensive: its reporter established the absence with `ToolSearch`, repeatedly.
- **`panelSurfaceFullyWithheld` returns false** when the policy is inactive, when any one panel tool survives (the documented `PRESET=safe` + `ALLOW=panel_graph_outline` opt-back-in is a *configured* surface, not a capability to retract wholesale), and when the def list is empty — so a build that has the tools is never told it does not. The def list is injectable **only** so that last guard is reachable from a test; an unreachable guard is an untested one.

## Verification

**Four mutations, each caught by the intended assertions:**

| mutation | caught by |
|---|---|
| predicate forced `false` | the three preset/allow-list cases |
| empty-list guard removed | `does NOT retract when the def list is empty` |
| suffix unwired from the prompt | `systemAppendForBackend applies it to every return path` |
| retraction stubbed to `""` | the four content assertions |

14 tests. With no policy configured the suffix is `""`, so behaviour is unchanged for every existing path.

`npm test` reports the same three failures — `vitest`, `check:blog-stale`, `asset-counts` — that fail identically on clean `origin/main` in this environment; controlled per-file, including the two that newly appeared in this run (`model-management`, `panel-secrets-slots`), which pass in isolation on both. The failing set varies run to run, which is the concurrent-suite interference signature.


artokun:
Correct, and fixed. I inserted `panelSurfaceFullyWithheld` between `createPanelMcpServer`'s JSDoc and its declaration, so that comment was documenting my helper rather than the function it was written for.

Moved the helper and its own JSDoc above it, and verified rather than eyeballed — there are now **zero** characters between the original comment's `*/` and `export function createPanelMcpServer(`:

```
29488  * Whether the configured tool-surface policy withholds the ENTIRE `panel_*` surface.
29510 export function panelSurfaceFullyWithheld(
29527  * Build the per-tab live-graph MCP server for the Claude (in-process Agent SDK)
29534 export function createPanelMcpServer(
```

14 tests still green, typecheck clean, changelog gate passes.

Worth noting for anyone reading later: this is the second time today one of your findings was a hole in my *verification* rather than only in the code — on comfyui-mcp#2873 the "glyph stays" assertion was satisfied by the file's own doc comment, and my mutation had hidden it by deleting that comment too. Both were cheap to fix and neither was visible from a green run.

artokun:
**The same defect exists one step over, for the HEADLESS surface. Reporting rather than folding it in — the fix is a design decision, not a mechanical extension of this PR.**

This PR retracts the `panel_*` claim because the policy can remove that surface entirely. The policy filters the headless surface too (`src/tools/index.ts` applies `resolveToolSurfacePolicy()` in the same position), and `PANEL_SYSTEM_APPEND` names specific headless tools by hand.

Measured against `origin/main` — tools named in the persona that a preset withholds:

| tool | mentions | withheld by |
|---|---|---|
| `list_packs` | **10** | safe + readonly |
| `download_model` | 2 | safe + readonly |
| `get_image` | 2 | readonly only |
| `queue` | 2 | safe + readonly |
| `upload_image` | 2 | safe + readonly |
| `apply_manifest` | 1 | safe + readonly |
| `install_comfyui` | 1 | safe + readonly |
| `install_custom_node` | 1 | safe + readonly |
| `enqueue_workflow` | 1 | readonly only |
| `generate_image` | 1 | readonly only |

**10 distinct tools, 23 mentions.** Under `COMFYUI_MCP_TOOL_PRESET=safe` — the value `docs/configuration.mdx` recommends verbatim — the prompt still teaches the agent to reach for `list_packs` ten times over, and it is not there.

It also cuts against the filter's own stated intent, in `src/tools/index.ts`:

> Logged, never pushed to the model: **the point is that it does not learn these exist.**

The persona tells it they exist, in the same process, from a string the filter never sees.

**Why not in this PR.** `panel_*` is all-or-nothing, so one retraction is correct and testable. The headless surface is partially withheld, so the honest fix is per-tool — filter the persona text against the live policy, or generate those mentions from the surface rather than hand-writing them — and that is a design call about how the persona is authored, with a real prompt-size and translation cost (the persona is one of the twelve-locale catalog strings). Retracting wholesale here would be the "attaches a new false claim while retracting an old one" failure this whole override family exists to avoid.

**On the measurement.** I got this number wrong twice before trusting it: a loose scan said 12 by matching tool names inside comments, and a strict one said 0 because a heredoc collapsed `\b` to `\b`, making the boundary a literal backspace character that matches nothing. The table above comes from a scan that first proves it can find a case I confirmed by eye (`list_packs`, 10) and exits otherwise.

artokun:
**Pushed a precedence fix — my own change had created a case where the agent gets two different causes for one absence.**

Auditing the composition rather than only the new function, I found both retractions can fire at once. Measured:

```
bindFires: true    policyFires: true    bothFire: true
bindSaysRestart: true          policySaysOperator: true
```

That happens on a codex/gemini tab whose loopback bind failed **and** whose policy also withholds the surface. Before this PR only the bind retraction existed on that lane, so the combination is one I introduced.

It matters because the two prescribe different remedies. `NO_PANEL_TOOLS_OVERRIDE` tells the agent to say the tools "failed to start this run" and that the user "cannot reach the canvas until the orchestrator is restarted" — and a restart cannot restore a tool a policy withholds. That is the remedy-half failure this override family exists to prevent, reintroduced by composition rather than by either message being wrong on its own.

**The policy now wins.** It is the binding constraint either way (the tools stay withheld whether or not the bind succeeds), and it names a cause an operator can act on. `panelToolsRetraction` takes the fact as an optional third argument defaulting to `false`, so every existing caller and test is unchanged.

One policy read feeds both the prompt suffix and the suppression, so the two cannot disagree about whether the surface was withheld — the wiring test pins that both sites read the same variable.

**Three more mutations, seven total, each caught by the intended assertions:**

| mutation | caught by |
|---|---|
| precedence guard removed (both fire) | `the policy retraction suppresses the bind retraction` |
| precedence inverted | that case plus 2 regression cases |
| flag dropped at the call site | `systemAppendForBackend applies it to every return path` |

19 tests, up from 14. The wiring test also caught my own refactor mid-change — it failed the moment I hoisted the policy read, which is what it is for.

PR #2877, opened 2026-09-05
fix(2876): do not advertise a nonexistent refresh_nodes install action

### LINKED ISSUE #2876: Fix impossible refresh_nodes remediation in download status
download_model action:"status" reported that newly downloaded ipadapter files were on disk but not visible, then instructed the agent to call install_comfyui action:"refresh_nodes". That action does not exist in install_comfyui's current JSON schema. The actual panel_refresh_nodes call was blocked because the open tab was one patch behind the installed panel bundle, leaving the reported remediation impossible from the active session.

### Steps to reproduce
1. Download a model into a model category not yet visible in the connected ComfyUI inventory.
2. Call download_model action:"status".
3. Observe remedy: install_comfyui action:"refresh_nodes".
4. Describe install_comfyui; its action enum has no refresh_nodes.

### Exact error (scrubbed)
panel_refresh_nodes returned stale_bundle: tab 0.15.173, installed panel 0.15.174.
download_model status recommended a nonexistent install_comfyui action:"refresh_nodes".

### Fix
Change download_model's status remediation to name panel_refresh_nodes only when the live bundle can execute it; for a stale panel bundle, instruct a hard refresh first. Do not advertise install_comfyui action:"refresh_nodes" unless that action is actually implemented and declared.

### Environment
Windows 11 / ComfyUI 0.34.3 / frontend 1.49.6 / RTX 4090 22.5 GB / comfyui-mcp 0.52.195 / live panel 0.15.173, installed panel 0.15.174.

artokun:
## What

`download_model` `action:"status"` no longer tells the agent to call `install_comfyui (action:"refresh_nodes")` when a just-landed file is on disk but not yet in the live listing.

- That action is not on `install_comfyui`'s schema.
- The note names `panel_refresh_nodes` only when the live panel bundle can execute it (advertised tab version ≥ the `refresh_nodes` floor, and not behind the installed pack).
- A stale tab (reporter: live 0.15.173, installed 0.15.174) is told to hard-refresh first (`Ctrl+Shift+R`); `panel_refresh_nodes` is only the follow-up after the tab loads the installed bundle.
- An unobserved or too-old live bundle gets hard-refresh / restart, not a tool it cannot run.

## Why

#2876: status reported ipadapter files on disk but not visible, then instructed `install_comfyui action:"refresh_nodes"`. `describe_tool` has no such action. The real `panel_refresh_nodes` call was blocked because the open tab was one patch behind the installed panel, so the advertised remediation was impossible.

Fixes #2876

## Tests

```
npx vitest run src/__tests__/services/placement-stale-listing.test.ts
```

26 passed. New cases fail on main (they require the fake install action to be gone, `panel_refresh_nodes` only when the live bundle can run it, and hard-refresh first for tab 0.15.173 vs installed 0.15.174).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ All notable changes to this project are documented here. This project adheres to
 
 ## Unreleased
 
+### MCP
+
+#### Fixed
+
+- **`download_model` `action:"status"` does not advertise `install_comfyui action:"refresh_nodes"` when a just-landed file is on disk but not yet listed (#2876).** That action is not on `install_comfyui`. The note names `panel_refresh_nodes` only when the live panel bundle can run it; a tab behind the installed pack is told to hard-refresh first.
+
 ## [0.52.196] - 2026-09-04
 
 ### MCP
--- a/src/__tests__/services/placement-stale-listing.test.ts
+++ b/src/__tests__/services/placement-stale-listing.test.ts
@@ -15,9 +15,20 @@
 // listing, refresh. Outside ⇒ genuinely misplaced, move.
 
 import { describe, expect, it } from "vitest";
-import { isUnderRoot, notVisibleVerdict } from "../../services/model-resolver.js";
+import {
+  isUnderRoot,
+  notVisibleVerdict,
+  staleListingRefreshRemedy,
+} from "../../services/model-resolver.js";
 
-const BASE = { wanted: "example.safetensors", category: "loras", baseUrl: "http://127.0.0.1:8188" };
+const BASE = {
+  wanted: "example.safetensors",
+  category: "loras",
+  baseUrl: "http://127.0.0.1:8188",
+  // Pin the unknown-bundle branch so a live panel on this machine cannot
+  // rewrite the refresh sentence (#2876).
+  panelBundle: {},
+};
 
 describe("a file inside the server's own models root is not called misplaced", () => {
   // The reporter's exact shape: one path, spelled two ways.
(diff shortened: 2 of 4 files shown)


artokun:
**Independent verification — this gets both sites, and closes the defect class.**

I had audited comfyui-mcp#2876 before this PR existed and found the bad remedy in **two** places, 27 lines apart in the same function and phrased differently — `model-resolver.ts:2459` and `:2486`. A fix greping for one sentence would have found one and left the other, which is this codebase's most persistent defect family.

On this branch:

```
  occurrences of `install_comfyui (action:"refresh_nodes")`: 0    (was 2 on main)
```

Both gone, replaced with `panel_refresh_nodes` — the name that actually exists.

**Re-running a codebase-wide sweep** for the same shape (every `<tool> (action:"<x>")` recommendation checked against the tool's own declaration):

```
  tools discovered (control):                 38
  distinct tool(action) pairs referenced:    189      (was 190)
  real mismatches:                              0      (was 1)
```

The four remaining hits are all in `src/tools/vocabulary.ts`, inside doc comments that are *about* invalid tool references — the file's own worked examples, including one literally labelled `// NOT action:"list_skills"`. Those are correct as they stand.

**So as of this branch, the class is empty.**

One thing worth calling out because it exceeds the report: `minPanelVersionForCmd("refresh_nodes")` at :2449 gates the recommendation on whether the live bundle can actually run the command, and the fallback names a path that works when it cannot —

> …cannot execute `panel_refresh_nodes` yet. After the tab loads the installed pack, call `panel_refresh_nodes`, or restart ComfyUI…

That is the harder half of the reporter's request ("name `panel_refresh_nodes` only when the live bundle can execute it"), and it means the remedy stays executable in the stale-bundle condition they were actually in — rather than trading an impossible remedy for one that is merely unavailable.

PR #2930, opened 2026-09-10
fix(node-dev): report search matches relative to custom_nodes/ so read can open them

### LINKED ISSUE #2921: node_pack read cannot open a file returned by node_pack search
`node_pack` search returned `./deno-custom-nodes/web/js/deno_local_llm_refiner.js`, but a follow-up `node_pack` read with the matching pack/path failed with `NODE_DEV_ERROR: File not found under custom_nodes/: "web/js/deno_local_llm_refiner.js".` This blocks inspection of the discovered source file and forced a local code-path workaround.

### Steps to reproduce
1. Call `node_pack` action `search`, pack `deno-custom-nodes`, query `deno_local_llm_state`.
2. Take its returned path `web/js/deno_local_llm_refiner.js`.
3. Call `node_pack` action `read`, pack `deno-custom-nodes`, path `web/js/deno_local_llm_refiner.js`.

### Exact error
`NODE_DEV_ERROR: File not found under custom_nodes/: "web/js/deno_local_llm_refiner.js".`

### Environment
Linux; ComfyUI 0.34.6; NVIDIA RTX 5090 31 GB; comfyui-mcp 0.52.202; panel 0.15.182.

---
Filed from the ComfyUI panel by **codex** · model `gpt-5.6-terra` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

wangzhengzhuo05:
**What**

`node_pack` action `search` reported every match path relative to the search's own `path`
walk root, while `read` / `write` / `list_files` resolve their `path` argument against the
`custom_nodes/` base root. So a path returned by `search` could not be fed back into
`read`: it failed with `NODE_DEV_ERROR: File not found under custom_nodes/: "<path>"`.

Reproduction from the issue: search with `path` set to a pack folder returns
`web/js/<file>.js`, and the matching `read` looks for `custom_nodes/web/js/<file>.js`.
The default `path: "."` happened to work only because the walk root *is* the base root.

**How**

- In `searchNodePacks`, compute the base root and the walk root's offset from it:

  ```ts
  const baseRoot = customNodesRoot(resolvedBase);
  const relPrefix = relative(baseRoot, searchDir).split(/[\\/]/).join("/");
  ```

- Pass `relPrefix` into both engines and prepend it to the reported `file`:
  - `searchBuiltin`: `file: relPrefix ? `${relPrefix}/${rel}` : rel`
  - `searchWithRipgrep`: `file: relPrefix ? `${relPrefix}/${m[1]}` : m[1]`
- Behavior with the default `path: "."` is unchanged (the prefix is empty), and the
  `path`/`glob` filtering still matches against the search-dir-relative path so glob
  semantics are untouched.

**Tests**

- New builtin regression test: create `Pack/nodes.py`, search `{ query: "FooNode", path: "Pack" }`,
  assert the reported file is `Pack/nodes.py`, then feed that exact string to `readNodeFile`
  and assert the content comes back — the full round trip from the issue.
- New ripgrep-path regression test asserting the same root-relative result through the
  mocked ripgrep seam.
- `npx vitest run src/__tests__/services/node-dev.test.ts` → 84 passed
- `npx tsc --noEmit` → clean

**Mutation check**

Reverting only the two `file:` expressions to their pre-fix form makes both new tests fail
(2 failed, 82 skipped); restoring the fix makes them pass. The new tests genuinely guard
the change.

Fixes #2921

### DIFF
--- a/src/__tests__/services/node-dev.test.ts
+++ b/src/__tests__/services/node-dev.test.ts
@@ -315,6 +315,38 @@ describe("searchNodePacks", () => {
     expect(rgCallsSpy[0].args[rgCallsSpy[0].args.indexOf("--regexp") + 1]).toBe("hit");
   });
 
+  it("#2921: a match found inside a pack is returned root-relative so read can open it", () => {
+    const pack = join(customNodes, "Pack");
+    mkdirSync(pack, { recursive: true });
+    writeFileSync(join(pack, "nodes.py"), "class FooNode:\n    pass\n");
+    const { deps } = makeDeps(); // builtin engine
+    const res = searchNodePacks({ query: "FooNode", path: "Pack" }, deps);
+    expect(res.matches.length).toBe(1);
+    // Root-relative (custom_nodes/), NOT pack-relative — this is what read expects.
+    expect(res.matches[0].file).toBe("Pack/nodes.py");
+    // The exact round trip from the issue: the returned path must be readable.
+    const read = readNodeFile({ path: res.matches[0].file }, deps);
+    expect(read.content).toContain("FooNode");
+  });
+
+  it("#2921: ripgrep matches inside a pack are returned root-relative", () => {
+    const pack = join(customNodes, "Pack");
+    mkdirSync(pack, { recursive: true });
+    writeFileSync(join(pack, "nodes.py"), "hit here\n");
+    const rgCallsSpy: { args: string[]; cwd: string }[] = [];
+    const { deps } = makeDeps({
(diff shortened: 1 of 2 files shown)


wangzhengzhuo05:
Closing this to keep the tracker clean \u2014 it has been sitting without review for several days. This is no reflection on the change itself: the fix and its local verification still stand. If the issue is still present and you would like the patch revisited, say the word and I will rebase it onto current main and re-open. Thanks for your time.

Issue #1470, opened 2026-08-12
install_custom_node nightly fails for direct Git URL with incompatible checkout flags


artokun:
Installing a custom node from a direct GitHub URL with version:"nightly" cloned successfully, then failed and removed the clone. Exact Git error: `fatal: '--detach' cannot be used with '-b/-B/--orphan'`. Repro: call install_custom_node action install with URL `https://github.com/numz/ComfyUI-SeedVR2_VideoUpscaler` and version `nightly`. Workaround that succeeded: repeat the same install with ref:"main" instead of version:"nightly". Expected: for a direct Git URL, nightly should resolve to the repository default branch/HEAD (or use a valid checkout command), not combine detached checkout with branch creation flags. Environment: Windows 11, ComfyUI 0.31.1, comfyui-mcp 0.51.15, panel 0.11.44, Python 3.13.

artokun:
Recurrence on a newer build: comfyui-mcp 0.52.76 with comfyui-mcp-panel 0.15.58 on Windows 11, ComfyUI 0.33.3, and Python 3.13. A direct Git URL with `version: "nightly"` again clones and then runs `git checkout --detach --end-of-options nightly`, failing with `fatal: git checkout: --detach does not take a path argument 'nightly'`.

In the same install flow, retrying the existing clone without a version printed `Already exists` plus `ERROR: An error occurred while installing` from comfy-cli, while the outer result still reported it as installed via official comfy-cli. The reporter is behind the current releases (MCP 0.52.89; panel 0.15.69).

artokun:
Claimed by Codex swarm at 2026-08-24T13:21:12Z. Prior fix PR #1513 is merged and its remote branch is gone; the only matching local branch is stale (last touched 2026-08-13) with no worktree. Because this issue was re-reported against MCP 0.52.76, I will revalidate the nightly Git ref/install flow on current main v0.52.91 and either drive a focused fix or close it with current evidence. This claim is active; please choose another issue.

artokun:
Completed and released in v0.52.92. PR #2195 merged as 1f0edfe0762d4ddfbba030dfb390597ce3030077; release PR #2197 merged and tag published at https://github.com/artokun/comfyui-mcp/releases/tag/v0.52.92. The fix makes comfy-cli nightly ref probing fail closed on unknown Git states, resolves only exact origin/nightly, and prevents false-success HEAD fallback. Release CI, npm, packs, and MCP Registry workflows are green.

PR #2936, opened 2026-09-10
fix(node-management): detach to canonical remote ref so remote-only branches install

### LINKED ISSUE #2919: install_custom_node: `git checkout --detach <ref>` fails for a remote-only branch ("'--detach' cannot be used with '-b/-B/--orphan'"), and the clone is rolled back

`install_custom_node` (action:"install", source:"git", ref:"<branch>") clones the repo, fetches, then runs:

```
git -C <nodeDir> checkout --detach <ref>
```

When `<ref>` names a branch that exists only as a **remote-tracking** branch (i.e. it is not the clone's default HEAD, so no local branch of that name was created), git's DWIM "create a local branch from origin/<name>" kicks in, which conflicts with `--detach`:

```
fatal: '--detach' cannot be used with '-b/-B/--orphan'
```

The install then errors and the cloned directory is rolled back/removed, so the user ends up with nothing installed and has to `git clone -b <branch>` by hand.

Source: `dist/services/node-management.js` (~line 2214 in 0.52.202), the `execFileSync("git", ["-C", nodeDir, "checkout", "--detach", checkoutRef], …)` call.

### Steps to reproduce

```
install_custom_node action:"install"
  id:"https://github.com/DrBearJew/ComfyUI-SolAttn_triton"
  source:"git"
  ref:"main"
```

(that repo's default branch is `h3-community-r01`, so `main` is remote-only after clone)

Reproduced standalone:

```
$ git clone -q https://github.com/DrBearJew/ComfyUI-SolAttn_triton repo && cd repo
$ git fetch -q --all --tags
$ git checkout --detach main
fatal: '--detach' cannot be used with '-b/-B/--orphan'
```

### Exact error (scrubbed)

```
{"error":"NODE_MANAGEMENT_ERROR","message":"Cloned \"https://github.com/DrBearJew/ComfyUI-SolAttn_triton\" but could not check out \"main\": Failed to check out git ref \"main\" for custom node \"https://github.com/DrBearJew/ComfyUI-SolAttn_triton\": Command failed: git -C ~\\AppData\\Local\\Comfy-Desktop\\ComfyUI-Installs\\ComfyUI\\ComfyUI\\custom_nodes\\ComfyUI-SolAttn_triton checkout --detach main\nfatal: '--detach' cannot be used with '-b/-B/--orphan'\n"}
```

### Fix

applied locally: yes (patched the installed `dist/services/node-management.js`)  ·  upstream-only: no

Resolve the ref to a concrete commit before detaching — `checkout --detach <sha>` never triggers DWIM. Try the local name first, then `refs/remotes/origin/<name>`:

```diff
+        // A ref that exists only as a remote-tracking branch (e.g. "main" when the
+        // clone's default HEAD is another branch) makes bare `checkout --detach <name>`
+        // trip git's DWIM branch creation and fail with
+        // "'--detach' cannot be used with '-b/-B/--orphan'". Resolve to a concrete
+        // commit first, trying the local name then origin/<name>.
+        let detachTarget = checkoutRef;
+        for (const candidate of [`${checkoutRef}^{commit}`, `refs/remotes/origin/${checkoutRef}^{commit}`]) {
+            try {
+                const sha = execFileSync("git", ["-C", nodeDir, "rev-parse", "--verify", "--quiet", candidate], {
+                    cwd: comfyuiBase,
+                    encoding: "utf-8",
+                    timeout: GIT_CLONE_TIMEOUT,
+                    env: nonInteractiveGitEnv(),
+                }).toString().trim();
+                if (sha) {
+                    detachTarget = sha;
+                    break;
+                }
+            }
+            catch {
+                // candidate does not resolve; try the next one
+            }
+        }
-        execFileSync("git", ["-C", nodeDir, "checkout", "--detach", checkoutRef], {
+        execFileSync("git", ["-C", nodeDir, "checkout", "--detach", detachTarget], {
             cwd: comfyuiBase,
             encoding: "utf-8",
             timeout: GIT_CLONE_TIMEOUT,
             env: nonInteractiveGitEnv(),
         });
```

Verification: the resolution step was verified standalone against the repro repo — `git rev-parse --verify --quiet "main^{commit}"` / `refs/remotes/origin/main^{commit}` yields `842c4eaa7d91dbaef3fee3ccdbf36a39521e82fc`, and `checkout --detach <sha>` does not hit the DWIM path. End-to-end re-run of the tool was NOT possible in this session because the orchestrator process holds the old `dist` and only reloads on a full process restart.

(Unrelated aside from the repro: that particular repo contains both `README.md` and `readme.md`, which collide on a case-insensitive filesystem and leave a dirty worktree after clone on Windows. That is upstream's problem, not this bug, but it means an install of that repo may also want a `-f`/clean step.)

### Environment

OS: Windows 11 Pro 10.0.26200 · ComfyUI 0.34.6 (Desktop, local) · GPU: NVIDIA RTX 5070 12 GB · CUDA 13.0 · torch 2.12.1 · python 3.13 · comfyui-mcp 0.52.202 · panel 0.15.182

---
Filed from the ComfyUI panel by **claude** · model `claude-opus-5` · effort `high` — stamped by comfyui-mcp from the panel's provider/model selection, not self-reported.

Related (same root cause): #1470

wangzhengzhuo05:
## Summary

Fixes #2919. `install_custom_node` (action:install, source:git) with a `ref` that names a branch existing only as a **remote-tracking** branch (i.e. it is not the clone's default HEAD, so no local branch of that name exists) fails because:

```
git checkout --detach main
fatal: '--detach' cannot be used with '-b/-B/--orphan'
```

git's DWIM "create a local branch from origin/<name>" kicks in for the bare name and conflicts with `--detach`. The install then errors and rolls back the clone.

## What

In `runGitCheckout` (`src/services/node-management.ts`), after `git fetch --all --tags` and before the detach, a bare branch/tag name on the **non-version path** is canonicalized when it does not exist locally:

- If `refs/heads/<name>` or `refs/tags/<name>` exists locally (via `gitShowRefExists`) → keep the bare name (no DWIM risk).
- Else if `refs/remotes/origin/<name>` exists (via `gitRemoteRefFor`) → use that canonical fully-qualified ref path. `git checkout --detach refs/remotes/origin/main` never triggers DWIM.
- Else → keep the bare name; a genuinely missing ref fails with the normal git error (behavior unchanged).

The `#1470` version-derived-channel branch (`refFromVersion === true && isGitHeadChannel`) is untouched. A probe failure (`GitRefProbeError`) is now rethrown instead of being misreported as a checkout error.

## Why

The repository's clone default HEAD can differ from the requested branch (e.g. the repro repo in #2919 defaults to `h3-community-r01`, so `main` is remote-only after clone). The user then gets nothing installed and has to `git clone -b main` by hand.

## Tests

- New `#2919` test: `ref: "main"` with a remote-only `main` asserts the checkout uses `["-C", <dir>, "checkout", "--detach", "refs/remotes/origin/main"]`.
- The existing explicit-tag test (`checkout --detach "v1.2.3"`) still passes unchanged, proving locally-present refs keep the bare name.
- Two cm-cli tests updated only to account for the new `show-ref` probe call (checkout assertions preserved in intent).

## Verification

- `npx vitest run src/__tests__/services/node-management.test.ts` → **244 passed** (243 baseline + 1 new).
- `npm run build` (tsc) → clean.
- **Mutation check**: with the canonicalization neutralized, the new `#2919` test fails (checkout uses the bare `main`); restoring the fix makes it pass — the regression test genuinely guards the remote-only-branch path.

### DIFF
--- a/src/__tests__/services/node-management.test.ts
+++ b/src/__tests__/services/node-management.test.ts
@@ -2059,6 +2059,48 @@ describe("node-management service", () => {
       ]);
     });
 
+    it("#2919 detaches to the canonical remote ref for a remote-only explicit branch", async () => {
+      stubFetch({ installedBody: {} });
+      let cloned = false;
+      mockedExists.mockImplementation((p: unknown) => {
+        const s = String(p);
+        if (s.includes("requirements.txt") || s.includes("install.py")) {
+          return false;
+        }
+        if (s.includes(".venv") || s.includes("cm-cli.py")) return false;
+        if (s.includes(NODE_DIR_UTILS)) return cloned;
+        return false;
+      });
+      const missingGitRef = () =>
+        Object.assign(new Error("fatal: 'refs/heads/main' - not a valid ref"), {
+          status: 1,
+        });
+      mockedExec.mockImplementation(((bin: string, args: string[]) => {
+        if (bin === "git" && args[0] === "clone") cloned = true;
+        if (bin === "git" && args[2] === "show-ref") throw missingGitRef();
+        if (bin === "git" && args[2] === "for-each-ref")
+          return "refs/remotes/origin/main\n";
+        return "";
+      }) as never);
+
+      const res = await installCustomNode({
+        id: "https://github.com/teskor-hub/comfyui-teskors-utils",
+        ref: "main",
+      });
+
+      expect(res.mechanism).toBe("git-clone");
+      const checkoutCall = mockedExec.mock.calls.find(
+        (c) => c[0] === "git" && (c[1] as string[])[2] === "checkout",
+      );
+      expect(checkoutCall?.[1]).toEqual([
+        "-C",
+        NODE_DIR_UTILS,
+        "checkout",
+        "--detach",
+        "refs/remotes/origin/main",
+      ]);
+    });
+
     it("cleans up a clone when the version-derived nightly probe is unknown", async () => {
       stubFetch({ installedBody: {} });
       let cloned = false;
@@ -2433,7 +2475,7 @@ describe("node-management service", () => {
       });
 
       expect(res.mechanism).toBe("comfy-cli");
-      expect(mockedExec).toHaveBeenCalledTimes(3);
+      expect(mockedExec).toHaveBeenCalledTimes(4);
       expect(mockedExec.mock.calls[0][1]).toEqual([
         "--json",
         "--workspace",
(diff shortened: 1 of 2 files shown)


wangzhengzhuo05:
Closing this to keep the tracker clean \u2014 it has been sitting without review for several days. This is no reflection on the change itself: the fix and its local verification still stand. If the issue is still present and you would like the patch revisited, say the word and I will rebase it onto current main and re-open. Thanks for your time.

PR #2940, opened 2026-09-12
fix(node-management): preserve params.base when dispatching a Manager install_model task

### LINKED ISSUE #2922: download_model omits base for legacy-UI Manager install_model payload
On a remote ComfyUI running pip ComfyUI-Manager v4 in --enable-manager-legacy-ui mode, download_model(action=download) returned HTTP 500 even for an exact model-list whitelist URL. comfyui-mcp constructed a base hint in model-resolver, but installModelViaManager rebuilt taskParams and dropped params.base. The legacy Manager whitelist checker unconditionally reads item['base'], causing KeyError: 'base'.

### Steps to reproduce
1. Run ComfyUI 0.33.0 with pip comfyui_manager and --enable-manager-legacy-ui.
2. Connect comfyui-mcp remotely.
3. Call download_model for the whitelisted SDXL IP-Adapter URL and target_subfolder ipadapter.
4. Observe /v2/manager/queue/install_model HTTP 500 and Manager traceback KeyError: 'base'.

### Exact error (scrubbed)
comfyui_manager/legacy/manager_server.py check_whitelist_for_model: x['base'] == item['base']; KeyError: 'base'.

### Fix
Applied locally: yes. Upstream-only: no. Preserve params.base when installModelViaManager builds taskParams, and provide the SDXL family hint at the URL/filename resolution call site. After reload, the same download succeeded and produced an 847,517,512-byte model file listed by ComfyUI.

Patch:
```diff
--- a/dist/services/model-resolver.js
+++ b/dist/services/model-resolver.js
@@
             type: managerType,
             save_path: managerSavePath,
+            base: /sdxl/i.test(resolvedFilename) ? "SDXL" : "SD1.5",
             trayCategory: modelType,
--- a/dist/services/node-management.js
+++ b/dist/services/node-management.js
@@
         type: params.type,
         save_path,
+        ...(params.base ? { base: params.base } : {}),
     };
```
A stronger upstream implementation should resolve base from Manager model-list metadata when matching a whitelisted URL instead of relying only on filename inference.

### Environment
Linux; ComfyUI 0.33.0; frontend 1.49.6; Python 3.13.7; PyTorch 2.11.0+cu130; NVIDIA GeForce RTX 5060 Ti 15.5 GB; comfyui-mcp 0.52.65; panel 0.15.32.

wangzhengzhuo05:
## What

`download_model(action=download)` returns HTTP 500 for an exact model-list whitelisted URL on a remote ComfyUI running ComfyUI-Manager v4 in `--enable-manager-legacy-ui` mode.

## Why

`installModelViaManager` rebuilds `taskParams` from scratch and dropped the `base` hint that model resolution had already computed. The legacy Manager UI's whitelist checker reads the key unconditionally (`check_whitelist_for_model`: `x['base'] == item['base']`), so the queued task 500s with `KeyError: 'base'` before the download is ever attempted.

## How

- `InstallModelParams` gains an optional `base` field with the reason documented inline.
- `installModelViaManager` forwards it into the queued task params, and only when it is present and non-blank, so no `base: undefined` reaches Manager.
- the `installModelViaManager` call site in `model-resolver.ts` supplies the family hint from the resolved filename (`SDXL` when it matches `sdxl`, otherwise `SD1.5`).

Resolving `base` from Manager's own model-list metadata instead of filename inference would be a stronger follow-up; this change keeps the task from failing in the meantime.

## Tests

New `src/__tests__/services/manager-install-model-base.test.ts`:

- forwards `base` into the queued install-model task params when it is set;
- sends no `base` key at all when the hint is absent.

```
npx vitest run src/__tests__/services/manager-install-model-base.test.ts \
  src/__tests__/services/manager-install-model-cause.test.ts \
  src/__tests__/services/manager-dialect-cache.test.ts
-> Test Files 3 passed (3), Tests 47 passed (47)

npx tsc --noEmit -p tsconfig.json
-> clean
```

Mutation check: deleting the new `...(params.base && ...)` spread makes the forwarding test fail (`1 failed | 1 passed`); restoring it passes again, so the test genuinely guards this fix.

Fixes #2922

### DIFF
--- a/src/__tests__/services/manager-install-model-base.test.ts
+++ b/src/__tests__/services/manager-install-model-base.test.ts
@@ -1,0 +1,153 @@
+// #2922 — the legacy Manager UI's whitelist checker reads item['base']
+// unconditionally, so a queued install_model task that omits `base` 500s with
+// `KeyError: 'base'` before the download is attempted. installModelViaManager
+// must therefore preserve a provided `base` hint in the dispatched task params,
+// while never sending the key at all when the hint is absent/blank (Manager
+// validates keys it receives, so `base: undefined` must not reach it).
+
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+// Kept mutable to model the runtime target the panel applies. vi.hoisted makes
+// it available to the mock factory.
+const target = vi.hoisted(() => ({ base: "http://127.0.0.1:8188", generation: 0 }));
+
+vi.mock("../../config.js", () => {
+  const config = {
+    comfyuiPath: "/fake/comfy",
+    resolvedPort: 8188,
+    comfyuiHost: "127.0.0.1",
+    comfyuiSsl: false,
+    githubToken: undefined as string | undefined,
+  };
+  return {
+    config,
+    getComfyUIBaseUrl: () => target.base,
+    getComfyuiTargetGeneration: () => target.generation,
+    getComfyUIAuthHeaders: () => ({}),
+    isLoopbackHost: (host?: string) => host === "127.0.0.1" || host === "localhost",
+    isRemoteMode: () => false,
+  };
+});
+
+// node-management pulls in comfy-cli → workspace-env, which calls
+// promisify(execFile) at module load; keep the subprocess surface inert.
+vi.mock("node:child_process", () => ({
+  execFile: vi.fn(),
+  execFileSync: vi.fn(),
+  spawnSync: vi.fn(() => ({ status: 0, stdout: "{}", stderr: "" })),
+}));
+vi.mock("node:fs", async (importOriginal) => ({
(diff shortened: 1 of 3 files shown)


wangzhengzhuo05:
Closing this out.

It has been waiting 6 days without any maintainer review, and we are winding down this contribution effort — a stale PR sitting in the queue costs maintainers attention without giving them a decision to make, so it is better closed than left open indefinitely.

If the direction is useful, the branch is intact on the fork (`fix/manager-base-taskparam`, +164/−0 across 3 file(s)) and can be picked up as-is.

What it contains: *fix(node-management): preserve params.base when dispatching a Manager install_model task* — `download_model(action=download)` returns HTTP 500 for an exact model-list whitelisted URL on a remote ComfyUI running ComfyUI-Manager v4 in `--enable-manager-legacy-ui` mode.

It was verified locally before opening (target test suite green, plus a mutation check that reverts only the fix and confirms the new regression test fails). Thanks for the time.

